PR #399, opened 2018-01-19
l10n: Init Vietnamese translation for gcoderef

vnwildman:
Signed-off-by: Trần Ngọc Quân <vnwildman@gmail.com>


jethornton:
For some reason this is being used even for US English...

PR #401, opened 2018-01-21
fix null pointer dereference for bidirectional pins on smart serial s…

rene-dev:
…laves.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/sserial.c
+++ b/src/hal/drivers/mesa-hostmot2/sserial.c
@@ -1872,7 +1872,7 @@ void hm2_sserial_write_pins(hostmot2_t *hm2, hm2_sserial_instance_t *inst){
                         break;
                     case LBP_BOOLEAN:
                         buff = 0;
-                        if (*pin->boolean ^ *pin->invert){
+                        if (*pin->boolean ^ ((conf->DataDir == LBP_OUT)?(*pin->invert):0)){
                             buff = (~0ull >> (64 - conf->DataLength));
                         }
                         break;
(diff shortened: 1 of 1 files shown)


src/hal/drivers/mesa-hostmot2/sserial.c
```diff
@@ -2016,7 +2016,9 @@ int hm2_sserial_read_pins(hm2_sserial_remote_t *chan){
                 break;
             case LBP_BOOLEAN:
                 *pin->boolean = (buff != 0);
-                *pin->boolean2 = (buff == 0);
+                if(conf->DataDir == LBP_IN){
+                    *pin->boolean2 = (buff == 0);
+                }
```
SebKuzminsky: Do you want to specify a default value here for when DataDir is not LBP_IN, like you did in the hunk above?

SebKuzminsky:
The pin direction handling in 2.7's sserial looks the same as master's before this PR.  Does 2.7 have this bug as well?  If so that's where the fix should go, then we'll merge 2.7 into master to get the fix there too.

src/hal/drivers/mesa-hostmot2/sserial.c
rene-dev: that is not required, as boolean2 is not created in the other case: https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/drivers/mesa-hostmot2/sserial.c#L1310-L1316
for the other direction it is required, because in case of outputs its the invert pin.
the bug was introduces here: https://github.com/LinuxCNC/linuxcnc/commit/31029ef2166d1f52745baa5b458358dd72fe91d4

SebKuzminsky: Cool.

SebKuzminsky:
@rene-dev reports on IRC that 2.7 doesn't have this bug.
http://tom-itx.no-ip.biz:81/~tom-itx/irc/logs/%23linuxcnc-devel/2018-01-21.html

PR #406, opened 2018-02-02
don't bundle yapps2

jepler:
.. there seems to be no reason to do this.


jepler:
@SebKuzminsky I see that I left 80% of the task unfinished.  feel free to just squash my commit into yours and merge.  thanks!

SebKuzminsky:
I merged a version of this branch into master at be22b73d4b5772.

PR #415, opened 2018-03-04
moved saveToolTable form iocontrol to tool_parse.cc, removed old tool…

rene-dev:
…table file syntax, added pocketno to tooltable struct so non-random changers report the correct pocket number. should fix #500


rene-dev:
Still work in progress....

maknoll:
I am currently implementing a rack toolchanger wich is managed with pockets and M6 remap.
I fell over that problem and am glad that you already have a fix for it. I hope this gets merged quickly.

Just wanted to comment that there are people out here with ATCs on LinuxCNC machines and this is an actual issue.
With a modern CAD-CAM workflow it is unreasonable to do toolnumber==pocket, random ATC or not.

rene-dev:
It needs testing. the one bug in this PR is, that remapped T do not work for some reason.
but IMHO a T code should never be remapped, as prepare should happen in the background, and not block.

maknoll:
I have applied this PR to master, build it and it now runs in our mill. It indeed fixes all issues and ATC now finally runs on Pockets. I will report if i suffer any additional bugs.

I agree that T remap is unreasonable. Would rather have the pocket issue fixed and open this one instead.

rene-dev:
you are using a M6 remap only?

maknoll:
Yes, no toolchange preparation whatsoever. Just move to tool location and pick up.

rene-dev:
ok, please check if all numbers match, including while loading-only and unloading-only a tool, by changing to and from tool 0.
Also check if the parameters, are right: http://linuxcnc.org/docs/html/gcode/overview.html#sub:system-parameters
for selected and current tool and pocket.
and, if the stuff passed to python is correct. IMHO you dont need those, as they are available as parameters...

andypugh:
On 26 May 2018 at 18:46, Rene Hopf <notifications@github.com> wrote:

> It needs testing. the one bug in this PR is, that remapped T do not work
> for some reason.
> but IMHO a T code should never be remapped,
>

Remapping T to use the normal T-only convention for a lathe tool-change
seems like one of the primary uses of remapping to me.


-- 
atp

rene-dev:
Ah, thanks andy, that makes sense. I will look into why it doesn't work.

rene-dev:
I created a new pull request to 2.7: https://github.com/LinuxCNC/linuxcnc/pull/528

PR #416, opened 2018-03-07
fixing configure.ac for python3

rene-dev:


rene-dev:
work in progress. currently fails to run halcompile. to test it:
```
sudo apt-get install python3-pip
pip3 install yapps
./autogen.sh && ./configure --with-python=python3 --with-boost-python=boost_python-py35 && make clean && make
```

PR #418, opened 2018-03-07
Configure python3

jepler:
After this change, the following configure invocation works for me on debian stretch:
~~~~
$ ./configure --with-python=python3 --with-boost-python=boost_python-py35 --enable-non-distributable=yes
~~~~
(`--enable-non-distributable` is needed on my usual development system for reasons unrelated to python3)

However, as expected, actually building LinuxCNC after this `./configure` invocation fails for many reasons.

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1562,7 +1562,7 @@ if test "$PYTHON" = "none"; then
 fi
 
 AC_MSG_CHECKING([python version])
-if ! $PYTHON -c 'import sys;raise SystemExit,sys.hexversion<0x2070000'
+if ! $PYTHON -c 'import sys;raise SystemExit(sys.hexversion<0x2070000)'
 then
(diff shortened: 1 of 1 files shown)


jepler:
FWIW I do not plan to work further on this at this time.

rene-dev:
Any reason this is not merged? as far as Im aware its 2 and 3 compatible.

zultron:
Looks like this might be superceded by #465.  @jepler?

zultron:
Sorry, #465 was submitted by @TurBoss, and has been subsequently closed.

rene-dev:
yes, I have seen that. closed, but not merged. why?

TurBoss:
sorry was having marge conflicts so I wanted to start a new PR but forgot

here is my branch

https://github.com/TurBoss/linuxcnc/blob/configure-python3/src/configure.ac

TurBoss:
sorry completely forgot to open a new one

thanks

PR #420, opened 2018-03-16
Bspi updates

sleepybishop:
been working on bspi driver, found a few bugs.

for context:
https://forum.linuxcnc.org/24-hal-components/34049-bspi-problems#106773

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/bspi.c
+++ b/src/hal/drivers/mesa-hostmot2/bspi.c
@@ -199,7 +199,7 @@ int hm2_bspi_write_chan(char* name, int chan, rtapi_u32 val)
 
 EXPORT_SYMBOL_GPL(hm2_bspi_setup_chan);
 int hm2_bspi_setup_chan(char *name, int chan, int cs, int bits, float mhz,
-                        int delay, int cpol, int cpha, int clear, int echo)
+                        int delay, int cpol, int cpha, int clear, int echo, int samplelate)
 {
     hostmot2_t *hm2;
     rtapi_u32 buff = 0;
@@ -239,6 +239,7 @@ int hm2_bspi_setup_chan(char *name, int chan, int cs, int bits, float mhz,
 
     buff = (echo != 0) << 31
         |  (clear != 0) << 30
+        |  (samplelate != 0) << 29
         | ((delay <= 0)? 0x10 : (rtapi_u32)((delay*board_mhz/1000.0)-1) & 0x1f) << 24
         | (cs & 0xF) << 16
         | (((rtapi_u16)(board_mhz / (mhz * 2) - 1) & 0xFF)) << 8
--- a/src/hal/drivers/mesa-hostmot2/bspi.c
+++ b/src/hal/drivers/mesa-hostmot2/bspi.c
@@ -177,7 +177,7 @@ EXPORT_SYMBOL_GPL(hm2_bspi_write_chan);
 int hm2_bspi_write_chan(char* name, int chan, rtapi_u32 val)
 {
     hostmot2_t *hm2;
-    rtapi_u32 buff;
+    rtapi_u32 buff = val;
     int i, r;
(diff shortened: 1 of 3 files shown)


andypugh:
Why reduce the maximum number of bspi interfaces?

sleepybishop:
not reducing, according to regmap the base addresses for bspi modules are
0x5500 BSPI FIFO port 0
0x5540 BSPI FIFO port 1
0x5580 BSPI FIFO port 2
0x55C0 BSPI FIFO port 3

and then
0x5600 BSPI channel descriptor setup registor 0

so theres only room for 4 bspi interfaces

unless we do a big refactor and resuse the dbspi address space, as far as i cant tell the only difference is one parameter in hostmot2.vhd (gatedcs= true) for dbspi. if we set that via other means then there is enough address space for 4 more bspi instances. but that would be a pretty big change

andypugh:
> so theres only room for 4 bspi interfaces

4 interfaces per FPGA card. But a system can have many FPGA cards.

sleepybishop:
right, ok ill revert that change.

jepler:
Also the memory map file is really just a guideline for how things typically are.  A firmware could move those to any other address and have room for more BSPI instances.  The hostmot2 driver will read the correct addresses from the IDROM, and does not use the addresses and ranges shown in the regmap.txt.

sleepybishop:
i removed all commits that deal with addresses of the registers, i have more work to do to sort all that out.

src/hal/drivers/mesa-hostmot2/bspi.c
```diff
@@ -197,9 +197,21 @@ int hm2_bspi_write_chan(char* name, int chan, rtapi_u32 val)
...
+    rtapi_u32 buff = 0; // data is dont care when writing to count register
+    int r;
+    r = hm2->llio->write(hm2->llio, hm2->bspi.instance[i].count_addr, &buff, sizeof(rtapi_u32));
+    if (r < 0) {
+        HM2_ERR("BSPI: hm2->llio->write failure %s\n", hm2->bspi.instance[i].name);
+    }
+
+    return r;
+}
+
```
jepler: This added function appears to be unused.  If that is correct, can you please remove it?

src/hal/drivers/mesa-hostmot2/bspi.c
```diff
@@ -197,9 +197,21 @@ int hm2_bspi_write_chan(char* name, int chan, rtapi_u32 val)
...
 EXPORT_SYMBOL_GPL(hm2_bspi_setup_chan);
 int hm2_bspi_setup_chan(char *name, int chan, int cs, int bits, float mhz,
-                        int delay, int cpol, int cpha, int clear, int echo)
+                        int delay, int cpol, int cpha, int clear, int echo, int samplelate)
```
jepler: you change the function signature here, but don't change the declaration or use sites until a later commit.  Please move to the appropriate commit.

Symptom: Compiling at this ref,
~~~~
hal/drivers/mesa-hostmot2/bspi.c:213:5: error: conflicting types for ‘hm2_bspi_setup_chan’
 int hm2_bspi_setup_chan(char *name, int chan, int cs, int bits, float mhz,
     ^~~~~~~~~~~~~~~~~~~
In file included from hal/drivers/mesa-hostmot2/bspi.c:25:0:
hal/drivers/mesa-hostmot2/hostmot2.h:1304:5: note: previous declaration of ‘hm2_bspi_setup_chan’ was here
 int hm2_bspi_setup_chan(char *name, int chan, int cs, int bits, float mhz,
     ^~~~~~~~~~~~~~~~~~~
Makefile:1005: recipe for target 'objects/rthal/drivers/mesa-hostmot2/bspi.o' failed
~~~~

sleepybishop:
@jepler, those two issues should be addressed now, thank you.

jepler:
@sleepybishop thank you for these fixes.

Issue #392, opened 2018-01-03
libnml should ask pkg-config about libtirpc location


l29ah:
As the internal glibc sunrpc implementation is deprecated long time ago and removed in the recent versions.
```
In file included from libnml/cms/cms.cc:37:0:
libnml/cms/cms_xup.hh:19:39: fatal error: rpc/rpc.h: No such file or directory
```
```
∞ pkg-config --cflags libtirpc       
-I/usr/include/tirpc
```

Petross404:
> ∞ pkg-config --cflags libtirpc       
-I/usr/include/tirpc

This works?

spike77453:
sunrpc seems to be removed from glibc >= 2.27 so linuxcnc doesn't compile anymore :(

jepler:
It appears that for master branch, it would be acceptable to change unconditionally to libtirpc, because the package libtirpc-dev is in ubuntu 12.04 and newer and debian 7.0 and newer.  However, a fix targeted to 2.7 must adapt to either situation, because 2.7 still targets ubuntu 10.04 which does not live package libtirpc-dev.

jepler:
.. I upated http://wiki.linuxcnc.org/cgi-bin/wiki.pl?MinimumSoftwareVersions with information about libtirpc-dev based on a bit of googling.  Feel free to update it (or just make a correction here) if I got it wrong.

Petross404:
> sunrpc seems to be removed from glibc >= 2.27 so linuxcnc doesn't compile anymore :(
I think it got removed one or two versions earlier ;)

awigen:
It was removed in 2011 glibc-2.14: https://sourceware.org/git/?p=glibc.git;a=commit;h=7b57bfe5988e476ea40934457dfd1c8a231e2391

rene-dev:
fixed by #587

PR #433, opened 2018-05-03
Compile against libtirpc

spike77453:
This is probably not the correct way to do it, but it fixes my build issues (#392).
It's probably a good idea to check for libtirpc or even include a configure flag (--with-libtirpc) but I currently can't wrap my head around linuxcnc's autotools setup.


jepler:
Thank you for offering a patch, though, I appreciate it, and it is going to help users who have the same problem as you're encountering.

You're right that we can't accept this PR as-is because it should be properly integrated into configure (and into debian dependency management stuff in debian/) but I will leave it open so that people who need a quick fix can get it.

awigen:
This is now superseeded by https://github.com/LinuxCNC/linuxcnc/pull/587 which fallsback to TI-RPC if libc does not provide RPC.

rene-dev:
fixed by #587

PR #442, opened 2018-06-15
Proposed fix for "FIXTHIS too verbose"

KimK:
I don't like the idea of having two different styles
for the G-code and M-code tables, but there are just
too many G-codes available to not have it be verbose.
So I ask the user to pick the blank cells for G-codes.

I could "reverse" the M-codes table too, if consistency
between them is thought to be important.

The tables are broken up to avoid trying to print
one table 100 lines long.

Signed-off-by: Kim Kirwan <Kim@KimKirwan.com>


SebKuzminsky:
That looks great, thanks!

andypugh:
71, 71.1, 72, 72.1 _will_ be allocated at some point, but at the moment there is an existing remap for those who choose to install it, so those need to still be remappable. 
Arguably even if/when they do get mainlined there is no reason to not add the remappable extras for those who prefer different behaviour.

PR #446, opened 2018-06-19
Replace `python-notify` with our own PyQt5 DBus system notification module 

KurtJacobson:
`python-notify` when used with Qt5 on Debian 9 causes gtk errors which prevent QtScreen from starting. This replaces that package with a pure Qt module that sends requests over DBus to the system notification server.

Depends on:
  * python-dbus
  * python-dbus.mainloop.pyqt5 (optional)

If `python-dbus.mainloop.pyqt5` is not installed any notification actions with callbacks will not work, but the notifications will still be shown.

Tested on:
  * Debian 9
  * Ubuntu 18.04


KurtJacobson:
For @c-morley!

Issue #451, opened 2018-06-26
mesa_7i65 comp causes kernel panic


pcw-mesa:
## Here are the steps I follow to reproduce the issue:

Run Linuxcnc with the halfile/hardware for mesa_7i65 component

## This is what I expected to happen:
Run normally

## This is what happened instead:
Causes a kernel panic about 1/3 time you launch or leave LinuxCNC
Driver seems to work properly if it launches OK

## It worked properly before this:
Not sure 2.8 also reportedly has this issue

## Information about my hardware and software:

LinucCNC 2.7.12 or LinuxCNC 2.7.14

  
  2.7.12 binary version is from linuxcnc.org
  2.7.14 binary version is built from source locally
 Hardware is 5I20 card + 7I65

andypugh:
Both PCW and myself think that patch works. I recommend pushing to 2.7. Or even 2.6 :-)

PR #456, opened 2018-07-05
tram.c: Fix pointer arithmetic thinko

jepler:
Because the type of tram_read_buffer is uint32_t, it's not
appropriate to perform arithmetic on it with the old read size
which is in bytes.

This didn't usually affect hm2 EXCEPT FOR bspi, because bspi
is the only user of the code that reaches hm2_allocate_tram_regions
with the old_tran_read_size being nonzero.

(The analysis is just the same for the case of the write buffers)

For those following along at home, I added the bad code
in cf4f68997553422c40dea22ad916b82ff0254a7f and all 2.7 and master
branches are likely affected.

Closes: #451


andypugh:
I am glad you found this, I probably wouldn't have noticed the problem. (I think pointer-arithmetic tries to be too clever). Presumably this was memsetting 4x as much memory as necessary, with fairly random consequences?

PR #458, opened 2018-07-08
2.7

KimK:
Docs: Correct G28,G30. Improve G92.

The G28 & G30 parameter count were still set to six,
from the old way of XYZABC, now XYZABCUVW so add counts.

G92 never mentioned parameters until near the end,
so I copied from G28/G30 to add to G92. Hopefully
made some other useful improvements while there too.


KimK:
Gaaah! Some branchy thing went wrong? I'll redo it.

andypugh:
On 8 July 2018 at 06:58, Kim Kirwan <notifications@github.com> wrote:

> Gaaah! Some branchy thing went wrong? I'll redo it.
>
Top Tip!

git push --dry-run will give you a log 123456..789abcd or similar, then
git log -p  23456..789abcd will show you exactly what you are about to
push.
Saves surprises :-)


-- 
atp

PR #459, opened 2018-07-08
Docs: Correct G28,G30. Improve G92. (2nd go)

KimK:
The G28 & G30 parameter count were still set to six,
from the old way of XYZABC, now XYZABCUVW so add counts.

G92 never mentioned parameters until near the end,
so I copied from G28/G30 to add to G92. Hopefully
made some other useful improvements while there too.

Signed-off-by: Kim Kirwan <kim@kimkirwan.com>


SebKuzminsky:
Thanks Kim!

PR #462, opened 2018-07-11
Some more Python Interface doc cleanup

KurtJacobson:
See commit messages.

### DIFF
--- a/docs/src/config/python-interface.txt
+++ b/docs/src/config/python-interface.txt
@@ -497,7 +497,9 @@ commanded output position.
 non-zero means limits are overridden.
 
 *units*:: '(returns float)' -
-units per mm, deg for linear, angular
+joint units per mm, or per degree for angular joints. +
+(joint units are the same as machine units, unless set otherwise
+by the configuration parameter [JOINT_n]UNITS)
 
 *velocity*:: '(returns float)' -
 current velocity.
--- a/docs/src/config/python-interface.txt
+++ b/docs/src/config/python-interface.txt
@@ -175,7 +175,7 @@ current feedrate override, 1.0 = 100%.
 currently loaded gcode filename with path.
 
 *flood*:: '(returns integer)' -
-Flood status, either 'FLOOD_OFF' or 'FLOOD_ON'.
+Flood status, either FLOOD_OFF or FLOOD_ON.
 
 *g5x_index*:: '(returns integer)' -
 **** replace with constants list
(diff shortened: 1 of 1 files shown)


c-morley:
The description says units per mm or degrees - but isn't it machine units (engineering units) ?

docs/src/config/python-interface.txt
```diff
@@ -471,7 +471,7 @@ parameter, reflects [JOINT_n]FERROR.
 non-zero means max hard limit exceeded.
 
 *max_position_limit*:: '(returns float)' -
-maximum limit (soft limit) for axis motion, in machine units.configuration
+maximum limit (soft limit) for axis motion, in machine units. configuration
```
SebKuzminsky: Should this say "for joint motion"?

docs/src/config/python-interface.txt
```diff
@@ -484,7 +484,7 @@ configuration parameter, reflects [JOINT_n]MIN_FERROR.
 non-zero means min hard limit exceeded.
 
 *min_position_limit*:: '(returns float)' -
-minimum limit (soft limit) for axis motion, in machine units.configuration
+minimum limit (soft limit) for axis motion, in machine units. configuration
```
SebKuzminsky: "for joint motion"?

KurtJacobson:
@c-morley The joint units will be the same as machine units for a trivkins machines, and probably many others, but it is not guaranteed to be the same as machine units. I don't fully understand it, but according to @rene-dev on IRC "non-triv machines can have different unis, like a robot. all joints are angular, but xyz abc is mm/deg"

c-morley:
@KurtJacobson: I agree that I am not up on non-triv at all and could be completely wrong but,
say for instance your machine is built completely in inches - It doesn't seem right to say the joints will be in mm/degree. I would think it's really joint units/degree, joint units being what ever is set in [JOINTS]. If it is actually mm then that would be the first time that I know of that linuxcnc used a specific unit internally.
But as I said this I am talking above my pay grade :) so could be absolutely missing something.

KurtJacobson:
@c-morley Now I understand your concern, I did not read your first comment properly! LinuxCNC does indeed use millimeter internally, and all other units are converted to millimeter. So, strange as it seems, the value of [JOINT]UNITS is the number of machine units that fit into one millimeter. For an inch machine the units value would be 0.03937in/mm, for a metric machine it would be 1mm/mm. For rotary axes the internal units is deg, so if you configure an axis in grad the units value would be 1.1111grad/deg etc. It took me quite some time to figure out was was going on, and that is what prompted me to try and clarify it in the docs.

c-morley:
@KurtJacobson  just revisiting your comments.
I wonder if we should add your description of how to calculate joint units - I find it non-intuitive.?

KurtJacobson:
@c-morley it probably would be good to include a description of how the joint units are calculated, but I doubt anybody even used them, so it may not be worth the effort.

KurtJacobson:
Thank you @c-morley!

PR #463, opened 2018-07-12
Re-wording of LinuxCNC User Introduction doc

ascurtis:
Proposed re-wording of 'LinuxCNC User Introduction' documentation page as per Forum discussion here: https://forum.linuxcnc.org/32-documents/34802-linuxcnc-is-poorly-documented?start=100#113774

Hopefully I've got the formatting right in the asciidoc!

Feel free to spin/fold/mutilate as you deem necessary.

Quick and dirty print of the page as it should appear once uploaded for reference:
[LinuxCNC User Introduction - AC rewrite.pdf](https://github.com/LinuxCNC/linuxcnc/files/2186832/LinuxCNC.User.Introduction.-.AC.rewrite.pdf)

Cheers, Andrew.

### DIFF
--- a/docs/src/user/user-intro.txt
+++ b/docs/src/user/user-intro.txt
@@ -4,180 +4,163 @@
 
 == How LinuxCNC Works
 
-When you start LinuxCNC from the CNC menu or by typing 'linuxcnc'
-in the terminal the LinuxCNC script starts the
-<<cha:starting-linuxcnc,Configuration Selector>>. When you pick a configuration
-LinuxCNC reads the <<cha:ini-configuration,INI file>> then loads the 
-<<cha:basic-hal-reference,HAL files>> in the order that they are listed in the
-INI file.
+LinuxCNC is a suite of highly-customisable applications for the control of a Computer Numerically
+Controlled (CNC) mills and lathes, 3D printers, robots, laser cutters, plasma cutters and other automated
+devices. It is capable of providing coordinated control of up to 9 axes of movement.
 
-When you create a configuration with the
-<<cha:stepconf-wizard,Stepper Configuration Wizard>> or
-<<cha:pncconf-wizard,Mesa Hardware Wizard>> or by picking a sample
-configuration from the <<cha:starting-linuxcnc,Configuration Selector>>
-LinuxCNC creates some directories and files in your user directory. In the
-following example the name of the configuration is 'My Lathe'.
+At its heart, LinuxCNC consists of several key components that are integrated together to form one
+complete system:
 
-* linuxcnc
-** configs
-*** My_Lathe
-**** My_Lathe.ini (this is read by the linuxcnc script)
-**** My_Lathe.hal (this hal file is loaded before the gui)
-**** post_gui.hal (this hal file is loaded after the gui)
-**** linuxcnc.var (stores the <<gcode:parameters,parameters>>)
-**** linuxcnc.var.bak (a backup parameter file)
-**** tool.tbl (the tool table file)
-** nc_files
-*** *.ngc (G code files)
+* a Graphical User Interface (GUI), which forms the basic interface between the operator, the software
+and the CNC machine itself;
+* the <<cha:hal-introduction,Hardware Abstraction Layer>> (HAL), which provides a method of linking all
+the various internal virtual signals generated and received by LinuxCNC with the outside world; and,
+* the high level controllers that coordinate the generation and execution of motion control of the CNC
+machine, namely the motion controller (EMCMOT), the discrete input/output controller (EMCIO) and the
+task executor (EMCTASK).
 
-There can be other directorys and files in a configuration but the above is
-usually the minimum files used.
-
-LinuxCNC can control machine tools, robots, or other automated devices. It can
-control servo motors, stepper motors, relays, and other devices related to
-motion control. LinuxCNC can control up to 9 axes in coordinated motion.
-
-There are five main components to the LinuxCNC software: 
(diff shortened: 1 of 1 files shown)


andypugh:
On a quick skim I think this is an improvement in clarity and elegance of prose. 
However I think that the explicit mention of Mesa cards to the exclusion of the others (Pico, General Mechatronics, Motenc, Servo-to-go) is a problem. A link to the "Supported Hardware" Page on the Wiki would be better.

ascurtis:
Hi Andy,

How about something along the lines of:

> While a stepper motor system under parallel port control is illustrated, a LinuxCNC system can also take advantage of a wide variety of dedicated hardware motion control interfaces, such as those offered by Mesa Electronics, General Mechatronics, Vital Systems and Servo-2-Go for increased speed and I/O capabilities.

Each of the above manufacturers listed in the above text could be a hyperlink off to their respective pages.

The Supported Hardware page on the Wiki appears to have a few dead links and '?'s printed here and there. Some of the entries for supported hardware even mention that the hardware actually isn't supported. Is the Wiki still the preferred place to steer people to for a current list of hardware in this case?

andypugh:
Pico are active on the forum and the mailing list. Leaving Jon out seems wrong. So I would be generic and link to somewhere. 

I don't _like_ that the list of supported hardware is on the Wiki. I would rather it was somewhere maintained and "official"

ascurtis:
Not fussed either way. If it's easier for now I can simply keep it generic and just say:

> While a stepper motor system under parallel port control is illustrated, a LinuxCNC system can also take advantage of a wide variety of dedicated hardware motion control interfaces for increased speed and I/O capabilities. A full list of interfaces supported by LinuxCNC can be found on the [link]Supported Hardware[link] page on the Wiki.

Maybe a better solution further down the track would be to have the supported hardware listed on its own page in the docs section as a parent to Hardware Drivers section.

ascurtis:
Modified the paragraph in question as detailed above.

KimK:
I agree that this version is clearer, I prefer it. However, an error crept in during the rewrite: In the folder tree diagram (in text), the line "Folder: My_CNC" I believe should be "Folder: configs", at least if the "traditional" names are intended. Otherwise, great, thanks!

ascurtis:
Thanks KimK.

The 'configs' folder appears to be one level up from the 'My_CNC' folder. 

[image](https://user-images.githubusercontent.com/41104842/43041874-881501b6-8db0-11e8-835a-804f3963031b.png)

I can include it in the tree structure in the documentation if it helps clarify things, but maybe it's adding more detail than is necessary in this case?

KurtJacobson:
@ascurtis I think this is a significant improvement over the current user intro, thank you for your efforts!
Re: Directory structure, you can use the `tree` command to create a nice compact hierarchical listing of a dir structure:
```
~/linuxcnc
├── configs
│   ├── my_cnc
│   │   ├── custom_gvcp.hal
│   │   ├── custom.hal
│   │   ├── custom_postgui.hal
│   │   ├── my_cnc.hal
│   │   ├── my_cnc.ini
│   │   ├── postgui_call_list.hal
│   │   ├── shutdown.hal
│   │   └── tool.tbl
│   └── my_cnc.pncconf
└── nc_files
    ├── examples
    └── ...
```
It does not format right here on github, so I don't know how it would look in the PDF or HTML pages, but you could possibly use this to show the hierarchy, and then simply describe the individual files in separate paragraphs instead of the nested list items. Maybe it would even be possible to link the names in the tree to the descriptions ...

ascurtis:
Cheers Kurt.

Is this the `tree ` function you mention?:

https://github.com/allati/asciidoctor-extension-monotree

I'm currently using asciidoclive.com to preview how the mods look. Even though it doesn't have the exact formatting used on the linuxcnc.org webpage, it gets close enough for the purpose of checking. Asciidoclive doesn't understand the `tree` syntax though. The PDFs are created in MSWord, so again the formatting is largely an approximation of how it would look in reality.

KurtJacobson:
@ascurtis That asciidoctor tree plugin looks very good, but I just meant the shell `tree` command. You may have to install it (on debian `sudo apt install tree`), then you just say something like `tree ~/linuxcnc` and it will print a nice dir structure to the terminal. I use it all the time for quickly getting an idea of a projects structure.

I think it looks great in the HTML docs
dir_tree_html

but not so much in the PDF version ... 
dir_tree_pdf

I would suspect there is a way to make it render correctly in the PDF. If you think the tree view might be good I will investigate.

Cheers!
Kurt

ascurtis:
Sure, I'd be happy with the added tree rendering, provided it carried over to the PDF documentation properly.

I suppose the other (lo-tech) thing that could be done is just to add a screenshot of the folder/file list from the file manager?

KurtJacobson:
Great!

A screenshot would work, or it could be a vector image, then it would print crisply. If I can't get it to work in the PDF then I will try that, else call what you have good enough (which it is!).

c-morley:
Are we calling this ready to go?

ascurtis:
I think Kurt was looking into adding the tree structure to the text? Otherwise I'm happy as-is.

KurtJacobson:
I can't figure out how to get the dir tree to render properly in the PDF docs, so lets call it good as-is.

c-morley:
Sounds good - always can do more later.

KurtJacobson:
Thanks @c-morley, that was quick!

ascurtis:
Seconded. Thanks for all your help.

PR #471, opened 2018-07-21
Configs comments: Avoid single quotes/apostrophes

KimK:
When using the syntax highlighter in gedit, single quotes and
apostrophes cause the highlighting to go bright magenta and
stay that way until a matching single quote or apostrophe is
found.

It would be nice to think that this could be fixed by adjusting
the syntax highlighter, but since that has not been done by the
experts, it's probably a tricky job. I think we *do* want the
magenta highlighting to appear in spots "like this" and 'like
this', and I'm not sure what it would take to make that work.

So I'll just go ahead and fix the offending config files.

The old joke says it best:
Patient: Doctor, it hurts when I do *this*.
Doctor: Don't *do* that.

Signed-off-by: Kim Kirwan <kim@kimkirwan.com>


KimK:
Oops, Sorry, I should have titled it "Avoid double quotes/apostrophes", or "Avoid double quotes/single quotes" or some such. I hope that doesn't confuse anyone. The fixes are all OK. Thanks.

KurtJacobson:
@KimK What version of Gedit and what language spec are you using? If the syntax highlighter does not like apostrophes in a comment string that is a bug in the highlighter, not the config files! Gedit 3.22.0 (Debian Stretch) highlights these config files correctly when using the default ini language spec. The Gedit language spec format has not changed, so you should be able to use the new spec with your version of Gedit.

KurtJacobson:
Here is the ini.lang file that appears to work correctly: [ini.lang.txt](https://github.com/LinuxCNC/linuxcnc/files/2216554/ini.lang.txt)

This file includes several others, so it may be the fix was in one of those, if so I can upload those as well.

Replace `/usr/share/gtksourceview-3.0/language-specs/ini.lang` with it or put it in `~/.local/share/gtksourceview-3.0/styles/language-specs/`. If you add in in `.local` you may have to change the language id and name to avoid conflicts with the old spec.

KimK:
Oops, thanks Kurt, I think I was viewing with the wrong type selected, and I'd like to get a clean install going to be sure of what is going on before answering your version questions, so I'll close this for now, but it may reappear in a different form in the future.

PR #472, opened 2018-07-25
Dont crash on corrupt position.txt

SebKuzminsky:
The local hackspace had a machine with a corrupt position.txt file: it existed, but was empty.  As a result, Task would segfault at startup, every time anyone launched LinuxCNC.

Investigation revealed two problems:
1. There was no error in the log complaining about the corrupted position.txt file.
2. Task's convoluted startup skipped inihal setup because of the failure of emcPositionLoad(), so Task would segfault later when trying to access the inihal data structures.

This PR adds the missing error logging to emcPositionLoad(), and fixes the initialization bug in Task that let it run without inihal set up.

### DIFF
--- a/src/emc/nml_intf/emc.hh
+++ b/src/emc/nml_intf/emc.hh
@@ -446,6 +446,8 @@ extern int emcMotionUpdate(EMC_MOTION_STAT * stat);
 
 extern int emcAbortCleanup(int reason,const char *message = "");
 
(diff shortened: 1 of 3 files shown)


KurtJacobson:
Thank you @SebKuzminsky, this is great! I have run into this several times in the past, once when I had first started using LCNC ...

PR #477, opened 2018-07-31
configure.in:  require `yaml2` executable

### LINKED ISSUE #475: `configure.ac` missing checks for `inttool-extract`, `yapps` executables
Commit 8c375852 added a target that runs `intltool-extract` to the Makefiles, and the `intltool` package was added to Debian packaging in d20f8eb1, but a similar check for the executable in `configure.ac` is missing.

Commit fc6eb7f2 removed a bundled version of `yapps` and added the build dep to the Debian packaging, but similarly didn't add it to `configure.ac`.

I know this is true for master, but haven't checked 2.7.

zultron:
Fixes #475; 2.7 counterpart for `inttool-extract` only in #476

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -889,6 +889,12 @@ then
     AC_MSG_ERROR([fuser not found])
 fi
 
+AC_PATH_PROG(YAPPS, yapps, "none", $SPATH)
+if test $YAPPS = "none"
+then
+    AC_MSG_ERROR([yapps not found])
+fi
+
 # MANDB empty is handled in doc/Submakefile
 AC_PATH_PROG(MANDB, mandb, "")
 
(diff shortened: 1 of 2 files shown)


SebKuzminsky:
The commit message for bf985533a99e1 is wrong (talks about intltool), but the patch is right (checks for yapps).  Please fix the commit message and I'll merge it.

SebKuzminsky:
I merged #476 into 2.7 and merged 2.7 into master, so if you rebase bf98553 onto master you won't need the merge in this PR, simplifying it some.

zultron:
Jeezus H. Beezus, @SebKuzminsky you caught the commit with the wrong name, but missed the PR misnamed "configure.in:  require `yaml2` executable"!  I can't use the "copy'n'paste" excuse for that one.

Fixed PR title.  :D

PR #354, opened 2017-11-25
adding optional I parameter to G33.1 which scales the spindle speed u…

rene-dev:
…p on the return move

Signed-off-by: Rene Hopf <renehopf@mac.com>

Issue #167, opened 2016-10-02
Warn users that spindle is too fast in G33 / G95


robEllenberg:
## Here are the steps I follow to reproduce the issue:
1. Have a machine with a slow K axis (ex. max velocity = 60ipm)
2. Run the following G code snippet

```
M3 S1000
G0 X0 Z1
G33 Z0 K0.1
M5
M2
```
## This is what I expected to happen:

Some form of warning that the machine can't reach the required feed rate. To execute that motion properly, the Z axis needs to move at 1000 rpm \* 0.1 in / rev = 100 ipm (greater than my example limit of 60 ipm).
## This is what happened instead:

The cut proceeds without complaint, and cuts the wrong thread pitch.
## It worked properly before this:

Probably never, but this seems to have been an issue with 2.6.x as well.
## Information about my hardware and software:

Use a simulation config with a slow Z axis (the stock axis simulator will work).

robEllenberg:
I have a fix in progress here:

https://github.com/robEllenberg/linuxcnc-mirror/tree/feature/nominal-vel-during-sync

It does two things:
1. Pop up an operator error dialog if the spindle speed is too fast for the axis limits
2. Attempt to reduce the spindle speed to a safe speed by issuing a new speed command before the sync motion starts. Obviously this does nothing if the spindle is manually controlled or fixed speed.

It's not ideal because the warning only appears right before the cut starts, but at least the user will have some feedback this way.

zultron:
Extra documentation on this issue [here][1] and [here][2].

[1]: http://www.mail-archive.com/emc-developers@lists.sourceforge.net/msg16946.html
[2]: http://www.mail-archive.com/emc-developers@lists.sourceforge.net/msg17087.html

zultron:
Nice, the above-linked branch passes my regression test by scaling spindle speed.  See the top commit in [this branch][1].

[1]: https://github.com/zultron/machinekit/tree/spindlesync-exceeds-maxvel-lcnc

robEllenberg:
Updated branch [here](https://github.com/robEllenberg/linuxcnc-mirror/tree/feature/spindle-sync-overhaul-2.7-rebase) with some fixes:

- New spindle sync algorithm that seems to have less ripple with high resolution encoders (needs testing to confirm)
- Fixed interaction with blending / position sync motion
- Inform the user via a GUI info message if the spindle speed is being reduced.

zultron:
[...]

PR #479, opened 2018-08-02
Fix synched spindle moves exceeding linear axis limits

zultron:
See  "Warn users that spindle is too fast in G33 / G95" #167.

This is @robEllenberg's work, when too high spindle speed during spindle-synched move would exceed linear axis limits, at the time of the motion, give a warning and attempt to reduce spindle speed.

On top, I added an interp-side warning, shown at program start time, and a regression test.

Issue #167 was pointed out in related issue #354.


andypugh:
There seems to be quite a collection of open issues around this. 
How do things stand for a merge? And into which branch?

zultron:
I haven't looked at this for a long time.  Maybe @robEllenberg remembers this better?  This has been in Tormach's tree since the PR was first opened, so it's had a few years of testing, albeit only in that single environment.

robEllenberg:
There are some additional fixes that need to be backported before this can be merged into 2.7 (and those fixes also need to make it into 2.8). I'd like to hold off on the merge until I can straighten them out.

rene-dev:
any update on this? sounds like a good feature for the upcoming 2.8 release.

robEllenberg:
Agreed, I think this is an important feature for 2.8. The branch as it stands needs some additional threading fixes backported. Depending on how that goes I'll either update the PR or make a new branch based on the current 2.8.

andypugh:
Did this go in? I am suspecting not?

robEllenberg:
As far as I know this didn't get merged. The last time I touched this I ran
into a bunch of merge conflicts trying to backport my changes from the
PathPilot fork, and simply haven't revisited this in a while. I still
believe that these changes would be a valuable improvement for 2.8. The
good news is that in the intervening time I've created a few test cases
(for runtests) to verify performance of spindle synchronization in sim, so
it will make it easier to verify that it's working as expected. If I can
lean on a few people with hardware to put the new code through its paces
(e.g. Gene), then we can make some progress here.

andypugh:
Looking at the merge conflicts reported here in the web editor it doesn't look too hard, but I think that a rebase on to 2.9 would be needed. And then all the multi-spindle stuff is likely to cause trouble. 
But I think it's worth a try. (And I have a lathe)

tests/motion/spindlesync-exceeds-maxvel/test-ui.py
```diff
@@ -0,0 +1,74 @@
...
+# Spindle synchronized motion at 2x Z axis maxvel
+c.mdi('G33 Z-5 K%.3f' % max_pitch)
+
+# While above MDI command runs, take an average of pitch while the Z
+# axis is in motion
+running = True
+timeout = 5
+avg_sum = 0; avg_count = 0
+while timeout > 0:
+    if abs(h['pitch']) > 0.001:
```
petterreinholdtsen, line 47: I tried porting the spindlesync-exceeds-maxvel test to the latest master, and ran into a problem with h['pitch'] never being above 0.001.  Where is it supposed to be set?

petterreinholdtsen:
https://github.com/petterreinholdtsen/linuxcnc/tree/479-test-spindlesync-exceeds-maxvel-on-master is my port of the test to python 3 and upgrade of the ini file to get linuxcnc to accept it.

petterreinholdtsen:
I notice pull request #581 also claim to solve #167.  Perhaps this pull request has been superseeded by #581?

andypugh:
This does look to have been superseded.

PR #480, opened 2018-08-10
docs: fix command for launching pyvcp example

KurtJacobson:
Jin^eLD reported on IRC that the command to launch the tiny.xml  PyVCP example did not work. If using halcmd instead of `halrun -I` it does work. `halrun` with the -I flag should be the same as `halcmd` (if I understand it correctly) but it does not seem to work. This may be a bug in `halrun` ...

### DIFF
--- a/docs/src/gui/pyvcp.txt
+++ b/docs/src/gui/pyvcp.txt
@@ -41,7 +41,7 @@ image::images/pyvcp_mypanel.png[]
 If you place this text in a file called tiny.xml, and run
 
 ----------------------------------------------
-halrun -I loadusr pyvcp -c mypanel tiny.xml
+halcmd loadusr pyvcp -c mypanel tiny.xml
 ----------------------------------------------
 
 PyVCP will create the panel for you, which includes two widgets, a


SebKuzminsky:
`halrun` starts the realtime environment (including HAL), while `halcmd` does not.  So `halrun` will work if LinuxCNC is *not* running, and `halcmd` will work if LinuxCNC *is* running.  Maybe the docs should mention both possibilities?

KurtJacobson:
@SebKuzminsky It does not seem to make any difference if LCNC is running or not.

With LCNC running:
```
kurt@TC0:~$ halrun -I loadusr pyvcp -c mypanel tiny.xml  
halrun: too many arguments <loadusr pyvcp -c mypanel tiny.xml>
```

Without LCNC running:
```
kurt@TC0:~$ halrun -I loadusr pyvcp -c mypanel tiny.xml
halrun: too many arguments <loadusr pyvcp -c mypanel tiny.xml>
```

This seems to work fine either way:
`halcmd loadusr pyvcp -c mypanel tiny.xml`

SebKuzminsky:
Heh, you're right, that `halrun` command is just totally broken.

`halcmd loadusr` works without the realtime environment loaded, I guess because it doesn't use realtime.  `halcmd loadrt` only works if the realtime env is started (for example via

Issue #164, opened 2016-09-17
Excessive jitter spindle position tracking


robEllenberg:
Looking at axis velocity / acceleration during G33 motions, there is a noticeable jitter in velocity, and large acceleration swings. The image below (zoomed in significantly) shows this occuring.

![image](https://cloud.githubusercontent.com/assets/1210940/18227405/173f2818-71f1-11e6-824f-4d4e7042b08a.png)

This is strange to see in a simulation config, because the encoder signal is very high resolution, and the the simulated spindle speed signal is clean, constant, and smooth. This obviously isn't representative of real machines with real encoders, but I believe this ideal case shows a flaw in the position error calculation.
## Here are the steps I follow to reproduce the issue:
1. Load an axis sim config with a simulated spindle encoder (I used AXIS as well, but it doesn't matter)
2. Configure HALScope with two channels, Zvel and Zacc
3. Run this G code program

```
G90 G20 G64 P0.005
G0 Z0
S240 M3
G33 Z-.5 K0.1
G0 Z0.1
M5
M2
```

Here's a [branch](https://github.com/robEllenberg/linuxcnc-mirror/tree/feature/spindle-sync-jitter-fix) with this change implemented. With a perfect spindle encoder, there is no jitter in the motion tracking:

![image](https://cloud.githubusercontent.com/assets/1210940/18227457/d8248036-71f2-11e6-9937-24200ee54c6d.png)

With encoder resolution of 256 PPR, there is still jitter, but not as much:

![image](https://cloud.githubusercontent.com/assets/1210940/18610876/53b35e08-7cf6-11e6-9a8e-fe99d9183ed9.png)

Compare this to stock 2.7 performance with the same simulated spindle encoder:

![image](https://cloud.githubusercontent.com/assets/1210940/18610895/c1c38576-7cf6-11e6-95d3-aa4805b4afba.png)

For reference, here's a [test branch](https://github.com/robEllenberg/linuxcnc-mirror/tree/test/spindle-tracking-jitter) with stock 2.7 spindle tracking and the quantized spindle encoder.

rene-dev:
Im not sure its related, but I noticed that the velocity display in gremlin is not stable during a g33.
Any reason this has not been merged?

zultron:
@rene-dev, this is another issue without an actual PR.  I submitted #479 for #167; you wanna go for this one?  ;)

PR #481, opened 2018-08-12
Feature/spindle sync jitter fix

rene-dev:
fix for #164

### DIFF
--- a/lib/hallib/sim_spindle_encoder.hal
+++ b/lib/hallib/sim_spindle_encoder.hal
@@ -5,16 +5,23 @@ setp sim_spindle.scale 0.01666667
 loadrt limit2 names=limit_speed
 loadrt lowpass names=spindle_mass
 loadrt near names=near_speed
+loadrt quant names=sim_spindle_encoder
 
-# this limit doesnt make any sense to me:
+# Bound spindle acceleration to something reasonable
 setp limit_speed.maxv 5000.0 # rpm/second
 
 # encoder reset control
 # hook up motion controller's sync output
 net spindle-index-enable motion.spindle-index-enable <=> sim_spindle.index-enable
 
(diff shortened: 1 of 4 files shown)


SebKuzminsky:
I made a throw-away integration branch containing this PR merged into master, and I ran @robEllenberg's recipe (from #164) for showing the Z-velocity jitter (using the sim/axis/axis.ini config).  Here is the result:
spindle-sync-jitter-481-saa

For reference, here's the result from the current tip of master with the quantized spindle encoder commits from this PR cherry-picked in (same config):
spindle-sync-jitter-quant-encoder-saa

So, while I agree that master has a lot of Z-velocity jitter during this spindle-synced move, I don't see the proposed PR making the situation better.  Am I missing something here?

samcoinc:
It has been a while since I have played with this branch but I think there is an extra pin that needs to be set.

This was from rob
   - Add a HAL pin, motion.spindle-tracking-gain (range 0.0 to 1.0) to
   control how aggressive position tracking is. 1.0 is stock behavior, 0.0 is
   no position tracking.

IIRC I was playing with .5
sam

SebKuzminsky:
There is no new hal pin in this PR.

samcoinc:
hmm - this is what I had been testing
feature/spindle-sync-overhaul-2.7-rebase
 
from robs..
https://github.com/robEllenberg/linuxcnc-mirror/tree/feature/spindle-sync-overhaul-2.7-rebase

Are there a lot of differences?  The last commits on robs are from Nov 07, 2017
This is confusing (for me)..

sam

SebKuzminsky:
This is confusing for me too.  @rene-dev can you explain this PR to me?

rene-dev:
whoops, I think I picked the wrong branch. It was supposed to be the real fix mentioned above.

rene-dev:
according to rob, there is more to it. He will collect

PR #482, opened 2018-08-14
 docs: add missing `rapidrate` method to linuxcnc.command docs, and convert tabs to spaces

KurtJacobson:
The file had a significant number of tabs mixed in with the predominant space indentation style. I took the liberty to replace each tab with four spaces, which are preferred since they make more money than tabs. I would be happy to open a PR for just the addition of the `rapidrate` method if the tabs to spaces change is too invasive.

Thanks,
Kurt


SebKuzminsky:
Thanks for making these two logical changes into two distinct commits, makes it easier to review.

KurtJacobson:
Absolutely! Thanks for the merge @SebKuzminsky.

PR #483, opened 2018-08-17
 docs: clarify value of stat. units attributes 

KurtJacobson:
yet more py interface doc clean up ...

### DIFF
--- a/docs/src/config/python-interface.txt
+++ b/docs/src/config/python-interface.txt
@@ -94,7 +94,7 @@ status of adaptive feedrate override (0/1).
 current value of the analog input pins.
 
 *angular_units*:: '(returns float)' -
-reflects [TRAJ] ANGULAR_UNITS ini value.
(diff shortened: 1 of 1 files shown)


docs/src/config/python-interface.txt
```diff
@@ -242,7 +242,7 @@ axis limit masks. minHardLimit=1,
 maxHardLimit=2, minSoftLimit=4, maxSoftLimit=8.
 
 *linear_units*:: '(returns float)' -
-reflects [TRAJ]LINEAR_UNITS ini value.
+machine linear units per mm, reflects [TRAJ]LINEAR_UNITS ini value.
```
SebKuzminsky: Isn't this "machine linear units per inch" on imperial machines?

SebKuzminsky:
Thanks for demonstrating so clearly that you were right and I was confused.  Much appreciated.

PR #492, opened 2018-08-27
Create Spindle_monitor.comp

blazini36:
Component to trigger an output based on the commanded vs actual speed of spindle. Useful for triggering E-stop if spindle speed drops below set threshold (cutter jam) to avoid damage. 

Required lowpass component to filter spindle-feedback input pin during testing.

Originally written by Andypugh, modified and tested by myself.


blazini36:
I’m not all that familiar with the git process but I believe I committed it to the wrong spot in the directory. Should be in src/hal/components.

SebKuzminsky:
Thanks for the heads up -- I should have checked that.  I moved the file where it belongs.

PR #499, opened 2018-09-24
Remove rtai_shm again

NTULINUX:
If you're using RTAI, compile rtai_shm into
rtai_sched by enabling it as a built-in,
not as a module.


SebKuzminsky:
Thanks for fixing this so quickly.

Issue #159, opened 2016-09-05
Reportedly, pncconf does not work on preempt-rt because hal.kernel_version does not exist


jepler:
see thread at https://forum.linuxcnc.org/forum/39-pncconf/30751-problems-with-prempt-rt-kernel-error-and-7i76e?limitstart=0

I did not test this for myself, but I think it's an accurate bug report that will affect any uspace user.

jepler:
@c-morley can you take a look at this?

c-morley:
I will try to look at it this weekend

Chris


________________________________

c-morley:
I did not get time to look into this, but looking at the error message it seems that the python module linuxcnc, with preempt_rt doesn't have kernal_version as an attribute:
AttributeError: 'module" object has no attribute 'kernel_version'

I assume this might be because there is no hard requirement for  kernel version?
Maybe we could just have it report the current kernel version so the test passes?

Chris M

jepler:
Yes, uspace realtime is not tied to one specific kernel, so that is why I coded it so that hal.kernel_version is not defined unless RTAPI_KERNEL_VERSION is (see commit 2c5ed60 "let python know the required kernel version"):
~~~~
#ifdef RTAPI_KERNEL_VERSION
    PyModule_AddStringConstant(m, "kernel_version", RTAPI_KERNEL_VERSION);
#endif
~~~~

In 2.7, uspace is realtime with "any RT-Preempt kernel".  In 2.8, this will depend on compile-time configuration details *AND* runtime installed packages besides the kernel.  For instance, I could write something like this:
~~~~
#ifdef RTAPI_KERNEL_VERSION
    PyModule_AddStringConstant(m, "kernel_version", RTAPI_KERNEL_VERSION);
#else
    if(rtapi_is_realtime())
    {
        uname(&uts);
        PyModule_AddStringConstant(m, "kernel_version", uts.release);
    }
    else
    {
        PyModule_AddStringConstant(m, "kernel_version",
            "Any RT-Preempt kernel"
#ifdef USPACE_RTAI
            " or any RTAI kernel with LXRT support"
#endif
#ifdef USPACE_XENOMAI
            " or any Xenomai kernel with posix skin support"
        );
    }
#endif
~~~~

[...]

PR #500, opened 2018-10-01
WIP doc changes prior to making stretch release 'official'

jepler:
@SebKuzminsky and I recently discussed that based on his experience running systems based on the stretch live images with preempt-rt, as well as feedback we've both seen on IRC, he is comfortable making it an official and recommended OS for 2.7.  This PR is a place for us to workshop the doc changes and figure out what's missing before we do this.  It's absolutely not ready to merge in its current state and before work is done on wlo and buildbot too.

### DIFF
--- a/docs/src/getting-started/getting-linuxcnc.txt
+++ b/docs/src/getting-started/getting-linuxcnc.txt
@@ -34,8 +34,11 @@ Image.
 
 Download the Live/Install image by clicking here:
 
-http://www.linuxcnc.org/linuxcnc-2.7-stretch.iso
+http://www.linuxcnc.org/stretch/linuxcnc-stretch-uspace-amd64.iso
 
+Or, if you are installing on a system which does not support 64 bit operating systems,
+
+http://www.linuxcnc.org/stretch/linuxcnc-stretch-uspace-i386.iso
 
 === Download using zsync
 
@@ -55,8 +58,12 @@ sudo apt-get install zsync
 . Then run this command to download the iso to your computer
 +
 ----
-zsync http://www.linuxcnc.org/linuxcnc-2.7-stretch.iso.zsync
+zsync http://www.linuxcnc.org/stretch/linuxcnc-stretch-uspace-amd64.iso.zsync
 ----
+
+Due to technical limitations, the downloaded file may have a revision number in
+the filename, so that the local file ends up being named e.g.,
+"linuxcnc-stretch-uspace-amd64-r12.iso".
 
 .zsync in Windows
 
@@ -103,7 +110,7 @@ a computer.  The image is too large to fit on a CD.
   then use this command:
 +
 -----
-dd if=linuxcnc-2.7-stretch.iso of=/dev/sde
+dd if=linuxcnc-stretch-uspace-amd64.iso of=/dev/sde
 -----
 
 The https://www.debian.org/CD/faq/#write-usb[Debian FAQ] provides more
--- a/docs/src/getting-started/getting-linuxcnc.txt
+++ b/docs/src/getting-started/getting-linuxcnc.txt
@@ -32,9 +32,9 @@ Image.
 
 === Normal Download
 
-Download the Live/Install CD by clicking here:
+Download the Live/Install image by clicking here:
 
-http://www.linuxcnc.org/linuxcnc-2.7-wheezy.iso
+http://www.linuxcnc.org/linuxcnc-2.7-stretch.iso
 
 
 === Download using zsync
@@ -55,7 +55,7 @@ sudo apt-get install zsync
 . Then run this command to download the iso to your computer
 +
 ----
-zsync http://www.linuxcnc.org/linuxcnc-2.7-wheezy.iso.zsync
+zsync http://www.linuxcnc.org/linuxcnc-2.7-stretch.iso.zsync
 ----
 
 .zsync in Windows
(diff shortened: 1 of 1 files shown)


andypugh:
Will it not work at all with a DVD? I have some machines that seem to refuse to boot from USB (even trying really very hard, and trying all the USB ports). IIIRC the D610MO will only USB boot from back-panel ports, for example. For my most recent LinuxCNC installation I actually went out and bought a £25 USB DVD drive (and that made it very simple)

KurtJacobson:
@andypugh not exactly sure what problem you are referring too, but I have installed jepler's Debian 9 R10 ISO from a DVD a bunch of times (15+) on various machines with no problems whatsoever.

jepler:
I haven't tried it with physical optical media.  It *does* boot in qemu when given as the "-cdrom" argument, which is a good sign.  Perhaps I shouldn't preemptively delete those sections, but it would be nice if someone verified it works with physical optical media.

jepler:
for those just driving by, the latest images can be found at http://www.linuxcnc.org/testing-stretch-rtpreempt/ though this will change when we make the images official.

KurtJacobson:
This comment is made from a Live session running from a DVD with the Debian 9 R11 ISO.

jepler:
@KurtJacobson thanks so much for testing!  I'll leave the references to using a DVD as install media intact.

KurtJacobson:
@jepler I proceeded to install from the DVD, and It Just Works (TM). I have had trouble with other Linux distros recognizing my non-free WiFi card, but, as usual, your ISO worked seamlessly. It's the best thing since sliced bread :). 

The only time I had any trouble installing it was when I did not have an Internet connection, that was probably with R8 or R9. I guess that is one more thing to test.

jepler:
I've heavily updated this PR again.  I think everything is right, except the location of the .iso files themselves.  Feedback please!

andypugh:
Not a lot to say, other than "go for it" :-)

docs/src/getting-started/getting-linuxcnc.txt
```diff
@@ -111,9 +103,13 @@ a computer.  The image is too large to fit on a CD.
   then use this command:
 +
 -----
-dd if=linuxcnc-2.7-wheezy.iso of=/dev/sde
+dd if=linuxcnc-2.7-stretch.iso of=/dev/sde
 -----
 
+The https://www.debian.org/CD/faq/#write-usb[Debian FAQ] provides more
+information about how to work with USB media, including how to use
+win32diskimager to write the image to USB media on Microsoft Windows.
```
KurtJacobson: It might be worth mentioning https://etcher.io/ here. It is by far the easiest to use tool for writing boot-able images to USB, and is open source and cross platform.

andypugh:
Just one thing. If this is uspace then then I think that the pncconf "test/tune axis" button will error, due to  #159 only being fixed in master.

jepler:
@SebKuzminsky is complete breakage of pncconf on this image something you'd want to make a new 2.7 release to address?

jepler:
(well, not complete, you just can't test/tune afaik)

SebKuzminsky:
Yes, this should be fixed in 2.7 and I'll make a new 2.7 release.

SebKuzminsky:
I cherry-picked 0034fdf into 2.7 and verified that the Test/Tune Axis dialog pops up in pncconf now.

jepler:
@rene-dev your well meaning commit message unintentionally closed their this PR. No harm done, as I can still reopen it.

rene-dev:
Oh, Im sorry. I really have no idea where the wrong number came from.

andypugh:
Where are we with this? We are now in a position where the current Live-CD image doesn't work properly (repositories at Debian have moved to an archive). I think this is also causing problems with the buildbot (the Wheezy instances fail at apt-get)

SebKuzminsky:
@andypugh I think there are three things we should do, all independent:
[ ] Fix the `/etc/apt/sources.list` files on the Wheezy buildbot machines (done).
[ ] Fix the `/ect/apt/sources.list` files on the Wheezy/RTAI/LinuxCNC-2.7 ISO.
[ ] Make a new RTAI/LinuxCNC ISO (last time i tried this RTAI 5.0.1 on Linux 4.4.0 was not stable enough to use).

andypugh:
My own attempts to make an RTAI kernel on 4.9.80 have met with little success. (kernel panic on boot). 
The main problem I have is, I suspect, finding a working .config. 
There are a set of configs in the ShabbyX repo, and in Seb's repo. But those all seem to be 3.x kernels. 
Has anyone found a .config for 4.9 kernels?

andypugh:
The Download link on wlo has been changed to point to the Stretch ISO, 
http://www.linuxcnc.org/downloads/
but that (and this) still point to an image in the "testing" repository. 
Should the docs (and wlo) point to a file in the /iso/ folder?

rene-dev:
any update on this? is there any reason stretch isn't the default iso?
wheezy is old, and cant even run on machines a few years old.

phillc54:
I have tried both the r12 and r13 versions and they are missing the dirmngr package which is required to add the archive signing keys for the Buildbot.
Apart from that they work well.

andypugh:
Where are we with this? I suspect this PR has been overtaken by events?

SebKuzminsky:
IMO, we shouldn't outright switch from Wheezy/RTAI/32-bit to Stretch/RT-Preempt/64-bit as the one install method we recommend, because some machines will get bad latency with RT-Preempt.

We should recommend to our users that they try booting the "Live" part of the Stretch ISO and run `latency-histogram`.  If that's good enough for their needs they should install that.  If it's not good enough, try with the RTAI ISO.  (And if that's still not good enough, get a different computer.)

andypugh:
That is my intention for 2.8. But probably not Stretch.

andypugh:
Can we conclude that with a 2.8 release on Buster that this PR has been overtaken by events and can  be closed?

PR #501, opened 2018-10-05
docs: fix formatting of spindle dict section heading

KurtJacobson:
This fixes a minor formating error that caused the new spindle dictionary section to be at the wrong level in the TOC.


KurtJacobson:
@rene-dev Thanks!

PR #502, opened 2018-10-06
BugFix: spindle dict item `override_enabled` should use underscore instead of dash

KurtJacobson:
This makes the `override_enabled` key consistent with the use of underscores elsewhere and with the docs.


andypugh:
The HAL pins all use hyphens. 
I think I was following the wrong convention in the Python interface.

KurtJacobson:
@andypugh You correctly used underscores on all the other dict items, you just missed that one.

KurtJacobson:
Oops, looks like I need to change something in the tests too.

KurtJacobson:
I have also tested axis, gmoccapy, gscreen, touchy and tklinuxcnc as well and all seem fine.

Issue #426, opened 2018-03-21
Arc Blending Bug


ToddZuercher:
The New Trajectory Planner pauses under certain circumstances.
https://forum.linuxcnc.org/20-g-code/34213-axis-short-stop-very-short-about-every-15-20-sec-when-cutting-circle?start=10#107672

## Here are the steps I follow to reproduce the issue:

 1. A helical arc, followed by 3 or more regular arcs.
 
## This is what I expected to happen:
Smooth blended motion between all the moves.

## This is what happened instead:
Linuxcnc will pause between the 1st and 2nd regular arcs. (Decelerate to 0 then continue.)

## It worked properly before this:
? Disabling the new TP with the ini setting ARC_BLEND_ENABLE = 0 makes the pause go away.

## Information about my hardware and software:
Running on multiple systems (real hardwar and VMs) using the latest Wheezy, and Stretch Linuxcnc images with the latest 2.7 version of Linuxcnc (from Buildbot.)

pcw-mesa:
I see the same pauses on that gcode with the new TP disabled, on both 2.7 and master
so if there's an issue, it seems to be common to both TPs

ToddZuercher:
Did you try running the simplified G-code I posted with both planners?
https://forum.linuxcnc.org/media/kunena/attachments/3190/arc-bug.ngc
Try it with in G20 with about F150 (or how ever fast you can on the config your testing with where the acceleration constraints won't get in the way too much.) 
The difference between the new and old TP is startling.

pcw-mesa:
Yes, and I get the same results as I had before, that is both the new and old TPs have similar
glitches
Though in some cases they dont occur on old vs new TP and this is very feed rate dependent

robEllenberg:
I can confirm this is a TP bug, luckily one I have a fix for in a local branch. It's a regression in one of the later 2.7 releases. What happens is that the TP fails to create an arc blend between segments (because they aren't planar), but fails to fall back to parabolic blending. Instead, it forces the blend to be treated as tangent, which means it needs to slow way down at the corner to obey acceleration constraints. 

I'll send word as soon as I have a hotfix branch ready (hopefully this weekend, the next few days are quite busy).

andypugh:
Is this ready for merge?

rene-dev:
https://github.com/LinuxCNC/linuxcnc/pull/509

rene-dev:
This should be fixed in 2.7 and 2.8-pre now.

PR #512, opened 2018-10-18
Rellenberg/bug447 fix blend tangent fallback

### LINKED ISSUE #447: unnecessary TP slowdown on some arc-to-line transitions
Originally reported here: https://www.cnczone.com/forums/tormach-pathpilot-/354884-tormach.html

The g-code in comment number 6, when run on a 2.7.13 sim/axis machine, produces this result:
tp-slowdown-2
Note how machine velocity drops to 0 on exit from the first arc.

The partial slowdowns on entry to the first arc and exit from the second arc are expected, since the tool path makes sharp corners there, but the transition from the first arc to the following straight-feed looks tangent (or at least nearly so), and with G64 in effect the speed should not drop to 0.

This problem is reproducible on 2.7 and master.

This g-code demonstrates the problem:
~~~~~
; from https://www.cnczone.com/forums/tormach-pathpilot-/354884-tormach.html
; comment #6

g64   ; best speed, tolerances be damned
g91.1 ; incremental arc centers

;N1710 T9 G43 H9 M6
(0.500 End Mill HSS)
N1720 S3450 M3 M8
N1730 G0 X1.6009 Y0.7267
N1740 Z0.15
N1750 Z0.1075
N1760 G1 Z-0.215 F20.
N1770 G3 X1.5 Y0.35 Z-0.215 I0.6525 J-0.3767
N1780 G1 Y-0.35
N1790 G3 X1.5647 Y-0.5915 Z-0.215 I0.483 J0.
N1800 G0 Z0.15

m2
~~~~~

rene-dev:
should fix #447 and #426

### DIFF
--- a/src/emc/tp/blendmath.c
+++ b/src/emc/tp/blendmath.c
@@ -446,7 +446,13 @@ int calculateInscribedDiameter(PmCartesian const * const normal,
     PmCartesian planar_x,planar_y,planar_z;
 
     //Find perpendicular component of unit directions
-    // FIXME use plane project?
+    // FIXME Assumes normal is unit length
+    
+    /* This section projects the X / Y / Z unit vectors onto the plane
+     * containing the motions. The operation is done "backwards" here due to a
+     * quirk with posemath. 
+     *
+     */
     pmCartScalMult(normal, -normal->x, &planar_x);
     pmCartScalMult(normal, -normal->y, &planar_y);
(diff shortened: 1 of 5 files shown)


robEllenberg:
There's one more fix this branch needs here, to respect exact-path and exact stop mode: 
https://github.com/robEllenberg/linuxcnc-mirror/commit/2d4df632e5c5576f53d0e5664af466b3215e5846

PR #517, opened 2018-10-27
Allow userspace components to include files relative to the location of the .comp file

jepler:
This is a test for #515.


petterreinholdtsen:
Should this pull request still be used?  Issue #515 is closed with a comment suggesting this pull request should be cancelled.

jepler:
Components are divided into two categories, "realtime" and "userspace".  The original issue #515 was half fixed in #516, for "realtime" components, which were the original problem. This added a test for "userspace" components, but didn't fix it for that case. Now, I've updated this PR and actually implemented the fix.

PR #519, opened 2018-10-27
Demux bugfixes

jepler:
- the default personality was 0, leading to a useless component
 - the maximum size was 1, meaning using a personality above 1 would lead to memory corruption
 - the calculation was incorrect, so when e.g., bit 02 was set a wrong bit number would be calculated


andypugh:
is that a record? More bugs than lines of code?

PR #184, opened 2016-10-20
pncconf: don't refer to kernel_version when not is_kernelspace

jepler:
This is intended to fix #159.  Tested on an XU4 (no mesa board actually attached), now it works to open the "tune" windows (but nothing can be done since there's no board attached)

PR #521, opened 2018-10-29
pncconf: don't refer to kernel_version when not is_kernelspace

jepler:
Fix issue #159 for 2.7.  This is the same as #184, just cherry picked.  Noted by @andypugh in #500.

Testing performed: Make selections in pncconf until it is possible to arrive at the "X Motor" screen with the "Test / Tune Axis" button.  Click it.

Before the change, the reported error occurs.

After the change, an error occurs when trying to load hm2_pci, which is expected since no such device is attached.

### DIFF
--- a/src/emc/usr_intf/pncconf/pncconf.py
+++ b/src/emc/usr_intf/pncconf/pncconf.py
@@ -1638,7 +1638,7 @@ class App:
                 return True
             else:
                 return False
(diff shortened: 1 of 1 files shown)


jepler:
Looks like @SebKuzminsky fixed this independently, so this PR is no longer required.

PR #522, opened 2018-10-30
Create public header for hostmot2 public APIs

jepler:
This is trying to address what I think is the actual problem underlying #515. While I didn't test this with a package, I did test that with an RIP build "mesa_7i65.comp" can be copied out of the tree and built with halcompile --compile. buildbot shows green for kernel-mode RTAI builds, and I tested locally with uspace builds on debian stretch.

This is closely related to #516 but is rebased for 2.7 and the bits about "pktuart" are removed.


jepler:
@SebKuzminsky pending the results of buildbot and travis I think this is ready for merge.  I had initially dropped the two added tests on the floor when I initially created this PR, so we definitely want to wait and see how they shake out.

jepler:
For reasions previously discussed, this PR isn't suitable for 2.7 in its current form.

Issue #523, opened 2018-11-03
angular only movement doesn't honour max velocity setting


c-morley:
## Here are the steps I follow to reproduce the issue:

home AXIS's 9 axis sample config
Set the max Velocity slider to 0
in MDI enter: g0 A20

## This is what I expected to happen:

linuxcnc to wait till max Velocity slider is above 0

## This is what happened instead:

Linuxcnc moves A axis to 0 

## It worked properly before this:

Not sure but 2.7.14 has the same problem

## More info:

Angular movement does honour rapid override setting.
It honours max Velocity setting if another linear axis is included.

rene-dev:
using the same max value for angular and linear moves does not make sense. So really, you need 2 sliders, for max linear and angular velocity?

c-morley:
Max velocity was a hack to create a way to slow down rapids a long time ago.
Now we have a proper rapid overrride for this.

But it's a surprise that the max velocity didn't limit angular velocity,

Thinking out loud:
if there is linear : limit base on max linear velocity.
if there is angular only limit based on max angular velocity.

So really to make a slider makes sense it should be in percentage.

Chris M

rene-dev:
ah, yes. I like the idea of a factor. but of what? fastest linear/angular axis?

c-morley:
Thinking out loud again:
if there is linear : limit base on percentage of max linear velocity.
if there is angular only limit based percentage of max angular velocity.

basically the same as feed and rapid overrides,
Or else just get rid of it all together ...

zultron:
I have a fix from another branch for this, but I haven't tested it with Axis or the other UIs.

If someone wants to try it out, I submitted it in PR #525.

zultron:
I [raised this issue][1] on emc-developers back in 2017-01, asking for input about how to treat the slider value, as a linear velocity or as a percentage.  Summarizing the feedback:

- @cradek pointed out that the touchy GUI needs a linear velocity for continuous jogs
- @samcoinc likes to know the velocity limit in linear units, not just a percent scale, when adjusting the slider in the GUI

These issues are easily addressed in this implementation.  Since the limits ended up being separate values for linear and angular velocity, touchy jogging will be unaffected.  In `axis.py`, the angular velocity is simply scaled by the computed percentage of the slider value, and the result is sent to motion, so the appearance of the GUI is unchanged.

Also, at the time, @gmoccapy volunteered to update the gmoccapy GUI.  I hope he's still agreeable to do so!

[...]

PR #525, opened 2018-11-06
Apply 'max velocity' to pure rotary motion

zultron:
**NOTE:**  This commit addresses #523; I haven't checked to see whether GUIs need updates in order for this to work.

Add max angular velocity to UI parallel to max linear velocity, plumb
through task to motion, and in `tpGetMaxTargetVel()`, limit
pure-rotary motion.  UIs may then give user control of pure rotary
motion with a 'max velocity' slider.

- UI
  - `.ini` file:  `[TRAJ]MAX_ANGULAR_VELOCITY`
  - Python module
    - Add 2nd arg to `command.maxvel(vel_lin, vel_ang)`
    - Add `stat.max_angular_velocity` attribute
    - Update `axis.py` to compute angular velocity scale
      proportionately to linear velocity
  - `halui`: add `mav_*` pins in parallel to `mv_*` pins
  - `inihal`:  add `traj_max_angular_velocity` field
- NML
  - Add `velocity_angular` field to `EMC_TRAJ_SET_MAX_VELOCITY` message
  - Add `maxAngularVelocity` field to `EMC_TRAJ_STAT` message
- task
  - Add 2nd arg to `emcTrajSetMaxVelocity(maxvel_linear, maxvel_angular)`
- NML (previous commit)
- motion
  - Add `emcmot_config_t.limitVelAng` field
  - Add `emcmot_command_t.vela` field
- tp
  - Add 3rd arg to `tpSetVlimit()`

- Tests:
  - `motion/jogwheel`:  add ABC axes to test `halui`
  - `motion/max_velocity`:  add test for max velocity

Ugliness:

The [`.ini` docs][1] place `MAX_LINEAR_VELOCITY` and
`MAX_ANGULAR_VELOCITY` in the `[DISPLAY]` section; the former is the
maximum possible value on the max velocity slider.  The `MAX_VELOCITY`
param is in the `[TRAJ]` section, and is the set value on the slider.
This naming makes it hard to add the missing angular equivalent to
`[TRAJ]MAX_VELOCITY`.

[1]: http://linuxcnc.org/docs/2.7/html/config/ini-config.html

### DIFF
--- a/src/emc/ini/inihal.cc
+++ b/src/emc/ini/inihal.cc
@@ -134,6 +134,7 @@ int ini_hal_init(void)
 
     MAKE_FLOAT_PIN(traj_default_velocity,HAL_IN);
     MAKE_FLOAT_PIN(traj_max_velocity,HAL_IN);
+    MAKE_FLOAT_PIN(traj_max_angular_velocity,HAL_IN);
     MAKE_FLOAT_PIN(traj_default_acceleration,HAL_IN);
     MAKE_FLOAT_PIN(traj_max_acceleration,HAL_IN);
 
@@ -152,6 +153,7 @@ int ini_hal_init_pins()
 {
     INIT_PIN(traj_default_velocity);
     INIT_PIN(traj_max_velocity);
+    INIT_PIN(traj_max_angular_velocity);
     INIT_PIN(traj_default_acceleration);
     INIT_PIN(traj_max_acceleration);
 
@@ -192,23 +194,26 @@ int check_ini_hal_items()
     const value_inihal_data &new_inihal_data = new_inihal_data_mutable;
 
     if (CHANGED(traj_default_velocity)) {
-        if (debug) SHOW_CHANGE(traj_default_velocity)
+        if (debug) SHOW_CHANGE(traj_default_velocity);
         UPDATE(traj_default_velocity);
         if (0 != emcTrajSetVelocity(0, NEW(traj_default_velocity))) {
             rcs_print("check_ini_hal_items:bad return value from emcTrajSetVelocity\n");
         }
     }
-    if (CHANGED(traj_max_velocity)) {
-        if (debug) SHOW_CHANGE(traj_max_velocity)
+    if (CHANGED(traj_max_velocity) | CHANGED(traj_max_angular_velocity)) {
+        if (debug) SHOW_CHANGE(traj_max_velocity);
+	if (debug) SHOW_CHANGE(traj_max_angular_velocity);
         UPDATE(traj_max_velocity);
-        if (0 != emcTrajSetMaxVelocity(NEW(traj_max_velocity))) {
+        UPDATE(traj_max_angular_velocity);
+        if (0 != emcTrajSetMaxVelocity(NEW(traj_max_velocity),
+				       NEW(traj_max_angular_velocity))) {
             if (emc_debug & EMC_DEBUG_CONFIG) {
                 rcs_print("check_ini_hal_items:bad return value from emcTrajSetMaxVelocity\n");
             }
         }
     }
     if (CHANGED(traj_default_acceleration)) {
-        if (debug) SHOW_CHANGE(traj_default_acceleration)
+        if (debug) SHOW_CHANGE(traj_default_acceleration);
         UPDATE(traj_default_acceleration);
         if (0 != emcTrajSetAcceleration(NEW(traj_default_acceleration))) {
             if (emc_debug & EMC_DEBUG_CONFIG) {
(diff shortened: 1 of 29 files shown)


zultron:
This won't work right out of the box:

- Configurations need to add a `MAX_ANGULAR_VELOCITY` parameter in the `[TRAJ]` section
- UIs need to read the parameter when reading `MAX_VELOCITY`
- In my case, the UI treated the max velocity slider as a percentage, and used the setting to scale both linear and angular velocity before calling `command.maxvel(scaled_val_lin, scaled_val_ang)`

zultron:
I see the changes to the Axis GUI actually **are** in place.  However, they're probably untested, and other GUIs haven't been updated.

zultron:
Thanks for the thoughtful comments, @c-morley , and the encouragement, too.  I'll need to find time to think about your `.ini` questions; I remember during implementation it was difficult to figure out in which sections the various parameters belonged.

I'll also have to find time to update the other GUIs besides Axis.  I'm hoping someone will jump in to help out there; it should be easy for someone with prior experience.

andypugh:
On Fri, 9 Nov 2018 at 04:01, John Morris <notifications@github.com> wrote:

> Thanks for the thoughtful comments, @c-morley
> <https://github.com/c-morley> , and the encouragement, too. I'll need to
> find time to think about your .ini questions; I remember during
> implementation it was difficult to figure out in which sections the various
> parameters belonged.
>

2.8 provides as opportunity to move things around in the INI file.  We are
already introducing significant INI file layout changes for JA, and there
is a script to handle the reformatting.
I think that it an INI file entry should move, this is the time to do it.

-- 
atp

zultron:
> 2.8 provides as opportunity to move things around in the INI file. We are already introducing significant INI file layout changes for JA, and there is a script to handle the reformatting. I think that it an INI file entry should move, this is the time to do it.

Good point.  This patch doesn't change existing INI params, but adds one new one.  Like the `MAX_VELOCITY` param, it defaults to 0.1 * googol, so its absence won't break existing configs.

c-morley:
I can help with Gmoccapy and Gscreen GUI changes - but I'm fairly busy for the next month.
Is this something we are trying to get into 2.8 ? I heard a tentative date of end of DEC....

gmoccapy:
I can add that to gmoccapy, just mention witch branch I need to test?

Norbert

Am 10.11.2018 um 05:32 schrieb c-morley:

KurtJacobson:
@zultron This looks great! Thank you.

One comment on the python interface changes: I see that you added a second argument to `comand.maxvel` for setting the angular max velocity. I would prefer to see separate command methods for setting the max linear velocity and max angular velocity.

There are several advantages, one being that UIs that currently use `command.maxvel` would not be effected, and would continue to work as before.

Second, it is most likely the max vel values will be set by UI sliders or the equivalent, meaning that the linear vel and max angular vel will be set independently. The new `command.maxvel` method with two arguments would require the UI programmer to first get the value of the override he did not want to change, resulting in something like this pseudo code:

```python
def on_mav_slider_changed(new_mav):
    """Set max angular velocity."""
    stat.poll()
    current_mlv = stat.max_velocity
    command.maxvel(current_mlv, new_mav)
```

I also think having separate command methods is more consistent with the rest of the changes.

andypugh:
Where are we with this one? Do we want to introduce a new INI item inside the conversion script?

zultron:
I haven't worked on this one for a while, but it's not ready.  I think @KurtJacobson is right, that the commands should be separated so that the GUIs don't have to be modified.

What's a conversion script, for updating user configurations?  That could work:  Add `[TRAJ]MAX_ANGULAR_VELOCITY`, copying the value of `[TRAJ]MAX_LINEAR_VELOCITY`.

Another way to handle the transition is to make this a feature that must be turned on; by default, max angular velocity would simply track max linear velocity, as it does right now.  Then users who care about this can add it to their configurations.

andypugh:
Because of the Joints_Axes and multispindle changes that are in master / 2.8 there is a script that automatically edits the INI and HAL files to deal with the new HAL pin names and config. It would be easy to make this also add a new MAX_ANGULAR_VELOCITY. 
https://github.com/LinuxCNC/linuxcnc/blob/master/scripts/update_ini#L227

petterreinholdtsen:
Is there any hope to brush up this pull request to a point where it can be applied to master?  Is it complete now, or is more work needed before it can be merged into master?

andypugh:
Discussing this at the Norway meeting, we think that this is now too old to merge. 
We like the idea, but this PR isn't mergable.

Issue #485, opened 2018-08-18
TRAJ LINEAR_UNITS in Axis


jethornton:
If you change the [TRAJ] LINEAR_UNITS to in or imperial you get mm. This can be seen by opening the Axis simulation and changing the units. The manual states LINEAR_UNITS = <units> - Specifies the machine units for linear axes. Possible choices are (in, inch, imperial, metric, mm).

PR #527, opened 2018-11-12
Fix for issue #485

phillc54:
Add 'in' and 'imperial' as valid inifile entries for [Traj]LINEAR_UNITS

Signed-off-by: Phillip Carter <phillcarter54@gmail.com>


rene-dev:
Would it not be better to update the docs instead? I dont understand why you need more than 2 names for a setting that does the same thing. or am I missing something here?

rene-dev:
I updated the docs to mm or inch only.

Issue #400, opened 2018-01-20
M6 Remap does not handle pockets correct!


gmoccapy:
If you do remap M6 code, according to the stdglue.py file it should transfer as parameters 

self.params["tool_in_spindle"]
self.params["selected_tool"]
self.params["current_pocket"]
self.params["selected_pocket"]

But LinuxCNC does not handle the pocket relation correct.
Just place tool 3 in pocket 8 and call T3 M6, that should return pocket 8 but does return pocket 3 instead!

Any hint would be appreciated to solve this.

Norbert

andypugh:
On 20 January 2018 at 21:18, Norbert Schechner <notifications@github.com>
wrote:

> But LinuxCNC does not handle the pocket relation correct.
>
> Just place tool 3 in pocket 8 and call T3 M6, that should return pocket 8
> but does return pocket 3 instead!
>
It might work better if you turn on random tool changer.

The stdglue.py does say that the result is probably nonsense:
https://github.com/LinuxCNC/linuxcnc/blob/master/nc_files/remap_lib/python-stdglue/stdglue.py#L163

 The problem is that the tool table version passed about in NML does not
contain pocket info, and what you actually get (on a non-random changer) is
the index of the tool array.

It was one of the things I tried to fix in
https://github.com/LinuxCNC/linuxcnc/tree/andypugh/tooltable but I was
never brave enough to merge that without somebody else testing it, and
nobody ever did.

-- 
atp

gmoccapy:
I played a little bit with the code and pushed my changes as:

newbynobi__master_tool_pockets

I reached a state, that the pockets are reported correct at the remap module, but now the iocontrol pin are wrong :-(

As mentioned I am completely new to C++ so I may produce rubbish, but I am trying hard and Rene_dev told me I must learn C++, so here I am learning.

So could someone push me in the right direction?

Norbert

P.S. I added two configs (axis and gmoccapy) both containing a diagnostics and a rack tool changer tab. (axis is not so beautiful to embed the rack changer but does work after extending the screen.

To use the python tool changer just click on the open button to open the changer cover, after that you can click on an image to load or unload a tool (like you would do on a real machine)
Watch the tool file to see the pockets changing by 100 places.

rene-dev:
The code he is talking about is here: https://github.com/LinuxCNC/linuxcnc/compare/newbynobi_master_tool_pockets#diff-22efa6c95bb01316fb5517cec373f87dL25
[...]

Issue #239, opened 2017-03-03
interp's named variables for tool info are wrong


SebKuzminsky:
Forum user jtc reports a bug with #<_current_pocket>:

https://forum.linuxcnc.org/10-advanced-configuration/32412-bug-in-current-pocket#88904

The [documentation for #<_current_pocket>](http://linuxcnc.org/docs/devel/html/gcode/overview.html#sub:system-parameters) says:

```
#<_current_pocket> - Return pocket number of the current tool.
```

However, on startup the value is 1, which is wrong.  And after Tx M6 the variable holds the index into Interp's tool table, which may differ from the pocket number in the tool table file and thus is also wrong.

To further complicate the issue, Random and Non-random tool changes differ in their behavior here.

SebKuzminsky:
The branch "2.7-tool-info" adds a test that demonstrates the current bogus behaviors.

rene-dev:
https://github.com/LinuxCNC/linuxcnc/blob/c8a718ea374289d661fbf5239096de53eddd3975/src/emc/rs274ngc/interp_namedparams.cc#L664
this returns the index number, not the pocket number.

rene-dev:
fixed in 2.8 and master by dab2a051b30e600ef95fbe911053386ef44631d0

PR #528, opened 2018-11-16
2.7 pocket fix

rene-dev:
this fixes https://github.com/LinuxCNC/linuxcnc/issues/400 and https://github.com/LinuxCNC/linuxcnc/issues/239
It also works on master, but we agreed that it should go in 2.7 as well.


Lcvette:
This is awesome!  Rene, hazzy and I are excited to use your table in the new gui project! You should stop by the qtpyvcp irc and check us out!

rene-dev:
looks like the remap is still broken, there is no test for them. I will add a workaround for that. but the interp variables are ok, just use them in the m6 remap.

rene-dev:
I got another bug reported: on prepare, the pocket numbers are delayed by one in the remap. No idea why the tests didnt catch that...

Lcvette:
Ah gotcha!  well I am glad to see you are back at it again!  Hazzy and Turboos and myself are VERY excited bever the less for the new tool teable when you have time for it and are eager to be able to implement it into the new gui for the ATC functionality which we are struggling with due to the old tool table.  Between Your Efforts on the new tool table and Hazzy's and Turboss's efforts on the QtPyVCP, LCNC is on its way to a really awesome place!  so at least from me and I know Hazzy and Turboss, a VERY BIG THANK YOU!!!!

TurBoss:
@rene-dev : hi,
    I just noticed that pocket_prepped return the index of stat.tool_table wich is a list that skips empty slots

does this make any sense

KurtJacobson:
@TurBoss I think 4e86bab fixed that problem by adding `pocketno` to the tool table struct so we don't have to rely on the index anymore, which is only valid if there are no "gaps" in the tooltable.

plmicia:
Hi Guys,
I am not sure but I think I found where pocket no. is linked with Remap.
Only one line to modify in 'interp_find.cc' (insrc/emc/rs274ngc):
```
int Interp::find_tool_pocket(setup_pointer settings, int toolno, int *pocket)
{
   if(!settings->random_toolchanger && toolno == 0) {
      *pocket = 0;
        return INTERP_OK;
    }
    *pocket = -1;
    for(int i=0; i<CANON_POCKETS_MAX; i++) {
        if(settings->tool_table[i].toolno == toolno)
//it was:            *pocket = i;
//it should be:
                     *pocket = settings->tool_table[i].pocketno;
    }
    CHKS((*pocket == -1), (_("Requested tool %d not found in the tool table")), toolno);
    return INTERP_OK;
}
```
Now when I issue Mx Tx command it returns correct pocket values in both iocontrol and remap :-) (at least for me). I have not done any intensive tests so I am not sure if this change has any side effects.
My compile conditions:
- Random_toolchanger = 0
- Commit like in #528 [4e86bab](https://github.com/LinuxCNC/linuxcnc/commit/4e86bab515df37e2713c537c0161518109b0dbd2)

I will be appriciate if somebody could doublecheck that and commit if it is ok.
@gmoccapy @rene-dev @SebKuzminsky @andypugh @KurtJacobson @lcvette

rene-dev:
Dont do that, it has very bad side effects. I think I have a fix for it.

Lcvette:
> this is not the new tooltable, thats next, after another cleanup... this only fixes interp and remaps getting the wrong pocket number, and to have a set of tests that represent the expected behavior.

@rene-dev 

Turboss and I are at the last obstacle of the atc portion of our gui project (for now) and it is dealing with the Interp and Turboss has some questions on it and we were hoping perhaps you could enlighten us so we could complete this leg of the project.  

Turboss's question:  "rene-dev: how can we use the bundled boost python interpreter from within systems python?

I would like to read parameters"

we are in the QtPyVCP IRQ on matrix where we collaborate, if you have a spare moment and could maybe nudge us in the right direction we would be grateful!

Thanks Rene!

Chris

rene-dev:
It should all work now, even remaps now get the correct pocket number. the interp variables are not correct if you remap T. This is expected, as they are set by the internal T call, which is not executed if you have a remap. Im still looking for a nice way of fixing this.

gmoccapy:
As you mentioned the T-remap does not work as it should, but remaping only M6 all is good.
I checked with all my possible configs and all do work as expected, even if you change tool pocket relation within a gui. I tested with non random and random tool changer!

Very good job rene-dev, thank you.

Norbert

mydani:
Hi Rene,

I think in interp_convert.cc in line 5256 is an issue, the variable pocket (little misleading) contains the tool table index, but not the actual pocket. Instead, the find_tool_pocket(...) should be used to find the real pocket for the given tool and this should be assigned to settings->selected_pocket.

Can you confirm?

Regards,
Daniel

aShure:
T0M6 and M61Q0 are broken: pocket from the previous tool is kept, not updated to pocket 0.

rene-dev:
This is now in 2.8 and in master.

PR #530, opened 2018-11-25
Fix locale string extraction error

Solitarily:
This is a bug that caused the string that “pncconf” needs to translate not to be extracted, now I fixed it.


c-morley:
Thank you for your fix.

PR #531, opened 2018-11-25
Add a symbol to the string so that it can be added to linuxcnc.pot

Solitarily:
Add a symbol to the string so that it can be added to linuxcnc.pot


jepler:
Looks good to me.

PR #532, opened 2018-11-26
Change the string "Home" to "home" to distinguish translations.

Solitarily:
There are several places where the code uses the "Home" string, but they represent different meanings, such as "machine reference point" and the keyboard "Home" button. So I changed it to "home" so that I can distinguish when translating. This change will not affect translations in other countries, because "home" is already used in "tklinuxcnc.tcl", which has the same meaning as the string I changed, and stands for "machine reference point".

### DIFF
--- a/src/emc/usr_intf/pncconf/private_data.py
+++ b/src/emc/usr_intf/pncconf/private_data.py
@@ -445,7 +445,7 @@ class Private_Data:
         bhome = [[_("Main Axis"),home],[_("Tandem Axis"),home2]]
         bshared = [[_("Main Axis"),home_limits_shared],[_("Tandem Axis"),home_limits_shared2]]
         self.human_input_names = [ [_("Unused Input"),S.UNUSED_INPUT],[_("Limits"),blimits],
-            [_("Home"),bhome],[_("Limts/Home Shared"),bshared],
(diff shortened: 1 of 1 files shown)


jepler:
I looked into the issue of translation "context" a little bit.  In GNU gettext, there are functions such as 'pgettext' which take two arguments, with one of them being a context: https://www.gnu.org/software/gettext/manual/html_node/Contexts.html -- read the link for more info about how this works.

Unfortunately, in Python's wrapping of gettext, a similar function is not supported.

The implementation of 'pgettext' is to concatenate <context> <ctrl-d> <msgid>, look it up in the catalog.  If it's there, that string is returned; otherwise, the original <msgid> is returned.

It is possible to "monkey patch" python's gettext module to support pgettext, and this is definitely something to consider, but it is also very expedient to adopt a convention like this if it helps improve the quality of the translations with such a small amount of effort.

jepler:
so e.g., for instance if someone contributed an implementation of pgettext, we would establish that there is a 'keycap name' context, so you would pgettext("keycap", "Home") instead of just writing _("Home")

jepler:
@c-morley any objection to merging this?

c-morley:
I agree long term something better would be nicer. Pncconf's life span is coming to an end fairly soon.
This change is a good compromise.
Thank you.

PR #534, opened 2018-11-26
Add toggle to display/hide pyVCP panel

chaotix-:
Hiding the pyVCP panel can be useful for small screens

Ctrl-E was the first free shortcut I found. Unfortunately P, Y, V and C are all already in use.

### DIFF
--- a/share/axis/tcl/axis.tcl
+++ b/share/axis/tcl/axis.tcl
@@ -380,6 +380,12 @@ setup_menu_accel .menu.view end [_ "Large coordinate fo_nt"]
 	-command clear_live_plot
 setup_menu_accel .menu.view end [_ "_Clear live plot"]
 
+.menu.view add checkbutton \
+	-variable show_pyvcppanel \
+	-accelerator [_ "Ctrl-E"] \
+	-command toggle_show_pyvcppanel
+setup_menu_accel .menu.view end [_ "Show pyVCP pan_el"]
+
 .menu.view add separator
 
 .menu.view add radiobutton \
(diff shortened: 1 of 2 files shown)


chaotix-:
You are right about two and a half points. :-) Only pressing the keyboard shortcut will not generate an exception because ctrl-e will not be bound if there is no panel.

I added the menu option as an afterthought and did not think about vcp_frame not being defined anymore when using the menu, so yes it will throw an exception. That will be easy to fix.

I'm not really familiar with Tkintern, but I will see if I can find a fix for not showing the menu item and the shortcut in the quick reference.

jepler:
Thanks, I appreciate you taking the time to look into it and improve this pull request.

jepler:
In `axis.py` you can see how items are sometimes deleted from menus, look for the `widgets.menu_view.delete` calls.  However, this approach is becoming a maintenance problem as the effect of a call like `...delete(0,5)` conditional on `if lathe_backtool...else` means that the numeric value for deleting the vcp menu item can't be chosen as a simple hard-coded number ☹️

chaotix-:
Thanks for your help.

Yeah. I figured out how to delete an entry with a fixed position, but I have not found a method yet to delete a entry based on the label or what the label should be like.

According to https://stackoverflow.com/questions/21906815/tkinter-menu-delete-item deleting by label should be possible, but using "Show pyVCP pan_el" does not work. Neither does adding a "-label foo" to axis.tcl and then calling .delete("foo").

But I have figured out how to add a shortcut to the quick reference dynamically. Although it will always be at the last position in the right column as I just append it to help2.

chaotix-:
OK. I got it. I have to drop the "_", so it just has to be "Show pyVCP panel".

Should I use a hack like `_("Show pyVCP pan_el").replace("_", "")` so it just has to be translated once?

PR #536, opened 2018-12-02
Add international support for classicladder

Solitarily:
Add internationalization support for classicladder and add specific symbols for strings that need to be translated.


c-morley:
Thank you for your work!

Issue #539, opened 2018-12-09
Add to Readme purpose and type of this software


matkoniecz:
I would be happy to make a PR if that would be desirable (let me know if that is the case). For start 

> LinuxCNC controls CNC machines. It can drive milling machines, lathes, 3d printers, laser cutters, plasma cutters, robot arms, hexapods, and more. http://linuxcnc.org/

may be a good initial readme content (currently it has solely disclaimer and link to build instruction).

matkoniecz:
And for the type of software I am still trying to confirm what it is.

- text on http://linuxcnc.org/ makes clear that it is a software
- http://linuxcnc.org/docs/2.7/html/getting-started/getting-linuxcnc.html - distributed as live CD so it is either separate OS or software hard enough to install that it is distributed as live CD or maybe software directed toward people with Windows and unwilling to install linux.
- further installation instructions suggest that it is actually impossible to install as normal software and requires separate OS (`Reboot, and select the Linux 3.2.0-4-rt-686-pae kernel. `)

andypugh:
LinuxCNC is a software  application that is dependent on a specific type of non-standard Linux kernel to work. (ie one with real-time extensions, PREEMPT-RT, RTAI or Xenomai) 
The kernel patches do not interfere with the other OS functions in any way, so the Linux install appears normal.

matkoniecz:
And it is a monolithic application, so it is impossible to install just parts that do not require special kernel features?

Or it possible to install just some tools?

andypugh:
No, it is made up of many parts, and these can be used independently in a modular way. 
But it doesn't really do anything useful without the realtime capability.

matkoniecz:
So it impossible to generate gcode with http://linuxcnc.org/docs/2.4/html/gui_image-to-gcode.html export it and use it on machine that operates CNC cutter?

Is it limitation of LinuxCNC or is my understanding of gcode as instructions executed by a CNC machine incorrect?

rmu75:
Usually, a linuxcnc system is a fully functional linux system.  You can install and run whatever linux stuff you like.

As andypugh said, operating a real machine requires the use of a non-standard kernel that has certain realtime-capabilities, but those capabilities do not interfere with other OS functions in any way.

Software that requires proprietary kernel modules may be a problem, but this stuff is to be avoided anyways.

andypugh:
I no longer believe that I understand the context of this discussion. 
[...]

PR #540, opened 2018-12-10
add to readme purpose of this software

matkoniecz:
fixes part of #539
text is copied from repository description at GitHub


andypugh:
I think that this makes sense, it does seem the most logical place to look for a "what does this software do" if browsing github.

PR #541, opened 2018-12-13
add ~/.local/bin to PATH in linuxcnc start script

KurtJacobson:
There is a regression in bash shipped with Debian Stretch which results in `~/.local/bin` not being on the path. This is the location to which pip installs python entry points, so is needed for linuxcnc to be able to find and launch some python based displays distributed via pip .


jethornton:
I think this should test for the existence of the directory first.

`
if [ -d $HOME/.local/bin ]; then
    if [[ "$PATH" != *".local/bin"* ]]; then
        PATH=$HOME/.local/bin:$PATH
    fi
fi
`
I'll test that today on a Debian 9 PC.

KurtJacobson:
@jethornton Good point! I have updated the PR. I have tested on Mint 19.1 and it works as expected, only adding .local/bin if it exists and is not already in the PATH.

PR #544, opened 2018-12-19
Fix spindle override in halui.cc

antbarney:
I found this little issue well setting up my pendant and saw the spindle override was always resetting to 1% on first use of the wheel. looks like it was a simple miss when updating these for having multiple spindles.


andypugh:
Yes, looks like an error on my part.

PR #552, opened 2019-01-31
Axis: justify text correctly for jog speed scales and max velocity scale

phillc54:
Signed-off-by: Phillip Carter <phillcarter54@gmail.com>


andypugh:
Looking at screenshots I can see the problem you are trying to address. 
But, also looking at screenshots, it seems to depend on the language (and label length) used. 
Does your change work in other locales?

phillc54:
No, I have not tried other locales nor do I know how to.
I just changed the left, right packing to the same as the feed, rapid and spindle overrides.

andypugh:
So, do you want to merge this without checking other languages / locales?

phillc54:
Sorry Andy, I have been really busy on other stuff and haven't had time to do anything with this.
I think it best if I close this PR and leave it be as it is only a cosmetic change and no-one seems bothered by it.

PR #554, opened 2019-02-01
Zultron/lcnc pr docker cleanup fixes

zultron:
Here are a few cleanups to the `scripts/linuxcnc.in` and `scripts/realtime.in` templates.

The most significant update is ignoring defunct zombie processes when cleaning up on exit.  These hang around in Docker containers, and aren't a problem except where they trigger false positives in scripts like these.  More info [here][1].

The others are minor:  Use the autoconfigured `grep` executable, and fix a log message.

[1]: https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/

### DIFF
--- a/scripts/linuxcnc.in
+++ b/scripts/linuxcnc.in
@@ -528,9 +528,9 @@ function KillTaskWithTimeout() {
 	if [ $WAIT -gt 0 ] ; then
 	    echo "Could not kill task $KILL_TASK, PID=$KILL_PID"
 	fi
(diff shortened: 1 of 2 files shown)


zultron:
Tests passed.

zultron:
Hold off on this, one more fix coming up

zultron:
There, `runtests` now run in Docker as well.  Tests pass, ready for review.  Thanks!

PR #557, opened 2019-02-05
Fix in carousel.comp for individual sensor position detection

inoxix:
The current position with encoding mode "single" was not detected correctly.


andypugh:
If you say it works, then that’s good enough for me.

PR #568, opened 2019-03-09
Add Tk options and a doc string to pyvcp table widget.

yoleg:


andypugh:
Looks reasonable, but can you give an example of what the kw argument does?

yoleg:
@andypugh The kw argument means that any attributes or child tags that are not listed in the arguments get passed along to Tk's Frame class. So any valid Frame arguments become valid options. The docstring I added already contains a couple of examples:  `highlightthickness="10" `and `highlightbackground="#CCCCCC"`. The pyvcp_label class already does the same thing with Tk's Label class (and personally I think all the classes should do this).

andypugh:
Thanks for the pull request.

PR #580, opened 2019-04-25
time.comp: add support for pausing timer.

itaib:


mozmck:
itaib, I just found your change here after I modified time.comp in a similar manner in 2.7.  My change makes the not component unnecessary because time.N.start is connected to halui.program.is-running.  I think the component should be the same in all branches, and I kind of like eliminating the not component, so would it be a problem for you if I changed your version in master to my version?

itaib:
no problem.

mozmck:
I merged my changes with yours and I think it's better yet - thanks!

itaib:
I believe you need to use your code from 2.7.
right now, pausing/unpausing resets the counter.

PR #581, opened 2019-05-02
Fixes for issues 68, 164, 167, 546, and 550

### LINKED ISSUE #68: G33 issue with Ver 2.7 TP
1. Run the G code below on Lathe.

G20 G64 G8 G18
M3 S400
G0 x.5 z.21
G1 x.490 z.2
/G4 P.01 ( need pause or G0 above for ver 2.7 TP)
G33 k.086 x.500 z-.550
G0 x.510
G0 z.21
m5
m2
1. If I have a G1 move leading into a G33 it makes a radius then 'Z' over accelerates then slows.
2. I can use G0 or put G4P.01 before G33 then OK. Only happens with spndl speed over 200~ rpm  

```
 feed to the G33 starting point then accel to K value.
```

   G1 moves toward G33 start point makes a radius then 'Z' over accelerates then slows.
   After the over accel it slows to proper rate for the remainder of G33 move.

  In previous versions or with old (Pre 2.7) TP

   Machine 1: Using version 2.7.4 Ubuntu Lucid with new TP, Mesa 5i20 & servo card
   Machine 2: Using version 2.7.4 Debian Wheezy with new TP,  Mesa 5i25/7i77  
- I am using this AXIS user interface (GUI)

### LINKED ISSUE #164: Excessive jitter spindle position tracking

### LINKED ISSUE #167: Warn users that spindle is too fast in G33 / G95

### LINKED ISSUE #546: Acceleration violations in linuxCNC 2.7 and 2.8
Both current versions of LinuxCNC have acceleration constraint violations in some circumstances
While testing a forum posters gcode files ( https://forum.linuxcnc.org/9-installing-linuxcnc/35744-not-a-bug-report ) i found that I could get acceleration violations with one of the test files: \

http://freeby.mesanet.com/1.ngc

I have aggressive velocity and acceleration limits on my test configuration: 20 IPS and 200 IPS^2
and that may play a part, but this error is reproducible in both  2.7 and 2.8 (first error around line 920)

http://freeby.mesanet.com/28ferror.png
http://freeby.mesanet.com/28ferrorb.png

The second plot has the ferror limits widened enough so the following error fault is not generated

### LINKED ISSUE #550: Acceleration violation with very short XYZ motion and large ABC motion
Following errors are likely to occur due to acceleration violations when running this G-code snippet:
```
G0 X0 Y0 A0
G91 G20

G1 X0.1 Y0.1 A1.0 F100
G1 X0.000000005 Y0.000000005 A100.0 F10000
G90
G1 X0 Y0 A0 F10000
M2
```

 1. Run LinuxCNC in simulation or real hardware on a machine with at least one linear and at least one rotary axis. It's easier to see with Sam's HAL setup to compute accelerations for each axis.
 2. Run attached program
 3. Acceleration violation (or following error) at line 5

Not sure, but likely for most 2.7 releases

Running simulation config in a user-space build (on Ubuntu 18.04) in tests/trajectory_planner/circular_arcs/configs/XYZ.ini (bit of a misnomer since it has an A axis). The particular platform should not matter for this issue, though it would be interesting if it was affected by compiler / choice of kernel.

robEllenberg:
This branch combines a bunch of old work, and some recent fixes from other branches to address some outstanding TP issues in 2.7

Fix #68 (reduce large blend arcs in G33 and prevent blends between synced and non-synced modes)
Fix #164 (smoother error correction and better velocity estimation in position sync)
Fix #167 (warns users if axis feeds are not fast enough for a G33 move at the current spindle speed)
Fix #546 (Fix some edge cases in TP including a regression in parabolic blend fallback)
Fix #550 (Fix mismatch between canon and TP's idea of what a zero-length move is, since they each have to arrive at that conclusion independently)

Note: all runtests and TP unit tests pass in a Ubuntu 18.04 x64 uspace build


samcoinc:
- tp: disallow blending when entering position-synch mode

Have not had time to play with this - but does this mean that things like a
fuse won't work any more?  (like combining a bunch of G33 commands together
to create a variable pitch/diameter thread?

samcoinc:
example
https://youtu.be/ACvRilmIKDQ?t=75

samcoinc:
ok - maybe I should have read down more?


   - canon: Enable smarter blending in spindle-synchronized motion.

sorry

robEllenberg:
You should still be able to blend between position-sync moves, but the
transition in and out (i.e. G1 to G33) won't blend (since you need to wait
for index anyway).

Rob

mozmck:
@robEllenberg what is the status of these fixes?  Have they been merged somewhere?  I have a problem on 2.7 with a pause in motion on an ellipse and I'm wondering if a fix for this is somewhere in these fixes.

robEllenberg:
I opened a pull request for #68 and backported the fix for #704 in another PR for 2.7. I need to pull out the blending fix from this branch and merge it separately because the spindle-sync stuff in the branch is too big a change for a point release in 2.7. My plan is to

petterreinholdtsen:
Is there any chance that this huge collection of independent changes can be split up into individual change sets that are easier to review, to increase the chance of these fixes making it into the master branch?

petterreinholdtsen:
Rene has volunteered to try to figure out this patch.

petterreinholdtsen:
@rene-dev, did you have any luck figuring out this patch?

petterreinholdtsen:
Any news here?

SebKuzminsky:
#550 was fixed in 2.7 by @robEllenberg here: https://github.com/LinuxCNC/linuxcnc/commit/10c3678ab069f856532e5512a93bfaffcf930491

BsAtHome:
Last part was fixed with #4441.

PR #583, opened 2019-05-07
fix importing of certain stls in vismach

sleepybishop:
i'm trying to integrate a 6dof robot arm with linuxcnc, been having problems with kinematics so i thought i'de try and model in vismach. once in a while an stl would fail to import, with the small changes here the failed stl imports went away.

submitting for merge consideration in case it's helpful, if not please disregard.

### DIFF
--- a/lib/python/vismach.py
+++ b/lib/python/vismach.py
@@ -919,7 +919,7 @@ class AsciiSTL:
                         dx2 = t[2][0] - t[0][0]
                         dy2 = t[2][1] - t[0][1]
                         dz2 = t[2][2] - t[0][2]
-                        n = [y1*z2 - y2*z1, z1*x2 - z2*x1, y1*x2 - y2*x1]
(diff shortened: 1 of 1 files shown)


andypugh:
Are you convinced that this is always better than the original version? 
As far as I can tell the only sample config that uses an STL is VMC_Toolchange.

sleepybishop:
yes i'm certain, before this code path would error out because [x1,x2,y1,y2,z1,z2] were not defined.

an example stl to trigger the problem can be found here:
https://github.com/sleepybishop/linuxcnc/blob/fishybearG-2.8/configs/sim/axis/vismach/ar2/ar2_link3.stl

andypugh:
OK, if the original was actually buggy and crashy, then we should merge it. 
Next question, though. If this is a bugfix then it should go in to 2.7 then be merged up to 2.8 and 2.9. 
Can you possibly re-submit it against 2.7?
(if that's too much trouble then I can do it)

PR #584, opened 2019-05-09
Add yapps2 as optional yapps binary

awigen:
Fedora (and others) do not package YAPPS. When installing from pip the
binary is named yapps2. Modify the binary search to search for either
binary name.

Also fix a spelling mistake.

Signed-off-by: Alex Wigen <alex@wigen.net>


zultron:
+1

PR #585, opened 2019-05-10
Use autoconf provided name for yapps

awigen:
While autoconf will correctly identify yapps or yapps2, the provided
variable should be used in the Makefile.

Signed-off-by: Alex Wigen <alex@wigen.net>


andypugh:
Hmm, Travis seems unhappy, but that does not immediately appear to be related to your change.

awigen:
Sorry about that, I forgot to export the YAPPS variable

PR #589, opened 2019-05-13
Rellenberg/bugfixes 546 550 167 68 164 for 2.7

rene-dev:


rene-dev:
Im only creating this PR to see the problems that will come when 2.7 with this is merged into master.
Some of it is the multi spindle support, and my rigid tapping fast retract.
but all in all it looks manageable.

PR #593, opened 2019-05-17
Enhanced Renes changes to the non-random tool changer so the tool id …

mydani:
…/ pockets are correct with internal as well as remapped m6. Non-random means:

Tool is loaded from and returned to the same pocket as given in the tool table
Tool 0 is special -> no tool.
Pocket 0 is special -> do not return to ATC but manual change instead (ToDo).
Current_Pocket -> original (tool table) pocket of the tool currently in spindle.

Signed-off-by: mydani <mydani@gmx.de>

### DIFF
--- a/src/emc/nml_intf/canon.hh
+++ b/src/emc/nml_intf/canon.hh
@@ -475,7 +475,7 @@ extern void USE_SPINDLE_FORCE();
 extern void USE_NO_SPINDLE_FORCE();
 
 /* Tool Functions */
-extern void SET_TOOL_TABLE_ENTRY(int pocket, int toolno, EmcPose offset, double diameter,
+extern void SET_TOOL_TABLE_ENTRY(int tool_index, int pocket, int toolno, EmcPose offset, double diameter,
                                  double frontangle, double backangle, int orientation);
 extern void USE_TOOL_LENGTH_OFFSET(EmcPose offset);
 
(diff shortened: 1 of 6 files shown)


mydani:
Tests are still left untouched, might fail. Will adapt them ina new commit.

rene-dev:
It might work, but adding the index everywhere is not the way to go.
the index should go entirely, as it is not required, and will be removed in a future version.

rene-dev:
there are no tests checking for what you are fixing. you need to cherrypick and fix sebs toolinfo test from my branch to actually test the tool info.

mydani:
> It might work, but adding the index everywhere is not the way to go.
> the index should go entirely, as it is not required, and will be removed in a future version.

Can you give me a pointer to the recent implementation? Tool table index = pocket sucks imho ;)

rene-dev:
https://github.com/LinuxCNC/linuxcnc/pull/528/
there are a few issues left, and a lot of unpushed code which fixes them that Im still cleaning up.
the tool number is the only thing that should get passed around, as its unique, the only thing that the interpreter cares about, and you can get everything from the tool number.
This is because Im also preparing the code for the new sqlite based tooltable, which has pockets, but no index. this also allows to add arbitrary fields, like max rpm, tool hours, ...

mydani:
Sounds very reasonable. Let me know when you need some support or a reviewer. I will close the pull request and keep the SW running at my machines until your branch is available.

PR #594, opened 2019-05-21
num_sserials=N is not used in hostmot2.c

sleepybishop:
num_sserials config parameter doesn't seemed to be used by the code, this PR removes it from the docs.


andypugh:
I think that this is right and 
https://github.com/LinuxCNC/linuxcnc/commit/ade72db61014152e1f78dd6112dd321474217b5c#diff-91a3d2867cf0072d0f9d88cbb4c4f1ff
was wrong

num_sserials is used internally, but the parameter string parsing looks for "sserial_port_" rather than "num_sserials"
https://github.com/LinuxCNC/linuxcnc/blob/0f91c553a238c3f5e8a52285044761c2dcfd7de5/src/hal/drivers/mesa-hostmot2/hostmot2.c#L415

PR #603, opened 2019-06-09
docs -  Correct the Python Interface example function ok_for_mdi

manymachines:
This PR corrects the logic in the ok_for_mdi() example function.  I will submit a separate request for 2.8.


jethornton:
Tested and looks good to me.

PR #604, opened 2019-06-09
fix bug in vismach stl import

sleepybishop:
fix undefined variables causing some stl imports to fail in lib/python/vismach.py

resubmitting #583 against 2.7 per @andypugh 's suggestion


sleepybishop:
@andypugh what's the process for merging up to 2.8/2.9 ?

andypugh:
I already merged it to 2.8. I was about to merge to 2.9 when I decided that I really should go to sleep.

PR #605, opened 2019-06-09
docs - Correct the Python Interface example function ok_for_mdi

manymachines:
This PR corrects the logic in the ok_for_mdi() example function.   I filed a separate request for 2.7.

### DIFF
--- a/docs/src/config/python-interface.txt
+++ b/docs/src/config/python-interface.txt
@@ -555,7 +555,7 @@ c = linuxcnc.command()
 
 def ok_for_mdi():
     s.poll()
-    return not s.estop and s.enabled and s.homed and (s.interp_state == linuxcnc.INTERP_IDLE)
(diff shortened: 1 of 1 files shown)


jethornton:
Both PR's are on Master and it looks like your repo was not up to date. You should git pull --rebase before submitting the PR and git checkout 2.7.

manymachines:
I changed the base branch of the PR to 2.8.  Is that better?  If so I can try making the same change to the other PR that was intended for the 2.7 branch.

manymachines:
I filed the separate PR for 2.8 because the 2.7 example will not work in 2.8.

For example consider a JOINTS=4 machine with trivkins coordinates=XYYZ in 2.8.

s.axes in 2.8 will return 3
s.joints (new in 2.8) will return 4.  This is what we need to compare with s.homed.count(1) for 2.8.

PR #613, opened 2019-06-17
docs: Correct alphanumeric order of Python interface attributes/keys

manymachines:
This change corrects the order of the listed Python interface attributes and dictionary keys.

For example the list of linuxcnc.stat attributes contains 82 items where 3 of them appear in a random order.  
The joint dictionary had 1 out of order item out of 22 items (see jointType).  Also corrected the spindle dictionary key ordering.


andypugh:
Thanks.

PR #614, opened 2019-06-18
Added support for Devuan

Centuriondan:
These commits:
1) adds support for detecting Devuan GNU/Linux versions into the debian/configure script
2) fixes the "clean" stage so it doesn't try to run autoconfig.sh  (breaks building with pbuilder because deps are resolved after the clean stage.
- this meant making some of the clean steps result be ignored because it tries to clean files that haven't yet been created.

### DIFF
--- a/debian/configure
+++ b/debian/configure
@@ -156,16 +156,16 @@ PYTHON_IMAGING="python${PYTHON_VERSION}-imaging | python-imaging | python-pil"
 PYTHON_IMAGING_TK="python${PYTHON_VERSION}-imaging-tk | python-imaging-tk"
 
 case $DISTRIB_NAME in
-    Debian-10|Debian-10.*)
+    Debian-10|Debian-10.*|Devuan-3|Devuan3.*)
         DOC_DEPENDS="$DOC_DEPENDS, asciidoc-dblatex"
         PYTHON_GST=python-gst-1.0,gstreamer1.0-plugins-base
         EXTRA_BUILD=python-yapps
         ;;
-Debian-9.*|Debian-10|Debian-10.*|Raspbian-9.*|Debian-testing|LinuxMint-19.*|Ubuntu-18.*)
+Debian-9.*|Debian-10|Debian-10.*|Raspbian-9.*|Debian-testing|LinuxMint-19.*|Ubuntu-18.*|Devuan-2.*)
         DOC_DEPENDS="$DOC_DEPENDS, asciidoc-dblatex"
         PYTHON_GST=python-gst-1.0,gstreamer1.0-plugins-base
         ;;
-Debian-8.*|Raspbian-8.*) # Jessie
+Debian-8.*|Raspbian-8.*|Devuan-1.*) # Jessie
         ;;
 Debian-7.*|Raspbian-7) # Wheezy
         EXTRA_BUILD=libgnomeprintui2.2-dev
--- a/debian/rules.in
+++ b/debian/rules.in
@@ -44,17 +44,9 @@ clean: debian/control
 	dh_testdir
 	dh_testroot
 	rm -f build-stamp
-
-	cd src && ./autogen.sh
-ifeq "$(kernel_version)" "uspace"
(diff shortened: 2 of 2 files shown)


andypugh:
Are you sure about the "clean" changes? I have not noticed a problem with it, and it seems likely that the code was put in for a reason.

SebKuzminsky:
The `autogen` and `configure` steps produce the `Makefile.inc`, without which `$(MAKE) clean` will fail.

SebKuzminsky:
I should also mention that we currently build packages using pbuilder and it works just fine, eg: http://buildbot.linuxcnc.org/buildbot/builders/4031.deb-stretch-rtpreempt-amd64/builds/800

Centuriondan:
> The `autogen` and `configure` steps produce the `Makefile.inc` file, without which `$(MAKE) clean` will fail.  
That is why I made the "rm" commands result ignored by prefixing the "-" 

It gets autogen gets run during the build phase anyway.

Centuriondan:
> I should also mention that we currently build packages using pbuilder and it works just fine, eg: http://buildbot.linuxcnc.org/buildbot/builders/4031.deb-stretch-rtpreempt-amd64/builds/800

Interesting.... because for me it fails because the configure step fails due to missing dependencies.  Is your pbuilder environment populated with some of the dependencies already??

SebKuzminsky:
@Centuriondan and I spoke about this some on IRC and discovered a difference in how we build debs, that may be responsible for this problem.

The [LinuxCNC Buildbot](http://buildbot.linuxcnc.org/buildbot/waterfall) uses this package build process:

1. Configure source for the current distro: `cd debian; ./configure`
1. Install missing build dependencies: `pbuilder-satisfydepends`
1. Build source package: `dpkg-buildpackage -S`
1. Build binary package (in pbuilder): `pbuilder build`

In contrast, @Centuriondan tells me he uses git-buildpackage, using a process like this:

1. Configure source for the current distro: `cd debian; ./configure`
1. Build source & binary packages (in pbuilder): `gbp buildpackage --git-pbuilder`

However, i just tried building our master branch (without @Centuriondan's PR) and it built fine.  I ran these commands:

1. `sudo git-pbuilder create`
1. `cd linuxcnc-dev`
1. `git checkout master`
1. `cd debian; ./configure uspace; cd ..`
1. `sudo gbp buildpackage --git-pbuilder --git-ignore-branch`

This seemed to build the dsc on the host machine (not in pbuilder), similar to the process our buildbot uses.  @Centuriondan, please correct me if i'm wrong in that assessment.

Building the dsc of course requires making clean, to ensure  that any old build artifacts lying around in your working directory don't get accidentally mixed in to the clean pbuilder build.  My build machine has the linuxcnc build dependencies installed, which is why the dsc build worked.  A clean pbuilder environment is expected to not have these dependencies installed, so if you somehow build the dsc inside pbuilder it'll fail.  This must be how @Centuriondan does his build, though i haven't been able to replicate it.

The second commit in this PR ("fix clean routine") breaks `debian/rules clean` by making it try to run `make clean` without running autogen and src/configure to create `Makefile.inc`.  This results in an error:
~~~
$ fakeroot ./debian/rules clean
dh_testdir
dh_testroot
rm -f build-stamp
cd src && /usr/bin/make clean -s
Makefile:82: *** Makefile.inc must specify RTPREFIX and other variables.  Stop.
make: [debian/rules:47: clean] Error 2 (ignored)
rm -f Makefile.inc
rm -f src/config.log src/config.status
dh_clean
~~~
The `make clean` invocation errors out before cleaning up, and old object files and other junk is left behind in the build directory while constructing the source package.  I think this current approach is a mistake.
I would welcome a PR with just the first commit ("Added support for Devuan"), but the second commit shouldn't go in, as it currently exists.  The Devuan support PR should be made against 2.8.

Centuriondan:
@SebKuzminsky Your right, it does build them locally first, but I don't have ALL the deps installed locally.  (And shouldn't need them for a source build).  FWIW, Devuan builds the source package from the source in a buildhost again without the deps being installed.  

It still seem wrong to me to have make clean running autoconf to generate files it needs to delete.
my approach makes deletion of the generated files non-fatal if they fail because the files aren't there.

That said, what I'll have to do to for Devuan will be a derivation from LinuxCNC's buildbot setup anyway.

PR #619, opened 2019-06-28
Add support for Catch2-based unit tests for interp

robEllenberg:
Adds catch2.hpp and a minimal framework to do unit testing on interp. If this gets general approval, I have some additional unit tests based on this to backport from another branch. Currently the unit test build is implemented in Meson as a parallel build system.


sebastiankuzminsky:
Why are you adding a parallel build system?

robEllenberg:
@sebastiankuzminsky Great question! It seems like an odd choice, I know, but it was both because of my frustration with the current build system, and curiosity about meson as the up-and-coming "build system of the future". I did a parallel build system to avoid risking disruption to the existing kernel-space builds. The unit tests have minimal dependencies, and are only a few files each of actual code, so it was a good toy problem to try a new build system for. In the future, it could grow to be the main build for LinuxCNC (like various gnome projects have done), or we could migrate it all to CMake. Porting from Meson to CMake (especially for a small build) is probably easier than the equivalent port from autotools, too.

As for why I chose Meson vs other options, I wanted to get some experience with it and see if the claims were true. So far, I've been impressed with it. It's easy to diagnose and add missing dependencies, and it has useful intermediate structures (e.g. file name / include path containers that remember the full path, dependency definitions as an object). The syntax that is intuitive (at least for a python user). I'm not yet sure how well it will scale, but it's been fast and robust so far. Dependencies are easy to manage too because it requires all source files to be explicitly defined in the build files. This is in theory more painful than globbing sources with Make, but it's very good about not making you repeat yourself. A properly-laid-out dependency tree only needs one mention of a source file name, for example.

src/emc/pythonplugin/python_plugin.cc
```diff
@@ -48,8 +49,7 @@ namespace bp = boost::python;
 	}								\
     } while (0)
 
-
-extern const char *strstore(const char *s);
+static const char *strstore(const char *s);
```
mozmck: Where was strstore defined before?

src/emc/sai/saicanon.cc
```diff
@@ -143,178 +101,117 @@ void print_nc_line_number()
...
-           fprintf(_outfile, control,                                       \
-                   arg1,arg2,arg3,arg4,arg5,arg6,arg7,arg8,arg9,arg10,arg11,arg12,arg13,arg14); \
-          } else
+#define PRINT(control, ...) do \
+{ \
+    _outfile = _outfile ?: stdout; \
+    fprintf(_outfile,  "%5d ", _sai._line_number++); \
+    print_nc_line_number(); \
+    fprintf(_outfile, control, ##__VA_ARGS__); \
+} while (false)
```
mozmck: Nice!

src/emc/sai/saicanon.cc
```diff
@@ -519,147 +408,142 @@ void STRAIGHT_PROBE(int line_number,
...
                                                    CANON_COUNTERCLOCKWISE);
 }
 
 void SET_SPINDLE_SPEED(int spindle, double rpm)
 {
-  PRINT2("SET_SPINDLE_SPEED(%i, %.4f)\n", spindle, rpm);
-  _spindle_speed[spindle] = rpm;
+  PRINT("SET_SPINDLE_SPEED(%i, %.4f)\n", spindle, rpm);
+  _spindle_speed = rpm;
 }
```
mozmck: Will these removal of [spindle] affect the multi-spindle stuff?  It looks at quick glance that before it was setting rpm and etc for the given spindle - so how are these changes handling it?

src/emc/sai/saicanon.cc
```diff
@@ -1001,40 +889,40 @@ extern int GET_EXTERNAL_QUEUE_EMPTY()
 /* Returns the system value for spindle speed in rpm */
 double GET_EXTERNAL_SPEED(int spindle)
 {
-  return _spindle_speed[spindle];
+  return _spindle_speed;
 }
```
mozmck: Same here as above - does this remove multi-spindle support??

src/emc/task/taskmodule.cc
```diff
@@ -270,58 +271,55 @@ BOOST_PYTHON_MODULE(emctask) {
...
+            .BOOST_PYENUM_VAL(EMC_ABORT_TASK_EXEC_ERROR)
+            .BOOST_PYENUM_VAL(EMC_ABORT_AUX_ESTOP)
+            .BOOST_PYENUM_VAL(EMC_ABORT_MOTION_OR_IO_RCS_ERROR)
+            .BOOST_PYENUM_VAL(EMC_ABORT_TASK_STATE_OFF)
+            .BOOST_PYENUM_VAL(EMC_ABORT_TASK_STATE_ESTOP_RESET)
+            .BOOST_PYENUM_VAL(EMC_ABORT_TASK_STATE_ESTOP)
+            .BOOST_PYENUM_VAL(EMC_ABORT_TASK_STATE_NOT_ON)
+            .BOOST_PYENUM_VAL(EMC_ABORT_TASK_ABORT)
+            .BOOST_PYENUM_VAL(EMC_ABORT_USER)
+            ;
```
mozmck: Is boost_pyenum a new dependency?

src/emc/sai/saicanon.cc
robEllenberg: This is a mistake, good catch! Anything like this that steps on multispindle is a mistake, due to bad resolution of rebase conflicts. Multispindle should not be affected by my changes, so I'll go through and fix these up.

src/emc/rs274ngc/interp_internal.hh
```diff
@@ -95,183 +98,195 @@ static inline bool equal(double a, double b)
...
+    G_92_2 = 922,
+    G_92_3 = 923,
+    G_93 = 930,
+    G_94 = 940,
+    G_95 = 950,
+    G_96 = 960,
+    G_97 = 970,
+    G_98 = 980,
+    G_99 = 990,
+};
```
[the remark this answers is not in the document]
robEllenberg: Modern compilers seem to accept final commas, and it makes version control nicer (since adding a new line doesn't affect the previous lines).

src/emc/nml_intf/emc_nml.hh
```diff
@@ -1486,7 +1486,7 @@ class EMC_TASK_STAT:public EMC_TASK_STAT_MSG {
     int activeGCodes[ACTIVE_G_CODES];
     int activeMCodes[ACTIVE_M_CODES];
     double activeSettings[ACTIVE_SETTINGS];
-    CANON_UNITS programUnits;	// CANON_UNITS_INCHES,MM,CM
+    int programUnits;	// CANON_UNITS_INCHES, MM, CM
```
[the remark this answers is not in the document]
robEllenberg: Good catch, it should stay as an enum (likely a rebase issue)

robEllenberg:
@mozmck Agreed, multi-spindle unit tests would be a great exercise for anyone interested in trying out the unit test framework. I still need to document how to create a unit test, and use the helper utilities, since there's a bunch of under-the-hood stuff that's necessary to make them run smoothly. For now, an enterprising developer could copy one of the existing tests and tweak it to their purposes.

I'll have more documentation and some improvements (tests for emccanon, interp stack trace printing), but that will take a few calendar weeks to clean up and rebase.

@andypugh Are you interested in taking a stab at multi-spindle tests? if so, it would be great to get feedback on the framework, and I can patch up any gaps you run into.

andypugh:
I am a bit busy right now. 
How do these tests differ from the existing tests?

robEllenberg:
They're based in Catch, which is a power unit test framework. It's not a replacement for runtest, but the unit tests are much easier to write due to the variety of assert conditions and helper macros.

PR #623, opened 2019-07-11
Clean up our use of yapps

SebKuzminsky:
LinuxCNC's halcompile uses yapps in two ways:
* At build-time we use the yapps **executable** to **construct** the parser halcompile uses to ingest `.comp` files.  This constructed parser is included in the halcompile executable.
* At run-time the halcompile program **imports** the yapps runtime module to help it **run** the `.comp` parser.

Thus we need to build-depend on the yapps executable and runtime-depend on the yapps runtime (which is in a separate debian package).

In the olden days, yapps was not packaged for the distros we wanted LinuxCNC to run on, so we included a copy of yapps in the linuxcnc git repo.  This problem has since been remedied by Debian (and Ubuntu), and today all distros we support include yapps packages.  Unfortunately the yapps packaging is slightly different on Debian Buster than it was on earlier distros. 

This PR does X things:

1. Make LinuxCNC build-depend on the distro's packaged yapps executable, and use that to construct the halcompile parser.
2. Remove our copy of yapps, since it's no longer used.
3. Minor tweaks to halcompile so it can deal with both yapps 2.1 (in pre-Buster) and yapps 2.2 (in Buster and probably later).
4. Minor tweaks to our packaging scripts so they can runtime-depend on whatever the yapps runtime deb is called on the different distros.


jethornton:
Looks good to me.

PR #624, opened 2019-07-18
Add some build-time dependencies found with Raspbian

andypugh:
Just the things I needed to apt-get install to compile. Not sure of the configure file formatting though.


andypugh:
There is also a run-time dependency on python-glade2.

jepler:
I think this is covered by #672 -- let us know if not.

PR #633, opened 2019-08-11
Cleanup docs

simaoamorim:
Doing some cleanup on the docs folder seems in order, given that some files were used as announcements (like the docs/translator-needed_es.txt) others should be kept updated but were moved (like the changelog from docs/NEWS to debian/changelog) or just got forgotten (like docs/TODO).
Also relocated some help files from docs into docs/help and created an archive folder for the old UPDATING file, which can be stored there for reference if needed.

If there is any interest in doing some more deep cleanup just let me know.
Simão A.


simaoamorim:
The [include/](https://github.com/LinuxCNC/linuxcnc/tree/master/include) and [rtlib/](https://github.com/LinuxCNC/linuxcnc/tree/master/rtlib) folders are empty, only containing a .gitignore file. Are these folders still used for anything or can they be removed?

The [bin/](https://github.com/LinuxCNC/linuxcnc/tree/master/bin) folder contains a profiling script for axis which is used by [configs/sim/axis/profile_axis.ini](https://github.com/LinuxCNC/linuxcnc/blob/master/configs/sim/axis/profile_axis.ini) and I think it can be moved to configs/sim/axis/bin/profile_axis safely (somehow I think the bin/ directory on the same folder of the INI file that calls it gets searched for, but I'm not 100% sure about that). I'll test this when I get a chance.

Finally what about the [./TODO](https://github.com/LinuxCNC/linuxcnc/blob/master/TODO) file whose most recent blame line dates back to 2009? I'm guessing it is highly outdated?

Best regards,
Simão A.

andypugh:
> On 11 Aug 2019, at 23:29, Simão Amorim <notifications@github.com> wrote:
> 
> The include/ and rtlib/ folders are empty, only containing a .gitignore file. Are these folders still used for anything or can they be removed?

My impression has always been that these are populated during a rip build. But I have not looked in to it in detail.

simaoamorim:
It would make sense if true. I'll test that.

simaoamorim:
@andypugh you're wright, the bin/, include/ and rtlib/ folders got populated in a rip build.

What about removing them from the repository (add them to the .gitignore) and include the necessary mkdir's into the autogen script?

andypugh:
I have no strong opinions on the subject, but other than neatness, is it worth the trouble?

simaoamorim:
I know it’s only a matter of neatness, but I’ll give it a try because it’s a small change
and if it works the repo gets a bit cleaner ;)

andypugh:
Hmm, are you still interested in this? I committed your changes and broke the buildbot...
http://buildbot.linuxcnc.org/buildbot/builders/4016.deb-wheezy-i386/builds/3824
It looks like the deb-build process explicitly references files that you deleted. (that's a guess)

simaoamorim:
@andypugh indeed, I looked at the build log and the deb-build tried to ```install docs/UPDATING```, which I moved...
```bash
install -m 0644 -o root ../docs/UPDATING /tmp/buildd/linuxcnc-2.9.0~pre0.1278.g9162192/src/../debian/tmp/usr/share/doc/linuxcnc/UPDATING
install: cannot stat `../docs/UPDATING': No such file or directory
```
I can see if I can fix that in the deb-build configs...
Maybe installing the docs/ folder **AS IS** instead of hard-coding the files inside would solve the issue and, at the same time, give more flexibility when changing / refactoring documentation??
(What's the point in hard-coding documentation files in a deb-build, anyway?)

andypugh:
I have a fix in the pipeline, just checking it now.

andypugh:
Well, just removing the hard-coded mentions of the removed file isn't enough. 
And I have no idea what the problem is. 
The package build complains of files in debian/tmp that shouldn't be there, but I have no idea how these changes could cause that.

andypugh:
I have reverted all the changes. Just to stop the buildbot bombarding me with blamelist messages.

simaoamorim:
Yeah, don’t mess with the winning team...
I’ll take a look into it and try to fix it

It only happens with the deb-build right? Because the rip build runs fine, no complaints...

Simão Amorim

andypugh:
Looks like Jepler fixed it. 

Thanks Jepler.

PR #636, opened 2019-08-20
Update configure for raspbian 10

jethornton:
This updates configure for raspbian 10 and is tested on the Rpi4.


JonnyTech:
Wow!!!

PR #642, opened 2019-09-04
Added missing include required by recent versions of asm/uaccess.h

neilwhelchel:
This fixes a compile error with recent versions of the linux kernel, in my case 4.14.134:
`
./arch/x86/include/asm/uaccess.h:33:9: error: dereferencing pointer to incomplete type
  current->thread.addr_limit = fs;
`
I have not tested this against older versions of the kernel, but it seems safe to include traps.h as it has an include guard.

### DIFF
--- a/src/rtapi/rtai_rtapi.c
+++ b/src/rtapi/rtai_rtapi.c
@@ -64,6 +64,7 @@
 #include <linux/kernel.h>
 #include <linux/slab.h>		/* replaces malloc.h in recent kernels */
 #include <linux/ctype.h>	/* isdigit */
+#include <asm/traps.h>		/* current */
 #include <asm/uaccess.h>	/* copy_from_user() */
 #include <asm/msr.h>		/* rdtscll() */
 


andypugh:
I did an experiment to see if another possible fix worked last night. 
(I should have done it in an experimental branch, and really I think that a fix belongs in 2.8 or maybe even 2.7)
Based on https://lkml.org/lkml/2017/9/30/189 I put in 0dbbfb3ae106feea27dab4753ae40c21b6ea06d3

But I think I will roll that back and use kernel version macros. so that the change only happens post 2.14. Thoughts.

andypugh:
Thanks for bringing this to to our attention.
In the end I have decided to use the fix recommended here:
https://lkml.org/lkml/2017/9/30/189
via 5ffa41e504ab9138a1287560ded1749b2d5f7aa5
But please don't let this dissauade you from further pull requests.

neilwhelchel:
Hello Andy, good catch! Strangely enough I knew about how the headers are being changed around in the kernel, but I totally missed that when made this fix. I guess I had a bad case of tunnel vision.

PR #647, opened 2019-09-11
pncconf servo tandem tandem_check finalize

sphasse:
switch the last few cases of tandem logic to use the tandem_check function, which works for both stepgen and pwm configurations


c-morley:
Will need to add a tandem A2 signal to select before tandem A axis will work.

PR #649, opened 2019-09-28
Hm2 rpspi update for RPi4 and 64bit

BsAtHome:
The hostmot2 RPi SPI driver update to support the new Raspberry Pi 4 and will also run properly on 64-bit linux kernels. Additionally, the 7C80 and 7C81 Mesa cards (specific cards for the RPi) are now recognised and functional.
The changes do away with the detection of the RPi version and uses IO register hardware address resolution the same way as the libbcm_host by reading the appropriate /proc/device-tree entries from the kernel's architecture defined DTS configuration files. The changes are confirmed to function on RPi3 and RPi4 on the forum thread.

The change should probably be added to both master branch as well as the 2.8 branch.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/hm2_rpspi.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_rpspi.c
@@ -99,19 +99,12 @@ typedef struct hm2_rpspi_struct {
 	int		spiceid;	// The SPI CE id [012]
 } hm2_rpspi_t;
 
-typedef enum {
-	RPI_UNSUPPORTED,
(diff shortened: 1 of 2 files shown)


andypugh:
I think you are right that it should go in 2.8 (though mozmck should have the final say) 
But this pull request is against master...

BsAtHome:
Yes, this request is for master.
Should I create a separate pull request for 2.8 branch? Basically, just two files changed, so that should be trivial to create in a PR.

andypugh:
Merged to 2.9, so we will wait for a merge up from there rather than having a separate commit.

PR #651, opened 2019-10-02
Backport the RPi4 fixes and 7C80/7C81 board additions to the 2.8 branch

BsAtHome:
This is the same as PR #649, but specific for the 2.8 branch. There are no code differences, other than the branch it targets. Two separate PRs should make it easy to do the merge automatically for each target branch.


andypugh:
This will automatically end up in 2.9 next time there is an upstream merge.

PR #665, opened 2019-12-09
Mesa: take a pile of updates from Peter

jepler:
I've skimmed these and didn't spot any problems.  Feeding them to CI seems like the next step.

    
     * absenc: incompatible change: require version 4, support new register
     * eth: add support for 7i94, 7i95, 7i97
     * inmux: add support for inputs from multiplexor chips
     * inm: add support for MPGs
     * pwmgen: Add support for "offset mode"
     * xymod: Add support for galvanometer hardware interface


jepler:
@SebKuzminsky do you want to take a minute to look at this or should I just add it?

micges:
I think port names and numbering is wrong on 7i95 code in hm2_eth.c

andypugh:
I have had a look through the bits that I have touched in the past, and it looks OK in those areas. 
But we should wait for confirmation from PCW of the pin numbering that micges mentions.

jepler:
PCW confirmed pin numbering is OK on IRC.

PR #669, opened 2020-01-07
debian/: Update deps for Debian 10 & Raspbian 10

jepler:
I think the gir1.2-vte dependency is not strictly correct (it's gtk3 only) but this dependency will be part of the way forward for gscreen's terminal widget when it transitions to gtk3.

This yields a working package which can be installed, and AXIS runs. (on a pi4 with raspbian buster)


jepler:
Looks like 2.7 debian/configure is pretty different, higher chance of carnage. 2.8 cherry picks without conflict.

SebKuzminsky:
Can we close this now that we have #672?

PR #673, opened 2020-01-12
Pentakins

pkmcnc:
Add pentapod kinematics. Machine examples https://youtu.be/gMyvfx-qML0  https://youtu.be/vvqsVhICXkc


andypugh:
I can't see any way for a new kins to break existing code.

PR #679, opened 2020-02-08
axis: withdraw instead of iconify

phillc54:
some window managers don't deiconify after iconify but they will deiconify after withdraw

Signed-off-by: Phillip A Carter <phillcarter54@gmail.com>


andypugh:
Jepler: Do you have the time and inclination to opinionate on this?

jepler:
What window manager? Withdraw is correct, no icon should be shown for axis before its main window is completely initialized.

jepler:
https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#changing_window_state

> Once the window has been provided with suitable properties, the client is free to change its state as follows:
> 
>  *   Withdrawn -> Normal - The client should map the window with WM_HINTS.initial_state being NormalState.
>  *   Withdrawn -> Iconic - The client should map the window with WM_HINTS.initial_state being IconicState.
>  *   Normal -> Iconic - The client should send a ClientMessage event as described later in this section.
>  *   Normal -> Withdrawn - The client should unmap the window and follow it with a synthetic UnmapNotify event as described later in this section.
>  *   Iconic -> Normal - The client should map the window. The contents of WM_HINTS.initial_state are irrelevant in this case.
>  *   Iconic -> Withdrawn - The client should unmap the window and follow it with a synthetic UnmapNotify event as described later in this section. 

An application requesting Withdrawn -> Normal is a supported state transition since, oh, at least 1994.  We can take this change if necessary, but it is compensating for broken software.  I would really prefer to *name* the broken software and file a bug with that software, even if they close or ignore it.

phillc54:
Specifically it is muffin, the wm for Cinnamon.
There was a thread on a Tk list here <http://74.208.161.176/tk/tktview?name=ee3361a3ad>

jepler:
Thanks @phillc54

jepler:
I have to confess that I read the change backwards.  I think using withdraw is probably more correct, and assumed that correct code had been written in the first place, so double thanks @phillc54 ! and sorry for my failure in understanding such a small change. :(

PR #680, opened 2020-02-09
Improve German translation

be-we:


andypugh:
Thanks for this. However I think it probably belongs in the 2.8 branch, so that it will be in the next release. I tried changing branches, but then the pull request seemed to include hundreds of other files. 
Do you feel like re-submitting agains the 2.8 branch?

rene-dev:
The translations look a lot better than what we have.
I never noticed how bad they were, all my machines are set up in English.
I cherry picked the commit to 2.8 manually, so I'm closing this PR.
Thanks Benjamin!

PR #683, opened 2020-02-16
G7x

mark-v-d:
This implements the G70/G71 and G72 lathe cycles.


c-morley:
I see you added a requirement for a more recent version of C++ - does list limit the code to only newer distributions?

mark-v-d:
It requires gcc 5.2 or later, so distributions using an older compiler can't use it.

andypugh:
Wheezy has 4.7.2 
Stretch appears to come with 4.6.3 _and_ 6.3.0

What is it about your new content that requires 5.2 given that the rest of LinuxCNC appears to compile with 4.7.2?

(I have been away with christmas, work trips and a ski holiday, I was just about to get to checking this when the gcc query arose)

mark-v-d:
I used some constructs not present in C++11. Fortunately these are all
in the header files and I just copied the relevant parts to the source.
I think it will now work with gcc 4.7 as well, but it needs the -std=c++11
option I think because c++98 seems to be the default for that compiler.

I use gcc-9.2 myself, for which the default is c++14, so I never noticed.

andypugh:
OK, let's see how it goes. I think 2.9 is far enough away from release to merge then test.

andypugh:
The buildbot seems to be unhappy.....
(I can't recall which one this was)
emc/rs274ngc/interp_g7x.cc:25:32: warning: unknown option after ‘#pragma GCC diagnostic’ kind [-Wpragmas]
emc/rs274ngc/interp_g7x.cc:27:5: error: missing space between ‘""’ and suffix identifier
emc/rs274ngc/interp_g7x.cc:29:27: error: expected declaration before end of line
make: *** [objects/emc/rs274ngc/interp_g7x.o] Error 1

See also this, with a lot of issues...
http://buildbot.linuxcnc.org/buildbot/builders/1630.rip-stretch-rtpreempt-amd64/builds/1454/steps/compile/logs/stdio

andypugh:
More info here:
http://buildbot.linuxcnc.org/buildbot/waterfall
Some platforms compiled cleanly, it seems. Some not (and it does not seem to correlate exactly with age)

mark-v-d:
I've pushed two fixes. I guess some systems use a compiler which does not compile with c++11 as default. The other error I can't fully understand, but the fix is to change all occurences of 1i to 1.0i

andypugh:
Where are the fixes pushed to?

mark-v-d:
OK, I have a better fix. I've removed all literals, the only c++14 thing left is the std::make_unique.

mark-v-d:
I've pushed them to the original branch,
https://github.com/mark-v-d/linuxcnc-1/tree/g7x

andypugh:
I just found them. Do you want to make a new pull request?

mark-v-d:
If that's the simplest thing, I'll do that.

andypugh:
I am awaiting advice on how to proceed from wiser heads.

mark-v-d:
OK, I've made a pull request. It only shows the four fixes I've done.

I'm off to bed now, regards Mark.

jepler:
FWIW here are the versions of important packages on various systems we have supported in the past.  http://wiki.linuxcnc.org/cgi-bin/wiki.pl?MinimumSoftwareVersions

also FWIW I think it's worth opening the discussion on whether to drop some old distros in master branch.

mark-v-d:
I just had a look at the build logs, and as expected gcc must be version
4.7 or later. Don't be fooled by the versions using gcc-4.4, they don't
even start compiling. I use too many features from c++11 to backport it
to c++98, so this is as good as it gets.

PR #685, opened 2020-02-25
axis: allow pyvcp panel at bottom of windows

phillc54:
Allows a PyVCP panel to be at the bottom of the AXIS user interface as an alternative to being at the right side.


c-morley:
Any reason not to add this to 2.8?

phillc54:
Not that I can think of, it shouldn't break any existing configs.

andypugh:
So, do we merge this request? Or do you want it in 2.8?

c-morley:
I'll cherry-pick later for 2.8

PR #688, opened 2020-02-27
Make G7x code build on precise

jepler:
This makes it build on a freshly installed precise/uspace system.  No testing was performed.

### DIFF
--- a/src/emc/rs274ngc/interp_g7x.cc
+++ b/src/emc/rs274ngc/interp_g7x.cc
@@ -8,58 +8,6 @@
 #include <complex>
 
 
-////////////////////////////////////////////////////////////////////////////////
-// Ancient compiler work-arounds
-////////////////////////////////////////////////////////////////////////////////
-
-#if __cplusplus >= 201402L
-    using namespace std::string_literals;
-#else
-    namespace std {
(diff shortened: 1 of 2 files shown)


jepler:
@mark-v-d your comment would be appreciated.

mark-v-d:
The conversion from "dkldkldf"s -> "dkldkldf" is incorrect, this will
cause linuxcnc to crash. This should be std::string("dkldkldf") instead,
which I also did in the follow-up patch.

Changing the 1i to std::complex<double>{0,1} is ok, I used a constexpr
in the follow-up patch.

Moving the initialisation of some members to the constructor is OK.

Removing the override is OK.

Replacing std::make_unique is OK.

mark-v-d:
Replacing emplace with emplace_back is incorrect

If you can't use emplace, use insert. For example
        insert(n,std::unique_ptr<round_segment>(new round_segment(
                distance>0,(*p)->ep(),center,(*n)->sp())));

mark-v-d:
I've just pushed your changes to my archive. I think that updates the g71_test branch
and builds on the buildbot. The version I pushed works for me.

andypugh:
I think this one was superseded?

PR #686, opened 2020-02-26
G7x

mark-v-d:
Implemens th G70/G71 and G72 lathe cycles.

PR #689, opened 2020-02-27
G7x precise mk2

jepler:
This builds on the code in #686 to achieve compatibility with the compiler on precise.

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -40,7 +40,7 @@ AC_SUBST([RTAPI_APP_CFLAGS])
 
 m4_include([m4/ax_cxx_compile_stdcxx_11.m4])
 # make this optional - it will likely only result in warnings
-AX_CXX_COMPILE_STDCXX_11(noext,optional)
+AX_CXX_COMPILE_STDCXX_11(noext,mandatory)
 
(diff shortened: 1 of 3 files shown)


jepler:
@mark-v-d any concerns about the changes introduced by this patch?  
b809e50  in particular.

mark-v-d:
Looks OK, except for flip_state{}. Empty braces means default
initialisation, which is an indeterminate value for int. So that should
be flip_state{0}.

mark-v-d:
Same for finish{}, which must be finish{0}

mark-v-d:
Ignore my remarks, that is c++03

andypugh:
Code details aside, we now seem to have clean(ish) compiles on all platforms. 
However I am seeing problems with the tests, as is the buildbot. 

My test machine is running an experimental RTAI kernel, so might not be representative. But I am getting one failed runtest with 2.8 and 115 fails (113 pass) with master. 
I am investigating, but could you run the tests (./scripts/runtests ) on your dev system to see what you get? It takes about 15 minutes to run all tests.

andypugh:
OK, that's a good sign. I will have a look with a preempt-rt system and do some git bisecting to see what is going on.

PR #693, opened 2020-03-01
Axis preview failing with wrapped rotary axis.

curtdutt:
When a wrapped rotary axis absolute position is > 360 or < -360, axis preview will fail to load a file. This is because axis issues hidden "g53 g0 ..."  commands to the preview interpreter prior to sending the file contents to the preview interpreter. Axis uses the absolute coordinates as part of those commands which is not correct for a wrapped_rotary type axis.

Included is a sim file wrapped_rotary to demonstrate this issue as well as demonstrate the fix.

To Test:
  1. linuxcnc wrapped_rotary.ini
  2. home
  3. mdi "g0 a180"
  4. mdi "g0 a0"
  5. ctrl-r (reload axis.ngc)


andypugh:
I have been sat on this as I thought it was an issue report rather than a pull request. 
Sorry about that.

curtdutt:
No problem at all.

andypugh:
I have tested this. There is a definite bug and this is a fix. 
This is such a bug that the fix belongs in 2.8.

andypugh:
I have cherry-picked this in to 2.8, so I am closing this pull-request.

andypugh:
And thanks!

PR #698, opened 2020-03-17
added chinese translation for Getting Started(2.8)

ButterflyTech:
my mistake, I create a new pull request for 2.8


andypugh:
That looks better. I did start to edit the base of the other one, but then became unsure what GitHub was about to do.

andypugh:
I have noticed that these documents are not being picked up by the docs build system. I will fix that before the 2.8 release. 
Do you fancy also making a Chinese version of the "quick reference"? It already exists in Vietnamese and French. 
https://github.com/LinuxCNC/linuxcnc/tree/master/docs/html
(That particular page is written directly in HTML)

PR #700, opened 2020-03-18
Fix segmentation fault on single point path.

mark-v-d:
Fixes a segmentation fault. We cannot do anything unless there are at least two points in the path.


andypugh:
No code survives contact with the enemy :-)

PR #703, opened 2020-03-22
Fix the doc build again?

jepler:
I tested this locally by running "debian/configure uspace noauto && debuild -b" and fixed each issue as I identified it.  I didn't check into what was different vs when andypugh worked on fixing it.


andypugh:
Was it the moved files that was causing the "tmp not empty" issue? That part had me stumped, and it was getting very late.

andypugh:
Closed by b3fa2975fde671f10087cb85671e097f35c45d6b

PR #706, opened 2020-04-01
Fix tests in state tags

zultron:
@andypugh has test-merged the state tags work into master and reported test failures.  This should fix them.

Also silence a compiler warning.  (This ought to be filed as a separate PR to mainline branches.)


zultron:
- The `g64` test fix follows aa4d20b9
- The `motion-logger` test fix follows bf873146
- I'm not sure yet why the `m98m99` failures

zultron:
- The `m98m99` failures are explained by the 314669b4 commit log

zultron:
Updated commit log to document **why** the changes.

andypugh:
Thanks for this.

andypugh:
There are three places where the code refers to "NAX_ARITY" and I am fairly sure I have already seen a commit float past that fixed some of them. 
I would guess that the cleanest fix would be to cherry-pick this to master, then add another commit to fix the other two? (to make this branch easier to merge back to master)

phillc54:
I just gave it a try and all runtest passed except startup-gcode-abort

PR #708, opened 2020-04-04
build 2.8 on Buster in general and on RPi4 in particular

SebKuzminsky:
This branch makes two tweaks to the `platform-is-supported` script that the buildbot uses to decide what buildslaves should build this branch:

1. Cherry-pick the "let the buildbot build on Buster" commit from master.

1. If we can't find the kernel config, assume it's a vanilla (non-realtime) kernel rather than failing.  (This is needed for the RPi4 disk image I'm using in the buildbot).


andypugh:
The buildbot is a strange and mysterious place to me. The changes seem OK, with that caveat.

PR #711, opened 2020-04-07
XHC-WHB04B-6 Initial commit to Linuxcnc master

alkabal:
Hi

I have revised this driver from MK by Raoul Rubien

All seem to work great regading original function and i have enhanced some of them.

Button enhancement :
Single function button Macro10 are now used as a standard button for toggle Dro Absolue/Relative and can also be used as Macro14 with push and hold the Fn button like the other button with macro.

Single function button Continuous (Con mode) have now a macro function that toggle the Mist state regading the hal status.

Single function button Continuous (Step mode) have now a macro function that toggle the Mist state regading the hal status.

Button Reset/Stop/Play can also be used as Macro11/12/13 if needed.

Function enhancement :
At startup or after each time you set Lead = the Mpg/Step/Con mode is allways revert to Mpg mode.
Add 5mm and 10mm for step mode (try to make this a configurable option) 

Mpg mode = feedrate override using the wheel.
Con mode = "velocity" mode
Step mode = as usual
Lead position = Spindle override from wheel


andypugh:
Can you add a manual page? 
This: https://github.com/LinuxCNC/linuxcnc/blob/master/docs/man/man1/xhc-hb04.1
Generates this: http://linuxcnc.org/docs/2.8/html/man/man1/xhc-hb04.1.html

And maybe squash it all in to one commit?

rene-dev:
There is a german guy that wrote a similar driver: http://www.talla83.de/xhcwhb04b.html
would it be possible to have one driver supporting the XHC-WHB04B and xhc-hb04?

alkabal:
Yes I have started the revised driver using this one + last MK version.

I have removed some useless code, add enhancement and correcting some wrongly manages stuff.

For 4 axis version the problem is I does not have this device, but yes I like the idea to do all in one.
Thanks for interest


Ps : yes i have started to write the help file

rene-dev:
I have the 4 axis version, and can test it.

alkabal:
WHB04B-4 or HB04 ?

rene-dev:
both

alkabal:
Good

do you know a working driver for WHB04B-4 ?

alkabal:
> {Anything} = false always returns false, as it is an assignment.
> You need (mIsStepMode_5_10 == false)

Unfortunately does not work i think the problem is from something like class

alkabal:
Not possible to use the original one ?
https://github.com/machinekit/machinekit-docs/blob/master/docs/drivers/xhc-whb04b-6.asciidoc

andypugh:
I hadn't found the asciidoc version. Yes, start with that one. 

And that " = false" definitely needed fixing, even if that was not the problem you were seeing.

alkabal:
fix is done thanks !

alkabal:
@rene-dev

> I have the 4 axis version, and can test it.

Please can you try to start this driver ? (or compile if this is easy for you)
https://forum.linuxcnc.org/38-general-linuxcnc-questions/38774-pendant-whb04b-on-linuxcnc-2-7-debian?start=10#163095
https://forum.linuxcnc.org/38-general-linuxcnc-questions/38774-pendant-whb04b-on-linuxcnc-2-7-debian?start=10#163149

Br

PR #714, opened 2020-04-08
XHC-WHB04B-6 Squashed commit

alkabal:
Hi

I have revised this driver from MK by Raoul Rubien

All seem to work great regading original function and i have enhanced some of them.

Button enhancement :
Single function button Macro10 are now used as a standard button for toggle Dro Absolue/Relative and can also be used as Macro14 with push and hold the Fn button like the other button with macro.

Single function button Continuous (Con mode) have now a macro function that toggle the Mist state regading the hal status.

Single function button Continuous (Step mode) have now a macro function that toggle the Mist state regading the hal status.

Button Reset/Stop/Play can also be used as Macro11/12/13 if needed.

Function enhancement :
At startup or after each time you set Lead = the Mpg/Step/Con mode is allways revert to Mpg mode.
Add 5mm and 10mm for step mode (try to make this a configurable option)

Mpg mode = feedrate override using the wheel.
Con mode = "velocity" mode
Step mode = as usual
Lead position = Spindle override from wheel


alkabal:
Old Pr request : https://github.com/LinuxCNC/linuxcnc/pull/711

rene-dev:
I tested it with the 3 axis version, and it works!
why did you map y to c?
https://github.com/LinuxCNC/linuxcnc/pull/714/files#diff-849d9fb6ee682e492b7fddb62b896115R52
this was the only thing I had to change.

andypugh:
So, 2.8 or Master? As a standalone driver I am thinking 2.8.

rene-dev:
I see no problem with 2.8.
I tested it, and all works. example config would be good.

alkabal:
i appologize i use this particular setting for my lathe tailstock

I'm very happy if this work !!

alkabal:
I have updated exemple file

For now the B args does not work combined to mIsStepMode_5_10  I'm not able to found the solution myself,i'm really sorry

alkabal:
i thinks with minor change it can work with 2.7 too

andypugh:
2.7 is pretty much closed. I am determined that the next release will be 2.8. And the only folk who can't upgrade are those running Ubuntu Lucid from 2010.

andypugh:
Rene: When you think this is ready, can you push it to 2.8?

alkabal:
Other things can be added, i have removed the spindle speed input because these was badly implemented in place of feedrate override value, 
I think now is possible to add these signal again and link to display only using Fn macro-3 and macro-4 (direct spindle increase decrease)

what did you think ?

alkabal:
Hi

I can try another attemp for squash all in one file, but the most easy for me is to remove fork and restart with fresh fork.

rene-dev:
no, its ok like this. tomorrow I will test again, and then merge it. do you think its ready?

alkabal:
Hi

For me this is ready except the B option from loadrt : using the var mIsStepMode_5_10 does not work
unfortunatly i don't found the way for make it work, i suspect something like class sharing variable

And maybee need to check if the man page work with web server.

Thanks

rene-dev:
Hi, I manually merged into 2.8. please make any future pull requests updating this driver into 2.8 branch.
Thanks!

andypugh:
On Fri, 17 Apr 2020 at 19:54, Rene Hopf <notifications@github.com> wrote:

> Hi, I manually merged into 2.8. please make any future pull requests
> updating this driver into 2.8 branch.
>

Can you un-merge? That has brought a lot of stuff from master in, 177 files
changed...

-- 
atp

andypugh:
@alkabal Do you feel enthusiastic about making this compile with older versions of Linux? 
We currently support the following in 2.8, and your component does not compile on some of them:
I know that we are talking in terms of supporting OS distributions 3 years past end of support, but machinists tend to not want to interfere with working systems.

alkabal:
Hi

I'm not closed, but i'm not sure knowing what is needed to do.

need to do all debian 7 to 9 ? + Ubuntu 12.04 ?

this is only compiling problem ?
I think this is not possible with RPI4 so need to use other computer ?

andypugh:
If an RPI is all you have, then this is a but much to ask. 
Do you have any other PC? I do a lot of testing in VMs in my Mac (using VMWare Fusion)

alkabal:
Yes i have other computer (laptop only) do not worries 

Thanks for Vmware suggestion !

alkabal:
Debian strech already compile with success ?

(Some time ago i have do the compilation with debian stretch but with machinekit)

andypugh:
The interesting one is probably Wheezy. If you can set up a VM with the LinuxCNC LiveCD as a test case that would be useful. 
http://www.linuxcnc.org/iso/linuxcnc-2.7.14-wheezy.iso

alkabal:
With yout link

at first step apt-get update/upgrade : 
Capture

alkabal:
so i can't install dependancy for compiling master branch

alkabal:
ubuntu 12.04 do a lot of error

Capture

phillc54:
You could try tricking the wheezy archive by temporarily changing the date on your machine then try the update.

sudo date -s "1 JAN 2018"
sudo apt-get update

andypugh:
You can ignore the fact that there are no security updates on Wheezy, I think. 
But Ubuntu 12.04 seems to be showing the sorts of problems we are talking about. 
Is line 79 of pendant-types.h valid C++? And is it valid in C++98?
(Also, this might not matter but C++ header files are normally .hh)

12.04 is running gcc 4.6, which I think defaults to C++98 but can be set to (slightly) newer versions with a command-line parameter ( -std=c++98, -std=c++03 should work, -std=c++11 and -std=c++14 probably not). One of these could probably be added to the Submakefile. 

I am not any sort of expert on this, most of this is what I have learned through Google in the last few minutes.

phillc54:
I did manage to build on Wheezy but had to comment out the logging as that is not available in the Wheezy lib_usb.h

alkabal:
Hi

Thanks a lot for your help.

I’m really sorry but for one or two days I can’t work on this.

I apologize for compilation problem I have change nothing from the first authors regarding to this and I’m really un-experimented for setting dev environment.

Br

andypugh:
Yes, I just finished making it build on Wheezy too. 
I needed to comment out the logging, as you say. And also add a -std=c++11 to the Submakefile

But that seemed to be all that was needed.

andypugh:
More on this. With Wheezy the ./configure script automatically adds -std=c++11 to the Makefile.inc
So for Wheezy only the debug logging change needs to be made. 
I will try to find a Precise VM

andypugh:
I managed to confuse myself bouncing around between various VMs and SSH windows to real machines.
Buster configure shows: checking whether g++ supports C++11 features by default... yes
Whereas Wheezy configure has:  checking whether g++ supports C++11 features by default... yes

alkabal:
Hi

No more need to check old distro ? you have do all needed change ?

andypugh:
Yes, all sorted out now. 
I ended up simply not trying to compile if for Precise. Precise was always a "build it yourself" option, never an ISO. I think that the overlap between users running Precise and users needing to run the XHC-WB06 is probably very small. 
But it compiles and works on anything Wheezy or later.

alkabal:
Good job, many thanks

c-morley:
Does this driver require a new library? 2.8 will not compile on mint 19.
lots of error similar to:

objects/hal/user_comps/xhc-whb04b-6/usb.o: In function `XhcWhb04b6::Usb::Usb(char const*, XhcWhb04b6::OnUsbInputPackageListener&, XhcWhb04b6::Hal&)':
/home/chris/emc/src/hal/user_comps/xhc-whb04b-6/usb.cc:190: undefined reference to `libusb_alloc_transfer'
/home/chris/emc/src/hal/user_comps/xhc-whb04b-6/usb.cc:191: undefined reference to `libusb_alloc_transfer'

andypugh:
It needs libusb1.0 (and presumably -dev), but it shouldn't even try to compile it without that.

andypugh:
The build system is making a terrible job of the HTML docs: 
http://linuxcnc.org/docs/2.8/html/man/man1/xhc-whb04b-6.1.html
The manpage is mainly usable.

phillc54:
Maybe it needs to be in Groff format. If that is the case I would be happy to have a go at it.

andypugh:
The system is set up to convert asciidoc to groff for manpage, and in this case that is working well. It is PDF and HTML that are not working. 
(http://linuxcnc.org/docs/2.8/pdf/LinuxCNC_Manual_Pages.pdf page 92 for the PDF rendering).
I think we need to split or remove the tables.

phillc54:
That is strange as it does do tables well for PDF and HTML from other asciidoc sources.

andypugh:
I think the tables are too long, as hinted at in the build failure messages:
http://buildbot.linuxcnc.org/buildbot/builders/1635.rip-buster-rtpreempt-i386/builds/111/steps/compile/logs/warnings%20%2855%29

andypugh:
One tiny fix I will be pushing, configure --help should tell you how to configure building both types of docs when setting  --enable-build-documentation  The answer is:
./configure --enable-build-documentation=pdf,html
(or yes, or Yes, or html,pdf, or Y or y)

phillc54:
If there is something I can do to help let me know and I can do it tomorrow.
I see a couple of PlasmaC warnings there too, I will get rid of them as well.

andypugh:
It's done. I removed the tables and now the docs contain the info, albeit in a less elegant format

alkabal:
I apologize for inconvenience !

PR #716, opened 2020-04-09
Update kernel version PREEMPT_RT check

cascade256:
When trying out the in development 5.4-rt kernel, I found that the version string changed and LinuxCNC was unable to detect the patched kernel. This PR adds a check for the new version string and lets LinuxCNC detect the patched kernel.


andypugh:
Thanks.

PR #729, opened 2020-04-15
Switch to use mktemp(1) instead of tempfile(1)

dwrobel:
tempfile(1) is not available at all on Fedora while mktemp(1) is
part of the coreutils package available virtually on all Linux
distributions.


rene-dev:
lets see what the buildbot says

PR #732, opened 2020-04-16
hostmot2: add rcpwmgen support

jepler:
Testing performed: only build-tested on buster/uspace


jepler:
This is a code drop from peter, I didn't write it.

rene-dev:
looks good, is this for a specific card?

andypugh:
I think it needs a hint of documentation in docs/src/man9/hostmot2.9

(Or, better, the info in that file should become an asciidoc version) 

Otherwise the only hint that an rcpwmgen is for controlling rc servos and brushless ESCs is in the code.

PR #735, opened 2020-04-17
Fix doc build

jepler:
An error building the documentation was introduced at 74fc4491014b9de4141c44f58e391960e681343b.  Fix it, fix a spurious "rm" error after a real error, and have travis build the docs so we get these errors sooner.


jepler:
If you want to substitute your own version, please do and just close this one up :+1: .  But you can also use it as a key for where all the problems are :)

PR #737, opened 2020-04-17
GitHub actions

jepler:
Travis-ci results aren't showing up in pull request info for some reason, for months.

This adds github actions, which have their own squirrelyness but .. let's give it a try anyway.


jepler:
Here's the successful github actions run with this branch: https://github.com/jepler/linuxcnc/actions/runs/80876697

PR #740, opened 2020-04-17
halcompile: allow "?:" ternary operations in array size calculations

jepler:
This old commit was found in one of my source trees.  Apparently it would enable the size of an array of pins to have a "?:" ternary operator in its size, probably something like `arr[4 : personality & 1 ? 4 : 3]`, making an array be size 3 or 4 depending on specific needs.

I don't know anymore why I thought it would be useful.


andypugh:
As a relatively heavy user of halcompile I have never felt the need for this. But at the same time, someone might have use for it, and what harm can it do?

PR #742, opened 2020-04-18
Reduce use of sprintf, strcpy and strcat

rene-dev:
jeplers https://github.com/LinuxCNC/linuxcnc/pull/495 but for 2.8


andypugh:
I think that the compiler actually checks if you do anything with the return value of snprintf. 
Otherwise it gives the "possible truncation" warning.

rene-dev:
Its actually right about the warning, but at least with this patch it doesn’t write to unallocated memory. Im trying to find a sane way of silencing it.

andypugh:
Just to avoid duplication of work, I am running through all the truncation warnings. 
So let me know if you are further along than me :-)

rene-dev:
I also ran https://clang-analyzer.llvm.org/ to find more potential problems.
it generates a html report file. some of them I already fixed.

PR #743, opened 2020-04-18
unsupport old distros in master

jepler:
If -- any only if -- there is consensus, we can include this patch to stop building older distros on buildbot.

### DIFF
--- a/scripts/platform-is-supported
+++ b/scripts/platform-is-supported
@@ -100,7 +100,12 @@ if cpu not in supported_cpus:
     sys.exit(1)
 
 if distributor == 'Ubuntu':
-    if release_major < 12:
+    if release_major < 18:
+        print "release is too old!"
+        sys.exit(1)
+
+if distributor == 'Debian':
+    if release_major < 10:
         print "release is too old!"
         sys.exit(1)
 


andypugh:
Requiring Debian 10 might be a step too far. But I can see an argument for dropping anything out of LTS _at_the_point_of_release

SebKuzminsky:
Maybe this is known to all involved, but in case not, note that each branch has its own copy of this file, which controls what distros that branch builds on.  This PR is for master, so 2.7 and 2.8 would be unaffected.

Current builds:

2.7: 
- Ubuntu >= 10
- Debian (unspecified, so >= 7 Wheezy)

2.8:
- Ubuntu >= 12
- Debian (unspecified, so >= 7 Wheezy)

master:
- Ubuntu >= 12
- Debian (unspecified, so >= 7 Wheezy)

c-morley:
If it possible to have the makefile only make the driver if the distribution is current enough?
I seems I bit crazy to have to a user update a distribution and possible hardware so that a driver that might not be used can compiled.

andypugh:
Conditional compilation is an interesting idea for the new XHC driver. It has a submakefile and appears to only compile if it finds a USB library.
Perhaps a check for gcc-6 ? (or whatever it takes)

gmoccapy:
Master should only build on recent OS, Imho Mint 20, etc. and Debian 10

We should take also the python 3 changes to master asap even if that brakes a lot, isn't master for THAT?

rene-dev:
Agree. Anything older doesn't ship with a glade version that supports python3, and glade is an important part of customizing linuxcnc.

andypugh:
Can I suggest a new line at the  bottom of:
to illustrate this proposal? 
It would also be good to expand it to the right for new and upcoming OSes.
I have thought about this, and think we should leave 2.6 and Lucid on there. It is a useful resource for telling folk with very old ISO installs who are resistant to an OS upgrade what LinuxCNC version they can run.

rene-dev:
I already started, but got confused by the wiki syntax. Will continue today.

andypugh:
Also bear in mind that we can potentially backport Glade from our own server. As we do with python-gtksourceview2.
And, for that matter, glade-gtk2 in Wheezy: http://linuxcnc.org/dists/wheezy/base/binary-amd64/

So, as long as it is only a small package list we can specify things as dependencies and also provide them.

rene-dev:
I updated the wiki page.
glade is one of the problems, there might be more. we need to test which versions work and which doesnt, see https://github.com/LinuxCNC/linuxcnc/issues/820
additional to python3 there are now the gtk3 related dependencies to be checked.

andypugh:
The Wiki table says that Stretch is supported, but the PR requires Buster.
Which way do we intend to go?

jepler:
Debian 9 ("stretch") is currently "oldstable" and receives limited updates.  For a version of LinuxCNC targeted for release in 2021 or 2022 I do not know why supporting stretch would be a priority.

> Debian 9 has been superseded by Debian 10 ("buster"). Security updates have been discontinued as of July 6th, 2020.
>
> However, stretch benefits from Long Term Support (LTS) until the end of June 2022. The LTS is limited to i386, amd64, armel, armhf and arm64. All other architectures are no longer supported in stretch. For more information, please refer to the LTS section of the Debian Wiki.

jepler:
It appears that some sort of "protection" against attacks used by the host of the wiki software may prevent edits to the MinimumSupportedVersions page of the wiki.  I don't think I can do anything about it, unless we wished to move the hosting of the wiki to somewhere that we have more control. (wiki is on dreamhost shared hosting with the limitations that entails)

jepler:
Unless someone has come forward to offer to support Debian before 10 and Ubuntu before 18, I'd like to get this PR merged.

andypugh:
I guess the 2.9 release manager is the final arbiter, perhaps we should elect one. 

Actually, it might be time for an actual almost-formal meeting. There are decisions to make about Py3 and Gtk and making changes that will break existing glade panels (because, for one thing, apparently H-box is no-more)

andypugh:
After a bit of discussion elsewhere, I think the time has come.

PR #750, opened 2020-04-20
Squash many compiler warnings.

andypugh:
Mainly truncation warnings, but a couple of others too.

### DIFF
--- a/src/emc/pythonplugin/python_plugin.cc
+++ b/src/emc/pythonplugin/python_plugin.cc
@@ -36,8 +36,12 @@ namespace bp = boost::python;
 #define ERRMSG(fmt, args...)					\
     do {							\
         char msgbuf[MAX_ERRMSG_SIZE];				\
-        snprintf(msgbuf, sizeof(msgbuf) -1,  fmt, ##args);	\
-        error_msg = std::string(msgbuf);			\
+        size_t ret = snprintf(msgbuf, sizeof(msgbuf) -1,  fmt, ##args); \
+        if (ret > sizeof(msgbuf)){                                      \
+            snprintf(msgbuf, sizeof(msgbuf), "Error message too long"); \
+        } else {                                                        \
+            error_msg = std::string(msgbuf);                            \
+        }                                                               \
     } while(0)
 
 #define logPP(level, fmt, ...)						\
--- a/src/emc/rs274ngc/interp_internal.cc
+++ b/src/emc/rs274ngc/interp_internal.cc
@@ -476,14 +476,14 @@ int Interp::call_level(void) { return _setup.call_level; }
 
 std::string toString(GCodes g)
 {
-    char buf[10]={};
+    char buf[15]={};
     int dec_value = g%10;
     if (dec_value)
     {
         // Has a decimal
-        snprintf(buf, 10, "G%d.%d", g/10, dec_value);
+        snprintf(buf, 15, "G%d.%d", g/10, dec_value);
     } else {
-        snprintf(buf, 10, "G%d", g/10);
+        snprintf(buf, 15, "G%d", g/10);
     }
     return buf;
 }
--- a/src/emc/usr_intf/sockets.c
+++ b/src/emc/usr_intf/sockets.c
@@ -306,7 +306,7 @@ int sockPrintfError(int fd, const char *format, .../*args*/ )
   va_list ap;
   int size = 0;
 
-  strncpy(buf, huh, sizeof(huh)); // note: sizeof(huh) < MAXMSG
+  strncpy(buf, huh, sizeof(buf));
 
   va_start(ap, format);
   size = vsnprintf(buf + (sizeof(huh)-1), sizeof(buf) - (sizeof(huh)-1), format, ap);
--- a/src/hal/components/sampler.c
+++ b/src/hal/components/sampler.c
@@ -115,7 +115,7 @@ int rtapi_app_main(void)
     samplers = hal_malloc(MAX_SAMPLERS * sizeof(sampler_t));
     /* validate config info */
     for ( n = 0 ; n < MAX_SAMPLERS ; n++ ) {
-	if (( cfg[n] == NULL ) || ( *cfg == '\0' ) || ( depth[n] <= 0 )) {
+	if (( cfg[n] == NULL ) || ( *cfg[n] == '\0' ) || ( depth[n] <= 0 )) {
 	    break;
 	}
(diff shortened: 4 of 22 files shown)


src/emc/usr_intf/sockets.c
```diff
@@ -306,7 +306,7 @@ int sockPrintfError(int fd, const char *format, .../*args*/ )
   va_list ap;
   int size = 0;
 
-  strncpy(buf, huh, sizeof(huh)); // note: sizeof(huh) < MAXMSG
+  strncpy(buf, huh, sizeof(buf));
```
rene-dev, line 309: I looked at this for a while as well, and I think the old code is correct, but its a not obvious what it does.
Its supposed to copy huh to buf, so len of huh is correct.

src/hal/components/sampler.c
```diff
@@ -115,7 +115,7 @@ int rtapi_app_main(void)
     samplers = hal_malloc(MAX_SAMPLERS * sizeof(sampler_t));
     /* validate config info */
     for ( n = 0 ; n < MAX_SAMPLERS ; n++ ) {
-	if (( cfg[n] == NULL ) || ( *cfg == '\0' ) || ( depth[n] <= 0 )) {
+	if (( cfg[n] == NULL ) || ( *cfg[n] == '\0' ) || ( depth[n] <= 0 )) {
```
rene-dev, line 118: now they both do the same thing.

src/hal/utils/halrmt.c
```diff
@@ -2852,7 +2858,7 @@ static cmdResponseType setLoadRt(char *s, connectionRecType *context)
   char *args[MAX_TOK+3];
   int i;
   
-  if (s == '\0') return rtCustomHandledError;
+  if (*s == '\0') return rtCustomHandledError;
```
rene-dev, line 2861: same here, depending on how its called maybe also check for null pointer?

src/emc/usr_intf/sockets.c
andypugh: strncpy copies at most size bytes, and stops at the /0. So The effect should be exactly the same, except in cases where (for some reason) huh became longer than buf.

src/hal/components/sampler.c
andypugh: I am not sure they do, cfg is an array of pointers, to we are checking if each element is a null pointer, and if it points to an empty string.

jepler: Maybe you need `*(cfg[n]) == '\0'`, with added parens so there's no question of operator precedence between * and []?

src/emc/usr_intf/sockets.c
jepler: in strncpy/strlcpy the size should always be "about" the destination buffer, even if the source & destination are demonstrably the same size or the source is demonstrably shorter.

src/emc/pythonplugin/python_plugin.cc
```diff
@@ -36,8 +36,12 @@ namespace bp = boost::python;
 #define ERRMSG(fmt, args...)					\
     do {							\
         char msgbuf[MAX_ERRMSG_SIZE];				\
-        snprintf(msgbuf, sizeof(msgbuf) -1,  fmt, ##args);	\
-        error_msg = std::string(msgbuf);			\
+        size_t ret = snprintf(msgbuf, sizeof(msgbuf) -1,  fmt, ##args); \
+        if (ret > sizeof(msgbuf)){                                      \
```
jepler, line 40: Should be `ret >= sizeof(msgbuf)`  The returned value doesn't count the '\0', so if you have buf[4] and write 'abcd' to it, snprintf will truncate to 'abc\0' and return 4.

src/emc/rs274ngc/interp_internal.cc
```diff
@@ -476,14 +476,14 @@ int Interp::call_level(void) { return _setup.call_level; }
 
 std::string toString(GCodes g)
 {
-    char buf[10]={};
+    char buf[15]={};
     int dec_value = g%10;
     if (dec_value)
     {
         // Has a decimal
-        snprintf(buf, 10, "G%d.%d", g/10, dec_value);
```
jepler, line 484: instead of a constant, better to write sizeof(buf).

Since this is C++ code returning a std::string, you could consider using boost::format here (though we don't use it yet).  `return (boost::format("G%d.%d") % (g/10) % (dec_value)).str()` or so.

src/hal/utils/halcmd.c
```diff
@@ -753,11 +753,12 @@ static int replace_vars(char *source_str, char *dest_str, int max_chars, char **
...
-                    *detail = info;
-		    return -4;
-                }
+            {
+                size_t ret = snprintf(info, sizeof(info), "%s", var);
+                if (ret > sizeof(info)) return -7;
+                *detail = info;
+            return -4;
+            }
 		if (strlimcpy(&dp, replacement, strlen(replacement), &buf_space) <0)
```
jepler, line 762: Your editor is reformatting code, changing leading whitespace. It makes it hard to interpret the intended change.

src/hal/utils/halrmt.c
```diff
@@ -1855,21 +1855,27 @@ static void getThreadInfo(char *pattern, connectionRecType *context)
...
         unsigned int runtime_pin_value;
 
-        snprintf(name, sizeof(name), "%s.time",tptr->name);
-        pin = halpr_find_pin_by_name(name);
-        if (pin) {
-            if (pin->signal != 0) {
-                sig = SHMPTR(pin->signal);
-                dptr = SHMPTR(sig->data_ptr);
+        size_t ret = snprintf(name, sizeof(name), "%s.time",tptr->name);
+        if (ret > sizeof(name)) {
```
jepler, line 1859: ret >= sizeof(name)

src/emc/pythonplugin/python_plugin.cc
andypugh: I was wondering about that, but decided to follow examples I found on the internet

src/hal/utils/halcmd.c
andypugh: Looking at the LinuxCNC coding style docs, I see the problem. I have Geany set up to indent with 4 spaces, but as far as I can tell that means that tabs get displayed with a width of 4. 
I will look in to this in more detail, but might just have to accept that files with mixed tabs and spaces will always be hard to follow on my screen.

andypugh:
I have made the changes and will push directly.

PR #751, opened 2020-04-21
Fix honoring {C,CXX,CPP,LD}FLAGS passed to the 'configure'

dwrobel:
This fixes linking against libreadline v5 in a non standard directory.
    
    On Fedora multiple readline libraries can be installed in parallel.
    Old GPL-2 compatible libreadline v5 has its own header and library in the
    following directories:
    
     $ rpm -ql compat-readline5-devel | grep -e 'readline\.h' -e 'libreadline\.so'
     /usr/include/readline5/readline/readline.h
     /usr/lib64/readline5/libreadline.so
    
    As per 'configure --help' the following flags (excerpt) should be supported:
     Some influential environment variables:
      LDFLAGS     linker flags, e.g. -L<lib dir> if you have libraries in a
                  nonstandard directory <lib dir>
      CPPFLAGS    (Objective) C/C++ preprocessor flags, e.g. -I<include dir> if
                  you have headers in a nonstandard directory <include dir>
    
    Despite passing above directories via 'CPPFLAGS' and 'LDFLAGS' to 'configure' the linking
    process fails as follows:
    
     gcc -Wl,-rpath-link,../lib -ltirpc  -o ../bin/halcmd objects/hal/utils/halcmd.o
       objects/hal/utils/halcmd_commands.o objects/hal/utils/halcmd_main.o
       objects/hal/utils/halcmd_completion.o ../lib/liblinuxcncini.so.0
       ../lib/liblinuxcnchal.so.0 -lreadline
     /usr/bin/ld: cannot find -lreadline
    
    This was due to not honoring flags passed to configure.
    
    This also fixes the error of passing C specific compielr option to C++ compiler:
       cc1plus: warning: '-Werror=' argument '-Werror=implicit-function-declaration'
       is not valid for C++
    

jepler:
I reverted it.  The problem seems to affect any RTAI kernel-mode configuration.  @dwrobel please feel free to resubmit if you can verify it works on one of the affected platforms.  Unfortunately, the github CI does not presently test these builds.

dwrobel:
I'm sorry for the trouble I'll try to check also the RTAI configuration.

> Unfortunately, the github CI does not presently test these builds.

Would be awesome to have it included in the `CI` one day.

dwrobel:
>please feel free to resubmit if you can verify it works on one of the affected platforms.

@jepler I submitted https://github.com/LinuxCNC/linuxcnc/pull/850 which works for me fine with `RTAI` as well.

PR #756, opened 2020-04-22
Fix threads.0 test

dwrobel:
Fixes errors when the unversioned python
interpreter defaults to python3.

Fixes the following errors:
  File "/home/sw/projects/machinekit/linuxcnc/tests/threads.0/checkresult", line 6
    print "result contained %d lines, not the expected 3500 lines!" % (len(l))
                                                                  ^
SyntaxError: invalid syntax
*** /home/sw/projects/machinekit/linuxcnc/tests/threads.0: FAIL: checkresult exited with 1

  File "tests/threads.0/checkresult", line 7
    raise SystemExit, 1 # failure
                    ^
SyntaxError: invalid syntax

### DIFF
--- a/tests/threads.0/checkresult
+++ b/tests/threads.0/checkresult
@@ -3,8 +3,8 @@ import sys
 
 l = [int(line.strip()) for line in open(sys.argv[1])]
 if len(l) != 3500:
-    print "result contained %d lines, not the expected 3500 lines!" % (len(l))
-    raise SystemExit, 1 # failure
+    print("result contained %d lines, not the expected 3500 lines!" % (len(l)))
+    raise SystemExit(1) # failure
 
 lineno = 1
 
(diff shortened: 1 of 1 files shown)


andypugh:
Whilst the Python changes look good, I am curious what this fixes. As far as I can tell the tests run and pass.

dwrobel:
>I am curious what this fixes. 

The error, when running the test in my environment, is in the git commit message. Basically some of the tests doesn't work at all in the environment where the unversioned `/usr/bin/python` defaults to `python3` like we have e.g. in `Fedora`:
```
$ /usr/bin/python --version
Python 3.7.6
```
In other words using `/usr/bin/python` you can't assume what version of python it will be.

So, I'm fixing tests to be compatible, at the source code level syntax, with both `python2` and `python3`.

Additionally, if the python script doesn't not include any `linuxcnc` specific modules _(e.g. `hal`)_ I'm leaving the existing interpreter _(usually it's unversioned `/usr/bin/python`)_. As the code should work on both `python2` and `python3` interpreters.

If the python script includes any `linuxcnc` specific modules then I'm changing it to the `/usr/bin/python2` as otherwise they are not portable on the system where the `/usr/bin/python` defaults to `python3`. On such a system you would see errors which I'm usually pasting on the `git` message.

I hope this helps, but please let me know if you have any other questions.

dwrobel:
PRs from #756 to #802 fixes all python scripts to work in an environment where the `/usr/bin/python` defaults to `python3`. Additionally all source codes were modified to be compatible with both `python2` and `python3`. Each '_Python script_' file has also been compiled to verify the syntax using:

>$ python2 --version
Python 2.7.17

>$ python3 --version
Python 3.7.6

> $ cd linuxcnc/tests
$ find . -type f -exec file {} \; | grep 'Python script' | awk '{sub(/:/, "", $1); print "python2 -m py_compile " $1}' | bash -e
$ find . -type f -exec file {} \; | grep 'Python script' | awk '{sub(/:/, "", $1); print "python3 -m py_compile " $1}' | bash -e

I would appreciate if those PRs could be reviewed. In case you find them useful it might be worth to add such a verification _(somewhere in the makefile or at least CI flow)_ to make sure that there will be no further regression in that area.

dwrobel:
>PRs from #756 to #802 fixes all python scripts

I missed to push this one https://github.com/LinuxCNC/linuxcnc/pull/806.

PR #758, opened 2020-04-22
Fix tests/build/header-sanity

dwrobel:
Fixes the following error in case where the libtirpc is used
as an RPC implementation provider:

+ g++ -std=c++0x -DULAPI -I/home/sw/projects/machinekit/linuxcnc-rip/include -E -x c++ /home/sw/projects/machinekit/linuxcnc-rip/include/cms_xup.hh
/home/sw/projects/machinekit/linuxcnc-rip/include/cms_xup.hh:19:10: fatal error: rpc/rpc.h: No such file or directory
   19 | #include <rpc/rpc.h>  /* struct XDR */
      |          ^~~~~~~~~~~


rene-dev:
the shebangs of the tests need to be changed to linuxcnc-python, a new script that uses the python version set at configure

dwrobel:
>the shebangs of the tests need to be changed to linuxcnc-python

I'll take care of it.

PR #776, opened 2020-04-24
Silence "psabi" warnings on raspbian

jepler:
Raspbian warns by default about changes in compiler ABI that would prevent mixing code built by different compiler versions.  These warnings are not important to us, and are generated simply by including/using C++ features.  Silence them.


jepler:
Correction: these are technically "notes", not warnings.

PR #807, opened 2020-04-27
Add pycheck (Python check) Makefile target

dwrobel:
It is designed to be used for checking compatibility of python
    scripts (currently only from the 'tests' subdirectory) with
    both python2 and python3 interpreters.
    
    Usage:
      $ make pycheck
    
    Successfull output:
      $ make pycheck
      Reading 200/200 dependency files
      Done reading dependencies
      Reading 212/212 realtime dependency files
      Done reading realtime dependencies
      Checking *.py files for python2 and python3 compatibility...
      Checking "Python scripts" (by MIME type) files for python2 and python3 compatibility...    

### DIFF
--- a/scripts/travis-build-test.sh
+++ b/scripts/travis-build-test.sh
@@ -2,5 +2,5 @@
 cd src
 ./autogen.sh
 ./configure --with-realtime=uspace --disable-check-runtime-deps --enable-build-documentation
-make -O -j$(nproc) V=1
+make -O -j$(getconf _NPROCESSORS_ONLN) default pycheck V=1
 ../scripts/rip-environment runtests
--- a/scripts/travis-build-test.sh
+++ b/scripts/travis-build-test.sh
@@ -1,6 +1,6 @@
-#!/bin/sh -e
+#!/bin/sh -ex
 cd src
 ./autogen.sh
 ./configure --with-realtime=uspace --disable-check-runtime-deps --enable-build-documentation
-make -O -j$(nproc)
+make -O -j$(nproc) V=1
 ../scripts/rip-environment runtests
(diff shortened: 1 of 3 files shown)


rene-dev:
Hi,
Thanks for all the PRs, are you on irc or something? we should coordinate to avid duplicate work.
I already had the idea for such a check, and it should cover the entire project.
there seems to be a problem with the awk part, it doesnt extract the path properly.

dwrobel:
>are you on irc or something?

I subscribed to: `emc-developers@lists.sourceforge.net` and I'm registered/joined to `#linuxcnc-devel` as `dwrobel`.

>I already had the idea for such a check, and it should cover the entire project.

I fully agree, however it should be gradually extended - to cover only that parts which is ready to be verified _(otherwise the CI will fails permanently)_. BTW, I added to the latest version checking of `hal` _(apart from `../tests` which was already there)_.

>there seems to be a problem with the awk part, it doesnt extract the path properly.

I verified that on `Fedora 31`, `Ubuntu 16.04`, and `Debian unstable` and it works for me. I would appreciated if you could share some more details what doesn't work and on which system.

Here is an output excerpt from [debian:unstable ](https://hub.docker.com/_/debian/):
```
$ cat /etc/debian_version 
bullseye/sid
$ find ../tests  -type f -exec file {} \; | grep 'Python script' | awk '{sub(/:/, "", $1); print $1}' | head -n3
../tests/limit3/checkresult
../tests/limit3/sunny-day/checkresult
../tests/halmodule.0/test.py
```

rene-dev:
`find ../tests -type f -exec file {} \; | grep 'Python script' | awk '{sub(/:/, "", $$1); print $$1}'`
`../tests/motion-logger/mountaindew/test-ui.py Python script, ASCII text executable
../tests/motion-logger/basic/test-ui.py Python script, ASCII text executable
../tests/motion-logger/startup-gcode-abort/test-ui.py Python script, ASCII text executable`
if I run everything I get
`IOError: [Errno 2] No such file or directory: 'Python'`
im running buster with
`GNU Awk 4.2.1, API: 2.0 (GNU MPFR 4.0.2, GNU MP 6.1.2)`

PR #811, opened 2020-04-29
Fix some major bugs in xhc-whb04b-6 user component

jasenk2:
1.There is wrong code for LEAD position of feed rotary switch.
2.There was major mistake in parse procedure Pendant::processEvent values are tested against button codes for invalid code but object is send as null to actual Pendant::processEvent so segfault coredump was produced

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,4 +38,3 @@ oprofile*
 position.txt
 *.9
 *.glade.h
-/build/
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ oprofile*
 position.txt
 *.9
 *.glade.h
+/build/
--- a/configs/.gitignore
+++ b/configs/.gitignore
@@ -13,3 +13,5 @@ position*.txt
 *.ini.expanded
 hallib
 *_cmds.hal
+/Rack_Tool_Changer_simplified/
+/MPG 1 DZIŁAJACY/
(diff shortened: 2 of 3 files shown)


alkabal:
Well i really need to check what is true for lead but for the rest i think you are right already.

Can you add this patch line1425 file pendant.cc ?
replace : 
            axisJogStepSize = feedButton.stepSize() * 0.001 * mHal.getFeedOverrideMaxVel();
with : 
            axisJogStepSize = feedButton.stepSize() * 0.0001 * mHal.getFeedOverrideMaxVel();

Can you check something at same time ?

The args B from cmd line does not work fine with associated var mIsStepMode_5_10

Thanks

jasenk2:
ok Alkabal i will try

.gitignore
```diff
@@ -38,3 +38,4 @@ oprofile*
 position.txt
 *.9
 *.glade.h
+/build/
```
jepler: The .gitignore file commited to git reflects only files that are created during the build process.  This path /build/ is not.  You can add your own ignores in $GIT_DIR/info/exclude -- see man gitignore(5) for more information.

configs/.gitignore
```diff
@@ -13,3 +13,5 @@ position*.txt
 *.ini.expanded
 hallib
 *_cmds.hal
+/Rack_Tool_Changer_simplified/
+/MPG 1 DZIŁAJACY/
```
jepler: see above comment.

PR #827, opened 2020-05-04
Fix linking error on gcc10

dwrobel:
Fixes the following linking error:
 /usr/bin/ld: objects/hal/classicladder/files_project.o:
 /home/sw/projects/rpmbuild/BUILD/linuxcnc-bb2a4cc21a7fed9574059d46eb4507c40ba3bf55 \
 /src/hal/classicladder/files_project.c:163: multiple definition of `FileName'; \
 objects/hal/classicladder/files.o:/home/sw/projects/rpmbuild/BUILD/ \
 linuxcnc-bb2a4cc21a7fed9574059d46eb4507c40ba3bf55/src/hal/classicladder/files.c:1303: \
 first defined here

### DIFF
--- a/src/hal/classicladder/files.c
+++ b/src/hal/classicladder/files.c
@@ -1300,7 +1300,7 @@ void DeleteTheDefaultSection( )
 	SectionArray[ 0 ].Used = FALSE;
 }
 
-char FileName[500];
+static char FileName[500];
 void LoadAllLadderDatas(char * DatasDirectory)
 {
 	ClassicLadder_InitAllDatas( );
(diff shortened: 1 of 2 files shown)


dwrobel:
I assumed that those variables could be `static` and not be the common global variable. If the assumption is wrong please let me know I'll rewrite the patch.

rene-dev:
I wonder why only gcc 10 complains, this should always be a linker error.  did you check if classicladder still works?

dwrobel:
>I wonder why only gcc 10 complains,

See https://gcc.gnu.org/gcc-10/porting_to.html, long story short:
```
GCC 10 defaults to -fno-common
```

>did you check if classicladder still works?

All tests still passes. If there are more specific tests which needs to be performed, please let me know.

PR #830, opened 2020-05-05
Fix gcc version 10 too old: skipping hal/user_comps/xhc-whb04b-6

dwrobel:


andypugh:
Are you sure that is better? 
-dumpfullversion is certainly an improvement, but I have a feeling that alphabetical sorting might be more robust than >=

dwrobel:
>alphabetical sorting

As per man `gcc(1)` the `--dumpfullversion` gives us numbers:
>The output is always three numbers separated by dots, major, minor and patchlevel version.

We're also interested in comparing numbers. So, I'm not quite convinced why we need to complicate it and involve alphabet?

dwrobel:
What I could do is to replace `sed` by `awk` and use:

```
gcc -dumpfullversion | awk -F '.' '{print $1*10000+$2*100+$3}'
```

PR #832, opened 2020-05-05
Fix SyntaxWarning: "is" with a literal - pyngcgui.py

dwrobel:
Fixes the following warning when compiling for python3.8:

 linuxcnc/lib/python/pyngcgui.py:1424: SyntaxWarning: "is" with a literal. Did you mean "=="?
   if comment is '':


dngarrett:
merged to master

PR #836, opened 2020-05-06
Remove duplicate checking for dblatex

dwrobel:

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1136,11 +1136,6 @@ fi
 
 # Programs required only for building the PDF documentation
 if ( test "$BUILD_DOCS_PDF" = "yes" ) ; then
-    AC_PATH_PROG(DBLATEX,dblatex,"none")
-    if ( test "none" = "$DBLATEX" ) ; then
-	AC_MSG_ERROR([no dblatex, PDF documentation cannot be built])
-    fi
-
     AC_PATH_PROG(PDFLATEX,pdflatex,"none")
     if ( test "none" = "$PDFLATEX" ) ; then
 	AC_MSG_ERROR([no pdflatex, PDF documentation cannot be built])


src/configure.ac
```diff
@@ -1131,15 +1131,8 @@ Documentation cannot be built.])
     if ( test "none" = "$GS" ) ; then
 	AC_MSG_ERROR([no gs, documentation cannot be built])
     fi
-fi
 
 # Programs required only for building the PDF documentation
-if ( test "$BUILD_DOCS_PDF" = "yes" ) ; then
```
jepler, line 1137: Doesn't this change make pdflatex required for just building html docs only?  I bet dblatex and pdflatex are only needed for pdf docs, and the first check for dblatex is the one to delete (and the version check moved down)

dwrobel: >Doesn't this change make pdflatex required for just building html docs only?

No, it just removes the duplicate check for the `dblatex`. The intention was to not introduce any functional changes to configure.ac.

More precisely, the first check for `dblatex` happens here:
https://github.com/LinuxCNC/linuxcnc/blob/f8a0a5ea981ff5e32ce5dd145b6743c3751428bd/src/configure.ac#L1103-L1106

and is under the [if ( test "$BUILD_DOCS" = "yes" ) ; then](https://github.com/LinuxCNC/linuxcnc/blob/f8a0a5ea981ff5e32ce5dd145b6743c3751428bd/src/configure.ac#L1067) anyway.

jepler: Why was the 'fi ... if' removed in this PR?

In origin/master I can `configure --enable-build-documentation=html` even if there's no `pdflatex` program; and the doc buld will finish.

In your branch, this fails at the configure stage stating that there's no pdflatex.

As far as I can tell, it's because you deleted the `fi ... if` and made the pdflatex check apply even when building only html docs.

```
$ sudo chmod 000 /usr/bin/pdflatex
$ ./autogen.sh  && ./configure --enable-build-documentation=html && make -j5 docs
...
configure: error: no pdflatex, PDF documentation cannot be built
```

dwrobel: Ok, got it.

PR #837, opened 2020-05-06
Fix configure "integer expression expected" error

dwrobel:
Fixes error when micro version part contains non-digits characters.

Specifically for:

 $ dblatex --version
 dblatex version 0.3.11py3

which is available on Fedora the configure script produces:

 checking dblatex version... ./configure: line 7138: \
   test: 11py3: integer expression expected

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1108,8 +1108,9 @@ EOF
     AC_MSG_CHECKING([dblatex version])
     set -- `dblatex --version`; DBLATEX_VER=$3
     set -- `echo $DBLATEX_VER | sed 's/[[.-]]/ /g'`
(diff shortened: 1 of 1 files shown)


jepler:
Since the check is for now-impossibly-old versions of dblatex, it can probably also be dropped if necessary.

dwrobel:
>Since the check is for now-impossibly-old versions of dblatex, it can probably also be dropped if necessary.

The check might become again useful also for relatively new versions, e.g. `0.3.11py3` seems to be completely [broken](https://bugzilla.redhat.com/1833095), especially with the new [inkscape >1.x](https://bugzilla.redhat.com/1833047).

PR #839, opened 2020-05-07
Fix generating pdf documentation using python3

dwrobel:


dwrobel:
Note: this leaves intact python code in the documentation text.

phillc54:
The six module imported into docs/source/image-wildcard appears to be missing from the Buildbot.
https://github.com/LinuxCNC/linuxcnc/issues/860#issuecomment-650475898

PR #842, opened 2020-05-08
Fix generating html documentation using python3

dwrobel:
Adds a possibility to owerwrite 'checklink' by using:

  $ ac_cv_path_CHECKLINK=<path-to-executable> ./configure

Note: Specifying '/usr/bin/true' as follows:

  $ ac_cv_path_CHECKLINK=/usr/bin/true ./configure

can be used to build the html documentation on machines
(usually builders) which by design disables network access
during a build process.


jepler:
Checklink is supposed to be excluding http(s) URLs.  Is this not working as intended?  Are additional exclusion patterns needed?
```
../docs/src/checkref:    checklink --quiet --exclude "http(s?)://" $F | tee $OUT 2>&1
```

dwrobel:
> Is this not working as intended?

First, the change fixes the proper interpreter [here](https://github.com/dwrobel/machinekit/commit/dfbf4c3f6301aa25d80f4af18c3ef3cb6f7bd589#diff-d2fa4818a8181b36b02edeaf193d88dbR502).

Secondly, `w3c-linkchecker` is not packaged for Fedora, so you end-up in something like:
```
configure: error: no checklink, HTML documentation cannot be built
install with "sudo apt-get install w3c-linkchecker"
```
Normally you might expect that you can override it passing to the `configure` `ac_cv_path_CHECKLINK=your-favorite-tool` but that wasn't supported, so I added it.

Taking advantage of it I could you use it either with:
-  passing `/usr/bin/true` for building `rpm` packages _(based on the assumption that someone already had verified all the `URLs`)_,
- specifying my script which would internally use [linkchecker](https://linkchecker.github.io/linkchecker/) - that's the equivalent of `checklink` we have available in Fedora - which might check all `URLs'` _(including externals)_ - as that makes sense when you have plans to verify modified documentation.

jepler:
Thanks for clarifying the motivation of the link checker change. I did not question the main motivation of this change but incidental changes should also be well motivated.

PR #845, opened 2020-05-10
Add the preselection of the parallel ports.

nicokid:
Signed-off-by: nicokid <nicola@quargentan.com>


jepler:
Hi and thanks for your pull request!

I looked at this but I don't understand what the intended change is.  Can you tell me more?  You can add screenshots or other images in the comments if it's helpful.

On the page "Parallel port 1", the "output pins presets" disappeared ("before" image is first, "after" image is second), and I didn't find it anywhere else.  Was this intentional?

Screenshot_2020-05-17_09-15-50
Screenshot_2020-05-17_09-17-22

On the "parallel port 2" page, some items moved around but no items were added or removed.

Screenshot_2020-05-17_09-16-00
Screenshot_2020-05-17_09-17-33

nicokid:
Hi jepler,
Sorry for my mistake. After resettinig my repository I had to rebuild the pport1 window mixing 3 different versions. I corrected the problem.

For the pport2 window you can see there is a small arrow near the number 1. If the /proc/ioports file exists and inside there is the reference to the parallel port, then the list of selectable addresses appears.
Nicola.

jepler:
Thanks, that is clearer. I'll check it again soon.

jepler:
It's worth noting that a sufficiently recent Linux kernel will always show the addresses in /proc/ioports as "0000" for non-root users reading the file.  https://github.com/torvalds/linux/commit/51d7b120418e99d6b3bf8df9eb3cc31e8171dee4

This affects Linux 4.19.0 on my everyday system, for example.  I get a list box with "0x0000" in it twice.

May I suggest that you filter out 0x0000 if it is ever returned as an address.

In general, I had assumed it was preferable to enter port *numbers* 0, 1, 2, ... which theoretically works in conjunction with the kernel's parallel port detection.  It's possible that even though this was added years ago the documentation and examples have not caught up.

(In this case, port numbers are read from `/proc/sys/dev/parport/parport#/base-addr` as root, so it "should work" even if the base-addr file is made inaccessible for non-root users for security in a future kernel version)

nicokid:
Hello,
my debian 7.11 still works with /proc/ioports so today I added a parallel card on a pc and then I installed ubuntu 20.04. Unfortunately I discovered that python-gtk2 no longer exists on ubuntu and so I can't compile linuxcnc...
Ok, I can remove 0x0000 and read **/proc/sys/dev/parport/parport#/base-addr** if the **/proc/ioports** file does not exist. I have no idea how the parallel port is handled by linuxcnc. But from what I've seen it uses addresses and not port numbers. There are no addresses in /proc/sys/dev/parport/parport#/.
We can also check if the device /dev/parport# exists. Is it possible to indicate the port number instead of the address in linuxcnc?

Nicola

jepler:
Right, as documented in `man hal_parport`,
> The port_addr parameter of the configuration string may be either the physical base address of a parallel port or specified as the detected parallel port via Linux parport_pc driver.  In which case, a port_addr of 0 is the first parallel port detected on the system,1is the next, and so on.

nicokid:
Hello Jeff,
yes I read the documentation. I modified the script to read the file /proc/sys/dev/parport/parport#/base-addr when I can't find the address in /proc/ioports.
But at this point can I completely ignore /proc/ioports and read just /proc/sys/dev/parport/parport#/base-addr?
Thanks.

jepler:
If you are aware that the parallel port indexes 0, 1, 2, ... can be used then I am not sure why you want to use/display parallel port addresses 0xabcd at all.  Can you tell me why?

nicokid:
Yes, perhaps the user does not understand .
I'm not a fan of editable combo boxes....
Do you think we can completely remove the custom text box and leave only the automatic search?
Or I can add a check box to enable or disable (grayed out) the custom text box and the combo box.

jepler:
If there's no need for the user to type something in, then the editable
textbox should be removed.

nicokid:
Hello,
I was a little busy.
I need a few days to make this change.
Nicola.

jepler:
Thank you for sticking with us through this process!

src/emc/usr_intf/stepconf/stepconf.py
```diff
@@ -918,26 +918,72 @@ def find_parport(self):
...
+                        except:
+                            print ("Unable to open %s" % baseaddr )
+                            continue
+                        # read base-addr file
+                        try:
+                            for line in in_file:
+                                # get init_address and number of port
+                                lline=line.split()
+                                dec_address=lline[0].strip()
+                                port_number=lline[1].strip()
```
dwrobel: The code says here *"number of port"* but I couldn't find that on the [documentation](https://www.kernel.org/doc/html/v5.6/admin-guide/parport.html) which mention that this file contains only addresses. In my case the file contains "888 1912" _(kernel : 5.6.17rt9)_.

Apart from that running `stepconf` on the existing configuration changed it as following:
```
$ git diff parport.stepconf parport/parport.hal
diff --git a/parport.stepconf b/parport.stepconf
index 352a87f..5e351d7 100644
--- a/parport.stepconf
+++ b/parport.stepconf
@@ -28,13 +28,14 @@
   <property name="floatsout" type="float" value="10.0"/>
   <property name="halui" type="bool" value="False"/>
   <property name="halui_list" type="eval" value="[]"/>
-  <property name="ioaddr" type="string" value="0x378"/>
+  <property name="ioaddr" type="string" value="0"/>
   <property name="ioaddr2" type="string" value="1"/>
   <property name="ioaddr3" type="string" value="2"/>
   <property name="ladderconnect" type="bool" value="True"/>
   <property name="ladderhaltype" type="int" value="0"/>
   <property name="laddername" type="string" value="custom.clp"/>
   <property name="latency" type="float" value="22000.0"/>
+  <property name="lparport" type="eval" value="['0']"/>
   <property name="machinename" type="string" value="parport"/>
   <property name="manualtoolchange" type="bool" value="True"/>
   <property name="md5sums" type="eval" value="[]"/>
diff --git a/parport/parport.hal b/parport/parport.hal
index ff63696..0cb56a1 100644
--- a/parport/parport.hal
+++ b/parport/parport.hal
@@ -1,9 +1,9 @@
-# Generated by stepconf 1.1 at Sat Jun  6 13:56:45 2020
+# Generated by stepconf 1.1 at Mon Jun 15 22:59:59 2020
 # If you make changes to this file, they will be
 # overwritten when you run stepconf again
 loadrt [KINS]KINEMATICS
 loadrt [EMCMOT]EMCMOT base_period_nsec=[EMCMOT]BASE_PERIOD servo_period_nsec=[EMCMOT]SERVO_PERIOD num_joints=[KINS]JOINTS
-loadrt hal_parport cfg="0x378 out"
+loadrt hal_parport cfg="0 out"
 loadrt stepgen step_type=0,0,0
 loadrt pwmgen output_type=1
 ```

So the `0x378` value is basically gone.

Comment (nicokid) [reply src/emc/usr_intf/stepconf/stepconf.py]: 888 decimal == 378 hex.
1912 decimal == 778 hex.
Do you have a card with two parallel ports? Can you post how your directory is made?
**ls /proc/sys/dev/parport/**

Nicola.

Comment (dwrobel) [reply src/emc/usr_intf/stepconf/stepconf.py]: >Do you have a card with two parallel ports?

No, it is built-in parallel port.

>Can you post how your directory is made?

```
$ ls -R /proc/sys/dev/parport/
/proc/sys/dev/parport/:
default  parport0

/proc/sys/dev/parport/default:
spintime  timeslice

/proc/sys/dev/parport/parport0:
autoprobe  autoprobe0  autoprobe1  autoprobe2  autoprobe3  base-addr  devices  dma  irq  modes  spintime

/proc/sys/dev/parport/parport0/devices:
active

```

nicokid:
Hi,
strange, I don't understand why together with the address 888 there is also 1912.

About you question, as documented the **hal_parport** parameter accepts both the port number and the address. Jeff finds it useless to use the address.
I don't know if this can cause problems.
Nicola.

PR #847, opened 2020-05-10
apply G54-G59.3 rotation to halui.axis.x.pos-relative and halui.axis.y.pos-relative

a-downing:
When using coordinate system rotation, the value reported by halui.axis.x.pos-relative and halui.axis.y.pos-relative are incorrect and do not match the value reported by the axis DRO. This change applies the rotation to these values.


andypugh:
Thanks.

PR #848, opened 2020-05-11
Fix TypeError: a bytes-like objectis required, not'str’ (fixes #840)

### LINKED ISSUE #840: Stepconf broken in master with python2
build_ini.py has on line 38:
importlib.reload(sys)

Which seems is python3 only

In stepconf.py line 53 there is:
reload(sys)

Would that be python2 only?

changing the build-ini.py to reload(sys) fixes things in python2

I'm actually not not sure of the python version policy for master are we supporting both or transitioning to python3 only?

dwrobel:
Fixes the following exceptions:

For python3:
  File "/home/sw/projects/machinekit/linuxcnc/lib/python/stepconf/pages.py", line 91, in on_button_fwd_clicked
    self['%s_finish'%name]()
  File " fhome/sw/projects/machinekit/linuxcnc/lib/python/stepconf/pages.py", line 881, in finished_finish
    self.a.buid_config()
  File "/home/sw/projects/machinekit/linuxcnc/bin/stepconf”, line 795, in buid_config
    self.d.save(base)
  File "/home/sw/projects/machinekit/linuxcnc/bin/stepcont", line 680, in save
    d.writexml(open((filename, "wb"), addindent=" ", newl="\n")
  File "/usr/lib/python3.7/xml/dom/minidom.py", line 1791, in writexml
    writer.write('<?xml version="1.0" ?>'+newl)
  TypeError: a bytes-like objectis required, not'str’

Above generates empty ~/.stepconf-preferences which in turns triggers the next exception
(applicable for both python2/3):

  File "/home/sw/projects/machinekit/linuxcnc/bin/stepconf", line 1576, in <module>
    app = StepconfApp(False)
  File "/home/sw/projects/machinekit/linuxcnc/bin/stepconf", line 770, in __init__
    self.d.load_preferences()
  File "/home/sw/projects/machinekit/linuxcnc/bin/stepconf", line 515, in load_preferences
    d = xml.dom.minidom.parse(open(filename, "r"))
  File "/usr/lib64/python2.7/xml/dom/minidom.py", line 1918, in parse
    return expatbuilder.parse(file)
  File "/usr/lib64/python2.7/xml/dom/expatbuilder.py", line 928, in parse
    result = builder.parseFile(file)
  File "/usr/lib64/python2.7/xml/dom/expatbuilder.py", line 211, in parseFile
    parser.Parse("", True)
xml.parsers.expat.ExpatError: no element found: line 1, column 0

For the last one the warning-dialog is now diplayed depicting the problem to the user.

### DIFF
--- a/src/emc/usr_intf/stepconf/stepconf.py
+++ b/src/emc/usr_intf/stepconf/stepconf.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env linuxcnc-python
 # -*- encoding: utf-8 -*-
 #
 #    This is stepconf, a graphical configuration editor for LinuxCNC
@@ -49,6 +49,11 @@ import hal
 import shutil
 import time
 from multifilebuilder_gtk3 import MultiFileBuilder
+
+try:
+    from defusedexpat import pyexpat as expat
+except ImportError:
+    from xml.parsers import expat
 
 if sys.version_info[0] == 3:
     import subprocess
@@ -512,7 +517,7 @@ class Data:
         filename = os.path.expanduser("~/.stepconf-preferences")
         if os.path.exists(filename):
             version = 0.0
-            d = xml.dom.minidom.parse(open(filename, "r"))
+            d = xml.dom.minidom.parse(open(filename, "rt"))
             for n in d.getElementsByTagName("property"):
                 name = n.getAttribute("name")
                 text = n.getAttribute('value')
@@ -577,7 +582,7 @@ class Data:
         n2.setAttribute('name', "machinename")
         n2.setAttribute('value', str("%s"%self.machinename))
 
-        d2.writexml(open(filename, "wb"), addindent="  ", newl="\n")
+        d2.writexml(open(filename, "wt"), addindent="  ", newl="\n")
 
     def load(self, filename, app=None, force=False):
         def str2bool(s):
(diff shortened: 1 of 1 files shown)


src/emc/usr_intf/stepconf/stepconf.py
```diff
@@ -50,6 +50,11 @@
 import time
 from multifilebuilder_gtk3 import MultiFileBuilder
 
+try:
+    from defusedexpat import pyexpat as expat
+except ImportError:
+    from xml.parsers import expat
```
jepler, line 56: this change seems unrelated to the main purpose of the PR. Can you try to separate these out more cleanly in the future?  (note: defusedexpat is not packaged in debian, but since it apparently degrades gracefully to the "normal" expat all is well)

dwrobel:
>this change seems unrelated to the main purpose of the PR.

We need `expat` as it is used [here](https://github.com/LinuxCNC/linuxcnc/pull/848/files#diff-f9fc25de596bb6877b5af1f9110e3cbfR777) to catch errors in the configuration file.

>defusedexpat is not packaged in debian

If for any reason you prefer to remove importing `defusedexpat`, please let me know I can easily remove it.

dwrobel:
>Can you try to separate these out more cleanly in the future?

Importing `defusedexpat` removed in https://github.com/LinuxCNC/linuxcnc/pull/853.

PR #849, opened 2020-05-11
Fix using dumpfullversion for old compilers and clang

dwrobel:
Fixes (non-fatal) error:
 gcc: fatal error: no input files
produced by compilers which does not support -dumpfullversion
(includes old gcc and all clang).

It also uses $(CC) instead of hardcoded 'gcc' value.


jepler:
Does this lead to the component being built or unbuilt when CC is clang?  If unbuilt, is that important to you?

dwrobel:
>Does this lead to the component being built or unbuilt when CC is clang?

It will build unconditionally when `CC=clang`.

Issue #846, opened 2020-05-10
Master build appears to be broken with Buster. 


andypugh:
Git bisect blames:
https://github.com/LinuxCNC/linuxcnc/commit/94fa1b2338a8c3586547ab50233e1d51c0957d98

The error is:
scripts/Makefile.build:50: *** CFLAGS was changed in "/home/andypugh/linuxcnc-dev/src/Makefile". Fix it to use ccflags-y.  Stop.
make[1]: *** [Makefile:1544: _module_/home/andypugh/linuxcnc-dev/src] Error 2
make: *** [Makefile:479: modules] Error 2

gcc version 8.3.0 (Debian 8.3.0-6)

PR #850, opened 2020-05-11
Fix honoring {C,CXX,CPP,LD}FLAGS passed to the 'configure'

dwrobel:
This fixes linking against libreadline v5 in a non standard directory.

On Fedora multiple readline libraries can be installed in parallel.
Old GPL-2 compatible libreadline v5 has its own header and library in the
following directories:

 $ rpm -ql compat-readline5-devel | grep -e 'readline\.h' -e 'libreadline\.so'
 /usr/include/readline5/readline/readline.h
 /usr/lib64/readline5/libreadline.so

As per 'configure --help' the following flags (excerpt) should be supported:
 Some influential environment variables:
  LDFLAGS     linker flags, e.g. -L<lib dir> if you have libraries in a
              nonstandard directory <lib dir>
  CPPFLAGS    (Objective) C/C++ preprocessor flags, e.g. -I<include dir> if
              you have headers in a nonstandard directory <include dir>

Despite passing above directories via 'CPPFLAGS' and 'LDFLAGS' to 'configure' the linking
process fails as follows:

 gcc -Wl,-rpath-link,../lib -ltirpc  -o ../bin/halcmd objects/hal/utils/halcmd.o
   objects/hal/utils/halcmd_commands.o objects/hal/utils/halcmd_main.o
   objects/hal/utils/halcmd_completion.o ../lib/liblinuxcncini.so.0
   ../lib/liblinuxcnchal.so.0 -lreadline
 /usr/bin/ld: cannot find -lreadline

This was due to not honoring flags passed to configure.

This also fixes the error of passing C specific compielr option to C++ compiler:
   cc1plus: warning: '-Werror=' argument '-Werror=implicit-function-declaration'
   is not valid for C++


dwrobel:
Comparing with old #751 it adds [ifeq ($(BUILD_SYS),uspace)](https://github.com/LinuxCNC/linuxcnc/pull/850/files#diff-7d087b07875afb09211b54c9cc23c35aR119) guard to make sure that none of the following:
 - CPPFLAGS
 - CXXFLAGS
 - CXXFLAGS
 - CFLAGS
 - LDFLAGS

would affect `RTAI` compilation.

dwrobel:
Should fix https://github.com/LinuxCNC/linuxcnc/issues/846.

jepler:
Thanks, I will submit it to the buildbot CI system before merging.

BTW, do you think that CPPFLAGS is redundant and should be merged into CXXFLAGS?  I could not determine why we have the two variables.  It is nothing to do with flags for the C preprocessor, sometimes called cpp, as far as I can tell.

dwrobel:
> I could not determine why we have the two variables.

It's almost self explanatory, see what `configure` print outs:
```
$ ./configure --help | grep CPPFLAGS
  CPPFLAGS    (Objective) C/C++ preprocessor flags, e.g. -I<include dir> if
```
so, basically `CPPFLAGS`  is not compiler specific, while `CXXFLAGS`:
```
$ ./configure --help | grep CXXFLAGS
  CXXFLAGS    C++ compiler flags
```
is only for `c++`. In other words if you would like to specify location for the `readline5` headers _(e.g. -I/usr/include/readline5)_ it's useful to just pass it using `CPPFLAGS` as this way all compilers will pick up this parameter.

jepler:
Thanks!

PR #853, opened 2020-05-12
Remove import of defusedexpat

dwrobel:


jepler:
Thanks, but it's harmless to us to leave it there so I'm going to close this up un-merged.

However, I hope you'll be more mindful in the future about putting unrelated changes in a single pull request, just in case the unrelated change is one that is not so innocuous.

PR #856, opened 2020-05-17
M99-endless-main-program test fails erratically, skip it

jepler:


andypugh:
Do we know _why_ it fails erratically? 
Though I doubt that M99 sees a lot of use.

jepler:
No, I didn't investigate closely. It is one missing message at shutdown. I support someone getting to the bottom of it, but a flaky test is worse then no test when it gets in the way of finding actual new regressions.

zultron:
I created #857 to track this one.  @jepler I'd appreciate if you could provide more information about the failures, if not a link to CI, then information about the environment where you saw the failure.  Thanks!

PR #861, opened 2020-05-19
Fix generating config files

dwrobel:
Removes .encode() as it produces bytes() not str() which
generates errors during parsing of configuration files.


dwrobel:
> The history of these changes seems to indicate that the "encode" was added for py3 compatibility at 879a219

But it's not clear why they were added only to some of the lines and not to all of them?
In practise those changes broke `python3` version.

jepler:
@dwrobel can you please provide more information about how to reproduce the problem you are fixing?  With the information provided I can't determine whether this change is appropriate or not.

I configured LinuxCNC master branch (8c87078c1f) `--with-python=python3` and verified that stepconf says `#!/usr/bin/python3`.  Then I proceeded to breeze through all pages of stepconf until an output file was written.  I repeated this process with the following locales enabled: C.UTF-8, de_DE.UTF-8 and en_US.UTF-8.  I did not encounter any errors, though several PyGTKDeprecationWarnings unrelated to character encodings were displayed.

dwrobel:
If you would generate a brand new configuration _(I generated conf-861 using only defaults)_ then please search for `^b'#`:
```
$ grep -e "^b'" conf-861/*.hal
conf-861/conf-861.hal:b'# If you make changes to this file, they will be'
conf-861/conf-861.hal:b'# overwritten when you run stepconf again'
conf-861/custom.hal:b'# This file will not be overwritten when you run stepconf again'
conf-861/custom_postgui.hal:b'# This file will not be overwritten when you run stepconf again'
conf-861/postgui_call_list.hal:b'# If you make changes to this file, they will be'
conf-861/postgui_call_list.hal:b'# overwritten when you run stepconf again'
conf-861/pyvcp_options.hal:b'# If you make changes to this file, they will be'
conf-861/pyvcp_options.hal:b'# overwritten when you run stepconf again'
```
Then run `linuxcnc` and select the aforementioned configuration. You should see something similar to my output:
```
$ linuxcnc
LINUXCNC - 2.9.0~pre0
Machine configuration directory is '/home/sw/linuxcnc/configs/conf-861'
Machine configuration file is 'conf-861.ini'
Starting LinuxCNC...
Found file(REL): ./conf-861.hal
./conf-861.hal:2: Unknown command 'b'# If you make changes to this file, they will be''
    /home/sw/linuxcnc_debug.txt
and
    /home/sw/linuxcnc_print.txt
```
Basically the error:
`./conf-861.hal:2: Unknown command 'b'# If you make changes to this file, they will be''`
is the result of using `.encode('utf-8')` which this PR is trying to fix.

dwrobel:
FYI: At the time I prepared that fix I was able to generate the configuration using 'stepconf' and then run linuxcnc to drive 3 stepping motors connected to the parallel port.
Everything worked fine with python3 version:
```
$ python --version
Python 3.7.7
```

jepler:
I requested Rene to look at this via irc as well, but nothing happened.

PR #878, opened 2020-05-30
Add vectorized version of linuxcnc icon

dwrobel:
Image source taken from:
 https://forum.linuxcnc.org/media/kunena/attachments/20660/linuxlogo.png
then vectorized using inkscape and will be installed as:
 /usr/share/icons/hicolor/scalable/apps/linuxcncicon.svg


dwrobel:
LinuxCNC icons are now much more readable _(tested in GNOME 3)_ than the png version.

andypugh:
Do you have any idea how compatible this is with different distributions and desktop environments? I am generally in favour of vector over bitmap, but would prefer bitmap in everything over nothing in some cases. 
I have tested in Wheezy with XFCE. and that works. 
I am not sure that the Penguin + Text is better than just the penguin. I feel that a vectorised version of the existing logo might be better as the new SVG looks a little cramped. 

The original artwork is here: http://www.fennetic.net/portfolio/

dwrobel:
> Do you have any idea how compatible this is with different distributions and desktop environments? I am generally in favour of vector over bitmap, but would prefer bitmap in everything over nothing in some cases.

I found references, in the [icon-theme](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#changes) specification, about SVG support dated back to 2002 year. So, hopefully it is long enough to be widely supported.

> I have tested in Wheezy with XFCE. and that works.

Good to hear that. It also works flawlessly on Fedora 31 _(GNOME 3.24.2)_ for me.

> I am not sure that the Penguin + Text is better than just the penguin. I feel that a vectorised version of the existing logo might be better as the new SVG looks a little cramped.
> 
> The original artwork is here: http://www.fennetic.net/portfolio/

Yeah, it would nice if you could find the original svg file. The page mentioned: _"illustrations done in Inkscape for EMC2"_, although it contain a link only to the [png](http://www.fennetic.net/portfolio/thumbnails/tn_chips2.png) version.

andypugh:
The _parts_ of the logo can be found as SVG here: http://www.fennetic.net/art/originals/
(chips2.svg)

dwrobel:
> The _parts_ of the logo can be found as SVG here: http://www.fennetic.net/art/originals/
> (chips2.svg)

I've tried to reconstruct the icon from the pieces. Please do not shoot if the result is not satisfactory :-).

>I am not sure that the Penguin + Text is better than just the penguin.

I added the text as well to not only keep the original design, but IMHO it also looks better for a longer names _(e.g. Integrator Infor.)_, plase have a look at the screenshot:

linuxcnc-icons-gnome-screenshot

andypugh:
We are getting in to the area of art criticism here. I like vector graphics, I like the logo above. But I think that as an icon it is too small and fussy. 
But this is a matter of taste, I am not going to argue if you disagree. 

Screenshot 2020-06-06 at 00 48 32

dwrobel:
>But I think that as an icon it is too small and fussy.

While I agree that it doesn't look nice it doesn't necessarily mean that we could easily start blaming the icon itself.

The SVG version is saved as 256x256 while on your screen-shot is looks like it is 64x64 [px]. So, the key question why it got downscaled so drastically? Not to mention, that if it's SVG then it can be both {down/up}-scaled without degrading quality.

Based on above would you consider to start blaming your Desktop Environment?

andypugh:
The screenshot might be a bit of an extreme case, that's a 2560x1440 desktop with default icon sizes. I don't know what the typical resolution is for a LinuxCNC system. Probably lower than that. 

As I said, I don't want to get into an argument about artistic merits. I am in favour of a scalable icon.

dwrobel:
> that's a 2560x1440 desktop with default icon sizes

That means that the size of the icons on the screen is less than 64x64, maybe 48x48?

But looking closely at your screen-shot, It looks to me, that you are not using the SVG version of the icon for linuxcnc, am I right?

Could you please compare the bottom part of the linuxcnc icon in your [screen-shot](https://user-images.githubusercontent.com/1010653/83930453-99c1d380-a78f-11ea-91ae-22541fd47eb4.png) with my [screen-shot](https://user-images.githubusercontent.com/228873/83918657-7a1fb080-a779-11ea-8f09-847548f911b3.png) - more precisely, you should not see the the pengiun's foot below the "LinuxCNC" text in the SVG version.

andypugh:
It's definitely linuxcncicon.svg, but your first version.

dwrobel:
Then it looks more like a Desktop Environment problem. FYI, on screen 1920x1080 - Gnome displays icons 96x96 [px]. On the one hand, extrapolating that to 2560x1440 resolution - you would need to have icons 128x128 to get comparable user experience to what I'm observing. On the other hand, even if I scale down my screen-shot to the lower resolution all icons still looks better than yours. Maybe the scaling algorithm is broken in your DE?

andypugh:
XFCE from the old LinuxCNC Live-CD. 
And yes, it was showing all icons at 32x32.

andypugh:
Just found https://forum.linuxcnc.org/show-your-stuff/32672-linuxcnc-logo?start=20#95199
We had an SVG logo 3 years ago....

PR #877, opened 2020-05-30
Improve searching LinuxCNC components from GNOME menu

dwrobel:
Signed-off-by: Damian Wrobel <dwrobel@ertelnet.rybnik.pl>

PR #879, opened 2020-06-04
ci: build debs

jepler:
This might have detected regressions in the packaging system that widely affected different architectures, such as #877.


jepler:
(It wouldn't have, because Debian Jessie was the newest distro affected by the particular problem in #877)

PR #880, opened 2020-06-05
Improve searching LinuxCNC components in GNOME menu

dwrobel:


jepler:
I locally checked this against the jessie version of of desktop-file-validate, 0.22.

Issue #579, opened 2019-04-23
Uncommanded moves when hitting stop


sheetcam:
## This is what I expected to happen:
When I press stop the machine should come to a stop as quickly as practical then stop the spindle.

## This is what happened instead:
This issue is intermittent but sometimes when using stop to stop a running program the machine stops the spindle and moves to some seemingly random point. I have seen two different symptoms. It either moves at around feed rate to a point then stops or it moves slowly and just keeps going until it hits a limit or you hit emergency stop. On my router this mainly seems to happen on X and Y. On my mill X,Y,Z all move. Every time I have seen this the axes have moved in a negative direction.

If you use feed hold then stop it works correctly every time. Only stopping while moving shows the problem.

The attached code triggers the fault every time on my mill. Run it until at least line 700 then hit stop. The exact line number is not critical. Code size seems to have an effect. I tried shortening the code to make testing easier but the fault went away.

I am not sure when this started but it has been around for at least 6 months.
LinuxCNC latest buildbot, RT-Preempt, using Axis. Note using a hardware stop button or the GUI have the same effect.
Both machines use Mesa Ethernet Anything IO cards
[problem.txt](https://github.com/LinuxCNC/linuxcnc/files/3107704/problem.txt)

andypugh:
That sounds bad!
Have you been able to determine if this is a commanded move from LinuxCNC or a problem with the external hardware? 

Big question:  Step/dir or something else? 

Does the position display in the GUI change? And are you displaying commanded or feedback?

sheetcam:
Sorry, I forgot to mention these are both servo systems. This is a commanded move. You can see the axes move in Axis and the run button is disabled until it stops moving.

SebKuzminsky:
My first guess is faulty hardware.  Try running memtest86 overnight on the machine that displays the problem.

sheetcam:
This fault shows up on two different machines with different make computers. The only hardware similarity is that both run Mesa Anything IO, which I doubt are the problem.

andypugh:
Which Mesa cards?

sheetcam:
7I80HD-16 + 7I48

I ran some more tests on the mill with a different file and the length definitely has an effect. At 302k bytes long it works correctly every time and at 306k bytes it fails every time I hit stop. This file was originally much longer and cut without any problems, though I kept well away from the stop while it was actually cutting.

[...]

Issue #393, opened 2018-01-06
Joint Value Jump ON_ABORT_COMMAND


KaushikM19:
Hello all,
Whenever I use ON_ABORT_COMMAND in RS274NGC section, joints motor-pos-cmd value run away on abort.

Linuxcnc: Master 

Configuration file:
[axis_mm.TXT](https://github.com/LinuxCNC/linuxcnc/files/1608690/axis_mm.TXT)

GCode:
[GCODE_FILE.TXT](https://github.com/LinuxCNC/linuxcnc/files/1608692/GCODE_FILE.TXT)

Kaushik

petterreinholdtsen:
Might be related to #241.

Issue #241, opened 2017-03-04
unexpected move after program stop when remap on_abort_command is used


propcoder:
Sorry, not enough time to describe everything in detail.

## Here are the steps I follow to reproduce the issue:
Use standard sim config and add remapping or use some remap demo config
add [RS274NGC]ON_ABORT_COMMAND
run some program, stop while running. Watch the surprise move.
TBD

## This is what I expected to happen:
Normal stop.
## This is what happened instead:
unexpected move after program stop

## Information about my hardware and software:
TBD
 * I am using this Linux distribution and version: Debian Wheezy x86
 * I am using this kernel version: TBD
 * I am running LinuxCNC 2.7.8
   * [+] A binary version from linuxcnc.org (including buildbot.linuxcnc.org)

alexfloca:
hi, reported this a couple of years ago, i have just disabled the  ON_ABORT_COMMAND

KMevada:
Even I got the same issue while using ON_ABORT_COMMAND
It seems that vector chained_points  in emccannon is holding some point and not clearing it out. When we call ON_ABORT_COMMAND this generates LINEAR move msg in flush_segments().

Did u find any solution to it?

KAUSHIk

PR #882, opened 2020-06-08
Fix #865, '"crazy move" when pressing STOP button / ESC'

### LINKED ISSUE #865: "crazy move" when pressing STOP button / ESC
1. load a reasonably complex g-code file (torus.png)
 2. run program
 3. interrupt while doing plunge

Motion is aborted

Feed or rapid move to some pseudo-random location

unsure.

Tested on RPI4 with linuxcnc 2.9 pre 8756d78d648d77c68952d1737f632d2eaea83cae

zultron:
This fixes #865.  It includes a test for the problem, and the top commit explains the problem:

```
LCNC [issue #865][1] found a race condition where an abort during
readahead resulted in unintended motion.  The `emcTaskAbort()`
function appears to be responsible for aborting interp, since it runs
`emcTaskPlanClose()`, but that by itself doesn't flush readahead
segments.

When `emcTaskStateRestore()` restored state from the `3D_Chips.ngc`
test program, executing `G64P.1`, the resulting `SET_MOTION_CONTROL_MODE()`
canon call ran `flush_segments()`, causing the unwanted motion to be
queued.

This patch fixes the problem by calling `emcTaskPlanReset()` alongside
`emcTaskPlanClose()` in `emcTaskAbort()`.  For good measure, it also
adds the call in a couple of other places alongside
`emcTaskPlanClose()` where the purpose is obviously to abort and clear
out interp and all queue.

[1]:  https://github.com/LinuxCNC/linuxcnc/issues/865

```


zultron:
Thanks to @rmu75 for finding and reporting this, plus finding a good fix for it.

Thanks to @andypugh for giving me enough hints to reproduce it:  the `3D_chips.ngc` file is special, since it is very long and contains the needed `G64`, and the "having enuf logs collected" hint was also crucial.

Thanks to @pcw-mesa and @jepler for the bisect.  I don't know how you managed to tickle it reliably enough for a bisect, but you did indeed find the commits that tickled this bug for the first time.

While I'm at it, thanks to @robEllenberg for state tags!

zultron:
Was submitting this PR against the `2.8` branch correct?

zultron:
> Was submitting this PR against the `2.8` branch correct?

I originally thought it's not needed in `2.8` because the original report was triggered by state tags, only in master.  However, #241 and #393 both appear to be the same problem, except triggered by `ON_ABORT_COMMAND`, so perhaps it should even be applied to `2.7`?

andypugh:
Does your new test pass or fail on 2.8 without the fix? 
(I can test myself later, but am at work now)

zultron:
> Does your new test pass or fail on 2.8 without the fix?
> (I can test myself later, but am at work now)

It passes on 2.8 without the fix.  It would probably be easily copied and adapted so that it failed with `ON_ABORT_COMMAND` in 2.8, though, if the theory that #241, #393, #579 are the same problem.

zultron:
> It passes on 2.8 without the fix. It would probably be easily copied and adapted so that it failed with `ON_ABORT_COMMAND` in 2.8, though, if the theory that #241, #393, #579 are the same problem.

Ah ha!  I copy'n'pasted the test and added an `ON_ABORT_COMMAND` to the `.ini` file, and reproduced that problem.  It turns out that problem was indeed fixed by @mozmck's and @robEllenberg's commit rev. b353bd3.

I think this patch is like belt & suspenders on top of that one, taking care of the problem in task's abort, rather than more narrowly within interp.  Should I resubmit the PR against the 2.7 branch?

zultron:
Last bit of info:  My fix doesn't fix the `tests/abort/on_abort_command-crazy-move`, so it's NOT belt & suspenders.

That test is included in this PR as well.  I think we can safely close out those open, related issues.

mozmck:
I would like to see these fixes in 2.7, but that's up to Seb ultimately I think.

zultron:
> I would like to see these fixes in 2.7, but that's up to Seb ultimately I think.

Most importantly, your commit b353bd30 went into 2.7.  The `tests/abort/on_abort_command-crazy-move` test from this PR confirms it (including failing without b353bd30, right on!); this test isn't critical, but there's an argument for putting it into 2.7 for completeness.

The other fix and test are only tickled by state tags.  I believe it's still a bug in the pre-state tags code; it's just there wasn't anything to tickle it before then, and I'm pretty certain nothing new will be introduced to 2.7 that would tickle it in the future.

Then, up to @SebKuzminsky to err on the side of pedantism or on the side of letting sleeping dogs lie.  I'm good either way.

mozmck:
Ah, I see - that makes sense then.  Thanks for all the work!

jepler:
Thank you so much!  Please confer with release managers and find out what other branches we would like this on.  @SebKuzminsky ?

rmu75:
Thanks for fixing this. I didn't really find this problem, andypugh alerted #linuxcnc-devel of a message on the user list, and I just tried to reproduce...
I don't remember/find the original reporter at the moment.

SebKuzminsky:
@zultron Please prepare a PR for 2.7 and i'll make another bugfix release.

PR #883, opened 2020-06-09
debian/control:  Replace missing build-deps

zultron:
This patch adds a few build deps to `debian/control` that are checked for in `configure.ac`, fixing packaging and restoring the ability to install all build deps with

    ./debian/configure uspace && mk-build-deps -i debian/control


zultron:
Actually, I have some more related patches to go with this; I'll file a new PR

PR #885, opened 2020-06-09
Build system tweaks

zultron:
This PR contains a few tweaks to the build system:

- Tweak build console output
- Add mode settings to `Makefile.*.in` for us emacs users
- Remove a redundant `gcc -I../include` arg
- Silence warning from the make macro `cc-option`

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -1107,7 +1107,7 @@ $(sort $(RTOBJS)) : objects/rt%.o : %.c
 	$(ECHO) Compiling realtime $<
 	@rm -f $@
 	@mkdir -p $(dir $@)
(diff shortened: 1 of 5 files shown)


jepler:
I assume that the change to cc-option is to address diagnostics with `make --warn-undefined-variables`, which is laudable.

cc-option was lifted from linux, and they used the optional $(2) for an alternative (so it's like an if/else situation) (however, linux upstream seems to use a newer and more convoluted way to do it last time I looked).  However, it seems we don't use it.  With cxx-option we specify $(2) in one case and not in another.  A fix would be to add a comma so that $(2) is defined to the empty string instead of being undefined.

zultron:
> A fix would be to add a comma so that $(2) is defined to the empty string instead of being undefined.

You're right, that would have been the better fix.

Thanks for the merge!

Issue #128, opened 2016-07-29
tests, packaging: make it possible to package the tests and run them from an installed system


jepler:
This is tricky due to the tests which need to compile and install components in the course of running.  The rest is probably just drudgery.

This would be good to have so that we can build once (to .deb) then install and run tests.  It's particularly good with uspace-plus, because you build once and then run on up to 4 different underlying RTOSes. (that will need a pile of buildbot work which is probably its own issue)

zultron:
I probably have a patch for this for a 2.8-pre branch.  Want a PR?

jepler:
Yes please

zultron:
See #888.  (Very lucky number here in China!)

It turns out I didn't actually have a patch, but I've spent so much time with tests in the past that it wasn't terribly difficult.

To solve the problem where tests must compile and install comps, the tests simply use `sudo halcompile --install` for non-RIP builds.  In Machinekit, I added code to load comps from a user `$MODULE_DIR`, but that wouldn't work for kthreads.

PR #888, opened 2020-06-10
Run tests against system install

zultron:
This PR addresses #128:  It updates the tests to run against a system install (including packages).

The tests aren't actually placed in a separate package, however.  Perhaps this isn't a strict requirement, or if it is, this work has been tested with the `tests/` directory copied out of the rest of the source tree, so it is able to run disembodied from the repo.


zultron:
Note some improvements to the `runtests` script:
- `runtests -p` prints out the last lines of `result` and `stderr` files, helpful for debugging CI failures
- `runtests -s` stops after the first error, helpful to limit spew when something's breaking lots of tests

Also note that tests against installed packages are now enabled in the GH Actions config.

.github/workflows/ci.yml
```diff
@@ -65,8 +65,10 @@ jobs:
         fetch-depth: 0
     - name: Build Debian package
       run: |
+        set -x
         git fetch --recurse-submodules=no https://github.com/linuxcnc/linuxcnc refs/tags/*:refs/tags/*
         ./scripts/travis-install-build-deps.sh
         debuild -uc -us -I -i -j$(nproc)
-        sudo apt install -y lintian
+        sudo apt-get install ../*.deb
```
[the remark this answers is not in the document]
zultron: That's correct.  The way the build system is set up, with the "RIP" build distinct from a system install, it's not possible to run tests from within `debian/rules`.  It would be possible to test uspace threads (but not kthreads) for a system install only with a build system overhaul.

tests/halcompile/personalities_mod/test.sh
```diff
@@ -1,17 +1,11 @@
 #!/bin/sh
-DIR=../../../src/hal/components ;# use in-tree components
-halcompile --personalities=2 --install $DIR/lincurve.comp
-halcompile --personalities=2 --install $DIR/logic.comp
-halcompile --personalities=2 --install $DIR/bitslice.comp
+${SUDO} halcompile --personalities=2 --install lincurve_test.comp
+${SUDO} halcompile --personalities=2 --install logic_test.comp
+${SUDO} halcompile --personalities=2 --install bitslice_test.comp
```
jepler, line 4: This won't work unless the user has passwordless sudo set up, right?

Can we put in an executable `skip` file `sudo -n true > /dev/null` so that in that situation the test is skipped instead of failing or, worse, pausing for user input?

zultron: That is correct.  Great idea.

jepler: Would love to see you add this (the sudo-skips) if you get motivated.  Thanks for the PR!

Issue #704, opened 2020-03-24
G33 uncontrolled movement when direction is reversed


andypugh:
## Here are the steps I follow to reproduce the issue:

 1. Open a simulator, for example sim/axis-mm
 2. MDI M3 S20
 3. MDI G33 K1 Z-10 ; results in controlled spindle-synched motion
 4. MDI G33 K1 Z10 ; results in a rapid move, possibly followed by a small spindle-synched motion
 5. MDI G33 K1 Z -10 ; also now misbehaves

## This is what I expected to happen:

 Two or more synched moves. 

## This is what happened instead:

Rapid moves after the first move. 

## It worked properly before this:

Broken since at least 2.7.15. (My initial thought was that it is related to multispindle, but it seems not) 

The first G33 sets index-enable and behaves as expected. Successive G33 moves do not toggle index-enable. This seems to be by-design. JMKs famous rat-trap fusee demo couldn't have worked if there was a re-synch to index between successive G33 moves. 

However, in MDI mode, with a pause between the moves, the spindle has moved on before the G33 command is issued, and the axis attempts to catch up.

andypugh:
Closed by b3fa2975fde671f10087cb85671e097f35c45d6b

Issue #639, opened 2019-09-03
G33.1 rapid move not synchronized


papagno-source:
## Here are the steps I follow to reproduce the issue:

 1.run in MDI
 2.run m3s10
 3.run g33.1 z-50 k2
 4.when z axis move in negative direction press abort
 5.run m3s10
 6.run g33.1 z-50 k2

The axis z not wait synchronized with encoder , but go in radip move,  for stop in indeterminate position and continued after with reduce velocity for synchronized at encoder for finisch rigid tapping

The problem i have test in master2.8.pre0 and actual master 2.9.pre0

Can replace problem in axis , but have test on real machine is the same.

Very dangerous.

robEllenberg:
I see this too in the stock axis config. Interestingly, doing another motion after the abort (e.g. a rapid back to Z0) makes it behave normally. I suspect a static variable in the rigid tap routine in the TP is not being cleared correctly on abort.

papagno-source:
Yes confirm , if after abort send another motion command type g0 z0 , the g33.1 move correct.
If after abort , you move in manual every axis , the g33.1 have again problem, probably not is reset some variable.Only after motion command this variable is reseted.

andypugh:
Closed by b3fa2975fde671f10087cb85671e097f35c45d6b

PR #889, opened 2020-06-13
G33.1 Fix for #639 & #703

robEllenberg:
Reset emcmotStatus->spindleSync on abort / empty queue
Otherwise successive spindle synched moves do not wait for index, so there
is no reset and a rapid move to catch up ensues.

Signed-off-by: andy pugh <andy@bodgesoc.org>


robEllenberg:
@andypugh I just tried this out in sim, it seems to fix #704 in MDI, when two G33 moves in the same direction causes a sudden rapid move to "catch up". I tested in MDI like this:

`
G0 Z0
M3 S1000
G33 Z-1 K0.01
G33 Z-2 K0.01 (Wait a few seconds then run this command, causes a big rapid down to ~Z-1.4 before resuming spindle-sync)
`

andypugh:
Yes, it was already used to close #704. I was wondering if it has any relevance to this one.

PR #890, opened 2020-06-13
Issue 68 spindle sync blend fixes 2.7

robEllenberg:
This fixes the bad blend behavior in #68, by forcing exact-path mode when changing from a non-synched move (e.g. feed, rapid) to a synched one (threading, rigid-tapping). Note that blends are allowed between the feed / rapid move AFTER a synched move (like in 2.6).


jepler:
The travis run failed with:
> The job exceeded the maximum log length, and has been terminated.
I think this can/should be ignored.

andypugh:
I merged this as part of an attempt to clear up all 2.8 issues.

PR #894, opened 2020-06-28
Do our best to ensure no use of sincos in realtime modules

jepler:
`sincos` is a nightmare, because the compiler loves to convert a call to sin and a call to cos with provably identical arguments into a call to the function `sincos`.  This leads to tragic results when the optimization is applied to a sincos implementation itself.

Change things around so that:
 * Every deliberate use is via a function called `pm_sincos`
 * In realtime modules, disable special treatment of sin/cos calls by telling the compiler they do not have the semantics of the "built in" functions.  (This will also have the effect of de-optimizing any calls that are with constants; for instance, `double d = cos(0)` could have been compiled into the assignment of a compile-time constant to `d`, but we have no reason to think such optimization is important)

I only built this change on a uspace system, however after this change I have no references to the non-qualified `sincos` function in real time modules.  As expected, some remain in userspace programs such as the interpreter.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -811,6 +811,7 @@ EXTRA_CFLAGS += -msse
 EXTRA_CFLAGS += -g
 endif
 endif
+EXTRA_CFLAGS += -fno-builtin-sin -fno-builtin-cos -fno-builtin-sincos
 
 ifeq "$(USE_STUBS)" "1"
 MATHSTUB := rtapi/mathstubs.o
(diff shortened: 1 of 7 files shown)


jepler:
Looks like all architecture tests pass: http://buildbot.linuxcnc.org/buildbot/builders/0000.checkin/builds/6943

jepler:
@SebKuzminsky 2.8 thoughts?

jepler:
@andypugh I removed the offending commit and rebased the branch

andypugh:
I have a feeling that this removes all reference to HAVE_SINCOS and HAVE___SINCOS from the code, so perhaps they should disappear from the configure script too?

SebKuzminsky:
This seems like a good idea for 2.8.

jepler:
The equivalent of this was merged up from 2.8.

PR #896, opened 2020-06-29
Remove sincos() usage (2.8 version)

jepler:
If we're still merging up then #894 can be closed when this is merged.

This also adds the removal of the configure checks for sincos, like andy suggested.


andypugh:
Looks good to me, it Travis agrees.

PR #897, opened 2020-07-01
Fix tests rtapi printf

jepler:
This was broken by #888 on older architectures, since a compiler flag was included unconditionally.  Also, a problem was introduced in the build system.


zultron:
Thanks for catching and fixing that, @jepler!

PR #900, opened 2020-07-08
Arc Radius and State Tags Pins

rodw-au:
All state tags are published to real time motion pins with a number of enhancements.

Extends State tags to include additional values including tool radius, arc radius and a pin that is true when cutting a circle. These are useful for plasma cutting. Publishes all state tags to pins. Refer **man interp**
Also publishes a directional  heading pin which may be useful to command a tangential knife or saw.

Note interp.feed-rate-upm is similar to motion.feed-upm except it does not factor in feed overrides. Some existing applications  already allowed for feed overrides so it was difficult to rewrite them to use the motion.feed-upm pin.

interp.feed-rate-ups is shown in units per second to be compatible with hal velocity pins thus allowing direct comparison between commanded velocity and the original feed rate set in G code which is desirable for plasma

See screen dump for pins published.
arcradius

Many thanks to Andy for his guidance and advice  and Reinhard for his ideas and support.
Andy has been encouraging me about the arc radius for several years and through the magic of State Tags its now possible.


src/emc/motion/control.c
```diff
@@ -2055,6 +2074,60 @@ static void output_to_hal(void)
...
+         Refer interp-internal.hh for values and
+         interp_write.cc for the CANON PLANE definitions
+         axes in order from 0-8 are: x,y,z,a,b,c,u,v,w
+         
+         Do we need to adjust for external offsets here???
+     */
+     hal_float_t x0 = 0;
+	 hal_float_t y0 = 0;
+     switch(emcmotStatus->tag.fields[GM_FIELD_PLANE]){
+		 case 170:	 //XY plane
```
dwrobel, line 2088: Could [G_17](https://github.com/LinuxCNC/linuxcnc/blob/cac12e29525cf822904fd8fd2a731b6653726588/src/emc/rs274ngc/interp_internal.hh#L213) be used instead of cryptic 170? If so then the same would apply for other cases as well.

andypugh: C switches are always integer. 
170 is used to mean G17.0 (etc) throughout the rest of the code, so is used consistently here. 
The comment _could_ say "G17 XY plane" I suppose.

rodw-au: No, the #define can't be used.
We are using a State Tag here which is a structure in motion. That is clear from the code.
G_17 is defined in src/emc/rs274ngc/interp_internal.hh as being internal to the interpreter and thus not available in motion.

When I review this again, I think the confusion is coming from the from the original state_tags code. 
Instead of using G_17 to denote a plane, it should be setting the tag to  PLANE_XY (= 1) which is #defined in canon.hh
Perhaps that should be fixed now before State Tags is used widely. I'm happy to oblige as that should allow an equate to be used in motion.

robEllenberg: Rod, that's a reasonable change to make. For context, state tags started out as a slightly-compressed copy of the modal state arrays in the interpreter. That's why you see a lot of use of those values (e.g. 170 for XY plane). They weren't originally designed to be understood by motion, so there wasn't much need to convert the modal state to a more neutral format.

It would definitely be an improvement if state tags were less interp-specific, and it should be an easy change, as long as all of the code that uses state tags understands the new scheme (e.g. Interp::active_modes() and Interp::restore_from_tag()).

rodw-au: Rob, thanks for that. I'll have a go tonight. The only procedure I've changed where I've included new tags is Interp::write_state_tag so I might need to add a bit more code elsewhere.

I had a chuckle becasue Interp::write_state_tag maps the canon variables to G_17 et al!
`switch(settings->plane) {
    case CANON_PLANE_XY:
	state.fields[GM_FIELD_PLANE] = G_17;
	break;
    case CANON_PLANE_XZ:
	state.fields[GM_FIELD_PLANE] = G_18;
	break;
    case CANON_PLANE_YZ:
	state.fields[GM_FIELD_PLANE] = G_19;
	break;
    case CANON_PLANE_UV:
	state.fields[GM_FIELD_PLANE] = G_17_1;
	break;
    case CANON_PLANE_UW:
	state.fields[GM_FIELD_PLANE] = G_18_1;
	break;
    case CANON_PLANE_VW:
	state.fields[GM_FIELD_PLANE] = G_19_1;
	break;
    }`

Some of the tags I don't fully understand so I don't want to change too much!

dwrobel: @andypugh

> C switches are always integer.

[G_17](https://github.com/LinuxCNC/linuxcnc/blob/cac12e29525cf822904fd8fd2a731b6653726588/src/emc/rs274ngc/interp_internal.hh#L213) is equal to `170`, please check the link I provided?

> 170 is used to mean G17.0 (etc) throughout the rest of the code, so is used consistently here.

Unfortunately not, If you will `grep` on the existing code you would get:
```
$ git grep -- 'state.fields\[GM_FIELD_PLANE' 
src/emc/rs274ngc/interp_write.cc:       state.fields[GM_FIELD_PLANE] = G_17;
src/emc/rs274ngc/interp_write.cc:       state.fields[GM_FIELD_PLANE] = G_18;
src/emc/rs274ngc/interp_write.cc:       state.fields[GM_FIELD_PLANE] = G_19;
src/emc/rs274ngc/interp_write.cc:       state.fields[GM_FIELD_PLANE] = G_17_1;
src/emc/rs274ngc/interp_write.cc:       state.fields[GM_FIELD_PLANE] = G_18_1;
src/emc/rs274ngc/interp_write.cc:       state.fields[GM_FIELD_PLANE] = G_19_1;
```
Which means that if elsewhere `G_17` is assigned to `state.fields[GM_FIELD_PLANE]` then for symmetry 
`switch(emcmotStatus->tag.fields[GM_FIELD_PLANE])` should use `G_17` as well _(not `170`)_. Otherwise, if someone will change the order in the [GCodes enum](https://github.com/LinuxCNC/linuxcnc/blob/cac12e29525cf822904fd8fd2a731b6653726588/src/emc/rs274ngc/interp_internal.hh#L199) and [G_17](https://github.com/LinuxCNC/linuxcnc/blob/cac12e29525cf822904fd8fd2a731b6653726588/src/emc/rs274ngc/interp_internal.hh#L213) will not longer be `170`, then you will be in a trouble.

rodw-au: Yes, agreed. your grep output has found the block I referred to in my earlier post.The G_17 equate is defined private to the rs274ngc folder so its not wise to  use in  motion. 

The solution which has been endorsed by Rob is to store settings->plane in state_tags and modify  interp::active_modes to map the plane to gcode[3] via a similar switch() statement here:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/rs274ngc/rs274ngc_pre.cc#L2174
That will allow the 170 to be replaced with the equate PLANE_XY which is defined in cannon.hh which I think is already loaded in control.c

I am working on these changes. I just want to be sure I don't miss something.

andypugh:
I am trying to clear up some of the forgotten pull requests. Is this still wanted?

andypugh:
OK. I think it still has value, so if you find the time, please try again.

PR #903, opened 2020-07-13
docs: Use pngs, never gifs

jepler:
This works around a problem where imagemagick declines to convert files to pdf; I don't know why a gif uses this conversion path but a png doesn't.
   
Note that this change confuses dependency analysis.  If you built docs before this change, you'll get an error that the (no longer referenced) gif files don't exist.  "make clean" or "git clean" can correct the problem.


andypugh:
Thanks for this, I imagine it was tedious?

jepler:
Thanks for catching that!   I prepared it against 2.8 and expected github to guess that the PR should be against 2.8.

PR #907, opened 2020-07-18
pncconf 7i92/G540 update

cdedwards:
Added dual G540's to the 7i92 device so that pncconf knows about them.


c-morley:
I see you have set the firmware with connectors 2 and 3 - by the manual it should be 1 and 2.
Is there a reason?

cdedwards:
> I see you have set the firmware with connectors 2 and 3 - by the manual it should be 1 and 2.
> Is there a reason?

umm nope. cut and paste from the 5i25 and a quick change to the device name. I can see if I can change it to 1/2 though.

cdedwards:
I'll redo the pull request with the proper input names in it.

PR #908, opened 2020-07-19
Update private_data.py

cdedwards:
Changed definitions of the 7i92 to include G540x2


c-morley:
Thanks for your work!

PR #909, opened 2020-07-20
Add Kalman, linear quadratic estimation (LQE), filter component

dwrobel:

### DIFF
--- a/src/hal/components/filter_kalman.comp
+++ b/src/hal/components/filter_kalman.comp
@@ -1,0 +1,143 @@
+//
+// Copyright (C) 2020  Damian Wrobel <dwrobel@ertelnet.rybnik.pl>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+//
+
+component filter_kalman "Unidimensional Kalman filter, also known as linear quadratic estimation (LQE)";
+license "GPL-2.0-or-later";
+author "Dmian Wrobel <dwrobel@ertelnet.rybnik.pl>";
+
+description
+"""
+Useful for reducing input signal noise (e.g. from the voltage or temperature sensor).
+
+More information can be found at https://en.wikipedia.org/wiki/Kalman_filter.
+
+Adjusting \\fBQr\\fR and \\fBQk\\fR covariances:
+
+Default values of \\fBRk\\fR and \\fBQk\\fR are given for informational purpose only. The nature of the
+filter requires the parameters to be individually computed.
+
+One of the possible and quite practical method (probably far from being the best) of
+estimating the \\fBRk\\fR covariance is to collect the raw data from the sensor by
+either asserting the \\fBdebug\\fR pin or using \\fBhalscope\\fR and then compute the covariance
+using \\fBcov()\\fR function from \\fBOctave\\fR package. Ready to use script can be found at
+https://github.com/dwrobel/TrivialKalmanFilter/blob/master/examples/DS18B20Test/covariance.m.
+
+Adjusting \\fBQk\\fR covariance mostly depends on the required response time of the filter.
+There is a relationship between \\fBQk\\fR and response time of the filter that the lower
+the \\fBQk\\fR covariance is the slower the response of the filter is.
+
+Common practice is also to conservatively set \\fBRk\\fR and \\fBQk\\fR slightly larger then computed
+ones to get robustness.
+""";
+
+pin    in   bit  debug = FALSE             "When asserted, prints out measured and estimated values.";
+pin    in   bit  passthrough = FALSE       "When asserted, copies measured value into estimated value.";
+pin    in   bit  reset = FALSE
+"""When asserted, resets filter to its initial state and returns 0 as an estimated value (\\fBreset\\fR pin
+has higher priority than \\fBpassthrough\\fR pin).""";
+pin    in float     zk                     "Measured value.";
+pin   out float xk_out                     "Estimated value.";
+param  rw float     Rk = 1.17549e-38       "Estimation of the noise covariances (process).";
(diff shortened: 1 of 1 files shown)


dwrobel:
A real example of filtering voltage from the [THCAD-300](http://store.mesanet.com/index.php?route=product/product&product_id=128) sensor. The setup of `Rk` and `Qk` covariances is as follows:

```
$ halcmd show param | grep kalman | grep float
    35  float RW         0.0001  filter-kalman.0.Qk
    35  float RW    0.006856344  filter-kalman.0.Rk
```

The input signal is from the stabilized power-supply and changes between `49V` to `50V` and the time constant of the filter is `~7ms` _(versus THCAD output)_.

filter-kalman-thcad-300-halscope
[Halscope raw data.](https://github.com/LinuxCNC/linuxcnc/files/4947406/filter-kalman-thcad-300-halscope.txt)

rodw-au:
@dwrobel  could you explain  your  hardware for the example plot? and why you tackled this project? I have found some plasma cutters to be quite noisy similar to your plots but the better machines (eg. my Thermal Dynamics and Hypertherm) generate much cleaner signals that probably don't warrant any filtering. Careful attention to designing for noise also helped a lot. I certainly could have used this with my old Everlast!

andypugh:
I will merge this, as I see no reason not to have it. But I wonder if there is scope for a more general filter which can do Kallman, Butterworth, Chebyshev depending on a setting parameter. This would make it easier to experiment on a live system without HAL rewiring.

dwrobel:
> @dwrobel could you explain your hardware for the example plot?

@rodw-au 
>The input signal is from the stabilized power-supply and changes between 49V to 50V

`THCAD` input is connected directly to that laboratory stabilized power-supply _(there is no other load connected to that power supply)_. `THCAD` output is connected to the `7I92` encoder input _(working in `counter-mode` `0`)_ and converted to voltage using `lincurve` as following:
```
loadrt lincurve count=1 personality=3
addf lincurve.0 servo-thread
setp lincurve.0.x-val-00 0
setp lincurve.0.y-val-00 -41.711988649799004
setp lincurve.0.x-val-01 117600
setp lincurve.0.y-val-01 0
setp lincurve.0.x-val-02 963400
setp lincurve.0.y-val-02 300
```
`val-01` and `val-02` comes directly from the THCAD sticker, while the `val-00` is a linear extrapolation based on `val-01` and `val-02`. If you see a voltage change _(49V->50V or 50V->49V)_ on the input signal it means I pressed a button on the power-supply to simulate it.

>Careful attention to designing for noise also helped a lot.

>and why you tackled this project?

As you see it is pure laboratory environment. FYI, even if I short-circuted the `THCAD` input I was still observing a fluctuation on the input frequency/voltage signal.

With the filter added, I'm able to observe a stable voltage on the `plasmac` panel, even with the increased precision of displaying a voltage _([from %3.0 to %3.1](https://github.com/dwrobel/machinekit/commit/2bb3df2a78b9df0a2dd5b173f89edec122234943#diff-453668bad3b9e33a34a8e110dcead6baR827))_.

rodw-au:
Damien, Thanks I am very interested in this. I have spent countless hours studying arc voltage over the last 3 years.  I've not used a THCAD-300 but have used the THCAD-10 and THCAD-5. I used this component I wrote years ago to parse the THCAD 
https://forum.linuxcnc.org/plasma-laser/32585-scalethcad-component 
I've since extended this to include moving averages of a variable number of readings. That helped us clean up a probing signal (ohmic sensing) on a water table so we could get reliable  probing on a water table so your filter might be helpful there.

I'm not aware of anybody using lincurve with the THCAD before. I did a lot of tests with two 48v switch mode power supplies in series to give me 96 volts which was in the expected arc voltage range. Out of the blue the signal was contaminated with noise I had not seen before much like your plots (see Green line)
Live default mm

I also found that there was a 50Hz AC ripple in the signal. 
THCAD volts

I purchased a USB Oscilloscope to explore the source of this noise and concluded that it was some kind of earthing issue. I fitted an EMI filter on my control panel and also a similar filter on the arc voltage in my Everlast plasma cutter itself. I also fully earthed my table to a star ground in accordance with Hypertherm's white paper on grounding. This resolved the noisy signal and I was able to remove the filter in the plasma cutter without affecting the signal. I've never needed to add a grounding rod.

Then I upgraded to a Thermal Dynamics A120 plasma cutter and immediately observed much smoother voltage signals (red line below)

Screenshot at 2020-04-26 09_02_10

This plot shows a dv/dt based on my moving average code (in green) which I used to enable a void hold signal (also shown) when the arc voltage skyrockets when crossing a kerf. You can see the dv/dt is able to detect the kerf crossing almost immediately the voltage starts to rise.

There is so much the Linuxcnc community has to learn about plasma and its exciting to see somebody else getting down to this level. I'm happy to collaborate with like minded people to lock down some of the ideas I have. I have not had a chance to get back to my arc radius PR but that will enable centripetal limiting for improved hole quality and I even have an idea to offset the torch position to account for the wider kerf based on the work in that PR.

rodw-au:
@dwrobel After I reflect on the background I covered earlier, how confident are you that the power supply to the 7i92 was clean? did it have an EMI filter on the AC input? I think its possible that the signal noise is not arriving via the laboratory  power supply signal but by a ground loop (if thats what they call it) and entering via the 7i92 power feed. (Which is what I observed)

dwrobel:
@rodw-au 
>how confident are you that the power supply to the 7i92 was clean?

Both `7i92` and `THCAD` are powered from `USB +5V` socket from the PC.

Please bear in mind that,
>even if I short-circuted the THCAD input I was still observing a fluctuation on the input frequency/voltage signal.

I also measured length of the `3` consecutive impulses _(beetwen rising edges and at `30MHz` sampling rate)_ on the following outputs of `THCAD`:
- **FREF+**:
```
1000.005 ns  999.995kHz
1000.000 ns 1000.000kHz
 999.999 ns 1000.001kHz
```

- **FOUT+**:
```
8333.647 ns 119.995476468 kHz 
7733.221 ns 129.312216873 kHz
8000.040 ns 124.999381271 kHz
```

Based on above, whatever the quality of the `USB +5V` is - it does not have any negative impact on the **FREF+** output and with short-circuited `THCAD` input the same should be observed on **FOUT+**, but it is not. That's why I came with the conclusion that the source of the noise is the **FOUT+** output and it would be good to try to minimize it.

rodw-au:
I think its great to have another filter and your method certainly appears effective. Lowpass is often used (I think Plasmac includes it) and I think there is also a place for the moving average as its been used in university papers. Its just the noise does not look right to me based on my experience. When I get a chance, I'll include some fresh plots from my machine. (7i76e, thcad-10 and thcad-5).

Also what was the THCAD hardware divider set on? 1/32 is recommended for Linuxcnc. I wonder if that makes a difference?
It would be interesting to get some feedback from Peter Wallace.

PR #918, opened 2020-08-04
Rootless preempt-rt mode (Work in Progress)

dwrobel:
Allows to run linuxcnc in `preempt-rt` mode without root _(setuid bit)_.

Benefits:
 - allows to always run using `SCHED_FIFO` policy _(on both rt and non-rt kernels)_,
 - `CI` flow can use and tests exactly the same code flow as will be used on the production,
 - `latency-test` _(on modern hardware)_ should return much more reliable results on non-rt kernel,
 - will help packaging `linuxcnc` in Fedora _(see: [setuid removal](https://fedoraproject.org/wiki/Features/RemoveSETUID))_ and maybe other distros where `setuid` binaries are not welcome.

How to run it:
- In order to run it in place, just set the following capabilities:
```
$ sudo setcap cap_ipc_lock,cap_net_admin,cap_sys_rawio,cap_sys_nice+ep  bin/rtapi_app
```
- If you're going to use parallel port, just add yourself to the appropriate group _(as you would do for accessing printer)_.

TODO list:
  - [ ] add something like `make setcap` target to set capabilities on `rtapi_app`,
  - [ ] decide how to clean up the API: like rtapi_is_realtime(),
  - [ ] add an option to not invoke `iptables` in hm2_eth.c _(it's much easier and more reliable to configure the rules using firewalld/NetworkManager e.g. on Fedora >=32 firewalld switched from iptables to [nftables backend](https://firewalld.org/2018/07/nftables-backend))_,
  - [ ] test with pci cards _(currently tested with parport and ethernet based mesa card)_,
  - [ ] very likely other things were missed - would appreciate to get a feedback.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,8 +29,12 @@ jobs:
         ./configure --with-realtime=uspace --disable-check-runtime-deps
         make -O -j$((1+$(nproc))) default pycheck V=1
         # Note that the package build covers html docs
-        ../scripts/rip-environment runtests -p
-    
+        sudo setcap cap_ipc_lock,cap_net_admin,cap_sys_rawio,cap_sys_nice+ep ../bin/rtapi_app
+        timeout --signal=9 3600 ../scripts/rip-environment runtests -vp || \
+          ([ -e ~/linuxcnc_debug.txt ] && (echo linuxcnc_debug.txt; cat ~/linuxcnc_debug.txt); \
+           [ -e ~/linuxcnc_print.txt ] && (echo linuxcnc_print.txt; cat ~/linuxcnc_print.txt); \
+           false)
+
   htmldocs:
     runs-on: ubuntu-16.04
     steps:
--- a/src/hal/drivers/mesa-hostmot2/hm2_eth.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_eth.c
@@ -368,7 +368,7 @@ static bool chain_exists() {
     return result == EXIT_SUCCESS;
 }
 
-static int iptables_state = -1;
+static int iptables_state = 0;
 static bool use_iptables() {
     if(iptables_state == -1) {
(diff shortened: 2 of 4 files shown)


jepler:
Things to make sure of as you work on this:
 * Top priority: Identify parts of this which can be separated into pull requests which can be incorporated one by one, so that you have to carry less "outstanding work" as the process continues and you can monitor community acceptance of the coding choices you're making instead of getting the real feedback at the end.
  For instance, these immediately strike me as separate PRs that can be evaluated on their own merits:
    * re-working how hm2_eth does firewalls
    * re-working how harden_rt can work without root but with capabilities (if you can keep as-root and as-capabilities intact, or as a configure-time flag)
 * It must remain possible to run LinuxCNC in a pure simulator mode with no elevated privileges required to build it, run the tests, run UIs in simulator mode, etc.
 * It must work on whatever range of platforms we will commit to supporting in our main development branch
 * It must be furnished with the debian packaging work completed as well.  I am not familiar with how capabilities are represented in debian packages, so you may have to educate yourself.
 * It must not regress on preventing non-rtapi_app processes from sending packets on the ethernet interface while LinuxCNC is running.

I don't know what's the point of protecting a system against a setuid binary but allowing a binary with cap_sys_rawio, but linux security decisions are a land of contrasts.

andypugh:
It's been quite a while, but this might still be a good idea. Any comments?

petterreinholdtsen:
It sound like a great idea to reduce the privileges needed to run linuxcnc.  Any hope to get this patch out of draft / work in progress mode?

andypugh:
Is this still a live project? It seems that this work would be good to introduce, but it has been "draft" for 6 years. 

If this is still something that you are interested in then we can leave it in the PR list, but otherwise it is just going to sit here looking untidy.

BsAtHome:
Fixed with #3964.

PR #919, opened 2020-08-11
Andypugh/spanish docs

andypugh:
This took a fair bit of fiddling to put the traslated manpages and README files where I think that they should be.


andypugh:
Various errors in the docs prevent building. Closing this.

PR #926, opened 2020-08-17
Documentation fixes

havardAasen:
I readded the English documents in docs/ folder. I hope the English version can be the standard and translated is optional. The second commit is supposed to fix some warnings asciidoc emits during build.


andypugh:
Thanks for these, I had missed them. 
I don't think the Spanish translator intended the work to be merged as-is, so he seems to have translated in-place probably planning to re-arrange later.

PR #930, opened 2020-08-24
Create linuxcnc-documentation-es.desktop

phillc54:
Probably requires a Spanish name and comment


andypugh:
I wondered what the problem was, this file has existed on my machine for some time but was lost amongst the untracked files.

PR #931, opened 2020-08-25
Fixes for man-page

havardAasen:
This fixes most of the lintian output for man-pages. There is one I didn't managed to fix
`W: linuxcnc-uspace: manpage-section-mismatch usr/share/man/man1/thermistor.1.gz:1 1 != 9`
I played with `src/hal/user_comps/Submakefile` with the only result in duplicating the file. One correct the other wrong.

I believe there are still some things to look into.
Some man-pages in `docs/man/es` is still in English. I can delete those that are not translated yet.

There is also man-pages with section 'Reporting Bugs' these have various sites ore email listed, most is regarding sourceforge. I can go through and remove those sections so we at least don't miss inform people.

There was also mentioned in the mailing list about the warnings a2x created during build. Are those warnings generated when a2x is converting docs/src/man/*/*.txt into man-page? If that is the case, can one solution be to replace those *.txt files with man-page in  troff/groff style?


andypugh:
Thanks. 
The docs that have an _es suffix but are still in English are there to avoid broken links in the Spanish docs.

PR #932, opened 2020-08-25
Change Standards-Version depending on distribution

havardAasen:
The change in debian/changelog is because of a warning, replicated with this command
`dpkg-parsechangelog --all | head`

I updated the default Standards-Version to 3.9.3 which is what lintian in Ubuntu Precise claims is correct, This is also the lowest version we should hit with LinuxCNC 2.8.
This site [http://wiki.linuxcnc.org/cgi-bin/wiki.pl?MinimumSoftwareVersions](http://wiki.linuxcnc.org/cgi-bin/wiki.pl?MinimumSoftwareVersions) does mention which Standards-Version the various distribution supports, though it differs from what lintian says, I have made the changes based on output from lintian.

This change is also mostly targeted the Buildbot, which is why Ubuntu 14-16 and LinuxMint 17-18 is still at default version (3.9.3). I also don't know which version they actually prefer.

I did some cleaning, removed some duplicated entries and swapped tabs witch spaces. I removed Debian-testing since Python2 is almost completely removed. We can perhaps re add it as a separate entry with a warning that it is not supported and exit the script.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -745,7 +745,7 @@ linuxcnc (1:2.7.5) unstable; urgency=medium
   * build: make failure copying images an error
   * packaging: interface with udev better
 
- -- Sebastian Kuzminsky <seb@highlab.com>  Tue, 12 July 2016 21:47:18 -0600
+ -- Sebastian Kuzminsky <seb@highlab.com>  Tue, 12 Jul 2016 21:47:18 -0600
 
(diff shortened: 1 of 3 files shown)


andypugh:
Thanks

havardAasen:
I must admit I did not check that.

PR #934, opened 2020-08-25
QTvcp python 3

satiowadahc:
- Allows QTVCP to run without need of python 2
- Python styling to PEP8


rene-dev:
Thanks!
is there anything that doesnt work? did you test all the UIs?

satiowadahc:
If I'm going to be honest I've only tested my own. I plan on converting a machine over to focal fossa and trying again. But I'm being pulled to another project already so if anyone else has a chance to test or critique, help is much appreciated

satiowadahc:
*Realization at 2am, need to go back and modify messages saying to install python3-pyqt5.* instead of python-pyqt5.*

c-morley:
No problem - there are a couple things I need to adjust in your code and I'll add those changes.
Thanks for the heads up!

c-morley:
I see you removed lib/python/qt5_graphics from git ignore. I am wondering why? it is a generated file from src/emc/usr_inf/gremlin/qt5_graphics.

satiowadahc:
I must have missed the comment that its auto generated. If I recall right (Don't have my work computer handy)
```super(Lcnc_3dGraphics, self).__init__(parent=parent, s=linuxcnc.stat(), lp=None)```
needed the s and lp in order for gcode graphics to load. Having said that I'm now wondering if gcodegraphics needs to be changed instead of this line

c-morley:
There is a new branch to try: qtvcp_py3
There is still some trouble with probing; the subpocess complains about importing linuxcnc - seems to be the wrong version

satiowadahc:
```File "/usr/lib/python2.7/dist-packages/qtvcp/widgets/nurbs_editor.py", line 91```  I see this is pointing to python 2.7 in your branch

rene-dev:
> the subpocess complains about importing linuxcnc - seems to be the wrong version

this means its not running the same python version linuxcnc was compiled for.

c-morley:
> > the subpocess complains about importing linuxcnc - seems to be the wrong version
> 
> this means its not running the same python version linuxcnc was compiled for.
Here is the error - it's a little awkward to debug as it's a subprocess invoked by a widget.
Does this ring a bell? I haven't started research yet.

Error parsing return data from sub_processor. Line=b'Traceback (most recent call last):\n
  File "/home/chris/emc/lib/python/qtvcp/widgets/basic_probe_subprog.py", line 21, in <module>\n    from qtvcp.core import Action\n  
File "/home/chris/emc/lib/python/qtvcp/core.py", line 4, in <module>\n    
import linuxcnc\nImportError: /home/chris/emc/lib/python/linuxcnc.so: undefined symbol: PyUnicode_FromString\n' (basic_probe.py:213)

c-morley:
> `File "/usr/lib/python2.7/dist-packages/qtvcp/widgets/nurbs_editor.py", line 91` I see this is pointing to python 2.7 in your branch

can you explain more - surely it shouldn't be in the python2.7 folder - I use RIP so this error has not shown it's self to me.

c-morley:
pushed more work.
python version error fixed - I forgot to change the subprocess loading command to python3

c-morley:
I'm closing this pull request. I regret not being  able to use your work directly, but it helped immensely in the process.
In the future if you could make things like PEP8 into separate commits as much as possible that would help me inspect the code.
Thank you for your contribution.

PR #935, opened 2020-08-28
docs: Use pngs, never gifs

htasta:
This works around a problem where imagemagick declines to convert
files to pdf; I don't know why a gif uses this conversion path but
a png doesn't.

Note that this change confuses dependency analysis.  If you built docs
before this change, you'll get an error that the (no longer referenced)
gif files don't exist.  "make clean" or "git clean" can correct the
problem.


htasta:
this fixes the build errors on 2.7

PR #941, opened 2020-09-23
Replace *get_cpuinfo_revision(char *revision)

elovalvo:
As reported in  https://www.raspberrypi.org/documentation/hardware/raspberrypi/revision-codes/README.md

"As of the 4.9 kernel, all Pis report BCM2835, even those with BCM2836, BCM2837 and BCM2711 processors. You should not use this string to detect the processor. Decode the revision code using the information below, or cat /sys/firmware/devicetree/base/model"

So the get_cpuinfo_revision() function is no longer correct and above all does not work with RPi4.

Therefore, I think it is more appropriate to completely replace this function, which is now obsolete, with a simpler one that can also adapt to future models of the Raspberry.
Also it's possible to delete cpuinfo.h

### DIFF
--- a/src/hal/drivers/cpuinfo.c
+++ b/src/hal/drivers/cpuinfo.c
@@ -22,13 +22,12 @@ SOFTWARE.
 
 #include <stdio.h>
 #include <string.h>
-#include "cpuinfo.h"
 
 char *get_cpuinfo_revision(char *revision)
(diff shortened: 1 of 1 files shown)


andypugh:
Thanks. 
(Doing it this way gets your name in the commit history)

PR #943, opened 2020-09-30
Gtk3 port

rene-dev:
This branch does the following:

- port gladevcp to Gtk3
- WIP on all new gremlin, based on a openGL core profile(gtk3 doesnt allow legacy profiles)
- Not compatible with python2 or gtk2 anymore.
- change default python to python3
- fix dependencies and packaging for gtk3 and python3 on focal, mint20 and buster
- anything older than focal/buster doesn't fully work due to bugs in Gtk, and incompatibility of older glade with python3.

Currently mostly working:
- Gmoccapy
- Touchy
- basic gcode rendering in new gremlin(no backplot, no tool, no buttons)

I'm keeping track of bugs/TODOs using the gtk3 label:
https://github.com/LinuxCNC/linuxcnc/issues?q=is%3Aissue+is%3Aopen+label%3Agtk3

I'm encouraging everybody to test this, and report any bugs.


satiowadahc:
I've been using this as the base for most of my testing with python3. One small annoyance that I can't seem to find the source of. 
From Scratch:
```bash
cd debian
./configure uspace
cd ..
dpkg-checkbuilddeps
```
There is an extra new line, if the OS is not recognized, before `Section: libs`

andypugh:
I think that maybe we should try to reach an agreement that we should allow breaking existing configs to get free of Py2 and GTK2. 
Maybe even an actual IRC meeting (like we did, briefly, several years ago.). 
Or just a proposal on the dev mailing list. (and if no-one objects constructively then take it as accepted)

c-morley:
I don't think anyone really thinks we can/should keep supporting py2 in the future.
The real questions are how/when will we transition - as the transition will be very disruptive.
Will we do a final py2 release first?
Will we wait until we have every program converted to py3?
How will we warning/inform users (in production/semi-production) of master that we will break everything.
Keeping in mind if a user is using master in older distributions, it may require the user to upgrade to a new distribution to have python3 support.
We should start warning people soon.
We need to discuss the plan, the decision has already been made for us by the python/distribution developers.

andypugh:
> We do need to bite the bullet and do it at some stage, I have no objections. It may be worth considering making the version number 3 to make it more distinguishable.

My opinion is that we should keep the 2, as a reminder that we are EMC2. So the next versions will be 2.9, 2.10. 2.11....
I don't think that going back to EMC2 at this point would be helpful, there was enough user-confusion when we changed last time.

andypugh:
> I don't think anyone really thinks we can/should keep supporting py2 in the future.
> The real questions are how/when will we transition - as the transition will be very disruptive.
> Will we do a final py2 release first?

I  think that 2.8,nnn should be the last Py2 version. We should announce the imminent withdrawal of Py2 in Master on the mailing list and forum, and then just get on with it. 

But I think that we should have this discussion on the developers list too. 
> Will we wait until we have every program converted to py3?
> How will we warning/inform users (in production/semi-production) of master that we will break everything.
> Keeping in mind if a user is using master in older distributions, it may require the user to upgrade to a new distribution to have python3 support.
> We should start warning people soon.
> We need to discuss the plan, the decision has already been made for us by the python/distribution developers.

andypugh:
See also cf8ebfe3a5f232b499a85d3fee67c9ee29f4abbd

rene-dev:
one final release before this is merged is probably the best way to go. a good time is soon, but it does need more testing by other users and on a real machine. touchy is close to working, gmoccapy is a little more work. axis fully working, QT should not be a problem.

andypugh:
> See also [cf8ebfe](https://github.com/LinuxCNC/linuxcnc/commit/cf8ebfe3a5f232b499a85d3fee67c9ee29f4abbd)

This means that buildbot won't be building Master on old distributions. (I think). Which will limit the exposure of legacy systems running buildbot releases to any changes in Master.

saxa:
I think GTK3 is nearly a must, or better since now there is also a GTK 4 maybe would be worth start working on that directly.

rene-dev:
superseded by https://github.com/LinuxCNC/linuxcnc/pull/1164

PR #957, opened 2020-10-10
hal: Remove component type magic numbers

ChrisNisbet01:
Please consider these changes for inclusion in the project. 
I was browsing the code and saw these mysterious magic numbers used for the component type, so I figured it might be better to replace them with something more descriptive.

Replaced the component type magic numbers with an enumerated type.
The new types have the same values as before.
i.e.
0 == user
1 == realtime
2 == something else (not sure where this value is ever assigned)

There should be no functional change.

Please note that although I've checked that the code still builds, I'm not at all familiar with how to run/test the code to ensure it's all still OK.

### DIFF
--- a/src/hal/utils/halcmd_commands.c
+++ b/src/hal/utils/halcmd_commands.c
@@ -1272,7 +1272,7 @@ int do_unloadusr_cmd(char *mod_name)
     next = hal_data->comp_list_ptr;
     while (next != 0) {
 	comp = SHMPTR(next);
-	if ( comp->type == 0 && comp->pid != ourpid) {
+	if ( comp->type == COMPONENT_TYPE_USER && comp->pid != ourpid) {
 	    /* found a userspace component besides us */
 	    if ( all || ( strcmp(mod_name, comp->name) == 0 )) {
 		/* we want to unload this component, send it SIGTERM */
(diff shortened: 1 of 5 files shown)


src/hal/utils/halcmd_commands.c
```diff
@@ -1640,16 +1640,16 @@ static void print_comp_info(char **patterns)
...
-            if(comp->type == 2) {
+            if(comp->type == COMPONENT_TYPE_OTHER) {
                 hal_comp_t *comp1 = halpr_find_comp_by_id(comp->comp_id & 0xffff);
                 halcmd_output("    INST %s %s",
                         comp1 ? comp1->name : "(unknown)", 
                         comp->name);
             } else {
                 halcmd_output(" %5d  %-4s  %-*s",
-                    comp->comp_id, (comp->type ? "RT" : "User"),
+                    comp->comp_id, (comp->type == COMPONENT_TYPE_REALTIME) ? "RT" : "User",
```
rene-dev, line 1650: there is a functional change here, old code checks > 0 new code == 1

rene-dev: but it does the correct thing, so I do approve this PR, lets wait for the tests to run.

ChrisNisbet01: Well spotted. I did wonder which would be best, and decided that this was probably what was desired.

PR #958, opened 2020-10-10
halcmd: fix "tune" command locking.

ChrisNisbet01:
Please consider these changes for inclusion in your project.

The "tune" command wasn't actually applying the desired locks
because it was locking a bitwise 'and' of values 1 & 2, which
gives a value of 0. This should have been a bitwise 'or' of the
values.
Also changed the associated 'unlock' functions so that the "tune"
command unlocks the same locks as it locked.
Created a new #define to represent the locks to locks for the "tune"
command to help make it clearer that that the lock and unlock
functions should alter the same locks.

Note to maintainers - I figured that the lock/unlock functions are always called in pairs for each command, so if the lock function was called for the "none" command, the unlock function would also be called for the "none" command. That is why you'll see the "none" command get a mention in the unlock functions whereas it wasn't before. If I figured wrong, sorry.

You'll also see some formatting changes in the affected code. I hope that's not a big problem.

As mentioned in my other PR, I have checked that the code compiles, but have not tested it (totally new to this code).

### DIFF
--- a/src/hal/hal.h
+++ b/src/hal/hal.h
@@ -147,6 +147,9 @@ RTAPI_BEGIN_DECLS
 #define HAL_LOCK_PARAMS   4     /* locking of parameter set commands */
 #define HAL_LOCK_RUN      8     /* locking of start/stop of HAL threads */
 
+/* locks required for the 'tune' command */
+#define HAL_LOCK_TUNE (HAL_LOCK_LOAD | HAL_LOCK_CONFIG)
+
 #define HAL_LOCK_ALL      255   /* locks every action */
 
 /***********************************************************************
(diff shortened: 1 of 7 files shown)


rene-dev:
to me this looks like a @jepler question...

jepler (approved):
Seems fine but may I suggest writing a test for the correct behavior?  

No testing performed.

ChrisNisbet01:
Meh, my rushed attempt at a test wasn't much good. Probsbly not that surprising given it's not something I'm familiar with on this repo. It might be better for somebody else to deal with it.

jepler:
Thank you!

PR #964, opened 2020-10-20
changed pocket_number to mod_pocket in direction logic

inoxix:
Tested and works now with a turret tool changer where tool number is grater than the effective tool positions. PR should go to 2.8 branch and master.


rene-dev:
Why is this change required?

inoxix:
It resolves a bug in component.

andypugh:
Thanks

PR #968, opened 2020-10-26
Add keep-integrator bit to PID

0wn3r:
Not sure if someone may need this but I found a case when it may be wanted to save the integrator value. My case is dual loop feedback with linear and rotary encoder, with an imperfect ballscrew. Ballscrew has an error of 0.2 mm on 750 mm of travel and when PID is disabled (machine OFF) this integrated value is lost. After disabling machine the only way to start work again is to close LinuxCNC and rehome.
integrator_save
Part of .hal file responsible for this offset:
net x-enable        =>  pid.xscale.enable
net x-scale-pos-fb        =>  pid.xscale.feedback
net x-pos-cmd       =>  pid.xscale.command xscale_sum2.in0
net x-scale-output      pid.xscale.output   => xscale_sum2.in1

net x-pos-corrected xscale_sum2.out       =>  pid.x.command
Thanks to DaBit for this use case!

### DIFF
--- a/docs/man/man9/pid.9
+++ b/docs/man/man9/pid.9
@@ -91,6 +91,9 @@ The difference between command and feedback.
 When true, enables the PID calculations.  When false, \fBoutput\fR is zero,
 and all internal integrators, etc, are reset.
 .TP
+\fBpid.\fIN\fB.keep-integrator\fR bit in
+When true, internal integrator is not reset when \fBenable\fR is false.
+.TP
 \fBpid.\fIN\fB.index\-enable\fR bit in
 On the falling edge of \fBindex\-enable\fR, pid does not update the
 internal command derivative estimate.  On systems which use the encoder
--- a/src/hal/components/pid.c
+++ b/src/hal/components/pid.c
@@ -41,7 +41,8 @@
     'command' minus 'feedback'.  'enable' is a bit that enables
     the loop.  If 'enable' is false, all integrators are reset,
     and the output is forced to zero.  If 'enable' is true, the
-    loop operates normally.
+    loop operates normally. 'keep-integrator' can be used to
+    save the integrator value when 'enable' is false.
 
     The PID gains, limits, and other 'tunable' features of the
     loop are implemented as parameters.  These are as follows:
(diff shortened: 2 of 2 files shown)


andypugh:
So, you have a PID which has position as an output used as an offset to the commanded position, and this leads to an f-error?
I can see how that could happen in such a situation. 

I don't see any reason (other than increased complexity and shared memory usage) to reject this idea, bit feel it is worth a bit of discussion around the reported problem. 

This feels like an unusual arrangement. More typical would be to use the measured position as the position feedback to a position PID which (optionally) controls a velocity PID loop closed by the motor encoder. 
When the machine does a machine off -> machine on transition the commanded position is set to the feedback position. In that situation you would want the offset to be zero so the existing behaviour seems like it should be correct, ie: Consider a 1m axis with a 2mm error. 
. Machine startup, X = 0, offset at 0 = 0, offset at +1000 = 2.
. Stop the machine at X = 500. New measured = 500, new actual = 500, offset = 0. Offset at -500 = -1, offset at +500 = +1. 

Which is not to say that there are not applications where you would want to freeze / save the integrator. 

I don't think that I support merging this in to 2.8, though. As a new feature rather than a bugfix I would prefer to see it go in to master.

0wn3r:
Yes, I'm getting the f-error in this transition. 
I understand that I should scales PID as the velocity setpoint for the servo drive, but reason behind current scheme is that rotary encoder has resolution 2 to 4 times higher per mm and I don't want to lose it. Also scales feedback is not that clean on this machine (due to some mechanical issues) and this may cause instability with higher P term which is needed to overcome f-error.
I hope that I have saved some images with f-error plots on rotary and linear encoder. I will attach them to show this situation.

0wn3r:
Found plots:
x-pid-tuned
y axis work1
That's why I'm not sure that using scales as only position PID is a good idea.

andypugh:
As I said, I can see that there might be a use for this change, but I would prefer to see it as a PR to Master than to 2,8.

0wn3r:
Scales are 1um, looks like I am overthinking on this?

rene-dev:
yes. 1µ looks good. can your machine actually position this accurate?
the motor encoder needs more resolution, to get a good velocity estimation, because velocity estimation on encoders is not trivial and can be noisy.
the position controller should be able to position the machine within few counts, providing the machine can do that.

andypugh:
We are wandering off-topic. 

Regardless of whether this scenario is a good use of such a pin, do you see a reason not to add the pin for a potential future application which might require it?

rene-dev:
I dont have a problem adding it, I just dont see a use case for it. Like I said, I dont think this is one, and doing it differently would actually get better results.

0wn3r:
I think if use case is ever to come up (mine looks to be not really correct), this can be brought back again.

jepler:
Closing based on the discussion. Please feel free to open a fresh PR in the future, if you wish to return to the topic.  Thank you!

Issue #979, opened 2020-11-17
QTVCP Deprecation Warnings


satiowadahc:
Here are the steps I follow to reproduce the issue:

 1. Using our own custom UI, based off qtvcp with some pyqtvcp
 2. Running CI testing, which tests 90% of our functionality
 3. Currently off Rene-Devs branch of limuxcmc...  Will try again momentarily off master

## This is what I expected to happen:
```bash
../../../../usr/lib/python3/dist-packages/qtvcp/core.py:118
  /usr/lib/python3/dist-packages/qtvcp/core.py:118: PyGIDeprecationWarning: GObject.SIGNAL_RUN_FIRST is deprecated; use GObject.SignalFlags.RUN_FIRST instead
    'toolfile-stale': (GObject.SIGNAL_RUN_FIRST, GObject.TYPE_NONE, (GObject.TYPE_PYOBJECT,)),

../../../../usr/lib/python3/dist-packages/qtvcp/core.py:142
  /usr/lib/python3/dist-packages/qtvcp/core.py:142: PyGIDeprecationWarning: Since version 3.11, calling threads_init is no longer needed. See: https://wiki.gnome.org/PyGObject/Threading
    GObject.threads_init()

../../../../usr/lib/python3/dist-packages/qtvcp/core.py:143
  /usr/lib/python3/dist-packages/qtvcp/core.py:143: PyGIDeprecationWarning: GObject.timeout_add is deprecated; use GLib.timeout_add instead
    GObject.timeout_add(100, self.update)

/usr/lib/python3/dist-packages/qtvcp/widgets/origin_offsetview.py:191: DeprecationWarning: This method will be removed in a future version of Python. Use 'locale.format_string()' instead.
    self.tabledata[row][column] = locale.format(tmpl, i[column])

/usr/lib/python3/dist-packages/qtvcp/widgets/origin_offsetview.py:187: DeprecationWarning: This method will be removed in a future version of Python. Use 'locale.format_string()' instead.
    self.tabledata[row][column] = locale.format(degree_tmpl, rot)

 /usr/lib/python3/dist-packages/qtvcp/qt_action.py:209: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
    self.cmd.program_open(str(fname))


```

## Information about my hardware and software:

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`):
 Ubuntu 20.04, Replicable on Linux Mint and Lubuntu
[...]

PR #983, opened 2020-11-19
QTVCP warnings cleanup

satiowadahc:
- Cleans up Py_ssize_t implicit type conversion warnings
- Removes literal comparison with is
- Removes repeated calls to Python.h

- removes pep8 from previous pr
Issue #979


c-morley:
Thanks for new pull request!
I checked on my system for python2 and it seems to work other then predictably I needed to uncomment the set_timer() function in core.py.
Without that function the GObject messages aren't sent to the widgets.
Is that a deal breaker for python3?

satiowadahc:
Yep. Threads init is no longer in gobject. So we could just stick a version check there?

c-morley:
ok well that fixes it in python2 but in python3 it won't update AFAICT - self.update() will never be called.

c-morley:
Somehow this breaks DRO display in AXIS and Qtvcp on my buster python3 system

c-morley:
removing commit 3b95bb2 fixes AXIS but not qtvcp

satiowadahc:
So I'll spend more time in axis later on. It seems courier font is breaking DRO. out of curiosity why does glcanon use miniGL but qt uses OpenGL?

c-morley:
I don't see update_start() What branch are you basing from?
The reason we override the function set_timer() in core.py is so the GObject thread is initialized after QT thread (AFAIU).
Otherwise we get a segmentation fault.

AXIS used miniGL originally.
QTVCP use openGL as that is the common library used with pyqt and I was working from pyqt examples to port gremlin over.

satiowadahc:
https://github.com/satiowadahc/linuxcnc/blob/c66379494e4400546798318256162f99a0e3997f/lib/python/hal_glib.py#L242

Not sure if I mentioned it, but it seems specifying the font in axis, broke the DRO.

c-morley:
update_start is for HAL pin update and is unral\ted to QTVCP. (it uses it's own class)
set_timer is used for linuxcnc status messages. gladeVCP and qtvcp both use it but need the threads initialized at different times.

c-morley:
update_start() is for GPin class, not the _GStat class.

c-morley:
Sorry i was so late getting back to this - I see you fixed AXIS DRO but qtvcp's is still broken.
I'll see if I can find time to figure out why - surely related to the AXIS fix

satiowadahc:
I finally can get back to this! gonna start figuring out if they just changed the names of fonts somewhere

satiowadahc:
So @rene-dev actually figured it out. Switched to openGL instead of miniGL on glNav.py. Seems to work now on both for me on python3

c-morley:
I get this error when running qtvcp:

  File "/home/chris/emc-dev/lib/python/qt5_graphics.py", line 666, in initializeGL
    self.realize()
  File "/home/chris/emc-dev/lib/python/qt5_graphics.py", line 476, in realize
    self.font_base, width, linespace = glnav.use_pango_font('courier bold 16', 0, 128)
  File "/home/chris/emc-dev/lib/python/glnav.py", line 142, in use_pango_font
    if w and h: glDrawPixels(w, h, GL_LUMINANCE, GL_UNSIGNED_BYTE, a)
  File "/usr/lib/python2.7/dist-packages/OpenGL/latebind.py", line 45, in __call__
    return self._finalCall( *args, **named )
  File "/usr/lib/python2.7/dist-packages/OpenGL/wrapper.py", line 782, in wrapperCall
    pyArgs = tuple( calculate_pyArgs( args ))
  File "/usr/lib/python2.7/dist-packages/OpenGL/wrapper.py", line 355, in calculate_pyArgs
    yield converter(args[index], self, args)
  File "/usr/lib/python2.7/dist-packages/OpenGL/GL/images.py", line 400, in __call__
    return arrayType.asArray( arg )
  File "/usr/lib/python2.7/dist-packages/OpenGL/arrays/arraydatatype.py", line 141, in asArray
    return cls.getHandler(value).asArray( value, typeCode or cls.typeConstant )
  File "/usr/lib/python2.7/dist-packages/OpenGL/arrays/arraydatatype.py", line 52, in __call__
    typ, repr(value)[:50]
TypeError: ("No array-type handler for type <type 'array.array'> (value: array('b', [-64, -128, -128, -128, -128, -128, -12) registered", <OpenGL.GL.images.ImageInputConverter object at 0x7f4a50b57dd0>)

Using qt 5.5.1 which might ne a problem.

satiowadahc:
```bash
TypeError: ("No array-type handler for type <type 'array.array'> (value: array('b', [-64, -128, -128, -128, -128, -128, -12) registered", 
```
I might be miscounting, but I think there is a `]` missing after the -12

satiowadahc:
Closing this as the pressing issues were resolved outside of it.

PR #987, opened 2020-11-21
Revert bad change inside driver XHC-WHB04B-6 lcnc 2.8

alkabal:
Hi

I don't know why the user jasenk2 have done this change but this broke all the work done for use :

-s
Enable spindle override value using wheel with feed button to Lead mode.

-f
Enable feedrate override value using wheel after Lead mode or at first start (display MPG mode)

Best regards


andypugh:
I asked on the mailing list of anyone had an XHC pendant to check this with, but there was no response. 

There are two very pull requests with similar titles. If this one was committed to 2.8 and then merged to master, would #986 be redundant?

alkabal:
Yes you are right i have done this by mystake

For lead wit WHB06-6 i really get 1C as you can see here, now the driver never get the lead mode
Capture

alkabal:
We need to have a log from user with WHB04B-4

for debug component started with : 
loadusr -W xhc-whb04b-6 -HsfBUn

If  xhc-whb04b-4 user need 9B, fr now i don't know how to manage this !

itaib:
I have a 4 axis xhc-whb04b. what needs to be done?

alkabal:
Unfortunately this is not the good unit, but many thanks for proposition !

alkabal:
Is someone have any suggest for make both value work ?

I think this its no so difficult but for now i'm sorry but i don't have a dev version installed for make some test.

alkabal:
Hi

Anyone using 4 axis model can do the test ? 

I'm a little disappointed i have ask for add these code last years to source with many change done myself and these change broke the stuff i have do...

br

satiowadahc:
We could add a udev rule that maps either device descriptor to a placeholder device, and use that device in the script?

alkabal:
H
Feel free to try something if you think about pid vid I think they are the same for both device 

yesterday I have instal a Run in place so i plan to check how to make the both value available.

thanks a lot for interesting

alkabal:
Hi

Proposed patch that seem to work fine (test done inverting value 0x1c/0x9b in function)

satiowadahc:
I can try it in the morning, I can't seem to locate the USB initiliazer. Where do you grab the USB handle?

alkabal:
the more easy way for see if this work is to try to use the device !
If the patch work the spindle feed override change in the gui using the wheel after rotate feed button to lead

but for see more debug data you need to change arguments for start the component

something like that
loadusr -W xhc-whb04b-6 -HsfBUn
or
loadusr -W xhc-whb04b-6 -HsfBUh

alkabal:
But what is your device ?
Hb04

xhc-whb04b-4

xhc-whb04b-6

alkabal:
Hello all

Any news about accepting this PR ?

br

andypugh:
Is it definitely right? It does not seem to have had particularly extensive testing, only with with your specific pendant.

alkabal:
The code manage both case and i have do the double test 

This patch allow the full feature use 

Br

andypugh:
Thanks.

PR #1013, opened 2020-12-28
Add to axis a HAL abort pin

alkabal:
Hi

It is possible to request this change or a similar change for allow panel and other glade interface to know axis GUI ask to stop the process ?

Br

### DIFF
--- a/src/emc/usr_intf/axis/scripts/axis.py
+++ b/src/emc/usr_intf/axis/scripts/axis.py
@@ -744,12 +744,12 @@ class LivePlotter:
             if kind in (linuxcnc.NML_ERROR, linuxcnc.OPERATOR_ERROR):
                 icon = "error"
                 comp["error"] = True
+                time.sleep(0.3)
             else:
                 icon = "info"
             notifications.add(icon, text)
             error = e.poll()
         self.error_after = self.win.after(200, self.error_task)
-        time.sleep(0.3)
         comp["error"] = False 
 
     def update(self):
--- a/src/emc/usr_intf/axis/scripts/axis.py
+++ b/src/emc/usr_intf/axis/scripts/axis.py
@@ -356,6 +356,8 @@ class Notification(Tkinter.Frame):
         button.configure(command=lambda: self.remove(widgets))
         frame.pack(side="top", anchor="e")
         self.widgets.append(widgets)
+        if iconname == "error":
+            comp["error"] = True
 
     def remove(self, widgets):
         self.widgets.remove(widgets)
@@ -366,6 +368,7 @@ class Notification(Tkinter.Frame):
             widgets[0].destroy()
         if len(self.widgets) == 0:
             self.place_forget()
+            comp["error"] = False
 
 def soft_limits():
     def fudge(x):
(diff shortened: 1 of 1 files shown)


rene-dev:
what is the reason for the sleep?

alkabal:
Without that the signal is not see as changed (too quickly revert to false
I can understand if this is not the good way to do this !!

alkabal:
i think about adding a hal abort confirmation for replace the sleep and for make the process mores secure but the most important is found the way for allow this pr to be accepted if this is possible.

Br

alkabal:
Hello

I ask for this patch mostly because the .axisrc or USER_COMMAND_FILE=xxxx does not work at all for Replace the task_stop method with a modified version.

TclCommands.original_task_stop = TclCommands.task_stop
TclCommands.task_stop = my_task_stop

def my_task_stop(*event):
    probe_user_comp["abort"] = True
    self.original_task_stop(*event)
    probe_user_comp["abort"] = False

Did you know why ?

Br

alkabal:
if hal_present == 1:
    # Add an probe.user.abort pin
    probe_user_comp = hal.component("probe.user")
    probe_user_comp.newpin("abort", hal.HAL_BIT, hal.HAL_OUT)
    probe_user_comp.ready()

alkabal:
At the end i have found a working way from .axisrc.

```
root_window.tk.call('.toolbar.program_stop','configure','-command','my_task_stop')

def my_task_stop(*event):
    #print 'hijacked stop command'
    probe_user_comp["abort"] = True
    if s.task_mode == linuxcnc.MODE_AUTO and vars.running_line.get() != 0:
        o.set_highlight_line(vars.running_line.get())
    c.abort()
    c.wait_complete()  
    time.sleep(0.3)
    probe_user_comp["abort"] = False

# Replace the add/remove methods with a wrapped version that manages an error pin
TclCommands.task_stop = my_task_stop
commands = TclCommands(root_window)

if hal_present == 1:
    # Add an probe.user.error pin
    probe_user_comp = hal.component("axisui.user")
    probe_user_comp.newpin("abort", hal.HAL_BIT, hal.HAL_OUT)
    probe_user_comp.ready()
```

alkabal:
But if you like the idea to add this i can suggest to add a error pin at same time.

Br.

Ps ; i hope to all a good last years day !

alkabal:
Well it seem to be not perfect for now error is tripped also using (DEBUG,xxxx)

alkabal:
Raa this continue to do not work like attempted for error pin that now is never triggered.

If this feature is something wanted, please can you help ?
If this is not wanted i can close this PR and do the stuff with axisrc, i like to know what is your opinion please.

Br

andypugh:
I haven't really looked at this, but the idea of having a pin seems reasonable. 
It is probably simplest to set the pin high when there is an abort, and simply leave it high until axis resumes. 
Then the code that uses the pin needs to be triggered in an edge, not on the absolute state.

alkabal:
Hi

Thanks a lot for interesting.

You mean move the  comp["abort"] = False in other section like" def task_run" "def task_resume" (outside of a possible better way to do this abort pin work fine for the moment with pause)

But i have some trouble to get axis error pin working like axisrc (coded by kiall) https://github.com/linuxcnc-probe-screen/probe-screen-ng/blob/master/.axisrc

Br

kiall:
> But i have some trouble to get axis error pin working like axisrc (coded by kiall) https://github.com/linuxcnc-probe-screen/probe-screen-ng/blob/master/.axisrc
> 
> Br

I went ahead and submitted the code from that `.axisrc` file as a pull request: https://github.com/LinuxCNC/linuxcnc/pull/1024

alkabal:
Hi again

Maybee this is better to do something like c-morley done here
https://github.com/LinuxCNC/linuxcnc/commit/0e8ab4534f9d33fa04e679067389589d39b86a07


???

andypugh:
I think that this pull request has become rather confused. 
Feel free to re-open it if you want to re-work it. I am in favour of the pin, in principle, but the PR needs to be clearer and easier to evaluate.

PR #1015, opened 2020-12-29
docs: quick fix

snowgoer540:


andypugh:
I think that this looks correct, but the original is so consistent that it looks deliberate.

snowgoer540:
I pushed a change that Phill already merged that had some doc clarifications with it.    Afterward I realized my find/replace got a little carried away (ie I forgot to change replace to MIN_LIMIT, so both MAXIMUM_LIMIT and MINIMUM_LIMIT got changed to MAX_LIMIT). So this push just changed the accidental changes to MIN_LIMIT like they should have been.
  

  
Sorry for the inconvenience.
  

  
-Greg      
  
  
  
    
  

  
  
>

PR #1024, opened 2021-01-03
AXIS: Add an axisui.error pin

kiall:
Add an `axisui.error` pin, mirroring the behaviour of the gmoccapy error pin exactly. The pin will be set high / True upon error, and will be set low / False when all error notifications have been dismissed by the operator.

This would allow us (https://github.com/linuxcnc-probe-screen/probe-screen-ng) to remove the requirement that users install a `.axisrc` file that injects this behaviour. Here's the `.axisrc` file we currently require users to install: https://github.com/linuxcnc-probe-screen/probe-screen-ng/blob/a00dff94257a8a53c764da400bcfad415fc2b95a/.axisrc

### DIFF
--- a/docs/src/gui/axis.txt
+++ b/docs/src/gui/axis.txt
@@ -1057,6 +1057,15 @@ Type  Dir    Name
 float OUT               axisui.jog.increment
 ----
 
+Axis has a Hal output pin that indicates when an error has occurred. The
+'axisui.error' pin will remain 'TRUE' until all error notifications have
+been dismissed.
+
+----
+Type  Dir    Name
+bit   OUT    axisui.error
+----
+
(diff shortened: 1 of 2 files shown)


alkabal:
Whoo thanks a lot for correcting my attempt !

Can you check for add abort pin ?

Br

alkabal:
Abort pin seem to work fine !

alkabal:
Errata error pin seem to work fine !

alkabal:
Imo this work fine for error pin.

Ps: I have see for the other confused Pr about abort pin and i will open a new one soon, thank you.

andypugh:
Let's see what the buildbot makes if it, it looks inoccuous.

PR #1028, opened 2021-01-06
Updated propostion for tool_probe_m6 in branch vesaprobe_qt_work

alkabal:
Well i try to PR this unfinished version, i hope we can review this together

Br

### DIFF
--- a/nc_files/remap_lib/python-stdglue/stdglue.py
+++ b/nc_files/remap_lib/python-stdglue/stdglue.py
@@ -571,6 +571,9 @@ def tool_probe_m6(self, **words):
         # user sets toolchange position prior to toolchange
         # we will return here after only for XY
 ############################################################## TODO ACTUAL SYSTEM DOES NOT SHOW CONFIRMATION POPUP IF YOU M6Tx SAME TOOL AS ACTUAL
+
+############################################################## IMO TODO check if tool is 0 for reset tool number only after goto to toolchange pos
+
         try:
             self.selected_pocket =  int(self.params["selected_pocket"])
             emccanon.CHANGE_TOOL(self.selected_pocket)
@@ -628,6 +631,8 @@ def tool_probe_m6(self, **words):
 
             # Search probe
             self.execute("G38.3 Z#<_ini[TOOLSENSOR]MAXPROBE> F#<_ini[TOOLSENSOR]SEARCH_VEL>")
+            # Wait for results
+            yield INTERP_EXECUTE_FINISH
             # Check if we have get contact or not
             tool_probe_result_check_sub(self, AbsoluteFlag, switchUnitsFlag, feed_backup, METRIC_BASED)
 
@@ -635,6 +640,8 @@ def tool_probe_m6(self, **words):
                      print ("------------G38.5 used WITH SPRING SETTER-------------")
                      # Spring mounted latch probe
                      self.execute("G38.5 Z#<_ini[TOOLSENSOR]REVERSE_LATCH> F[#<_ini[TOOLSENSOR]SEARCH_VEL>*0.5]")
+                     # Wait for results
+                     yield INTERP_EXECUTE_FINISH
                      # Check if we have get contact or not
                      tool_probe_result_check_sub(self, AbsoluteFlag, switchUnitsFlag, feed_backup, METRIC_BASED)
 
@@ -643,6 +650,8 @@ def tool_probe_m6(self, **words):
 
             # Final probe
             self.execute("G38.3 Z-[#<_ini[TOOLSENSOR]TS_LATCH>*2] F#<_ini[TOOLSENSOR]PROBE_VEL>")
+            # Wait for results
+            yield INTERP_EXECUTE_FINISH
             # Check if we have get contact or not
             tool_probe_result_check_sub(self, AbsoluteFlag, switchUnitsFlag, feed_backup, METRIC_BASED)
 
@@ -706,9 +715,6 @@ def tool_probe_retract_and_check_sub(self, AbsoluteFlag, switchUnitsFlag, feed_b
                  
 # Check if we have get contact or not
 def tool_probe_result_check_sub(self, AbsoluteFlag, switchUnitsFlag, feed_backup, METRIC_BASED):
-
-            # Wait for results
-            yield INTERP_EXECUTE_FINISH
             
             if self.params[5070] == 0 or self.return_value > 0.0:
                 # if we switched units for tool change - switch back
@@ -740,5 +746,4 @@ def tool_probe_restore_sub(self, AbsoluteFlag, switchUnitsFlag, feed_backup, MET
                     print ("switched Units back to imperial")
 
             print(AbsoluteFlag, switchUnitsFlag, feed_backup, METRIC_BASED)
-          
-          
+
--- a/nc_files/remap_lib/python-stdglue/stdglue.py
+++ b/nc_files/remap_lib/python-stdglue/stdglue.py
@@ -556,7 +556,6 @@ def tool_probe_m6(self, **words):
 
         yield INTERP_EXECUTE_FINISH
 
-
         try:
 
 # TODO allow the autoprobe section to be a configurable option from some interface (restoring manual mode)
@@ -589,7 +588,7 @@ def tool_probe_m6(self, **words):
                 self.params["_backup_offset"] = self.params[5363]
             elif self.params["_coord_system"] == 593:
                 self.params["_backup_offset"] = self.params[5383]
-            print (self.params["_backup_offset"])
+            print ("backup_offset", self.params["_backup_offset"])
 
             self.execute("G10 L2 P0 Z0")                           # reset G5x offset for correct tool measurement
 
@@ -609,6 +608,7 @@ def tool_probe_m6(self, **words):
                 yield INTERP_ERROR
 
             if self.params["_ini[TOOLSENSOR]SETTER_WITH_SPRING"] == 1:
+                     print("------------G38.4 used WITH SPRING SETTER-------------")
                      # Spring mounted latch probe                                                                         # DO NOT WORK FINE WITHOUT SPRING MOUNTED PROBE AND SETTER
                      self.execute("G38.4 Z#<_ini[TOOLSENSOR]REVERSE_LATCH> F#<_ini[TOOLSENSOR]SEARCH_VEL>")
                      # Wait for results
@@ -672,11 +672,6 @@ def tool_probe_m6(self, **words):
             msg = "%d: '%s' - %s" % (e.line_number,e.line_text, e.error_message)
             print (msg)
             yield INTERP_ERROR
-
-    except Exception as e:
-        print (e)
-        self.set_errormsg("tool_probe_m6 remap error: %s" % (e))
-        yield INTERP_ERROR
 
 
 
--- a/nc_files/remap_lib/python-stdglue/stdglue.py
+++ b/nc_files/remap_lib/python-stdglue/stdglue.py
@@ -473,17 +473,19 @@ def index_lathe_tool_with_wear(self,**words):
 # REMAP=M6 modalgroup=10 python=tool_probe_m6
 #
 # auto tool probe on m6
+# move to Z max - configurable offset
 # move to tool change position for toolchange
-# wait for acknoledge of tool change
+# backup offset and position
+# wait for acknowledge of tool change
 # move to tool setter probe position
 # probe tool on tool setter
 # move back to tool change position
-# set offsets
+# set tool offsets
+# move to Z max - configurable offset
+# goto position backup
 # based on Versaprobe remap
 #
-# param 5000 holds the work piece height
-# param 4999 should be set to 1 if the 
-# machine is based in imperial
+# param _backup_offset holds the work piece height
 #
 # required INI settings
 # (Abs coordinates/ machine based units)
@@ -495,13 +497,20 @@ def index_lathe_tool_with_wear(self,**words):
 #TOOL_CHANGE_QUILL_UP = 1
 
 #[TOOLSENSOR]
-#X = 5.00
+#X = -5.00
 #Y = -1
 #Z = -1
-#PROBEHEIGHT = 2.3
+
+#SETTER_HEIGHT = 40
 #MAXPROBE =  -3
-#SEARCH_VEL = 20
-#PROBE_VEL = 5
+#TS_LATCH = 0.3
+#REVERSE_LATCH = 2
+#SEARCH_VEL = 300
+#PROBE_VEL = 10
+
+
+# What about checking machine is homed before allow this?
+# TODO ver's use a configurable setterheight that can be set manually or using probe
 
 def tool_probe_m6(self, **words):
 
@@ -516,15 +525,20 @@ def tool_probe_m6(self, **words):
     # record current position; probably should record every axis
     X = emccanon.GET_EXTERNAL_POSITION_X()
     Y = emccanon.GET_EXTERNAL_POSITION_Y()
-    Z = emccanon.GET_EXTERNAL_POSITION_Z()
-    print ("X",X)
-    print ("Y",Y)
-    print ("Z",Z)
+#    Z = emccanon.GET_EXTERNAL_POSITION_Z()
+    print ("X ",X)
+    print ("Y ",Y)
+#    print ("Z ",Z)
+    if self.current_pocket == 0:                                      # if current tool is 0 do not restore Z
+        Z = 0
+        print ("Z if TOOL 0 ",Z)
+    else:                                                             # if current tool is GT 0 try restore Z with all offset including new tool offset
+        Z = self.params[5422]
+        print ("Z if TOOL 1 ",Z)
 
     # Z up first if required
     if self.tool_change_quill_up:
         self.execute("G53 G0 Z0")
-        yield INTERP_EXECUTE_FINISH
 
     # turn off all spindles if required
     if not self.tool_change_with_spindle_on:
@@ -556,6 +570,7 @@ def tool_probe_m6(self, **words):
 
         yield INTERP_EXECUTE_FINISH
 
+
         try:
 
 # TODO allow the autoprobe section to be a configurable option from some interface (restoring manual mode)
@@ -565,13 +580,11 @@ def tool_probe_m6(self, **words):
             # move to tool setter position (from INI)
             self.execute("G90")
             self.execute("G53 G0 X#<_ini[TOOLSENSOR]X> Y#<_ini[TOOLSENSOR]Y>")
-            yield INTERP_EXECUTE_FINISH
-
             self.execute("G53 G0 Z#<_ini[TOOLSENSOR]Z>")
-            yield INTERP_EXECUTE_FINISH
+
 
             if self.params["_coord_system"] == 540:
-                self.params["_backup_offset"] = self.params[5223]    # backup G5x offset for correct tool measurement
+                self.params["_backup_offset"] = self.params[5223]         # backup G5x offset for correct tool measurement
             elif self.params["_coord_system"] == 550:
                 self.params["_backup_offset"] = self.params[5243]
             elif self.params["_coord_system"] == 560:
@@ -590,7 +603,7 @@ def tool_probe_m6(self, **words):
                 self.params["_backup_offset"] = self.params[5383]
             print ("backup_offset", self.params["_backup_offset"])
 
-            self.execute("G10 L2 P0 Z0")                           # reset G5x offset for correct tool measurement
+            self.execute("G10 L2 P0 Z0")                                  # reset G5x offset for correct tool measurement
 
             # set incremental mode
             self.execute("G91")
@@ -607,10 +620,10 @@ def tool_probe_m6(self, **words):
                 self.set_errormsg("tool_probe_m6 remap error:")
                 yield INTERP_ERROR
 
-            if self.params["_ini[TOOLSENSOR]SETTER_WITH_SPRING"] == 1:
+            if self.params["_ini[TOOLSENSOR]SETTER_WITH_SPRING"] == 1:    # DO NOT WORK FINE WITHOUT SPRING MOUNTED PROBE AND SETTER
                      print("------------G38.4 used WITH SPRING SETTER-------------")
-                     # Spring mounted latch probe                                                                         # DO NOT WORK FINE WITHOUT SPRING MOUNTED PROBE AND SETTER
-                     self.execute("G38.4 Z#<_ini[TOOLSENSOR]REVERSE_LATCH> F#<_ini[TOOLSENSOR]SEARCH_VEL>")
+                     # Spring mounted latch probe                                                                
+                     self.execute("G38.4 Z#<_ini[TOOLSENSOR]REVERSE_LATCH> F[#<_ini[TOOLSENSOR]SEARCH_VEL>*0.5]")
                      # Wait for results
                      yield INTERP_EXECUTE_FINISH
 
@@ -621,10 +634,8 @@ def tool_probe_m6(self, **words):
                          self.set_errormsg("tool_probe_m6 remap error:")
                          yield INTERP_ERROR
 
-            # Latch probe
-            self.execute("G1 Z#<_ini[TOOLSENSOR]TS_LATCH> F#<_ini[TOOLSENSOR]SEARCH_VEL>")
-            # Wait for final retract
-            yield INTERP_EXECUTE_FINISH
+            # Final Latch probe
+            self.execute("G1 Z#<_ini[TOOLSENSOR]TS_LATCH> F[#<_ini[TOOLSENSOR]SEARCH_VEL>*0.5]")
 
             # Final probe
             self.execute("G38.2 Z-[#<_ini[TOOLSENSOR]TS_LATCH>*2] F#<_ini[TOOLSENSOR]PROBE_VEL>")
@@ -641,22 +652,11 @@ def tool_probe_m6(self, **words):
             # set back absolute state
             self.execute("G90")
 
-            # Z up first if required
-            if self.tool_change_quill_up:
-                 self.execute("G53 G0 Z0")
-                 yield INTERP_EXECUTE_FINISH
-
-            # return to recorded positon
-            self.execute("G53 G0 X{:.5f} Y{:.5f}".format(X,Y))
-            yield INTERP_EXECUTE_FINISH
-            self.execute("G53 G0 Z{:.5f}".format(Z))
-            yield INTERP_EXECUTE_FINISH
-
-
             # adjust tool offset from calculations
             proberesult = self.params[5063]
             probeheight = self.params["_ini[TOOLSENSOR]SETTER_HEIGHT"]  # TODO using global_var from some interface allow to set and save this value using probe or by hand
             workheight = self.params["_backup_offset"]                  # load G5x backup_offset after correct tool measurement
+
 
             adj = proberesult - probeheight                             # IMO better to save tool offset without G5x offset and to do this separatly
             self.execute("G10 L1 P#<selected_tool> Z{}".format(adj))
@@ -666,6 +666,25 @@ def tool_probe_m6(self, **words):
 
             # apply tool offset
             self.execute("G43")
+            # Wait for acceptance
+            yield INTERP_EXECUTE_FINISH
+
+            # Z up first if required
+             if self.tool_change_quill_up:
+                 self.execute("G53 G0 Z0")
+
+            # return to recorded positon
+            self.execute("G53 G0 X{:.5f} Y{:.5f}".format(X,Y))
+#            Zcalc = Z + self.params["_backup_offset"]
+#            print("Zcalc", Zcalc)
+#            self.execute("G53 G0 Z{:.5f}".format(Zcalc))
+
+            if Z <> 0:
+                  print("******************RESTORING Z", Z)
+                  self.execute("F100")
+                  self.execute("G1 Z{:.5f}".format(Z))                   # Value ar ok but this code seem to be not executed
+                  print("******************Why not move to ", Z)
+
 
 
         except InterpreterException as e:
--- a/nc_files/remap_lib/python-stdglue/stdglue.py
+++ b/nc_files/remap_lib/python-stdglue/stdglue.py
@@ -509,6 +509,10 @@ def index_lathe_tool_with_wear(self,**words):
 #SEARCH_VEL = 12
 #PROBE_VEL = 0.4
 
+
+# What about checking machine is homed before allow this?
+# TODO ver's use a configurable setterheight that can be set manually or using probe
+
 def tool_probe_m6(self, **words):
 
     # only run this if we are really moving the machine
@@ -517,7 +521,15 @@ def tool_probe_m6(self, **words):
         yield INTERP_OK
 
     IMPERIAL_BASED = not(bool(self.params['_metric_machine']))
-    
+
+    # Saving G90 G91 at startup
+    if bool(self.params["_absolute"]) == 1:
+        print ("Absolute G90")
+        AbsoluteFlag = True
+    else:
+        print ("Incremental G91")
+        AbsoluteFlag = False
+
     # cancel tool offset
     self.execute("G49")
 
@@ -535,9 +547,6 @@ def tool_probe_m6(self, **words):
         Z = self.params[5422]
         print ("Z if TOOL 1 ",Z)
 
-    # Z up first if required
-    if self.tool_change_quill_up:
-        self.execute("G53 G0 Z0")
 
     # turn off all spindles if required
     if not self.tool_change_with_spindle_on:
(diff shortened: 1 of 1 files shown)


alkabal:
At the end all is already done for switch to G38.3 and G38.5 for bypass built in error management and allow to restore value correctly

I have not try for now but i if this is possible to do a sub for the error checking it can be better

alkabal:
For now i do not have the _metric_machine available in my 2.8 version, but i have noticed that a value can be found in :

alkabal:
[TRAJ]
LINEAR_UNITS = mm

c-morley:
I haven't tested this - just looked at code.

recording/returning to user Z position after tool change - My instincts tell me this is not the way to go, but my guess is it depends how you wish to setup your tool change routine.
Returning to absolute Z,   I think, will always be safe - unless you set up the routine to not automatically move to a toolchange position before toolchange and then picked a spot without enough head room for new tool. Which is up to the main gcode program  programmer.
I assume that the usual thing would be (automatic or not) to first move Z to the highest position
What does using user Z get you?

param [ '_metric_based'] in not available - there is no sense in having two params that give the same (but inverted) info

We need to discuss the removal and restoring of the user offsets.
essentially you are changing the probed numbers to machine units, originally the probed data was in current user units.
This makes a difference because the workpiece (currently params[5000]) must be set with the same reference - I assume it was user units in versaprobe.
I guess the first question is why do it one way or the other.

probing g38.3/5 Good thinking! This covers the probably common error of probe not tripped at least.

we can probably make the resetting of states on error a common function to tidy up the code a bit
There is also the possibility of m70/m72 to save states - i haven't tried it but would be much cleaner and save more states.

alkabal:
Hi

Thanks a lot for reply

About _metric_based change i have too late see i can also use _imperial in the ocode.. So i think i need this parts need to be restored from your code but i suspect soemthing is wrong
```
            if switchUnitsFlag:
                if IMPERIAL_BASED:
                    self.execute("G21")
                    print ("switched Units back to metric")
                else:
                    self.execute("G20")
                    print ("switched Units back to imperial")
```


For the G38.3 G38.5 Thanks to Kiall for the idea ! I have tested and this work fine.

Kiall seem to be in the process to test M73 M71 + ocode save/retstore 

Recording/returning to user Z position after tool change : i have try that because your idea to return to last Z in G53 make me think i can be better to return to last Z, but at the end i think this is not a good idea and we need to return to same value as quill up (if diameter change this can be a issue returning to  Z is possible to near to the workpieces imo) 

What does using user Z get you? :  nothing that does not work lol at all ! I think better is to return to highest Z and after that to return to last X Y but not return to Z last position. 

I guess the first question is why do it one way or the other. : i'm open to any suggestion for make things more regular it's for the moment difficult for me to know what is the best mstly because i see many different things and i really dislike to add the workpieces height in the tooltable

I appologize but i have some lack of knowledge for params[5000]  vs user units in versaprobe or bad understanding about "you are changing the probed numbers to machine units, originally the probed data was in current user units."

alkabal:
I see params[5000] i a user available var, how did you set it ?

c-morley:
>and i really dislike to add the workpieces height in the tooltable
not sure of your meaning  here

>I see params[5000] i a user available var, how did you set it ?
When I converted the versa code I had, it maintains a work height setting (probing the work piece height).
It originally would send this number in an oword routine for toolsetter probing.
Since we are not using owords I chose a temporary param (5000) to exchange this number, so the remap could use it.
When our project process is finalized I would like to add it as named param instead - but that requires some C++ code.
 
So far I can see we (have or need to add - chosen names are arbitrary):
_imperial_based
_probed_work_height
_enable_tool_probe

possibly:
_probed_toolsetter_height

I see you added something about no_spring_probe but I wonder if you could just set search_prob_velocity to zero and that would force remap to skip the fast search?

alkabal:
Restoring initial G90 G91 work fine  this is why i have move some of them and add other

 adj = proberesult - probeheight + workheight
            self.execute("G10 L1 P#<selected_tool> Z{}".format(adj))

alkabal:
did you check the use of something like #<_ini[TRAJ]LINEAR_UNITS> ?

I think this allow to do not need to add _imperial_based

alkabal:
with versprobe we do

self.usepopup = float(self.inifile.find("PROBE_SCREEN", "TOOLCHANGE_POPUP_STYLE"))

alkabal:
if self.inifile.find("TRAJ", "LINEAR_UNITS") not in ["metric", "mm"]:

alkabal:
_probed_work_height at the end how did you think to use this value ? you do not use G5x offset for that ?

alkabal:
if self.params["_ini[TRAJ]LINEAR_UNITS"] not in ["imperial", "inch"]:
	       METRIC_BASED = 1

c-morley:
>did you check the use of something like #<_ini[TRAJ]LINEAR_UNITS> ?
>I think this allow to do not need to add _imperial_based
> if self.inifile.find("TRAJ", "LINEAR_UNITS") not in ["metric", "mm"]:

This is what the C++ code does - why do you not like it?

>with versprobe we do
>self.usepopup = float(self.inifile.find("PROBE_SCREEN", "TOOLCHANGE_POPUP_STYLE"))

For What?

>_probed_work_height at the end how did you think to use this value ? you do not use G5x offset for that ?
I have to assume here a bit - I'd need to look at the code that I converted.
The whole idea is you probe a work piece height and probe a toolsetter height to get the difference.
as long as both are referenced in the same way, it doesn't matter if it's in user coordinates or absolute coordinates.
But the remap and the work height probe routine must agree on the reference.

It looks like you are removing the user coordinates offset, probing, calculating difference, then resetting the user offset.
i don't undserstand why you need to do this, but if you do then you need to make sure the workpieceheight (param[500]), 
is calculated with user coordinates offsets removed too.

alkabal:
I did "not like" it mostly because i do not use your branch so this param isn't available for now, and it is needed to use C++ vs directly read the ini ??


come back with  param (5000) work piece height
Originally vers do : 
```
adj = proberesult - probeheight + workheight
self.execute("G10 L1 P#<selected_tool> Z{}".format(adj))
```

I have change for : 
```
            adj = proberesult - probeheight                             # IMO better to save tool offset without G5x offset and to do this separatly
            self.execute("G10 L1 P#<selected_tool> Z{}".format(adj))

            adj = workheight
            self.execute("G10 L2 P0 Z{}".format(adj))                   # restore G5x offset after correct tool measurement
```

But your welcome for any suggestion for better use

Another time sorry i'm unsure about that : 
_but if you do then you need to make sure the workpieceheight (param[500]),
is calculated with user coordinates offsets removed too._

From test done this seem to work fine using : 
```

    # Down probe to workpiece for measuring it vs Know tool setter height
    def on_btn_probe_workpiece_released(self, gtkbutton, data=None):
        if self.error_poll() == 0:
             # Start psng_probe_workpiece.ngc
             if self.ocode("o<psng_probe_workpiece> call") == -1:
                 return
             a = self.probed_position_with_offsets()
             pwres = float(a[2])
             self.display_result_z(pwres)
             print("workpiecesheight =", pwres)
             print("setterheight", self.halcomp["setterheight"])
             self.spbtn_block_height.set_value(pwres)                                                                    # this call update automatically the offset for workpiece
             self.add_history(gtkbutton.get_tooltip_text(), "Z", 0, 0, 0, 0, 0, 0, 0, 0, pwres, 0, 0)
             if self.ocode("o<psng_hook_end> call") == -1:
                 return
```

```
o<psng_probe_workpiece> sub

o90 if [#<_hal[probe.setterheight]> LE 0]
     o<psng_hook_end> call
    (ABORT,TOOLSETTER HEIGHT NOT PROBED/CONFIGURED)
o90 endif

(cancel Z offsets But use tool offset)
G43          ; force tool offset for correct calc
G10 L2 P0 Z0 ; reset G54 offset for correct pieces measurement and replace by new value

#<z>=#<_z> (save current Z position)

o<psng_start_touch_probing> call
o500 if [#5070 EQ 0]
     o<psng_hook_end> call
    (ABORT,FAILURE : PROBE FOR WORKPIECE HEIGHT DOES NOT FOUND THE CONTACT POINT)
o500 endif

#<touch_result> = #5063
(print, probe Z=#<touch_result>)

; "G10 L2 P0 Z%s" % blockheight is done inside python code due to multiple possibility for set workpieces height

(back to start point)
G1 Z#<z> F#<_ini[TOOLSENSOR]RAPID_SPEED>   

o<psng_probe_workpiece> endsub
```

at end is done : 
`"G10 L2 P0 Z%s" % blockheight    # is value pwres

alkabal:
for be more exact originally vers add the blockheight to the tooltable offset, and if you change the value after that the value is registered as G5x offset... A lot of trouble with that when i try the original stuff

alkabal:
Imo is really better to save the real tool offset and to set the workpieces offset in other way

c-morley:
>I did "not like" it mostly because i do not use your branch so this param isn't available for now, and it is needed to use C++ vs >directly read the ini ??

yes that a good work around for 2.8. I am interested in making things standard in linuxcnc  _imperial_based is more clear about what you are checking while looking at code. It can be used in Oword code -checking the INI file can not be done in Qword if the value is not a number.

 >adj = proberesult - probeheight 

In your code what are these based referenced from ? G5x? machine ?

c-morley:
>param [ '_metric_based'] in not available - there is no sense in having two params that give the same (but inverted) info

Sorry I must be getting tired. _metric_machine  is the param I added to master.
I'll get some sleep soon and re-look over our discussion. Thank you.

alkabal:
Ok i understand why you want to do this way : perfect !

Well another time this is something I’m not sure but I think this is done in machine unit

alkabal:
Did you suggest than is better for me to switch to master ? (For the moment I do not want to use a rip)

alkabal:
I know too how is hard to be sleepless...

c-morley:
>Did you suggest than is better for me to switch to master ? (For the moment I do not want to use a rip)

only the _metric_machine has been added to master - my branch is based from master but you would need to RIP for it.

probably don't need to yet - we are getting a long ways with discussion IMHO

alkabal:
> >
> > with versprobe we do
> > self.usepopup = float(self.inifile.find("PROBE_SCREEN", "TOOLCHANGE_POPUP_STYLE"))
> 
> For What?
I come back with some things not clarified, this one is only a exemple for read ini.


> I see you added something about no_spring_probe but I wonder if you could just set search_prob_velocity to zero and that would force remap to skip the fast search?

Well you are right but what append only using 0 for Feedrate ? the G38 is automatically bypassed or we need to add a check ? (and good to notice to remove also the fast G38.3 i have missed that, but if we keep the setting like i have done with ini user can keep using a fast (but slow) and slow (more slow) probing and exclude only the retract probing G38.5

I can agree with what you think is better, i'm open to all suggestion ! And i really would like to help creating some standard and secured remap, i'm in the process to do this for myself and i need to take decision about keeping my logic for myself, change for other logic (i think we try to do that), use full python or use ocode...

> It looks like you are removing the user coordinates offset, probing, calculating difference, then resetting the user offset.
> i don't undserstand why you need to do this, but if you do then you need to make sure the workpieceheight (param[5000]),
> is calculated with user coordinates offsets removed too.

I really hope a review for this

alkabal:
for information i have test the code on my real machine !

alkabal:
if you agree with the direction of this code i think we now need to : 

--Review error system misunderstanding yield INTERP_ERROR
--Review comment in the code and the code
--Speak about  _enable_tool_probe (i can start to code that)
--Speak about _probed_toolsetter_height like if NONE use the one from ini ? (also need to save this value in pref file)
--Speak about G5x offset for _probed_work_height or other way
**--Speak about prevent measurement if actual tool is 0
--Add to hal_manualtoolchange a popup if you ask same tool as actual tool for prevent unwanted move without asking confirmation**
--Speak about a possible use of optional ocode for system like hook from PSNG  before and after move (i can details why i think this is needed)

c-morley:
Can you explain the rational of the g38.5 for spring probes. I'm not understanding the process.
Why do you need latch distance from the INI?
Still don't understand why you need to remove and reset user coordinates ( Though I do see an error in my process - toolsetter height from INI is in machine coordinates and everything else was in user coordinates) This is messy in the code and if we can eliminate it that would be a big plus.

>--Review error system misunderstanding yield INTERP_ERROR

tell me more.

alkabal:
Adding this work fine for bypass measurement but imo we need to ass a pin or param like you suggested before
```
        if Popen('halcmd getp probe.use_tool_measurement',stdout=PIPE,shell=True).communicate()[0].strip() == 'FALSE':
            self.execute("(DEBUG, TEST INHIBIT AUTO LENGHT)")
            yield INTERP_OK
            # apply tool offset
            self.execute("G43")
```

For G38.5 from some very old test using only something like a touchplate can cause problem with false contact

alkabal:
For user coordinates i have more or less reproduce what is done inside vers probe but if you think we can do without that your welcome. but maybe this is only because vers integrate at some point the workpieces height in the tooltable.

alkabal:
--Review error system misunderstanding yield INTERP_ERROR

I have finally see the other issue (from you) related to this code with remap

alkabal:
Test done for backup clear restore G5x offset, without this the resuting toollength is wrong (adding the offset value)

alkabal:
Test done for do not reset user offset, this is possible if we read it same to have god result using 

`adj = proberesult - probeheight + self.params["_backup_offset"]`

But i suspect a possible issue : my tooltable is with negative number and i'm unsure if this is like that for all ?

alkabal:
If this is correct for that allow to read only offset without the need to clear/restore.

alkabal:
for latch distance from ini i have think to do that for metric/imperial machine config because this is harcoded in vers probe with a check

alkabal:
Sorry for many many separated message !

I have check again the orginal code from vers and i think now i understand more, the same calc is done without resetting offset, but the trouble is due to the use of #<_hal[probe.blockheight]> rather than reading the actual real offset, so if you use probe screen for set workpieces height all is ok but if you use axis touch off button or change cordinates this is really not correct (i think lol)

alkabal:
I think i can't do something better without review, now this is in your hand ^^

swapping the line commented out allow to work with actual 2.8.1 without any change in linuxcnc code.

I have not really tested with branch vesaprobe_qt_work !

Bye

c-morley:
Why have you put so many yield INTERP_EXECUTE_FINISH in ?
They are only useful for queue breakers - unpredictable out comes.
Taking them (most of them) out should make no difference.

What I think would help, is writing Pseudo code of what you think needs to happen in remapped tool change.
There seems to be feature creep that keeps making this more and more complicated.

To make my position clear - I was trying to make a simple as possible remap - that follows linuxcnc toolchange standards as much as possible and that did not use NGC code.
My intention was for electronic tool setter as per versa probe 2.

Now I'm certainly open to some compromises.
Are we both working towards roughly the same goal?

alkabal:
> Are we both working towards roughly the same goal?

i hope ! 

Without entering in the details I like to do something robust with much as possible error detection and preventing things like measure tool 0.(so at this point unfortunately this start to be not simple as wanted)
If we can found the good compromise this can be perfect imo I like also to respect standard as much as possible (imo I have some difficulty to do that lol)


As you can see I’m a hobby coder without much knowledge so I have put many things in code that you right can be useless but I will have many issue keeping the situation under control without using many INTERP_EXECUTE_FINISH that allow the code to wait as much as possible before do other things  and execute some status checking without false positive errors

Add to this some problems with INTERP_ERROR in subcode that do nothing, maybe you can start to understand why I get here


For ngc code you mean calling ococde ? 
i really like this full python attempt but I like a customizable ocode or something like that for allow user to easily add some custom code before going to toolsensor position 
maybe is possible to read some line from a ini special section and use them as custom user code, allowing to not use ocode ??? (What is this crazy idea lol)

alkabal:
For the moment I have write this code using my own vision of what I like without real restrictions but I’m open for all suggestions or limitations.


From my opinion  at this stage the best things to do now :
First says me what you want to be changed
Second someone like you (a regular dev) try to execute my code (with urgence button in the hand)
Third this person comment out all code thinked as useless and try again. (Or ask for change if found a issue)

Br

c-morley:
So my advice is to keep things very simple till you get it to work well - then move on to making it do more.
And we could always make more then one remaps - one for each type of toolsetter (spring/no spring) for instance, since it's not likely that both would  be used on one machine.

I think you are trying to use  INTERP_EXECUTE_FINISH in a way that it doesn't work (but it's possible that it has unintentional effect)
You seem to use it as if it means 'wait till last command is done'.AIUI  It is not meant for this and in fact should make no difference to normally executed gcode commands. Are you saying that it does in fact make a difference?

The reason not to use oword subroutines is again to keep it simple and easier to set up - ie only one file needs to be found.
If someone wants custom behavior then they need to make a more complicated remap -but at least they would have a simple example to work from.

alkabal:
Yes at some time i get difference using  INTERP_EXECUTE_FINISH but that seem to work fine without this.
At the end i think adding one = unintentional effect so need to add other for "cancel" unintentional effect...  

The last one work fine.

About ocode or "other custom code" command this is something i like to be optional : if a file is specified in ini use it else do not use. 
For my own i need to manage probe input at start using M64 before machine goto toolsensor, i like to change that for direct HAL or using popen method with setp/sets) and i like to restore M65 at the end. (off course I also want to test the toolsetter connection and cancel the whole job if toolsetter is not ready)

```
# 3D Probe and Tool Setter management
def tool_probe_manager_sub(self):
            if Popen('halcmd getp  your_3Dtouch_and2_switch_in0_',stdout=PIPE,shell=True).communicate()[0].strip() == 'TRUE':
                self.execute("M64 P1")      # connected to your_3Dtouch_and2_switch_in1 or use popen method and remove motion.digital from hal
                #Popen('halcmd sets your_3Dtouch_and2_switch_in1',stdout=PIPE,shell=True).communicate()[0].strip()
```

alkabal:
```
# 3D Probe and Tool Setter management
def tool_probe_manager_sub(self):
            if Popen('halcmd getp  your_3Dtouch_and2_switch_in0_',stdout=PIPE,shell=True).communicate()[0].strip() == 'TRUE':
                self.execute("M64 P1")      # connected to your_3Dtouch_and2_switch_in1 or use popen method and remove motion.digital from hal
                #Popen('halcmd sets your_3Dtouch_and2_switch_in1',stdout=PIPE,shell=True).communicate()[0].strip()

            if Popen('halcmd getp  your_Probe_and2_switch_in0_',stdout=PIPE,shell=True).communicate()[0].strip() == 'TRUE':
                (ABORT,TOOL SETTER MISSING OR ALREADY TRIGGERED)
```

alkabal:
This one is more exact :

```
# 3D Probe and Tool Setter management
def tool_probe_manager_sub(self):
            if Popen('halcmd getp  your_combined_or2_in0_',stdout=PIPE,shell=True).communicate()[0].strip() == 'TRUE':
                self.execute("M64 P1")      # connected to your_3Dtouch_and2_switch_in1 or use popen method and remove motion.digital from hal
                #Popen('halcmd sets your_3Dtouch_and2_switch_in1',stdout=PIPE,shell=True).communicate()[0].strip() # in0 is the probe input, in1 is the on/off input for probe

            if Popen('halcmd getp  your_combined_or2_in1_',stdout=PIPE,shell=True).communicate()[0].strip() == 'TRUE':
                (ABORT,TOOL SETTER MISSING OR ALREADY TRIGGERED)
```

alkabal:
Hi

With the latest version i have sometime (not always) a issue here XYZ can be executed at the same time resulting diagonal move.
https://github.com/LinuxCNC/linuxcnc/blob/61378b4bcc9f2cfe8e96cf9e9d85926dcf4e0781/nc_files/remap_lib/python-stdglue/stdglue.py#L632

alkabal:
https://github.com/LinuxCNC/linuxcnc/pull/1028#issuecomment-760216360

Using this code call from line 630 without yeld INTERP_EXECUTE_FINISH does not work at all

copy past this to line 630 work fine
```
            if Popen('halcmd getp  your_combined_or2_in0_',stdout=PIPE,shell=True).communicate()[0].strip() == 'TRUE':
                self.execute("M64 P1")      # connected to your_3Dtouch_and2_switch_in1 or use popen method and remove motion.digital from hal
                #Popen('halcmd sets your_3Dtouch_and2_switch_in1',stdout=PIPE,shell=True).communicate()[0].strip() # in0 is the probe input, in1 is the on/off input for probe

            if Popen('halcmd getp  your_combined_or2_in1_',stdout=PIPE,shell=True).communicate()[0].strip() == 'TRUE':
                (ABORT,TOOL SETTER MISSING OR ALREADY TRIGGERED)
```

alkabal:
https://github.com/LinuxCNC/linuxcnc/blob/61378b4bcc9f2cfe8e96cf9e9d85926dcf4e0781/nc_files/remap_lib/python-stdglue/stdglue.py#L197

Here is used the  INTERP_EXECUTE_FINISH but is is for "queue breakers - unpredictable out comes." ?

If i understand correctly i can need to use some of them at some special place in the code ?

alkabal:
**If you touch the toolsetter during non probe move the restore value is not executed**

alkabal:
Hi
did you know a way for reas g5x_offset[2] from stdglue, this can be better for remove the whole coord_backup.

Br

c-morley:
Sorry for missing in action....

You can read linuxcnc status (and g53 offsets) with the linuxcnc python library.

I have purchased a tool setter probe and when I have it wired in will continue this work.

alkabal:
You really welcome ! (great news for your order)

With your confirmation for use linuxcnc.stat i have removed the whole backup coord offset and use g5x_offset now.

alkabal:
Hello all

Have you some interest about trying to continu this project ?

Best regards

alkabal:
I have rewrite many things and i'm now near to be happy with python remap

For now i have a problem with this parts of the code :

         elif self.params["selected_tool"] == self.params["tool_in_spindle"]:
               self.execute("(DEBUG,SAME TOOL AS ACTUAL PLEASE CHECK + NEED TO UNPAUSE OR ABORT)")
               yield INTERP_EXECUTE_FINISH
               if not self.stat.paused:
                  self.cmd.auto(linuxcnc.AUTO_PAUSE)
                  yield INTERP_EXECUTE_FINISH

All work fine excepted the high time needed for have pause active (like 5 seconds)
Removing any "yield INTERP_EXECUTE_FINISH" does not allow a more faster pause (many problem with message not displayed and with the next code executed too early)



Have you any suggest ?

The stuff wanted is "display message to user", (wait the time needed or some acknowledgement), after that "pause machine imediately" and "wait until pause is released" before continu to other code.

c-morley:
try:
emccanon.PROGRAM_STOP()

There is a comment in the source that it is actually a pause message

alkabal:
A hundred of thanks that work perfectly

alkabal:
Well

I'm near to be really happy with this full python remap, now i try to find a way for manage double probe output to motion.probe-input

I have a fully working way using M64 M65 (with or2 + 2*and2)
I know i can create some hal output from stglue and use them in hal but that can't work without adding more things in hal (other and2 or lut5)
I can create a component for manage multpiple probe with multiple off input

The stuff wanted need to be easy to link in hal, need to allow user to use mostly ini file for add the remap.
At this point the remap need only a few config in hal :
```
Regular link for motion.probe-input or without double input (with or2 + 2*and2)

net manual-tool-change-loop    iocontrol.0.tool-change      => iocontrol.0.tool-changed
net manual-tool-prep-loopback  iocontrol.0.tool-prepare     => iocontrol.0.tool-prepared
```

For now i do not see other way for manage that without using Digital Output

I dream about patching Linuxcnc for manage motion.probe-input and motion.setter-input with different on/off management using hal and internal Linuxcnc command emccanon



Can i know what did you think about that ?

alkabal:
Little correction, imo if we add motion.probe-input and motion.setter-input separated we do not need really something for manage on/off 
I think (probably wrong) we can only add a check in macro/program for verify if the probe triggered is the good one using a new numbered parameters params[5071] + old params[5070] and keep all the actual stuff unchanged excepting adding a second separated probe input for tool setter.

In that way i hope we do not need to change anything for G38.x code the goal is to have something that does not brake any actual config and do not more need to deal with probe status.

c-morley:
controlling probe and setter detection with an M code seems perfectly reasonable.
A 'probe connected' input pin might be useful too.

alkabal:
Oh yes another great idea i haven’t much explorer that stuff excepted for M6.

But what append with the stglue calling a custom Mcode ? i hope the yeld interpret_execut_finish can work and wait. Lol

Conclusion is having a second motion input for setter is something not wanted ?

c-morley:
I think it would be most helpful for the integrator, as using two probes on one machine is more painful then it should be.
A second HAL pin to connect to would be mush easier. This could be done with external HAL code/components but that is overly complicated.
It would be very easy to add a second probe pin to motion that is a logical OR with the current pin.

Adding a completely different tool setter probing code into linuxcnc would be much bigger project and I'm not sure what the advantage would be - but I have little experience with tool probe setters.
 
I will test out your remap code as soon as work slows down - hopefully this weekend.
Thanks for hanging in!

alkabal:
Well I have says motion setter pin but second probe is same for me !

Yes I will also think this is really too complicated actually than it need to be.

If you want to try my attempt I will post soon the last revision

alkabal:
For now I’m close to be happy but I have a big problem found at the instant 

When the machine is in pause in the remap, if i  touch accidentally my pendant wheels the remap is totally cancelled without any error or message sent.

I have always see many problem for detecting all the way of stop see my psng code here:

https://github.com/alkabal/probe-screen-ng/blob/0cc54a23e9a993f2f8187e2901db0fa9fa604a32/psng/python/base.py#L132

I suspect this problem of synching all the stop possible is not totally clean actually I always found some issue with that.

alkabal:
Did you know if your idea about adding a second probe can allow to have a numbered Parameters 5071  ?

Allow to read probed value on the same 506x value ?

alkabal:
My attempt for use user custom M-code 100 to 199 does not produce any effect except the execution of the ON_ABORT_COMMAND

[stdglue.txt](https://github.com/LinuxCNC/linuxcnc/files/7605185/stdglue.txt)

Code is attached as txt file but i like to replace M64 M65 with M1xx + need to review one check in "tool_probe_manager_sub" i suspect i will need to use "if self.params["call_number"] == 2:" and add a "if self.params["call_number"] > 2:"

Any feedback really welcome

alkabal:
problem fixed with custom M-code

[stdglue.zip](https://github.com/LinuxCNC/linuxcnc/files/7605960/stdglue.zip)

alkabal:
[stdglue.zip](https://github.com/LinuxCNC/linuxcnc/files/7612486/stdglue.zip)

Sorry i have do some mistake with renaming in last post

c-morley:
So I finally got a bit of time to try this.
The code got very much more complicated - makes it hard to assess.
Probably my fault for taking so long to get back to this :)

Looks like you tied it to AXIS gui. (I had to modify it to try it) I am hoping for generic code.

It would be helpful I think to wrap the INI reading with a try/except case to either pick a reasonable default if missing or give a helpful error message. ie. I hadn't set a probe tool number - error that I had to look in remap source code to figure out why.
In my case I changed it to default to 0 if the entry is missing.

Why does TS_height need to be only positive? all my g53 Z movement is negative.

MAX_PROBE - you add a negative in the remap, so the INI entry is positive which doesn't make sense to me.
Maybe you mean this to mean max probing distance? ie relative rather then absolute?

In my test it moved to two positions for tool changing? one requires an unpause, one a toolchange acknowledge. why two?

Does this routine require a reference tool?

Maybe a README to explain the INI settings would be helpful as I am sure I don't understand the relationship of settings.

alkabal:
thanks for checking, i appologize for the difficulty.

I agree 100% for having a generic code for all GUI, i need help for that if possible.
Ok i will try to do something for INI reading with a try/except but if you already code something like that can you share this ?

TS_height is in my head the distance from the table to tool-setter so i think it was always positive but we can remove the check "<=" if you think this is better.

For MAX_PROBE vs ini setting this is something keep from versa probe logic, relative mode so yes this is max probing distance for first probe move. (changed as your request in attached version)

Too much pause due to work in progress states. (changed in attached version)

Sorry i do not really see how to add more information for ini setting excepted to speak about custom Mcode (i think only
this part is tied it to AXIS gui)

You do not need a reference tool.
You only need to know the tool-setter height from table and take care about maximum tool lenght + tool-setter G53 Z position + max probing distance


Imo the more important things is to solve https://github.com/LinuxCNC/linuxcnc/pull/1028#issuecomment-977264327


[stdglue.zip](https://github.com/LinuxCNC/linuxcnc/files/7699194/stdglue.zip)

alkabal:
checking again i allways see some issue for wait before use tool_probe_manager_sub

exemple ["call_number"] = 4 and 5 are executed before the latch movement

alkabal:
[stdglue.zip](https://github.com/LinuxCNC/linuxcnc/files/7700087/stdglue.zip)

c-morley:
This was my code for giving a default setting if entry missing
        try:
            self.probe_number = int(self.params["_ini[TOUCH_PROBE]PROBE_NUMBER"])
        except:
            self.probe_number = 0

c-morley:
You could probably use the python HAL module to read and write HAL pins, rather then using Mcodes

alkabal:
Thanks for code

> You could probably use the python HAL module to read and write HAL pins, rather then using Mcodes

Yes i think this is possible but i have choose to use Mcode mostly because you speak about it some time ago, but i noticed this is easily customisable for user and can be used from Gcode or other method (only one file need to be edited for all probe management from all source)

alkabal:
[stdglue.zip](https://github.com/LinuxCNC/linuxcnc/files/7738478/stdglue.zip)

Update with Try except for ini entry
No more check executed before the latch movement

alkabal:
Hello

At the end can i ask what did you think about this project ?

If the direction of this attempt is not the good one i agree to close this PR. (for my own i will think about going back to ngc remap with probescreen)

If you like we can keep open and try to find solution : 
-allowing all gui to be cleared.
-and most wanted by me :  
> When the machine is in pause in the remap, if i touch accidentally my pendant wheels the remap is totally cancelled without any error or message sent. I have always see many problem for detecting all the way of stop see my psng code here: https://github.com/alkabal/probe-screen-ng/blob/0cc54a23e9a993f2f8187e2901db0fa9fa604a32/psng/python/base.py#L132 I suspect this problem of synching all the stop possible is not totally clean actually I always found some issue with that.

alkabal:
Hello
any idea why i can't get tool length from my python remap ?
i have try using self.stat.tool_offset but i get all 0 i think because i'm in a tool change process...

i have add a pause in the python code before and after using popen, checking in hal i can read the offset with "halcmd getp halui.tool.length_offset.z motion.tooloffset.z"

but trying to call that from python i get allways 0 (test with float/str and without)
float(Popen('halcmd getp halui.tool.length_offset.z',stdout=PIPE,shell=True).communicate()[0].strip())
float(Popen('halcmd getp  motion.tooloffset.z',stdout=PIPE,shell=True).communicate()[0].strip())
float(Popen('halcmd getp  halui.tool.length_offset.z, shell=True, stdout=PIPE).stdout.read())
float(Popen('halcmd getp  motion.tooloffset.z, shell=True, stdout=PIPE).stdout.read())


i have try to add self.stat.poll()

Br

alkabal:
It is because the value are negative ?

alkabal:
trying a second tool change allow to see some value but they seem related to old tool this is crazy

c-morley:
I would really try to avoid using any remap code that relies on calling Popen or halui or even motion.
I think you are playing with fire as far as keeping things in sync.
If you are using pure python then use python bindings to the interpreter and canon.
Mostly it means you can call these functions from python - you can see we have already used some of them.
https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/rs274ngc/interpmodule.cc
https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/rs274ngc/canonmodule.cc

alkabal:
many thanks i well check that !

alkabal:
A little illustration

https://www.youtube.com/watch?v=zBfd6Wx8JwQ

alkabal:
At this state this work correctly but not in all situation. like M61 (and also using the python remap with BYPASS_MEASURE = 1 but i need to update stglue remap with your last recomendation)

for having something clean i need to have this associated but i'm unsure what is the best choice, can i ask if you have a suggestion ?

G49 G43 but they are not available for remap
M61 M6 only and they are available for remap
other way ?

Also some questioning about the necessity to add some reset like in ON_ABORT_COMMAND, for now my brain is burned after resolving so much problem... (i rewrite probe_screen at this moment)

Thousands of thanks for support.


Ps : i also think about changing the value TS_POS_Z for go to tool sensor for a automatically calculated value #TS_POS_Z = [#<_z_axis_length_abs>  - [[#<_ts_height> + #<_ts_max_tool_lgt>] - #<_latch>]]
Imo opinion this is more elegant

alkabal:
I close this topic because for now i have switched back to a ngc remap for use with PSNG

They are too much things complicated for use use a full python remap with some advanced function.


Thanks a lot for all support !!

PR #1029, opened 2021-01-08
calculator.glade: increased button size for touchscreens

hansu:
I increased the button size of the calculator widget for a better usability with a touch screen.
Additionally I added some symbols and swapped the CLR button to the right.
It looks now so:
Bildschirmfoto_2021-01-08_11-37-53

I don't know if this is used anywhere else than in gmoccapy. If yes we should may create a separate version for gmoccapy?

**It would be nice if the Ok and Cancel button would be in the same size as the other buttons. But this buttons seems to be drawn by gmoccapy.** 
@gmoccapy can you do this?


andypugh:
Can I suggest in->mm and mm->in buttons?

Hans470:
If we add some buttons then also backspace button would be nice. But maybe we should keep it simple? Or is that function often needed?

andypugh:
I have the buttons on my own lathe interface. 
They are probably most useful in the UK where we have a lot of legacy imperial measuring and machine tools, but generally work in metric. (For example I have a metric CNC lathe, and machine in metric, but all my bore micrometers are imperial) 
So touching-off a boring tool generally involves the calculator on my phone. 
The US might also find it useful, as they also have largely imperial machines but sometimes need to make metric parts.

c-morley:
The calculator widget is used in Gscreen as well but your changes look good. IMHO the metric conversion buttons and back button would be good features - I added then to the qtvcp version.

Hans470:
How do we proceed now? @andypugh, can you add the conversion buttons?

andypugh:
I would have to re-acquaint myself with Glade, it's been a long while. And I have plenty of other diversions. 
Actually.. It would be very useful if someone could work out how to make packages for Glade-gtk2 for Buster. 
(ie, a deb package of Glade 3.8 ). We could then put it on the LinuxCNC repo so that it could be readily installed without having to compile it.

Hans470:
I though you have already done this for your lathe.
Why don't use that package? http://www.linuxcnc.org/dists/wheezy/base/binary-amd64/glade-gtk2_3.8.0-0ubuntu6_amd64.deb
Still working on Buster.

andypugh:
> Why don't use that package? http://www.linuxcnc.org/dists/wheezy/base/binary-amd64/glade-gtk2_3.8.0-0ubuntu6_amd64.deb

Partly because that is 3.8.0 and I would hope that 3.8.6 is a bit less buggy.

andypugh:
There is now a glade-gtk2 package (Glade 3.8.6) on the LinuxCNC server. 
sudo apt-get install glade-gtk2

Hans470:
Nice, thanks!

hansu:
It looks actually so:
calc-1

hansu:
I found two bugs that only appear with german or similar locale settings I think:
1. Repeated clicking on the "="-button when a value with a comma is displayed makes the result miltiplied by 10
2. Entering a comma value inside braces leaded to "Error"
These are fixed in https://github.com/LinuxCNC/linuxcnc/pull/1029/commits/cad67ec3f4836f5becf55c42550027aef9b705f8

And finally I added the conversion buttons whicht @andypugh suggested:

image

hansu:
I think this should better go to 2.8. Do you agree with that?

andypugh:
2.8 should only really be getting bug-fixes. (That's the idea of a stable branch) 
But I am open to persuasion if you want to state your case.

hansu:
I don't care but I think at least the bug fixes could be in 2.8 even if they are not critical.

I just changed/improved a little bit the behaviour: When doing a new calculation, beginning with a digit, the input area is cleared. Of course it is still possible to add operations to the last result. Please tell me if you like this behaviour.

hansu:
Reminder @c-morley, @gmoccapy . So busy?

gmoccapy:
Sorry, yes I have been very busy, I will check on this on Sunday and merge the commit.

Norbert

hansu:
:disappointed:

c-morley:
This code prints the answer to the terminal - is that just for debugging? Should it not be removed?
Otherwise looks good.

c-morley:
only prints to terminal if you convert mm->inch or vice versa

hansu:
@c-morley 
it prints to the glade text field  that is used for calculation and then applied to the axes...

c-morley:
`
def displayMmInch( self, widget ):
        self.eval_string = "("+ self.eval_string + ") / " + locale.format("%f", float(25.4))
        print("Mn->Inch: {}".format(self.eval_string))
        self.compute()

    def displayInchMm( self, widget ):
        self.eval_string = "("+ self.eval_string + ") * " + locale.format("%f", float(25.4))
        print("Inch->Mm: {}".format(self.eval_string))
        self.compute()
`

Looks like it prints to terminal too

hansu:
Ah I thought in general. Yes these lines print debug messages and should be removed.

c-morley:
Thank you for your work and determination.
It saddens me this took 5 months to get in - it's a common problem with linuxcnc unfortunately.

hansu:
Thanks for merging!

PR #1030, opened 2021-01-08
Mb2hal: add fnct_01, fnct_05, add bit-invert for coils, change pin names

Hans470:
I added the function codes:
-   fnct_01_read_coils
-   fnct_05_write_single_coil

Tested with devices from SAMSON (5578 and SAM Home Gateway).

For fnct_01_read_coils I added an inverting bit an therefore added sub-pins `.bit` and `.bit-inv`. I did the same to fnct_02_read_discrete_inputs and added sub pins to the other functions like it is already implemented in e.g.  fnct_03_read_holding_registers (`.int` and `.float`).
The question is: Let it in to have a consequent structure or leave it out for compatibility?

### DIFF
--- a/docs/man/es/man1/mb2hal.1
+++ b/docs/man/es/man1/mb2hal.1
@@ -33,7 +33,10 @@
 MB2HAL is a generic userspace HAL component to communicate with one or more
 Modbus devices.
 .PP
-See the Documents for more information on mb2hal
+See 
+.UR http://linuxcnc.org/docs/html/drivers/mb2hal.html 
+.UE
+for more information.
 .SH AUTHOR
 John Thornton
 .SH LICENSE
--- a/docs/man/man1/mb2hal.1
+++ b/docs/man/man1/mb2hal.1
@@ -33,7 +33,10 @@
 MB2HAL is a generic userspace HAL component to communicate with one or more
 Modbus devices.
 .PP
-See the Documents for more information on mb2hal
+See 
+.UR http://linuxcnc.org/docs/html/drivers/mb2hal.html 
+.UE
+for more information.
 .SH AUTHOR
 John Thornton
 .SH LICENSE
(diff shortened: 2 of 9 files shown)


Hans470:
By the way: Is it possible for hal components to have two inputs where it is allowed to conned only one of the two? Then we could have a `.float` and an `.int` input pin for the write operations.

andypugh:
It is possible, though a bit of a hack to detect if a pin has been connected, so it might be possible to detect the condition. 
But it is generally easier to just sum the two inputs and document that generally it is expected that only one will be used. 
As an example, mux_generic accepts both bit and integer selection. Typically you would use only one, but the unused one can be used as a "shift" and this could be the same with your proposal.

Hans470:
Thanks Andy, good idea.

Hans470:
done :heavy_check_mark:

andypugh:
I have practically zero experience with CL. Do you think that this pull request is ready to be merged?

Hans470:
What do you mean with CL? Classicladder?
Yes I think it is ready to be merged, but I thought someone else could have a look at it and do some little testing...

andypugh:
@jethornton I think you use Classic Ladder?

Hans470:
Mb2hal is a separate HAL component/driver using the [libmodbus library](https://libmodbus.org/docs/v3.0.8/), independent of Classicladder...

andypugh:
Sorry, I have been consistently failing to read the title. 
I saw "coils" and assumed ladder.

Hans470:
Just a question about the development workflow regarding the branches.
I am not really clear about the meaning and usage of the master branch. I can see only merges fro e.g. 2.8 into master but not the other way round. The bugfixes are merged into the current 2.8 I see. And only new features go into the master or experimental stuff? Just wondering if a better would have made a pull request to 2.8...

andypugh:
Bugfixes go in to where they most belong. Currently that is 2.8 (as there probably won't be any more 2.7 releases). 
Anything that goes in to 2.8 is merged upstream into master. (and anything that goes in to 2.7, too). 
At some future point a 2.9 branch will be created from Master, and then that will be stabilised (bug fixes, new-but-safe features) and then eventually released.

PR #1032, opened 2021-01-08
Try to add tool description to hal_manualtoolchange.py

alkabal:
Hi

Add tool description to hal_manualtoolchange.py but is not tested for the moment.

### DIFF
--- a/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py
+++ b/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py
@@ -3,6 +3,9 @@ import sys, os
 import gettext
 BASE = os.path.abspath(os.path.join(os.path.dirname(sys.argv[0]), ".."))
 gettext.install("linuxcnc", localedir=os.path.join(BASE, "share", "locale"), unicode=True)
+CONFIGPATH = os.environ["CONFIG_DIR"]
+toolfile = os.path.join(CONFIGPATH, tooltable)
+self.tooledit1.set_filename(toolfile)
 
 import linuxcnc, hal
 
@@ -33,8 +36,9 @@ def stop_polling_hal_in_background():
     _after = None
 
 def do_change(n):
+    tooldescr = self.tooledit1.get_toolinfo(n)[16]
     if n:
-        message = _("Insert tool %d and click continue when ready") % n
+        message = _("Insert tool %d %s and click continue when ready") % n, tooldescr
     else:
         message = _("Remove the tool and click continue when ready")
     app.wm_withdraw()


andypugh:
Please don't make pull requests for untested code. 

I like the idea, but please test it first. 

Before making a pull request it should be tested for actually working, and the LinuxCNC testsuite should be run. 

If the testsuite (runtests) doesn't pass then the buildbot will not make packages.

alkabal:
I apologize i have do this for expose the idea to better coder for try to help the user from forum.

I know this code work 100% from python probe screen but for the moment i haven't a dev environment for compile the whole Lcnc, i work only with some interpreted code like axisrc stglue python code for gladevcp.


Feel free to close this pr.

Br

alkabal:
the code use some import that is possible to be missed in the file !

c-morley:
Looks like you are using gmoccapy py/gladevcp widget code in this :
self.tooledit1.set_filename(toolfile)

Not going to work for this widget.
You need to find the INI file, to find the tooltable path, then read the file to get the comments.

looking in gladevcp tooledit_wudget  _reload() for ideas- it looks complicated because its filling widget data and must check if the info is current - not necessary for the toolchange program.

alkabal:
Thanks, i'm in the process to get rip so i can do some test soon (i hope)

At the same, i have see something that is for me a bad things, if you ask same tool as actual the system does not ask any confirmation, it is wanted to patch that ?

c-morley:
Why would you want a confirmation on a tool already in the spindle?

alkabal:
I find this dangerous with autolength, but maybe this is only because i do many test for dev

alkabal:
And maybe this i unwanted due to post processor that possibly call several time same tool.

I ask mostly for know what is wanted.

c-morley:
> find this dangerous with autolength, but maybe this is only because i do many test for dev

please explain why.

>And maybe this i unwanted due to post processor that possibly call several time same tool.

Well the operator/programmer is also responsible to not run stupid G code :)
Currently if you call the same tool again, it will probe again, which seems a waste of time.
So as an operator, I would edit out the extra tool changes.

alkabal:
Agree with you for checking post pro result and maybe change post pro for that but i think not all are able to do this easily

> find this dangerous with autolength, but maybe this is only because i do many test for dev
> 
> please explain why.

simply because with all other M6 you wait for a popup so noo move until confirmation is done except if you ask same tool, user like me with unprotected machine can have something in the way.

I do not want to force this change lol i have do that in my vers probe version and just asking for opinion for the hal_manualtoolchange, and because this is the good moment for ask before try to edit the code for description.

alkabal:
For the moment i fail don't find working solution. Sorry, i can close this PR or keep open if other wan't to try ?

Br

alkabal:
> Currently if you call the same tool again, it will probe again, which seems a waste of time.

At the end i think 2 things with code like it is  :
1 Yes waste of time and imo not good without user confirmation or cancel.
2 Very usefull feature allowing user to create soem macro for checking if the tool lenght is near the same for determinate if the tool is broken inside a big program

andypugh:
Probing the same tool again could be a good way to check that the tool hasn't broken. 

There is probably a lot to be said for probing the tool before putting it away, and screaming for help if the length has changed. 
(breakage or pull-out) 

Not that this logic belongs in here, just thinking aloud.

alkabal:
Yes your 100% right this type of stuff ca be useful and need to be encouraged.

And also yes nothing to do with the first subjects for add tool description, but useful (for me) to speak about that’s.

gmoccapy:
If you like, you may take a look in the gmoccapy.py file to check the way I do it in my GUI. May be you can just take that code and get the tool description. I do listen on Hal status tool in spindle has changed and the I do read the info.NorbertVon meinem Huawei-Tablet gesendet-------- Ursprüngliche Nachricht --------Von: alkabal <notifications@github.com>Datum: Di., 12. Jan. 2021, 13:24An: LinuxCNC/linuxcnc <linuxcnc@noreply.github.com>Cc: Subscribed <subscribed@noreply.github.com>Betreff: Re: [LinuxCNC/linuxcnc] Try to add tool description to hal_manualtoolchange.py (#1032)

Yes your 100% right this type of stuff ca be useful and need to be encouraged.



And also yes nothing to do with the first subjects for add tool description, but useful (for me) to speak about that’s.

alkabal:
Hi
Thanks for pointing that I have already try two days ago using your code but without success (lack of knowledge)

I will retry to do something later !

c-morley:
Since there is no usable code and a indication of work stoppage -I'm closing this. Feel free to open it again if you take this up again.

PR #1033, opened 2021-01-08
Moving one return INTERP_OK in stglue and deleting one

alkabal:


alkabal:
I don't know how to ask to propagate this in other branch ?

PR #1044, opened 2021-01-15
Added new version of Raspberry Pi4 and Raspberry Pi 400

elovalvo:
Added new version of Raspberry Pi4 and Raspberry Pi 400
Added also different version of 3B and 3B+
Revised for 3B


andypugh:
Sorry to be picky, but it's best to do this now rather than later. 
If you view the patch here on Github it appears that the indentation is inconsistent. 
Can you match tabs/spaces to the original file?

PR #1049, opened 2021-01-20
xhc-whb04b-6 add is-homed hal pin

alkabal:
Hi

after some user report some problem with teleop mode i have investigate more and add some hal pin "joint._.is-homed" for allowing pendant working with unhomed machine and prevent some lag when machine is homed.

https://forum.linuxcnc.org/49-basic-configuration/40581-how-to-configure-a-xhc-whb04b-pendant?start=0#195293


alkabal:
Hello again

also any news for this PR ?

andypugh:
It seems OK, though I am in no position to test it.

Issue #996, opened 2020-12-05
Restructure documentation of HAL components


hansu:
Hi,

I would like to restructure the documentation of the HAL components a little bit. I think a central overview would be nice.

Now some components are described in [Basic HAL Reference](http://linuxcnc.org/docs/2.8/html/hal/basic-hal.html) and  some in [HAL Components](http://linuxcnc.org/docs/2.8/html/hal/components.html). But a complete overview one get only on the [man page](http://linuxcnc.org/docs/2.8/html/man/man9/).

And some entries in the [table of contents](http://linuxcnc.org/docs/2.8/html/) are a little bit mixed up with the titles of the corresponding pages.

My idea is basically:
1. Move the sections 
*5. Logic Components* and *6. Conversion Components* from **Basic HAL Reference** to **HAL Components** 
2. Add the remaining components from the man page
3. Display the components in groups

What do you think about that?

@andypugh by the way: Have a look at http://linuxcnc.org/docs/2.8/html/man/man9/demux.9.html - there is a part of the NAME section in the SYNOPSIS section instead of the synopsis

andypugh:
I agree that the HAL components document should list every component in a structured way. 
It would be even better if some way could be found to display the manpage information from the master doc. The obvious way would be to link to the online manpage. (But that doesn't work offline) 

The manpages are auto-generated by halcompile, which is why they are simply grouped together alphabetically. It is difficult to do it any other way. 

With special regard to demux, it seems that, for some reason the triple-quoted two-line description gets split into the synopsis: https://github.com/LinuxCNC/linuxcnc/blob/2.8/src/hal/components/demux.comp
Which is probably an easy fix, but also means that there is a way around the untidy double-synopsis in http://linuxcnc.org/docs/2.8/html/man/man9/bldc.9.html

Hans470:
Maybe it is possible to generate or extend the html from the manpages? Then it would be availabe offline
But first, can you please explain how the html code of the  is generated?

andypugh:
At the moment we do not distribute offline html docs. The docs packages contain only PDF. 
However, I think that offline html is being looked at. (I have heard "Sphinx" mentioned, but all I can find is https://forum.linuxcnc.org/32-documents/38889-survey-the-state-of-the-documentations-and-ways-to-improve-it#164933 so possibly I saw it in passing on the IRC)

The docs generation process it almost indistinguishable from magic :-)

[...]

PR #1052, opened 2021-01-23
 2.8 some little fixes in docs

Hans470:
See https://github.com/LinuxCNC/linuxcnc/issues/996


andypugh:
Just to check, you think that these should go in to 2.8?

Hans470:
Yes

andypugh:
Thanks.

andypugh:
Can you look in on the buildbot in an hour or so and check that the docs build went cleanly?

Hans470:
Hmm a little thing in the PDF generation may went wrong.
But why are the contents on the webserver not updated yet? Is that not done automatically?
```
Formatting manual pages as PDF
....
man9/toggle.9:18: .sy request not allowed in safer mode
man9/toggle.9:18: can't open `3.eps': No such file or directory
man9/toggle2nist.9:30: .sy request not allowed in safer mode
man9/toggle2nist.9:30: can't open `4.eps': No such file or directory
	| ps2pdf - ../docs/LinuxCNC_Manual_Pages.pdf
```

andypugh:
The docs on the web server update automatically, but only if the compilation succeeds. 
It looks like something went wrong. 
http://buildbot.linuxcnc.org/buildbot/builders/0000.checkin/builds/7544
I have tried to force a rebuild, but I don't know how to force a branch (using the IRC buildbot chatbot)

andypugh:
Do the docs build cleanly for you? I rather assumed that you had tested that before submitting the pull request. 
(./configure --enable-build-docs=pdf,html )

Hans470:
I just cloned the current 2.8 (after that merge) into a clean directory and made a build.
There were even lots of other warnings that don't belong to my changed parts.
But the HTMLs and PDFs were generated correctly.
[build_stdout.log](https://github.com/LinuxCNC/linuxcnc/files/5874525/build_stdout.log)
[build_stderr.log](https://github.com/LinuxCNC/linuxcnc/files/5874524/build_stderr.log) (sorry for the partly german messages)

andypugh:
Yes, there are hundreds of warning messages. I actually managed to get rid of all of them and get an all-green build last year. Briefly.

Hans470:
Hmm seems like we have a new task...
But the question is, which warnings are real warnings and which depens on my system. E.g. some missing latex packages that are not covered by the configure script. 
Do you also get rid of the messages like `Cannot convert '.....png' to pdf` ?

andypugh:
As I recall the png-to-pdf thing is only a problem on some OS versions. 
I rather wish that we could switch to using asciidoctor-pdf, as that makes the PDFs in a single step and also makes the Chinese docs properly.

Hans470:
But what exactly causes the buildserver not to update the webserver cannot really be determined?

andypugh:
When I edited the info about the Pi images a day or two ago, the changes showed up about an hour later. 

You _can_ tell what has gone wrong by looking at the buildbot grid: http://buildbot.linuxcnc.org/buildbot/grid
The branch is at the top, and you want to see a green, or at least only yellow 2000.docs build. 
If you can click on the box to look at the log.

Hans470:
Hmm whatever causes the buildserver not to update the webserver - after your commit my changes are on the webserver.
Strange...
And on the grid view I can only see the last 5 builds, so nothing to debug the regarding 2.8 build.

andypugh:
Is it possible that the new stylesheet is changing "-" to a hyphen? 
(ie, ASCII 2D to E2 88 92)
Compare the LinuxCNC Wiki URL in my locally-made version here:
http://www.bodgesoc.org/sendkeys.1.html
To the one here:
http://linuxcnc.org/docs/2.8/html/man/man1/sendkeys.1.html

This same problem has appeared in many places in the docs. It is especially troublesome where it leads to invalid HAL pin names. For example in:
http://linuxcnc.org/docs/2.8/html/man/man1/halui.1.html
Where all the HAL pin names are invalid.

hansu:
I have no idea why that happens. It only adds the line
` <link rel="stylesheet" href="../../asciidoc.css" type="text/css" /> <link rel="stylesheet" href="../../linuxcnc.css" type="text/css" />`
 to the `<head>`-section.

My local build produces no hyphens, see here: 
[halui.1.html.tar.gz](https://github.com/LinuxCNC/linuxcnc/files/6090503/halui.1.html.tar.gz)

andypugh:
No, mine does not create hyphens either, which makes it a rather tricky thing to try to fix.

hansu:
Yes the same as with the empty images (https://github.com/LinuxCNC/linuxcnc/pull/1068#issuecomment-774645683) :-/

PR #1055, opened 2021-01-26
QTVCP - White Space Formatting

satiowadahc:
No actual code changes, just conforming the Files on the root folder to PEP8 white space standards. Will make follow up merges for sub-folders to ease the review process

### DIFF
--- a/lib/python/qtvcp/core.py
+++ b/lib/python/qtvcp/core.py
@@ -7,9 +7,11 @@ import inspect
 import sys
 if sys.version_info.major > 2:
     from gi.repository import GObject
-else: import gobject as GObject
+else:
+    import gobject as GObject
 
-import _hal, hal
+import _hal
+import hal
 from PyQt5.QtCore import QObject, QTimer, pyqtSignal
 from hal_glib import GStat
 from qtvcp.qt_istat import _IStat as IStatParent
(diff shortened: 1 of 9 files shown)


andypugh:
My impression is that we have typically not encouraged changes to code that only change formatting or (nonfunctional) whitespace as it complicates the commit history. 
Though the only statement I have found is 
http://www.linuxcnc.org/docs/html/code/contributing-to-linuxcnc.html#_follow_the_style_of_the_surrounding_code
Which suggests that dedicated reformatting commits are OK.

satiowadahc:
Reasoning I want this to go through is really for my own purpose. I run CI testing on my custom GUI that uses several instances from QTVCP. The CI complains a lot about non conforming white space. and On a previous pull request c-morley asked for the standardization's to be put in a separate pull request. 

After this, I'd like to go through and check the other complaints from PEP & PyLint such as name shadowing outer scope and cls/self member items. In my experience when PyLint complains about 60-70% of the time its a valid time to review code. The other 30-40% is fluff like white space, broad exceptions clauses, or boost not showing an object is an array, which you can denote with pylint ignores.

andypugh:
OK, seems reasonable. Especially if we want to see green boxes in the buildbot.

c-morley:
As long as we don't make a habit of it. It creates problems with stashed work/branches not merging cleanly.

satiowadahc:
Duly noted. I am a huge fan of JetBrains IDEs for merging, notably Pycharm and Clion.

They make the Process much cleaner. Couple that with GitKraken for monitoring GitFlow and the end result makes for easy Git management. And aside from Clion free:)

PR #1073, opened 2021-02-08
WJ200_vfd: Modified parity argument validation

damiodj:
Changed argument for parity to be in line with libmodbus
expected values (o for odd, e for even, n for no parity)


andypugh:
Thanks

andypugh:
Do you think that this also belongs in 2.8? It seems to fall into the category of bug fixes.

damiodj:
yes, you are correct, this should also apply to 2.8

PR #1082, opened 2021-02-12
QTVCP - Multiple Periods in File Names Causing Crashes

satiowadahc:
1) Our Millwright were trying to save files with multiple periods. This ended with some unwanted behavior and crashed Linuxcnc on occasion.

2) Class methods by definition can't access themselves, renaming local variable to clear warnings


c-morley:
Thank you

c-morley:
It seems this breaks on paths with ../ which are legal. Could you address this?

c-morley:
reference:
https://forum.linuxcnc.org/qtvcp/41425-file-editor-open-error

But hey maybe it's a unicode problem ... I'll look into it further.

c-morley:
the problem you were trying to fix was something like 'test.2.ngc' ?

satiowadahc:
My apologies I forgot about that case. Yes `test.2.ngc` was one I tested repeatedly. or `test.mm.dd.yyyy.ngc`

c-morley:
No problem. it showed a problem with file_manager that would add weird /../home paths.
In fact I never did change your code (I probably should cover the possibility I guess)
Thanks

satiowadahc:
Can I suggest this for edge case protection? https://github.com/satiowadahc/linuxcnc/tree/cw-period

c-morley:
That looks good. Do you want to pull request or shall I just add the changes myself.

c-morley:
Can you remember why multiple .. broke linuxcnc?
I have reports that in 2.8 they were being used successfully and of course they are not happy to change.

PR #1085, opened 2021-02-15
QTVCP - corecting my mistake

satiowadahc:


satiowadahc:
@c-morley is there any interfaces I haven't thought to check?

c-morley:
That should be it. Most use the default file dialogs/widgets.
I'll try to remember when I add Jura's woodpecker screen.

Thanks again

phillc54:
This commit isn't quite correct, even with a valid file it returns. This is an example of a valid file that has been given a temp filename:
/tmp/emcflt-nhNtDS.d/shape_testing_3_metric.ngc
I have commented it out in QtPlasmaC

c-morley:
Yes I guess it would be better just to check the filename rather then the whole path

c-morley:
how about this:

def file_loaded(self, obj, filename):
        if os.path.basename(filename).count('.') > 1:
            self.lastLoadedProgram = ""
            return

phillc54:
I am not 100% sure what the issue was.

c-morley:
it counts the . in the file path - any more then one will fail.
He just wants to test the filename that the user gives so they don't do program.2.ngc

phillc54:
OK, then you solution seems good.
Is the a reason for not allowing something like program.2.ngc

c-morley:
Caused crashes apparently.
https://github.com/LinuxCNC/linuxcnc/pull/1082

I'll push the change shortly - sorry for the problems

phillc54:
No worries at all, thanks.

c-morley:
ok pushed. thanks for testing code

PR #1086, opened 2021-02-15
QTVCP - Error Message Correction

satiowadahc:
Audio_player actually uses system espeak, Error asked for python version.


c-morley:
ahh yes. after python3 is complete i can go back to python libraries.
Thanks

phillc54:
This commit is not quite correct, even with a valid file it returns.
This is an example of the temp name of a valid file:
/tmp/emcflt-nhNtDS.d/shape_testing_3_metric.ngc

phillc54:
Sorry guys, I commented on the wrong commit... :-(

PR #1087, opened 2021-02-17
fix typos in docs/ and tests/

luzpaz:
Found via codespell


docs/src/gui/qtvcp_widgets.txt
```diff
@@ -1147,13 +1147,13 @@ It has a the ability to scroll the names with hardware such as a MPG. +
...
 
-the function 'getCurrentSelected()' will return a python touple, containing +
+the function 'getCurrentSelected()' will return a python tuple, containing +
 the file path and whether it's a file. +
 
 [source,python]
 ----
 temp = FILEMANAGER.getCurrentSelected()
-print 'filepath={}'.fomat(temp[0])
+print 'filepath={}'.format(temp[0])
```
luzpaz, line 1156: please review

c-morley: yes .format is right

andypugh:
It looks like one of the spelling corrections causes a test to fail. 
(Which means that the test needs to be corrected too)
`-    8 N..... MESSAGE(" I should complain if I were'nt called but I guess I can not do that")
+    8 N..... MESSAGE(" I should complain if I weren't called but I guess I can not do that")`
This file needs to be fixed to match:
https://github.com/LinuxCNC/linuxcnc/blob/master/tests/remap/fail/epilog/expected#L9

luzpaz:
That's grammatically incorrect as well. 
`I should complain if I weren't called but I guess I can not do that` should be
`I should complain if I wasn't called but I guess I can not do that`  
Can i change that as well ?

andypugh:
I think that "were" might be correct?
https://www.quora.com/Which-of-these-is-right-If-I-were-a-rich-man-or-If-I-was-a-rich-man-Why
Though in the event that the code there is executing then it probably isn't executing in the subjunctive mood.

luzpaz:
`I weren't` ? Doesn't sound right, lol. 
Anyway, here is the requested fix 97f809e

PR #1092, opened 2021-02-26
Typo correction

alkabal:


andypugh:
I am in two minds about this one. It is clearly a typo, but it is doing no harm and will only be seen by developers, who will know what it is meant to say. 
The downside is small, but it is an extra commit in the history that has no effect on function.

alkabal:
﻿
﻿
﻿
I understand feel free to close this pr !

Issue #368, opened 2017-12-03
Does not stop on probe contact during non-probe move


rene-dev:
## Here are the steps I follow to reproduce the issue:

 1. launch simulator, turn on, home
 2. command a non-continuous jog, turn feed down to slow it down.
 3. trip probe: http://linuxcnc.org/docs/html/hal/tools.html#_simulate_probe

## This is what I expected to happen:
stop moving
## This is what happened instead:
continues to move, and crash into a possibly very expensive probe
## It worked properly before this:
works in 2.7

## Information about my hardware and software:
 * I am using Debian 9
 * I am running A binary I built myself

rene-dev:
its more complicated than this: it does stop moving while jogging only if the machine is not homed.

rene-dev:
its here: https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/motion/control.c#L583
in JA world jogs are in teleop, not free.
not sure how to abort teleop jog, and how to check if jogging is active.
checking for (joint->kb_jjog_active || joint->wheel_jjog_active) doesnt work, because the axis is jogged, not the joint.

jmelson:
OK, in my old 2.7 version, the behavior was that motion would stop if the probe CHANGED to tripped status during the move.  Having the probe tripped at the START of the move would not inhibit motion.
Makes sense, so you can jog off the probe, but you HAVE to be careful.

I'm pretty sure the behavior was the same for G1 moves, but have less experience with that.

Jon

SebKuzminsky:
@jmelson did you intend to close the issue?

I think there's still some discussion to be had here and a decision to come to.

jmelson:
On 12/05/2017 12:03 PM, Sebastian Kuzminsky wrote:
>
> @jmelson <https://github.com/jmelson> did you intend to 
> close the issue?
>
> I think there's still some discussion to be had here and a 
> decision to come to.
>
>
ACCK!  Did I hit the wrong button?  No, I did NOT intend to 
close the issue!

Sorry,

Jon

Lcvette:
hey guys, i am experiencing the same issue, probe triggered during a non probing positioning move does not stop and causes a crash. I am assuming there was never a solution found for this or if there was it is not anywhere i have found while looking.  this is a bad situation obviously and i assume as old as this is maybe a solution is somewhere just not where i have found?  thanks in advance if you can direct me!

Chris

jmelson:
[...]

PR #1096, opened 2021-02-28
Addresses bug #368, stop on probe

ZincBoy:
Fixed stop on probe behavior.  In the original code the call to tpGetExecId would not return a 0 after a ngc program was run or an o word was executed from MDI.  This led to the probe input being ignored during commanded moves, a dangerous situation.  After removing the check on tpGetExecId, motion will stop when the probe contact is triggered during o word functions as well as ngc programs.

Has been compiled and tested in simulation.

### DIFF
--- a/src/emc/motion/control.c
+++ b/src/emc/motion/control.c
@@ -624,11 +624,10 @@ static void process_probe_inputs(void)
         int i;
         int aborted = 0;
 
-        if(!GET_MOTION_INPOS_FLAG() && tpQueueDepth(&emcmotDebug->coord_tp) &&
-           tpGetExecId(&emcmotDebug->coord_tp) <= 0) {
-            // running an MDI command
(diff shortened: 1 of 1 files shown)


andypugh:
Thanks

andypugh:
There might be a problem with this change. 

https://forum.linuxcnc.org/10-advanced-configuration/5596-manual-tool-change-tool-lengh-touch-off?start=110#204713

ZincBoy:
I tested with both the spindle on and off and the probing functions work as expected.  I also do not see unexpected stop on non-probe moves when the spindle is on or off.

Prior to this change, the stop on non-probe move behavior did not work as soon as a program or macro was run.  Now people will get non-probe move errors if there is noise on the probe input when previously it would be ignored.  This will most likely crop up when the spindle drive is enabled if the drive is noisy.

PR #1099, opened 2021-03-05
seems to be not java as of 2021

matkoniecz:
removal of java is useful as its presence was suggesting that documentation is extremely outdated


andypugh:
Thanks.

PR #1105, opened 2021-03-13
updated po readme

hansu:
- Updated the file list as kbabel is now lokalize

>Otherwise, format your changes as a patch with "git format-patch"
and send it to the emc-developers mailing list

Is that still a possible procedure? Otherwise it should be removed.

And I wonder why the linuxcnc.pot is on the gitignore. I thought that is the catalogue that holds all translations that are needed and should be updated if new strings are added to the GUIs.


andypugh:
Yes, it should still be possible to email a patch to the mailing list.

PR #1114, opened 2021-03-23
Add QtPlasmaC To Pncconf and Stepconf

phillc54:
These two commits add QtPlasmaC to Pncconf and Stepconf. I got a bit carried away with Stepconf and also added tandem joints. I can split the Stepconf commit into two if that would be preferable.


phillc54:
I think it was a mistake to have these combined, I will split them up.

PR #1116, opened 2021-03-26
stepconf: add qtplasmac

phillc54:
add QtPlasmaC to Stepconf

Add tandem joints (removed this)

Propose change to minlim and maxlim in ini file

### DIFF
--- a/docs/src/config/stepconf.txt
+++ b/docs/src/config/stepconf.txt
@@ -199,6 +199,11 @@ to your charge pump input. The charge pump output is connected to the base
 thread by Stepconf. The charge pump output will be about 1/2 of the maximum
 step rate shown on the Basic Machine Configuration page.
 
+* 'Plasma Arc Voltage' - If you require a Mesa THCAD to input a plasma arc
+voltage then select Plasma Arc Voltage from the list of output pins.
+This will enable a THCAD page during the setup procedure for the entry of
+the card parameters.
+
 == Parallel Port 2 Setup
 
 .Parallel Port 2 Setup Page
--- a/src/emc/usr_intf/stepconf/Submakefile
+++ b/src/emc/usr_intf/stepconf/Submakefile
@@ -9,6 +9,8 @@ STEPCONF_GLADEFILES := \
 	spindle.glade \
 	options.glade \
 	halui_page.glade \
+	ubuttons.glade \
+	thcad.glade \
 	axisx.glade \
 	axisy.glade \
 	axisz.glade \
--- a/src/emc/usr_intf/stepconf/build_HAL.py
+++ b/src/emc/usr_intf/stepconf/build_HAL.py
@@ -44,7 +44,11 @@ class HAL:
         print(_("# overwritten when you run stepconf again"), file=file)
 
         print("loadrt [KINS]KINEMATICS", file=file)
-        print("loadrt [EMCMOT]EMCMOT base_period_nsec=[EMCMOT]BASE_PERIOD servo_period_nsec=[EMCMOT]SERVO_PERIOD num_joints=[KINS]JOINTS", file=file)
+        # qtplasmac requires 3 spindles
+        if self.d.select_qtplasmac:
+            print("loadrt [EMCMOT]EMCMOT base_period_nsec=[EMCMOT]BASE_PERIOD servo_period_nsec=[EMCMOT]SERVO_PERIOD num_joints=[KINS]JOINTS num_spindles=[TRAJ]SPINDLES", file=file)
+        else:
+            print("loadrt [EMCMOT]EMCMOT base_period_nsec=[EMCMOT]BASE_PERIOD servo_period_nsec=[EMCMOT]SERVO_PERIOD num_joints=[KINS]JOINTS", file=file)
         port3name=port2name=port2dir=port3dir=""
         if self.d.number_pports>2:
              port3name = ' '+self.d.ioaddr3
@@ -67,12 +71,15 @@ class HAL:
             print("loadrt sim_parport names=%s"%name, file=file)
         if self.a.doublestep():
             print("setp parport.0.reset-time %d" % self.d.steptime, file=file)
-        encoder = SIG.PHA in inputs
-        counter = SIG.PHB not in inputs
-        probe = SIG.PROBE in inputs
+        if not self.d.select_qtplasmac:
+            encoder = SIG.PHA in inputs
+            counter = SIG.PHB not in inputs
+            probe = SIG.PROBE in inputs
+            pwm = SIG.PWM in outputs
+            pump = SIG.PUMP in outputs
+        else:
+            encoder, counter, probe, pwm, pump = False, False, False, False, False, 
         limits_homes = SIG.ALL_LIMIT_HOME in inputs
-        pwm = SIG.PWM in outputs
-        pump = SIG.PUMP in outputs
         if self.d.axes == 0:
             print("loadrt stepgen step_type=0,0,0", file=file)
         elif self.d.axes in(1,3):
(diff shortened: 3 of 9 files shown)


c-morley:
So a couple observation - tested a little bit.
The button pages need to have a vertical scroll bar as it's forcing stepconf to be taller then usual.
The qtplasmic setting on the options page need to be reformatted so they are not double spaced.
The classicladder realtime component should be loaded in the main HAL file rather then the custom file.
setting self.d.classicladder true should cause this automatically.
Is the spindle page needed in plasma?

I got this error when trying to run a sim:

Starting LinuxCNC...
twopass:invoked with <> options
twopass:found ./my-mill.hal
twopass:found /home/chris/emc-dev/lib/hallib/plasmac.tcl
twopass:found ./custom.hal
using old debounce component
it is recommended to convert to the new dbounce component

twopass:pass0: loadusr classicladder estop.clp
Shared memory conflict or RT component not loaded!
ERROR CLASSICLADDER-   Ladder memory allocation error
creating ladder-state
twopass: Error in file /home/chris/emc-dev/lib/hallib/plasmac.tcl:
    Pin 'spindle.0.on' was already linked to signal 'spindle-on'


I will keep my eye on this request and test again.

c-morley:
How do you set tandem joints?

c-morley:
oh I figured it out - unfortunate that pncconf and stepconf use different methods.

phillc54:
Yes, I was just playing around with it in steconf and it grew and grew, I should have done it separate to the qtplasmac stuff but I ended up too committed. I can pull that out and rework it separatley if you prefer.

c-morley:
It might make it easier to follow changes later for other developers. I have no strong feeling about pulling it out if it's difficult.

phillc54:
I will have a look at it after I fix the other errors.

Is it actually worthwhile having having tandem joints in stepconf?

c-morley:
oh yes I think so -it's quite popular - i think there is actually a request for it IIRC

phillc54:
"unfortunate that pncconf and stepconf use different methods."
Is that in reference to the user input, the code implementation or both?

c-morley:
Users input is what i was referring to.
In pncconf (I think) it's obvious how to get tandem axis - X ans X2 (not saying it's the best way)
In pncconf's case you even need to click through a tree node to pick either main or tandem - but that was more about showing the user less signals at a time then anything.
X and X2 etc should be sufficient.

phillc54:
I only did it that way because I couldn't open the glade files and was hand editing. I only had v3.8.6 for LnuxCNC. Last night I added 3.22 and now I can open them so I can fix that as well.

phillc54:
OK, I removed tandem joints and will work on tht after this is sorted.
I think I have fixed your requested changes.
I didn't see the classicladder issue...  I guess it may have been fixed by some of the changes.

c-morley:
Ok I pulled and tested again.
this is what I found and my opinons - for what that is worth...

If one adds (leaves) the spindle pwm signal on the parport - the config fails as it tries to add the addf for pwm:

HAL: ERROR: function 'pwmgen.make-pulses' not found
twopass: Error in file ./my-plasma.hal:
    addf failed

I removed the bad signal and plasmac tries to load (my system is too old for plasmac so I'll need to test with a different system for that.)
So improvement but we should catch that error possibility.

Did you work on layout ? that seemed the same.

I noticed that plasmac's new signals are added after 'Unused' - unused should be last so it's easy to find.


Rant warning.... (I hate TCL and dislike two pass.)

I disagree with loading components and addf ing all over the place.
addf sequence can be very important and it's hard to immediately see all the components and the sequence they are run.
Similarly I much prefer to have all the components loaded in one place - there is actually a fair amount of code used to make sure of this in stepconf and pncconf.
 
I realize twopass was made so as to help with this but I think it makes it easier (for users) to build sloppy logic.
If I was to use it, I would only use it for user custom code so they can be sloppier and still get a working config.
But the addf problem still remains.

TCL is evil and if it was up to me I would wipe it out of linuxcnc. :)
This is of course my personal opinion and I realize with no alternative, TCL is useful and here to stay. (sigh)

I'm not sure if I'm making a mountain out of a mole hill about addf - other's opinion?
I do feel that since most people will use this as a reference for anything they build after, we should try to keep to the best practice.
Best practice of creating configs is probably a separate conversation in it's self though. If I every redo pncconf I'll be looking for feedback!

I'm also not keen on calling HAL files that are stored somewhere else I think the copy should be in the config so a user doesn't have to search all over for it - these are usually new users using stepconf.

rant over...thanks for that

I see if you select plasmac that the classicladder option is hidden - does plasmac always require classicladder?
if not, the classicladder option should be still available.

I hope this is helpful to you.

phillc54:
I think I have fixed the issues you pointed out, I did miss some others as well, sometimes I can't see the forest for the trees...

I did do the layout but I needed to decrease the height of the buttons frame. One thing I can't grasp is why the StepConf window starts up offscreen at the bottom and can't be resized smaller until after the base page is displayed...

I have use a .tcl hal file up to now because of the logic that is not available in .hal files, also none of the nets that are done in the .tcl should really be changed and a bonus is that .tcl file seems to scare most folk off :)

If you are happy for me to add most of what appears in /lib/hallib/plasmac.tcl then I could add all the nets to the end of the main hal file and the loardrt and addf up near the beginning. If I do that here I will need to do it in PNcConf as well.

c-morley:
yes the opening tallness I could not figure out either - it happens with regular stepconf too. I think i'd need to remove all the pages on by one to figure it out. I believe it happened when someone switched to gtk3 - put it was a problem when I updated the gtk2 code - it's really easy to have a setting wrong that forces it and very hard to find that setting.

yes only tcl allows logic - I was whining - I wished he used python for his projects rather the TCL.
The loading/addf of components all over the place is a more important complaint.
users always mess with the code :)

yes a copy of plasmac.tcl in the config - moving the loadrt to the main file would be a nice bonus :)

phillc54:
There you go, nice tidy hal files, I think it must be close now.

I had a play with the height, set the height of window1  which changes but the outside border stays the same, weird...

c-morley:
Hey that looks good!
Better push it before we find something else :)

I see you use multiple postgui files - I'm quite happy how it's cleaner and more obvious.
I'll have to modify pncconf and stepconf use use it more.

In your opinion - are we to the point where a help button with context help would be useful? For the future I mean.

phillc54:
The multiple postgui is only for the sim panel in sim mode, I kept changing my mind whether to put it in the normal postgui but your recent push to allow multiple posgui files convinced me to do it this way.

I was thinking the same about the help while doing this and seeing it in pncconf, I think it would be good.

You probably should try the pncconf PR as well, I have probably messed something up there as well.

c-morley:
Ha I sort of figured out the problem with the large display at start up, but not sure how to fix it.
It's the parport input pin combo box pop up menu. it;s too large i guess but I have no idea why it is large when you can resize it or why that changes after the basic page.
in pages.py in the initialize function, id I add return just after population of th eoutput pin combo box - the widow is small at start up.

Gtk3 is weird

phillc54:
Haha, yep it is...

I was fooling around with parse_geometry last night and it seemed to help as well.

c-morley:
ya i see expanding then collapsing the driver setting on the basic page will make the window small too - completely independent of the parport combobox. - wow
It seems the window sizing function is deeply confused and i can't figure out how to force it to do the right thing. - i give up :)

phillc54:
This seems to trick it into working, in stepconf.py, after window.show() in the StepconfApp class __init__ function:

        window.set_position(Gtk.WindowPosition.CENTER)
        window.reshow_with_initial_size()

I found I could do it with parse_geometry() but apparently all the geometry stuff is deprecated in Gtk3 3.2+, so now we are expected to use stuff that doesn't appear to work... Weird is an understatement :)

PR #1121, opened 2021-04-04
halmodule: add pin/sig/param info methods DO NOT merge; discuss

c-morley:
looking for opinions/ code critique.
I from time to time need a list of all the pins in HAL in python and its awkward using other methods.
So added some code to halmodule.

Then I though one might often need the current value of all the pins too, so I change the pin method to collect name and value.
collecting both at the same time should be more efficient then having a method for each. 
I chose a list of dicts so in the future we could add other info into the dict without breaking current code.

The signal/param methods show the original way I just got names.

Questions: 
Any better method name then 'get_pins_info' ?
Am I getting to fancy with a list of dicts?

sample code for interactive python (helps if you have some pins already in HAL)
`import hal`
`a = hal.component('')`
`#find a current list of dics of hal pins`
`listdict = get_pins_info()`
`print listdict[0].get('name')`
`print listdict[0].get('value')`

`#find a list of signal/parameter names:`
`print get_signal_names()`
`print get_params_names()`

### DIFF
--- a/src/hal/halmodule.cc
+++ b/src/hal/halmodule.cc
@@ -1064,8 +1064,127 @@ PyObject *get_value(PyObject *self, PyObject *args) {
 
 }
 
+/*######################################*/
+/* Get all Pin names     */
+PyObject *get_pins_info(PyObject *self, PyObject *args) {
+    int next;
+    int type;
+    char str_n[] = "name";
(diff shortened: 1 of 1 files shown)


c-morley:
A version of this has been pushed to master.

PR #1122, opened 2021-04-04
Update components.txt

robertspark:
Docs >> Hal >> components.txt
Added hyperlinks to various components.
Added conv_bit_float to the list

### DIFF
--- a/docs/src/hal/components.txt
+++ b/docs/src/hal/components.txt
@@ -16,10 +16,10 @@ man axis (or perhaps 'man 1 axis' if your system requires it.)
 ----
 
 
-axis:: AXIS LinuxCNC (The Enhanced Machine Controller) Graphical User Interface.
+axis:: AXIS LinuxCNC (The Enhanced Machine Controller) Graphical User Interface. link:../man/man9/axis.9.html[Details] 
 axis-remote:: AXIS Remote Interface.
-comp:: Build, compile and install LinuxCNC HAL components.
-gladevcp:: Virtual Control Panel for LinuxCNC based on Glade, Gtk and HAL widgets.
+comp:: Build, compile and install LinuxCNC HAL components. 
+gladevcp:: Virtual Control Panel for LinuxCNC based on Glade, Gtk and HAL widgets. link:../man/man9/gladevcp.9.html[Details] 
 gs2:: HAL userspace component for Automation Direct GS2 VFD's.
 halcmd:: Manipulate the Enhanced Machine Controller HAL from the command line.
 hal_input:: Control HAL pins with any Linux input device, including USB HID devices.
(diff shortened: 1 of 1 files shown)


hansu:
I would suggest doing PR's from a separate branch. Because all changes you do on the PR's source branch (in this case the master) are linked in this PR. Further it eliminates the need for a merge/PR of your repo when pulling the changes from the upstream repo.

hansu:
And I would suggest aiming the new PR to branch 2.8 as it would be good to have it also in the current doc pages. Thanks ;-)

robertspark:
there doesn't seem to be any interest in updating or improving the docs as this PR sat for ~ 5 days.  I understand if it was code changes, but all I did was add some hyperlinks to the man docs and added some additional components to the list.  there doesn't seem to be an interest, and I did wonder if it was a little closed shop when I saw 20 odd PRs listed with some quite long dates they appeared to have been sitting around.  hence thought it best not to waste my time on doing any more.  I kind of thought that there would be a will and a simple review.   I did the changes to master as I presumed that 2.8 was stable and the process of getting changes approved or considered for that would really be a difficult process.   I may create a doc similar to what I did with for uccnc but this time probably do it in ascidoc format so it is more flexible.  https://drive.google.com/drive/folders/0B9eqGqWQVBbWNVhsZkdfQWRaYTg?usp=sharing

thanks for your guidance on this one Hand as I learnt about ascidoc and Troff / Groff

hansu:
I wouldn't say so at all. But you are right - it is sometimes a bit frustrating how the PR's are handled. But I don't know the internal apportionment who merges which part of the software. But keep in mind that the project is maintained by people in their free time. 

Don't give up. May you ask again in the forum's thread? At least JT is one with merge permissions.

You can also host your own docs version with "github pages" like I did in #1127 
That is also a good thing to give a quick overview over the changes to not grab into the sources.

andypugh:
On Fri, 9 Apr 2021 at 22:03, Rob Clegg ***@***.***> wrote:

> there doesn't seem to be any interest in updating or improving the docs as
> this PR sat for ~ 5 days.
>
There is lots of interest. But someone has to look at them all, look
through the code (which is probably unfamiliar to them)  and often set up a
relevant test config to check them.

I spent 5 hours trying to merge two pull-requests that altered the same
code last weekend.

I only really have time during _some_ weekends.

Now, as a docs change, this is easier than a core code change, so could be
done by inspection. But 5 days is just the gap between weekends.

I know it is frustrating. I can only apologise. I have a message that I
have left in my inbox since 10/10/20 because I know that I will have to
dedicate a full day to it, and haven't found the day.

It definitely helps if you discuss the PR here on the dev list.
Explaining why, and what testing you have done.
(For example linking to PDF and HTML docs created after your changes would
be an excellent proof of "not breaking things")

We are not a "little closed shop". We might be a "little open shop"

Don't ascribe to malice that which can be explained by disorganisation and
shortage of time.

I consider myself the gatekeeper of PR's to 2.8 because I volunteered to be
2.8 release manager and nobody said "no".
Seb is (I think) the 2.7 release manager.

And, as far as defined roles go, that is all we have. We are frustratingly
informal, but I would be cautious about trying to change that, as the risk
is that when someone has a defined role the pressure can make leaving the
project the only option for them if their personal situation changes only
slightly. [1]

[1] As an example, I used to use every lunch break at work to read / reply
to all of the forum. Then COVID and "work from home" happened, and the
lunchbreak disappeared. And if I was going to spend an hour on the forum it
was using hour at home, where my other stuff is...

-- 
atp

PR #1124, opened 2021-04-07
fix build debian packages python 3

TurBoss:
small change that allows building debian packages with python 3 support

### DIFF
--- a/debian/configure
+++ b/debian/configure
@@ -76,8 +76,8 @@ EXTRA_FILES=
 EXTRA_BUILD=
 KERNEL_VERSION=$TARGET
 DRIVERS=drivers.files.in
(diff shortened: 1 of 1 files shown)


SebKuzminsky:
This PR breaks packaging on platforms that don't have python3 installed, such as Debian Stretch.

rene-dev:
the python version needs to be moved down to the os specific part. or even better should be specified, some OS versions can build with both versions.

rene-dev:
This is included in the gtk3 branch, which defaults to python3, and also does packaging.
https://github.com/LinuxCNC/linuxcnc/pull/1164

TurBoss:
Thanks!

PR #1125, opened 2021-04-07
stepconf: add tandem joints for x and y axes

phillc54:
Allow tandem joints for X and/or Y axes.

Reduce initial window height.

### DIFF
--- a/src/emc/usr_intf/stepconf/build_HAL.py
+++ b/src/emc/usr_intf/stepconf/build_HAL.py
@@ -80,13 +80,11 @@ class HAL:
         else:
             encoder, counter, probe, pwm, pump = False, False, False, False, False, 
         limits_homes = SIG.ALL_LIMIT_HOME in inputs
-        if self.d.axes == 0:
-            print("loadrt stepgen step_type=0,0,0", file=file)
-        elif self.d.axes in(1,3):
-            print("loadrt stepgen step_type=0,0,0,0", file=file)
-        elif self.d.axes in(2,4):
-            print("loadrt stepgen step_type=0,0", file=file)
 
+        stepgens = '0'
+        for axis in range(1, len(self.d.axislist)):
+            stepgens += ",0"
+        print("loadrt stepgen step_type={}".format(stepgens), file=file)
 
         if encoder:
             print("loadrt encoder num_chan=1", file=file)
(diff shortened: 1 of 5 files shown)


c-morley:
So I did a quick run of this -couple things right away
in the INI the heading [AXIS_Z] etc are missing for axes
In the HAL file:
loadrt stepgen step_type=0,0,0
doesn't have enough stepgens

I see sim-hardware will need more pins added for tandem axes -  I could help with that.

phillc54:
I fixed the axis headings in the INI file and the stepgen count in tha HAL file.

I am a bit lost /unsure with the sim-hardware HAL file.

phillc54:
I think I found it, sim_axis_hardware.comp will need X2 and Y2 copies of everything?

c-morley:
Yep that's it


From: Phillip Carter ***@***.***>
Date: 2021-04-12 4:53 p.m. (GMT-08:00)
Cc: c-morley ***@***.***>, Review requested ***@***.***>
Subject: Re: [LinuxCNC/linuxcnc] stepconf: add tandem joints for x and y axes (#1125)


I think I found it, sim_axis_hardware.comp will need X2 and Y2 copies of everything?

phillc54:
I changed the sim comp and now I am not sure what is happening. I can run a sim but I cannot get it to home, the axes just continue on as if there is no home switch.

c-morley:
It looks because the signals are different - upper vrs lower:
In the sim_hadware.hal file:
net fake-both-home-z 
net fake-both-home-Z

I just havent found exactly where the problem is generated

c-morley:
in build_hal,py:

`            for j in self.d.axislist:
                print("net fake-both-{}            sim-hardware.{}bothsw-out".format(j.upper(), j.upper()), file=f1)`
should be j.lower(),j.upper()
There are other lines too

phillc54:
> in build_hal,py:
> 
> ` for j in self.d.axislist: print("net fake-both-{} sim-hardware.{}bothsw-out".format(j.upper(), j.upper()), file=f1)`
> should be j.lower(),j.upper()
> There are other lines too

That is so obvious now :)

I think it is sorted, I can now home correctly.

c-morley:
I agree that seemed to fix things.
On another note -not related to your work.
If I set home position to 0 and table travel to 0 to 8 (say for y axis) for a simulation, linuxcnc starts with the limit switches tripped.
Which shouldn't be as the switch is at -.01
I think this is related to hysteresis of the simulated limit switch.
Do you see the same thing?

phillc54:
> I agree that seemed to fix things.
> On another note -not related to your work.
> If I set home position to 0 and table travel to 0 to 8 (say for y axis) for a simulation, linuxcnc starts with the limit switches tripped.
> Which shouldn't be as the switch is at -.01
> I think this is related to hysteresis of the simulated limit switch.
> Do you see the same thing?

I haven't looked too hard but if I use all home I have a heck of a time trying to home unless I can move all joints away from 0 a bit.

phillc54:
I do see the limits tripped isue if I have all-limits-home. If I have separate all-limits and all-home it seems ok. It appears to be coming fom the lut5 when it homes.
I have a fix for the all-home issue I was having in a sim. If you are happy with the tandem stuff then I will push it and make another branch with the fix.

c-morley:
ya I say go ahead and push it

PR #1127, opened 2021-04-08
docs: reworked iocontrol/io/iov2 man pages

### LINKED ISSUE #1108: iocontrol - io - iov2 man pages
I assume that the man page for iocontrol is still included because of compatibility reasons?
If yes I would suggest only add a note like "See IO man page" instead of duplicating the content. So it looks as if there are two components.
And what does IOv2 differently? It would be nice if that could be documented.

hansu:
closes #1108

### DIFF
--- a/docs/man/man1/io.1
+++ b/docs/man/man1/io.1
@@ -1,1 +1,121 @@
-.so man1/iocontrol.1
+.TH "IOCONTROL - IO" "1" "2007-08-25" "LinuxCNC Documentation" "HAL Component" 
+.de TQ
+.br
+.ns
+.TP \\$1
+..
+
+.SH NAME
+iocontrol \- accepts NML I/O commands, interacts with HAL in userspace
+
+.SH SYNOPSIS
+
+\fBloadusr io [\-ini \fIinifile\fB]
+
+.B [EMCIO]
+.br
+.B EMCIO = iov2
+
+.SH DESCRIPTION
+These pins are created by the userspace IO controller, usually found in $LINUXCNC_HOME/bin/io
+.P
+The signals are turned on and off in userspace - if you have strict timing requirements or simply need more i/o, consider using the realtime synchronized i/o provided by \fBmotion\fR(9) instead.
+.P
+The inifile is searched for in the directory from which halcmd was run, unless an absolute path is specified.
+
+.SH PINS
+
+.TP
+\fBiocontrol.0.coolant\-flood
+(Bit, Out) TRUE when flood coolant is requested
+
+.TP
+\fBiocontrol.0.coolant\-mist 
+(Bit, Out) TRUE when mist coolant is requested
+
+.TP
+\fBiocontrol.0.emc\-enable\-in 
+(Bit, In) Should be driven FALSE when an external estop condition exists.
+
+.TP
+\fBiocontrol.0.lube 
+(Bit, Out) TRUE when lube is requested.  This pin gets driven True when
+the controller comes out of E-stop, and when the "Lube On" command gets
+sent to the controller.  It gets driven False when the controller goes
+into E-stop, and when the "Lube Off" command gets sent to the controller.
+
+.TP
+\fBiocontrol.0.lube_level 
+(Bit, In) Should be driven FALSE when lubrication tank is empty.
+
+.TP
+\fBiocontrol.0.tool\-change 
+(Bit, Out) TRUE when a tool change is requested
+
+.TP
+\fBiocontrol.0.tool\-changed 
+(Bit, In) Should be driven TRUE when a tool change is completed.
+
+.TP
+\fBiocontrol.0.tool\-number
+(s32, Out) Current tool number
+
+.TP
+\fBiocontrol.0.tool\-prep\-number 
+(s32, Out) The number of the next tool, from the RS274NGC T-word
+
+.TP
+\fBiocontrol.0.tool\-prep\-pocket
+(s32, Out) This is the pocket number (location in the tool storage
+mechanism) of the tool requested by the most recent T-word.
+
+.TP
+\fBiocontrol.0.tool\-prepare 
+(Bit, Out) TRUE when a T\fIn\fR tool prepare is requested
+
+.TP
+\fBiocontrol.0.tool\-prepared 
+(Bit, In) Should be driven TRUE when a tool prepare is completed.
+
+.TP
+\fBiocontrol.0.user\-enable\-out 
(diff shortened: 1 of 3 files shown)


hansu:
For a quick insight see

https://hansu.github.io/linuxcnc-doc/man/man1/iocontrol.1.html
https://hansu.github.io/linuxcnc-doc/man/man1/io.1.html
https://hansu.github.io/linuxcnc-doc/man/man1/iov2l.1.html

andypugh:
The description should probably mention that iocontrol is largely concerned with tool changing. 
You should probably strip out anything that I wrote, and the authorship. I threw the manpages together is a rush just to fill some gaps in the docs build process. 

I think that the iov2 page should include _all_ of the pins that it creates, rather than just the extra ones. 

I don't recall how the copyright statement got there, possibly it should be "The LinuxCNC Project", or yourself now.

hansu:
Thanks for reviewing.
Ok I'll add the remaining pins to iov2 and work on the description.
Do you think I should remove `loadusr io [-ini inifile]` from the synopsis section? Because it is generally invoked by the INI file. Or is it still a common use case?
Do you know if it's possible to include also the subsections (`.SS`) in the TOC?
And do you may know how to get rid of the linebreak after "See" in this context?:
```
.TP 0
See 
.UR http://wiki.linuxcnc.org/cgi-bin/wiki.pl?ToolchangerProtocolProposal 
.UE
```

andypugh:
I can't find any use of "loadusr io" in the demo configs. But it is technically correct. It appears that the iocontrol file is loaded by the "linuxcnc" script in most cases, by reference to the INI file. 
Which is actually a little odd. Halui tends to be loaded the same way. 

As for the URL, you are way ahead of me in even managing to get URLs to show in troff-based docs.

hansu:
>As for the URL, you are way ahead of me in even managing to get URLs to show in troff-based docs.

Ok I just thought because you made some crazy stuff in the bldc man page where I got some inspiration from. But that was 10 years ago I just figured out :smile:

hansu:
I made a new version. Please see the links [above](https://github.com/LinuxCNC/linuxcnc/pull/1127#issuecomment-816718808).

andypugh:
Thanks

hansu:
I'm sorry Andy I didn't pushed the changes from the preview yet. I wanted to wait until you agree and then amend.
Can you please reopen this PR?
And have you any idea why the coresponding issues was not closed automatically?

andypugh:
That's annoying, as I spent an hour or so merging to Master and doing test builds. 
I don't see a way to re-open the PR. Can you make a new one? 

Also, can you take a look at the current master docs (ie, after the merge that I made), as Dewey made some changes relating to his tool table work?

The link in the manpage should probably now be to http://linuxcnc.org/docs/2.8/html/config/iov2.html rather than the Wiki page. Or possibly both.

gmoccapy:
Hallo,

just to metion, I had made a lot of testing on IO_V2, while Michael Haberler did the coding. The io_v2 has been made to be able to remap toolchange parts and export the toolchange signals to allow an interaction between tool-change and tool-changed. All that is now possible with io_v2.

I do not see any need for io_v2, also the documentation gives no clue to the differences.
Why do we want to stay with this one?

Norbert

hansu:
@andypugh 
> I don't see a way to re-open the PR. Can you make a new one?

Of course.

>Also, can you take a look at the current master docs (ie, after the merge that I made), as Dewey made some changes relating to his tool table work?

Which section exactly?

>The link in the manpage should probably now be to http://linuxcnc.org/docs/2.8/html/config/iov2.html rather than the Wiki page. Or possibly both.

This link is in the iocontrol man page. In the iov2 man page there is a link to the wiki as there are some more detailed informations like the timing diagram e.g.  Or is this unnecessary?


@gmoccapy 
>I do not see any need for io_v2, also the documentation gives no clue to the differences.

You mean io (v1) ?

gmoccapy:
No, I refer to io_v2, as far as I know, io_v1 can do all the parts now witch has been supposed to be done by io_v2. 
I just am thinking about, because on all machines I used io_v2, I am using now io_v1, so I do not see the need any more to maintain two different versions.

Norbert

hansu:
I like the idea of maintaining only one iocontrol component.
But iov2 offers some more pins so there have to be some differences in function.

gmoccapy:
You are right, but none ever used the addition hal pin, as all the error handling must be done in your own remap procedure.

hansu:
I can't say anything about that as I don't have the luxury like a toolchanger. Therefore I haven't dealt with that yet.

andypugh:
> > Also, can you take a look at the current master docs (ie, after the merge that I made), as Dewey made some changes relating to his tool table work?
> 
> Which section exactly?

https://github.com/LinuxCNC/linuxcnc/commit/2dbb2f640fb87e7fe40c6d83cee381f8643233be#diff-8a067337e52076e31c98cbeded99d0d806df5eb723c0e0509c1cd547194285a9

hansu:
>Also, can you take a look at the current master docs (ie, after the merge that I made), as Dewey made some changes relating to his tool table work?

Unfortunately the iocontrol man pages looks now a little bit weird after your merge.
Fixed in https://github.com/LinuxCNC/linuxcnc/pull/1139

hansu:
And we should decide what to do with iov2.
@gmoccapy  (https://github.com/LinuxCNC/linuxcnc/pull/1127#issuecomment-821995049) and @dngarrett  says (" ... ` c) iov2 (ioControl_v2.cc -- no maintainer, deprecated)`...") that iov2 is decprecated and has no maintainer. But @jethornton  has just introduced a new iov2 doc page (https://github.com/LinuxCNC/linuxcnc/commit/69c4d11e8517c4305bc761b0725ce359746cff8d) ... :confused:

c-morley:
I think we should be careful about deciding code is deprecated just because there is no maintainer - most of out code is really like that.
IIRC iov2 adds an easy way to handle errors to the tool change process, which seems a good idea.
IIRC if you short circuit the error handling pins then it works the same as iov1.

To me it seems we should switch to using V2 and deprecate V1 as V2 does the same and more.

As for using remap to get V1 to do everything that V2 does; That may be true but that is a high bar to add to users.
Error handling would seem to be a common need for auto tool changers so why make every user invent their own system.
Using a common system allows GUI builders to code messages etc that work with many machines.

Thanks so working on these docs!

hansu:
Ok I meant more what to do with the documentary than with the component itself. Mark as deprecated or not. Seems like "not"

PR #1128, opened 2021-04-09
Display SI-standard time unit in latency-histogram.

arvidbrodin:
latency-histogram uses a mix of uS and uSec as units of time. But
large case S is the unit Siemens, and second should never be
shortened to "sec" or "Sec".

This patch changes to [greek letter mu]s in the GUI but uses "us" in
the terminal, since not all terminals support UTF-8.


andypugh:
Admirably pedantic. :-)

arvidbrodin:
Haha, yep, sure want LinuxCNC to look good. :)

Thanks!

PR #1136, opened 2021-04-17
stepconf: prevent limit swich errors when all home switches selected

phillc54:
Prevent false limit switch errors when "All home" or "All limits + homes" as a parport input.

> If I set home position to 0 and table travel to 0 to 8 (say for y axis) for a simulation, linuxcnc starts with the limit switches tripped.
> Which shouldn't be as the switch is at -.01
> I think this is related to hysteresis of the simulated limit switch.

I did see this but if I set the home switch position 0.02 or greater outside the table travel then it seems OK.

### DIFF
--- a/src/emc/usr_intf/stepconf/build_HAL.py
+++ b/src/emc/usr_intf/stepconf/build_HAL.py
@@ -266,7 +266,7 @@ class HAL:
             print("net all-limit-home => lut5.0.in-4", file=file)
             print("net all-limit <= lut5.0.out", file=file)
             for j in self.d.axislist:
-                print("net homing-{0} <= joint.{1}.homing => lut5.0.in-{1}".format(j.upper(), self.d.axislist.index(j)), file=file)
+                print("net {0}homing <= joint.{1}.homing => lut5.0.in-{1}".format(j, self.d.axislist.index(j)), file=file)
 
         # connect the joints
         for j in self.d.axislist:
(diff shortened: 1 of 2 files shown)


c-morley:
Yes this sounds like a problem between stepconf automatically adjusting the home switch so it's not on the soft limits and sim-hardware comp's setting for hysteresis of the simulated limit switch.

ahh yes here in BUILD_INI:

        # linuxcnc doesn't like having home right on an end of travel,
        # so extend the travel limit by up to .001in or .01mm

It's machine unit dependent.


Anyways I didn't test your code (I'm working again) but it looks fine to me.
Thanks for doing this!

andypugh:
This shouldn't actually be necessary in 2.8 or master. 
https://github.com/LinuxCNC/linuxcnc/commit/86df9b81584664ede66aa453c3b34f6f2aecd259
background here:
https://github.com/LinuxCNC/linuxcnc/pull/1047

phillc54:
> This shouldn't actually be necessary in 2.8 or master.
> [86df9b8](https://github.com/LinuxCNC/linuxcnc/commit/86df9b81584664ede66aa453c3b34f6f2aecd259)
> background here:
> #1047

This commit fixes an issue in only a sim config with shared switches, it has no affect on a machine config.

I think I do see what you are referring to. So we should be able to remove the "little bit extra" on MAX_LIMIT and MIN_LIMIT in both PcCconf and StepConf ?

andypugh:
> I think I do see what you are referring to. So we should be able to remove the "little bit extra" on MAX_LIMIT and MIN_LIMIT in both PcCconf and StepConf ?

Yes, we certainly _should_ be able to.

PR #1145, opened 2021-05-01
pncconf: add plasma firmwares for 5i25, 7i76e, 7i92, and 7i96

phillc54:
These firmwares change the encoder from a single A, B, and I channel to three A channels.


c-morley (approved):
Looks good to me!

PR #1148, opened 2021-05-07
Fix: miss "mM"  type in config abs_encoder.c

annhan:


andypugh:
This is a bug-fix and belongs in the 2.8 branch. Assuming that the tests pass I will apply it there.

andypugh:
This has been manually applied to 2.8. 
https://github.com/LinuxCNC/linuxcnc/commits/2.8

PR #1149, opened 2021-05-07
uspace: rtapi: set useful process titles

jepler:
This allows the realtime threads to be easily distinguished from the utility threads with `ps`:
```
$ ps -C rtapi_app H -o tid,comm
  TID COMMAND
31568 rtapi_app
31569 rtapi_app:mesg
31570 rtapi_app:T#0
```
(though one can also follow the rule that the first two are the utility threads and the following are the realtime threads from fastest to slowest)

pthread_setname_np is nonportable but a cursory search shows it's available on Linux and FreeBSD.


andypugh:
I would say that, if you think it is useful, just push it.

PR #1150, opened 2021-05-07
uspace: hal: Show involuntary context switches as hal pin

jepler:
Linux accounts the number of "involuntary context switches", occasions where it interrupted a thread in the middle of CPU-bound work.  In a properly functioning RT system, this number would always be 0.  So, it can be used to diagnose whether involuntary context switches are a cause of poor RT performance.

This may not be a great idea to do unconditionally, since it adds an additional syscall to each thread period (at the end).

The same value can be seen in /proc/<TID>/status as nonvoluntary_ctxt_switches, so it's also possible to check it manually.

Typical (running in non-realtime mode)
```
halcmd: show pin thread1.involuntary
Component Pins:
     5  s32   OUT             2  thread1.involuntary
```

### DIFF
--- a/src/hal/hal_lib.c
+++ b/src/hal/hal_lib.c
@@ -74,6 +74,11 @@ MODULE_LICENSE("GPL");
 #include <time.h>
 #endif
 
+#ifndef __KERNEL__
+#include <sys/resource.h>
+#include <sys/time.h>
+#endif
+
 char *hal_shmem_base = 0;
 hal_data_t *hal_data = 0;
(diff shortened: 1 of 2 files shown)


andypugh:
It's probably useful, to the approximately zero% of users who would ever think to look. 

It probably needs at least some documentation to be useful?

jepler:
Another problem with this is that, with the semi-mythical lxrt (uspace+rtai) build, it's not permitted to call getrusage.

I'll close this up without merging.

PR #1155, opened 2021-05-15
Alkabal patch add an axisui.abort pin

alkabal:
Add an axisui.abort pin. The pin will be set high / True after user kill operation from Axis GUI.

This pin is reverted to low / False after 0.3ms

This help to inform user component program if action is killed from GUI for prevent bad things.

If the time.sleep(0.3) is the not the good way i apologize i haven't found other working method.


alkabal:
I appologize because i don't understand what append with "Build CI"

Issue #864, opened 2020-05-21
halmeter and halscope still using gtk2


rene-dev:
they are written in c, and should be ported to gtk3.
pncconf and this are the last parts that depend on legacy gtk versions.

havardAasen:
I've done some work on this. not complete but I'm not sure if I can swap out `CList` without breaking anything. It still works with gtk2 and can be implemented in this master branch.

Isn't classicladder still using some outdated gtk functions?

rene-dev:
yes, classicladder also uses gtk2. breakign stuff isnt an issue.
can you create a PR to my gtk3 branch? https://github.com/LinuxCNC/linuxcnc/pull/943

PR #1156, opened 2021-05-15
change halscope to Gtk3

rene-dev:
Gtk2 is officially End of life. https://blog.gtk.org/2020/12/16/gtk-4-0/
This is the first step in merging GTK3 changes in master.

this closes https://github.com/LinuxCNC/linuxcnc/issues/864
Thanks a lot to @havardAasen 

The list in the halmeter and halscope now can be searched.

Im making this a separate PR, as unlike Norberts any my work on glade, gmoccapy and the other UIs this is already fully functional, and doesn't break any configurations, but you do need to install libgtk-3-dev.

this also changes CI version to ubuntu 18.04 as this is now what master requires according to
https://github.com/LinuxCNC/linuxcnc/pull/743/
and


rene-dev:
Does anyone know why the packaging fails?

havardAasen:
It's lintian that throws an error. It complains about the timestamp in
debian/changelog
The newest entry has an older timestamp then the previous entry.

tir. 18. mai 2021, 00:18 skrev Rene Hopf ***@***.***>:

rene-dev:
ok, that has nothing to do with these change. the file is the same in master.

rene-dev:
ah, this is probably related to ubuntu 18.04

havardAasen:
I think so too.

Would it be possible to add a new changelog entry called "unreleased", with
a timestamp which always gets updated when a release has been made, or
merged into master? Perhaps have this as part of the release routine?

Developers would also have the possibility to update the changelog when
committing changes and fixes.

tir. 18. mai 2021, 22:13 skrev Rene Hopf ***@***.***>:

rene-dev:
ah, fixed now. That would be a good idea, sadly I dont know anything about debian packaging, and how this is done properly.
I think the issue is that the changelog was added at a later date, because it wasnt done in time.

PR #1159, opened 2021-05-17
rs274: Temporarily force the LC_NUMERIC setting to C as needed

### LINKED ISSUE #1158: M70/M72 restore of modal state fails with comma ',' in feed paramte
In a subroutine for `REMAP=M6 modalgroup=6 ngc=...` I use M70/M72 to store and restore the modal state. When the feedrate changes between these in the subroutine I get the following G-Code error message which is displayed when loading a program: `M7x: restore_settings failed executing ' F123,0': Bad character ',' used.` The saved feedrate was F123 here. 
When the feedrate does not change this is not an issue, so a workaround is to separately store and restore the #<_feed> variable. 

My current idea where this originates: `snprinf` of a float changes behavior depending on the locale. I'm in Germany, where a decimal point is a comma. 
This is where it probably happens:  [linuxcnc/src/emc/rs274ngc/interp_convert.cc in line 2752](https://github.com/LinuxCNC/linuxcnc/blob/380bcd3b2582dfd9a6f936354a7d1faa7ded05bd/src/emc/rs274ngc/interp_convert.cc#L2752).

One solution would be to use `setlocale` prior to `snprintf` and then restore the locale as suggested here: https://stackoverflow.com/a/14753747

For completeness: 

 0. Use the german or french locale
 1. Put a subroutine for M6 in place in the ini file (`REMAP=M6 modalgroup=6 ngc=...`). Example here: https://forum.linuxcnc.org/10-advanced-configuration/5596-manual-tool-change-tool-lengh-touch-off?start=30#48235
 2. In the subroutine do an M70, change feed, M72
 3. load a gcode program

### This is what I expected to happen:
Uneventful restoration of the modal state.

### This is what happened instead:
G-Code error messae: `M7x: restore_settings failed executing ' F0,0': Bad character ',' used.`

### It worked properly before this:

### Information about my hardware and software:

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`): debian. The linuxcnc iso
 * I am running ...
   * [ x] A binary version from linuxcnc.org (including buildbot.linuxcnc.org)
 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`): 2.8.1
 * I am using this user interface (GUI) (e.g., AXIS, Touchy, gmoccapy, etc): gmoccapy
 * I am using this interface hardware vendor and chipset (e.g., parallel port, ethernet port, FPGA card): mesa 7i96

jepler:
Code in the gcode interpreter assumes that number parsing/formatting
is done according to the C locale, as gcode part programs always use
"." as the decimal separator.

However, some of the user interfaces set the LC_NUMERIC to the user's
locale and then call into the interpreter to prepare previews of the
part program.

The last time I addressed this, it was by changing one routine to work via
an ostringstream, which is not affected by locales by default.
This time I took a different approach, which is to temporarily force the
LC_NUMERIC to C at entry to certain functions, and to restore it on exit.

If UIs or other embedders of the interpreter also use threads, they
must use `uselocale` to switch from program-wide to per-thread locale
settings, or the behavior of their program becomes undefined.

This hopefully fixes multiple bugs in affected locales, and closes #1158.
However, I did not actually set up a testing environment that would allow
me to check it.  Additionally, saving & restoring parameters will now
take place in the C locale.

### DIFF
--- a/src/emc/rs274ngc/interp_convert.cc
+++ b/src/emc/rs274ngc/interp_convert.cc
@@ -1271,6 +1271,7 @@ int Interp::convert_axis_offsets(int g_code,     //!< g_code being executed (mus
 
 int Interp::convert_param_comment(char *comment, char *expanded, int len)
 {
+    FORCE_LC_NUMERIC_C;
     int i;
     char param[LINELEN+1];
     int paramNumber;
@@ -2737,6 +2738,7 @@ int Interp::gen_settings(
     double *float_current, double *float_saved,  // S, F, other
     std::string &cmd)                            // command buffer
 {
+    FORCE_LC_NUMERIC_C;
     int i, val;
     int g64_changed = 0;
     char buf[LINELEN];
@@ -2864,6 +2866,7 @@ int Interp::gen_settings(
  */
 int Interp::gen_m_codes(int *current, int *saved, std::string &cmd)
 {
+    FORCE_LC_NUMERIC_C;
     int i,val;
     char buf[LINELEN];
     for (i = 0; i < ACTIVE_M_CODES; i++) {
--- a/src/emc/rs274ngc/interp_internal.hh
+++ b/src/emc/rs274ngc/interp_internal.hh
@@ -13,6 +13,7 @@
 #ifndef INTERP_INTERNAL_HH
 #define INTERP_INTERNAL_HH
 
+#include <locale.h>
 #include <algorithm>
 #include "config.h"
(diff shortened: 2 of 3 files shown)


SebKuzminsky:
Good find, good fix.

andypugh:
Tested by the original reporter, and confirmed to work. 

Thanks.

PR #1161, opened 2021-05-18
Alkabal patch add an axisui.abort pin

alkabal:
Add an axisui.abort pin. The pin will be set high / True after user kill operation from Axis GUI.

This pin is reverted to low / False after 0.3ms

This help to inform user component program if action is killed from GUI for prevent bad things.

If the time.sleep(0.3) is the not the good way i apologize i haven't found other working method.

### DIFF
--- a/docs/src/gui/axis.txt
+++ b/docs/src/gui/axis.txt
@@ -1057,6 +1057,14 @@ Type  Dir    Name
 float OUT               axisui.jog.increment
 ----
 
+Axis has a Hal output pin that indicates when an abort has occurred. The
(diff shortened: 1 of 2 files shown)


alkabal:
Hello

I like if this pin can be available, really usefull for passing stop button status from Axis GUI to user component like probe screen, due to the use of some macro the stop button can be available, if user use this stop button the user component can't know what append. after that the status can be unpredictable.

I think adding this pin can't break any existing config.

This is something near the use of https://github.com/LinuxCNC/linuxcnc/pull/1024

Br

PR #1164, opened 2021-05-23
Master gtk3

rene-dev:
supersedes https://github.com/LinuxCNC/linuxcnc/pull/943

This branch does the following:

port gladevcp,touchy and gmoccapy to Gtk3
Not compatible with python2 or gtk2 anymore.
change default python to python3
fix dependencies and packaging for gtk3 and python3 on focal, mint20 and buster
anything older than focal/buster doesn't fully work due to bugs in Gtk, and incompatibility of older glade with python3.
Currently mostly working:

Gmoccapy
Touchy
I'm keeping track of bugs/TODOs using the gtk3 label:
https://github.com/LinuxCNC/linuxcnc/issues?q=is%3Aissue+is%3Aopen+label%3Agtk3

I'm encouraging everybody to test this, and report any bugs.


havardAasen:
Should the build system be updated to Ubuntu 20.04? I believe that yapps is still python2 in Ubuntu 18.04

rene-dev:
can somebody figure out the packaging and dependencies, I have no idea why it fails.

phillc54:
It was intended that after the changover to python3/Gtk3 that the PlasmaC config would be removed and users would move to QtPlasmaC. This has also been notified on the forum. Would it be best if I did that work on this branch or should I wait until it has been merged into master.

gmoccapy:
IMHO you should do the changes on this branch, as after merging it to master, we will for sure brake some user configs, so also the plasma stuff will cause then unconvinience, but at the same time.

Norbert

andypugh:
Is it as simple as changing DISPLAY = in the INI? 

If it is, then an extra increment could be added to the update_ini script to swap the interface. 
If there are further changes added, then those could be put in too if they are regex-able.

phillc54:
> Is it as simple as changing DISPLAY = in the INI?

No, there are more changes necessary:
[http://linuxcnc.org/docs/devel/html/plasma/qtplasmac.html#qt_modify-config](url)
 
> If it is, then an extra increment could be added to the update_ini script to swap the interface.
> If there are further changes added, then those could be put in too if they are regex-able.

There is an included python script that can do the job:
[https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/usr_intf/qtplasmac/qtplasmac-plasmac2qt.py](url)

rene-dev:
If PlasmaC doesnt work with python3, is unmaintained, and there is already a replacement it should be removed in this branch.
Currently I see good progress, Norbert and I decided to merge this end of July. Are there any objections?
I already talked to the qtpycvp developers, and they cant wait for this to get merged.

andypugh:
> There is an included [python script](https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/usr_intf/qtplasmac/qtplasmac-plasmac2qt.py) that can do the job:

Putting that same script in the update_ini script will run automatically and only once, if set up as a new INI file version. 
Which might be prepared.

phillc54:
> > There is an included [python script](https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/usr_intf/qtplasmac/qtplasmac-plasmac2qt.py) that can do the job:
> 
> Putting that same script in the update_ini script will run automatically and only once, if set up as a new INI file version.
> Which might be prepared.

Thanks, I will have a look to see if we can utilize this method.

phillc54:
> Putting that same script in the update_ini script will run automatically and only once, if set up as a new INI file version.
> Which might be prepared.

I totally confused myself trying to add it into the existing update_ini script. I did manage to get a working setup by adding a stanza into /scripts/linuxcnc.in after the check for V1.1 but I am not sure if it is appropriate to have it in this file.

[linuxcnc.in stanza](http://home.iprimus.com.au/phillcarter/temp/linuxcnc.in.txt)

andypugh:
> I did manage to get a working setup by adding a stanza into /scripts/linuxcnc.in after the check for V1.1 but I am not sure if it is appropriate to have it in this file.

I have no problem with it being there, if it works. But does the user see the message if they start LinuxCNC from the GUI menu or a desktop shortcut?

phillc54:
> I have no problem with it being there, if it works. But does the user see the message if they start LinuxCNC from the GUI menu or a desktop shortcut?

The messge is probably unnecessary, the migration GUI will appear and the user can either continue with the migration or cancel.

andypugh:
Sounds good to me.

phillc54:
@rene-dev thanks for your work on this.

c-morley:
This has broken qtvcp on gtk2 systems.
I don't remember a warning about this.
That's super annoying.
And we should probably discuss merge strategies from 2.8 to 2.9

PR #1171, opened 2021-06-01
Increasing Hal Size as we Maxed out pin count.

satiowadahc:
On this ever increasing complexity of a build, Hal in formed us we were out of memory. 
```bash
HAL: ERROR: Insufficient memory for signal 'j2-master-ctrl8'
``` 
We increased this on our current repo. The Hal File that caused this has 330 nets, in addition to the built in Hal pins. As computers increase in memory size I see no reason to not use the available memory. 

Future thought if I had more time would be to be able to give a compile flag to change the size at compile time.

Note: Tested for about 2 days and it no issues yet.

### DIFF
--- a/src/hal/hal_priv.h
+++ b/src/hal/hal_priv.h
@@ -291,7 +291,7 @@ typedef struct {
     int funct_ptr;		/* pointer to function */
 } hal_funct_entry_t;
 
-#define HAL_STACKSIZE 16384	/* realtime task stacksize */
+#define HAL_STACKSIZE 65536	/* realtime task stacksize */
 
 typedef struct {
     rtapi_intptr_t next_ptr;		/* next thread in linked list */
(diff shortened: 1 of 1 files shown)


andypugh:
Are you working with current master? I already increased this by a factor of 3 back in february: 
c439f5a586ddd005cca167ef8f4c78f2bbde5f97

satiowadahc:
Recent. I look at he history and it was only February you changed it. After version number was implemented the only changes to this were size increases. 7 size increases in 18 years

andypugh:
It seems odd that a tripling of the memory should prove inadequate so soon. 
But I didn't change the stack size. 
Would it be possible to try your config with the existing 256*4096 allocation, but with your stack size change?

satiowadahc:
We're still adding more pins as we go, so maybe I'll leave this one open for a while and decrease it after we're done adding pins.

jepler:
@satiowadahc any updates?

satiowadahc:
We definitely need the increased memory size, and stack size. It might be possible to try 512*4096. 

The mill is currently down as I'm attempting to convert it to RTAI and Debian, so I can't test 512*4096 until its up and running again.

satiowadahc:
If your curious we have an EL6695 Ethercat Card that has a few hundred pins.

andypugh:
> If your curious we have an EL6695 Ethercat Card that has a few hundred pins.

But I would expect the current HAL size to handle more than hundreds of pins. At 8 bytes a pin 256 x 4096 would be 130,000 pins. So I am wondering where the overhead is?

andypugh:
It's been quite a while and I haven't heard of any other users running out of pins. Did you find that you really needed all that memory, or did you find that the cause was something else?

satiowadahc:
We still need the memory, however being a unique mill I'm fine with keeping this as a one off branch I manually update from master as needed.

PR #1172, opened 2021-06-03
Support for PI500 and equivalent VFDs via Modbus RTU

oMtQB4:
I'm using the Klinger Born PI500 7R5G3 and the control of the VFD is working properly. Should also been working with the Sourcetronic ST500 and Powtran PI500 since they are from the same supplier with different visual appearances.

The code is based on the WJ200 VFD modbus driver within LinuxCNC which I had been using until a few weeks ago.


rene-dev:
I think there is a typo, the company name is Klinger Born.
I cannot test it, but looks good to me.

andypugh:
If it is a standalone compoinent, and it is going in to Master, then I see no reason not to merge.

src/hal/user_comps/pi500_vfd/Submakefile
```diff
@@ -0,0 +1,31 @@
...
+$(call TOOBJSDEPS, $(PI500_SRCS)) : EXTRAFLAGS += $(PI500_CFLAGS)
+
+USERSRCS += $(PI500_SRCS)
+
+../bin/pi500_vfd: $(call TOOBJS, $(PI500_SRCS)) ../lib/liblinuxcnchal.so.0 ../lib/liblinuxcncini.so.0
+	$(ECHO) Linking $(notdir $@)
+	$(Q)$(CC) $(LDFLAGS) -o $@ $^ $(PI500_LIBS) 
+
+hal/user_comps/pi500_vfd/pi500_vfd.c: hal/user_comps/pi500_vfd/pi500_vfd.comp ../bin/halcompile
+	$(ECHO) "Preprocessing wj200_vfd.comp"
```
andypugh, line 16: I think that this probably shouldn't say WJ200 any more? 

Does this comp actually need its own submakefile? I created a userspace component in .comp format a few months  ago and just had to add it to the /user-comps/Submakefile rather than make a new one.

andypugh (approved):

src/hal/user_comps/pi500_vfd/Submakefile
oMtQB4: Thanks for the review. You are totally right. The WJ200 shouldn't be there anymore.

Regarding the Submakefile, I think it is more readable than put everything in the same Submakefile. Putting it all into one file is more error prone to changes.

oMtQB4: I corrected the wj200 into pi500

oMtQB4 (commented):
Changed wj200 to pi500 in a log

jepler:
The previous CI failed, possibly due to this branch being out of date. I've attempted to bring it up to date.

PR #1175, opened 2021-06-06
debian/configure: Drop outdated support for rtai and xenomai

jepler:
There are no packaged rtai or xenomai kernels available.

This does NOT drop support for building with src/configure.

It can be restored if there are ever properly packaged rtai or xenomai kernels again.


andypugh:
I am a little confused by the assertion that there are no RTAI packages. 
http://www.linuxcnc.org/dists/buster/base/binary-amd64/
Has kernel and RTAI application  packages.

rene-dev:
oh, I thought rtai was non-functional on buster.

andypugh:
It's complicated. Less than 1% of the time it crashes the kernel on exit. Which means that it can't reliably make it through runtests. 
But it is perfectly usable if you can accept that problem.

jepler:
Please feel free to revert this change if you like.  (xenomai should still go away though)

PR #1184, opened 2021-06-10
Create IEC Timers

satiowadahc:
Created Standalone RT HAL components for IEC timers T_ON, T_OFF, T_P. 
I was loading classic ladder for the sake of two timers. this saves a little bit and makes hal files more pedantic to read.


andypugh:
You seem to have rather done this the hard way, could these not have been very simple .comp files? 
I am not suggesting that you re-do them, but it is something to consider in the future, especially as .comp files get compiled automatically just by existing in the hal components folder. 
A very tiny nit is that the comments for t_off describe it as a t_on, and t_p has "t_pelse" in the comment.

c-morley:
I wonder if using s32 rather then u32 pins would be more useful - u32 are fairly rare in linuxcnc. do you need the range of u32?

rene-dev:
I agree with @c-morley and @andypugh

satiowadahc:
I was going off of debounce.c. I'll check out the .comp syntax if I need to do something like this again. 

Documentation updated.

Timers are in ns as they get time from threads. u32 gives up to 4.2 seconds of delay. Logically having negative time doesn't make sense as the elapsed timer can only increment from 0 and the pulse timer can't go backwards in time, so allowing negatives would not be ideal. I could modify it to be ms which might be more practical in some cases, then s32 could work (24 days) but still would need error checking for negative numbers, would this be preferred?

rene-dev:
What's wrong with using float and seconds? I always prefer SI base units.

satiowadahc:
Converted to Float per request.
I tend to work strictly on micro-controllers and floats are expensive memory and calculation wise, hence I'm used to avoiding them. Doesn't matter as much on a processor.

c-morley:
would abs() the time rather then error-ing with negative number be more user friendly?

andypugh:
I assumed that it avoided floats so that it could run in a non-fp base thread. If it is useful to do that then perhaps not using floats was right? (Sorry, I didn't see the float comment at the time)

andypugh:
If you _are_ using floats then the use_fp bit in hal_export_funct needs to be set.

rene-dev:
@andypugh Non fp threads are a thing from the past. @satiowadahc even many cheap microcontrollers have a fpu now.

andypugh:
> @andypugh Non fp threads are a thing from the past. @satiowadahc even many cheap microcontrollers have a fpu now.

I think that the base thread on most LinuxCNC systems is still non-fp.  Base-thread FP defaults to on if you use "loadrt threads" for example, but the default in motion.c is still "off": https://github.com/LinuxCNC/linuxcnc/blob/2e75b091a23e3312b9e57a88941714701f3709b8/src/emc/motion/motion.c#L44 
The repository does not seem to contain many overrides for the base_thread_fp modparam.

satiowadahc:
The origin of this came from our machinists who don't know programming very well but can do some PLC code. 
@c-morley In that regards, personal preference is to fail fast, fail loud, instead of just changing the sign.
@rene-dev I'm just a programmer I don't get much say on what hardware I use, thats on the engineers :P 
@andypugh Bit Set. Any other Requests for this PR?

andypugh:
In your application, are you running this in the base thread or the servo thread? I can not think of many applications for running such a thing in the base thread, as nothing useful is likely to respond to it (as they will be in the servo thread). 
If it typically runs in the servo thread then I have nothing more to say, but if it runs in the base thread then we need to sort out whether existing configs are setting base_thread_fp. My expectation is that they are not, in general.

satiowadahc:
I rarely run anything on base thread, almost everything is on servo-thread with the odd exception of some custom python hal components run on their own threads.

rene-dev:
Like andy, I also prefer comp for such a simple component, but if you don't want to change it now, let's just keep it like this.

PR #1185, opened 2021-06-13
qtvcp: don't show unnecessary startup error messages

phillc54:


c-morley (approved):
Looks good. Thanks for doing this.

PR #1191, opened 2021-06-18
Fix source comment typos

luzpaz:
Found via `codespell`


andypugh:
Is there any way to limit these changes to only text that a user will see? I tend to the opinion that spell-checking comment text in the code adds unnecessary noise to the commit history.

PR #1192, opened 2021-06-18
Halscope

havardAasen:
Some minor changes to the halscope files.

The largest change is the creation of dialog windows. I believe this is a more modern way of doing it. I also find it easier to read. The only user visible change is the message dialogs. The title size has changed and the buttons has become wider.

### DIFF
--- a/src/hal/utils/meter.c
+++ b/src/hal/utils/meter.c
@@ -41,7 +41,7 @@
     any responsibility for such compliance.
 
     This code was written as part of the EMC HAL project.  For more
-    information, go to www.linuxcnc.org.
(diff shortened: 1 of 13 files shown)


andypugh (commented):
Does this work with GTK3? Perhaps this belongs as a push to the master-gtk3 branch?

havardAasen:
Yes, it works with GTK3. Halscope should be identical in both master and master-gtk3, so it shouldn't matter which branch it gets merged into.

rene-dev:
@andypugh I merged his work on porting halscope to gtk3 weeks ago into master. So far no complaints.

PR #1196, opened 2021-06-23
Display SI standard units in halscope

arvidbrodin:
Fixes:

"kelvin-hertz" (KHz) -> kilohertz (kHz)
"mSec" -> ms
"uSec", "uS" -> us
"nSec" -> ns
"Mhz" -> "MHz".

### DIFF
--- a/src/hal/utils/scope_files.c
+++ b/src/hal/utils/scope_files.c
@@ -276,7 +276,7 @@ void write_log_file (char *filename)
 	log->order=INTERLACED;
 
     /* write data */
-    fprintf(fp, "Sampling period is %i nSec \n", sample_period_ns );
+    fprintf(fp, "Sampling period is %i ns \n", sample_period_ns );
 
 	/* point to the first sample in the display buffer */
 	start = ctrl_usr->disp_buf ;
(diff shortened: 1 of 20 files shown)


andypugh:
I am generally in favour. 
Is "µs" too hard?

arvidbrodin:
"µs" would be better of course, but just replacing u with µ in the source file results in a compile error:

```
Localizing linuxcnc.pot
/usr/bin/xgettext: Non-ASCII string at src/hal/utils/scope_horiz.c:1161.
                   Please specify the source encoding through --from-code.
```

I might look into this further--maybe the solution is to not localize units at all?--but that change should probably be in its own commit then, in case it causes further issues and needs to be revoked?

andypugh:
The whole point of SI is that it should not need localisation. (or, localization :-) ) 

So, it seems that it is possible to fix it in every language _but_ English in the .po file?

arvidbrodin:
Well, I removed the "translation ornaments" in the source file (the "_()") and it compiles with the µ character and seems to work - at least on my debian system:
Halscope-SI

Do you want me to push this change too?

andypugh:
Yes, I think so. At least add it to the pull request and we can see what the checking system makes of it.

arvidbrodin:
I took an additional look at e.g. src/po/ru.po and the units do have translations, e.g. "KHz" -> "КГц". So removing the translation altogether might not be a great idea. I'll look into making the translation play nice instead.

arvidbrodin:
Wait, I can of course use the chinese translations that were in place before I ran 'make pofiles'. Give me a moment to amend my last commit...

arvidbrodin:
Just for clarity: I believe this change is ready to be accepted now, since my force-push of 2f81b5b.

andypugh:
OK, lets see how it goes.

PR #1197, opened 2021-06-23
Check INI folder for hal files

satiowadahc:
For ease of scripting; check the folder with the INI files for Hal Files

### DIFF
--- a/scripts/linuxcnc.in
+++ b/scripts/linuxcnc.in
@@ -828,6 +828,19 @@ else
                 break
               fi
             done
+            for pathdir in $INI_DIR ; do
+              foundfile=$pathdir/$CFGFILE
+              if [ -r $foundfile ] ; then
+                # use first file found in HALLIB_PATH
+                if [ "${pathdir:0:1}" = "." ] ; then
+                  foundmsg="Found file(REL): $foundfile"
+                else
+                  foundmsg="Found file(lib): $foundfile"
+                fi
+                break
+              fi
+            done
+
           fi
         fi
         [ -d $foundfile ] && foundmsg=""
--- a/scripts/linuxcnc.in
+++ b/scripts/linuxcnc.in
@@ -831,12 +831,7 @@ else
             for pathdir in $INI_DIR ; do
               foundfile=$pathdir/$CFGFILE
               if [ -r $foundfile ] ; then
-                # use first file found in HALLIB_PATH
-                if [ "${pathdir:0:1}" = "." ] ; then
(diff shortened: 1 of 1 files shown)


andypugh:
Can you elaborate on the purpose?

dngarrett:
HALLIB_PATH already includes the '.' directory as the first dir searched:

https://github.com/LinuxCNC/linuxcnc/blob/master/scripts/linuxcnc.in#L53

satiowadahc:
I sometime run linuxcnc from a script 
```bash
# Set up testing for various mills
...

# Need these where ever we call the script
ln -s "$BASE_DIR/linuxcnc/configs/$MACH/$MACH""_postgui.hal" "$SRCDIR/$MACH""_postgui.hal"
ln -s "$BASE_DIR/linuxcnc/configs/$MACH/$MACH.var" "$SRCDIR/$MACH.var"
ln -s "$BASE_DIR/linuxcnc/configs/$MACH/sim.tbl" "$SRCDIR/sim.tbl"

...
  linuxcnc "$LCNC_INIFILE"

```

Trying to set up similar in python.
I can start linuxcnc with 
```
linuxcnc ~/repo/linuxcnc/configs/test.ini
```
but it won't find ~/repo/linuxcnc/configs/test.hal

satiowadahc:
> HALLIB_PATH already includes the '.' directory as the first dir searched:
> 
> https://github.com/LinuxCNC/linuxcnc/blob/master/scripts/linuxcnc.in#L53

Removed the secondary check and updated the string.

satiowadahc:
As I'm sitting here thinking, is there currently a way to add another path like LIB:file.hal? CUSTOMLIB:file.hal?

dngarrett:
proposed method: 724d1265d

satiowadahc:
Can both be an option? I like the idea of being able to add a custom path but also check the same location as the INI file.

dngarrett:
With referenced commit e7cc19c using:
   $ linuxcnc -H somepath/dirname mypath/my.ini,

then [HAL]HALFILE= items will be found by searching 
directories in the order:

1) somepath/dirname
2) mypath -- this is the dir of the INI file (mypath)
3) system halfile library (HALLIB_DIR)

As always, files specifed by [HAL]HALFILE
will obey absolute path specifications or respect
the current directory in the search sequences
for relative paths.

I have no idea what you want if above is not responsive.  A
more complicated search method would require more
documentation and testing.

satiowadahc:
That is what I was thinking.

PR #1200, opened 2021-06-27
Fix changelog when running GitHub actions

havardAasen:
lintian throws errors when the distribution in the changelog is wrong and when the timestamp for the last release is older then the previous. This PR resolves this by creating a new changelog entry during the workflow. This will always use the current time as timestamp. The distribution is parsed from `/etc/os-release`.

It also reverts a previous commit that changed the timestamp for the 1.2.9.0~pre0 release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,8 @@ jobs:
         set -x
         git fetch --recurse-submodules=no https://github.com/linuxcnc/linuxcnc refs/tags/*:refs/tags/*
         ./scripts/travis-install-build-deps.sh
+        codename=$(grep VERSION_CODENAME /etc/os-release | cut -d = -f 2)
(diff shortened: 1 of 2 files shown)


rene-dev:
I believe this has already be fixed another way in master-gtk3 branch, it runs the ci on 20.04 and I remember a similar problem.

havardAasen:
That's partly true. The problem is the timestamp. When there is a new 2.8 release, which it is, the 2.8 changelog will contain a new timestamp. This will almost always be newer then the timestamp we have in the master branch changelog, We then merge 2.8 into master, and lintian will begin to complain.

The easiest solution might be to update the timestamp in master branch, but it needs to be done every time there is a new 2.8 release.

If you merged master into master-gtk3 now. master-gtk3 will also fail.

rene-dev:
Ok, thanks for the explanation.

PR #1201, opened 2021-06-30
pncconf: fix all-limit-home homing

phillc54:
All Limit Plus Home was not writing the correct homing configuration.

### DIFF
--- a/src/emc/usr_intf/pncconf/build_INI.py
+++ b/src/emc/usr_intf/pncconf/build_INI.py
@@ -462,7 +462,7 @@ class INI:
         minlim, maxlim = self.find_limits(letter)
         print("MIN_LIMIT = %s" % minlim, file=file)
         print("MAX_LIMIT = %s" % maxlim, file=file)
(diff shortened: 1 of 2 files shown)


andypugh:
Can you rebase against 2.8 if the fix is needed there?

phillc54:
> Can you rebase against 2.8 if the fix is needed there?

I could but I am not sure of the workflow require to achieve it.

andypugh:
Yes, please. Bug fixes belong in 2,8 and then will be merged up to master.

phillc54:
closing and creating new PR on 2.8 branch.

phillc54:
I have run into an issue that I am not sure how to resolve.

I need to add the  line:
`print("HOME_IS_SHARED = 1", file=file)`
The problem is that in master the print function is used `from __future__ import print_function` but in 2.8 is uses the print statement so the line needs to be:
`print >>file, "HOME_IS_SHARED = 1"`

I guess that I now need to create a separate PR for each branch?

PR #1205, opened 2021-07-09
axis.tcl typo correction

alkabal:
typo error :
Reload tool ta_ble > Reload tool table

### DIFF
--- a/share/axis/tcl/axis.tcl
+++ b/share/axis/tcl/axis.tcl
@@ -87,7 +87,7 @@ setup_menu_accel .menu.file end [_ "Edit _tool table..."]
 
 .menu.file add command \
 	-command reload_tool_table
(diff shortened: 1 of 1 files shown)


dngarrett:
The underline is intended and is used in
many menu entries in axis.tcl.

The -underline option is set by the proc
setup_menu_accel:
Ref: https://github.com/LinuxCNC/linuxcnc/blob/2.8/share/axis/tcl/accel.tcl#L150

Ref: https://www.tcl.tk/man/tcl8.4/TkCmd/menu.html
"If any of the entries in a menu have letters
underlined with the -underline option, then
pressing one of the underlined letters (or its
upper-case or lower-case equivalent) invokes that
entry and unposts the menu."

alkabal:
Ok sorry i will not know that, so i close this useless pr.

Thanks for information

PR #1206, opened 2021-07-10
Patch allowing to use Y axis as a Tailstock for lathe

alkabal:
This patch allow to display or mask Y button and Y Dro from Axis Gui regarding to :

[DISPLAY]
GEOMETRY = xyzc

### DIFF
--- a/src/emc/usr_intf/axis/scripts/axis.py
+++ b/src/emc/usr_intf/axis/scripts/axis.py
@@ -4063,7 +4063,8 @@ if lathe:
     widgets.view_x.pack_forget()
     widgets.view_p.pack_forget()
     widgets.rotate.pack_forget()
-    widgets.axis_y.grid_forget()
+    #patch for display Y used for some tailstock (with a W axis G33.1 G73 G81 G82 g83 does not work fine)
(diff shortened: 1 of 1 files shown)


andypugh:
A comment in the code says "#patch for display Y inside lathe for some tailstock use (with a W axis G33.1 G73 G81 G82 g83 does not work fine)"

Having proved to myself that, yes, you can peck-drill in Y in the G18 plane, I then found that you can peck-drill in W in the G17.1 plane. So what is the problem with the W plane?

alkabal:
Hi

For be honest, i haven't do more test with that after this topic : 
https://www.forum.linuxcnc.org/26-turning/38148-lathe-rigid-tapping-g33-1-with-tailstock-as-w-axis?start=10

The idea for propose a PR was derived from : 
https://www.forum.linuxcnc.org/26-turning/33810-is-a-y-axis-possible-on-a-lathe-settup?start=0

For now my machine was only used for play with code rather than executing some Gcode.

Maybe you are right and the best things to do is use G17.1 and use W axis, this is something i need to test again.

Br

PR #1211, opened 2021-07-19
 Add height override external pin

fupe:
Add external pin in qtplasmac handler for height override
height_override_plus
height_override_minus
height_override_reset


fupe:
make new one

PR #1226, opened 2021-08-10
configure.in: Modernize detection of python & boost

jepler:
Hopefully, this helps with autodetecting the right boost library version. It also enforces the minimum Python version of 3.4.
    
This brings us deeper into the autoconf "ecosystem" so the files config.guess et al had to be added.  Oh well!


jepler:
This is green in Github Actions and Buildbot now.  It's a net growth in code, but much of the addition is maintained by others (the copies of the m4 macros and the files from autoconf) so overall it's a simplification.

src/configure.ac
```diff
@@ -1064,28 +999,19 @@ install with "sudo apt-get install yapps2"])
 fi
 AC_SUBST(YAPPS)
 
+echo A
```
SebKuzminsky, line 1002: Oops.

src/configure.ac
```diff
@@ -1458,61 +1380,48 @@ AC_MSG_CHECKING([whether to check for runtime dependencies])
...
+    else
+        AC_MSG_RESULT(no)
+        AC_MSG_ERROR([BWidget not found!
+    install with "sudo apt-get install bwidget"])
+    fi
 
-AC_MSG_CHECKING([for python pango and cairo modules])
-if $PYTHON -c 'import sys;raise SystemExit(sys.hexversion<0x3040000)'
-then
+    AC_MSG_CHECKING([for BLT using tclsh8.5])
```
SebKuzminsky, line 1392: Shouldn't this use `$TCLSH` too?

SebKuzminsky (changes requested):
Two minor issues, but other than that thanks for the cleanup.

src/configure.ac
jepler: fixed

src/configure.ac
jepler: Modernized.  The reason was hidden deep in the git blame, something about ubuntu hardy :-/

PR #1227, opened 2021-08-10
qtvcp - DetachTabs backwards equality

satiowadahc:
Caution should be taken when copying from stack overflow ;) https://stackoverflow.com/questions/47267195/in-pyqt-is-it-possible-to-detach-tabs-from-a-qtabwidget

The drag event should not happen until after the length of setStartDragDistance is met. Default from system is 10 pixels. on a touch screen typically is about 50-150px.


c-morley:
Thank you

PR #1249, opened 2021-08-27
pncconf: fixes for python3 and board discovery

phillc54:
This seems to work with 7i96 and 7i76e, I have no other boards to test.

### DIFF
--- a/src/emc/usr_intf/pncconf/pncconf.py
+++ b/src/emc/usr_intf/pncconf/pncconf.py
@@ -1135,7 +1135,7 @@ PNCconf will use internal firmware data"%self._p.FIRMDIR),True)
                         #print(tempfunc,founddevice)
                         # this auto selects the sserial 7i76 mode 0 card for sserial 0 and 2
                         # as the 5i25/7i76 uses some of the sserial channels for it's pins.
-                        if boardname in ("5i25","7i92"):
+                        if boardname in ("5i25","7i92",'7i76e'):
                             if "7i77_7i76" in firmname:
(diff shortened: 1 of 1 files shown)


phillc54:
It regularly came back with an error that iocode is None.

c-morley:
Thanks Phill!

phillc54:
One issue I did notice with a 7i76e was sserial channel 0 on i/o connector 1.
This is discovered:
discovered

This is selected from the combobox:
selected

c-morley:
Ya that's one of those hack-it-to-make-it-user-friendly things :) I'll see if i can track down where it does it for 7i76 and see if it's practical to get discovery to do it too.

c-morley:
Looks like around 1138 is the code to auto select.
I bet the strings it looks for are slightly different on discovered boards.
Or else this function is not called at all when using discovery. But it's too late tonight to thread that needle....

phillc54:
Thanks for the hint, this seems to do the trick:
`
                        if boardname in ("5i25","7i92",'7i76e'):
                            if boardname == '7i76e' and 'discovered' in firmname:
                                if tempfunc == "TXDATA0": convertedname = _PD.SS7I76M0
                                else: convertedname = pinconvertsserial[tempfunc]
                            elif "7i77_7i76" in firmname:
`

c-morley:
Glad I could finally help a bit :)
Want me to commit the changes?

phillc54:
Yeah, thanks Chris, that would be good.

PR #1256, opened 2021-09-02
use python3 in all #!python lines

jepler:
Debian has chosen to not install anything at /usr/bin/python, unless python-is-python3 is installed.  pep394 states that "python3" is the dependable name.

Change all #!-lines to invoke `/usr/bin/env python3`. This adapts to the different possible locations of python3, such as /usr/bin or /usr/local/bin; this is the location of env on standard linux systems as well as on freebsd, in the case of the mythical freebsd port of linuxcnc.

Programs that are "installed" from locations in the source directory by make may have their #!-line changed at that time. This PR doesn't change that, so for instance `bin/axis` gets the first line of `#!/usr/bin/python3.9`, the exact value detected at configure time. This is more desirable than the `/usr/bin/env python3`, as it ensures it's the exact version that we built our C extensions for.


rene-dev:
Looks good to me.

PR #1270, opened 2021-09-23
Smallish changes to help Debian build experience

smoe:
Hello,

in cowbuilder the "-o root" for install was not permitted. Since everything will be installed as root already, I added a patch to remove that option.

Took the freedom to add myself as an uploader but please substitute me with anyone who you think may have an interest to upload an update of that package to Debian - yes, without me. There is this "Debian Maintainer" status that grants the permission to upload updates to what is already in Debian, you just need to be mentioned as uploader or maintainer and have a gpg key signed by to Debian Developers.

Otherwise I only got a few lintian warnings.

You have "focal" in debian/changelog, which I suggest to change to "UNRELEASED" (all caps) until the package is uploaded - to focal or unstable or both.

The lintian error
E: linuxcnc source: source-is-missing src/emc/usr_intf/gremlin/__pycache__/gremlin-runcpython-39.pyc
is apparently a symptom of something not cleaned up properly. Maybe you have an idea where to look.
 
I changed the date in d/changelog to a version that is younger than the new standard 4.6.0. 

In Debian there is a craze ongoing to change the good old which to "command -v" because of too many versions of which being around. So the current which gives a warning that I have prevented by asking command first.

In d/control there are some conventions like having the substitutions first. The build-depends are a bit excessive. I removed iptables and have some confidence that some more can go.

But congrats, cowbuilding and mostly lintian-clean, this is already darn good!
Steffen


rene-dev:
Thanks!
CI fails to install build deps, any idea why?

smoe:
Hm. This is unexpected. It cannot install the build-deps. But cowbuilder just worked. I nicified the d/control file such that there are npw lines that only show a comma (substitutions with nothing). Deep hunch. I am testing a respective "sed -e '/^\s*,\s*$/d' debian/control" locally that should be integrateable into configure.

smoe:
That may be a bit weird now, but if I get this right then this "install -o root" should go. But for regular installs from the source tree this is still valid. This has the effect that for the Debian package you now patch your own source tree. I now changed debian/source/format from native to quilt. My patch to undo the root should then be applied. We'll see.

smoe:
This turned out to be quite a sport, so I must admit. I'll get there and then we can discuss a rebase.

smoe:
Here we are. Looking back, it was likely the "Rules-Requires-Root: no" that demanded the patch and then I learned the hard way that a native package does not apply patches. Which makes some sense. I am not sure that you want to follow my substitution of debuild with ./debian/rules binary. But then again, the overall robustness should have improved and errors are not hidden that much.

jepler (approved):
Thank you!

jepler:
Looks like the change to `quilt` causes problems with our buildbot CI system:
```
 dpkg-source -I -b .
dpkg-source: error: can't build with source format '3.0 (quilt)': no upstream tarball found at ../linuxcnc_2.9.0~pre0.4651.g8a7f0deae.orig.tar.{bz2,gz,lzma,xz}
dpkg-buildpackage: error: dpkg-source -I -b . subprocess returned exit status 25
```

@SebKuzminsky do you have the time to look into this in the short term, or do we need to revert the change to quilt or find another way to do it, so that we can continue providing development version packages via buildbot?

smoe:
I had changed the debuild (which invokes dpkg-buildpackage) to "fakeroot ./debian/rules binary" since you only install the .deb files and do not work with the others (like .changes and .dsc). From what I understand one can also instruct the dpkg-source to build the source tarball but unless you directly feed your linuxcnc repository from the CI-generated files maybe a simple execution of the good ole "binary" target of d/rules is all you need. From what I recall, that change helped me with better error messages.

smoe:
After some time invested in the dpkg-source manpage that went nowhere, I came up with the following:
```
p=$(head -n 1 debian/changelog |cut -f1 -d' ')
v=$(head -n 1 debian/changelog |cut -f1 -d\) | cut -f2 -d\( )
d=$(basename $(pwd))
(cd .. && tar --exclude debian --exclude .git --exclude .github -czf ${p}_${v}.orig.tar.gz $d)
```
Still, I hope that you just do not need to run debuild but fall back to "fakeroot ./debian/rules binary". The above is annoying.

SebKuzminsky:
@jepler Leave it in master for now, I'll take a look at the buildbot...

Issue #1272, opened 2021-09-24
kernel_lockdown block iopl call


safarir:
## Here are the steps I follow to reproduce the issue:

run any rtapi_app on a system with secure boot

## This is what I expected to happen:

Get the latency test going

## This is what happened instead:

```cannot gain I/O privileges - forgot 'sudo make setuid'?```

and program exit

## It worked properly before this:

I change my hardware, it work fine on my older pc

## Information about my hardware and software:

Hardware: HP z240, i5-7500, 8GB ram, NVME SSD
Latest debian buster live cd `linuxcnc-2.8.2-buster.iso`

## Comment

The behavior is controlled by `src/rtapi/uspace_rtapi_app.cc`:

```
#if defined(__linux__) && (defined(__x86_64__) || defined(__i386__))
    if (iopl(3) < 0) {
        rtapi_print_msg(RTAPI_MSG_ERR,
                        "cannot gain I/O privileges - "
                        "forgot 'sudo make setuid'?\n");
        return -EPERM;
    }
#endif
```

This code seem to try to be there to allow access to the parallel port. However, with modern kernel and secure boot (kernel_lockdown), iopl is blocked, see reference 2.

```
When lockdown is in effect, a number of features are disabled or have their use restricted
[...]
The use of the ioperm and iopl instructions on x86.
```

I ran a small test c program, and ran it as root. iopl(3) fail in all case on my system. Once disabling secure boot, the iopl call work.

The current error message can be really misleading when this situation. Like the use in reference 1, I was looking at my setuid for a long time before realizing the problem.

Since using parallel port is not a requirement of linuxcnc, and less and less machine come without serial port, I recommend simply printing an warning message. This would warn people trying to use serial port, while allowing people with modern kernel and hardware to run linuxcnc without issue.

# References

1. https://forum.linuxcnc.org/9-installing-linuxcnc/41261-halrun-error-cannot-gain-i-o-privileges-error
2. https://www.mankier.com/7/kernel_lockdown#

PR #1273, opened 2021-09-24
Fix iopl issue with kernel_lockdown (secure boot)

safarir:
This is my quick fix for issue #1272 

Please let me know if you have a better idea

Thank you


jepler:
Thanks. The ci failed, but it was unclear why so I restarted it.

rene-dev:
@jepler reasin it failed: https://github.com/actions/virtual-environments/issues/3287

safarir:
Quick question, how will this fix be apply on the 2.7 and 2.9 branch. Is there some automatic merging happening somehow ? I could not find any instructions ...

rene-dev:
It will be merged into master. I don't think 2.7 still receives updates.

PR #1275, opened 2021-09-25
Fix "object has no attribute 'tolist'"

jepler:
This occurs since the switch to pyopengl (yay!) but only if python3-numpy is not also installed.

I was able to eliminate the conversion of the matrix to its 1D / list form entirely.

I think I got pretty good manual testing coverage of everything _except_ qt vismach in my testing, both with and without python3-numpy installed.  qt vismach should be the same as tk vismach, as I made edits with the same intended effect.

### DIFF
--- a/lib/python/qtvcp/lib/qt_vismach/primitives.py
+++ b/lib/python/qtvcp/lib/qt_vismach/primitives.py
@@ -1,4 +1,4 @@
-import itertools
+import copy
 import OpenGL.GL as GL
 from OpenGL import GLU
 import hal
@@ -164,9 +164,9 @@ class Track(Collection):
 
         view2world = invert(self.world2view.t)
 
-        px, py, pz = self.position.t[12:15]
+        px, py, pz = self.position.t[3][:3]
         px, py, pz = self.map_coords(px, py, pz, view2world)
-        tx, ty, tz = self.target.t[12:15]
+        tx, ty, tz = self.target.t[3][:3]
         tx, ty, tz = self.map_coords(tx, ty, tz, view2world)
         dx = tx - px;
         dy = ty - py;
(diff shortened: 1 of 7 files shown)


lib/python/qtvcp/lib/qt_vismach/primitives.py
```diff
@@ -654,8 +654,7 @@ def capture(self):
         # give a list of lists.
         # It would be better to change all the functions so a flat list could be used 
         # directly - i don't understand the code well enough
-        # self.t = GL.glGetDoublev(GL.GL_MODELVIEW_MATRIX)
-        self.t = list(itertools.chain.from_iterable(GL.glGetDoublev(GL.GL_MODELVIEW_MATRIX)))
+        self.t = glGetDoublev(GL_MODELVIEW_MATRIX))
```
rene-dev, line 657: extra brace here causes the test to fail

rene-dev:
looks good to me, thanks for looking into the issue

jepler:
I'll correct the error soon, it looks trivial

jepler:
I have a report on irc from @pcw-mesa that this didn't fix the problem, but I still think it's an improvement. (and don't see a problem here, running sim/axis/axis.ini and sim/axis/vismach/5axis/max5/max5kins.ini.  If you're still having a problem after this change, please report whether
 - installing python3-numpy makes a difference (I have no problem making this a hard dependency of linuxcnc)
 - the full traceback of where the error is still occurring
 - any steps you have to take before it occurs (my testing is just panning & rotating the view in axis, and running gcode to watch the backplot in vismach)

Feel free to report it here or in a fresh issue. If you mention me in a fresh issue I'll be more likely to see it.

PR #1278, opened 2021-09-26
Fix (and silence) a bunch of build-time diagnostics ("warnings")

jepler:


src/libnml/cms/cms_cfg.cc
```diff
@@ -305,7 +305,7 @@ int cms_copy(CMS ** dest, CMS * src, int set_to_server, int set_to_master)
 extern char *get_buffer_line(const char *bufname, const char *filename)
 {
     int line_len, line_number;
-    char linebuf[CMS_CONFIG_LINELEN];	/* Temporary buffer for line from
+    static char linebuf[CMS_CONFIG_LINELEN];	/* Temporary buffer for line from
```
petterreinholdtsen, line 308: Why do you make this variable static?  Why should its content survive from call to call?

jepler: Happily, this function is deleted in the current version!

Originally, a pointer within `linebuf` was returned from the function, which is undefined behavior since the storage scope of `linebuf` ends when the function returns:
```c
    char linebuf[LINELEN];	/* Temporary buffer for line from
					   file. */
    char *line = linebuf;
    // …
   return line;
```
By making it static, the storage remains valid for the lifetime of the program, though it is over-written by a subsequent call to `get_buffer_line`, so some care is still required in its use.

Removal: 35659250289e74ee15344885cba3ff91cb9b2dde (thanks @dwrobel !)

petterreinholdtsen:
Aha.  Then it is perhaps time to rebase the pull request on master and
see if we can get it merged before it is too late this time?

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
The *QUIETFLAGS Makefile variables seem scary, hiding coding errors that could cause security or stability issues.  Perhaps we should make an effort to get completely rid of them?

src/libnml/inifile/inifile.cc
```diff
@@ -315,9 +315,8 @@ IniFile::Find(const char *_tag, const char *_section, int _num, int *lineno)
            continue; // get next line to extend
         } else {
             if (extend_ct) {
-               strncpy(elinenext,line,newLinePos);
+                snprintf(elinenext, newLinePos, "%.*s", newLinePos, line);
```
petterreinholdtsen, line 318: Using newLinePos twice here seem redundant, as both uses will limit the string length.  Why did you both use snprintf() to limit the length, and %.*s to do the same?

petterreinholdtsen:
Note, I tried addressing one of the warnings in #1922 by completely rewriting the print methods to avoid statically allocated buffers.

src/emc/usr_intf/schedrmt.cc
```diff
@@ -1254,12 +1254,12 @@ int main(int argc, char *argv[])
     // process local command line args
     while((opt = getopt_long(argc, argv, "e:n:p:s:w:", longopts, NULL)) != -1) {
       switch(opt) {
-        case 'e': strncpy(enablePWD, optarg, strlen(optarg) + 1); break;
-        case 'n': strncpy(serverName, optarg, strlen(optarg) + 1); break;
+        case 'e': snprintf(enablePWD, sizeof(defaultPath), "%s", optarg); break;
+        case 'n': snprintf(serverName, sizeof(defaultPath), "%s", optarg); break;
```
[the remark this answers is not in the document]
jepler: fixed (and below)

src/libnml/cms/cms_cfg.cc
jepler: Dropped on rebase.

src/libnml/inifile/inifile.cc
jepler: dropped during rebase

jepler:
I've heavily revised this PR and would appreciate re-review @SebKuzminsky @petterreinholdtsen

src/emc/kinematics/Submakefile
```diff
@@ -27,7 +27,7 @@ TARGETS += ../bin/genserkins
 
 RDELTAMODULESRCS := emc/kinematics/rotarydeltakins.cc
 PYSRCS += $(RDELTAMODULESRCS)
-$(call TOOBJS, $(RDELTAMODULESRCS)): CFLAGS += -x c++ -Wno-declaration-after-statement
+$(call TOOBJS, $(RDELTAMODULESRCS)): EXTRAFLAGS += -DBOOST_ALLOW_DEPRECATED_HEADERS=1 -DBOOST_BIND_GLOBAL_PLACEHOLDERS=1
```
petterreinholdtsen, line 30: Hiding warnings about using obsolete boost features seem to me like a disaster waiting to happen.  At the moment we get a warning about a problem we need to solve sooner or later.  With the warnings gone, the incentive to fix it drop to almost non-existing.

src/libnml/rcs/rcs_print.cc
```diff
@@ -47,7 +47,7 @@ long rcs_print_mode_flags = PRINT_RCS_ERRORS;
 FILE *rcs_print_file_stream = NULL;
 char rcs_print_file_name[80] = "rcs_out.txt";
 
-char last_error_bufs[4][100];
+char last_error_bufs[4][256];
```
petterreinholdtsen, line 50: Perhaps 256 here should be replaced with error_buf_size?

petterreinholdtsen (commented):
The code changes look good, more unsure about the build option changes, as some of them only hide the warnings, but do not solve the issues.

src/emc/kinematics/Submakefile
jepler: I'd love to "fix" this.  However, it is a diagnostic that occurs whenever the header <boost/python.hpp> is used, as shipped by Debian!
```
$ gcc -c -x c++ -include boost/python.hpp /dev/null -I /usr/include/python3.9
In file included from /usr/include/boost/smart_ptr/detail/sp_thread_sleep.hpp:22,
                 from /usr/include/boost/smart_ptr/detail/yield_k.hpp:23,
                 from /usr/include/boost/smart_ptr/detail/spinlock_gcc_atomic.hpp:14,
                 from /usr/include/boost/smart_ptr/detail/spinlock.hpp:42,
                 from /usr/include/boost/smart_ptr/detail/spinlock_pool.hpp:25,
                 from /usr/include/boost/smart_ptr/shared_ptr.hpp:29,
                 from /usr/include/boost/shared_ptr.hpp:17,
                 from /usr/include/boost/python/converter/shared_ptr_to_python.hpp:12,
                 from /usr/include/boost/python/converter/arg_to_python.hpp:15,
                 from /usr/include/boost/python/call.hpp:15,
                 from /usr/include/boost/python/object_core.hpp:14,
                 from /usr/include/boost/python/args.hpp:22,
                 from /usr/include/boost/python.hpp:11,
                 from <command-line>:
/usr/include/boost/bind.hpp:36:1: note: ‘#pragma message: The practice of declaring the Bind placeholders (_1, _2, ...) in the global namespace is deprecated. Please use <boost/bind/bind.hpp> + using namespace boost::placeholders, or define BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior.’
   36 | BOOST_PRAGMA_MESSAGE(
      | ^~~~~~~~~~~~~~~~~~~~
/usr/include/boost/detail/iterator.hpp:13:1: note: ‘#pragma message: This header is deprecated. Use <iterator> instead.’
   13 | BOOST_HEADER_DEPRECATED("<iterator>")
      | ^~~~~~~~~~~~~~~~~~~~~~~
```+

src/libnml/rcs/rcs_print.cc
jepler: good catch, thanks.

src/emc/kinematics/Submakefile
jepler: however I can package it up and write an explanatory comment .. done

PR #1284, opened 2021-09-28
fix latency-test for no-base-thread case

jepler:
I noticed that none of the doc-suggested ways to run latency-test without a base thread actually worked.  This makes `latency-test 1ms -` work.  The problem was that the required xml file was only being written in the 'else' branch.


smoe (changes requested):
In line https://github.com/LinuxCNC/linuxcnc/blob/65638cc2dd59a93f7152701f747275586c085bfa/scripts/latency-test#L154 creates latexit.sh again.

smoe:
@hansu - This is one of the bitrotting gems we have among our PRs. If you find the minutes, please have a look, also at my patch to Jeff's patch and get this done. It is the very prominent latency-test script, so nothing mission critical and nothing too complicated.

smoe:
@andypugh - many thanks for the merge, but please also inspect the change I had done as a PR to @jepler 's branch on https://github.com/LinuxCNC/linuxcnc/pull/1587 .

hansu:
@smoe I looked into it yesterday but I couldn't find any issue with the existing version on my VM. I had to check it on the real system.

smoe:
It works, just a file is created twice and my initial patch was bad :-/

PR #1287, opened 2021-10-03
Better formatting

rgbond:
Clean up the formatting in this if statement.


smoe (approved):

jepler (approved):
Thanks, I think it's correct, and the only reason I had trouble verifying it was how confusing the old code was.  @rgbond are you submitting this PR as a part of hacktoberfest?

rgbond:
Thanks Jeff. No, not for hacktoberfest.

 

I found it while I was going through the code. I was convinced that the code was wrong before I finally figured it out. Just trying to save others the same confusion.

 

Bob

PR #1288, opened 2021-10-04
Update control.top.in

smoe:
Hello,

I tried to get my head around what LinuxCNC is and what it is not and how it relates to other parts of the Open Source world that I had heard about. This PR is just a request-for-discussion and, frankly, I wish you would adopt a few ideas from whatever this package description now evolves into for your home page. A keyword that description should still see is the Hardware Abstraction Layer. Please kindly "nativate" my write-up and squeeze HAL in somewhere :)

Best,
Steffen

### DIFF
--- a/debian/control.main-pkg.in
+++ b/debian/control.main-pkg.in
@@ -26,7 +26,49 @@ Depends: ${shlibs:Depends}, @KERNEL_DEPENDS@,
     tclreadline, procps, psmisc, tclx, @MODUTILS_DEPENDS@,
     mesa-utils, blt, udev
 Description: motion controller for CNC machines and robots
- LinuxCNC is the next-generation Enhanced Machine Controller which
- provides motion control for CNC machine tools and robotic
- applications (milling, cutting, routing, etc.).
+ LinuxCNC is a fully-realised CNC machine controller that can interpret
+ machine-control programs (such as G code), plan trajectories and finally
+ output low-level signals to machine control hardware
+ .
+ LinuxCNC is modular and in principle the HAL (hardware abstraction layer),
+ pluggable interpreters and variety of hardware drivers mean that it can
+ control almost any software-defined physical system.
+ .
+ LinuxCNC relies on a realtime kernel to support real-time motion control,
+ typically updating the position waypoints every 1mS and reacting to
+ input within tens of microseconds.
+ .
+ Modules exist to use the low-level parts of LinuxCNC as a motion-control
+ client for OpenPnP. Interfaces exist for ROS (the Robot Operating System)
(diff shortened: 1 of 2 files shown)


andypugh:
Let me try:

"LinuxCNC is a fully-realised CNC machine controller that can interpret G-code (but is extendable to parse other grammars such as STEP-NC or Gerber), plan trajectories and finally output low-level signals to machine control hardware such as step/dir on a variety of interface hardware, Modbus, EtherCAT, CAN and Mesa "smart-serial"

LinuxCNC is modular and in principle the HAL (hardware abstraction layer), pluggable interpreters and variety of hardware drivers mean that it can control almost any software-defined physical system. 

LinuxCNC relies on a realtime kernel to support real-time motion control. It could be configured to support buffered non-realtime motion control hardware but that would be missing the point. Other options exist in that space. 

Modules exist to use the low-level parts of LinuxCNC as a motion-control client for OpenPnP.  Interfaces exist for ROS {detail needed, I know little about this} with which LinuxCNC has considerable overlap. 

LinuxCNC supports arbitrary kinematics (the mapping between coordinate space and actuator space) and can control both parallel (eg Stewart platform) and serial (eg robot arm) systems in addition to conventional cartesian machine tools. 

LinuxCNC supports 9 cartesian axes and 16 "Joints" (individual actuators) though the joint limit is build-time configurable. 

The HAL layer supports a wide variety of motion control hardware. Typically the input data is position commands for all joints at a fixed clock rate, but commands can also be supplied from user-space and over network connections. 

In summary, if you want a computer to control things in the real world then LinuxCNC can do it. It's not always the best choice,   as you move away from machine tools and robots it becomes a less-good fit, but it can do it."

smoe:
Many thanks! I like many bits of it a real lot. What I would like to hope for it would be
 * an escalation in the level of detail (from rough sketches for the causal reader who is not really interested, then the one who wants to control his coffee machine (and should learn that LinuxCNC could boil it at the nanosecond of accuracy), and then concluding at some expert level)
 * allow a read without technical terms whenever possible (explanations become main text, not hidden in long parentheses like this one)

> In summary, if you want a computer to control things in the real world then LinuxCNC can do it. It's not always the best choice, as you move away from machine tools and robots it becomes a less-good fit, but it can do it."
Maybe there is a positive wording, maybe pointing to alternatives that are better suited for problems with less RT demands or a lower complexity or ...

smoe:
Put it as the description for the main  and the -dev package.

petterreinholdtsen:
I like the updated description a lot better than the original, and hope it will be included into master.

SebKuzminsky:
Thanks folks!

PR #1290, opened 2021-10-06
motion - Adding Misc Error

satiowadahc:
I had a need for a fault pin on a custom spindle hal component, and didn't want to load a full Spindle for one fault pin. This enables motmod to load up to 64 error pins, that can act as a fault for what ever purpose you might need.

usage:
```
loadrt motmod  num_misc_error=1
net myfaultnet  myfaultpin => motion.misc-error-00

```


satiowadahc:
@andypugh This is ready for review again.

PR #1291, opened 2021-10-08
Introducing d/watch, d/TODO and d/README.source

smoe:
The README.source describes the steps I am about to take to upload LinuxCNC to Debian. Please comment if you prefer me to do this somehow differently.

PR #1293, opened 2021-10-09
<strike>Not</strike> installing realtime init file for uspace

smoe:
The rm in d/rules was easiest for me. I presume the install routine should also be more selective about when to copy that file but I failed to find that.

This change removes about three linitan warnings in PR #https://github.com/LinuxCNC/linuxcnc/pull/1291 .

### DIFF
--- a/scripts/realtime.in
+++ b/scripts/realtime.in
@@ -1,4 +1,17 @@
 #!/bin/bash
+### BEGIN INIT INFO
+# Provides:          realtime
+# Required-Start:    $remote_fs $syslog
+# Required-Stop:     $remote_fs $syslog
+# Default-Start:
+# Default-Stop:      0 1 6
+# Short-Description: enable realtime capabilities when needed
+# Description:       The rtai module is loaded when needed and
+#  removed when the job has ended and is used throughout the LinuxCNC
+#  infrastructure. The script has no effect for the PREEMPT realtime kernel.
+#  The script is not meant to be started a boot time - it could be, though.
+# X-Interactive:     false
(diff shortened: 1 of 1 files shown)


smoe:
Sigh. The removal worked, but halrun is expecting that init file:
```
/usr/bin/halrun: line 83: /etc/init.d/realtime: No such file or directory
/usr/bin/halrun: line 142: /etc/init.d/realtime: No such file or directory
```
line 83 of halrun:
```
if $REALTIME status > /dev/null; then
    echo "$THIS: Realtime already running.  Use 'halrun -U' to stop existing realtime session." 1>&2
    exit 1
fi
```
and the other invocation in line 142 is
```
$REALTIME start || exit $?
```
I never worked with the rtai module or the preempt kernel. The easiest fix would be to just place a dummy script. But you may have a better idea?

smoe:
Ok, I eventually did a "grep -r REALTIME ." and - it is everywhere. So, I need to rethink this. From how I see this now, linuxcnc declares when it has realtime requirements and also stops them when these are no longer required. While I sense a problem with that approach as soon as you control two robots on the same machine, I still like it - should be turned into a semaphore, though.

smoe:
This is the kind of a problem that is best solved by walking a dog.  I "[currently](https://www.wordsense.eu/currere/)" like the idea to understand it as an init script. It should just
 * not be auto-initiated at boot time and we
 * need something for systemd (maybe not immediately)
 * it shall to be explicit about being useful only if the rtai module is used
I think we can get there, dog-time now :)

PR #1294, opened 2021-10-09
The patch is no longer needed

smoe:
The patch itself is missing, so it does not build. And it does not seem to be required anyway.
This also eliminates the quilt lintian warning in PR #1291 .

### DIFF
--- a/debian/patches/series
+++ b/debian/patches/series
@@ -1,1 +1,0 @@
-No_root_install.patch


SebKuzminsky:
I pushed this branch to our main repo so our buildbot would test it.  Package building on Buster (the only platform that's both supported by `master` and represented in the buildbot), worked fine:
http://buildbot.linuxcnc.org/buildbot/builders/4041.deb-buster-rtpreempt-amd64/builds/1261
http://buildbot.linuxcnc.org/buildbot/builders/4042.deb-buster-rtpreempt-rpi4/builds/1153

So I'm merging it.

Thanks @smoe :-)

Issue #30, opened 2016-01-22
G53 doesn't work right with G92 and coordinate system rotation


jepler:
Tested in 2.6.11 and master-ish with configs/sim/axis:
1. Start AXIS.  Power on.  Home all.  Clear G54, G92 offsets via menu
2. MDI: G54
3. MDI: G0 G53 X0 Y0
4. MDI: G92 X1 / G10 L2 P1 R45 / G0 G53 X0 Y0

Expected behavior: no motion at step 4, because the machine was already at G53 X0 Y0 after step 3
Actual behavior: motion to machine coordinates (.2929, -.7071)

jepler:
Instead of G0 G53 X0 Y0, you can also G91 G0 X0 Y0, which should cause motion since it's in incremental distance mode with all distances zero.  It takes you to the same machine location as the G0 G53 version.
![rotation-g92-wrongmotion](https://cloud.githubusercontent.com/assets/1517291/12522349/c18dfc1e-c115-11e5-9506-8cf8d0f61440.png)

jepler:
Thanks!  The best way to get a change accepted into linuxcnc is by preparing a pull request. Patches added to issues are likely to get lost.

andypugh:
Seems to be fixed by #1297

PR #1297, opened 2021-10-10
Address interpreter rotation issues

rgbond:
Changes to address G10 Rx xy rotation issues.
- Fix inconsistent application of G92 offsets in rotated systems.
  The G92 offset should always be applied after the rotation. In
  some cases it was being done at the same time as the G5x offsets
  and the tool offsets:
  convert_stop()
  find_ends(), G53 case. This fixes issue #30
  find_relative()
- Add rotation support to G28.1 and G30.1 in convert_savehome()
- Added test cases for the G53 and G28 changes


andypugh:
Gets my vote, but let's see if cradek has any insight.

rgbond:
Ok. But note that g53 doesn’t handle rotation correctly either. See issue #30.

andypugh:
@yamyamar Are you able to try out the changes in this PR to see if they help your situation?

yamyamar:
Andy,

I see that there are 13 files changed.
But how do I implement them into my linuxcnc?
Do I need to compile from source, or can I just replace those 13 files or some other way...?

Please advice as I do not have a clue how to implement these changes into my installation.

Thanks

Jan

andypugh:
@yamyamar At the present time the only way to test these changes would be to build LinuxCNC from source. 
Typically you would do this as a run-in-place install that can live alobgside the standard install and share the same config files.

rene-dev:
you need to checkout his repo: https://github.com/rgbond/linuxcnc/tree/rotations
and then checkout the branch where the changes are in:
`git checkout rotations`

rgbond:
I would clone the current tree then download the patches and install them with the Linux “patch” command. There may be better ways but I don’t know them.

Bob

yamyamar:
Thanks for your help.

The ckeckout and clone commands resulted in bad repositorys? 
But I have downloaded the Zip file provided under the source.
Unzipped it gave a full directory structure under linuxcnc-rotations. 
So I believe this will be good to start from.

Following the manual I have checked the deps, as a result an error; See pic.
But before it tells that uspace is ignored? 

If  I try just to install it complaints for a pkg-config error. see picture

My system is the standard system provided by linuxcnc 2.8.2-11
I did an apt update/upgrade.

Any ideas?
Thanks in advance.

Jan

rene-dev:
You are missing the build dependencies, install all the packages it tells you to install.

yamyamar:
All,
Compiled linuxcnc from source without any further big issues.
Didn't showed any errors at the end so I assume it went ok.

Now the 10 mlj question? How do we start llinuxcnc from this new install?
Original startup file is found in the /bin/ linuxcnc and file is not present in the local bin dir?
So executing linuxcnc in a shell is apparently not the way?

Thanks,
Jan

yamyamar:
All,

I found in the script folder linuxcnc as an executable. So my question above becomes obsolete.

Executing this file gives the opportunity to choose my existing config file.
But it stops on an  error on "opengl module"?

Choosing axis simulation or other ini files, all give the give the same error?

see down:

Did I missed something in the compilation?

Thanks for your reply.
Jan

LINUXCNC - 2.9.0~pre0
Machine configuration directory is '/home/cnc/linuxcnc/configs/LinuxCNC_machine'
Machine configuration file is 'LinuxCNC_machine.ini'
Starting LinuxCNC...
Found file(REL): ./LinuxCNC_machine.hal
Note: Using POSIX realtime
...
hm2_5i25.0: initialized AnyIO board at 0000:05:00.0
Found file(REL): ./handwheel.hal
Found file(REL): ./custom.hal
  File "/home/cnc/linuxcnc-rotations/bin/axis", line 24, in <module>
    from OpenGL.GL import *
ModuleNotFoundError: No module named 'OpenGL'
Running HAL shutdown script
task: 10 cycles, min=0.000012, max=0.010075, avg=0.006526, 0 latency excursions (> 10x expected cycle time of 0.010000s)
hm2_5i25.0: dropping AnyIO board at 0000:05:00.0
hm2/hm2_5i25.0: unregistered
RTAPI_PCI: Unmapped 65536 bytes at 0x7fb5505f0000
hm2_pci: driver unloaded
hm2: unloading
Note: Using POSIX realtime
    /home/cnc/linuxcnc_debug.txt
and
    /home/cnc/linuxcnc_print.txt


------------------
(program exited with code: 1)
Press return to continue


the log give:
24971
25006
Stopping realtime threads
Unloading hal components

andypugh:
Try 
sudo apt-get install python3-opengl

yamyamar:
Thank you all for the support.
We are finally in business;-)

Besides all the dep's given by dpkg-checkbuilddeps  the following libs also need to be installed in a manual way:

python3-gi
python3-cairo
python3-gi-cairo
python3-opengl, 
python3-configobj,

Jan

yamyamar:
All,

i tried out the adaptations but with no result.
I include a small program so if interested you can try out for yourself.
It's not written to be used at a machine, units are in mm rename extension to ngc.

As before I do not understand that when, after given a rotation of 45°, a new rotation of -45° will not result in  the original coordinate system.

Is it an option to have a rotation as G53?

Let me know if I can help in any way.
Jan
[1111.txt](https://github.com/LinuxCNC/linuxcnc/files/7527872/1111.txt)

yamyamar:
As I was playing around I changed the translations of G52 into G54.
File under.

Now it works as expected, so we can isolate the problem to using G52.







[1111.txt](https://github.com/LinuxCNC/linuxcnc/files/7527971/1111.txt)

JAn

yamyamar:
As I believe G52 and G92 has to perform in the same way.
This is what I found:

G54 this is the wanted behavior, take attention to the dashed lines.
G54

G52 acts like: beforeG52, set G54 to X100 Y100
G52

G92 acts like: before G92, set G54 to X100 Y100
G92


To put rotation into a Gcode I meant G68 instead of G53 as written above.

Jan

rgbond:
Hi Jan,

 

Thanks for the example files. I spent some time working with them and found a couple of things:

1)      I suggest that you add these two lines to the top of the file after the T1 command.

a.       G92.1

b.      G10 L2 P0 X0 Y0 Z0 R0

I had trouble debugging the files because the previous G92 and G54 state is preserved between runs

2)      Change “G10 L2 P0 R-45” to “G10 L2 P0 R0” With those two changes I was able to make both test files work (I think) with my patches.

 

Let me know how these changes work for you.

Bob

yamyamar:
Bob,

Thank you for your reaction.

Initiating all translations to 0 do not do the trick in my opinion
Mine were already on zero, I checked the var file for this.
On the screen you can actually see that they are all 0.
 

G52-2
The remark to change R-45 to 0 is in my opinion doubtfull as G10 L2 P0 refers to the coordinates of G52 which are relative coordinates.  But this is open for interpretation.
 
I include a test were the subprog is effectuated 2 times as follows;
G52
R45
o100
R0
o100
And it proves you are right. 

G52-3

I stay with the impression that a rotation of the original coordinate system and not taking into account further relative movements is not correct.  I think that the program works like this:  when encountering a rotation the program looks for the absolute coordinate system and effectuates the rotation regardless if a G52  was already active? (maybe different for G92?)
In my experience coordinate translations and rotations are added 1 after another until an absolute new coordinate system is encountered.
Why wouldn't it be possible to have something like: 
G54 Rot45 G52 Rot78 G52 Rot0... 

Secondly I do not understand why a G52 acts differently than & G92  as showed above.

Regards,
Jan

rgbond:
Hi Jan,

When a new g10 command is issued for the current g5x (say g54) system, or a different g5x system is commanded (say g55), the code takes these steps in this order:

1 removes the current g52/g92 translation, 
2 removes the current rotation, 
3 removes the current g5x offsets,
4 sets up the new g5x system, 
5 rotates the system if needed, 
6 reapplies the g52/g92 translation. 

So if the rotation changed the g52/g92 translation is now rotated from the previous direction.

My patches are primarily aimed at cleaning up corner cases in this scheme.

I don’t understand the differences you see between g52 and g92. Please explain a bit more. The code paths are pretty much identical except for the differences in persistence between runs.

Bob

yamyamar:
Bob,

Thank you for your explanation.

As far as I can see the program is reacting like you describe.

1 removes the current g52/g92 translation,
2 removes the current rotation,
3 removes the current g5x offsets,
4 sets up the new g5x system,
5 rotates the system if needed,
6 reapplies the g52/g92 translation.

In essence the buildup coordinate system is pulled down till machine coordinates and then build up again.

The linuxcnc description of G10 (https://linuxcnc.org/docs/html/gcode/g-code.html#gcode:g10-l2)
describes this behavior as you write it down

For me this is not always correct.
Let me try to explain why I'm not convinced in this general way of handling G10.

These are the points I believe are not correct in the manual:

General def of G10 : 
G10 L2 P- <axis R-> 

is in fact G10 L2 P- X Y Z R  
P from 0-10 
R on XY plane origin 0,0 X+ direction
Only Absolute (G91) coordinates for P1-P6 (see down)


And we note that G10 L2 P0 speaks about the **active coordinate system**.
I believe we need to distinguish P0 from the other (see below) 

In the manual:

    G10 L2 Pn does not change from the current coordinate system to the one specified by P, you have to use G54-59.3 to select a coordinate system.
G10 L2 P0 Rx is directly effectuated?  (P0 is different as it is always active, see manual)

    When a rotation is in effect jogging an axis will only move that axis in a positive or negative direction and not along the rotated axis.
This is a missed change, with a probe we can say that the piece on the machine is straight as it lays down. 
As a user we do not care if the machine needs to activate 2 motors to follow 1 axis.  But the setup of the machined part becomes very easy if we can use rotation for this way of working.
  
    If a G52 local offset or G92 origin offset was in effect before G10 L2, it will continue to be in effect afterwards.
This is useless to me when we install a new coordinate system from G54 to G55. But useful when we adapt a coordinate system from G54 to G54 
 
    When programming a coordinate system with R, any G52 or G92 will be applied after the rotation.
Again, this is useless to me when we install a new coordinate system from G54 to G55 we need to forget everything different then what we ask for. But useful when we adapt a coordinate system from G54 to G54 

    


Let me try to explain why I believe P0 needs a different treatment then P1-6

All coordinate system are setup with
G10 L2  

	P1-P6  XYZR (+ G52)  

On this coordinate system we can effectuate the following commands:

New command coordinate system, different then the actual G54/9. So G54/G59 with or without Rotation is installed.
	All commands are previously setup with G10 and wait before coming active on a G54/9 Command, (unless they were already active) 
(For me: everything is forgotten, and the new coordinate system is installed as demanded, I speak about a new start, so all G52 are not to be installed again, 
the coordinate system is absolute G90 and follows the G10 command)

Or we issue the same coordinate system. so for example G54 is active.
G10 L2 P1 XYZR,  the actual values that were not defined come active the ones that were active are overwritten. In this situation it makes sense to reinstall the relative movements of G52 
after the adaptation of the existing coordinate system

Or we issue P0 XYZR 
My interpretation is that we start from the actual coordinate system and that all coordinates are relative (G91).
This allows to setup the most complicated coordinate systems and to leave the control to the user.
Take into account that its possible to have non of the G54/9 active, but only G52, Would it be good behavior when the rotation now comes before the translation?



After writing this down I came to the conclusion that P0 needs to be treated as relative (G91)
And P1-P6 as absolute (G90), but with the possibility to adapt (not change). 
And if you allow adaptation of the latter, why not allow a G91 change so you get something as G10 L2 P1 G91 X6, as it moves the existing center 6 to the right, HAAs machines are allowed to do this).

Further if we can implement a G68 as a rotational Gcode we can
overcome the above as the rotation takes place the moment G68 is activated (G91 behavior)
Overcome the limit of the XY plane
Overcome the limit to rotate around 0,0

Share me your comments.

I'll come back for the difference I noticed between G52/92

Regards,

Jan

rgbond:
>> As far as I can see the program is reacting like you describe.
...
>>In essence the buildup coordinate system is pulled down till machine coordinates and then build up again.

Thank you for spending the time to check out my changes. Bringing up a build environment is a lot of work.

>> And we note that G10 L2 P0 speaks about the active coordinate system.
>> I believe we need to distinguish P0 from the other (see below)

Just a bit of clarification – in “G10 L2 P0”, currently “P0” is an alias for the last G5x command. So if you issue “G56”, P0 is G56. P0 is the coordinate system that the rest of the system outside of the interpreter (Axis, etc) uses.

>> After writing this down I came to the conclusion that P0 needs to be treated as relative (G91)
>> And P1-P6 as absolute (G90), but with the possibility to adapt (not change).

This is a substantial change. It is not something I can do as part of the changes currently under review.

Bob

yamyamar:
Bob,

The above was not a change I want to implement, it was solely a reflection of me that I chaired.
It made me think deeper over translations and Rotations in Gcode and  was very clarifying for me.

Possible, that my way is not the Linuxcnc way. And no regrets if the outcome in Linuxcnc is different. 
My thinking was just based on our practise.
We use movement and rotations; when pallets are changed, multiple item are milled on a table, to put an axis following a part and to position data more easy on the part(so the math becomes more easy), and just in 3 axis, and in 2 dimensions at a time.
When we mill with 5axis we leave the setting of the coordinate system over to the cam-computer.
As Linuxcnc, is capable in controlling 9axis, so for sure, the view can be different.

I'm using Linuxcnc as a preparation for making Haas programs, (at home)
I'm finishing another project before I move to making our new "Haas" move in my way;-)
I'll will try the same files I have send you so we can evaluate how a "Fanuc-clone"style controller is responding to G10 code.
Normally we would use rotation in Haas with  G68, but for the understanding we will try-out a G10 version.
I suspect it will move the linuxcnc way as both are "Fanuc-styled".

You can expect my findings beginning of December.

Regards,

Jan

rgbond:
Andy, Chris, Rene, how can we make progress on this? The changes fix real bugs...

c-morley:
I'm not sure I'm qualified to assess this.

Though I can confirm P0 is an alias for the active coordinate system, but there is a bug - if used in a gcode program it will use the active coordinate system at interpreter time rather then at motion time.

yamyamar:
Bob,

I'm not forgetting about this issue.

But programming time is still fully occupied, as I'm bending a Haas controller to fit into our needs.
Holiday's are coming, besides playing with the turkey I hope to have some spare moments to get along with LinuxCnc and my HomeMill;-)

I have the plan to write a small program (a letter "A" or something) for testing all the rotational an translating functions.
I would execute the same program (adapted for the language) on a Haas and take pictures of the graphical results for easy comparison with Linuxcnc. Feel free to indicate what you want to be tested.

I had a look at an "official" RS274 paper, but did not found the rotational/translation behavior as used in linuxcnc.
I suppose that the "rules" implied in linuxcnc are written down somewhere?
Are you able to address me to such a paper?

Thanks in advance.
Regards
Jan

rgbond:
Hi Jan,

 

I think that the syntax and semantics for rotation in Linuxcnc are unique to Linuxcnc.  I couldn’t find a clear exposition of the rules so my work was an attempt to make what I could figure out consistent across the platform. If you find a definitive reference I would love to see it.

 

You might check the tests that I included in the. They should run standalone with just the interpreter.

 

Bob

andypugh:
It's taken up semi-permanent residence at the bottom of my inbox too. It's a matter of finding the time at the right time.

yamyamar:
Bob,

I was diving into the definitions found in the rs274 document.
The most official one I could find is the NIST document https://www.nist.gov/publications/nist-rs274ngc-interpreter-version-3.
I was astonished to found out what shortcomings this standard has in regard to what the commercial market is offering as functions. So in my opinion to follow this standard to design Linuxcnc is like taken the Cassette recorder standards when designing a digital audio system? But this is another conversation. 
If interested try f.e. Heidenhein 
https://content.heidenhain.de/doku/tnc_guide/html/en/index/1242135142456/1242135142470/1242135142478/1242135142478.html 
pag 333, shows you the approach of moving and rotation of axis.

I made a small sketch and tried  the Linuxcnc way to solve this.
May idea would be; as the center is in the middle.
Center is G54
Move axis G52 to the right 
Rotate while executing a hole.
Anti rotate to set the axis again to 0°
G52 to the left 
Use polar coordinates to get the holes drilled.

To extend this thought say that we have 4 places on the machine were we execute this part.
Meaning G55, G56 and G57 are also used.

With my overview given above I'm not able to solve this as Linuxcnc does not allow me to rotate around a G52 setting.

Can you try to make the Gcode to mill this part?

Let me know,
Regards,
Jan



BOB

rgbond:
Hi Jan,

 

Sorry it took so long to get back to this. I don’t think you can do it with G52. I think  it may be possible to do it if you use a “temp” setup, say G59. The idea would  be “G54”, then “G0 x50, Y-30”, then use “G10 L2 P6 0 0 R0” followed by  “G59. That should set up the center point for the following operations. Then “x50” drill, “X0”, “G10 L2 P6 R45”, “X50”, etc. Use “G54 X0 Y0” to get back the center.

 

Bob

pltsi:
Hello.

I have been testing the PR on top master with simulator. This PR appears to fix #30, but not the problem that I'm facing.

As I read the PR description and LinuxCNC documentation, it is mentioned that G92 offset should be applied **after** rotation. However, the code in this PR still seems to calculate the rotation with G92 offset included. This can be clearly seen in axis, that the G92 offset (when displayed) is rotated with G10 L2 P1 Rx. My understanding based on the documentation is that the offset and it's direction should remain constant regardless of the applied rotation.

I'm using persistent G92 offset in my machine to adjust the location of my work surface and it's features. Applying rotation changes the direction of G92 offset and things don't work as expected anymore..

Also axis opengl drawing code in lib/python/rs274/glcanon.py seems to apply G92 offset before rotation, at least for the grid.

So, how exactly should LinuxCNC behave with G92 offset applied?

Best regards,
Sami

rgbond:
Thanks Sami.  I spent a bunch of time looking at the interpreter code, trying to figure out how to make things consistent. It is certainly possible that I misconstrued something.

 

>> However, the code in this PR still seems to calculate the rotation with G92 offset included 

Can you send some gcode that demonstrates what you mean by this? The code in the PR applies the G5x translation, then applies the rotation, then adds in the G92 offset. That means that the G92 offset will send the spindle off in a different direction with different rotations.

>> My understanding based on the documentation is that the offset and it's direction should remain constant regardless of the applied rotation

I’ve been looking for documentation that addresses this issue. Please send a reference.

 

>> Applying rotation changes the direction of G92 offset and things don't work as expected anymore..

This is what I would expect. The interpreter has inconsistent applications. The PR is an attempt to make it all consistent. Again, please send a bit of gcode that demonstrates where you see an issue so I can take a look at it.

 

Bob

yamyamar:
Bob,

That was my point, the above is a real world example.
This kind of drawing is what we give to the cnc operators.
Our commercial CNC-machines do allow to solve the above with a G52.

As I did not find the linuxcnc behaviour in the rs274 description. In fact I did not find any definition for accumulating rotation/translation?
And as I do not see the advantage in practise of the way linuxcnc implements the rotation/translation behaviour.
I can only say that I would program the behaviour in a different way.


Best regards,
Jan

yamyamar:
Bob,
As promised a test on a Haas for the rotation/G52 behavior.
HAAS g52

As before, a similar program on Linuxcnc
G52

Regards,
Jan

andypugh:
I lost track of this one. 
@rgbond Do you think that this is correct and should be merged?

rgbond:
Hi Andy, 

Yes, I think it is correct. At a minimum though, I would prefer that someone run the tests I provided against the current release (without my changes) to verify that the problems that I found and fixed are really problems. That should only take a few minutes…

Bob

andypugh:
I had forgotten that this addressed a reported issue. 
Sorry for the long delay in getting this merged.

PR #1300, opened 2021-10-12
qtvcp: add gcode file units to props

phillc54:
Adds the units system the gcode file was written in to gcode props


c-morley:
Thanks

PR #1305, opened 2021-10-14
Adjusted udev rules to flag xhc units as user accessable.

petterreinholdtsen:
This is an alternative to using the plugdev group, which provide access
to the user logged in on the console without having to statically grant
plugdev membership.


petterreinholdtsen:
[Sebastian Kuzminsky]
> This looks straight-forward enough to me.  Do you have any input on
> how the xhc udev rules are handled vs the shuttle rules?

Not really.  I assume you are talking about
debian/extras/lib/udev/rules.d/99-shuttle.rules?  As far as could tell,
it set the device files up to be readable by all, so no plugdev group,
and nothing useful to gain from a uaccess tag.  But remember, I have no
idea what these devices are used for nor which dongles the USB IDs
represent.  Perhaps its udev rule should be more like the xhc ones?

> This looks like a good bugfix, please retarget this PR for the 2.8
> branch.  We'll merge this PR into 2.8 then merge 2.8 up to master.

I can do that.  But your comment make me suspect the master branch is
not used like I am used to, as the development branch.  Where can I read
up on how the branches are used in the linuxcnc project?

-- 
Happy hacking
Petter Reinholdtsen

PR #1306, opened 2021-10-14
Added Appstream metadata announcing supported USB IDs

petterreinholdtsen:
See https://wiki.debian.org/AppStream/Guidelines for information about
the format and setup.

### DIFF
--- a/debian/copyright
+++ b/debian/copyright
@@ -30,6 +30,10 @@ Files: src/rtapi/*
 Copyright: the LinuxCNC developers
 License: LGPL-2.1+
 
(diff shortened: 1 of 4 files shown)


petterreinholdtsen:
[Sebastian Kuzminsky]
> The two USB ids in the metadata are for the XHC jog dongles.  Should
> this include the USB ids for the Shuttle devices too?

Yes.  Good catch, I did not look closely at the IDs.  I'll update the
pull request.  Should I aim for the 2.8 branch too?

-- 
Happy hacking
Petter Reinholdtsen

PR #1314, opened 2021-10-17
WIP: Provide man page images as PlantUML for easier maintenance.

petterreinholdtsen:
See https://plantuml.com/timing-diagram for information about the format
used in toggle.puml and toggle2nist.puml.

### DIFF
--- a/docs/man/images/toggle.puml
+++ b/docs/man/images/toggle.puml
@@ -1,0 +1,23 @@
+@startuml
+hide time-axis
+binary "in"  as in
+binary "out"  as out
+
+@0
+in is low
+out is low
+
+@1
+in is high
+out is high
+
+@1.5
+in is low
+
+@3
+in is high
+out is low
+
+@3.5
+in is low
+@enduml
--- a/docs/man/images/toggle2nist.puml
+++ b/docs/man/images/toggle2nist.puml
@@ -1,0 +1,42 @@
+@startuml
+hide time-axis
+robust "in"  as in
+binary "on"  as on
+binary "off"  as off
+binary "is-on"  as ison
+
+@0
+in is low
+on is low
+off is low
+ison is low
+
+@1
+in is high
+on is high
+
+@2.2
+in is {low,high}
+
+@3
+on is low
+ison is high
+
+@4
+in is low
+
+@6
+in is high
+off is high
+
+@7.2
+in is {low,high}
+
+@8
+off is low
+ison is low
+
+@9
+in is low
+
+@enduml


SebKuzminsky:
Good catch, currently the timing diagram postscript images just show up in the HTML docs:  http://linuxcnc.org/docs/2.8/html/man/man9/toggle2nist.9.html#DESCRIPTION

Can PlantUML output its timing diagrams as ascii-art, or are you proposing including the PlantUML source code in the manpage?  Or is there some other option that I'm missing?

The timing diagrams rendered by PlantUML look great, and I do like the idea of doc images generated from source code (instead of appearing in our repo in a hard-to-edit image format), but it feels a bit silly to pull in Java to generate them...  I'd be sold if puml was written in python, which we already used in the build.  As is, I'm conflicted.  @jepler, @andypugh, @rene-dev, do you have an opinion?

petterreinholdtsen:
[Sebastian Kuzminsky]
> Can PlantUML output its timing diagrams as ascii-art, or are you
> proposing including the PlantUML source code in the manpage?  Or is
> there some other option that I'm missing?

PlantUML can output ascii-art, as well as PNG, SVG, EPS, PDF, HTML LaTeX
and others.

-- 
Happy hacking
Petter Reinholdtsen

SebKuzminsky:
I typed your toggle2nist puml into the www.plantuml.com server.  The PNG and SVG look great:
https://www.plantuml.com/plantuml/svg/TP1D2i8m44RtSugXQojQnPzThfBIZlsWJQ2JeYBUtGYfZQ2xZvUNHuZHOCSm3JrrQ5eE6DhItg0q-dhIm0LSmMoLuQY6i-ET2v-Gpv18OY8V22rd5HfbCXk2YvjxVoM_B95vGXTCPflq3fUE_3B9LEigdJsYkdhJPyIxPEMfT6DFoCg7hx989Xv_fq__dtdEqgPrJVpD5m00

But the ascii art looks underwhelming:
https://www.plantuml.com/plantuml/txt/TP1D2i8m44RtSugXQojQnPzThfBIZXreCz19KH7lReIqIK5t7oylZn0mwktefw67ZfiMFGzjQH-iC4ezgSU2NO5e5Tb1pSwEJoma8iaGKMH42sGjPvKro61sm2vixEKEaXRHF5aJ1hEFUiUt3YHDCDMsYcUle6w-z1tm8LfBAjqu0cVbqwfCb8ddt-dB_sTUL-dMDU4xFm00

Am I doing something wrong here?

SebKuzminsky:
I can make myself be ok with it, especially if it makes good ascii-art (sorry, I mean utf-8-art) for the manpages.

petterreinholdtsen:
I created https://github.com/plantuml/plantuml/issues/872 instead of trying to reach the plantuml directly via email.

andypugh:
Is this good to merge, in your opinion? Or have you had better ideas since?

petterreinholdtsen:
[andypugh]
> Is this good to merge, in your opinion? Or have you had better ideas
> since?

Not until we figure out a better way to handle the ascii edition.  I
have not spend much time on coming up with one.
-- 
Happy hacking
Petter Reinholdtsen

smoe:
@petterreinholdtsen - maybe rename this branch to WIP like you have done for the po4a one.

hansu:
I have an alternative solution: Using https://github.com/Wren6991/asciiwave.
It works for the same sources are generated from) where the  [current images](http://linuxcnc.org/docs/2.8/html/man/man9/toggle2nist.9.html#DESCRIPTION)

So [this](https://github.com/LinuxCNC/linuxcnc/blob/master/docs/man/images/Images_info.md) produces this for HTML and PDF
image
and this for the man page
image

Looks quite nice, doesn't it? I only had to insert line breaks in the man page manually. But maybe this gets automatically solved when we generate the man pages from AsciiDoc.

hansu:
It's just a Python script. 
And if we use the sources, I consider generating also the postscript file during build. But for that I have to figure out if WaveDrom is working from the CLI because it is designed as web application.

smoe:
That is exactly why Debian prefers to keep things separate. Your added functionality is more likely to be merged back upstream for the benefit of all. Looking forward to it!

hansu:
But for https://github.com/Wren6991/asciiwave there are two additional python packages required. I don't think it is worth adding more build dependencies for that. I mean, how often are timing diagrams for man pages are changed or even created? Before I introduced these two as kind of template "how we could have images in the HTML representation of man pages", there was nothing like this. So I would suggest adding this script to the repo and maybe creating and optional target for the makefile.

smoe:
As a start I am fine with that and such a solution was already taken for other diagrams in the archive. However, if we can promote that software a bit more - why not add it to Debian and just use it - if you drag that script from within the source tree or as a package is not too much of a difference.

petterreinholdtsen:
What about dropping the graphical figure and only copying the text edition into the man page for use in both text and PDF editions?

hansu:
That would be the easiest, but maybe not the prettiest.
For those two images we have right know I dont care.
But in future maybe we can use conditional statements in asciidoc.

petterreinholdtsen:
[Hans Unzner]
> That would be the easiest, but maybe not the prettiest.
> For those two images we have right know I dont care.
> But in future maybe we can use conditional statements in asciidoc.

Sure, we could, but I am starting to suspect it is not wise.  The wisest
appear seem to be to make sure the man pages are identical across all
renderings.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
Yeah that's the option that makes least trouble. it seems introducing real images in man pages makes too much trouble.
Even though it looks really nice to have images there and for those simple components it makes no sense to create a separate doc page.
But we can create an overview page with timing diagrams of all components. I think that would be helpful.

silopolis:
Le sam. 27 août 2022 à 10:56, petterreinholdtsen ***@***.***>
a écrit :

> [Hans Unzner]
> > That would be the easiest, but maybe not the prettiest.
> > For those two images we have right know I dont care.
> > But in future maybe we can use conditional statements in asciidoc.


Sure, we could, but I am starting to suspect it is not wise. The wisest
> appear seem to be to make sure the man pages are identical across all
> renderings.



Seems like a valuable point 🤔... But identical they are, just rendered
differently.

PR #1316, opened 2021-10-17
Debian needs /usr/bin/python3 in shebang line

smoe:
Separated from PR #1296 .

### DIFF
--- a/debian/rules.in
+++ b/debian/rules.in
@@ -90,6 +90,8 @@ endif
 	(cd debian/tmp/usr/share/doc/linuxcnc/examples; \
 	    ln -sf ../../../linuxcnc/ncfiles ./nc_files)
 	dh_movefiles
+	# Fixing python path
+	sed -i '1s%^#!/usr/bin/-*python3.*%#!/usr/bin/python3%' debian/linuxcnc-uspace/usr/bin/* debian/linuxcnc-uspace-dev/usr/bin/halcompile
 ifeq "$(enable_build_documentation)" ""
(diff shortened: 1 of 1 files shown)


SebKuzminsky:
Thanks @smoe

SebKuzminsky:
I just reverted 63a2092.

I've proposed a different way to solve the problem in #1318: teach our configure script to pick python3 in the first place, instead of rewriting all the scripts afterwards or hand-feeding it the python3 name.  If @jepler and @rene agree I'll revert @petterreinholdtsen's 03f3627 too as it will be not needed.

jepler:
While we're at it, we might see whether AX_PYTHON_DEVEL has had the "python 3.1" bug fixed yet and grab a new version
```
                                ac_python_version=`$PYTHON -c "import sys; \
                                        print (sys.version[[:3]])"`
```

smoe:
I checked on http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_python_devel.m4  and that uses "split", indeed.

It explicitly mentions that users can set environment variables to override it. I tend to like @petterreinholdtsen 's patch even more. I feel that the requirement to constrain the shebang line to the major version is something Debian-specific. If you ever come up with a conda package for LinuxCNC, you may not want that since conda installs several version of Python at the same time.  So, to set the environment variable in a debian-specific part but leave everything else may pay off in yet unforseen ways. But then again - these are dear side-issues sinking too much time. Whatever works is fine.

PR #1317, opened 2021-10-17
remove python version checks from qtvcp

rene-dev:
they are no longer required. Can somebody please test this?


c-morley:
Is there a reason besides just wanting this?
It will make merging and cherrypicking more difficult.

c-morley:
I take that back - there is no version checking in 2.8

c-morley (approved):
I added a little change - looks good (and tests fine) to me otherwise

PR #1318, opened 2021-10-18
shebang "/usr/bin/python3", not "/usr/bin/python3.9"

SebKuzminsky:


rene-dev:
It should be #!/usr/bin/env python3 for best cross platform support.

petterreinholdtsen:
I suspect the approach in #1327 might be a better way forward.

petterreinholdtsen:
@smoe, is this patch still needed?

smoe:
From the Debian-build perspective this is no longer required. I would support the idea to just close this PR and wait if any problem resurfaces. A python3.9 would make problems when the system is updated but the linuxcnc install is not. But this is then also easily detected and fixed.

dwrobel:
Is there any plans to merge it?

From the Fedora perspective, having this merged would be very helpful.

Fedora [guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_interpreter_invocation) recommends using `#!/usr/bin/python3` saying that:
>Using #!%{python3} (#!/usr/bin/python3) rather than e.g. #!/usr/bin/env python ensures that the system-wide Python interpreter is used to run the code, even if the user modifies $PATH (e.g. by activating a virtual environment).

>From the Debian-build perspective this is no longer required. I would support the idea to just close this PR and wait if any problem resurfaces. 

I'm not an expert in Debian, buf I was able to find a similar [guidelines](https://www.debian.org/doc/packaging-manuals/python-policy/index.html#interpreter-location) for Debian which also recommends using `#!/usr/bin/python3`.

Interestingly, Debian has a similar comment w.r.t to `/usr/bin/env` saying that:
>This is not advisable as it bypasses Debian’s dependency checking and makes the package vulnerable to incomplete local installations of Python.

andypugh:
It seems that Debian doesn't need/want this but Fedora does. How do we decide which way to go?

smoe:
My reading is that both Fedora and Debian prefer #!/usr/bin/python3 . Debian just updated the default Python in unstable from 3.9 to 3.10. I think it would be a considerable win for Debian if this works since the way it was until last week (3.10 available but 3.9 default) we have 3.10 libraries linked against a 3.9 version and that causes problems.

smoe:
I am just not sure about what happens when the default python version changes - but that does not happen in a stable environment.

andypugh:
Will there be a Python 4?

andypugh:
Can you look at the conflict? I am not up to date enough to decide which to keep.

smoe:
> Will there be a Python 4?

I meant when the default becomes 3.11.

smoe:
> Can you look at the conflict? I am not up to date enough to decide which to keep.

I cannot push to this branch, @SebKuzminsky, could you please address a rebase?

hansu:
> > Can you look at the conflict? I am not up to date enough to decide which to keep.
> 
> I cannot push to this branch, @SebKuzminsky, could you please address a rebase?

You can just tell which lines to keep:
```
<<<<<<< better-shebang
AX_PYTHON_DEVEL([>='3.7'],,AC_MSG_ERROR([python-dev not found]))
PYTHON=`echo ${PYTHON%.*}`
=======
AX_PYTHON_DEVEL(,,AC_MSG_ERROR([python-dev not found]))
>>>>>>> master
```

petterreinholdtsen:
As this patch is no longer needed on Debian, which uses PYTHON=/usr/bin/python3 in debian/rules.in instead, and Fedora can adjust its build rules the same way, I believe it is time to close it without merging it.

PR #1325, opened 2021-10-21
Import LSB shell methods into realtime init.d script

petterreinholdtsen:
This allow systemd to hook into the script.  This also get rid of
a lintian warning on Debian.


petterreinholdtsen:
#1326 make more sense than this pull request.

PR #1330, opened 2021-10-24
qtplasmac: set kerf width and height per volt to 3 places on imperial

snowgoer540:


phillc54:
Nice change for the dinosaurs. :)

PR #1343, opened 2021-10-30
Moved use of compile time detected values (config.h) out of installed header files

petterreinholdtsen:
Values not detected at compile time, ie LINELEN, BUFFERLEN, MM_PER_INCH
and INCH_PER_MM, are moved into a new file linuxchc.h.

To avoid having to install the architecture specific config.h file,
restructure includes and introduce linuxcnc.h for definitions used
across the code.  Drop the redundant MM_PER_INCH from
interp_internal.hh and CMS_CONFIG_LINELEN from cms_cfg.hh and use the
value from linuxcnc.h instead.  Made sure all C files depending on
config.h values include it directly.  Make sure no header file include
config.h.

As uspace_common.h is not exposed to API clients, and is a code
fragment and not a normal include file, the use of config.h is left
as it is there.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -155,6 +155,7 @@ SUBDIRS := \
     emc/motion emc/ini emc/rs274ngc emc/sai emc emc/pythonplugin \
     emc/motion-logger \
     emc/tooldata \
+    emc \
     \
     module_helper \
     \
@@ -310,7 +311,7 @@ default: $(INFILES)
 # files in the source tree are the ones used when building linuxcnc.  The copy
 # in ../include is used when building external components of linuxcnc.
 HEADERS := \
-    config.h \
+    emc/linuxcnc.h \
     emc/ini/emcIniFile.hh \
     emc/ini/iniaxis.hh \
     emc/ini/inijoint.hh \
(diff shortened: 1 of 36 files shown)


petterreinholdtsen:
Note, I suspect some uses of LINELEN are better suited as PATH_MAX, as they seem to be file paths not text lines.  I did not dare to do this rewrite in this patch. Also, to get the code building on Hurd and to avoid arbitrary size limites, I would suggest to rewrite every use of LINELEN or PATH_MAX to handle any array length using alloca() or malloc() instead, but did not take the time required to try on this larger task.

petterreinholdtsen:
I do not understand what causes the error "install: cannot stat '../lib/python/qtvcp/designer/x86_64/qt5.5/*.gz': No such file or directory".

phillc54:
Those files were removed from master today, they were unndeeded python2 files.
https://github.com/LinuxCNC/linuxcnc/commit/ba62f3d9ba033958109d1805fa2dcceb420705e8

petterreinholdtsen:
Right.  Then I guess someone forgot to update the install target in the make file.

petterreinholdtsen:
A better workaround is provided in #1345.

SebKuzminsky:
I merged #1345, please rebase to drop 12c8c73 and i'll merge this.

petterreinholdtsen:
[Sebastian Kuzminsky]
> I merged #1345, please rebase to drop 12c8c73 and i'll merge this.

Done.

-- 
Happy hacking
Petter Reinholdtsen

SebKuzminsky:
Thanks @petterreinholdtsen !

PR #1354, opened 2021-11-04
Readmes are .adocs by heart

smoe:
I also had a look at the INSTALL files but those are better left as they are.


rene-dev:
https://github.com/LinuxCNC/linuxcnc/runs/4111517801?check_suite_focus=true#step:4:11912
you need to rename the readme here:
https://github.com/LinuxCNC/linuxcnc/blob/16822b4ea15208bd38bb3e7cba3f694005573249/debian/linuxcnc.files.in#L29

PR #1357, opened 2021-11-05
Fix typos in src/emc

luzpaz:
Found via `codespell -q 3 -S *.po,*.ts,./share,./docs/man/es,./configs/attic -L ans,doubleclick,dout,halp,parm`

### DIFF
--- a/src/emc/pythonplugin/testpp.cc
+++ b/src/emc/pythonplugin/testpp.cc
@@ -146,7 +146,7 @@ main (int argc, char **argv)
     }
 
     if (!PythonPlugin::instantiate(builtins ? builtin_modules : NULL)) {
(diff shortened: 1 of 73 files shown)


src/emc/pythonplugin/testpp.cc
```diff
@@ -146,7 +146,7 @@ main (int argc, char **argv)
     }
 
     if (!PythonPlugin::instantiate(builtins ? builtin_modules : NULL)) {
-	fprintf (stderr,"instantiate: cant instantiate Python plugin");
+	fprintf (stderr,"instantiate: can\'t instantiate Python plugin");
```
luzpaz, line 149: is this correct?

rene-dev: yes

PR #1360, opened 2021-11-09
d/README.source: Instruction how to prepare git snapshot

smoe:
When preparing from git snapshot the date and hash should be specified. This is also possible with uscan, but uscan should only inform us about new releases, not about every patch, so that dependency on single commits should be external to uscan as an exception.

Did not change it with this commit, but just to have mentioned it: We can introduce the epoch any time. I suggest to upload without the epoch. Could well imagine that someone with a running linuxcnc install from the current repository would not want to accidentally update it when all that was wanted is a system update.

The package currently does not build because of an update of asciidoc.

I thought I had created an ITP for LinuxCNC but apparently I have not: Could not find it at least. And since there is a package for LinuxCNC already for a couple of decades, maybe we also do not need it.

### DIFF
--- a/debian/README.source
+++ b/debian/README.source
@@ -14,7 +14,6 @@ required for Debian unstable.
 Steps
 -----
 
-
 1. Generate the source tarball, This will use the version as specified
    in the first line of tthe debian/changelog file. And maybe remove
    anything outdated, just to avoid confusions:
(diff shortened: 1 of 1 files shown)


SebKuzminsky:
I fixed the asciidoc problem in master a few hours ago.

smoe:
Will update. Moment.

PR #1364, opened 2021-11-09
Fix realtime script location

spike77453:
I think there might have been a mixup introduced in https://github.com/LinuxCNC/linuxcnc/commit/eb7997d2390fde047866d5ba6bd401dd03018e60#diff-ae87d3a8398b892f1dc4cda159b07e561414e9da8c18c17710c530b28a80cb6e regarding the location of the `scripts/realtime` file. This PR tries to address that.

Please note that I'm not sure if `lib/linuxcnc/` is a sound location for the file to begin with. Maybe this should be put into something like `${bindir}` or `$(datadir)/linuxcnc/`.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -614,6 +614,7 @@ install-dirs:
 	$(DIR) $(DESTDIR)$(initd_dir) $(DESTDIR)$(EMC2_RTLIB_DIR) \
 		$(DESTDIR)$(sysconfdir)/linuxcnc $(DESTDIR)$(bindir) \
 		$(DESTDIR)$(libdir) $(DESTDIR)$(includedir)/linuxcnc \
+		$(DESTDIR)/usr/lib/linuxcnc \
 		$(DESTDIR)$(docsdir) $(DESTDIR)$(ncfilesdir) \
 		$(DESTDIR)/etc/X11/app-defaults $(DESTDIR)$(tcldir)/bin \
 		$(DESTDIR)$(tcldir)/scripts \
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
As it is an executable, and its implementation should be hidden from its callers, I believe $(DESTDIR)$(libdir)/linuxcnc/ (aka /usr/lib/linuxcnc/) is an appropriate resting place for it.  If it is guaranteed to be a architecture neutral script for the rest of eternity, it could go into  $(DESTDIR)$(datadir)/linuxcnc/ (aka /usr/share/linuxcnc/).

spike77453:
Thanks @petterreinholdtsen for the feedback. Happy to leave the PR as is (i.e. putting `realtime` into `/usr/lib/linuxcnc/realtime`) but can also change it to `$(DESTDIR)$(datadir)/linuxcnc/`

`$(DESTDIR)$(libdir)/linuxcnc/` is a bit tricky though, since that's `/usr/lib/linuxcnc/` on debian but might be (depending on arch) `/usr/lib64/linuxcnc/` on distributions like RHEL, Fedora and others. The location of the script is hard-coded to `/usr/lib/linuxcnc/` in (at least) https://github.com/LinuxCNC/linuxcnc/blob/master/src/configure.ac#L719 where `$(libdir)` is not (easily) available.

Let me know what the preferred solution is and I'll change the PR accordingly.

spike77453:
It seems like this issue was fixed in https://github.com/LinuxCNC/linuxcnc/pull/1403. Thanks @dwrobel and @SebKuzminsky for taking care of this!

Issue #1363, opened 2021-11-09
Building debs in master fails with permissions issue after yesterday's update


rodw-au:
## Here are the steps I follow to reproduce the issue:

 1. cd ./src git pull
 2. cd ..
 3. dpkg-buildpackage -b -uc

## This is what I expected to happen:
debs would be built just like they did yesterday
**## This is what happened instead:**

executing regular installation with regular prefix but unterneath the debian folder
DESTDIR=/home/plasma2/linuxcnc-dev/debian/linuxcnc-uspace /usr/bin/make -C src install
make[2]: Entering directory '/home/plasma2/linuxcnc-dev/src'
emc/Submakefile:4: warning: overriding recipe for target '../include/linuxcnc.h'
emc/Submakefile:4: warning: ignoring old recipe for target '../include/linuxcnc.h'
install -o root -d -m 0755	/home/plasma2/linuxcnc-dev/debian/linuxcnc-uspace/usr/lib/linuxcnc/modules/linuxcnc \
	/home/plasma2/linuxcnc-dev/debian/linuxcnc-uspace/usr/bin \
	/home/plasma2/linuxcnc-dev/debian/linuxcnc-uspace/etc/linuxcnc
install: cannot change owner and permissions of ‘/home/plasma2/linuxcnc-dev/debian/linuxcnc-uspace/usr/lib/linuxcnc/modules/linuxcnc’: Operation not permitted
install: cannot change owner and permissions of ‘/home/plasma2/linuxcnc-dev/debian/linuxcnc-uspace/usr/bin’: Operation not permitted
install: cannot change owner and permissions of ‘/home/plasma2/linuxcnc-dev/debian/linuxcnc-uspace/etc/linuxcnc’: Operation not permitted
make[2]: *** [Makefile:798: install-kernel-dep] Error 1
make[2]: Leaving directory '/home/plasma2/linuxcnc-dev/src'
make[1]: *** [debian/rules:71: override_dh_auto_install] Error 2
make[1]: Leaving directory '/home/plasma2/linuxcnc-dev'
make: *** [debian/rules:39: binary] Error 2
dpkg-buildpackage: error: debian/rules binary subprocess returned exit status 2

## It worked properly before this:
(If the behavior changed after making a particular change in hardware or
software, describe the change you think is responsible.  E.g., "after upgrading
from LinuxCNC 2.7.3 to 2.7.4")

## Information about my hardware and software:

[...]

PR #1366, opened 2021-11-10
docs: Added -rfakeroot to dpkg-buildpackage and simplified

smoe:
This addresses (at least as a first shot) the issues @rodw-au experienced in https://github.com/LinuxCNC/linuxcnc/issues/1363 . Many thanks to Rod for his report.

### DIFF
--- a/docs/src/code/building-linuxcnc.adoc
+++ b/docs/src/code/building-linuxcnc.adoc
@@ -229,9 +229,7 @@ The first step is generating the Debian package scripts and meta-data
 from the git repo by running this:
 
 -----
-> cd linuxcnc-dev/debian
-> ./configure uspace
-> cd ..
+> ./debian/configure uspace
 -----
 
 [NOTE]
@@ -249,7 +247,7 @@ package by running `dpkg-buildpackage` (note that it needs to run from
 the `linuxcnc-dev` directory, *not* from `linuxcnc-dev/debian`):
 
 -----
-> dpkg-buildpackage -b -uc
+> dpkg-buildpackage -rfakeroot -b -uc
 -----
 
 
@@ -265,7 +263,10 @@ platform to build for.  The normal values for this argument are:
 `uspace`::
 
     Configure the Debian package for Preempt-RT realtime or for
-    non-realtime (these two are compatible).
+    non-realtime (these two are compatible). uspace is the default if no
+    other realtime configuration is selected.
+
+    The Debian package created will be linuxcnc-uspace.
 
 `noauto`::
 `rtai`::
@@ -290,6 +291,14 @@ platform to build for.  The normal values for this argument are:
     Configure the Debian package for the currently running RTAI kernel.
     You must be running an RTAI kernel on your build machine for this
     to work!
+
+`--help`::
+
+    presets a very short summary of extra options
+
+`no-docs`::
+
+    skips building of documentation
 
 `$KERNEL_VERSION`::
 
@@ -326,9 +335,7 @@ Then ask your LinuxCNC git checkout to generate its Debian package
 meta-data:
 
 -----
-> cd linuxcnc-dev/debian
-> ./configure uspace
-> cd ..
+> ./debian/configure uspace
 -----
 
 Finally ask `dpkg-checkbuilddeps` to do its job (note that it needs to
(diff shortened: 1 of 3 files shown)


rodw-au:
Thank you for fixing this. Much appreciated.

docs/src/code/building-linuxcnc.adoc
```diff
@@ -291,6 +292,14 @@ platform to build for.  The normal values for this argument are:
     You must be running an RTAI kernel on your build machine for this
     to work!
 
+`--help`::
+
+    presets a very short summary of extra options
```
SebKuzminsky, line 297: Type - should be "Presents".

SebKuzminsky (changes requested):
Typo.

SebKuzminsky:
The current (ie before this PR) build instructions work for me without `-rfakeroot` on latest master, b58e0b62ca545555d7.  There must be something about this problem that i'm not understanding.

rodw-au:
Others reported the same and I was able to create the debs without -rfakeroot on my chromebook that had not been building the debs previously. I had been building the debs on a regular basis on Debian 11 on my production machine for about a month until it failed with some kind of permissions issue when I raised issue  #1363. I suspect the issue may have been caused by a conflict with permissions on files that had been created in a previous build.
I  did observe a number of commits in relation to the .deb build process in the few days before I observed the issue so I assumed something had changed. I was able to build the debs when logged in as root.

rodw-au:
I wonder if its because my user and root have different passwords?
I usually always make the root password and user password identical but I didn't on this build.

SebKuzminsky:
> I wonder if its because my user and root have different passwords?

That shouldn't matter.

phillc54:
I haven't noticed any issues without -rfakeroot either.

smoe:
That is weird. Could it be the version of Debian that makes the difference? I am building on Debian unstable, which should be mostly equivalent to bookworm (version 12). I am not sure about what is the right thing to do, but since -rfakeroot with uttermost confidence does not do any harm (that is how packages were always built not too long ago) and with just a bit less of a confidence someone else will go through the same pain like @rodw-au , I suggest to just add it.

docs/src/code/building-linuxcnc.adoc
smoe: Done. It unfortunately took me another iteration to recognize that the sentence should start with a capital letter. Also  fixed for the following section.

smoe:
@petterreinholdtsen , does this ring a bell?

phillc54:
I build on both MX-Linux19 (a Debian10 spinoff) and MX-Linux21 (a Debian11 spinoff)

petterreinholdtsen:
[Steffen Möller]
> @petterreinholdtsen , does this ring a bell?

I suspect the handing of Rules-Requires-Root, which is fairly recent in
Debian, might be related.  Personally I use 'debuild' from devscripts,
which insert fakeroot when needed, so I never have to add fakeroot
settings manually.

-- 
Happy hacking
Petter Reinholdtsen

smoe:
>> [Steffen Möller]
>> @petterreinholdtsen , does this ring a bell?

> I suspect the handing of Rules-Requires-Root, which is fairly recent in Debian, might be related. Personally I use 'debuild' from devscripts, which insert fakeroot when needed, so I never have to add fakeroot settings manually.

@SebKuzminsky , I suggest to then go for the -rfakeroot.

SebKuzminsky:
It's a bit hard to tell from the bug report since it does not specify which git commits built and which failed, but my hunch is that @rodw-au 's build problem stems not from problems with the build instructions or with the contents of the git repo, but from root-owned files being accidentally created and later not being able to remove/clean those files as a non-root user.

I don't want to train our users to add `fakeroot` (or worse, `su`) where it's not needed.  So the updates to the docs that add `-r fakeroot` should be removed imo.

I *do* like the updates to the build docs that simplify running `debian/configure` instead of `cd debian; ./configure; cd ..`.  Please leave those in.

With that change (and maybe a change to the title of the PR) I think it will be ready to merge.

petterreinholdtsen:
Note, this patch is partly made obsolete by #1835 .  I strongly recommend rewriting the documentation to use 'debuild' instead of introducing -rfakeroot, as it is a more high level tool doing more quality checking automatically.

smoe:
Too outdated to chase this all up.

PR #1368, opened 2021-11-10
Fix typos in config/sim

luzpaz:
Found via `codespell -q 3 -S *.po,*.ts,./share,./docs/man/es,./configs/attic,*_fr.*,*_es.* -L ans,ba,bulle,componentes,doubleclick,dout,dum,fo,halp,inout,parm,parms,te,ue,wille`

### DIFF
--- a/configs/sim/axis/external_offsets/queuebuster.ngc
+++ b/configs/sim/axis/external_offsets/queuebuster.ngc
@@ -22,7 +22,7 @@ o<queuebuster> sub
 
     o<um66> if [#<use_m66> GT 0]
       (debug, m66 queuebuster follows)
-      m66p0l0   ;QB wait on input l0=imediate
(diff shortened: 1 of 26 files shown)


configs/sim/axis/external_offsets/queuebuster.ngc
```diff
@@ -22,7 +22,7 @@ o<queuebuster> sub
 
     o<um66> if [#<use_m66> GT 0]
       (debug, m66 queuebuster follows)
-      m66p0l0   ;QB wait on input l0=imediate
+      m66p0l0   ;QB wait on input l0=immediate
```
luzpaz, line 25: please review closely

rene-dev:
thanks. I think it would be a good idea to have a script in scripts/ that does all this, so it can be ocasionally checked.

PR #1369, opened 2021-11-10
Remove initd_dir since it's no longer used

spike77453:
Since https://github.com/LinuxCNC/linuxcnc/commit/eb7997d2390fde047866d5ba6bd401dd03018e60 was merged, `initd_dir` seems to be no longer necessary since `realtime` was the only file put into `/etc/init.d`

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -611,7 +611,7 @@ install: install-kernel-dep install-kernel-indep
 	$(ECHO) "Installed in DESTDIR='$(DESTDIR)' with prefix $(prefix)"
 
 install-dirs:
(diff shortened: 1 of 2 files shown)


rene-dev:
its still used here:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/Makefile#L614
or can that be removed as well?

spike77453:
> its still used here: https://github.com/LinuxCNC/linuxcnc/blob/master/src/Makefile#L614 or can that be removed as well?

Pretty sure it can be. I've updated the PR accordingly.

PR #1370, opened 2021-11-10
Refactor IEC Timers

mbjohnston:
The t_on, t_off, and t_p HAL components had some behaviour inconsistent with other standard components which made using them a bit difficult.

This PR converts them to the comp format so that they get the standard boilerplate behaviour - mainly `count=` and `names=` support and standardized functions for each instance loaded. It also renames them to `ton`, `tof`, and `tp`, respectively. This aligns better with the IEC 61131-3 standard, and I feel is an acceptable change since the arguments required to use them components is also changing (that is, this PR is already a breaking change).

The internal logic of each component is identical, other than two minor changes:
* Change `if(et > pt)` to `if(et >= pt)` in the "update outputs" section of ton and tof. This stops et from exceeding pt when the timer finishes, mainly aesthetic.
* Change tp to only trigger on a rising edge of q. Previously, if q was left high the component would start a new output pulse as soon as the last one finished.


andypugh:
Thanks

PR #1376, opened 2021-11-13
Dropped nroff manual pages, maintain them as asciidoc

petterreinholdtsen:
The asciidoc manual pages are converted to groff during the build process.

The migration was done using scripts/man2adoc-migrate, with manual editing to clean up the formatting, correct manual page redirects (files with only groff .so statements), and adjusted file names and content to get this working.

Adjusted build rules to use new names for intro man pages. To avoid a name conflict intro.3hal and intro.3rtapi was renamed to hal.3 and rtapi.3 (both could not be generated to intro.3).


silopolis:
Just to up this one before meeting and show support to this 👍

andypugh:
There is embedded (n/t/g)roff in some of the .comp file documentation sections. Will those need changing to adoc? 

Example: https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/components/carousel.comp

hansu:
Okay this gives us nice output for the docs. But if I want to access them as man pages in the terminal - how will that work? Or what did I miss?

petterreinholdtsen:
[andypugh]
> There is embedded (n/t/g)roff in some of the .comp file documentation
> sections.

There is even embedded HTML in them, side by side with the roff strings
using roff if statements.

> Will those need changing to adoc?

I suspect a good way forward is to change halcompile.g to generate adoc
files into docs/src/man/man9/ instead of groff files into
docs/man/man9/.  It is not part of this pull request because I was not
aware of the source of the generated man pages when I started.  As they
can not be converted using pandoc, I guess it is better left for a
different pull request.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[Hans]
> Okay this gives us nice output for the docs. But if I want to access
> them as man pages in the terminal - how will that work? Or what did I
> miss?

I suspect you missed the adoc->man conversion already taking place with
the files in docs/src/man/.  This pull request is about moving
everything maintained in docs/man/ into docs/src/man/ instead of having
manual pages in both locations and in two different formats.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
Ah okay great. That makes the Groff update a little bit unnecessary :smile:

smoe:
@petter, the error is shown a couple of thousand lines above. Somehow there is a title missing for the man page:
```
a2x: executing: "/usr/bin/asciidoc" --backend docbook -a "a2x-format=manpage"  --doctype manpage --attribute "mansource=LinuxCNC" --attribute "manmanual=LinuxCNC Documentation" --verbose  --out-file "/home/runner/work/linuxcnc/linuxcnc/docs/man/man1/axis-remote.1.xml" "/home/runner/work/linuxcnc/linuxcnc/docs/src/man/man1/axis-remote.1.adoc"

asciidoc: reading: /etc/asciidoc/asciidoc.conf
asciidoc: reading: /home/runner/work/linuxcnc/linuxcnc/docs/src/man/man1/axis-remote.1.adoc
asciidoc: FAILED: manpage document title is mandatory
```
The asciidoc-ed man pages are in your branch only, right? I just looked at the axis-remote.1 on my system and that seems fine?!?

petterreinholdtsen:
[Steffen Möller]
> @petter, the error is shown a couple of thousand lines above. Somehow
> there is a title missing for the man page:

Aha, thank you for tracking this down.  I had a look at the original
adoc manual pages, and notice this change is required to get this to
work:

diff --git a/docs/src/man/man1/axis-remote.1.adoc b/docs/src/man/man1/axis-remote.1.adoc
index d7ca90c547..c05e898357 100644
--- a/docs/src/man/man1/axis-remote.1.adoc
+++ b/docs/src/man/man1/axis-remote.1.adoc
@@ -1,3 +1,5 @@
+= axis-remote(1)
+
 == NAME
 
 axis-remote - AXIS Remote Interface

I guess pandoc is missing a key part required to convert the manual page
to adoc, so this will have to be done manually.

> The asciidoc-ed man pages are in your branch only, right?

Yes.

> I just looked at the axis-remote.1 on my system and that seems fine?!?

Of course, it is the original text before I ran pandoc to convert it to
adoc. :)

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
Did looked more into this, and discovered that pandoc is less strict about converting from man pages to adoc files, than a2x is about converting the converted adoc files back to man pages.  I found https://stackoverflow.com/questions/11411024/valid-asciidoc-to-convert-to-manpage-with-a2x useful in this regard.  This mean the automated migration script is incomplete, and we might have to update the generated adoc files manually for this to work.

In particular, I discovered that pandoc will convert empty HISTORY sections, while a2x refuse to convert adoc files with empty sections to man pages.

petterreinholdtsen:
It would be useful to have a decision up front if this is the direction we want to move in, as it is going to be some work to get the migration ready, and all changes to man pages should be done before the migration, and stopped completely until after the migration is done.  Who can make a commitment regarding this?

hansu:
I wouldn't start this until halcomp is ready to handle asciidoc.

petterreinholdtsen:
[Hans]
> I wouldn't start this until halcomp is ready to handle asciidoc.

Why not?  This will not affect halcom man pages.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
The goal for this is primary to unify the formatting language for both docs and man pages, right?
And secondary to make it easier to write man pages if I remember correctly.

So if we don't convert the man pages in the comp files, we only shift the split between asciidoc and groff to another point. So I would first extend halcomp to support asciidoc.

petterreinholdtsen:
[Hans]

> The goal for this is primary to unify the formatting language for both
> docs and man pages, right?  And secondary to make it easier to write
> man pages if I remember correctly.

Yes.  One step at the time.  Not trying to solve everything here.

> So if we don't convert the man pages in the comp files, we only shift
> the split between asciidoc and groff to another point. So I would
> first extend halcomp to support asciidoc.

I am very happy to learn you will look at this.  I suggest it is done in
a separate pull request.  I suspect we need to support both roff and
adoc components for a while, assuming there are components not in the
linuxcnc git repository that should keep working with linuxcnc.
-- 
Happy hacking
Petter Reinholdtsen

hansu:
The goal for this is primary to unify the formatting language for both docs and man pages, right?
And secondary to make it easier to write man pages if I remember correctly.

So if we don't convert the man pages in the comp files, we only shift the split between asciidoc and groff to another position. So I would first extend halcomp to support asciidoc.

silopolis:
Le dim. 17 juil. 2022 à 11:51, Hans ***@***.***> a écrit :

> So if we don't convert the man pages in the comp files, we only shift the
> split between asciidoc and groff to another position. So I would first
> extend halcomp to support asciidoc.
>

Echo echo! 😉

>

hansu:
>Yes.  One step at the time.

Sure. But I thougt of a different order =)


>I suspect we need to support both roff and
adoc components for a while, assuming there are components not in the
linuxcnc git repository that should keep working with linuxcnc.

I would make halcomp backward compatible so that the old comps are supported forever. I have in mind to add a section in the comp file that contains asciidoc. So if this section is present, the asciidoc is used, otherwise groff is used.

smoe:
This is listed as work for 2.9 on https://github.com/orgs/LinuxCNC/projects/1/views/1 - this is of course for @andypugh to decide, but I propose a) not to wait with 2.9 since this internal optimization does not contribute any function and b) to have this for 3.x since it changes a how developers interact with the documentation.

Another thought is that I better do not invest into translations of man pages.

hansu:
The project page is more a proposal how it could look like and this issue I even don't put in the To Do column. And yes, it's up to the release manager what should go in.

petterreinholdtsen:
[Steffen Möller]
> This is listed as work for 2.9 on
> https://github.com/orgs/LinuxCNC/projects/1/views/1 - this is of
> course for @andypugh to decide, but I propose a) not to wait with 2.9
> since this internal optimization does not contribute any function and
> b) to have this for 3.x since it changes a how developers interact
> with the documentation.

I thought Andy was the release manager for 2.8, not 2.9.  Is this wrong?
Note, also I am not propose to hold off 2.9 while waiting for this
change.  I only propose to hold off man page changes for the duration of
the conversion (aka a few days).

> Another thought is that I better do not invest into translations of
> man pages.

Note, the strings will not really change between nroff and adoc, so the
translations will  be carried over via the PO file.

-- 
Happy hacking
Petter Reinholdtsen

silopolis:
Le ven. 26 août 2022 à 09:49, Steffen Möller ***@***.***> a
écrit :

> This is listed as work for 2.9 on
> https://github.com/orgs/LinuxCNC/projects/1/views/1
>
Very glad to discover a project has been created without notice and without
dropping a single word about it in related #1960 issue... :/

hansu:
Sorry my friend, but I created this project and it was not possible for me to make it public, so I asked Andy to do so. And just today I noticed that is was made public when Steffen mentioned it. My plan was to announce it on the mailing list, when it is public. Furthermore I am on a little bicycle trip and haven't read all my mails.

silopolis:
Le ven. 26 août 2022 à 11:28, Hans Unzner ***@***.***> a
écrit :

> Sorry my friend, but I created this project and it was not possible for me
> to make it public, so I asked Andy to do so. And just today I noticed that
> is was made public when Steffen mentioned it. My plan was to announce it on
> the mailing list, when it is public. Furthermore I am on a little bicycle
> trip and haven't read all my mails.
>

Oh, OK ! Then, nice to see we two at least have the will to give these
tools a try :)

Have a nice ride mate ✌

hansu:
Can you provide an example branch with one or two man pages in the asciidoc format and the corresponding changes to the Submakefile?
Would be interesting to see how it looks like.
I think the first step is:
( man pages in ascidoc format ) -> Groff -> HTML/PDF
( comp files ) -> Groff -> HTML/PDF

As a second step i could imagine:
( man pages in ascidoc format ) -> Groff
( man pages in ascidoc format ) -> HTML/PDF
( comp files ) -> Groff -> HTML/PDF

Maybe we can skip the first step and jump direct to the second step? Shouldn't be a big thing.

petterreinholdtsen:
[Hans Unzner]
> Can you provide an example branch with one or two man pages in the
> asciidoc format and the corresponding changes to the Submakefile?

This pull request contain the complete set.  Can you use it instead of
only a small subset?

I am reluctant to spend more time on the task until I know if this is
the direction we want to be going.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
>This pull request contain the complete set.  

I missed the integration in the makefile...

petterreinholdtsen:
[Hans Unzner]
> I missed the integration in the makefile...

The makefile already build some man pages from adoc, specificly the
files in docs/src/man/.  As far as I can tell, no changes are needed in
the makefile.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
>The makefile already build some man pages from adoc, specificly the
files in docs/src/man/.  As far as I can tell, no changes are needed in
the makefile.

Oh sorry I didn't know that we support asciidoc man pages since 2015 and also have some in the asciidoc format :see_no_evil:

andypugh:
Persuading halcompile to create manpages in adoc format rather than Troff should be easy. 
For example swapping the //f and //R here for the adoc equivalent: https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/utils/halcompile.g#L856

hansu:
I would suggest to to that in small portions, otherwise it would be a huge amount to review and may would get stuck.

petterreinholdtsen:
[Hans Unzner]
> I would suggest to to that in small portions, otherwise it would be a
> huge amount to review and may would get stuck.

You are probably right.  I welcome others to help with the migration.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
I have rebased the branch on master and regenerted all adoc man pages
based on the latest edition of the groff files.  I believe it is ready
to go into master.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
The CI test failed, and I am not quite sure why, so it might need some
work.  In the process of debugging this, I discovered that asciidoc
actually support man page links (.so), and that this might be related to
the problem, so I have pushed an update generating the links from
asciidoc.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
For the record, I figured out the CI failures (some generated files were
not properly removed in git, and some files were not properly
generated), and this change should now be ready to go into master.

Note that the patches will have to be updated if someone change any of
the groff manual pages, so I would prefer for it to go in before this
happen.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
These changes seem not to be used in the current master docs.

Just a small check: http://linuxcnc.org/docs/devel/html/man/man1/io.1.html does have the section "SEE ALSO" but the source from where it should be generated from does not have it: https://github.com/LinuxCNC/linuxcnc/blob/master/docs/src/man/man1/io.1.adoc

petterreinholdtsen:
[Hans Unzner]
> These changes seem not to be used in the current master docs.
>
> Just a small check:
> http://linuxcnc.org/docs/devel/html/man/man1/io.1.html does have the
> section "SEE ALSO" but the source from where it should be generated
> from does not have it:
> https://github.com/LinuxCNC/linuxcnc/blob/master/docs/src/man/man1/io.1.adoc

How strange.  Not quite sure which job is building the content in <URL:
http://linuxcnc.org/docs/devel/html/ >, so no idea why it fail.  The
build of the man pages happen locally for me.  Is it also happening for
you?

-- 
Happy hacking
Petter Reinholdtsen

hansu:
Yeah locally they get builded from the adoc files.
But the section "SEE ALSO" is missing for some of them bit not all. Weird.
And there is sometimes a space before the second asterix i.e. `*iocontrol.0.coolant-flood *::` which leads not to the desired bold formatting.

petterreinholdtsen:
[Hans Unzner]
> Yeah locally they get builded from the adoc files.  But the section
> "SEE ALSO" is missing for some of them bit not all. Weird.  And there
> is sometimes a space before the second asterix
> i.e. `*iocontrol.0.coolant-flood *::` which leads not to the desired
> bold formatting.

The amount converted was a bit too much for me to review, and fixed the
issues I found.  If you see more, please send patches to fix them.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
> The amount converted was a bit too much for me to review, and fixed the issues I found. If you see more, please send patches to fix them.

There are quite a few, and I plan to fix them, but I think first it is more important to figure out why they don't show up on the website.

petterreinholdtsen:
[Hans Unzner]
> There are quite a few, and I plan to fix them, but I think first it is
> more important to figure out why they don't show up on the website.

I do not really know how the web site is updated.  I suspect it is a
buildbot job somewhere.

Perhaps the first step is to ensure 'make install' or similar is working
with the web editions?

-- 
Happy hacking
Petter Reinholdtsen

PR #1382, opened 2021-11-15
Backlash bug on YZAC axes + SSR invertion bug

3Dpassion:
sorry, wrong section


tests/remap/andypugh/toplevel.py
```diff
@@ -0,0 +1,3 @@
+#! /usr/bin/python
```
smoe, line 1: /usr/bin/python is not installed on my machine. There is https://packages.debian.org/sid/main/python-is-python3 to help with some extra convenience - but, nah, or?

jepler:
@3Dpassion Hi, it looks like this PR was opened in error.  Please feel free to open a fresh one in the future with your proposed changes. Thank you.

PR #1384, opened 2021-11-15
d/copyright: Reaction to FTPmasters

smoe:
I would like to list more LinuxCNC-folks.

### DIFF
--- a/debian/README.source
+++ b/debian/README.source
@@ -1,6 +1,17 @@
 Debian packaging of LinuxCNC
 ============================
 
+debian/copyright
+----------------
+
+Automated tools to check licenses within the source tree will likely find
+CC-2.5 referenced. But this refers only to some image files that were
+not adopted by the source tree, only the README describing them was 
+intentionally left unchanged.
+
+Steps
+-----
+
 The following steps were performed to transform Upstream's git repository
 into
  - a source tarball and
@@ -10,9 +21,6 @@ for Debian's main distribution.
 LinuxCNC is used across a wide span of releases, now supporting from
 Squeeze up to the latest version of Ubuntu. Not too many changes are
 required for Debian unstable.
-
-Steps
------
 
 1. Generate the source tarball, This will use the version as specified
    in the first line of tthe debian/changelog file. And maybe remove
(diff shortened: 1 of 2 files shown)


SebKuzminsky:
`share/qtvcp/images/widgets/writer/icons/README.txt` will mentions three sets of icons with CC-2.5 licensing, none of which are included in our repo.  It seems cleanest to remove mention of these files we don't have.  What do you think?

smoe:
d/copyright mentions the CC-3.0 so we are fine. We could explain this in d/README.source. Would prefer not to touch upstream's license declaration.

smoe:
Sidenote: src/hal/hal_lib.c is LGPL-2, not GPL. I think we have to crosscheck a bit more.

smoe:
@SebKuzminsky I am now happy enough with this. The individuals appearing as copyright holders throughout the source tree should all be mentioned and instructions how to retrieve commits from the repository are given. Please kindly review and - from my side this can now be merged.

smoe:
After some closer inspection it seems like I never completed my previous manual inspections. I now used a
```
find src/hal -name "*.comp* | xargs -r -- egrep -B2 -A2 'Copyright|copyright|Author|author'
```
as a filter and then manually went through it. I then did the same for src/hal -type f and merged. I likely missed someone or someone's year for which I then apologize. I still think this can be merged, nothing speaking against adding authors/years later.

rene-dev:
https://github.com/LinuxCNC/linuxcnc/blob/2e75b091a23e3312b9e57a88941714701f3709b8/src/hal/user_comps/xhc-hb04.cc
Frederic, Marius and myself are missing for some reason.

smoe:
@rene-dev Thank you. I had you and Frederic already in my local repository but  Marius was still missing. Anyone else coming to mind whom I have left out?

rene-dev:
why was this file not picked up by find?

rene-dev:
can this be merged already?

smoe:
Since you had asked why the three of you were not found I just repeated my greps on the whole tree and found your names a few more names and more years of others, so I suggest to let me iterate through what comes up and receive more feedback. As I said in my email to the list - this is (for me) just something technical to prepare for next upload to the distribution and I expect this to updated and likely also completely substitute with some automated routine in some near future.

smoe:
Iterated through them all again and I admit to need some extra motivation to repeat that effort. Merge and expect PRs, I suggest.

PR #1386, opened 2021-11-16
Fix typos in misc. source comments and docs

luzpaz:
Found via `codespell -q 3 -S *.po,*.ts,./share,./docs/man/es,./configs/attic,*_fr.*,*_es.* -L ans,ba,bulle,componentes,doubleclick,dout,dum,fo,halp,ihs,inout,parm,parms,ser,te,ue,wille,wont`


SebKuzminsky:
Thanks @luzpaz !

PR #1395, opened 2021-11-21
Enhance halshow tool

hansu:
Added several useful features to halshow.  
In detail:

1. Added buttons for manipulating the pins
2. Added context menu to
   - copy pin name
   - set value
   - unlink pin (if connected to signal)
   - remove pin from list  
  
halshow-changes-2

3. Added entry dialog to set int/float values

halshow-changes-3

4. Made text copy-able by right click and Ctrl+C
5. Moved entry dialog to main frame for greater width  

halshow-changes-1

I hope you like it :smile: 

For the buttons i used code from https://wiki.tcl-lang.org/page/Canvas+Buttons:
> License statement
>
> Tcl and extensions such as Tk, Expect, tcllib are distributed under the tems of this license , a BSD-type license, which is much less restrictive than the GPL.
> 
> It is safe to assume that any code posted on this Wiki (or that you post on this wiki) is, unless otherwise explicitly specified, under the same license.

I'm not 100% sure if it complies with the LinuxCNC license model. So please check this if someone knows more about that.

### DIFF
--- a/src/po/de.po
+++ b/src/po/de.po
@@ -4983,11 +4983,11 @@ msgstr "Pins aufklappen"
 
 #: tcl/bin/halconfig.tcl:177 tcl/bin/halshow.tcl:116
 msgid "Expand Parameters"
-msgstr "Parameters aufklappen"
+msgstr "Parameter aufklappen"
 
 #: tcl/bin/halconfig.tcl:179 tcl/bin/halshow.tcl:118
 msgid "Expand Signals"
-msgstr "Signals aufklappen"
+msgstr "Signale aufklappen"
 
 #: tcl/bin/halconfig.tcl:182 tcl/bin/halshow.tcl:121
 msgid "Erase Watch"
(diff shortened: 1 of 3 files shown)


andypugh:
Just the ability to take a miss-clicked item out of the list would be a big win. 

https://opensource.stackexchange.com/questions/304/can-i-take-bsd-licensed-code-and-distribute-it-under-gpl
Seems to say (in most of the answers, but not all) that BSD-licensed code can be incorporated into a GPL project. 

However I wonder if there is existing code in the project that could be used instead, for example in sim_pin:
https://github.com/LinuxCNC/linuxcnc/blob/master/scripts/sim_pin

rene-dev:
I just tested it, and I love it! Thanks so much.
License should be no problem.
Maybe the documentation should be updated.
@andypugh you can remove items by right click, remove.

hansu:
> However I wonder if there is existing code in the project that could be used instead, for example in sim_pin: https://github.com/LinuxCNC/linuxcnc/blob/master/scripts/sim_pin

Do you mean to avoid the license issue or in general to reuse existing code?
I didn't know the sim_pin script. I could borrow the input dialog from this for example. But changes in this script would might break halshow.

> Maybe the documentation should be updated.

Good point. I'll keep it in mind. 
I also thought about an info message which informs the user that there is now a right-click-menu available. Usually one wouldn't try a right click i guess.

andypugh:
> @andypugh you can remove items by right click, remove.

That was my point, you can with this but before there was no way. (Well, there was one way, you could save the layout, edit in a text editor, and import...)

snelweg:
This is great! Thank you so much for making this.

hansu:
Ah I forgot to mention one feature: There is now a command history, use with Up/Down keys.

PR #1396, opened 2021-11-21
smallish changes to debian/{rules.in,copyright} done for first upload

smoe:
Some harmless optical fixes and d/rules now deletes files created during build. That helps to execute dpkg-buildpackage twice in a row, e.g. after the first build failed.

### DIFF
--- a/share/qtvcp/screens/qtplasmac/qtplasmac_handler.py
+++ b/share/qtvcp/screens/qtplasmac/qtplasmac_handler.py
@@ -1,4 +1,4 @@
-VERSION = '1.221.161'
+VERSION = '1.221.162'
 
 '''
 qtplasmac_handler.py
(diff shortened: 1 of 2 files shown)


petterreinholdtsen:
Is the clean update really needed?  Running 'debuild && debuild' in master work just fine.

smoe:
debuild does not check for differences outside debian with the orig.tar.gz. The challenge is dpkg-buildpackage && dpkg-buildpackage. I tend to agree though that this cleaning should be done from within LinuxCNC, not within Debian.

jepler (approved):
thank you.

smoe:
@petterreinholdtsen , I do not recall all details, but an interrupted build may also trigger the one or other problem that the cleaning may have attempted to address.

petterreinholdtsen:
Perhaps this change be split, and the cleanup fixes discussed and adjusted in a separate pull request?  I believe some of the cleanup seem to undo stuff done by make (should go into 'make clean'), other stuff done by configure (should go into 'make distclean'), while d/rules clean should only undo what was done by d/rules.

smoe:
"undo" -> "redo/should do"?  I only had my Debian hat on when I prepared those changes and used the PR to inform you all about it. A bit of redundancy in debian/rules I understand is not of concern except that it masks the problems in the other clean routines. Whatever you consider worthwhile to move elsewhere I am very happy with.

SebKuzminsky:
The changes to `debian/copyright` are welcome, please make a PR with just that.
The changes to `debian/rules.in` might be better moved to the `clean` target in `src/Makefile`?  I would prefer to keep things organized by their different areas of responsibility (ie the "main" build system vs the debian packaging system).

petterreinholdtsen:
The d/copyright changes just went in.  Time to rewrite this to do cleanup in the parts responsible?

smoe:
I admit I would like to doge that one. Need to finish something else this week :(

PR #1400, opened 2021-11-24
pncconf: Add tab names for 7i92 mesa card

dwrobel:
Adds tab names (P1 and P2) for 7i92 mesa card as per:
 - http://www.mesanet.com/pdf/parallel/7i92man.pdf
 - http://www.mesanet.com/graphics/parallel/7i92colo.png


c-morley:
Thank you!

PR #1401, opened 2021-11-24
pncconf: Avoid using hardcoded number of pins (24)

dwrobel:
Avoid using hardcoded number of pins per connector
which results in rising IndexError exception
on lines where 24 is hardcoded, like the following
example:

 firmptype,compnum = self.d["mesa%d_currentfirmwaredata"% boardnum][_PD._STARTOFDATA+pin+(concount*24)]

### DIFF
--- a/src/emc/usr_intf/pncconf/pncconf.py
+++ b/src/emc/usr_intf/pncconf/pncconf.py
@@ -2458,6 +2458,14 @@ Clicking 'existing custom program' will avoid this warning. "),False):
         self.set_mesa_options(boardnum,title,firmware,numofpwmgens,numoftppwmgens,numofstepgens,numofencoders,numofsserialports,numofsserialchannels)
         return True
 
+    def get_ppc(self, boardnum: int) -> int:
+        """ Returns the number of pins per connector for a given 'boardnum' """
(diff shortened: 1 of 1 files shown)


c-morley:
Is your idea to allow cards with more then 24 pins? That would be a big change.

dwrobel:
> Is your idea to allow cards with more then 24 pins?

No, rather the opposite, with `17` pins. I've tested it with my `7i92` firmware _(see my [private_data.py](https://github.com/dwrobel/linuxcnc/commit/11feed4d06b2fc52eb8ad31fc8886dfd4aeb9baf))_ and the original `pncconf` didn't work at all.
With this change I didn't encountered any `IndexError` exceptions so far. My assumption is that this change should be transparent to existing configurations and should be helpful in case you have a different [breakout](https://www.amazon.com/Breakout-optical-coupler-Stepper-Driver/dp/B00WN4T23W) card.

c-morley:
pncconf functions all assume 24 pin data. Cards with less padded the firmware data.
Your data doesn't look like it padded the data - that is why you got the error.

I'm not against this change in principle.
This needs to be tested careful - that breaking the assumption doesn't create corner case problems.

c-morley:
Please open again, if this is really needed. Thanks

PR #1402, opened 2021-11-24
Add support for checking "text/x-script.python" in pycheck

dwrobel:
Fixes pycheck error on Fedora >=35:
 Checking "text/x-python" (by MIME type) files python3 compatibility...
 make: Leaving directory '/builddir/build/BUILD/linuxcnc-39d63664bee51ce4fed43d2f46ea71b5e1f4f3bc/src'
 usage: py_compile.py [-h] [-q] filenames [filenames ...]
 py_compile.py: error: the following arguments are required: filenames
 make: *** [Makefile:541: pycheck-python-script] Error 123
 error: Bad exit status from /var/tmp/rpm-tmp.DkTMua (%check)
     Bad exit status from /var/tmp/rpm-tmp.DkTMua (%check)

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -538,10 +538,10 @@ pycheck-python-files:
 	done
 
 pycheck-python-script:
-	@echo 'Checking "text/x-python" (by MIME type) files python3 compatibility...'
+	@echo 'Checking "text/x-python" and "text/x-script.python" (by MIME type) files python3 compatibility...'
 	@for d in ../tests; do \
 		filelist=$(shell mktemp) && \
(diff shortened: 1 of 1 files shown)


jepler (approved):
thanks. I was unable to determine if there is a canonical MIME type for .py files

I note that on my system there's at least one more related file type, based on the list of installed files in the mime database (debian bullseye):
```
/usr/share/mime/text/x-python3.xml
```
and perhaps a script variant of it.  I guess that since the "3" is trailing, the grep will match it either way, since it's looking for a substring and not a full match.

PR #1403, opened 2021-11-24
Unify realtime script installation

dwrobel:
Unifies installation of 'realtime' script in the same
directory as expected by the configure.ac script.

Fixes regression introduced on #1326 which causes 'realtime'
script being installed in directory referred by ${libdir}
while the configure.ac expected it to be installed on 'lib' directory.
On systems (e.g. Fedora, CentOS) where ${libdir} != 'lib'
this generates discrepancy.

Also fixes proper $(prefix)/lib/linuxcnc directory creation.
When this was not created it caused 'realtime' script to be installed
as $(prefix)/lib/linuxcnc instaed of $(prefix)/lib/linuxcnc/realtime.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -612,8 +612,8 @@ install: install-kernel-dep install-kernel-indep
 
 install-dirs:
 	$(DIR) $(DESTDIR)$(EMC2_RTLIB_DIR) \
-		$(DESTDIR)$(sysconfdir)/linuxcnc $(DESTDIR)$(bindir) \
-		$(DESTDIR)$(libdir) $(DESTDIR)$(includedir)/linuxcnc \
+		$(DESTDIR)$(sysconfdir)/linuxcnc $(DESTDIR)$(bindir) $(DESTDIR)$(libdir) \
+		$(DESTDIR)/lib/linuxcnc $(DESTDIR)$(includedir)/linuxcnc \
 		$(DESTDIR)$(docsdir) $(DESTDIR)$(ncfilesdir) \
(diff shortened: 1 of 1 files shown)


SebKuzminsky:
I verified that this works right when building debian packages, and I trust @dwrobel that it does the right thing on Fedora & CentOS.

dwrobel:
Thank you @SebKuzminsky for your confidence. Yes, with this small fix, it started to work agin on Fedora/CentOS however, w.r.t. to your original change #1326, more precisely:

>This PR moves the script from /etc/init.d/ to /usr/lib/linuxcnc/, in accordance with the Debian Filesystem Hierarchy: https://www.debian.org/doc/packaging-manuals/fhs/fhs-3.0.html#usrlibLibrariesForProgrammingAndPa

I'm not quite sure that the `lib` location is the most appropriate for executables. It looks that `libexec` sub-directory sounds a little bit more appropriate. I see both Debian and [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrlibexec) marks `libexec` as an optional directory, but in practice also Fedora uses [libexec](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_libexecdir) for helper programs. 

So, if there would be no objections nor better ideas - we could consider moving `realtime` to `/usr/libexec/linuxcnc` sub-directory to clearly mark that it is executable that is not intended to be run directly by the user.

spike77453:
:+1: for `/usr/libexec/linuxcnc`

PR #1406, opened 2021-11-25
Halshow: further improvements

hansu:
- Buttons on canvas will no move with window resizing (needed if you have very long signal names)
- Using now `halcmd` als subprocess to the command entry (more cababilities)
- Fix when one click again on "Set val" when entry window is still open
- Made strings "translatable" and added german translation for those
- New item on menubar - allows adding signals/pins/parameters by name

halshow-changes-4

### DIFF
--- a/src/po/de.po
+++ b/src/po/de.po
@@ -5081,7 +5081,7 @@ msgstr "Beobachtungsliste laden"
 
 #: tcl/bin/halshow.tcl:107
 msgid "Tree View"
-msgstr "Ansicht"
+msgstr "Menübaum"
 
 #: tcl/bin/halshow.tcl:288
 msgid "Test HAL command :"
@@ -21237,3 +21237,23 @@ msgstr "Setzen"
 #: tcl/bin/halshow.tcl
 msgid "Remove"
 msgstr "Entfernen"
+
+#: tcl/bin/halshow.tcl
+msgid "Add to watch"
+msgstr "Hinzufügen"
+
+#: tcl/bin/halshow.tcl
+msgid "Input value"
+msgstr "Eingabe"
+
+#: tcl/bin/halshow.tcl
+msgid "Add pin"
+msgstr "Pin hinzufügen"
+
+#: tcl/bin/halshow.tcl
+msgid "Add signal"
+msgstr "Signal hinzufügen"
+
+#: tcl/bin/halshow.tcl
+msgid "Add parameter"
+msgstr "Parameter hinzufügen"
--- a/src/po/de.po
+++ b/src/po/de.po
@@ -4768,7 +4768,7 @@ msgstr "Bearbeiten"
 msgid "Cut"
 msgstr "Ausschneiden"
 
-#: tcl/bin/genedit.tcl:94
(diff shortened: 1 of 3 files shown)


hansu:
While I was creating some new images for the docs, I recognized that my change in https://github.com/LinuxCNC/linuxcnc/pull/1406/commits/9b9e1804378948b0d99cb3b317c31d3b90159b1c creates some ugly output (the selection):

halshow-addf-ddt

So switch back to the simple hal-command that doesn't have the extra options like halcmd, or ...?
(There is also an additional *-v* needed to get this output)

hansu:
> Is there any way to add the tab-completion features of halcmd to the halshow command box? I generally avoid using that box just because there is no tab-completion.

Not really IMHO. With TCL I couldn't find a possibilty to generate an auto-completion event like tab without return. But I'm not a TCL expert.

With a little modification of halcmd I could reach the following behaviour, but this isn't really what you had in mind. It only lists the completion possibilities:
```
$ halcmd -c show pin axis.x.e
axis.x.eoffset
axis.x.eoffset-clear
axis.x.eoffset-counts
axis.x.eoffset-enable
axis.x.eoffset-request
axis.x.eoffset-scale
```

By the way:
What is the benefit of the prefix `<commandline>:0:` on errors?
`<commandline>:0: Unknown command 'foo'`
Couldn't it be just removed or replaced by "halshow:" ?

Another thing:
What about changing the title of halshow to HAL Show instead of HAL Configuration?
Maybe less confusing for new users.

@rene-dev no comment on this https://github.com/LinuxCNC/linuxcnc/pull/1406#issuecomment-979413452 ?

rene-dev:
I think the commandline text can be removed.
easiest way of implementing tab completion would be to just use a halrun.
there must be an easy way of embedding a terminal emulator in tcl.

hansu:
> there must be an easy way of embedding a terminal emulator in tcl.

Yeah the same came came also in my mind, but it doesn't seems so. The most suitable I found was [gush](https://wiki.tcl-lang.org/page/gush) - a shell implementation in tcl. But it only supports completion on the filesystem, not for sub-commands:

grafik

If I feel really boring I may implement it by hand in this way :stuck_out_tongue_winking_eye:

snowgoer540:
I think this commit introduced a bug.  When you execute a command in the test HAL command area, setp, for example, the show window goes blank.  In the past it would immediately show that it was setting the pin, or if you were showing a pin it would immediately show the result.

hansu:
@snowgoer540 
Yeah I know. Rene was a bit too fast with merging. i expected a little discussion before merging.

https://github.com/LinuxCNC/linuxcnc/pull/1406#issuecomment-979413452:
> (There is also an additional -v needed to get this output)   

  
------------

It is fixed in https://github.com/LinuxCNC/linuxcnc/pull/1410 (https://github.com/LinuxCNC/linuxcnc/pull/1410/commits/30a963912185395dc7b1f0ec9d07a0d98d82528c)

rene-dev:
I did pull your changes, and tested them, unfortunately did not find this issue. Anyway, that's what master is for.

PR #1410, opened 2021-11-27
Halshow: updated docs + little modifications

hansu:


jepler (changes requested):
Hi, can you answer a question about these changes?  Thank you!

hansu:
> Hi, can you answer a question about these changes? Thank you!

Sure, but what question? Or do you miss a description?

src/hal/utils/halcmd_main.c
```diff
@@ -220,7 +220,7 @@ int main(int argc, char **argv)
             fd = fileno(srcfile);
             fcntl(fd, F_SETFD, FD_CLOEXEC);
         } else {
-            halcmd_set_filename("<stdin>");
+            halcmd_set_filename("");
```
jepler, line 223: here you change the filename from <stdin> to the empty string...

src/hal/utils/halcmd_main.c
```diff
@@ -473,15 +473,17 @@ void halcmd_output(const char *format, ...) {
...
     va_start(ap, format);
     vfprintf(stderr, format, ap);
     va_end(ap);
 }
 
 void halcmd_error(const char *format, ...) {
     va_list ap;
-    fprintf(stderr, "%s:%d: ", halcmd_get_filename(), halcmd_get_linenumber());
+    if (*halcmd_get_filename() != '\0')
+        fprintf(stderr, "%s:%d: ", halcmd_get_filename(), halcmd_get_linenumber());
```
jepler, line 486: and here you check for an empty string. does this mean that "halcmd invalid command" will no longer produce a diagnostic message, but will rather print nothing?

jepler (changes requested):

src/hal/utils/halcmd_main.c
hansu: Because when you are running it from stdin, you know that you are running from stdin. I think there is no reason to display this. If you run a file, it makes quite sense to have the file name and line number.

src/hal/utils/halcmd_main.c
hansu: In case of stdin and commandline the string is set to empty. In these cases the error message will not include the filename. So the output will only contains the error message that is output with
`   vfprintf(stdout, format, ap);`
in line 488

src/hal/utils/halcmd_main.c
jepler: This style is something I'm familiar with from other tools, such as gcc (clang is similar)
```
$ echo 'f' | gcc -Wall -x c - -D'f=if(wrong code)'
<command-line>: error: expected identifier or ‘(’ before ‘if’
<stdin>:1:1: note: in expansion of macro ‘f’
```

The location

jepler:
So, more specifically, given these two behaviors:
```
jepler@bert:~/src/linuxcnc$ printf 'show pin\nzoooo\n' | halcmd -f
Component Pins:

<stdin>:2: Unknown command 'zoooo'
jepler@bert:~/src/linuxcnc$ halcmd zooo
<commandline>:0: Unknown command 'zooo'
```

It's important to me to preserve the first one, and I feel the second one does no harm.

hansu:
Thanks for reviewing.
I removed the changes in halcmd_main.c.

jepler (approved):
Thanks for addressing my review comments.  I didn't perform any testing for myself.

jepler:
Note that the conflicts in the de.po file have to be addressed by you before the PR can be merged.

hansu:
> Note that the conflicts in the de.po file have to be addressed by you before the PR can be merged.

Have you had a look at the conflict? I don't understand why GitHub makes a conflict out of this.

jepler:
Unfortunately, we mere mortals don't get a choice in what git(/hub) considers a merge conflict, we only get to resolve them by hand.

Issue #1182, opened 2021-06-06
master-gtk3 - gmoccapy - settings


gmoccapy:
settings of Themes and sound does not work

andypugh:
I have been trying to get themes to work with Touchy today, and I think that there may be problems outside our control. 

Touchy now works with _some_ themes, but there are lots of warnings of changed property names.

gmoccapy:
HLo Andy, can you share your touch, stuff, so I can try to adapt that to gmoccapy

Final solution need to be to use css style for customizing the view of the gui

andypugh:
I was hoping that I could have a look to see if the same (naive) fixes would help, but at the moment I can't even start the gmoccapy sample configs in master-gtk3

rene-dev:
why does the ui need to set its theme? whats wrong with using the system theme?
@andypugh which error do you get when starting gmoccapy?

andypugh:
> @andypugh which error do you get when starting gmoccapy?

First a dialog box:
https://photos.app.goo.gl/Xys9JQeVkxkSsmdaA
Then a second dialog box (With Gmoccapy open behind it) 
https://photos.app.goo.gl/GboNxENjSonM5Kh88
Closing the dialog box closes Gmoccapy:

````
andypugh@buster:~/linuxcnc$ linuxcnc
LINUXCNC - 2.9.0~pre0
Machine configuration directory is '/home/andypugh/linuxcnc-dev/configs/sim/gmoccapy'
Machine configuration file is '4_axis.ini'
Starting LinuxCNC...
Found file(REL): ./core_sim4.hal
Note: Using POSIX realtime

trivkins: coordinates:xyzc
   Joint 0 ==> Axis X
   Joint 1 ==> Axis Y
   Joint 2 ==> Axis Z
   Joint 3 ==> Axis C
trivkins: Recommend: kinstype=both

Unexpected realtime delay on task 0 with period 100000
This Message will only display once per session.
Run the Latency Test and resolve before continuing.
Found file(REL): ./spindle_sim.hal
Found file(REL): ./simulated_home.hal
Unexpected realtime delay on task 0 with period 100000
This Message will only display once per session.
Run the Latency Test and resolve before continuing.

Found an error!
The following information may be useful in troubleshooting:

Traceback (most recent call last):
  File "/home/andypugh/linuxcnc-dev/lib/python/gladevcp/iconview.py", line 188, in __init__
[...]

PR #1411, opened 2021-11-28
Feature/update gtk icons

PeterMue:
Since GTK 3.10 the usage of the so called "stock icons" is deprecated.
Instead named icons according to the [Freedesktop icon naming spec](https://specifications.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html) should be used.

I've changed them in gmoccapy related files.

This should fix the problems mentioned here: #1182 

Closes: #1182

### DIFF
--- a/lib/python/gladevcp/iconview.py
+++ b/lib/python/gladevcp/iconview.py
@@ -7,7 +7,7 @@
     it is based on a tutorial from ZetCode PyGTK tutorial
     the original code is under the BSD license
     author: jan bodnar
-    website: zetcode.com 
+    website: zetcode.com
 
     Copyright 2012 Norbert Schechner
     nieson@web.de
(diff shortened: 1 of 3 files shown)


lib/python/gladevcp/iconview.py
```diff
@@ -102,7 +102,7 @@ def __init__(self):
         # This will hold the path we will return
         self.path = ""
         self.button_state = {}
-        self.old_button_state = {}        
+        self.old_button_state = {}
```
jepler, line 105: in the future, it doesn't hurt to make whitespace-tidying changes a separate commit.  It's nice to see the cruft cleaned up in any case.

jepler (approved):
thank you

PeterMue:
Yes you're right, both commits are included:

- https://github.com/LinuxCNC/linuxcnc/pull/1548/commits/3b34639916e1b9242f662652143dfc24a0ecacdc
- https://github.com/LinuxCNC/linuxcnc/pull/1548/commits/868487e981816cee5678ae72f709bdd987e80fdb

hansu:
Perfect and thanks for your quick response!
Then I'll have to figure out why we still seem to have a non free desktop spec compliant icon name, which is causing this error: `gi.repository.GLib.GError: gtk-icon-theme-error-quark: Icon 'gtk-home' not present in theme Adwaita (0)`

PR #1414, opened 2021-11-28
qtvcp: led_widget fix off color and add border color

phillc54:
off_color was linked to getColor and setColor.

I also took the opportunity to add a border color property.

### DIFF
--- a/lib/python/qtvcp/widgets/led_widget.py
+++ b/lib/python/qtvcp/widgets/led_widget.py
@@ -37,6 +37,7 @@ class LED(QWidget, _HalWidgetBase):
         self._diameter = 15
         self._color = QColor("red")
         self._off_color = QColor("black")
+        self._border_color = QColor("black")
         self._alignment = Qt.AlignCenter
         self.state = False
         self.flash = False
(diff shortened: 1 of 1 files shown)


lib/python/qtvcp/widgets/led_widget.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, parent=None):
         self._diameter = 15
         self._color = QColor("red")
         self._off_color = QColor("black")
+        self._border_color = QColor("red")
```
c-morley, line 40: Does border_cokor need the color() around it

phillc54: Possibly not, I just assumed that it would be needed.

phillc54: Without it this error occurs:
  File "/home/phill/linuxcnc-dev/lib/python/qtvcp/widgets/led_widget.py", line 111, in paintEvent
    painter.setPen(QColor(self._border_color))
AttributeError: 'LED' object has no attribute '_border_color'

c-morley:
I tested and the line 111:
painter.setPen(QColor(self._border_color))

can be changed to :
painter.setPen(self._border_color)

What do you think of having the border color default to black?
I think the LED visually 'pops' nicer that way.

either way I say push away Phill

phillc54:
> On 29 Nov 2021, at 10:18 am, c-morley ***@***.***> wrote:
> 
> 
> I tested and the line 111:
> painter.setPen(QColor(self._border_color))
> 
> can be changed to :
> painter.setPen(self._border_color)
> 
> What do you think of having the border color default to black?
> I think the LED visually 'pops' nicer that way.
> 

I agree, black was my favourite as well, I went with red as it was the current default.

PR #1416, opened 2021-11-28
Disable fuzzy mode for po file generation

hansu:
Per default msgmerge uses fuzzy mode. That leads to lots of wrong translations if not manually checked and corrected after generation. Unfortunately we have a lot of these actually. Just search the po files for "fuzzy". I would recommend to rather have no translation instead of a wrong one.

### DIFF
--- a/src/po/Submakefile
+++ b/src/po/Submakefile
@@ -70,7 +70,7 @@ po/linuxcnc.pot:
 TARGETS += po/linuxcnc.pot
 
 pofiles: po/linuxcnc.pot
-	set -x; for i in po/*.po; do msgmerge -U $$i po/linuxcnc.pot; done
+	set -x; for i in po/*.po; do msgmerge -U -N $$i po/linuxcnc.pot; done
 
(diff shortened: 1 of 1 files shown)


hansu:
And I would strongly recommend to remove all fuzzy marked translations as they are very likely wrong.
The fuzzy option might be a good thing to help the translator when updating only one language.
But all fuzzy marked entries have to be reviewed. IMHO no fuzzy marked string should ever be committed.
While I'm currently going through the german translation - I didn't find any correct translation of the fuzzy ones. Some of them tell even the opposite.

If you would agree, I can do this for 2.8 as well as for master.
And I would go that far to throw a warning if there are fuzzy strings left.

petterreinholdtsen:
What exactly is the problem you are trying to solve here?  The fuzzy translations are only a hint/help to the translator _when doing manually check_, and is never used directly as a valid translation.  So they should be manually checked, and the fuzzy flag should only be removed once that is done.

hansu:
>The fuzzy translations are only a hint/help to the translator when doing manually check, and is never used directly as a valid translation.

You are totally right, sorry.
I was working previously on a translation where the fuzzy entries were taken into account. So I did not get the idea that it could be different here.

PR #1429, opened 2021-12-06
Added Norwegian Bokmål (nb) program translation

petterreinholdtsen:


petterreinholdtsen:
With #1428 out of the way and Weblate updated to point to the new POT files, it is quicker to start on the translation on weblate than to wait for this pull request to be merged, so I close it and continue on Weblate instead.

https://hosted.weblate.org/projects/linuxcnc/

PR #1430, opened 2021-12-06
witch defenestration

smoe:
We need a strategy how to get more feedback from the translators. The problem with [weblate](https://hosted.weblate.org/translate/linuxcnc) I see is that the instant reward is to increase the percentage of a translation, not to improve punctuation, spelling errors etc in the upstream code base.

### DIFF
--- a/docs/src/gui/gladevcp.adoc
+++ b/docs/src/gui/gladevcp.adoc
@@ -729,7 +729,7 @@ show_counts::
     Set this to False, if you want to hide the counts display in the middle of the widget. +
     default = True
 label::
-    Set the content of the label witch may be shown over the counts value. +
+    Set the content of the label which may be shown over the counts value. +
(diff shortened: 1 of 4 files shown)


silopolis:
Le lun. 6 déc. 2021 à 13:16, Steffen Möller ***@***.***> a
écrit :

> ... not to improve... the upstream code base.
>

Speaking of which 😝, I see a massive amount of work on source docs to be
done without clearly seeing/understanding the real, actual implications and
consequences on po4a'ed translations of such refactorings... Could some
eagle eye help me better grasp what's coming to us?

smoe:
Yip, I just prepared https://github.com/LinuxCNC/linuxcnc/pull/1444 . I expect that for po4a this is all not too bad. The  .po files now created will be matched, I expect. And a new paragraph, well, yes, that will then be a new paragraph and should not matter to much either. Also, any restructuring of the text should not matter since the weblate elements are mostly without context.

PR #1445, opened 2021-12-12
command.c: fixed typo (missing blank) in error message

smoe:
Am not completely confident about the punctuation. A "!" or "." seems to be the right thing to add, the translations on weblate add them.


jepler (approved):
thanks.

we don't have enough of a style guide to say whether these should be sentences with punctuation or not.  If there is trailing punctuation, it should not be "!".

PR #1447, opened 2021-12-12
Merged de translations of 2.8 into master

hansu:


hansu:
Too many conflicts due to new changes in Weblate.

PR #1449, opened 2021-12-13
MB2HAL errorcounter does not count high

andrewheeler82:
I've been using LinuxCNC for over 10 years now. Starting with the parallel port and now currently on the road with a Mesa 7i96. I run my VFD through MB2HAL and I noticed 2 strange things. I have created a patch for this, maybe the developer of MB2HAL could take a look at it. It's currently working so well for me.


SebKuzminsky:
I'm not the author of mb2hal, but this looks right to me.  Thanks for the fix!

PR #1453, opened 2021-12-18
gmoccapy: fix limit switch issue (#1000)

### LINKED ISSUE #1000: "ignore-limits"-checkbox in gmoccapy 3.0.9.1 (LinuxCNC 2.8) is not displayed correctly

 1. Drive machine into a limit switch (or simulate the switch)
 2. Set the checkbox "ignore limits" and turn on again
 3. Jog away (with keyboard) from the limit switch

If the limit switch is not active any longer, the checkbox has to be unchecked.
Turn on and off refreshes the checkbox so it display the correct state.

But it stays active checked even if you stop jogging while the limit switch is still active. But if you try to turn on again because you think the checkbox is checked: no way. Because it is internal (correctly) disabled, but the display is not updated. Unchecking and checking again leads to the expected behaviour.

#### A second thing:

After having such an event the "ignore limits" checkbox is always clickable in off-state. It should be grayed out, when no limit switches are active. Clicking leads to error "no limit switch is active, command will not be executed!" 

In 2.7's gmoccapy version it shows the same behaviour as in 2.8.
In axis of v2.7 this was working properly. 

## What to do?

I think this is done by updating the checkbox on the following events:

* turn on/turn off
* rising/falling edge of limit switch

hansu:
This fixes https://github.com/LinuxCNC/linuxcnc/issues/1000#issuecomment-753649684

resolves #1000 

@gmoccapy

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -2591,6 +2591,9 @@ class gmoccapy(object):
         if state:
             # sensitize ntb_jog when limit tripped
             self.widgets.ntb_jog.set_sensitive(True)
+        else:
+            # refresh immediately when limit is no longer active
+            self.widgets.chk_ignore_limits.set_active(False)
             
     def on_hal_status_mode_manual(self, widget):
         print ("MANUAL Mode")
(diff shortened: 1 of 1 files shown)


gmoccapy:
Have you checked you solution also with a non trivial kinematics?
IMHO you will need to check the kinematics type, as for jogging in a non trivial kinematics and state not homed, you will need to display the JOINT Button and in a trivial kinematics the axis BUTTON.

hansu:
@gmoccapy When there is any tripped limit switch, then it would be mentioned in `stat.limit`, right?
In that case I enable the "ignore limit" checkbox to be able to switch on the machine and move away from the limit switch.
The action is the same as when toggling the estop (except the button manipulation).
'Trivial kinematics' also has its limit switches on joints, so that makes no difference IMHO.

hansu:
Now it is working perfectly. I verified this in a real machine.

c-morley:
Ready to go then?

hansu:
Yes!

PR #1456, opened 2021-12-19
Stop removing POT files in clean target

petterreinholdtsen:
As the POT files are kept in git, it is best to not remove them in the
'clean' target.


jepler (approved):
Thanks. It seems that `TARGETS` is used for some other purposes, like `make install`, but only subsets that don't include the `.pot` files.

PR #1459, opened 2021-12-20
Drop useless makefile variable DOC_SRCS_TXT.

petterreinholdtsen:
It is identical to DOC_SRCS after renaming .txt to .adoc.


jepler (approved):
Thanks

PR #1460, opened 2021-12-20
Drop useless space from docs/src/Submakefile

petterreinholdtsen:
This space causes emacs to warn every time the file is saved.


jepler (approved):
Thanks

PR #1464, opened 2021-12-22
QTVCP - Update Function to setting the TITLE and ICON for QTVCP APP

annhan:
Need to add config in DISPLAY to INI file
With ICON : image file need to save in ini folder

[DISPLAY]
DISPLAY = qtvcp qtdragon
TITLE = MyQTVCP-qtdragon
ICON = silver_dragon.png


c-morley:
I'm a fan of the idea, but not of the implementation.
I think using qtvcp's INFO library to read the INI file directly is a better way and doesn't require modifying the linuxcnc startup script.

annhan:
> I'm a fan of the idea, but not of the implementation. I think using qtvcp's INFO library to read the INI file directly is a better way and doesn't require modifying the linuxcnc startup script.

ok c-morley, it's a good idea for me. thanks

PR #1469, opened 2021-12-24
pyhal component names were not being registered properly when using python3

curtdutt:


c-morley:
Can you give details on your change. I don't understand why it's needed. Actually I also don't know what uses pyhal.py

curtdutt:
Python3 uses unicode strings natively. python 2 uses c style strings
natively. this call converts the unicode string in python 3 to a c style
string prior to passing it to hal. Presumably this works with python2 as
well.

It looked like as part of the conversion to python 3 the other methods that
pass a string were changed to use the encode method as well. The init call
was missed and needed the encode method called as well.

I have pending code that will rely on the pyhal module. I will be
submitting a laser raster module for controlling laser engravers soon.

c-morley:
LOL for some reason I saw 'encoder' rather then 'encode'
Thanks for putting me straight.

curtdutt:
Thanks!

PR #1470, opened 2021-12-24
Added po4a workarounds to English documentation.

petterreinholdtsen:
The po4a system fail to handle some variants of AsciiDoc tables, see
https://bugs.debian.org/1002303 for the details.  This patch
reformat tables to get po4a to parse them correctly.

Also, use full image file name for some under specified file names.


petterreinholdtsen:
With the release of po4a 0.66-1 in Debian, the issue worked around here is no longer present in po4a.  So if all building of translated documentation will be done with po4a >= 0.66, this workaround is not needed.  I leave it to others to decide if a more strict po4a dependency is OK or not.

smoe:
I am not sure if po4a is broken or if these blanks at the end of lines are indeed needed. I made quite an effort to remove the blanks at the end of lines about everywhere and feel a bit bad now.

petterreinholdtsen:
Most trailing blanks (all you removed, I suspect) could go, so no need to feel bad.

SebKuzminsky:
Thanks for these fixes, and thanks for your patience and gentle reminders.

PR #1495, opened 2022-01-04
Fix returning address of local variable in get_buffer_line()

dwrobel:
Fixes the following issue, where function return a pointer
from a local stack allocated buffor:

 libnml/cms/cms_cfg.cc: In function 'get_buffer_line':
 cc1plus: warning: function may return address of local variable [-Wreturn-local-addr]
 libnml/cms/cms_cfg.cc:308:10: note: declared here
   308 |     char linebuf[LINELEN];      /* Temporary buffer for line from
       |          ^~~~~~~

As this function seems to be unused the simplest fix is to remove it.


c-morley:
I would like to merge this but have no idea if there is any danger in doing so (on say older systems?)
My guts says not dangerous as it's unused.
Since you maybe understand this code - could you assess how dangerous the change is?

dwrobel:
```
$ git grep get_buffer_line
src/libnml/cms/cms_cfg.cc:extern char *get_buffer_line(const char *bufname, const char *filename)
src/libnml/cms/cms_cfg.hh:    extern char *get_buffer_line(const char *buf, const char *file);
```
Besides function's prototype and body there is no other reference to this function in the whole repository.

>any danger in doing so (on say older systems?)

Whether it is a new or older system it should be safe to remove this code as it's simply unused.

c-morley:
Thank you.

PR #1497, opened 2022-01-08
Update xhc-hb04.tcl to fix  very slow default accelerations and document the mpg_accels setting. 

rszemeti:
Default axis accel if no mpg_accels were specified was defaulting to 1.0 units/second^2 ... modified to add documentation of the feature and to default to normal axis max accel (afraction and jfraction =1.0) if no individual mpg_accels are provided.


rszemeti:
This fixes the crazy slow accels with the xhc-hb04 wireless pendant if no mpg_accels values are provided in the ini. I have update the webpage at http://wiki.linuxcnc.org/cgi-bin/wiki.pl?Using_A_XHC-HB04_Wireless_MPG_Pendant with at least info about the feature (which was not documented either in the web page or the tcl file) and fixed the tcl so it defaults to the standard axis acceleration instead of 1 unit/second^2

andypugh:
Thanks.

PR #1498, opened 2022-01-08
Gmoccapy: fix jogging with max. velocity

### LINKED ISSUE #1473: Gmoccapy: Jogging with max velocity shouldn't be affected by halui.max-velocity setting
When jogging with keyboard it always jogs with the velocity set it gmoccapy even when the max. velocity is reduced by halui.max-velocity.
If using shift+button then the halui.max-velocity setting is applied. 

Procedure: 
1. Set Jogging speed in Gmoccapy to maximum
2.
   setp halui.max-velocity.direct-value 1
   setp halui.max-velocity.scale 1
   setp halui.max-velocity.counts 5 (for example)
3. Jog. Jog velocity is not reduced to halui.max-velocity.value

@gmoccapy We already talked about this issue, I only want to mention it here that it doesn't get lost.

hansu:
For details see https://github.com/LinuxCNC/linuxcnc/issues/1473

closes #1473

@gmoccapy


src/emc/usr_intf/gmoccapy/gmoccapy.py
```diff
@@ -551,9 +551,10 @@ def _get_pref_data(self):
         self.scale_feed_override = self.prefs.getpref("scale_feed_override", 1, float)
         self.scale_rapid_override = self.prefs.getpref("scale_rapid_override", 1, float)
 
-        # holds the max velocity value and is needed to be able to jog at
-        # at max velocity if <SHIFT> is hold during jogging
-        self.max_velocity = self.stat.max_velocity
+        # Holds the max velocity value and is needed to be able to jog at
+        # at max velocity if <SHIFT> is hold during jogging.
+        # The resulting jogging velocity will be automatically reduced to the maximum machine velocity. 
+        self.max_jog_velocity = 100
```
c-morley, line 557: Where does this get set to any other value? I think your reasonning is 100units/per second will be higher then system max and linuxcnc will limit it internally.
That assumption is not explicit nor future proof - 600 inches per minutes is totally doable.

You could get the actual value from getinfo.py 's function: get_max_jog_vel(self):

c-morley:
It seems, by reading Norberts comments, that he might have meant for the shifted jog speed to be capped by the current max velocity.
Is there a strong reason to have it one way or the other?

gmoccapy:
gmoccapy is supposed to jog with keyboard.
if only the key is pressed the machine should jog with the velocity set on the slider for jog vel, if shift is hold down at the same time, the machine should jog with max velocity from INI file.

IMHO it is not a good idea to hard code a max velocity, because as Chris already mentioned, machines may travel faster than the coded value. Isnt TRAY / Max_Vel supposed to set the max value?

Norbert

src/emc/usr_intf/gmoccapy/gmoccapy.py
hansu: > Where does this get set to any other value? I think your reasonning is 100units/per second will be higher then system max and linuxcnc will limit it internally. That assumption is not explicit nor future proof - 600 inches per minutes is totally doable.
> 


Yes it is not a very good solution, b I just did it like `continuous_jog()` is doing it to have the same behaviour like in Gscreen and other that uses this function. But it should be 100/60, it seems that I missed the "/60"

https://github.com/LinuxCNC/linuxcnc/blob/487e3de70b9f0487260de097f91615ed01e3cf69/src/emc/usr_intf/gscreen/emc_interface.py#L152-L165

https://github.com/LinuxCNC/linuxcnc/blob/487e3de70b9f0487260de097f91615ed01e3cf69/src/emc/usr_intf/gscreen/emc_interface.py#L23


> You could get the actual value from getinfo.py 's function: get_max_jog_vel(self):

Yes, that's much better, I didn't know. You should have mention that in https://github.com/LinuxCNC/linuxcnc/issues/1473 :wink:

hansu:
> It seems, by reading Norberts comments, that he might have meant for the shifted jog speed to be capped by the current max velocity. Is there a strong reason to have it one way or the other?

Maybe I am wrong, but what if in the HALFILE the halui.max-velocity reduces the max velocity. I think this file is processed before the GUI is loaded and so the reduced velocity is passed to max_velocity

https://github.com/LinuxCNC/linuxcnc/blob/487e3de70b9f0487260de097f91615ed01e3cf69/src/emc/usr_intf/gmoccapy/gmoccapy.py#L556

hansu:
> IMHO it is not a good idea to hard code a max velocity, because as Chris already mentioned, machines may travel faster than the coded value. Isnt TRAY / Max_Vel supposed to set the max value?

The underlaying C-Code would limit the velocity. And yes TRAJ / MAX_LINEAR_VELOCITY would be the best if we would get as constant. But getting it via linuxcnc.stat, it is reduced to the current max velocity. So the docs are not correct at this point. See https://github.com/LinuxCNC/linuxcnc/issues/1473#issuecomment-1007508644

hansu:
Now I have it! Just take the value for max value of the slider that Norbert has already taken from
 `getiniinfo.get_max_jog_vel()`. Thanks Chris for the hint!

c-morley:
Are we still good here - ready to merge?

hansu:
IMHO yes. But maybe @gmoccapy wants to approve this?

c-morley:
Norbert seems pretty busy these days. He can adjust the code after if necessary.
There seems to be conflicts no though.

hansu:
That's true indeed.
I am pretty sure that it's a good solution.

c-morley:
Just waiting for tests to finish...

PR #1500, opened 2022-01-08
halscope: save logs as csv files

havardAasen:
Changes the log format to csv.

During creation of the log, it temporarily changes the locale `LC_NUMERIC` to `C`, so the number format will be identical regardless of the locale that was set previously. I believe this will make it simpler if we ever want to read the values back in. It also makes it possible to use comma `,` as a delimiter if we choose. Currently the delimiter is set to semicolon `;`.

Remove some code that was never used, regarding logging an log functionality.

Require at least one active channel to create a log file.

Fixed some minor issues  when we create a log with less active channels then enabled, and if the active channels is not continuous from channel 1.


andypugh:
I think it's safe to make this change, given that we have no tools to make use of halscope logs in the current format.

PR #1510, opened 2022-01-16
Typos2 - changed all occurences of linuxCNC to LinuxCNC

smoe:
Weblate asked for separate translations for either variant, so I thought this should be fixed and then I was surprised that there are about 30 files with the lower-case one (1200 with the upper case).


SebKuzminsky:
Thanks @smoe!

Issue #1398, opened 2021-11-22
Gmoccapy error - combi_dro.py


LJspice:
## Here are the steps I follow to reproduce the issue:

 1. Install linuxcnc buster preempt-RT 2.8 ISO
 2. Change repo to 2.9/master, update/upgrade
 3. Install missing dependency gnome-icon-theme
 4. Open sim gmoccapy config (Lathe_C.ini, though others do it)
 5. Click spindle on button. Error

## This is what I expected to happen:

No error message, spindle shows as on

## This is what happened instead:

Error message. After error appears, jog buttons stop working

```
Traceback (most recent call last):
  File "/usr/bin/gmoccapy", line 2279, in _periodic
    self._update_vc()
  File "/usr/bin/gmoccapy", line 3773, in _update_vc
    diameter = int(self.dro_dic["Combi_DRO_0"].get_position()[1] * 2)
  File "/usr/lib/python3/dist-packages/gladevcp/combi_dro.py", line 633, in get_position
    positions = self._position()
TypeError: _position() missing 5 required positional arguments: 'object', 'p', 'rel_p', 'dtg', and 'joint_actual_position'
```

<details>
  <summary>Click to expand complete terminal log</summary>

```
linuxcnc@lcnc-lathe:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 10 (buster)
Release:	10
Codename:	buster
linuxcnc@lcnc-lathe:~$ uname -a
Linux lcnc-lathe 4.19.0-18-rt-amd64 #1 SMP PREEMPT RT Debian 4.19.208-1 (2021-09-29) x86_64 GNU/Linux
linuxcnc@lcnc-lathe:~$ linuxcnc
LINUXCNC - 2.9.0-pre0-5081-g8a9665315
Machine configuration directory is '/home/linuxcnc/linuxcnc/configs/sim.gmoccapy.lathe_configs'
Machine configuration file is 'lathe_C.ini'
Starting LinuxCNC...
Found file(REL): ./core_sim_lathe_C.hal
Note: Using POSIX realtime
 
identityKinematicsSetup: coordinates:XZC
   Joint 0 ==> Axis X
   Joint 1 ==> Axis Z
   Joint 2 ==> Axis C
identityKinematicsSetup: Recommend: kinstype=both
 
Found file(REL): ./spindle_sim.hal
Found file(REL): ./simulated_home_lathe.hal
[...]

PR #1512, opened 2022-01-16
Fix Combi-DRO

hansu:
This fixes two errors of the combi DRO:

1. Clicking on the DRO caused an exception
2. Calling of get_postion() failed, reported in #1398

### DIFF
--- a/lib/python/gladevcp/combi_dro.py
+++ b/lib/python/gladevcp/combi_dro.py
@@ -267,7 +267,7 @@ class Combi_DRO(Gtk.VBox):
         else:
             if not self.toggle_readout:
                 return
-            self.toogle_readout()
+            self.set_toggle_readout()
 
     # Get propertys
     def do_get_property(self, property):
@@ -523,7 +523,7 @@ class Combi_DRO(Gtk.VBox):
     # this will toggle the DRO around, mainly used to maintain all DRO
     # at the same state, because a click on one will only change that DRO
     # This can be used to change also the others
-    def toogle_readout(self, Data = None):
+    def set_toggle_readout(self, Data = None):
         '''
         toggles the order of the DRO in the widget
 
@@ -609,7 +609,7 @@ class Combi_DRO(Gtk.VBox):
         '''
         self._ORDER = order
         self._set_labels()
-        self.toogle_readout(Data=True)
+        self.set_toggle_readout(Data=True)
 
     # This will return the position information of all three DRO
     # it will be in the order Abs, Rel, DTG
--- a/lib/python/gladevcp/combi_dro.py
+++ b/lib/python/gladevcp/combi_dro.py
@@ -135,6 +135,9 @@ class Combi_DRO(Gtk.VBox):
         self.cycle_time = 150
         self.diameter = False
         self.actual = True
+        self.dtg = 0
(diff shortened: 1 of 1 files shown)


lib/python/gladevcp/combi_dro.py
```diff
@@ -629,13 +632,11 @@ def get_position(self):
...
               while the machine is moving, because of time delays
         '''
-        positions = self._position()
-        if self._ORDER == ["Rel", "Abs", "DTG"]:
-            return positions[1], positions[0], positions[2]
-        if self._ORDER == ["DTG", "Rel", "Abs"]:
-            return positions[2], positions[1], positions[0]
-        if self._ORDER == ["Abs", "DTG", "Rel"]:
-            return positions[0], positions[2], positions[1]
+        if self.diameter:
```
c-morley, line 635: Does this scale the position no matter what axis?
I'm probably missing some details but only X axis should get scaled.
If this reinforced somewhere else?

hansu: You have a Combi_DRO object for each axis. And `diameter` is a member of the Combi_DRO class, so you set it only for the X-axis object.
Here an example for gmoccapy ([go to file](https://github.com/LinuxCNC/linuxcnc/blob/2b07041d04501421cbd56a5fa0810e41beaa2214/src/emc/usr_intf/gmoccapy/gmoccapy.py#L1317-L1334)):
``` python
    def _make_lathe(self):
...
        # if we have a lathe, we will need an additional DRO to display
        # diameter and radius simultaneous, we will call that one Combi_DRO_9, as that value
        # should never be used due to the limit in axis from 0 to 8
        dro = Combi_DRO()
        dro.set_property("name", "Combi_DRO_9")
...
        joint = self._get_joint_from_joint_axis_dic("x")
        dro.set_joint_no(joint)
        dro.set_axis("x")
        dro.change_axisletter("D")
        dro.set_property("diameter", True)
...
```

lib/python/gladevcp/combi_dro.py
```diff
@@ -424,18 +427,18 @@ def _position(self, object, p, rel_p, dtg, joint_actual_position):
         # joint_actual_position = joint positions, not needed here
         
         try:
-            dtg = dtg[self.axis_no]
-            abs_pos = p[self.axis_no]
-            rel_pos = rel_p[self.axis_no]
+            self.dtg = dtg[self.axis_no]
+            self.abs_pos = p[self.axis_no]
+            self.rel_pos = rel_p[self.axis_no]
```
hansu, line 432: I can also take the scale factor at this position into account. Is maybe more logical here.

```suggestion
            dtg = dtg[self.axis_no]
            abs_pos = p[self.axis_no]
            rel_pos = rel_p[self.axis_no]
```            
```  
...
        if self.diameter:
            scale = 2.0
        else:
            scale = 1.0
...            
```
```diff
+           self.dtg = dtg * scale
+           self.abs_pos = abs_pos * scale
+           self.rel_pos = rel_pos * scale
```

hansu: I have implemented this. Much better i think.

c-morley:
Thank you

PR #1513, opened 2022-01-17
Changed [gG] [cC]ode to G-code where appropriate

smoe:
While translating I was wondering about why there was no consensus on the spelling on G-code. It was a ~310 to 140 decision to write it with a hyphen and a lower-case c. Headers and titles have "Code" start unchanged with a capital, but a hyphen was always added when missing.

Somewhat unfortunate, for some files I also removed blanks at the ends of lines, which makes reading this patch more difficult than it meant to be.


smoe:
I just found that my stash still had some linuxcnc -> LinuxCNC conversions, trying to distinguish the package from the script that is started and LinuxCNC as a whole.

SebKuzminsky:
Wow, thank you for your careful attention to detail!
I see you fixed a bunch of things in `index.tmpl`, any chance you can give similar treatment to the translated versions of that file?

snowgoer540:
I intended to make a comment on this last night, but I ran out of time, and I see it is committed now.  I am not sure that the 310 to 140 takes into account the entire project.  I know that in qtplasmac.adoc alone there are at least 140 instances of 'G-Code' and this was not part of your commit.

I would prefer to leave them as 'G-Code' (and 'M-Code' by proxy) as I took great care to align the entire document to 'G-Code' on purpose.  No specific reason, other than 'G-code' looked odd to me.  Additionally, I know there are many many instances of 'gcode' throughout the project which I think we can all agree is the worst option :) .

smoe:
> I intended to make a comment on this last night, but I ran out of time, and I see it is committed now. I am not sure that the 310 to 140 takes into account the entire project. I know that in qtplasmac.adoc alone there are at least 140 instances of 'G-Code' and this was not part of your commit.

I admit not to have explicitly grepped for "G-Code". Good point. This is likely because deep down I was sufficiently happy with it, just the "hyphen or not" we should kind of agree on as a community. But maybe it would not be a bad thing to also adjust the G-Codes, but this is not pressing IMHO.
 
> I would prefer to leave them as 'G-Code' (and 'M-Code' by proxy) as I took great care to align the entire document to 'G-Code' on purpose. No specific reason, other than 'G-code' looked odd to me.

Personally, I also prefer G-Code over G-code, since this somehow goes together as a name - nobody says I am programming in 'G'. But this is not about me, really. And it may also be different in translations, for instance the immune system's T cells are T-Zellen in German (with a hyphen).

> Additionally, I know there are many many instances of 'gcode' throughout the project which I think we can all agree is the worst option :) .

I did a grep on that and decided not to go there. The source code is full of them and should not be changed. In the documentation however, yes, let me do this now.

smoe:
To be continued on https://github.com/LinuxCNC/linuxcnc/pull/1515 .

smoe:
Sorry, I only read this now. Concerning the translated files I thought that this would now all be subject to weblate and I did not want to interfere with what @petterreinholdtsen has already done.
https://github.com/LinuxCNC/linuxcnc/pull/1515 now has a few patches to the other language templates.

hansu:
I tend to say that G-code or G code is the correct form.
But G-Code is felt more often used.
[Siemens](https://new.siemens.com/global/en/markets/machinebuilding/machine-tools/cnc4you/cnc4you-videos/ttt-gen/g-code-samples.html) and [FANUC](https://www.fanucamerica.com/support/training/cnc-robodrill/online-cnc-training-2018/g-code-programming-operation-lathe) uses "G-Code" as far as I found some examples.
For [Heidenhain](https://content.heidenhain.de/doku/tnc_guide/pdf_files/TNC640/34059x-07/diniso/892909-25.pdf) I found a document where they use "G code" and [Haas](https://www.haascnc.com/content/dam/haascnc/en/service/manual/operator/english---mill-ngc---operator's-manual---2017.pdf) seems to use "G-code".

So hard to say what is right or the most used variant :smile:

smoe:
Sounds like an email to write to Heidenhain. The majority-vote was just LinuxCNC-internal. We should likely use the spelling that is used in the standard's description. Does anybody know where/how to access that?

hansu:
Even in the standard they can't decide whether to use "G-code" or "G code":

petterreinholdtsen:
[Hans]
> Even in the standard they can't decide whether to use "G-code" or "G code":
> https://tsapps.nist.gov/publication/get_pdf.cfm?pub_id=823374

According to <URL: https://www.grammarly.com/blog/hyphen/ > "Use a
hyphen in a compound modifier when the modifier comes before the word
it's modifying", which seem to apply here.
<URL: https://linguapress.com/grammar/points/hyphens.htm > on the other
hand do not provide guidance in this case
Then again <URL: https://englishgrammarnotes.com/hyphen/ >
list "Compound Words" and "Prefixes to word" as two categories where
hyphen is to be used, and it could be said to apply.

My gut feeling is to use "G-code" to make it very clear the two words
express one single concept, the G-code.

-- 
Happy hacking
Petter Reinholdtsen

PR #1515, opened 2022-01-18
(g-code | gcode) -> G-code

smoe:
This is mostly in the documentation.
Also removed blanks at end of lines and obvious .adoc indentation errors where I encountered them.


docs/src/docs.xml
```diff
@@ -124,10 +124,10 @@
 <doc name="gui/tklinuxcnc_es.html" title="GUI: TkLinuxCNC"/>
 <doc name="gui/mini_es.html" title="GUI: Mini"/>
 <doc name="common/machining_center_es.html" title="Informations Centre d'usinage"/>
-<doc name="gcode/overview_es.html" title="Aperçu gcode"/>
+<doc name="gcode/overview_es.html" title="Aperçu G-code"/>
 <doc name="gcode/coordinates_es.html" title="Les systèmes de coordonnées"/>
 <doc name="gcode/tool_compensation_es.html" title="Les compensations d'outil"/>
-<doc name="gcode/main_es.html" title="Le langage Gcode EMC2"/>
+<doc name="gcode/main_es.html" title="Le langage G-code EMC2"/>
```
SebKuzminsky, line 130: Since you seem to be in a consistency-generating kind of mood...  "EMC2" is the old name for LinuxCNC.  The EMC(™)(®) corporation claimed that we were infringing on their legally protected name and threatened legal action unless we renamed ourselves, so we did, but possibly we missed some places, like this one.

smoe: I'll have a look. I really just wanted to start reading instead of just correcting bits here and there :) But for weblate to work, well, I panicking, just a bit.

The _es and _fr files are basically gone, so I am relaxed about those :)

smoe: Just checked - it is not so easy as in
```
./src/ladder/classic-ladder.adoc:To use Classic Ladder with HAL without EMC: 
```
which cannot be renamed, right?

SebKuzminsky: The "EMC" in that line should be "LinuxCNC".
http://linuxcnc.org/docs/devel/html/common/emc-history.html#_name_change

PR #1521, opened 2022-01-21
EMC2 -> LinuxCNC

smoe:
A follow-up to a comment from @SebKuzminsky in #1515 on remnants of the original name EMC in our source code and documentation.


SebKuzminsky:
Thanks @smoe !

PR #1531, opened 2022-01-25
qtplasmac: various enhancements

snowgoer540:
-MDI enhancements:
1. alt+return (or enter) opens the MDI
2. return (or enter) with no entry closes the MDI
3. add the ability to 'clear history'
4. add the ability to 'setp' pins/params
5. inhibit M4 commands in MDI
6. history doesn't log entries with errors (thanks phillc54)
7. history doesn't log consecutive duplicate entries (thanks phillc54)
-file editor now opens blank after file_clear
-show table view if editor is closed with no file loaded and no changes
-fix autorepeat turning off when MDI or editor were open and the main 
tab was left and returned to
-edit button is always active
-add hal pin to enable some tabs during run
-docs updates

### DIFF
--- a/share/qtvcp/screens/qtplasmac/qtplasmac_handler.py
+++ b/share/qtvcp/screens/qtplasmac/qtplasmac_handler.py
@@ -1,4 +1,4 @@
-VERSION = '1.221.157'
+VERSION = '1.221.158'
 
 '''
 qtplasmac_handler.py
@@ -172,6 +172,8 @@ class HandlerClass:
         KEYBIND.add_call('Key_Period', 'on_keycall_BNEG')
         KEYBIND.add_call('Key_Greater', 'on_keycall_BNEG')
         KEYBIND.add_call('Key_End', 'on_keycall_END')
+        KEYBIND.add_call('Alt+Key_Return', 'on_keycall_ALTRETURN')
+        KEYBIND.add_call('Alt+Key_Enter', 'on_keycall_ALTRETURN')
         self.axisList = [x.lower() for x in INFO.AVAILABLE_AXES]
         self.systemList = ['G53','G54','G55','G56','G57','G58','G59','G59.1','G59.2','G59.3']
         self.slowJogFactor = 10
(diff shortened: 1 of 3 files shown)


share/qtvcp/screens/qtplasmac/qtplasmac_handler.py
```diff
@@ -493,42 +501,60 @@ def mdi_line_patch(self):
...
             AUX_PRGM.load_halscope()
-        elif intext.upper() == 'CALIBRATION':
+        elif intext.lower() == 'calibration':
             AUX_PRGM.load_calibration()
-        elif intext.upper() == 'PREFERENCE':
+        elif intext.lower() == 'preference':
             self.new_openReturn(os.path.join(self.PATHS.CONFIGPATH, 'qtplasmac.prefs'))
             self.w.gcode_editor.readOnlyMode()
             self.w.preview_stack.setCurrentIndex(2)
+        elif intext.lower() == 'clear history':
```
c-morley, line 526: Might I suggest some of these changes would be a nice addition to the base widget. 
clear history, setp etc sounds like great ideas for the MDI basewidget.
While using uppercase was a choice I made to make sure people didn't accidentally do something they didn't want - I am not against changes -especially if you show that in the real world it works fine.

Please feel free to modify basewidgets if the use case is general enough. I'd be happy to give my opinion on a change if you think you need it, but I think you guys have a good 'feel' for  the idea of Qtvcp's infrastructure .
great ideas here!

snowgoer540: > Might I suggest some of these changes would be a nice addition to the base widget.
clear history, setp etc sounds like great ideas for the MDI basewidget.

Actually, that's a great idea.  Phill patched the MDI_LINE function a while back because M3, M4, and M5 commands in the MDI cause QtPlasmaC angst.  Due to tunnel vision while working in the handler, I admit I didn't think about pushing this back to the base widget as well.  Sorry about that.

>  While using uppercase was a choice I made to make sure people didn't accidentally do something they didn't want - I am not against changes -especially if you show that in the real world it works fine.

The change from intext.upper() to intext.lower() was just for consistency because our handler primarily compares strings in lower case.  I found myself writing "if" statements that used both upper() and lower() and that just seemed "wrong". 

I'd be honored to add these to the base widget, I will work on a PR in the next few days to add these features, and try harder to remember in the future :)

PR #1559, opened 2022-02-01
halshow: keep window on top

phillc54:
The argument _--ontop true_ will attempt to keep the halshow window above other windows.

### DIFF
--- a/docs/src/hal/halshow.adoc
+++ b/docs/src/hal/halshow.adoc
@@ -37,7 +37,7 @@ Notes:
 ----
 
 Example to limit number of decimal points for floats
-and used a file named my.halshow in the current directory:
+and use a file named my.halshow in the current directory:
 
 ----
 $ halshow  --fformat "%.5f" ./my.halshow
--- a/tcl/bin/halshow.tcl
+++ b/tcl/bin/halshow.tcl
@@ -43,6 +43,7 @@ set x [expr ($xmax - $masterwidth )  / 2 ]
 set y [expr ($ymax - $masterheight )  / 2]
 wm geometry . "${masterwidth}x${masterheight}+$x+$y"
 wm  minsize . $masterwidth $masterheight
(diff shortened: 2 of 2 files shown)


c-morley:
This is a great idea - but i think ot should be default like halmeter.

phillc54:
It has always frustrated me that it falls beneath other windows.
So change it to on top is default and have the option to revert it to the current behaviour?

c-morley:
You can always manually set it not-on-top in the window manager.
I wouldn't even bother with a switch - just set it always on top.
I would have done it a long time ago when I did halmeter, but tcl deters me.

phillc54:
> You can always manually set it not-on-top in the window manager.

I didn't know that...

c-morley:
right click the window header. Different options on different window managers but always-on-top is usually there.

phillc54:
Thanks,  that was embarrassing, it is now quite obvious.
I will rework it.

c-morley:
perfect.

hansu:
Yeh, great idea, thanks!

dwrobel:
The downside of it is, that now you cannot use the `SetVal` until the `Always on Top` is disabled - at least on the `GNOME 41.3`. It might be non-obvious behaviour for someone.

phillc54:
Oops, I messed the entrybox... fixed.

hansu:
Good catch Damian! And thanks Phillip for immediate fixing!

dwrobel:
On 01d26623a528b9a7d7553554d423d15027fe2446, when I pressed `SetVal` the `User input` window is created, but beneath the `HAL Show` window. The workaround is to either press `SetVal` for the second time or move the `HAL Show` window to uncover the `User input` window.

hansu:
I can't confirm that behaviour. On my Debian 10 with xfce the _User input_ window is shown after the first click.

PR #1561, opened 2022-02-01
gmoccapy: center button labels with property instead of spaces

hansu:
The text of some multi-line buttons was manually centred with spaces. But this is not helpful for the translated labels.
So I removed the spaces and added a label as child to be able to set the 'justify' property to 'center'.

### DIFF
--- a/lib/python/gladevcp/offsetpage.glade
+++ b/lib/python/gladevcp/offsetpage.glade
@@ -439,8 +439,14 @@
             <property name="visible">True</property>
             <child>
               <object class="GtkButton" id="zero_g92_button">
-                <property name="label" translatable="yes">Zero
+                <child>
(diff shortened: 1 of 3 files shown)


c-morley:
hansu - would you be able to assess  #1548 - modern icon for Gmoccapy?

hansu:
> hansu - would you be able to assess #1548 - modern icon for Gmoccapy?

Yeah I can have a look. But I don't think I'll have time for that in the next few days.

c-morley:
Thank you. I like support the changes, ust want to make sure it doesn't breal work you and others are doing in gmoccapy/gladevcp

hansu:
> Thank you. I like support the changes, just want to make sure it doesn't break work you and others are doing in gmoccapy/gladevcp

:heavy_check_mark:

PR #1562, opened 2022-02-02
Move forced newlines out of translatable text

petterreinholdtsen:
There is no use asking tranlators to remember these newlines, as
they always need to be there in any language.

Also removed some useless trailing spaces.

Based on comment in Weblate regarding the English text.


SebKuzminsky:
I think it would be clearer to put the "untranslatable newline" on the line it terminates, like this as for example:
```
message = _("**** GMOCCAPY INI Entry ****") + "\n"
message += _("user mode selected")
```
But this is already an improvement, so thank you!

PR #1567, opened 2022-02-03
Add gmoccapy.pot to target 'pofiles'

hansu:
Update also the `gmoccapy.pot` with "make pofiles".


smoe (approved):

smoe:
I would just also add pofiles to .PHONY .

PR #1571, opened 2022-02-06
()s added for op precedence, small bugs

smoe:
Since the ? is evaluated prior to the || to the left of it, the ()s around the left parts were needed. The problems were mostly harmless, except for when these are not :o)  

I was not completely sure about when the pmErrno should be set and when not. In my understanding this should also be set if the function already returns a regular value, e.g. for the square root of negative values.

In line 815 this catches a possible div0.

In line 1043 I added ()s for the check if the vector is normed. This may have been quite a tricky bug hidden in there.

1055 is just  a typo fixed, but one with semantics.

Some newlines were added for consistency.

Sorry for submitting this all as one patch, I truly set out only to add the ()s.

Steffen

### DIFF
--- a/src/libnml/posemath/_posemath.c
+++ b/src/libnml/posemath/_posemath.c
@@ -35,7 +35,7 @@
 #include "sincos.h"
 
 /* global error number */
-int pmErrno = 0;
+int pmErrno = PM_OK;
 
 #ifdef PM_PRINT_ERROR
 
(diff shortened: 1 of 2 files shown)


src/libnml/posemath/_posemath.c
```diff
@@ -1052,11 +1068,11 @@ int pmCartCartProj(PmCartesian const * const v1, PmCartesian const * const v2, P
     
     r1 = pmCartCartDot(v1, v2, &d12);
     r2 = pmCartCartDot(v2, v2, &d22);
-    if (!(r1 || r1)){
+    if (!(r1 || r2)){
```
SebKuzminsky, line 1074: Nice catch!

SebKuzminsky (approved):

Issue #1574, opened 2022-02-06
Integrate cppcheck in continuous build tests


smoe:
Just a proposal that I would help to implement for 2.9+0.1 if this is considered positive for the project.
There are some automated code-checkers that (at times) are surprisingly helpful. A very good stimulus is "cppcheck" - not perfect, but the one or other memory leak or undefined variable or plain error should be spotted with it.

I have tried this in src/hal and found some interesting bits like
```
src/hal/drivers/opto_ac5.c:432:13: error: Signed integer overflow for expression '1<<(31-i)'. [integerOverflow]
      mask=1<<(31-i);
```
and the following patch to it
```
$ vim !$
vim hal/drivers/opto_ac5.c
$ cppcheck hal/drivers/opto_ac5.c
Checking hal/drivers/opto_ac5.c ...
$ git diff !$
git diff hal/drivers/opto_ac5.c
diff --git a/src/hal/drivers/opto_ac5.c b/src/hal/drivers/opto_ac5.c
index 5d81aa5be3..19bdb40ebb 100644
--- a/src/hal/drivers/opto_ac5.c
+++ b/src/hal/drivers/opto_ac5.c
@@ -429,7 +429,7 @@ Device_DigitalOutWrite(void *arg, long period)
                        pDigital = &pboard->port[portnum].io[23];//one before what we want to check
                        for (i = 0;i < 2;i++)
                                {
-                                       mask=1<<(31-i);
+                                       mask = (unsigned int) 1 << (31-i);
                                        pDigital++;
                                
                                        if ( *(pDigital->pValue) ==0 ) {        pins |= mask;       }   
```
The bug (if this effectively is a bug, anyway) was that integers are signed by default.

SebKuzminsky:
I'm in favor of increased use of automated code checkers in our CI system.

PR #1575, opened 2022-02-06
int (also literals as it seems) are signed by default

smoe:
These patches eliminate an error found by cppcheck:
src/hal/drivers/opto_ac5.c:432:13: error: Signed integer overflow for expression '1<<(31-i)'. [integerOverflow]
      mask=1<<(31-i);

This is just a pilot PR to test my thought to have cppcheck integral to the continuous integration (https://github.com/LinuxCNC/linuxcnc/issues/1574).

Many thanks!
Steffen

### DIFF
--- a/src/hal/drivers/hal_gm.c
+++ b/src/hal/drivers/hal_gm.c
@@ -1126,7 +1126,6 @@ read(void *arg, long period)
 {
     	gm_device_t	*device = (gm_device_t *)arg;
     	card	*pCard = device->pCard;
-    	int		i;
 	hal_u32_t temp;
 	
       //basic card functionality: watchdog, switches, estop
@@ -1134,10 +1133,10 @@ read(void *arg, long period)
 
       //read parallel IOs
 	temp=pCard->gpio;
-    	for(i = 0; i < 32; i++)
+    	for(unsigned int i = 0; i < 32; i++)
 	{
-		*(device->gpio[i].in) = (hal_bit_t)((temp & (0x0001 << i)) == 0 ? 0 : 1);
-		*(device->gpio[i].inNot) = (hal_bit_t)((temp & (0x0001 << i)) == 0 ? 1 : 0);
+		*(device->gpio[i].in) = (hal_bit_t)((temp & ((unsigned int) 1 << i)) == 0 ? 0 : 1);
+		*(device->gpio[i].inNot) = (hal_bit_t)((temp & ((unsigned int) 1 << i)) == 0 ? 1 : 0);
 	}
 
       //Read Encoders
--- a/src/hal/drivers/hal_gm.c
+++ b/src/hal/drivers/hal_gm.c
@@ -1126,14 +1126,15 @@ read(void *arg, long period)
 {
     	gm_device_t	*device = (gm_device_t *)arg;
     	card	*pCard = device->pCard;
+	unsigned int i;
 	hal_u32_t temp;
-	
+
       //basic card functionality: watchdog, switches, estop
 	card_mgr(arg, period);
 
       //read parallel IOs
 	temp=pCard->gpio;
-    	for(unsigned int i = 0; i < 32; i++)
+    	for(i = 0; i < 32; i++)
 	{
 		*(device->gpio[i].in) = (hal_bit_t)((temp & ((unsigned int) 1 << i)) == 0 ? 0 : 1);
 		*(device->gpio[i].inNot) = (hal_bit_t)((temp & ((unsigned int) 1 << i)) == 0 ? 1 : 0);
--- a/src/emc/usr_intf/emcsched.cc
+++ b/src/emc/usr_intf/emcsched.cc
@@ -190,7 +190,7 @@ void SchedEntry::setTool(int t) {
 
 static void crcInit() {
   crc rmdr;
-  int i;
+  unsigned int i;
   uint32_t bit;
 
   for (i = 0; i < 256; ++i)
--- a/src/hal/drivers/opto_ac5.c
+++ b/src/hal/drivers/opto_ac5.c
@@ -402,6 +402,7 @@ Device_DigitalOutWrite(void *arg, long period)
     board_data_t			*pboard = (board_data_t *)arg;
     DigitalPinsParams			*pDigital;
     int					portnum=0;
+    unsigned int			i,j;
     unsigned long			pins, offset=DATA_WRITE_OFFSET_0,mask;
 
     // For each port.
@@ -412,7 +413,7 @@ Device_DigitalOutWrite(void *arg, long period)
 			pins=0;
 
    			// For each pin.
-			for(unsigned int j = 0; j < 24; j++, pDigital++)
+			for(j = 0; j < 24; j++, pDigital++)
 			{
 				if ((pboard->port[portnum].mask & mask) !=0) //is it an output?
 				{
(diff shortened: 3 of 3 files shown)


SebKuzminsky:
I'm in favor of the syntax change in this commit, ie of moving the variable definition from the function scope into the smaller scope of the for loop, with definition in the for loop itself.  However, this conflicts with a bug in our RTAI debs for Buster that @dngarrett pointed out recently:

https://github.com/LinuxCNC/linuxcnc/commit/3768a3c533cf138a41056a2ca8286eca6e65e894

The ideal way to fix this would be to fix that version of our RTAI packages so that we can use a modern (well, a less ancient) C standard.  I think @andypugh or @NTULINUX prepared those RTAI debs, do you have input for us on this topic?

smoe:
@SebKuzminsky Together with the additional instance if a (signed int) << shifted I have also moved the initialization of i and j back up to the start of the function.

@andypugh The effect of compiler-whatever options on RTAI is interesting, though. Not that I would understand this. Link-time optimization would be close to my heart since this reduces the memory footprint which I could imagine to reduce the jitter.

NTULINUX:
The RTAI CFLAGS outside of LIBM (musl math libm math library) are:

`-O2 -march=x86-64 -fomit-frame-pointer -pipe`

`-fomit-frame-pointer` helps decrease code size, maybe increase performance a tad, but makes debugging a bit more difficult. RTAI is pretty much hit-or-miss anyway, and frame pointers won't help you lol.

`-pipe` allows for faster compilation, does not affect generated code at all.

For the math library, it is the above CFLAGS but with these appended:

`-D__IN_RTAI__ -mcmodel=kernel -nostdinc -ffreestanding -fno-pic -fno-builtin -fno-common -fno-unwind-tables -fno-asynchronous-unwind-tables -frounding-math -mpreferred-stack-boundary=4 -mstackrealign -mfpmath=sse -msse -msse2`

`-D__IN_RTAI__` pertains to autoconf/Kconfig stuff, don't worry about that really:

`
dnl CAUTION: We need to have the CONFIG_RTAI_XX symbols always defined when
dnl the configuration header is read, but we want the Autoconf-produced
dnl symbols to be defined only when compiling RTAI. This way, we won't
dnl pollute the namespace with the latter when our configuration header is
dnl indirectly included by a client application. To achieve this, we ask
dnl autoheader to produce the following header structure:
dnl #define CONFIG_XX
dnl #define CONFIG_XX ...
dnl #ifdef __IN_RTAI__
dnl <Autoconf-defined symbols>
dnl #endif /* __IN_RTAI__ */
dnl This is quite a hack since we have to rely on the fact that
dnl all Autoconf-generated symbols are lexicographically sorted
dnl after CONFIG_RTAI_XX ones, but, well...
`

`-mcmodel=kernel` is generate code for kernel model.

`-fno-pic` is because you cannot build position independent code (PIC) in kernel space. PIC is a security feature, but does not work in cases such as the Linux kernel, or certain assembly instructions that is not PIC-friendly.

https://en.wikipedia.org/wiki/Position-independent_code

` -nostdinc -ffreestanding -fno-builtin -fno-common` These basically just mean "Don't use anything, from anyone, outside of this scope. Don't use compiler built-in functions, don't grab things from anywhere else, this is a free standing environment, don't even try looking outside the box."

`-fno-unwind-tables -fno-asynchronous-unwind-tables` are to reduce generated code size, may improve performance. Again, makes debugging more difficult but these probably won't help you anyway.

`-frounding-math -mpreferred-stack-boundary=4 -mstackrealign -mfpmath=sse -msse -msse2` These all work together to satisfy the build and runtime of floating point instructions. `-mstackrealign` makes sure that RTAI modules are aligned to kernel space, basically. Helps keeps things in order. `mfpmath=sse` `-msse` `-msse2` are x86 extensions which should be used to aid in floating point precision.

https://en.wikipedia.org/wiki/Streaming_SIMD_Extensions

SebKuzminsky:
@NTULINUX Thanks for that run-down.

Do you have any idea about the source of the `-std=gnu89` that @dngarrett reports here?  https://github.com/LinuxCNC/linuxcnc/commit/3768a3c533cf138a41056a2ca8286eca6e65e894#commitcomment-65892820 
It seems to somehow make it into LinuxCNC's CFLAGS when building on RTAI.

The ancient Wheezy RTAI does not have this problem, it builds LinuxCNC with `-std=gnu99` which accepts this syntax, eg: http://buildbot.linuxcnc.org/buildbot-admin/builders/1401.rip-wheezy-rtai-i386/builds/6346/steps/compile/logs/stdio

NTULINUX:
My guess is:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=51b97e354ba9fce1890cf38ecc754aa49677fc89

smoe:
This (too long) video https://www.youtube.com/watch?v=-G1FuEQqxVI just informed me that apparently the Linux kernel is about to upgrade to C99 .

andypugh:
Where are we with this? Master/2.9 is currently not targeting any OS older than Buster, so we may have the opportunity to drop some legacy limitations. 
I am still committed to keeping an RTAI kernel available and supported, but now that LinuxCNC is included in (future) Debian in the preempt-rt version we probably should not consider RTAI-specific issues to be blocking.

smoe:
I think this is fine. I had undone my compiler-incompatibility as mentioned in https://github.com/LinuxCNC/linuxcnc/pull/1575#issuecomment-1033637795 .

src/hal/drivers/hal_gm.c
```diff
@@ -1126,18 +1126,18 @@ read(void *arg, long period)
 {
     	gm_device_t	*device = (gm_device_t *)arg;
     	card	*pCard = device->pCard;
-    	int		i;
 	hal_u32_t temp;
-	
+	unsigned int i;
```
petterreinholdtsen, line 1130: Why do you change location of the declaration?

smoe: I wish you had asked me a few months ago. Looking at it now I feel like i may have ordered the variables by their "localness", i.e. these are declared function-wide but used over different regions of the code and i I may have perceived as the most local.

petterreinholdtsen:
[Steffen Möller]
> I wish you had asked me a few months ago. Looking at it now I feel
> like i may have ordered the variables by their "localness", i.e. these
> are declared function-wide but used over different regions of the code
> and i I may have perceived as the most local.

Is the idea that the compiler can not figure this out on its own?  Tried
to consult docs/src/code/style-guide.adoc to see if it had something
useful on the topic, but alas.

-- 
Happy hacking
Petter Reinholdtsen

smoe:
I likely have edited it back in after having tried this within the for loop itself and forgot where I got it from. While I tend to think this is an improvement as it is, this is not meant to have an effect on the compiler but solely beween us devs.

smoe:
Brought the order of variables back into the original order as suggested in https://github.com/LinuxCNC/linuxcnc/pull/1575#issuecomment-1180432626

petterreinholdtsen:
The updated pull request look good to me, even if it fail to clean up
some already existing white space issues (reported as red in 'git diff')
against marster in the lines it touces.  The code in question compile
without any compiler warnings, and all tests succeed.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
The change look good to me.  The current failing build tests are due to a transient error on github, and not related to the patch.  I guess a rebase and rebuild will show success.

jepler:
a matter of style "1u" also works when you want an unsigned literal value.

PR #1581, opened 2022-02-08
Idea to integrate cppcheck

smoe:
There may be better ideas how to address this. The plan is to iteratively add all directories for which cppcheck works. At some point the coverage should then be good enough to simplify by calling it on larger subtrees.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -261,27 +261,8 @@ jobs:
         set -x
         git fetch --recurse-submodules=no https://github.com/linuxcnc/linuxcnc refs/tags/*:refs/tags/*
         sudo apt-get -y install cppcheck shellcheck
-    - name: C
-      continue-on-error: true
-      run: |
-        echo -n "I (1/4): checking HAL folders with C code only from dir "; pwd
-        for d in src/hal/classicladder src/hal/components; do (cd $d && cppcheck -j $(nproc) --language=c --force *.h *.c); done
-        echo -n "I (2/4): checking EMC folders with C code only from dir "; pwd
-        for d in src/emc/motion-logger src/emc/tp; do (cd $d && cppcheck -j $(nproc) --language=c --force *.h *.c); done
-        # problematic: src/hal/drivers src/hal/user_comps src/hal/utits
-    - name: C++
-      continue-on-error: true
-      run: |
-        echo -n "I (3/4): checking EMC folders with C++ code only from dir "; pwd
-        for d in src/emc/canterp src/emc/ini src/emc/pythonplugin src/emc/tooldata; do (cd $d && cppcheck -j $(nproc) --language=c++ --force *.hh *.cc); done
-        # problematic iotask src/emc/sai src/emc/task
-    - name: C and C++ in same dir
-      continue-on-error: true
-      run: |
-        echo -n "I (4/4): checking EMC folders with both C and C++ code from dir "; pwd
-        for d in src/emc/rs274ngc; do (cd $d && cppcheck -j $(nproc) --language=c --force *.h *.c && cppcheck -j $(nproc) --language=c++ --force *.hh *.cc); done
-        # problematic src/emc/kinematics src/emc/nml_intf src/emc/usr_intf
+        scripts/cppcheck.sh
     - name: Shellcheck
       continue-on-error: true
       run: |
-        find . -name "*.sh" | xargs -r -L 1 shellcheck
+        scripts/shellcheck.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,6 +256,7 @@ jobs:
         submodules: true
         fetch-depth: 0
     - name: Perform Source Code checks that were successful in the past
+      continue-on-error: true
       run: |
         set -x
         git fetch --recurse-submodules=no https://github.com/linuxcnc/linuxcnc refs/tags/*:refs/tags/*
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -260,24 +260,28 @@ jobs:
       run: |
         set -x
         git fetch --recurse-submodules=no https://github.com/linuxcnc/linuxcnc refs/tags/*:refs/tags/*
-        sudo apt-get -y install cppcheck
-        
+        sudo apt-get -y install cppcheck shellcheck
     - name: C
+      continue-on-error: true
       run: |
         echo -n "I (1/4): checking HAL folders with C code only from dir "; pwd
         for d in src/hal/classicladder src/hal/components; do (cd $d && cppcheck -j $(nproc) --language=c --force *.h *.c); done
         echo -n "I (2/4): checking EMC folders with C code only from dir "; pwd
         for d in src/emc/motion-logger src/emc/tp; do (cd $d && cppcheck -j $(nproc) --language=c --force *.h *.c); done
         # problematic: src/hal/drivers src/hal/user_comps src/hal/utits
-        
     - name: C++
+      continue-on-error: true
       run: |
         echo -n "I (3/4): checking EMC folders with C++ code only from dir "; pwd
         for d in src/emc/canterp src/emc/ini src/emc/pythonplugin src/emc/tooldata; do (cd $d && cppcheck -j $(nproc) --language=c++ --force *.hh *.cc); done
         # problematic iotask src/emc/sai src/emc/task
-
     - name: C and C++ in same dir
+      continue-on-error: true
       run: |
         echo -n "I (4/4): checking EMC folders with both C and C++ code from dir "; pwd
         for d in src/emc/rs274ngc; do (cd $d && cppcheck -j $(nproc) --language=c --force *.h *.c && cppcheck -j $(nproc) --language=c++ --force *.hh *.cc); done
         # problematic src/emc/kinematics src/emc/nml_intf src/emc/usr_intf
+    - name: Shellcheck
+      continue-on-error: true
+      run: |
+        find . -name "*.sh" | xargs -r -L 1 shellcheck
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,3 +247,36 @@ jobs:
         set -e
         set -x
         eatmydata apt-get --yes --quiet install ../*.deb
+
+  cppcheck:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+        fetch-depth: 0
+    - name: Perform Source Code checks that were successful in the past
+      run: |
+        set -x
+        git fetch --recurse-submodules=no https://github.com/linuxcnc/linuxcnc refs/tags/*:refs/tags/*
+        sudo apt-get -y install cppcheck
+        
+    - name: C
+      run: |
+        echo -n "I (1/4): checking HAL folders with C code only from dir "; pwd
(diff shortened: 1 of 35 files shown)


smoe:
So, here my shot at https://github.com/LinuxCNC/linuxcnc/issues/1574 . 
This is admittedly much better than I had expected. I ran into one [false positive](https://trac.cppcheck.net/ticket/10047) that is trivial to see by eye so I am not worried. Also
```
stashf_wrap.h:26:5: error: syntax error: keyword 'if' is not allowed in global scope [syntaxError]
    if(result < 0) return result;
```
is somewhat too unconventional for cppcheck (and it is weird, indeed, the .h ending should maybe change to .include?)

But some limitations of cppcheck are easily outweighed by basically impossible to find gems like:

```
emcsh.cc:3338:4: error: Read and write operations without a call to a positioning function (fseek, fsetpos or rewind) or fflush in between result in undefined behaviour. [IOWithoutPositioning]
   fputc(77, inFile); // Request data resent
```
or 
```
saicanon.cc:325:30: error: Out of bounds access of nurbs_control_points, index 'nurbs_control_points.size()' is out of bounds. [containerOutOfBoundsIndexExpression]
  _sai._program_position_x = nurbs_control_points[nurbs_control_points.size()].X;
```

petterreinholdtsen:
This sound like a good idea.  Anything that can help us find bugs automatically is a good idea.

smoe:
I must admit that I do not really know how to integrate this properly in our build infrastructure. Run it on all the source tree and somewhere maintain a list of known issues that we do not want to fix?

petterreinholdtsen:
[Steffen Möller]
> I must admit that I do not really know how to integrate this properly
> in our build infrastructure. Run it on all the source tree and
> somewhere maintain a list of known issues that we do not want to fix?

I suspect src/Makefile would need an update with TARGETS += cppcheck or
similar, and the CI scripts and d/control need a new build dependency.

-- 
Happy hacking
Petter Reinholdtsen

smoe:
Note to self by our friendly Jitsie-committee: Rebase! Introduce variables! 
And more !!!s

jepler:
Hi! I like the general idea, but may I suggest to put the body of the logic in a script (a .sh file somewhere) so that not only can actions run the script but so can a developer locally.  Also, the branch has a conflict now.

rene-dev:
good idea, the Norwegian linuxcnc meeting agrees with @jepler

heeplr:
i've sent [a quick PR](https://github.com/smoe/linuxcnc/pull/3) so lots of errors won't trigger CI for now so this could be merged without breaking stuff.

heeplr:
@smoe can you rebase your branch/resolve conflicts? there were _a lot_ of commits since back then ;)

rene-dev:
best way ist using compile_commands.json
https://clang.llvm.org/docs/JSONCompilationDatabase.html

```
make clean
bear -- make -j
cppcheck --project=compile_commands.json
```

smoe:
> @smoe can you rebase your branch/resolve conflicts? there were _a lot_ of commits since back then ;)

Rebased. Also, I split the cppcheck invocations into multiple steps.  I do not expect that build error to be triggered by the rebase. Had a look at what cppcheck brings, and, seems like some of them are real:
```
44/46 files checked 76% done
interp_convert.cc:5900:14: error: Uninitialized variable: beta [legacyUninitvar]
        if ((beta < -small) || (beta > (M_PIl + small))) {
             ^
interp_convert.cc:5895:45: error: Uninitialized variable: gamma [legacyUninitvar]
        end_x = (px + (radius * cos(alpha + gamma)));
```
Yes, for the else branch beta is not set. Would not know how to set it, admittedly. 0.0?

```
nurbs_additional_functions.cc:969:23: error: When ii==nurbs_costant.size(), nurbs_costant[ii] is out of bounds. [stlOutOfBounds]
     a = nurbs_costant[ii]; //uj
                      ^
```
Yes.

```
stepgen.c:1207:55: error: Array 'user_step_type[1]' accessed at index 17, which is out of bounds. [arrayIndexOutOfBounds]
        master_lut[USER_STEP_TYPE][i] = user_step_type[i];
                                                      ^
stepgen.c:1208:31: error: Array 'user_step_type[1]' accessed at index 17, which is out of bounds. [arrayIndexOutOfBounds]
 used_phases |= user_step_type[i];
              ^
```
I am tempted to call this a false positive. Problem with designated initializers?

Anyway, first it needs to build again :-)

jepler:
```c
950                 for(unsigned ii=0; ii<= nurbs_costant.size(); ii=ii+6)
```
this loop check is probably just wrong, it's almost always an error to loop to `<= size`

jepler:
```c
 308 #define MAX_CYCLE 18
...
 319 int user_step_type[] = { [0 ... MAX_CYCLE-1] = -1 };
...
1206     for(i=0; i < MAX_CYCLE && user_step_type[i] != -1; i++) {
1207         master_lut[USER_STEP_TYPE][i] = user_step_type[i];
1208         used_phases |= user_step_type[i];
1209     }
```

I checked [GCC documentation for this gcc extension](https://gcc.gnu.org/onlinedocs/gcc-14.2.0/gcc/Designated-Inits.html) and I think this is a false positive. The range of this kind of initializer has an inclusive upper bound, so user_step_type[17] is initialized to -1 and is valid to access. 

> To initialize a range of elements to the same value, write `[FIRST ... LAST] = VALUE`.  This is a GNU extension.  For example,
> ```c
>     int widths[] = { [0 ... 9] = 1, [10 ... 99] = 2, [100] = 3 };
> ```
> If the value in it has side effects, the side effects happen only once, not for each initialized field by the range initializer.

This might be a false positive that occurs because the GCC extension is not properly implemented in cppcheck. Adding the explicit bound might fix it?
```c
int user_step_type[MAX_CYCLE] = { [0 ... MAX_CYCLE-1] = -1 };
```

heeplr:
> someone interested could report it

Thanks a lot for the investigation! Does one get a github notification when highlighted in a random issue? Then @danmar might see this.
My sf account is buried somewhere but I'll try to report it.

smoe:
> The bit about undefined beta/gamma:
> ...
> ERS is a macro which always ends with "return;", so the `end_x` line which depends on the value of `gamma` is unreachable from the else branch:

I was not aware of the ERS macro. Fooled me, too. False positive, I tend to think.

danmar:
> .... 
> reduced cppcheck test case for false positive legacyUninitvar -- someone interested could report it at [their issue tracker](https://trac.cppcheck.net/). A long time ago they fixed [a simpler case](https://trac.cppcheck.net/ticket/936), but not when the return is in a do/while.

I have reported https://trac.cppcheck.net/ticket/13227 thanks!

heeplr:
> Should we not possibly anyway just initialize beta and gamma with 0.0 and add a comment that the value will always be reset if used?

imho it probably wouldn't hurt. (Aeons ago I learned "All variable definitions which aren't static need to be initialized" to avoid nasty bugs. But not sure if that still holds.)

smoe:
Should this have been a separate PR? Possibly, sorry for that. I have added shellcheck and applied it on *.sh in the archive. This seems to work just fine, have also added some commits that address some of the shellcheck's concerns. These were morely about missing quotes. Missing checks if an operation was successful I have addressed by setting bash's "-e" flag. Most problems were in the test scripts, so I am a bit curious if those still work now :)

smoe:
> > Should we not possibly anyway just initialize beta and gamma with 0.0 and add a comment that the value will always be reset if used?
> 
> imho it probably wouldn't hurt. (Aeons ago I learned "All variable definitions which aren't static need to be initialized" to avoid nasty bugs. But not sure if that still holds.)

Done in https://github.com/LinuxCNC/linuxcnc/pull/1581/commits/4c4208047975ed7e906ccebf2a15f7869081a24e .

heeplr:
@smoe just want to say that i appreciate all your effort so much. This will add a good amount of solidness to the codebase.

If we can at some point remove the "continue-on-error" to make CI fail on warnings, less efforts needs to be taken when merging PRs aswell. A bright future lies ahead. All hail to static code analysis! ;)

smoe:
The next upload will have the separate scripts that @jepler and the Oslo folks suggested. Testing it, I was directed to ./src/hal/classicladder/edit_gtk.c
```
Checking edit_gtk.c ...
edit_gtk.c:423:76: error: Uninitialized struct member: ToolBarEle.VarType [uninitStructMember]
    DrawElement( cr, 0, 0, PIXELS_SIZE_IN_TOOLBAR, PIXELS_SIZE_IN_TOOLBAR, ToolBarEle, TRUE);
                                                                           ^
edit_gtk.c:423:76: error: Uninitialized struct member: ToolBarEle.VarNum [uninitStructMember]
    DrawElement( cr, 0, 0, PIXELS_SIZE_IN_TOOLBAR, PIXELS_SIZE_IN_TOOLBAR, ToolBarEle, TRUE);
                                                                           ^
edit_gtk.c:423:76: error: Uninitialized struct member: ToolBarEle.IndexedVarType [uninitStructMember]
    DrawElement( cr, 0, 0, PIXELS_SIZE_IN_TOOLBAR, PIXELS_SIZE_IN_TOOLBAR, ToolBarEle, TRUE);
                                                                           ^
edit_gtk.c:423:76: error: Uninitialized struct member: ToolBarEle.IndexedVarNum [uninitStructMember]
    DrawElement( cr, 0, 0, PIXELS_SIZE_IN_TOOLBAR, PIXELS_SIZE_IN_TOOLBAR, ToolBarEle, TRUE);
                                                                           ^
edit_gtk.c:423:76: error: Uninitialized struct member: ToolBarEle.DynamicInput [uninitStructMember]
    DrawElement( cr, 0, 0, PIXELS_SIZE_IN_TOOLBAR, PIXELS_SIZE_IN_TOOLBAR, ToolBarEle, TRUE);
                                                                           ^
edit_gtk.c:423:76: error: Uninitialized struct member: ToolBarEle.DynamicState [uninitStructMember]
    DrawElement( cr, 0, 0, PIXELS_SIZE_IN_TOOLBAR, PIXELS_SIZE_IN_TOOLBAR, ToolBarEle, TRUE);
                                                                           ^
edit_gtk.c:423:76: error: Uninitialized struct member: ToolBarEle.DynamicVarBak [uninitStructMember]
    DrawElement( cr, 0, 0, PIXELS_SIZE_IN_TOOLBAR, PIXELS_SIZE_IN_TOOLBAR, ToolBarEle, TRUE);
                                                                           ^
edit_gtk.c:423:76: error: Uninitialized struct member: ToolBarEle.DynamicOutput [uninitStructMember]
    DrawElement( cr, 0, 0, PIXELS_SIZE_IN_TOOLBAR, PIXELS_SIZE_IN_TOOLBAR, ToolBarEle, TRUE);
                                                                           ^
edit_gtk.c:423:76: error: Uninitialized variables: ToolBarEle.VarType, ToolBarEle.VarNum, ToolBarEle.IndexedVarType, ToolBarEle.IndexedVarNum, ToolBarEle.DynamicInput, ToolBarEle.DynamicState, ToolBarEle.DynamicVarBak, ToolBarEle.DynamicOutput [uninitvar]
    DrawElement( cr, 0, 0, PIXELS_SIZE_IN_TOOLBAR, PIXELS_SIZE_IN_TOOLBAR, ToolBarEle, TRUE);
                                                                           ^
```
The "error" is a bit drastic, not? Shall I find out how to override this one?

This would be the then last error, i.e. cppcheck could become enforced for the folders it is currently run on.

andypugh:
Where are we with this? Do you want to merge it?

smoe:
> Where are we with this? Do you want to merge it?

I just skimmed through the diff again and it looks fine. Yes, please merge it.

I guess we then just skim through the build logs from time to time and see what we want to do.

PR #1584, opened 2022-02-09
cppcheck warned about a comparison of pointers from different objects

smoe:
This may not be overly dramatical since hidden in a #debug conditional, but for the sake of clarity and to add src/emc/nml_intf to the whitelisted set of directories proposed in #1581, this change may be fine.

```
interpl.cc:88:36: error: Comparing pointers that point to different objects [comparePointers]
 ((void *) &temp_node.line_number) >
                                   ^
interpl.hh:26:9: note: Variable declared here.
    int line_number;  // line number it was on
        ^
interpl.cc:88:12: note: Address of variable taken here.
 ((void *) &temp_node.line_number) >
           ^
interpl.hh:37:7: note: Variable declared here.
 char commandbuf[MAX_NML_COMMAND_SIZE]; // the NML command;
      ^
interpl.cc:89:12: note: Address of variable taken here.
 ((void *) &temp_node.command.commandbuf)) {
           ^
interpl.cc:88:36: note: Comparing pointers that point to different objects
 ((void *) &temp_node.line_number) >
```

### DIFF
--- a/src/emc/nml_intf/interpl.cc
+++ b/src/emc/nml_intf/interpl.cc
@@ -85,8 +85,8 @@ int NML_INTERP_LIST::append(NMLmsg * nml_msg_ptr)
 #ifdef DEBUG_INTERPL
     if (sizeof(temp_node) < MAX_NML_COMMAND_SIZE + 4 ||
 	sizeof(temp_node) > MAX_NML_COMMAND_SIZE + 16 ||
-	((void *) &temp_node.line_number) >
-	((void *) &temp_node.command.commandbuf)) {
+	((uintptr_t) &temp_node.line_number) >
+	((uintptr_t) &temp_node.command.commandbuf)) {
 	rcs_print_error
 	    ("NML_INTERP_LIST::append : assumptions about NML_INTERP_LIST_NODE have been violated.");
 	return -1;


src/emc/nml_intf/interpl.cc
```diff
@@ -85,8 +85,8 @@ int NML_INTERP_LIST::append(NMLmsg * nml_msg_ptr)
 #ifdef DEBUG_INTERPL
     if (sizeof(temp_node) < MAX_NML_COMMAND_SIZE + 4 ||
 	sizeof(temp_node) > MAX_NML_COMMAND_SIZE + 16 ||
-	((void *) &temp_node.line_number) >
-	((void *) &temp_node.command.commandbuf)) {
+	((unsigned long) (void *) &temp_node.line_number) >
+	((unsigned long) (void *) &temp_node.command.commandbuf)) {
```
dwrobel, line 89: From the language perspective, you cannot safely cast from `void *` to `unsigned long` as there is no guarantee that `unsigned long` will be capable of holding a value of pointer.

If you really want to cast `void *` to `integer` type please consider using `uintptr_t` type.

smoe:
Technically according to https://docs.oracle.com/cd/E19253-01/817-6223/chp-typeopexpr-2/index.html for 32 and 64 bit the unsigned long would have been fine but your suggestion (in the second table) reflects the semantics and is hence better. Have many thanks for spotting that! I rebased to master and updated as suggested.

dwrobel:
>64 bit the unsigned long

You were refering to _"Solaris Dynamic Tracing Guide"_, while the [ c++ specification](https://eel.is/c++draft/basic.fundamental#:type,unsigned_long) doesn't guarantee that, see the `Minimum width N` for `long int`type, which only guarantees `32`.

PR #1585, opened 2022-02-10
Fix misleading/wrong info in build doc.

arvidbrodin:
'./configure --with-realtime=uspace' does look for (and finds)
any installed rtai kernels, but the build itself does NOT build
support for these. Kernel modules are not built alongside the
uspace libraries; neither is uspace builds usable for checking
that code changes build on rtai systems.

I don't know about xenomai, but the (few and old) sources I've
found on building LinuxCNC with xenomai support seems to use
something like '--with-kernel=/boot/config-3.2.27-xenomai+'
rather than '--with-realtime=uspace'.

I don't speak Spanish so I could not fully fix the Spanish
version of the doc.

### DIFF
--- a/docs/src/code/building-linuxcnc.adoc
+++ b/docs/src/code/building-linuxcnc.adoc
@@ -143,14 +143,11 @@ The most commonly used arguments are:
 
 `--with-realtime=uspace`::
 
-    Build for any realtime platform, or for non-realtime.
+    Build for Preempt-RT realtime platform, or for non-realtime.
     The resulting LinuxCNC executables will run on both a Linux kernel
     with Preempt-RT patches (providing realtime machine control) and
     on a vanilla (un-patched) Linux kernel (providing G-code simulation
-    but no realtime machine control).  If development files are installed
-    for Xenomai (typically from package libxenomai-dev) or RTAI (typically
-    from a package with a name starting "rtai-modules"), support for
-    these real-time kernels will also be enabled.
+    but no realtime machine control).
 
 `--with-realtime=/usr/realtime-$VERSION`::
 
--- a/docs/src/code/building-linuxcnc_es.adoc
+++ b/docs/src/code/building-linuxcnc_es.adoc
@@ -146,10 +146,7 @@ Los argumentos más utilizados son:
     Los ejecutables LinuxCNC resultantes se ejecutarán tanto en un kernel de Linux
     con parches Preempt-RT (que proporcionan control de la máquina en tiempo real) o
     en un núcleo Linux original (sin parches) (que proporciona simulación de código G
-    pero sin control de máquina en tiempo real). Si los archivos de desarrollo están instalados
-    para Xenomai (típicamente del paquete libxenomai-dev) o RTAI (típicamente
-    desde un paquete con un nombre que comienza por "rtai-modules"), también estarán habilitado
-    soporte para estos núcleos en tiempo real.
+    pero sin control de máquina en tiempo real).
 
 `--with-realtime=/usr/realtime-$VERSION`::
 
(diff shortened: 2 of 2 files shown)


smoe (approved):

smoe:
Thank you for spotting this. @SebKuzminsky, I propose to leave the Spanish as it is which does not seem to be affected by this bug - sidenote - we hope that such fixes in translations that are not communicated back to the original will be reduced via the upcoming crowd-translations on https://hosted.weblate.org/projects/linuxcnc/.

arvidbrodin:
Thank you, Steffen. Just for clarity: [i]after my fix[/i], the Spanish translation does not appear to be affected by this bug anymore. I.e. the Spanish translation happened to need less work than the English original.

andypugh:
I don't think that the original wording is incorrect. It is entirely possible to create a user-space build of linuxCNC that uses the RTAI realtime subsystem. (it even even performs quite well) 

```
andypugh@rm-one:~/linuxcnc-dev/src$ uname -a
Linux rm-one 4.19.195-rtai-amd64 #5 SMP PREEMPT Sun Jul 11 19:13:27 BST 2021 x86_64 GNU/Linux
andypugh@rm-one:~/linuxcnc-dev/src$ latency-test
Note: Using LXRT realtime
```

Note that the message says "using LXRT realtime" rather than the usual "POSIX realtime" or "POSIX non-realtime" 
(This system showed around 3k latency in this configuration, about twice the numbers seen with kernel space RTAI, but still very good)

smoe:
Nice insights. Many thanks!

arvidbrodin:
Ah, yes, it seems reasonable that a uspace build should work on any "real time enabled" kernel. Does this usage really require the rtai development files though? I'm assuming LinuxCNC uspace gets its realtime capability by changing its scheduling priority? (Guess I will have to try it!)

The text fooled me during development: I thought I had test-built for both uspace and rtai by simply doing a uspace build, which wasn't the case.

arvidbrodin:
Well, my assumption was wrong:

```
$ sudo apt remove rtai-modules-4.19.195
$ sudo apt remove linux-headers-4.19.195-rtai-amd64
$ uname -r
4.19.0-18-rt-amd64
$ cd ~/src/linuxcnc-dev/src/
$ make clean
$ ./configure --with-realtime=uspace
...
checking for rtai-config... none
...
$ make -j2
$ sudo make setuid

(reboot to rtai kernel)

$ uname -r
4.19.195-rtai-amd64
$ cd ~/src/linuxcnc-dev/
$ . scripts/rip-environment
$ latency-test
insmod: ERROR: could not load module /usr/realtime-4.19.195-rtai-amd64/modules/rtai_hal.ko: No such file or directory
$ sudo apt install rtai-modules-4.19.195
$ latency-test
```
So a "pure" uspace build of linuxcnc won't run in realtime on an rtai kernel at all.

It would be nice to come up with some wording for the manual that expresses that a uspace-on-rtai build is not the same as a kernel module-on-rtai build. Any ideas?

smoe:
Hm. I am now so confused that I do not even know if I should be confused. Concerning larger changes to the documentation I would prefer to first see the po4a changes accepted/functional.

arvidbrodin:
Yes, it is confusing that a "pure" uspace linuxcnc detects that it's running on an rtai kernel, loads some rtai kernel module and then fails to use any kind of realtime system at all.

But doing a uspace build on a non-rtai system and then trying to run that build on an rtai kernel is hardly a common use case. :)

andypugh:
> So a "pure" uspace build of linuxcnc won't run in realtime on an rtai kernel at all.

A uspace build built against the preempt-rt kernel won't run on an rtai kernel, no. 
Especially as I think that you removed all of /usr/realtime, which I believe contains some files that any rtai build (kernel mode or userspace) needs. For example the rtai-config script (which returns information that the system needs. Run "/usr/realtime-4.19.195-rtai-amd64/bin/rtai-config -h" to see what info this script can return. 

> It would be nice to come up with some wording for the manual that expresses that a uspace-on-rtai build is not the same as a kernelspace-on-rtai build. Any ideas?

Well. I am fuzzy on some of it myself. Not that there are more than a couple of folk in the world less fuzzy than me :-)

andypugh:
I think that the docs could perhaps be clearer, but the existing wording is more accurate than the proposal.

PR #1587, opened 2022-02-10
The second latexit.sh differs from the first

smoe:
I just went through some of the still open PRs and thought that since I am somewhat co-responsible for the delay for https://github.com/LinuxCNC/linuxcnc/pull/1284 I should send this fix+ping around.

The problem was that the .sh file was generated twice if base is set to 0, which likely is not too often, but anyway. The two are slightly different, so I got this fixed by moving the second .sh generation back into the else clause.

I had a look at the functions implementing the parsing of the arguments and feel a bit like investing into that. Another day :)

Best,
Steffen


smoe:
@andypugh - Not good?

hansu:
Maybe this was done automatically by Github when Andy deleted the branch 'LinuxCNC:jepler/latency-test'?
However, your fix should have gone to master as the branch your PR was targeting to was already merged.

smoe:
Not that I would understand it but the master branch is fine. Many thanks to you both!

andypugh:
I didn't deliberately close this. I am unsure what Github has done. 
Was the pull merged or not?

hansu:
>Was the pull merged or not?

I don't think so. And I wonder why @smoe deleted his branch...

smoe:
I checked the latest master locally and liked what I saw. latexit.sh is created only once. Or did I look at version that I had changed locally? Don't think so. Just checked again, looks good.

PR #1590, opened 2022-02-10
Disable Always on top by default (fixes #1589)

### LINKED ISSUE #1589: halshow: keep window on top #1559 does not work on GNOME
On https://github.com/LinuxCNC/linuxcnc/commit/01d26623a528b9a7d7553554d423d15027fe2446, when I pressed SetVal the User input window is created, but beneath the HAL Show window. The workaround is to either press SetVal for the second time or move the HAL Show window to uncover the User input window.

Observed on GNOME 41.3.

See also: https://github.com/LinuxCNC/linuxcnc/pull/1559#issuecomment-1028061537

dwrobel:
It looks that depending on the DE being used
enabling "Always on top" by default prevents other
windows to be also 'Always on top' - which, in fact,
makes sense.

So, let the user to conciously enable this option.

### DIFF
--- a/tcl/bin/halshow.tcl
+++ b/tcl/bin/halshow.tcl
@@ -45,7 +45,7 @@ set x [expr ($xmax - $masterwidth )  / 2 ]
 set y [expr ($ymax - $masterheight )  / 2]
 wm geometry . "${masterwidth}x${masterheight}+$x+$y"
 wm minsize . [int [expr $masterwidth*0.3]] [int [expr $masterheight*0.5]]
-wm attributes . -topmost yes
+wm attributes . -topmost no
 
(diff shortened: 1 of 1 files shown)


dwrobel:
>change this line

Changed _(I didn't notice any problem with that because as long as the `User input` window was the only top window - it didn't introduce any issue)_. Thanks!

hansu:
> > change this line
> 
> Changed _(I didn't notice any problem with that because as long as the `User input` window was the only top window - it didn't introduce any issue)_. Thanks!

I mean it is not necessary to have it on top if the main window is not on top. But you are right, it makes no problems.

phillc54:
Thanks, it is a pity all desktops didn't behave the same ... :(

PR #1591, opened 2022-02-10
offsetpage-widget: translation of 'black' results in a Gtk-Warning

hansu:
`<col id="13">black</col>` must not be translated. This results in a Gtk-Warning if translated. 
Here for locale DE: `Gtk-WARNING **: Don't know color 'Schwarz'`

### DIFF
--- a/lib/python/gladevcp/offsetpage.glade
+++ b/lib/python/gladevcp/offsetpage.glade
@@ -50,7 +50,7 @@
         <col id="10">True</col>
         <col id="11">False</col>
         <col id="12" translatable="yes"></col>
-        <col id="13" translatable="yes">black</col>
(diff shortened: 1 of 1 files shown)


smoe:
Should it better be `translatable="no"` to indicate that this was not just forgotten?

hansu:
I didn't check if that works, but this is the way glade designer creates non translatable strings.

hansu:
> Should it better be `translatable="no"` to indicate that this was not just forgotten?

Yeah that works also. I changed it.

And for "G54", "G92", ... the "translatable" can also be removed, what do you think?

hansu:
@c-morley ?

c-morley:
I agrree G5x should not be translated.

PR #1594, opened 2022-02-11
Reacting to cppcheck on omitted va_start

smoe:
The va_ routines are available with C99 so the #define better disappears to avoid confusions, I felt.

All the va-logic is required only for the first function, so I changed the structure accordingly.

I should not have added those "const" when I saw them. Shall I prepare a second PR for those?

```
motion/motion.c:221:5: error: va_list 'apc' used before va_start() was called. [va_list_usedBeforeStarted]
    va_copy(apc, ap);
    ^
motion/motion.c:222:67: error: va_list 'apc' used before va_start() was called. [va_list_usedBeforeStarted]
    if(level == RTAPI_MSG_ERR) emcmotErrorPutfv(emcmotError, fmt, apc);
                                                                  ^
motion/motion.c:224:5: error: va_list 'apc' used before va_start() was called. [va_list_usedBeforeStarted]
    va_end(apc);
    ^
```

### DIFF
--- a/src/emc/motion/motion.c
+++ b/src/emc/motion/motion.c
@@ -148,8 +148,8 @@ static int init_comm_buffers(void);
 static int init_threads(void);
 
 /* functions called by init_threads() */
-static int setTrajCycleTime(double secs);
-static int setServoCycleTime(double secs);
+static int setTrajCycleTime(const double secs);
+static int setServoCycleTime(const double secs);
 
 static int module_intfc(void);
 static int tp_init(void);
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
Perhaps lift the const and cast changes into its own commits?  I assume they get rid of some compiler warning.

src/emc/motion/motion.c
```diff
@@ -199,21 +199,24 @@ void reportError(const char *fmt, ...)
...
-    if(level == RTAPI_MSG_ERR) emcmotErrorPutfv(emcmotError, fmt, apc);
-    if(old_handler) old_handler(level, fmt, ap);
-    va_end(apc);
+   if(level == RTAPI_MSG_ERR) {
+        va_list apc;
+        //va_start(ap,fmt) - prepared by calling function
+        va_copy(apc, ap);
+	emcmotErrorPutfv(emcmotError, fmt, apc);
+        va_end(apc);
+        va_end(ap);
```
petterreinholdtsen, line 211: How will calling va_end(ap) affect the call to old_handler()?  As far as I can tell from rtapi/rtai_rtapi.c, old_handler() expect an active va_list variable, and closing calling va_end(ap) will break all calls to rtapi_msg_handler().

smoe:
Should I just close this?

PR #1603, opened 2022-02-13
halshow: remember window geometry and watchlist

hansu:
**Saves the window geometry after resizing, changing the tree view width, moving the window, show/hide tree and exit.**

If started from a GUI then the `CONFIG_DIR` can be read ans the settings file is written to this path.
If started from the bash, this environmental variable cannot is not set, so the file is placed in the home directory.
Do you agree with that behaviour or should it be always written to the home directory? (Axis is doing in the same way)
But I thought it would be better to have a own settings file for each configuration.
Maybe there is a possibility to obtain the config dir  even when started from the bash?

### DIFF
--- a/tcl/bin/halshow.tcl
+++ b/tcl/bin/halshow.tcl
@@ -960,9 +960,11 @@ proc watchReset {del} {
                 set ::watchlist [lreplace $::watchlist $place $place]
                 set watchlist_copy $::watchlist
                 set ::watchlist ""
+                set scrollbar_pos [lindex [$::cisp yview] 0]
                 foreach var $watchlist_copy {
                     watchHAL $var
                 }
+                $::cisp yview moveto [expr $scrollbar_pos * (1 + 1/double([llength $watchlist_copy]))]
                 setStatusbar "'$del' [msgcat::mc "removed from list"]"
             } else {            
                 watchHAL zzz
@@ -973,8 +975,10 @@ proc watchReset {del} {
 
 proc reloadWatch {} {
     set watchlist_copy $::watchlist
+    set scrollbar_pos [lindex [$::cisp yview] 0]
     watchReset all
     foreach item $watchlist_copy { watchHAL $item }
+    $::cisp yview moveto $scrollbar_pos
 }
 
 # proc switches the insert and removal of upper right text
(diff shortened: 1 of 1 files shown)


hansu:
Another question: Now the settings are saved action-based (after resize, move, hide/show and so on) and additionally on closing and on FocusOut. So this is a bit done twice. In a normal situation the two latter should be sufficient. Only if something crashes, the WindowDelete or FocusOut event could be missed. 
So how do you guys think we should handle the saving?

hansu:
Settings tab added.
@phillc54 now you can enable the always on top option individually :wink: 


grafik

phillc54:
This look really nice, the only issue I see is that if --fformat  or --iformat are specified on the command line then the format cannot be changed from the GUI.

hansu:
> This look really nice, the only issue I see is that if --fformat or --iformat are specified on the command line then the format cannot be changed from the GUI.

Yeah the command line overrides the settings of the GUI. I think that makes most sense. But I am open for other ideas.

phillc54:
> Yeah the command line overrides the settings of the GUI. I think that makes most sense.

Yes, I agree. Something I tried was to disabled the entries with the following code:
`    if {[string first "--fformat" $::argv] != -1} {\n`
`        $f1.::ffmts.label configure -text [msgcat::mc "Float (disabled by \"--fformat\" in command)"]`
`        $f1.::ffmts.entry configure -state disabled`
`    }`
`    if {[string first "--iformat" $::argv] != -1} {`
`        $f1.::ifmts.label configure -text [msgcat::mc "Integer (disabled by \"--iformat\" in command)"]`
`        $f1.::ifmts.entry configure -state disabled`
`    }`

after line 574:
`    addTextSetting $f1 ::ifmts "    [msgcat::mc "Integer"]"`

hansu:
Yead good idea. I can add this.


But maybe like this:
```diff
- if {[string first "--fformat" $::argv] != -1} {\n
+ if {[info exists ::ffmt]} {
```

hansu:
Do you agree with this solution?

phillc54:
Yes, I think it is nice with the entries disabled, it look good.

PR #1604, opened 2022-02-13
Introducing destructor to os_intf/inetnull.cc

smoe:
I am a bit uncertain about this one. Please discuss if this kind of hardening is indeed beneficial. And some good soul should bring this to the attention of the cppcheck devs.

Looked into this because of this error message, which is a false
positive that should be reported to cppcheck:

os_intf/inetnull.cc:45:5: error: Resource leak: fp [resourceLeak]
    return inet_file;

### DIFF
--- a/src/libnml/os_intf/inetnull.cc
+++ b/src/libnml/os_intf/inetnull.cc
@@ -58,6 +58,7 @@ int inet_file_close(INET_FILE * inet_file)
     if (NULL != inet_file) {
 	if (inet_file->fp) {
 	    fclose(inet_file->fp);
+	    inet_file->fp = NULL;
 	}
 	delete inet_file;
     }
@@ -95,4 +96,5 @@ int inet_file_rewind(INET_FILE * ifp)
 
 ~INET_FILE() {
     if (fp) fclose(fp);
+    fp = NULL;
 }
--- a/src/libnml/os_intf/inetnull.cc
+++ b/src/libnml/os_intf/inetnull.cc
@@ -20,6 +20,7 @@
 class INET_FILE {
   public:
     FILE * fp;
+    ~INET_FILE();
 };
 
 int inet_file_init(const char *agent_name, char *agent_version, int debug)
@@ -55,7 +56,9 @@ char *inet_file_gets(char *str, int maxlen, INET_FILE * inet_file)
 int inet_file_close(INET_FILE * inet_file)
 {
     if (NULL != inet_file) {
-	fclose(inet_file->fp);
+	if (inet_file->fp) {
+	    fclose(inet_file->fp);
+	}
 	delete inet_file;
     }
     return 0;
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
I did not test the code, but it seem to me like a clear file descriptor leak.

src/libnml/os_intf/inetnull.cc
```diff
@@ -55,7 +56,9 @@ char *inet_file_gets(char *str, int maxlen, INET_FILE * inet_file)
 int inet_file_close(INET_FILE * inet_file)
 {
     if (NULL != inet_file) {
-	fclose(inet_file->fp);
+	if (inet_file->fp) {
+	    fclose(inet_file->fp);
```
petterreinholdtsen, line 60: I believe this code need to do inet_file->fp = NULL; too.

smoe: A few lines down, just outside that patch, in line 62, the whole inet_file structure gets deleted.

petterreinholdtsen:
[Steffen Möller]
> A few lines down, just outside that patch, in line 62, the whole
> inet_file structure gets deleted.

I know, but it is irrelevant for my comment, which is based on good
defensive programming practice of _always_ zeroing out pointers when
they are no longer valid.

-- 
Happy hacking
Petter Reinholdtsen

src/libnml/os_intf/inetnull.cc
smoe: Done, also in the destructor (which felt a bit weird, admittedly), and rebased to master.

src/libnml/os_intf/inetnull.cc
```diff
@@ -66,7 +70,11 @@ int inet_file_eof(INET_FILE * inet_file)
     if (NULL == inet_file) {
 	return 1;
     }
-    return feof(inet_file->fp);
+    if (inet_file->fp) {
+        return feof(inet_file->fp);
+    } else {
+	return 1; // or raise an error
```
petterreinholdtsen, line 76: I checked the return value of feof() if given a bogus file pointer, and it should return -1 and set errno to EBADF.  Returning the same value as if inet_file is NULL seem to me to be the best way to handle bogus input, ie return 1 (and not raise an error).  A bogus file pointer can definitely be seen as at its end.  There is nothing more to read from it.

In short, I believe this patch should be ready to merge into master.

petterreinholdtsen (approved):

smoe:
Force-pushed a rebase to master, no changes.

smoe:
@SebKuzminsky This is ancient libnml code for which we do not have anyone resposible, I presume. Could you please have a look?

At some point I would very much like to discuss if LinuxCNC should separate the libnml code base from the rest of the source tree, but, well, that is for the future.

PR #1606, opened 2022-02-13
fix: translation of hal_mdihistory broke other translations

### LINKED ISSUE #1566: Translation of hal_mdihistory breaks Gmoccapy translation
In https://github.com/LinuxCNC/linuxcnc/commit/446e7addd3be542c1da5d6ac87eeaf3111f6181e I made the `hal_mdihistory.py` translatable.
But this breaks the Gmoccapy translation.

If i remove
https://github.com/LinuxCNC/linuxcnc/blob/13eb99bec7c95904bf25be9ab5a2f28354e7f0a5/lib/python/gladevcp/hal_mdihistory.py#L77
the gmoccapy translation works but the mdi history is not translated. Is there a way to have both or is that a problem with the two different translation domains?

hansu:
Sometimes less is more... :slightly_smiling_face: 

Gmoccapy translation in now working again and hal_mdihistory too.

fixes #1566

### DIFF
--- a/lib/python/gladevcp/hal_mdihistory.py
+++ b/lib/python/gladevcp/hal_mdihistory.py
@@ -39,12 +39,7 @@ try:
 except:
     pass
 
-import locale              # for setting the language of the GUI
 import gettext             # to extract the strings to be translated
-import sys
-
-BASE = os.path.abspath(os.path.join(os.path.dirname(sys.argv[0]), ".."))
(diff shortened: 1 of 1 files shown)


hansu:
What is the problem with this PR?

@c-morley ?

c-morley:
We will find out soon if there is one :)

hansu:
Exactly :-) Thank you!

PR #1613, opened 2022-02-16
motmod refactoring: Remove joint and axis internal data from shared mem

arvidbrodin:
This patch takes a first step in reducing the scope of internal data of motmod by making joints' and axes' internal data inaccessible from "the rest of the world".

Rationale: Reducing the scope of data makes the code easier to maintain and extend. (Specifically, this is part of preparation for adding jerk limited jog planning in the future.)


arvidbrodin:
I should add that this PR has been test built on uspace and rtai (realtime-4.19.195-rtai-amd64), and passes runtests (tested on uspace).

Issue #1616, opened 2022-02-17
Status id shadows python builtin name


satiowadahc:
http://linuxcnc.org/docs/devel/html/config/python-interface.html
https://docs.python.org/3/library/functions.html#id

Can id be renamed to motion_id? 

I see axis and src/emc/usr_intf/touchy/emc_interface.py use it.

Axis uses it to guess line number.
Touchy uses it to guess line number and display status?

Having seen its used to guess line number maybe motion_line would be more suiting?

satiowadahc:
Thought this was closed... closed with #1617

PR #1617, opened 2022-02-19
Emcmodule - Renaming status.id

satiowadahc:
#1616 
Id shadows built-in command from python. 

The naming implication is ambiguous as id actually corresponds to the currently executing motion id; not the status id which would be the memory of address of status.


andypugh:
Do you want to merge it?

satiowadahc:
Should be good to go. I can't find any other references to status.id

hansu:
@satiowadahc it seems this change https://github.com/LinuxCNC/linuxcnc/pull/1664/files also has to be fulfilled for your renaming task, right?

satiowadahc:
I appeared to have missed one. Thanks.

Issue #1599, opened 2022-02-12
OpenGL error messages at startup cause error messages


smoe:
## Here are the steps I follow to reproduce the issue:

 1. install Debian bullseye
 2. run linuxcnc
 3. press "OK" (cancel is fine)

## This is what I expected to happen:

no idea, it is the first time I started this

## This is what happened instead:
$ linuxcnc
LINUXCNC - 2.9.0~pre0
Machine configuration directory is '/home/moeller/linuxcnc/configs/sim.axis-1'
Machine configuration file is 'axis_mm.ini'
Starting LinuxCNC...
Found file(lib): /usr/share/linuxcnc/hallib/core_sim.hal
Note: Using POSIX non-realtime
Found file(lib): /usr/share/linuxcnc/hallib/sim_spindle_encoder.hal
Found file(lib): /usr/share/linuxcnc/hallib/axis_manualtoolchange.hal
Found file(lib): /usr/share/linuxcnc/hallib/simulated_home.hal
PYTHON: exception during 'this' export:
TypeError: 'Boost.Python.class' object is not iterable

The above exception was the direct cause of the following exception:

Traceback (most recent call last):

  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load

  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked

  File "<frozen importlib._bootstrap>", line 666, in _load_unlocked

  File "<frozen importlib._bootstrap>", line 565, in module_from_spec

  File "<frozen importlib._bootstrap>", line 763, in create_module

SystemError: _PyEval_EvalFrameDefault returned a result with an error set

link (updating variable file): Datei oder Verzeichnis nicht gefunden
note: MAXV     max: 53.340 units/sec 3200.400 units/min
note: LJOG     max: 53.340 units/sec 3200.400 units/min
note: LJOG default: 30.480 units/sec 1828.800 units/min
note: jog_order='XYZ'
note: jog_invert=set()
PYTHON: exception during 'this' export:
TypeError: No Python class registered for C++ class Interp

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/OpenGL/latebind.py", line 43, in __call__
    return self._finalCall( *args, **named )
TypeError: 'NoneType' object is not callable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/axis", line 3979, in <module>
    commands.set_view_p()
[...]

PR #1619, opened 2022-02-20
Avoiding crash of https://github.com/LinuxCNC/linuxcnc/issues/1599

smoe:
I admit to be just slightly curious if this now builds on github - it should. I admit not to be the ultimate fan of this solution, since the test should be what python3's version is. The pragmatic side of mine has won for now, if you have a better idea - please go for it. Whatever we come up with to upload to Debian, it needs to run on unstable and bookworm. This patch was tested on unstable and if this now also builds on on the build server then I consider it just fine.

### DIFF
--- a/src/m4/ax_python.m4
+++ b/src/m4/ax_python.m4
@@ -55,7 +55,7 @@
 AC_DEFUN([AX_PYTHON],
 [AC_MSG_CHECKING(for python build information)
 AC_MSG_RESULT([])
-for python in python3.10 python3.9 python3.8 python3.7 python3.6 python3.5 python3.4 python3.3 python3.2 python3.1 python3.0 python2.7 python2.6 python2.5 python2.4 python2.3 python2.2 python2.1 python; do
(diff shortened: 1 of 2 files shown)


src/m4/ax_python.m4
```diff
@@ -55,7 +55,7 @@
 AC_DEFUN([AX_PYTHON],
 [AC_MSG_CHECKING(for python build information)
 AC_MSG_RESULT([])
-for python in python3.10 python3.9 python3.8 python3.7 python3.6 python3.5 python3.4 python3.3 python3.2 python3.1 python3.0 python2.7 python2.6 python2.5 python2.4 python2.3 python2.2 python2.1 python; do
```
dwrobel, line 58: It would be appreciated not to break the existing systems in which the python3.10 is the default one _(e.g. Fedora >=35)_.

dwrobel:
>This patch was tested on unstable and if this now also builds on on the build server then I consider it just fine.

FYI, I'm building linuxcnc for Fedora 34 - which uses python3.9 and for Fedora >=35 which uses python3.10 without any issues. So, I'm not sure why this needs to be patched.

smoe:
Debian unstable has both Python 3.10 and python 3.9 installed. Everything newly built links against 3.10. But the version /usr/bin/python3 is 3.9. And this then provokes what https://github.com/LinuxCNC/linuxcnc/issues/1599 shows. I agree that this patch is not acceptable. Thank you!

Issue #1618, opened 2022-02-20
config.guess - is there a chance to remove this file from source tree?


smoe:
Dear @jepler,

src/config.guess is changed during the build. Ideally, the regular source tree remains invariant, so everything remains reproducible and we know where a change is coming from.

I just compared src/config.guess it with /usr/share/autoconf/build-aux/config.guess and from a first clance preferred the version in /usr/share/autoconf. I failed to grasp if there is any particular problem you were adressing with a change to that config.guess file that I have missed. If not then I would prefer to copy over the config.guess in debian prior to the execution of configure - I have seen this in a few other packages, too.

Best,
Steffen

jepler:
I don't recall why I had to add this file. If you can make our builds happy without it, please do.

jepler:
aeb1c09ea7d81fac04fde2b751a396fbb56b241c adds a number of files simply copied in from the autoconf ecosystem. I'd rather that src/autogen.sh copy them in so that it correctly serves not only people building debian packages but people building run-in-place or for non-debian packaging systems.

jepler:
Some [autoconf documentation](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/autoconf.html) seems to suggest "just take a random copy of these files from the internet":

> The scripts needed by Autoconf core macros can be found in $(datadir)/autoconf/build-aux of the Autoconf installation (see [Installation Directory Variables](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/autoconf.html#Installation-Directory-Variables)). install-sh can be downloaded from https://git.savannah.gnu.org/cgit/automake.git/plain/lib/install-sh. config.sub and config.guess can be downloaded from https://git.savannah.gnu.org/cgit/config.git/tree/.

I also found indications that maybe I need to use some macros & `autoreconf --install` would copy them but it didn't for me. https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/autoconf.html#index-install_002dsh

Please make it better if you know how. We don't know of love auto---- we just use it.

dwrobel:
Please find an alternative approach to address this issue in the https://github.com/LinuxCNC/linuxcnc/pull/1625.

smoe:
[...]

PR #1620, opened 2022-02-20
Avoiding configure-modified files in source tree

smoe:
Addressing https://github.com/LinuxCNC/linuxcnc/issues/1618

I admit not to have tested this locally, but if this builds then it should just be fine.


smoe:
My changes may raise some eyebrows.  Everything is fine if just debian/rules or dpkg-buildpackage is used. However, the manual step to copy config.* elsewhere from the local disk is not yet described in the documentation.

What I should not have added in this PR is that I changed "fakeroot debian/rules binary" to a mere "dpkg-buildpackage". I am bit surprised that this worked. The benefit I was after is that lintian can be invoked on the .changes file which checks the source tree and then all the .debs afterwards.

smoe:
@SebKuzminsky , could you please have a look since I am touching the .github/CI?

dwrobel:
> My changes may raise some eyebrows. Everything is fine if just debian/rules or dpkg-buildpackage is used. However, the manual step to copy config.* elsewhere from the local disk is not yet described in the documentation.

Please review the alternative approach https://github.com/LinuxCNC/linuxcnc/pull/1625, which doesn't change anything in the manual build steps. It is also not debian/dpkg specific.

smoe:
From a first glance at it that regular autoconf/automake setup of #1625 is a better approach, indeed.

PR #1625, opened 2022-02-21
Remove autogenerated files (fixes #1618)

### LINKED ISSUE #1618: config.guess - is there a chance to remove this file from source tree?

dwrobel:
Removes the following files automake can either
generate or install:
 - src/config.guess
 - src/config.sub
 - src/install-sh

Adjusts the src/.gitignore accordingly.

Switches to use autoreconf in autogen.sh.

Adds Makefile.am stub as this is required by automake and
renames Makefile to makefile (to keep precedence over
generated Makefile).

Removes everything related to TRIVIAL_BUILD as in a standard
automake project it is also not possible to invoke any make
targets prior to generating configure/Makefile files.

Future improvements of build system could:
 - move things from makefile to Makefile.am and then
   remove makefile,
 - switch to use a different build system.


smoe (approved):

src/autogen.sh
```diff
@@ -1,10 +1,3 @@
 #!/bin/sh
 rm -rf autom4te.cache
-AUTOGEN_TARGET=${AUTOGEN_TARGET-configure:config.h.in}
-set -e
-case :$AUTOGEN_TARGET: in
-*:configure:*) autoconf; touch configure ;;
-esac
-case :$AUTOGEN_TARGET: in
-*:config.h.in:*) autoheader; touch config.h.in ;;
```
jepler, line 9: I need to check why we had both of these before, I think it has something to do work make invoking this script itself

dwrobel: >I need to check why

Please, because It wasn't obvious for me and I chose simplicity over complexity.

smoe:
What is important to me is that all autogenerated files are removed upon a "make -C src clean" (debian/rules clean this is). Please kindly have a quick check on that before this gets accepted.

dwrobel:
> What is important to me is that all autogenerated files are removed upon a "make -C src clean" (debian/rules clean this is). Please kindly have a quick check on that before this gets accepted.

`clean` target cannot delete any generated files because it would be last time that `make` worked properly. Maybe you had in mind `distclean` target?

smoe:
Yes.

dwrobel:
> Yes.

But linuxcnc didn't support `distclean` so far, so please don't expect it will start working magically out of the box.

smoe:
My last automake project is so long ago that what you have just done already feels like magic to me :) I can live/continue work on whatever you now come with.

src/autogen.sh
jepler:
```
ifeq ($(TRIVIAL_BUILD),no)
configure: configure.ac
        AUTOGEN_TARGET=configure ./autogen.sh

config.h.in: configure.ac
        AUTOGEN_TARGET=config.h.in ./autogen.sh
```

I haven't been [comfortable with make rules that generate multiple targets](https://www.gnu.org/software/automake/manual/html_node/Multiple-Outputs.html)—read the docs about it, they're a hoot of workarounds piled on workarounds, so I allowed make to invoke it in this way. It ensures that each file is generated by a distinct invocation of autogen.sh. With your change,

 * autogen.sh will be invoked twice
 * to update all the same files
 * possibly in parallel

On the other hand, if the stanza for `config.h.in` isn't present, then `make` may believe it does not need to regenerate `config.h` from `config.h.in`:
```
config.h: config.h.in config.status
        @./config.status -q --header=$@                                         
```
... depending when it stats config.h.in.

I have to admit I am not confident whether this works as intended or not, users should just re-run `autogen.sh` manually when modifying configure.ac. But I'd rip out some amount of Makefile goo, trying to make it clear that make can't really re-invoke configure and restart itself as it actually needs to do.

dwrobel:
Conflct resolved.

src/autogen.sh
jepler:
```make
config.h.in: configure.ac
	AUTOGEN_TARGET=config.h.in ./autogen.sh
#…
Makefile: config.h  
```

I've been trying to disentangle this.

Here's what I think I see:
 * There are a lot of places where `Makefile` is specifically depended. Since it was renamed to `makefile` all these explicit dependencies need to be revised. (on left and right sides)
 * There's special machinery in make where, if a file used as input by make is out of date, make will regenerate it and then reinvoke itself. This seems to allow getting rid of the special `config.h: ...` rule that I noticed at first, because make will re-invoke itself as long as there's a correct chain from `configure.ac` to `[Mm]akefile` (whatever filename is being used)
 * leaving the rule is bad, because it'll invoke the `autogen` script twice in parallel to generate the same files. Who knows what you end up with in this case

It's unfortunate that build system bugs of the type I'm describing above are often hard to reproduce and can depend on which jobs are invoked in what order, the job parallelism, etc.  The problem of renaming Makefile but not changing the rules that refer to it could easily go unnoticed, because it doesn't produce any overt behavior most of the time -- you just get rebuilds that are incomplete.

smoe:
I presume why this is not getting forward is that we now have two Makefiles - one "Makefile" and a "makefile".

dwrobel:
I'll look into it, but first need to find some spare time.

dwrobel:
>Please see my comments on February 21 and 26.

I addressed the comments related to renaming `Makefile`->`makefile`.

However, I also removed the 'special' machinery from the `makefile` related to automatically call `autogen.sh` because the `autogen.sh` is something which is normally not distributed in a tarball - so there would be nothing to invoke.

Additionally, I didn't like to duplicate this kind of logic which is normally autogenerated by `autotools` in the `Makefile`. So, the current status is that it will start working again when all of the remaining logic would be moved from the `makefile` to the `Makefile`. I would understand if that would be yet another reason to reject this PR, but on the other hand there is too much legacy to remove it in one shot to make this project fully `autotools` compatible.

src/makefile
```diff
@@ -34,22 +34,6 @@ endif
 .SUFFIXES:
 .PHONY: default userspace modules clean modclean depclean install python pythonclean cscope cscopeclean
 
-# A "trivial build" is one which should not include dependency information
-# either because it should be usable before dependency information can be
-# generated or when it is invalid (clean, docclean) or when running as root
-# when the user must guarantee in advance that everything is built
-# (setuid, install)
-ifeq ($(MAKECMDGOALS),)
```
jepler, line 42: Can you please let me know the motivation of removing TRIVIAL_BUILD? It causes "make clean" to have to do a substantial amount of work. Additionally, we have found in the past that if (e.g.,) "sudo make setuid" does things like create dependency files and their containing directories, users end up with problems removing the files later with "make clean". Users are not always aware of whether invoking "make" will need to do these steps, since a point of make is to abstract away details like these.

src/makefile
```diff
@@ -271,24 +253,6 @@ $(sort $(patsubst %.cc,%.ii,$(CXXUSERSRCS))): %.ii: %.cc
 	$(ECHO) Preprocessing $< to $@
 	$(Q)$(CXX) -dD $(CXXFLAGS) $(EXTRAFLAGS) -E $< -o $@
 
-ifeq ($(TRIVIAL_BUILD),no)
-configure: configure.ac
```
jepler, line 275: Can you please let me know why you chose to remove this block, which helps re-generate and re-run configure when configure.ac is modified? This feels like a regression, since the user now has to manually re-run autogen.sh

jepler (changes requested):

src/makefile
jepler: I see that the initial comment explains this:
> Removes everything related to TRIVIAL_BUILD as in a standard automake project it is also not possible to invoke any make targets prior to generating configure/Makefile files.
However, as the automake'd Makefile is never actually used, does it require breaking this feature?

jepler:
I've offered a different approach to this, not invoking automake, at #1848.

src/makefile
dwrobel: There are a few reasons:
- `autogen.sh` is the next file to be removed as it doesn't make sense to keep an extra wrapper around the `autoreconf`,
- if you run `autoreconf` then the generated `Makefile` will contains this logic - however, it will become usable once the existing logic from the `makefile` will be moved to the `Makefile.am` then the `makefile` could be entirely removed - at that time the `LinuxCNC` project will become more `autotool` compliant.
- this is not a crucial functionality - so it should be acceptable to document that you need to run `autoreconf` after `configure.ac` was modified - which you need to run anyway if you start compiling the project from the scratch.

rene-dev:
@jepler does #1848 close this?

jepler:
Thanks for your work on this @dwrobel, it's appreciated even though I felt there were problems and ended up offering a different solution with different trade-offs. I hope you will consider doing more work to improve our build system in the future, and for my part I'll try to be open to new ideas.

dwrobel:
Based on above, I'm closing it.

PR #1628, opened 2022-02-26
Add debian/source/format

smoe:
This should not disturb for earlier releases and should need to be added
with every update for Debian/derivative distros. Should it be problematic somewhere, the file is easily deleted.

Somehow I feel like I had proposed this before and something spoke against it. Could not find it.


havardAasen:
> Somehow I feel like I had proposed this before and something spoke against it. Could not find it.

Could it be this #1277 you were looking for?

I agree that the file should exists, but looking at the mentioned pr and earlier versions, it seems that this was a native package.

PR #1633, opened 2022-02-28
Readme incorrectly stated XYZY when it should be XYYZ. Reported as #1…

JTrantow:
Minor update to Readme files. Gantry setup is XYYZ.

### DIFF
--- a/configs/sim/axis/gantry/README
+++ b/configs/sim/axis/gantry/README
@@ -1,10 +1,11 @@
 Axis gui sim configs for gantry applications
-using 2 joints for the Y coordinate
+using 2 joints for the Y coordinate.
 
-Use trivkins coordinates=XYZY kinstype=BOTH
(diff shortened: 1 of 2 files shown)


andypugh:
I was going to suggest that this be aimed at 2.8, but looking at 2.8 I see that gantry.ini is XYZY and gantry_mm.ini is XYYZ. 
It seems that they do not share a HAL file, so this is probably OK.

andypugh:
The checks failed "runtests" (mountaindew test) but I think that it is inconceivable that this PR is to blame.

JTrantow:
I was looking at gantry_mm. I will double check the other 2.9 gantry
configs and make sure they are all XYYZ.
If there is going to be a 2.8.3 before 2.9 goes mainstream I could direct
this for 2.8 too.

I concur with the runtest failure.

PR #1637, opened 2022-02-28
Branch issue1634

JTrantow:
Updated gantry configurations for /by_interface/parport/gantry

I ran these configs on a router with 2.8.2 LinuxCNC install.


andypugh:
I think that looks like a _lot_ more than you intended. 
I would suggest checking out 2.8 then cherry-picking your specific commits.

JTrantow:
Yes, that was a mistake. The only changes I wanted were in
configs/by_interface/par/gantry

I think I branched from the master and did the pull request to 2.8 or
2.8.pocket-fix. I will resubmit a PR later tonight or tomorrow.

Do pull requests for 2.8 automatically get into the master(2.9)? I think a
good gantry example would be helpful asap and going forward.

andypugh:
Yes. 2.8is periodically merged-up to master. It's not automatic, but it does get done.

Issue #366, opened 2017-12-02
Suggestions about Linuxcnc default editor


Solitarily:
I found that Linuxcnc's editor was gedit by default, while linuxcnc.iso did not include gedit when it was released, only Mousepad. Although you can solve the problem by installing gedit or reassigning it as a Mousepad, this can be annoying to novices. So I think it might be better to specify the editor as Mousepad by default.

PR #1640, opened 2022-03-02
Improved parallel port gantry config.

JTrantow:
Updated the gantry example with what I gather from the forum is best gantry practice.
1) Reduce from 4(XYZA) to 3(XYZY) coordinate.
2) Use negative HOME_SEQUENCE values for synchronous homing of the dual Y joints.
3) Rename signals such as adir, astep to y2dir, y2step.
4) Provide gantry_joint.ini that allows homing in joint space to verify hardware before running gantry.ini.


andypugh:
I am generally in favour apart from two minor points. 
1) EDITOR = gedit I don't like. The LiveCD installs geany which comes with far fewer dependencies. Installing gedit on the LiveCD install brings in several GB of Gnome Desktop (or did when I tried it) 
2) There is no real need to reduce the dir-setup time, and it is possible that some hardware needs more.

JTrantow:
I will roll back those changes and resubmit.
1. EDITOR = gedit was commented out #366 and I didn't look back at the log
to see why it was commented out.  I use nano but will change to geany.
2. dir-setup may have come from stepconf G540 selection. I will revert to
the old dir-setup.

PR #1214, opened 2021-07-22
fix start/stop/pause behaviour with xhc-whb04b-6

sundtek:
this fixes the unreliable {start/stop/pause} behaviour of the xhc-whb04b-6 pendant

https://github.com/LinuxCNC/linuxcnc/issues/1187
https://forum.linuxcnc.org/24-hal-components/42740-xhc-whb04b-4-6-issues

PR #1644, opened 2022-03-03
Alkabal 2.8 xhc whb04b 6

alkabal:
> Hi

I can propose this update
at first i was think that it need only to update things related tyo this topic : https://github.com/LinuxCNC/linuxcnc/pull/1214

Best regards


andypugh:
I don't have an XHC to check your work with. Are you 100% sure that this is good?

alkabal:
Hello

My unit work better with this patch (from some time now).

Best regards

PR #1646, opened 2022-03-04
docs: HAL Component List - generate unlisted components automatically

### LINKED ISSUE #996: Restructure documentation of HAL components

hansu:
As discussed in #996 it would be great to have a list of all HAL components which is always up-to-date.

I decided to use the current list as it already covers a lot of components.
A script checks this list against the existing man pages and generates tables for those that are missing and for those that are too much. It writes those into two tables each that are included in the components.adoc.
Moreover it generates the links for the man pages for the missing entries AND extracts the descriptions of the man pages and writes them into the table.
So all what you see in the tables **Not categorized (auto generated)** and **Obsolete (auto generated)** in 
https://hansu.github.io/linuxcnc-doc/devel/html/hal/components.html is generated.
(The current page as reference: http://linuxcnc.org/docs/devel/html/hal/components.html)
This can now be manually sorted into suitable categories. And it ensures that components that are added are always in this list.

The script detects missing man pages for these components.
- bldc_hall3
- gs2
- hal_ppmc
- pluto_servo
- pluto_step

I guess `bldc_hall3` is now `bldc` and `gs` is renamed to `gs2_vfd`. And the other three - are they still available and just don't have a man page?

Another question: should we keep the list of
3. HAL API calls 
and
4. RTAPI calls
here? I think it they are not of great benefit here.

@andypugh I would appreciate your opinion here ;-)

Closes #996

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,8 @@ build-stamp
 .tmp*
 # Ignore generated html files, 
 /docs/src/*/*.html
+docs/src/hal/components_gen1.adoc
+docs/src/hal/components_gen9.adoc
 # docs/html/.gitignore is for the html directory
 debian/*.debhelper.log
 rtlib/Module.symvers
--- a/.gitignore
+++ b/.gitignore
@@ -13,8 +13,7 @@ build-stamp
 .tmp*
 # Ignore generated html files, 
 /docs/src/*/*.html
-docs/src/hal/components_gen1.adoc
-docs/src/hal/components_gen9.adoc
+docs/src/hal/components_gen.adoc
 # docs/html/.gitignore is for the html directory
 debian/*.debhelper.log
 rtlib/Module.symvers
--- a/docs/po4a.cfg
+++ b/docs/po4a.cfg
@@ -111,6 +111,7 @@
 [type: AsciiDoc_def] src/hal/canonical-devices.adoc $lang:src/$lang/hal/canonical-devices.adoc
 [type: AsciiDoc_def] src/hal/comp.adoc $lang:src/$lang/hal/comp.adoc
 [type: AsciiDoc_def] src/hal/components.adoc $lang:src/$lang/hal/components.adoc
+[type: AsciiDoc_def] src/hal/components_gen.adoc $lang:src/$lang/hal/components_gen.adoc
 [type: AsciiDoc_def] src/hal/general-ref.adoc $lang:src/$lang/hal/general-ref.adoc
 [type: AsciiDoc_def] src/hal/hal-examples.adoc $lang:src/$lang/hal/hal-examples.adoc
 [type: AsciiDoc_def] src/hal/halmodule.adoc $lang:src/$lang/hal/halmodule.adoc
--- a/docs/src/Submakefile
+++ b/docs/src/Submakefile
@@ -336,7 +336,6 @@ DOTFILES=$(shell find . -name "*.dot") $(shell find ../docs/src/ -name "*.dot")
 .PHONY: svgs_made_from_dots
 svgs_made_from_dots: $(DOTFILES:.dot=.svg)
 
-docs: gen_complist
 ifeq ($(BUILD_DOCS_PDF),yes)
 docs: pdfdocs
 install-doc: install-doc-pdf
--- a/docs/src/Submakefile
+++ b/docs/src/Submakefile
@@ -336,6 +336,7 @@ DOTFILES=$(shell find . -name "*.dot") $(shell find ../docs/src/ -name "*.dot")
 .PHONY: svgs_made_from_dots
 svgs_made_from_dots: $(DOTFILES:.dot=.svg)
 
+docs: gen_complist
 ifeq ($(BUILD_DOCS_PDF),yes)
 docs: pdfdocs
 install-doc: install-doc-pdf
@@ -353,17 +354,17 @@ htmldocs: svgs_made_from_dots .htmldoc-stamp checkref_en
 .htmldoc-stamp: copy_asciidoc_files gen_complist $(HTML_TARGETS) .images-stamp .include-stamp
 	touch $@
 
-.PHONY: copy_asciidoc_files
+.PHONY: copy_asciidoc_files gen_complist gen_complist_links
 copy_asciidoc_files:
 	cp -f /etc/asciidoc/stylesheets/*.css $(DOC_DIR)/html
 	cp -f /etc/asciidoc/javascripts/*.js $(DOC_DIR)/html
 
 gen_complist:
-	python3 $(DOC_SRCDIR)/gen_complist.py
+	python3 $(DOC_SRCDIR)/gen_complist.py $(DOC_SRCDIR)/hal/components.adoc
 	touch $(DOC_SRCDIR)/hal/components.adoc
 
-.PHONY gen_complist_links:
-	python3 $(DOC_SRCDIR)/gen_complist.py links
+gen_complist_links:
+	python3 $(DOC_SRCDIR)/gen_complist.py $(DOC_SRCDIR)/hal/components.adoc links
 
 checkref: checkref_en checkref_de checkref_es checkref_fr checkref_hu checkref_nb checkref_vi checkref_zh_CN
 
--- a/docs/src/Submakefile
+++ b/docs/src/Submakefile
@@ -353,17 +353,12 @@ htmldocs: svgs_made_from_dots .htmldoc-stamp checkref_en
 .htmldoc-stamp: copy_asciidoc_files gen_complist $(HTML_TARGETS) .images-stamp .include-stamp
 	touch $@
 
-.PHONY: copy_asciidoc_files gen_complist gen_complist_links
-copy_asciidoc_files:
-	cp -f /etc/asciidoc/stylesheets/*.css $(DOC_DIR)/html
-	cp -f /etc/asciidoc/javascripts/*.js $(DOC_DIR)/html
+copy_asciidoc_files: $(wildcard /etc/asciidoc/stylesheets/*.css) $(wildcard /etc/asciidoc/javascripts/*.js)
+	cp -f $^ $(DOC_DIR)/html
 
-gen_complist:
+gen_complist: $(DOC_SRCDIR)/gen_complist.py $(DOC_SRCDIR)/hal/components.adoc $(MAN_HTML_TARGETS)
+	mkdir -p $(DOC_DIR)/html/hal
 	python3 $(DOC_SRCDIR)/gen_complist.py $(DOC_SRCDIR)/hal/components.adoc
-	touch $(DOC_SRCDIR)/hal/components.adoc
-
-gen_complist_links:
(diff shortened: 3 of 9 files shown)


andypugh:
> On 4 Mar 2022, at 16:51, Hans ***@***.***> wrote:
> 
> hal_ppmc
> pluto_servo
> pluto_step
> I guess bldc_hall3 is now bldc and gs is renamed to gs2_vfd. And the other three - are they still available and just don't have a man page?
> 

I think that they are listed in the drivers section rather than HAL components. 

Bldc_hall remains in place for backwards compatibility in case anyone has it in their HAL still. Though now we have update_ini it could be auto-replaced.

hansu:
>I think that they are listed in the drivers section rather than HAL components.

Yes but I want to make sure that they are still 'alive'. Maybe we should give them man pages.
And `bldc_hall3` is already gone.

But I wanted to know your opinion more about the generation in general what I introduced here.

andypugh:
Scanning what is there seems like  good idea, though I see some artefacts. 

"comp" was renamed as "halcompile" largely because if Linux couldn't find "comp" it gave a really misleading error message suggesting that the fix was to install "mh"
https://github.com/LinuxCNC/linuxcnc/commit/a0613f0834e5f14f00b2673683b961ef6fd9c08a

It seems that the manual page was not renamed at the same time. Which seems like an omission. 
hal_ppmc is the driver for Jon Elson's Parallel Port Motion Controller which is still very much a live product. 

Pluto probably still works, but it was flaky at best when first introduced arouinf 2007. Apparently they are still available but I would never recommend it now. https://www.knjn.com/FPGA-Parallel.html

andypugh:
There is a HAL component called "comp" which has a manpage, but comp.adoc is the documentation for halcompile, I think.

hansu:
> There is a HAL component called "comp" which has a manpage, but comp.adoc is the documentation for halcompile, I think.

That's correct. But it doesn't conflict with the generated table/links as I only take the man pages into account.
Nevertheless it would be good to rename the comp.adoc to prevent any confusion.

andypugh:
I have a feeling that I am misunderstanding the main question. Can you re-state it?

petterreinholdtsen:
Now that po4a is used with the documentation, perhaps this patch should be extended to list the new adoc files in po4a.cfg?

hansu:
The problem is that these files are only temporary auto generated ones which should not be comitted.

hansu:
Then we should add them to the po4a.cfg.

jepler:
sooo halcompile turns out to depend on USER_MODULE_DIR
```
src/hal/utils/halcompile.g:BINDIR = os.getenv('USER_MODULE_DIR', None) or os.path.join(BASE, "bin")
```
with this change, the userspace module used by this test gets installed to a location not on PATH.  Way down in the raw error logs:
```
2022-07-16T16:46:13.9045685Z         both.hal:1: execv(userspace_count_names): No such file or directory
2022-07-16T16:46:13.9046391Z         both.hal:1: waitpid failed userspace_count_names userspace_count_names
2022-07-16T16:46:13.9046905Z         both.hal:1: userspace_count_names exited without becoming ready
2022-07-16T16:46:13.9047596Z         Note: Using POSIX non-realtime
2022-07-16T16:46:13.9047988Z         + for HAL in *.hal
```

sorry this is such a tangle!

jepler:
Looks like this needs to be re-worked, there are multiple conflicts.

I'm surprised that components_gen#.adoc are both added as empty files in this PR, but also listed in .gitignore. Is this intentional or does it represent a mistake?

hansu:
>Looks like this needs to be re-worked, there are multiple conflicts.

Yes, it's on the top 3 of my LinuxCNC to-do list, but it's summer and I don't have so much time for that :innocent: .
`components_gen#.adoc` is only the generated part and should not be committed. It doesn't have to be an existing file (which is empty), but I didn't managed it without that file. It always complained about that missing file in the "depending" stage. I will reach out for someone with better makefile skills if I continue on that.

hansu:
Following problem: When I build this branch (also available here: https://github.com/LinuxCNC/linuxcnc/tree/hansu/doc-component-list) locally, all is building as expected. But on the CI I get following in the logs:
```
Formatting userkins.9 as HTML
Formatting watchdog.9 as HTML
Formatting wcomp.9 as HTML
Formatting weighted_sum.9 as HTML
Formatting xhc_hb04_util.9 as HTML
Formatting xor2.9 as HTML
gen_complist: Broken link: ../man/man1/axis.1.html
python3 ../docs/src/gen_complist.py ../docs/src/hal/components.adoc
gen_complist: Component  hal_ppmc  without link
gen_complist: Component  pluto_servo  without link
gen_complist: Component  pluto_step  without link
gen_complist_links: Added 268 link(s) to ../docs/src/hal/components_gen.adoc
gen_complist: Added 268 uncategorized and 217 obsolete entries to hal component list
gen_complist: Broken link: ../man/man1/axis-remote.1.html
gen_complist: Broken link: ../man/man1/gmoccapy.1.html
gen_complist: Broken link: ../man/man1/gscreen.1.html
gen_complist: Broken link: ../man/man1/halui.1.html
gen_complist: Broken link: ../man/man1/mdro.1.html
gen_complist: Broken link: ../man/man1/ngcgui.1.html
```
>`gen_complist: Broken link: ../man/man1/axis-remote.1.html`

That means gen_complist cannot find e.g. axis-remote.1.htm. But according to the above log the HTML manual page files were just created before. Is this maybe a parallelism problem? Or does someone has another hint?

jepler:
I initially wrote this advice because I assumed you were talking about a situation where make exits with an error, but that appears to not be the case. Leaving it as it might be helpful for the future:

When diagnosing a make parallelism error, I find it helpful to not the specific failed target listed by make:
```
make: *** [example.mk:1: sample-target] Error 99
```

Now, from a clean build, invoke make listing only that target. If a dependency of `sample-target` is not explicitly listed, then it won't be built, so the failure should become reproducible.

Anyway, what may be the actual problem: 
```python
                    if os.path.isfile(os.path.join('../docs/html/hal',link)):
                        complist_doc.add(comp_man)
                    else:
                        print('gen_complist: Broken link:', link, file=sys.stderr)
                        miss_in_man.add(comp_man.split(".")[0])
```
I found that if I start with `make gen_complist` alone, the directory `../docs/html/hal` has not yet been created, so trying to resolve `../docs/hal/html/../man/man1/axis-remote.1.html` does not work.

```
From 0333104d2bae892a9d62578aa90113ade7b99f37 Mon Sep 17 00:00:00 2001
From: Jeff Epler <jepler@gmail.com>
Date: Wed, 7 Sep 2022 09:22:55 -0500
Subject: [PATCH] Try to fix dependencies and non-error build errors

---
 docs/src/Submakefile | 13 ++++++-------
 1 file changed, 6 insertions(+), 7 deletions(-)

diff --git a/docs/src/Submakefile b/docs/src/Submakefile
index 6fd05e9079..ac23fbcbc4 100644
--- a/docs/src/Submakefile
+++ b/docs/src/Submakefile
@@ -352,16 +352,15 @@ htmldocs: svgs_made_from_dots .htmldoc-stamp checkref_en
 .htmldoc-stamp: copy_asciidoc_files $(MAN_HTML_TARGETS) gen_complist $(HTML_TARGETS) .images-stamp .include-stamp
 	touch $@
 
-.PHONY: copy_asciidoc_files gen_complist gen_complist_links
-copy_asciidoc_files:
-	cp -f /etc/asciidoc/stylesheets/*.css $(DOC_DIR)/html
-	cp -f /etc/asciidoc/javascripts/*.js $(DOC_DIR)/html
+copy_asciidoc_files: $(wildcard /etc/asciidoc/stylesheets/*.css) $(wildcard /etc/asciidoc/javascripts/*.js)
+	cp -f $^ $(DOC_DIR)/html
 
-gen_complist:
+gen_complist: $(DOC_SRCDIR)/gen_complist.py $(DOC_SRCDIR)/hal/components.adoc $(MAN_HTML_TARGETS)
+	mkdir -p $(DOC_DIR)/html/hal
 	python3 $(DOC_SRCDIR)/gen_complist.py $(DOC_SRCDIR)/hal/components.adoc
-	touch $(DOC_SRCDIR)/hal/components.adoc
 
-gen_complist_links:
+gen_complist_links: $(DOC_SRCDIR)/gen_complist.py $(DOC_SRCDIR)/hal/components.adoc $(MAN_HTML_TARGETS)
+	mkdir -p $(DOC_DIR)/html/hal
 	python3 $(DOC_SRCDIR)/gen_complist.py $(DOC_SRCDIR)/hal/components.adoc links
 
 checkref: checkref_en checkref_de checkref_es checkref_fr checkref_hu checkref_nb checkref_vi checkref_zh_CN
-- 
2.30.2

```
I've attempted to add proper dependency analysis as well as explicit creation of the required directory. After this, the sequence
```
jepler@bert:~/src/linuxcnc/src$ make clean; git clean -dxf ../docs; make gen_complist -j9
```
ends with the output
```
Formatting xor2.9 as HTML
mkdir -p ../docs/html/hal
python3 ../docs/src/gen_complist.py ../docs/src/hal/components.adoc
gen_complist: Component  hal_ppmc  without link
gen_complist: Component  pluto_servo  without link
gen_complist: Component  pluto_step  without link
generate_links ../docs/src/hal/components_gen.adoc
gen_complist_links: Added 53 link(s) to ../docs/src/hal/components_gen.adoc
gen_complist: Added 53 uncategorized and 3 obsolete entries to hal component list
```

hansu:
Great, thanks a lot for digging into my issue :slightly_smiling_face: 

>I found that if I start with make gen_complist alone, the directory ../docs/html/hal has not yet been created, so trying to resolve ../docs/hal/html/../man/man1/axis-remote.1.html does not work.

That makes totally sense! And the reason why it locally worked was because I only ran `make clean`, not `git clean`, so the folder was still existing.

So with this proper dependency analysis, this commit can be dropped as well: https://github.com/LinuxCNC/linuxcnc/pull/1646/commits/68c2610188053c8e58935bcb4c8d588cbdbb1be1

I only wonder why you removed the phony declaration of the target "copy_asciidoc_files". I thought it is good practise to declare targets in this way which don't correspond to files.

PR #1647, opened 2022-03-04
Gmoccapy: add pins for toolchanger confirm and abort

hansu:
Feature request according to https://forum.linuxcnc.org/gmoccapy/42219-signal-tool-change-complete-with-physical-panel-button#207414

This adds the pins
* gmoccapy.toolchange-confirm
* gmoccapy.toolchange-abort

to confirm and cancel the tool change and
* gmoccapy.warning-confirm

to confirm the warning dialogs.

### DIFF
--- a/docs/src/gui/gmoccapy.adoc
+++ b/docs/src/gui/gmoccapy.adoc
@@ -807,10 +807,11 @@ pin is exported.
 The settings page is unlocked if the pin is high.
 To use this pin, you need to activate it on the settings page.
 
-=== Error Pins
+=== Error/Warning Pins
 
- * gmoccapy.error
- * gmoccapy.delete-message
+ * gmoccapy.error _BIT OUT_
+ * gmoccapy.delete-message _BIT IN_
+ * gmoccapy.warning-confirm _BIT IN_ Confirms warning dialog like click on OK
 
 gmoccapy.error is an bit out pin, to indicate an error, so a light can lit or even the machine may
 be stopped. It will be reset with the pin gmoccapy.delete-message. gmoccapy.delete-message will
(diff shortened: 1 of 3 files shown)


c-morley:
You can't actually abort a tool change without aborting linuxcnc or at least it makes no sense to.
Either a tool change abort should abort linuxcnc or the option should be removed from the dialog. IMHO

hansu:
I only reflected Gmoccapy's functionality with this pin. In Gmoccapy you can abort a tool change by just clicking the "x" to close the tool changer dialog. But if I remove the abort pin, we should also think about the ability to abort in general.

c-morley:
An abort in a toolchange sounds like a real problem - particularly in a running program.
I would not promote the abort, unless it actually calls an abort to linuxcnc.
In qtvcp I hid the window controls and removed the abort button after realizing the problem it could be.

hansu:
Yes, aborting the toolchange also aborts the running program. But currently this is the only way to abort a program (using only the gmoccapy GUI) if the toolchange window is on top, because the main window is deactivated then.

I think following has to be done:
- allow clicking of the gmoccapy main window (to stop program) when toolchange window is popped up
- close toolchange window when `iocontrol.0.tool-change` becomes low
- remove `gmoccapy.toolchange-abort` pin
- maybe disable toolchange abort by closing the window

hansu:
I removed the abort pin and let the toolchange dialog close when the tool-change signal goes low (when the program is aborted).
I didn't remove the possibility to abort the tool change (and with it also the program) by clicking the "x" because it's an feature that Norbert introduced and and don't want to remove this without having talked to him.

c-morley:
Sounds good to me. It's a good upgrade.
Did i see you can commit now?

hansu:
> Did i see you can commit now?

Yes, you saw that correctly :slightly_smiling_face:

c-morley:
excellent! Welcome to the dark side :)
I approve of your judgement.
Please merge your work and thank you for your continued work on this project.

hansu:
Thanks :slightly_smiling_face:

PR #1649, opened 2022-03-05
Gmoccapy: fix gcode view issue (2.8)

hansu:
Fixes these three issues according to
https://forum.linuxcnc.org/gmoccapy/43045-after-saving-ngc-file-ask-for-do-you-want-to-exit-without-saving-the-changes
1. When editing a G-code file, a change was detected even when only the cursor changed.
2. After saving the changes, a warning for unsaved changes appeared again.
3. After creating a new file and 'save as', the save button was disabled.

### DIFF
--- a/lib/python/gladevcp/hal_sourceview.py
+++ b/lib/python/gladevcp/hal_sourceview.py
@@ -335,6 +335,9 @@ class EMC_Action_Save(_EMC_Action, _EMC_FileChooser):
 
 class EMC_Action_SaveAs(EMC_Action_Save):
     __gtype_name__ = 'EMC_Action_SaveAs'
+    __gsignals__ = {
+        'saved-as': (gobject.SIGNAL_RUN_FIRST, gobject.TYPE_NONE, ()),
+    }
 
     def __init__(self, *a, **kw):
         _EMC_Action.__init__(self, *a, **kw)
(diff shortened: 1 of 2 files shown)


c-morley:
Does this not break the feature for non Gmoccapy code?

hansu:
That's no problem, because only Gmoccapy used the 'change'-signal.

c-morley:
How do you know that? There are custom screens out in the wild that we can not predict what code they use.
2.8 is legacy code. We try hard not to break production code.
I also thought gscreen used the changed feature - but maybe it uses a different mechanism.

hansu:
Ah, you mean screens outside the LinuxCNC repository. I understand. I'll provide a solution without changing the interface. But for  the master branch we can have this solution, right?

hansu:
> I also thought gscreen used the changed feature - but maybe it uses a different mechanism.

Gscreen directly polls the get_modified() function of the SourceViews text buffer.

c-morley:
> Ah, you mean screens outside the LinuxCNC repository. I understand. I'll provide a solution without changing the interface. But for the master branch we can have this solution, right?

Yes. With released code one must be careful not to break code even if it was used wrong.
master branch one can deviate code much more, though I wouldn't break code without good reason.

hansu:
Now without changing the interface of hal_sourceview :slightly_smiling_face:

c-morley:
Thank you!

PR #1629, opened 2022-02-26
Improved Modbus to Hal docs + man page

hansu:
1. i created a "real" doc  page for MB2HAL instead of only the copy of the Howto.ini: https://hansu.github.io/linuxcnc-doc/drivers/mb2hal.html

2. And also added the Pins to the man page: https://hansu.github.io/linuxcnc-doc/man/man1/mb2hal.1.html
  To keep it short, i didn't write the pin names that are the same for two function codes, like this:

   ```
   fnct_01_read_coils:
   fnct_02_read_discrete_inputs:
       mb2hal.m.n.bit bit out
       mb2hal.m.n.bit-inv bit out
   ```
   Do you think this is clear? In the bash it looks okay, but in the HTML version are is more space between them, that might  be irritating...

PR #1661, opened 2022-03-11
gmoccapy: fix source typo

luzpaz:
Fix source typo `responce` to `response`

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -2188,16 +2188,16 @@ class gmoccapy(object):
             if pin.get():
                 self.halcomp["messages." + message[2] + "-waiting"] = 1
                 title = "Pin " + message[2] + " message"
-                responce = self.dialogs.show_user_message(self, message[0], title)
+                response = self.dialogs.show_user_message(self, message[0], title)
                 self.halcomp["messages." + message[2] + "-waiting"] = 0
         elif message[1] == "yesnodialog":
             if pin.get():
                 self.halcomp["messages." + message[2] + "-waiting"] = 1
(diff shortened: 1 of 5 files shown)


src/emc/usr_intf/gmoccapy/gmoccapy.py
```diff
@@ -2188,16 +2188,16 @@ def _show_user_message(self, pin, message):
...
         elif message[1] == "yesnodialog":
             if pin.get():
                 self.halcomp["messages." + message[2] + "-waiting"] = 1
                 self.halcomp["messages." + message[2] + "-response"] = 0
                 title = "Pin " + message[2] + " message"
-                responce = self.dialogs.yesno_dialog(self, message[0], title)
+                response = self.dialogs.yesno_dialog(self, message[0], title)
                 self.halcomp["messages." + message[2] + "-waiting"] = 0
-                self.halcomp["messages." + message[2] + "-response"] = responce
+                self.halcomp["messages." + message[2] + "-response"] = response
```
luzpaz, line 2200: this is probably wrong. Can we rename the variable something other than a typo of `response` ?

hansu: No, I think removing the typo is not wrong. Why do you think so?

By the way, I have also done these changes (in an open PR): https://github.com/LinuxCNC/linuxcnc/pull/1629/commits/7bf7745e894953afd64c90043e5ff720c3aba8d4 :smile:

luzpaz: So this PR is obsolete in that case?

hansu: Yeah kind of. Sorry.

PR #1666, opened 2022-03-13
Updated build instruction links

kentavv:
Updated paths

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,6 @@ See the file COPYING for more details.
 
 # The Build Process
 
-Refer to the file 'docs/src/code/building-linuxcnc.txt' for information
-about building and running the software, or look here:
-
-http://linuxcnc.org/docs/devel/html/code/building-linuxcnc.html
+Refer to the file 'docs/src/code/building-linuxcnc.adoc' for information
+about building and running the software, or look
+[here](https://linuxcnc.org/docs/devel/html/code/building-linuxcnc.html).
--- a/README.md
+++ b/README.md
@@ -40,6 +40,5 @@ See the file COPYING for more details.
 
 # The Build Process
 
-Refer to the file 'docs/src/code/building-linuxcnc.adoc' for information
-about building and running the software, or look
-[here](https://linuxcnc.org/docs/devel/html/code/building-linuxcnc.html).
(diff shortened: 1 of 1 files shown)


silopolis:
👍

README.md
```diff
@@ -40,7 +40,6 @@ See the file COPYING for more details.
 
 # The Build Process
 
-Refer to the file 'docs/src/code/building-linuxcnc.txt' for information
-about building and running the software, or look here:
-
-http://linuxcnc.org/docs/devel/html/code/building-linuxcnc.html
+Refer to the file 'docs/src/code/building-linuxcnc.adoc' for information
+about building and running the software, or look
+[here](https://linuxcnc.org/docs/devel/html/code/building-linuxcnc.html).
```
hansu, line 45: Hmm I don't know if it is clever to link to the devel docs. At some point 2.9 will be stable and the link has to be changed again.

kentavv: I switched the link to GitHub, but that could also be out of sync. If relative links always work in markdown that might be better. Maybe it's preferred to drop the link.

c-morley:
I think links will just need to be updated when 2.9 is actually made.
The link to dev should be still valid because dev is forever.

README.md
hansu: Yes, that makes also sense to link inside GitHub when coming from GitHub.

hansu:
> I think links will just need to be updated when 2.9 is actually made. The link to dev should be still valid because dev is forever.

Yeah I think we won't get rid of updating the links. 
But what do you think of adding a redirection from `linuxcnc.org/docs/2.9/ ` to  `linuxcnc.org/docs/devel/`? That would allow us to set all links now to /docs/2.9/ and they can stay so forever. Only if this is moved to a stable 2.9 branch, the links of the new master branch have to updated to /docs/2.10/

kentavv:
@andypugh The README refers to src/building-linuxcnc.txt, but the correct path appears to be src/building-linuxcnc.adoc. Just feedback from a first time builder trying to get started.

README.md
```diff
@@ -40,7 +40,5 @@ See the file COPYING for more details.
 
 # The Build Process
 
-Refer to the file 'docs/src/code/building-linuxcnc.txt' for information
-about building and running the software, or look here:
```
hansu, line 44: @kentavv  I'll add a link to http://linuxcnc.org/docs/2.9/html/code/building-linuxcnc.html as we have now a redirection from '2.9' to 'devel'

PR #1668, opened 2022-03-13
Follow-up to b016bf77225 (missing in previous commit)

luzpaz:


c-morley:
I really am a bad speller and proof reader it seems.
thanks

lib/python/qtvcp/lib/xembed.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, window, forward):
         d = display.Display()
         self.fw = drawable.Window(d.display, forward, 0)
 
-        self.foward = forward
+        self.forward = forward
```
luzpaz, line 66: Just double checking if this is kosher?

c-morley:
yes that should be done too.

PR #1669, opened 2022-03-16
fix for full circle interpretation

phillc54:
A bit of a corner case but it is repeatable.

The following gcode in the standard axis sim will show that a full circle is not being interpreted correctly.

`G91`
`G0 Z0.001`
`G90`
`F100`
`G0 X0.0 Y0.5`
`M3 $0 S1`
`G3 X0.0 Y0.5 I0.5`
`M5`
`G0 X0 Y0`
`M2`

### DIFF
--- a/src/libnml/posemath/_posemath.c
+++ b/src/libnml/posemath/_posemath.c
@@ -1846,7 +1846,7 @@ int pmCircleInit(PmCircle * const circle,
        2PI. */
     pmCartCartCross(&circle->rTan, &rEnd, &v);
     pmCartCartDot(&v, &circle->normal, &d);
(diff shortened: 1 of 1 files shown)


phillc54:
It may be that this PR is hiding some kind of timing issue. I did notice today that increasing the Z motion "fixes" the issue as does decreasing the MAX_ACCELERATION value for the Z axis and joint 2 in the ini file for the sim.

snowgoer540:
> Your email has been sat in my inbox awaiting urgent attention for a couple of weeks. I will try to get round to looking at it in detail in the next couple of days.

Andy, have you had any time to take a look at this and see if it is good to commit?

PR #1671, opened 2022-03-20
fix hal_lightbutton (#1266)

### LINKED ISSUE #1266: hal_lighbutton broken in master
Seems this may not have  been completed.
I noticed when using sim axis/panelui - it uses hal_lightbuttons.

The buttons don't show and there are errors produced in the terminal.
seems mostly about pango/cario coding

hansu:
Seems to work now. Would be nice if someone could double-check that.

closes #1266


c-morley:
works for me - thank you.

PR #1672, opened 2022-03-20
Gmoccapy & Gscreen: Stop keyboarg jogging on focus out event

### LINKED ISSUE #1511: Gmoccapy & Gscreen: Keyboarg jogging continues when window looses focus
If you switch to another window (or another window comes into the foreground) and a jogging key is held down - the jogging will continue until you go into the LinuxCNC window and press the jog key again.

One possibility to solve this would be to catch the Gtk `focus-out-event`  and stop all jogging on this.
Or we could check the status of the jogging keys periodically.

hansu:
closes #1511


gmoccapy:
Looks good for me, so please merge.

Norbert

PR #1677, opened 2022-03-24
docs: remove bad unicode characters

snowgoer540:
there are more in the .po files, as well as some of the language 
versions. I left those alone as to not mess with translation efforts, 
but can fix there too if needed.


smoe (approved):

docs/man/man1/hal_parport.1
```diff
@@ -90,17 +90,17 @@ The following lists the input and output pins by the type setting used in the cf
...
+Reads HAL \-out pins of port <p> and updates that port's physical output pins.
 .TP
 \fBparport.write\-all (funct)
 Reads HAL \-out pins of all ports and updates all physical output pins.
 .TP
 \fBparport.<p>.reset (funct)
-Waits until \fIreset\-time \fRhas elapsed since the associated write, then resets pins to values indicated by \fI\-out\-reset \fRand \fI\-out\-invert \fRsettings. reset must be later in the same thread as write. 'If '\fI\-out\-reset \fRis TRUE, then the reset function will set the pin to the value of \fI\-out\-invert\fR. This can be used in conjunction with stepgen’s doublefreq to produce one step per period. The stepgen stepspace for that pin must be set to 0 to enable doublefreq.
+Waits until \fIreset\-time \fRhas elapsed since the associated write, then resets pins to values indicated by \fI\-out\-reset \fRand \fI\-out\-invert \fRsettings. reset must be later in the same thread as write. 'If '\fI\-out\-reset \fRis TRUE, then the reset function will set the pin to the value of \fI\-out\-invert\fR. This can be used in conjunction with stepgen's doublefreq to produce one step per period. The stepgen stepspace for that pin must be set to 0 to enable doublefreq.
```
hansu, line 99: I think here we should use U+2019 [[1](https://en.wikipedia.org/wiki/Apostrophe#Unicode)] or U+02BC [[2](https://tedclancy.wordpress.com/2015/06/03/which-unicode-character-should-represent-the-english-apostrophe-and-why-the-unicode-committee-is-very-wrong/)]

docs/man/man1/hal_parport.1
```diff
@@ -90,17 +90,17 @@ The following lists the input and output pins by the type setting used in the cf
...
 .SH USAGE
 The hal_parport component is a driver for the traditional PC parallel port. The port has a total of 25 physical pins of which 17 are used for signals. The original parallel port divided those pins into three groups: data, control, and status. The data group consists of 8 output pins, the control group consists of 4 output pins, and the status group consists of 5 input pins.
 
-In the early 1990’s, the bidirectional parallel port was introduced, which allows the data group to be used for output or input. The HAL driver supports the bidirectional port, and allows the user to set the data group as either input or output. If configured as \fIout\fR, a port provides a total of 12 outputs and 5 inputs. If configured as \fIin\fR, it provides 4 outputs and 13 inputs.
+In the early 1990's, the bidirectional parallel port was introduced, which allows the data group to be used for output or input. The HAL driver supports the bidirectional port, and allows the user to set the data group as either input or output. If configured as \fIout\fR, a port provides a total of 12 outputs and 5 inputs. If configured as \fIin\fR, it provides 4 outputs and 13 inputs.
```
hansu, line 103: same here

docs/man/man1/hal_parport.1
```diff
@@ -147,7 +147,7 @@ For each parallel port handled by the hal_parport driver, a type can optionally
 
 If the type is not specified, the default is out.
 
-A type of epp is the same as out, but the hal_parport driver requests that the port switch into EPP mode. The hal_parport driver does not use the EPP bus protocol, but on some systems EPP mode changes the electrical characteristics of the port in a way that may make some marginal hardware work better. The Gecko G540’s charge pump is known to require this on some parallel ports.
+A type of epp is the same as out, but the hal_parport driver requests that the port switch into EPP mode. The hal_parport driver does not use the EPP bus protocol, but on some systems EPP mode changes the electrical characteristics of the port in a way that may make some marginal hardware work better. The Gecko G540's charge pump is known to require this on some parallel ports.
```
hansu, line 150: here also and so on ...
But I am neither a native english speaker nor a linguist. Maybe someone other could help out here.

snowgoer540: Yes, I agree, those are bad characters too.  I will either amend this PR, or start a new one to fix them as well.  Thanks!

snowgoer540: Sorry, I misunderstood your suggestion. 

For at least the docs I changed, my understanding is that U+0027 ( ' ) is the correct character as it is part of the Basic Latin Unicode block (encoded in one byte in UTF-8).   The look-alike characters can cause syntax errors in code and in terminal commands. 

When the English source files are converted to other languages, I am not so sure which characters are correct, that is why I left those files (_es, _fr, _CN, etc.) alone, as well as the .po files (those I left alone because I know they are being worked on).

hansu: >For at least the docs I changed, my understanding is that U+0027 ( ' ) is the correct character as it is part of the Basic Latin Unicode block (encoded in one byte in UTF-8). 

I am not sure. You can read through the references in https://github.com/LinuxCNC/linuxcnc/pull/1677#discussion_r834481961 and see that's is not that simple.  
But for the singe and double quotes i totally agree.

>as well as the .po files (those I left alone because I know they are being worked on).

Not on these which are in the repo, but they are handled by Weblate, so totally correct to not touch them.

hansu: i figured out that my concerns were without any reason. Asciidoc generates the correct character. (`'` turns into U+2019 )

PR #1678, opened 2022-03-24
Halshow: add filtering of tree -- need some feedback

hansu:
I added an entry to filter the tree view.
It supports regular expressions and can either search only inside the names or the full path.
Here an example for searching the full path:
image
But the problem is, that the search entry will be very small for the default window size.
So how can I shorten the strings? Leave the label "Filter" away? A symbol or abbreviation for "Full path"? Or a toggle button for "Full path"?
Something that doesn't break the intuitive usage.


silopolis:
Le jeu. 24 mars 2022 à 20:41, Hans ***@***.***> a écrit :

> I added an entry to filter the tree view.
> It supports regular expressions and can either search only inside the
> names or the full path.
>
Nice !


> But the problem is, that the search entry will be very small for the
> default window size.
> So how can I shorten the strings? Leave the label "Filter" away? A symbol
> or abbreviation for "Full path"? Or a toggle button for "Full path"?
> Something that doesn't break the intuitive usage.
>
First idea would be a symbol for filter, like the usual funnel and, yes,
another one for full path if you find one that "talks"

hansu:
>First idea would be a symbol for filter, like the usual funnel and, yes,
another one for full path if you find one that "talks"

Good idea, but looks a bit like a button

grafik
But for the other it is difficult to find sth. that is clear.
The only thing that comes into my mind is a toggle button with two different images:
grafik
grafik

silopolis:
Le ven. 25 mars 2022 à 18:39, Hans ***@***.***> a écrit :

> First idea would be a symbol for filter, like the usual funnel and, yes,
> another one for full path if you find one that "talks"
>
> Good idea, but looks a bit like a button
>
Not really for me.

But for the other it is difficult to find sth. that is clear.
>
What about something like the attached image ?

hansu:
> What about something like the attached image ?

Which attached image?

c-morley:
put the filter entry box on the very bottom where it can be full width?

hansu:
> put the filter entry box on the very bottom where it can be full width?

You mean above or below the command entry? I think one wouldn't associate that with the tree view.
Maybe I can put the checkbox in the second row. But I didn't want to waste so much space with that.

hansu:
I think I have it. I'll add a preferences button that makes the settings frame pop up.
And the filter icon can be eliminated by some text shown when the entry is empty.

Screenshot_2022-03-28_14-37-53

andypugh:
I am happy to merge this now and see how it goes, if you are.

hansu:
I haven't implemented my idea of the last comment. I'll merge it as soon I have implemented this.

PR #1680, opened 2022-03-24
tp: fix missing state tag during blended motion

snowgoer540:
TP was not posting a valid state tag during blend motion (G64). The 
state tag for a motion was not being copied into the blended motion when 
it was created. This caused it to report the default state tag and fail 
the validity check down the track. This fix simply copies the state tag 
from the previous motion when the blend is created.

This fixes the erroneous "Parameter file name is missing" or "Unknown 
oword number" error messages that would have been received in all GUIs 
when motion was Aborted during a blend move.

While I can take credit for getting to the root of this problem and 
figuring out how to repeat it 100% of the time, the credit goes to Rob 
Ellenberg for helping me understand what was going on and ultimately 
providing the solution.

### DIFF
--- a/src/emc/tp/tp.c
+++ b/src/emc/tp/tp.c
@@ -839,6 +839,9 @@ STATIC int tpInitBlendArcFromPrev(TP_STRUCT const * const tp,
     //In the future, radius may be adjustable.
     tcFinalizeLength(blend_tc);
 
+    // copy state tag from previous segment during blend motion
+    blend_tc->tag = prev_tc->tag;
+
     return TP_ERR_OK;
 }
 


hansu:
Nice!
It seems you are well familiar with the the interpreter. Do you think you can have a look at another interpreter related issue? 
Currently, the interpreter sends only line numbers for commands with motion. So all M-codes for example will not be highlighted. It would be nice if all lines could be highlighted (especially very helpful when stepping through).

snowgoer540:
Well, I am not sure I would go that far, but I am _becoming_ familiar with it...  As for the M code line highlighting, I wonder why this wasn't done up front?

hansu:
But at least you know where to look and you know the start point where the interpreter begins to read the file...

>I wonder why this wasn't done up front?

I am wondering too. I have no idea why.

snowgoer540:
Is there any discussion on this on the forum at all?

hansu:
I only hear now and then in the forum that users complain about that. 
And it also bothers me :)

spumco1234:
[Never used github before - set up an account just to reply to this conversation]

> I only hear now and then in the forum that users complain about that. And it also bothers me :)

It bothers me too.


> Is there any discussion on this on the forum at all?

I've brought it up indirectly  - perhaps accidentally - as I'm learning how LCNC works during my first implementation on a retrofit.  I got so focused on the hardware/electronics/HAL side of things, I kind of ignored the 'using it' part until very recently.

But the issue/question is:

Where is LCNC different from 'standard' or conventional CNC controls?  I'm not talking about Fanuc vs. Siemens vs. Other dialects; rather, certain conventions that are common to many, many OEMs implementations.

If you take a reasonably experienced CNC operator/programmer out of a working shop and tell them to run an LCNC-equipped machine, what functions would cause them to say "_WTH were they thinking?_"

Short list off the top of my head:
- Stepping through a program vs. single-block _Mode_.  (partial workarounds are available)
- Stepping/single-block does not stop at each line (see above), only at lines with motion.
- Feedhold doesn't work on non-motion lines (see above)
- No GOTO statements in programs

I'd be happy to continue this discussion in the forum, or in a new PR, or anywhere else.

hansu:
a little bit more discussion about his: https://github.com/LinuxCNC/linuxcnc/issues/1694

jallwine:
Does this tag tie into any of the line number values available in the Python linuxcnc.stat object?

http://linuxcnc.org/docs/2.8/html/config/python-interface.html

Any of the following?
current_line
motion_line
settings[0]

I've found that settings[0] seems to give the best number for what line number is running, but could current_line or motion_line have been jumping around because the blend between each line was causing it to show the wrong value? I'm not sure what the settings[0] number actually represents, but from observation it seemed to be the most reliable line number while running a program.

PR #1685, opened 2022-03-29
Added 'eatmydata' to optimize auto-builds

smoe:
This is a bit of an experiment. In principle we could run the build from a ramdisk. "eatmydata" gets a bit closer to that in that a process does not wait on a write to disk to be completed. Particularly with the upcoming humongous I/O with the po4a-generated documentation this is expected to boost build times. If this works. This PR will tell.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,10 +24,11 @@ jobs:
     - name: Build RIP & test
       run: |
         ./scripts/travis-install-build-deps.sh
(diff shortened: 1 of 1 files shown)


hansu:
I thought we push all po4a related to a separate branch ,not master?

smoe:
I had thought about it, but then again, introducing build-optimizations is kind of independent from docs, was only motivated by it, and we either want it or we don't. That is why I went for a PR against master.

PR #1687, opened 2022-03-29
Use more standard debian tools for github package build.

petterreinholdtsen:
Also use eatmydata a bit more to speed up build.


smoe (approved):
You could add "eatmydata" also to apt-get install.

PR #1689, opened 2022-03-30
Avoid '*' turning to <strong> in HTML.

petterreinholdtsen:


hansu:
Does this provoke a bug in po4a? It just sounds a bit weird as in https://github.com/LinuxCNC/linuxcnc/pull/1677 these characters were changed the other way around.

petterreinholdtsen:
Nope.  It produces the wrong HTML output from a2x.  But I did discover it via the reformatting from po4a.

petterreinholdtsen:
You can see the incorrect formatting in https://github.com/LinuxCNC/linuxcnc/blob/master/docs/README.adoc .

hansu:
And `''` makes is italic where it supposed to be a monospace font  which  `` does

PR #1691, opened 2022-03-31
RFD: Reduce set of required packages -> sep packages for HAL and GUIs?

smoe:
I just wanted to give latency-test a spin and then ran into a Python conflict in Debian unstable since python3-opencv is not yet ready for python3.10.

And then I thought - waitaminute - opencv? Yip, very nice for integrating video. But what about those who only want to use HAL or (like me) aimed at insights on the latency of their hardware.

In this PR I propose to move some hard dependencies of the Debian package to "recommended", which means these get installed by default but if there is a problem then not. But what this PR is really asking is if there should be a more fine-grained set of packages for LinuxCNC. There could be linuxcnc-uspace-hal, maybe also linuxcnc-uspace-gui. The current linuxcnc-uspace package could depend on them all so there would not bit a difference in user experience.

### DIFF
--- a/debian/control.main-pkg.in
+++ b/debian/control.main-pkg.in
@@ -19,9 +19,6 @@ Depends:
     python3-opengl,
     python3-configobj,
     python3-xlib,
-    @PYTHON_IMAGING@,
-    @PYTHON_IMAGING_TK@,
-    @QTVCP_DEPENDS@,
     libgtksourceview-3.0-dev,
     tcl@TCLTK_VERSION@,
     tk@TCLTK_VERSION@,
@@ -31,6 +28,10 @@ Depends:
     tclx,
     procps, psmisc,
     udev
+Recommends:
+    @PYTHON_IMAGING@,
+    @PYTHON_IMAGING_TK@,
+    @QTVCP_DEPENDS@
 Suggests: mesaflash
 Description: motion controller for CNC machines and robots
  LinuxCNC is a fully-realised CNC machine controller that can interpret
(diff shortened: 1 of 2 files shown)


silopolis:
Le jeu. 31 mars 2022 à 12:33, Steffen Möller ***@***.***> a
écrit :

> But what this PR is really asking is if there should be a more
> fine-grained set of packages for LinuxCNC. There could be
> linuxcnc-uspace-hal, maybe also linuxcnc-uspace-gui. The current
> linuxcnc-uspace package could depend on them all so there would not bit a
> difference in user experience.
>

As you read my min 😉, needless to tell I'm all in for that! And even
further, split GUIs into separate packages so s/o running AXIS doesn't have
to pull Qt* etc. Likewise, move ClassicLadder to its own package and make
it optional.

So one installing linuxcnc-uspace would still pull the whole enchilada, but
another could just install linuxcnc-uspace-qtplasmac and pull just the
needed dependencies to control its cutter.

And I'm sure an orthodox Debianista would favor even finer grain...

andypugh:
I think that these ideas are appealing to programmers and developers. But it is worth remembering that many (most?) of our users are machinists who are already a little bit scared to be using Linux. 
I don't see this idea as much of a win for them.

smoe:
They would not. But they would just install linuxcnc-uspace as before or continue using the USB stick (seen this a few times on YouTube). I do not think that any harm would be done.

That all said, I would be very much like a thread somewhere about what we can do to further improve the experience with the LinuxCNC ISO. In my understanding, having the regular Debian package is only rendering the crafting of such an USB stick technically easier but the challenge to compile and present a nice working environment remains.

silopolis:
Le jeu. 31 mars 2022 à 23:25, Steffen Möller ***@***.***> a
écrit :

>  but the challenge to compile and present a nice working environment
> remains.
>

True. First improvement would be to add xfce extra packages to have a not
so bare desktop. Specifically, lack of a graphical network config tool by
default is the first abyss new users fall into just after logging in 😬



>

PR #1696, opened 2022-04-02
Update pyvcp_widgets.py

ALatSMT:
convert floating point to int


ALatSMT:
looks like someone fixed this in the dial widget also

c-morley:
Thank you

PR #1700, opened 2022-04-05
Debian uses -A and -B flags to decide if dpkg-buildpackages builds docs

smoe:
Hello,

The autobuilders of Debian are just a bit opinionated on how they think they want everything configured. And I can upload only one kind of Debian package, not multiple, to switch between "documentation builds" or not. This PR
 * resets the BUILD_DOCS configure flag in configure.ac if key packages like asciidoc or dblatex are not installed but does not error out of the build. This avoids the need to install asciidoc and dblatex when only compiling the binary packages (running dpkg-buildpackage -B)
 * fixes d/rules' install target to tolerate when files are not existing - was important both for -A and -B

I am not 100% confident that this is the last patch in this direction. We'll see. Without addressing this, we will not get the build demons to function.

### DIFF
--- a/debian/control.top.in
+++ b/debian/control.top.in
@@ -15,6 +15,8 @@ Build-Depends:
     desktop-file-utils,
     gettext,
     intltool,
+    asciidoc (>= 8.5),
+    dblatex (>= 0.2.12),
     libboost-python-dev,
     libepoxy-dev,
     libgl1-mesa-dev | libgl1-mesa-swx11-dev,
--- a/debian/control.top.in
+++ b/debian/control.top.in
@@ -15,8 +15,6 @@ Build-Depends:
     desktop-file-utils,
     gettext,
     intltool,
-    asciidoc (>= 8.5),
-    dblatex (>= 0.2.12),
     libboost-python-dev,
     libepoxy-dev,
     libgl1-mesa-dev | libgl1-mesa-swx11-dev,
--- a/debian/rules.in
+++ b/debian/rules.in
@@ -29,8 +29,8 @@ export DATE=$(shell LANG=C date --date='@$(TIMESTAMP)' '+%d\ %b\ %Y')
 export TIME=$(shell LANG=C date --date='@$(TIMESTAMP)' '+%T')
 
 kernel_version = @KERNEL_VERSION@
-enable_build_documentation = @ENABLE_BUILD_DOCUMENTATION@
 configure_realtime_arg = @CONFIGURE_REALTIME_ARG@
+enable_build_documentation = @ENABLE_BUILD_DOCUMENTATION@
 DESTDIR=$(CURDIR)/debian/@MAIN_PACKAGE_NAME@
 DEV_PACKAGE_NAME=@MAIN_PACKAGE_NAME@-dev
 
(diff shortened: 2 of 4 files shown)


smoe:
Good news, this is what was now successfully built (4th attempt) and can be installed for surprisingly many architectures from Debian: https://buildd.debian.org/status/package.php?p=linuxcnc

rodw-au:
While working with conditional docs packaging I wonder if it were possible for local builds to have some kind of configure flag to select the docs to build? They take a long time to build so it might be nice to support something that debian bots don't see like:
lang=all, lang=en, lang=de, lang=en de fr 
That would let users just build what they need and save a lot of time....

Also I think the docs will need updating for the dpkg-buildpackage syntax http://linuxcnc.org/docs/devel/html/code/building-linuxcnc.html#_building_debian_packages

silopolis:
Le mer. 6 avr. 2022 à 21:39, Rod Webster ***@***.***> a
écrit :

> While working with conditional docs packaging I wonder if it were possible
> for local builds to have some kind of configure flag to select the docs to
> build? They take a long time to build so it might be nice to support
> something that debian bots don't see like:
> lang=all, lang=en, lang=de, lang=en de fr
> That would let users just build what they need and save a lot of time....
>

Built docs for the first time this afternoon and had the same idea 👍
But now I wonder how would that be useful with the upcoming translation
toolchain 🤔

rodw-au:
I confirmed that linuxcnc-uspace installs with apt on unstable today.
I found an article about repo pinning today so I am keen to try that method to install on Bullseye. No need to do a full upgrade to unstable then. That would be cool!

silopolis:
Le jeu. 7 avr. 2022 à 09:52, Rod Webster ***@***.***> a
écrit :

> I confirmed that linuxcnc-uspace installs with apt on unstable today.
> I found an article about repo pinning today so I am keen to try that
> method to install on Bullseye. No need to do a full upgrade to unstable
> then. That would be cool!
>
Package pinning is another Debian magic ! You will love it :)

smoe:
There are issues reported with the Debian package that I do not understand, yet. See https://forum.linuxcnc.org/9-installing-linuxcnc/45634-current-master-2-9-building-debs-error and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1012789
The problem is reproducible. The first window to select a configuration works. Then starting., e.g., axis_mm fails.

smoe:
I just found https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1012825 that blames the tiff library for removing a symbol.

havardAasen:
Yes, the strange thing is that I think upstream means those symbols to be private, but their still exported and used by other libraries.
My thoughts is that if there had been a major SONAME bump, then `lbtk-img` would have been rebuilt, and picked up the missing symbol in `tiff`.

andypugh:
> My thoughts is that if there had been a major SONAME bump, then `lbtk-img` would have been rebuilt, and picked up the missing symbol in `tiff`.

Which would be less annoying if we used Tiff anywhere...

smoe:
In a way, these kind of problems are exactly where there is a "Debian unstable". A bit earlier this month the package arrived also in testing (https://tracker.debian.org/pkg/tiff). Until this is resolved, I helped myself by installing the version from the beginning of last month. As root, do
```
echo "deb https://snapshot.debian.org/archive/debian/20220501T085721Z/ sid main
" >> /etc/apt/sources.list
apt update
sudo apt install libtiff-dev=4.3.0-7 libtiff5=4.3.0-7 libtiffxx5=4.3.0-7
```
back as a regular user then start
```
linuxcnc
```
and it should just work.

jepler (approved):
I'm not thrilled by this, particularly
```
install:
	$(MAKE) install-docs || echo "Ignored."…
```
because it has the potential to hide a real problem. However, I don't have a better solution to offer at this time and it's an improvement over the status quo for building debian packages, which we value.

PR #1707, opened 2022-04-10
Copy paste mistake

TooTall18T:
The German part is a copy of line 2978 .


hansu:
The translations are now handled by Weblate. Changing them in this way would lead to conflicts.
I have done this change on Weblate now: https://hosted.weblate.org/translate/linuxcnc/linuxcnc/de/?q=+source%3A%3D%22Probe+is+already+tripped+when+starting+G38.2+or+G38.3+move%22+&sort_by=-priority%2Cposition&offset=1

For further translations/translation fixes please use Weblate. Thanks!

PR #1710, opened 2022-04-11
Synced with _es, weird redundancy in core-components

smoe:
My hunch is that there once was a git merge conflict that was not resolved properly. Someone with much vertical screen estate or better memory please check.

### DIFF
--- a/docs/src/config/core-components.adoc
+++ b/docs/src/config/core-components.adoc
@@ -6,28 +6,31 @@
 See also the man pages 'motion(9)'.
 
 [[sec:motion]]
-== Motion
+== Motion(((Motion)))
 
-These pins and parameters are created by the realtime 'motmod' module. +
-This module provides a HAL interface for LinuxCNC's motion planner. +
-Basically motmod takes in a list of waypoints and generates a nice 
-blended and +
-constraint-limited stream of joint positions to be fed to the motor drives. +
+These pins and parameters are created by the realtime 'motmod' module.
 
-Optionally the number of Digital I/O is set with 'num_dio'. +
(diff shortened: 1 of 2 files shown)


smoe:
On a sidenote - I typically tried not to touch the _en version too much. When po4a gives an error, though, it is very bad in telling what the error is, so you start simplifying and then desimplify again to make it nice for the eye again. This causes the one or other smallish diff that seems unmotivated.
From my weblate experience I recall that weblate counts line differences. When desimplifying, I have also started prepare one sentence per line when this felt "reasonably close" to how things are formatted already.

docs/src/config/core-components.adoc
```diff
@@ -78,235 +83,163 @@ Example:
 unlock_joints_mask=0x38 selects joints 3,4,5
 ----
 
+selecciona articulaciones 3,4,5
```
hansu, line 86: Can this be removed?

smoe: Yes, please remove in _es and _en. It means "Selects joints 3,4,5". I just copied it since I was uncertain if this should be in or out and the Spanish had it.

hansu:
```suggestion

```

docs/src/config/core-components_es.adoc
```diff
@@ -78,7 +81,7 @@ articulación(es). El bit menos significativo de la máscara selecciona la artic
 Ejemplo:
 
 ----
-   unlock_joints_mask = 0x38 (111000b) 
+unlock_joints_mask = 0x38 (111000b) 
 ----
 
 selecciona articulaciones 3,4,5
```
hansu, line 87:
```suggestion

```

PR #1711, opened 2022-04-13
Big merge request for all the _es <-> _en synced files.

smoe:
Merged with docs-devel and it looks fine from here.


smoe:
My branch had no diffs outside docs/ against master. I am somewhat uncertain what this now looks like after the merge with docs-devel. You will manage :)

docs/src/user/user-intro.adoc
```diff
@@ -32,7 +32,7 @@ complete system:
 The below illustration is a simple block diagram showing what a typical 3-axis, CNC mill with stepper
 motors might look like:
 
-.Simple LinuxCNC Controlled Machine
+//.Simple LinuxCNC Controlled Machine
```
hansu, line 35: special reason for this and following?

smoe: Limitation by po4a - the same text appearing in different contexts and this thing freaks out. I presume there is an image::images line beneath.

hansu: Oh ok. Yes there is an image below.  I just wonder as we are using image captions at at some more locations..

smoe: We need a style guide. ;) I am not exactly sure about what the difference is between these image captions and the first argument given to the image[] or how this relates to the alt="" argument. These comments are dead easy to undo, so I am not worried.
You will also see that I converted all alt="" to a first argument setting.

hansu: I think
`image::images/icons/home.png[Alt text]` 
should be the same as
`image::images/icons/home.png[alt="Alt text"]`
but completely different from
```
.Image caption
image::images/icons/home.png[]
```

smoe: That is what my intuition also told me. You will find though that in 99.9% of all cases in the LinuxCNC docs the first argument is identical to the first argument/the alt attribute. A majority of image calls has no caption at all and only the first argument/alt attribute.

silopolis:
Le mer. 13 avr. 2022 à 11:26, Steffen Möller ***@***.***> a
écrit :

>
> In docs/src/user/user-intro.adoc
> <https://github.com/LinuxCNC/linuxcnc/pull/1711#discussion_r849274678>:
>
> > @@ -32,7 +32,7 @@ complete system:
>  The below illustration is a simple block diagram showing what a typical 3-axis, CNC mill with stepper
>  motors might look like:
>
> -.Simple LinuxCNC Controlled Machine
> +//.Simple LinuxCNC Controlled Machine
>
> That is what my intuition also told me. You will find though that in 99.9%
> of all cases in the LinuxCNC docs the first argument is identical to the
> first argument/the alt attribute. A majority of image calls has no caption
> at all and only the first argument/alt attribute.
>
I've solved this by exercising my rephrasing skill in too many places 😅

docs/src/user/user-intro.adoc
hansu: >Limitation by po4a - the same text appearing in different contexts and this thing freaks out. 

You mean po4a can't handle it if two equal source strings come from different contexts?

silopolis:
Le mer. 13 avr. 2022 à 11:46, Hans ***@***.***> a écrit :

>
> In docs/src/user/user-intro.adoc
> <https://github.com/LinuxCNC/linuxcnc/pull/1711#discussion_r849293595>:
>
> > @@ -32,7 +32,7 @@ complete system:
>  The below illustration is a simple block diagram showing what a typical 3-axis, CNC mill with stepper
>  motors might look like:
>
> -.Simple LinuxCNC Controlled Machine
> +//.Simple LinuxCNC Controlled Machine
>
> Limitation by po4a - the same text appearing in different contexts and
> this thing freaks out.
>
> You mean po4a can't handle it if two equal source strings come from
> different contexts?
>
Not always well If I understood correctly :/

smoe:
> Le mer. 13 avr. 2022 à 11:26, Steffen Möller ***@***.***> a écrit :
> ***@***.**** commented on this pull request. ------------------------------ In docs/src/user/user-intro.adoc <[#1711 (comment)](https://github.com/LinuxCNC/linuxcnc/pull/1711#discussion_r849274678)>: > @@ -32,7 +32,7 @@ complete system: The below illustration is a simple block diagram showing what a typical 3-axis, CNC mill with stepper motors might look like: -.Simple LinuxCNC Controlled Machine +//.Simple LinuxCNC Controlled Machine That is what my intuition also told me. You will find though that in 99.9% of all cases in the LinuxCNC docs the first argument is identical to the first argument/the alt attribute. A majority of image calls has no caption at all and only the first argument/alt attribute.
> I've solved this by exercising my rephrasing skill in too many places

I have been guilty of that a few times, too. But then again: What is the benefit for our readers? A true caption can be multiple lines long. The title of an image is short. I would need to understand better how these captions/titles are rendered.

A consequence is that we will now conflict over these comments. And while this will not affect the .po file, it is still unfortunate.

hansu:
> Not always well If I understood correctly :/

But in this case I would rather remove the alt text than the caption. The alt text should be only shown when the image is missing. And even if the image is missing we still have the caption. So no really need to have  an alternative text when we have a caption.

smoe:
Please try at home:
```
$ cat << EOTEST > test.adoc 
:lang: en

= A header goes here.

.This is the caption
image::images/calibration-1.png["This is the first argument",alt="This is alt."]
EOTEST

$ asciidoc test.adoc
$ lynx --dump test.adoc
a header goes here.

   This is the first argument
   Figure 1. This is the caption
     __________________________________________________________________

   Last updated 2022-04-13 12:26:06 CEST
```
but this is because it is a text-only representation. The, .html is like
```
<img src="images/calibration-1.png" alt="This is the first argument" />
</div>
<div class="title">Figure 1. This is the caption</div>
```
So, "alt" is completely ignored with the first the first argument present. The caption is always shown and gets a "Figure 1" prefix.

My suggestion is to keep the captions commented out and we make it a mission of the style guide to make the captions true captions over time. A true caption in my understanding is something that explains the image without the need to read through the text.

If you want to change everything from first-argument to captions, then I understand that this is more invasive than letting it be how it is. And to have truly well formulated captions would be quite a feat.

silopolis:
Le mer. 13 avr. 2022 à 12:38, Steffen Möller ***@***.***> a
écrit :

> .This is the caption
> image::images/calibration-1.png["This is the first argument",alt="This is alt."]
>
> AFAI this is not correct. First argument IS alt text so its either:

.This is the caption
image::images/calibration-1.png[alt="This is alt."]

or

.This is the caption
image::images/calibration-1.png["This is alt."]

or

image::images/calibration-1.png[alt="This is alt.",title="This is the caption"]

or

image::images/calibration-1.png["This is alt.",title="This is the caption"]

docs/src/user/user-intro.adoc
```diff
@@ -103,41 +103,41 @@ AXIS:: <<cha:axis-gui,AXIS>>, the standard keyboard GUI interface. This is also
   Configuration Wizard is used to create a desktop icon launcher:
 
 [[fig:axis-graphical-interface]]
-.AXIS graphical interface
+//.AXIS graphical interface
 image::../gui/images/axis.png["Axis, the standard keyboard GUI interface",align="center"]
```
hansu, line 107: Here shouldn't be any problem as caption != alt text

hansu:
>If you want to change everything from first-argument to captions, then I understand that this is more invasive than letting it be how it is. 

No, I don't want that right now. I only will not turn existing captions into comments. I don't want to strip parts of the docs until we refactor them.

docs/src/user/user-intro.adoc
```diff
@@ -103,41 +103,41 @@ AXIS:: <<cha:axis-gui,AXIS>>, the standard keyboard GUI interface. This is also
...
 image::../gui/images/gmoccapy_3_axis.png["gmoccapy, a touch screen GUI based on Gscreen",align="center"]
 
 NGCGUI:: <<cha:ngcgui,NGCGUI>>, a subroutine GUI that provides wizard-style programming of G code. NGCGUI may be
   run as a standalone program or embedded into another GUI as a series of tabs. The following screen shot
   shows NGCGUI embedded into Axis:
 
 [[fig:ngcgui-graphical-interface-into-axis]]
-.NGCGUI graphical interface integrated into Axis
+//.NGCGUI graphical interface integrated into Axis
 image::../gui/images/ngcgui.png["NGCGUI graphical interface integrated into Axis",align="center"]
```
hansu, line 134: And for those I would suggest

```suggestion
.NGCGUI graphical interface integrated into Axis
image::../gui/images/ngcgui.png[align="center"]
```
or
```suggestion
.NGCGUI graphical interface integrated into Axis
//image::../gui/images/ngcgui.png["NGCGUI graphical interface integrated into Axis",align="center"]
image::../gui/images/ngcgui.png[,align="center"]
```
or
```suggestion
.NGCGUI graphical interface integrated into Axis
image::../gui/images/ngcgui.png["The NGCGUI graphical interface integrated into Axis",align="center"]
```
(slighly modifiyng the alt text should fix the po4a problem, right?)

hansu (changes requested):

silopolis:
Sorry sent too quickly :/

Le mer. 13 avr. 2022 à 12:38, Steffen Möller ***@***.***> a
écrit :

> My suggestion is to keep the captions commented out and we make it a
> mission of the style guide to make the captions true captions over time. A
> true caption in my understanding is something that explains the image
> without the need to read through the text.
>
> If you want to change everything from first-argument to captions, then I
> understand that this is more invasive than letting it be how it is. And to
> have truly well formulated captions would be quite a feat.
>
I don't think they're caption in that extended meaning but rather image
"titles"... as the parameter name suggests.
Alt test is to be displayed when no image is available or not possible to
display, read by text to speech and, once upon a time, important for SEO.
Title should briefly describe block (be it image, table, code block,...)
for sound content and for reference in appendix tables

silopolis:
Le mer. 13 avr. 2022 à 12:54, Hans ***@***.***> a écrit :

> +.NGCGUI graphical interface integrated into Axis
>
> +image::../gui/images/ngcgui.png[align="center"]
>
> Title must be on the line immediately above its corresponding block

> (slighly modifiyng the alt text should fix the po4a problem, right?)
>
Exact

smoe:
I recall to have shifted about ~20 or so captions to alt-texts. In part this could be motivated by the translation having it as an alt. Lacking the style guide, I considered it reasonable to have everything equal for all images. And frankly, since we need alt texts anyway, I am happy with how I now did it and suggest we come up with decent captions for all images as a To-Do item. I prefer a clear-cut out-commenting over a partial rewrite when what is truly missing is a real caption.

smoe:
> Le mer. 13 avr. 2022 à 12:54, Hans ***@***.***> a écrit :
> +.NGCGUI graphical interface integrated into Axis +image::../gui/images/ngcgui.png[align="center"] Title must be on the line immediately above its corresponding block (slighly modifiyng the alt text should fix the po4a problem, right?)
> Exact

Technically that is correct, just sometimes the authors uses the ".text" as some header to separate paragraphs, not as a caption. Style-guide.

smoe:
My hunch is that the .po file will just not feature an entry if it is commented out. Leaving it in, you also need to address all the translations. For a quick sync and generation of the .po files I suggest to follow my path. Uncommenting is trivial. The rest is for weblate.

smoe:
The changes proposed also need to be changed in the translations. What is special about this branch is that _es and _en are in sync. And since everything is not as easy as it seems, please run the migrate script locally. Often it is the nearby header that has the same wording which just just as bad as the first argument being the same.

docs/src/user/user-intro.adoc
hansu: > I recall to have shifted about ~20 or so captions to alt-texts. In part this could be motivated by the translation having it as an alt. Lacking the style guide, I considered it reasonable to have everything equal for all images. And frankly, since we need alt texts anyway, I am happy with how I now did it and suggest we come up with decent captions for all images as a To-Do item. I prefer a clear-cut out-commenting over a partial rewrite when what is truly missing is a real caption.

I had a closer look and these captions are introduced by you or Jérémie during this process. So in this case it is fine for me.
But why don't keep them in cases where they don't make problems (caption != alt) ?

smoe: caption != alt != header - please also check that.
And, the translation needs the same caption or alt or header.
There may be cases in which I have anticipated a problem and already commented something out prior to the error occurring. But otherwise, if there was no problem, then I would not have touched the file in the first place.

smoe:
I just noticed that all checks passed again. Great!

docs/src/user/user-intro.adoc
hansu: ah ok a header with the same name some lines before...

smoe: Lucky me :)

silopolis:
Another one for the pleasure:

This breaks:

```[[cap:point-final-compensation]]
.Point final de la compensation
image::images/comp-path_fr.png["Point final de la
compensation",align="center"]```

This passes:

```[[cap:point-final-compensation]]
.Point final de la compensation(((Point final de la compensation)))
image::images/comp-path_fr.png["Point final de la
compensation",align="center"]```

hansu:
> Another one for the pleasure: This breaks: ```[[cap:point-final-compensation]] .Point final de la compensation image::images/comp-path_fr.png["Point final de la compensation",align="center"]``` This passes: ```[[cap:point-final-compensation]] .Point final de la compensation(((Point final de la compensation))) image::images/comp-path_fr.png["Point final de la compensation",align="center"]```

Sure that it is not the other way around? :smile:

petterreinholdtsen:
For the record, as I understand the po4a limitation, it only affect the matching of source and translated strings when trying to generate the initial PO files in the migration.  I am not aware of any such limitation with using po4a to generate the translated documents using the source and PO files.  So the workaround to avoid identical strings in multiple contexts is only needed to generate the initial PO files for french and spanish, and is not needed in master moving forward.

smoe:
What Petter just stated was (mostly) obvious to me but maybe not to you all. IMHO the consequence is that you are investing too much into this now. You should merge my branch with docs-devel so @silopolis can sync most easily with it, and prior to merging with master you can then remove all the commented-out image headers again in a presumed-simple sed script, so master is not affected. This branch now hanging here is not good for @silopolis 's syncing.

hansu:
Okay. i created a PR to have an overview: https://github.com/LinuxCNC/linuxcnc/pull/1713

@smoe: Yeah sure, I haven't had the time yesterday to have a closer look at it. but it's on the way :sunglasses:

silopolis:
Le jeu. 14 avr. 2022 à 10:35, Steffen Möller ***@***.***> a
écrit :

> What Petter just stated was (mostly) obvious to me but maybe not to you
> all. IMHO the consequence is that you are investing too much into this now.
> You should merge my branch with docs-devel so @silopolis
> <https://github.com/silopolis> can sync most easily with it, and prior to
> merging with master you can then remove all the commented-out image headers
> again in a presumed-simple sed script, so master is not affected. This
> branch now hanging here is not good for @silopolis
> <https://github.com/silopolis> 's syncing.
>
Yes please, merge it if !
Like I said, I'll ync with it and make another check to verify and fix
whatever I may have broken with fr_sync

hansu:
Okay done, some comments here: https://github.com/smoe/linuxcnc/pull/1/files

PR #1718, opened 2022-04-24
make a pretty index.html for the PDF docs

SebKuzminsky:


andypugh:
The PDF index I like. 
I don't understand why the configure error message is in here too, nor why it splits the message half warn and half error?

hansu:
Is this just coincidence? I recently did something very similar to this and asked Andy how to bring it on the webserver:
https://github.com/LinuxCNC/linuxcnc/commit/019d08e7f49bc573e81c0df3f7627be05bab63f2

andypugh:
It's not entirely coincidence. I put in an index (generated on the server) a few weeks ago after the docs discussion, but it was overwritten next time the docs were generated. I asked Seb if he could add something on the buildbot in the upload script instead, and he came up with this.

hansu:
> I asked Seb if he could add something on the buildbot in the upload script instead, and he came up with this.

Ah okay, thanks! This looks good basically but I don't know why his generated index.html appears on the web server and my static one not...

SebKuzminsky:
> I don't understand why the configure error message is in here too, nor why it splits the message half warn and half error?

It's just something i happened to notice and fix while working on the PDF index thing.  The old `AC_MSG_ERROR` had an embedded newline (`\n`), which (on the Wheezy machine I was working on) did not get turned into an actual newline, and passed through as the two-character sequence `\` and `n`.  I tried to split it into two `AC_MSG_ERROR` messages, but only the first one was printed, then `configure` stopped with an error.  So I turned the first one into a `AC_MSG_WARN` which prints but does not exit.

I was too lazy to make this a second PR, but i should have mentioned my rationale in the original message....

> I wonder why his generated index.html appears on the web server and my static one not...

It takes a small update to the buildbot to copy the index to the right place for uploading to the server.

andypugh:
I say just push it.

SebKuzminsky:
I'll try to merge 2.8 to master tonight.

hansu:
Okay I merged the single japanese man page docs into one document, then it works with this solution: http://linuxcnc.org/docs/scratch/pdf/
Then I'll merge 2.8-docs-japanese to 2.8, okay?

andypugh:
It seems like that ought to work until we get something more permanent set up.

PR #1721, opened 2022-04-27
docs: added japanese PDF docs from MasaoSakai

hansu:
related to https://github.com/LinuxCNC/linuxcnc/pull/1718


hansu:
@andypugh ?

andypugh:
The comment that docs do not build is probably true only for older distros. 
As master is no-longer supporting anything older than Buster, and buster works with xetex (and, indeed, xetex is used):
https://github.com/LinuxCNC/linuxcnc/commit/57c946feaf6f08230e3297d5294f81e159671b67
Both the comments, and the workaround _might_ be outdated?

hansu:
Hmm but the japanese docs I want to add are static ones and don't need to be built. So I don't really see the relation to the commit you mentioned.

andypugh:
Sorry, I was just going through the commit diff and thought I would comment on something that I noticed. 

As far as I am concerned, this can go in.

PR #1729, opened 2022-05-05
docs: lowercase Documentation.pot

silopolis:
Merge after #1724

PR #1730, opened 2022-05-05
docs: convert INSTALL{,_es} to AsciiDoc

silopolis:
Merge after #1729 and pot and po files regenerated


docs/INSTALL.adoc
```diff
@@ -0,0 +1,127 @@
+:lang: en
+:toc:
+
+= Install LinuxCNC
```
hansu, line 4: The file name is still misleading...
```suggestion
= Build LinuxCNC for "run in place"
```

docs/INSTALL.adoc
```diff
@@ -0,0 +1,127 @@
...
+sudo make setuid                                  <6>
+----
+<1> Change to source directory
+<2> Builds the configure script (uses `configure.in`)
+<3> Configure the package for your system +
+    Running `configure` takes awhile. While running, it prints some
+    messages telling which features it is checking for.
+<4> Remove previous build(s) artifacts
+<5> Build/compile the software package
+<6> Set ownership and mode of built files
```
hansu, line 27: At first glance it looks as if you have to do the steps in the code block and the numbered items.
The same for section "Run LinuxCNC" and INSTALL_es as well.

```suggestion

[NOTE]
====
<1> Change to source directory
<2> Builds the configure script (uses `configure.in`)
<3> Configure the package for your system +
    Running `configure` takes awhile. While running, it prints some
    messages telling which features it is checking for.
<4> Remove previous build(s) artifacts
<5> Build/compile the software package
<6> Set ownership and mode of built files
====
```

silopolis: I don't think that callouts still work if embedded in an admonition block !?

docs/INSTALL_es.adoc
```diff
@@ -0,0 +1,133 @@
+:lang: es
+:toc:
+
+= Instrucciones de Instalación
```
hansu, line 4: Should be also "run in place" in spanish
@smoe Creación de LinuxCNC para "ejecutar en el lugar" ?

docs/INSTALL.adoc
silopolis: As agreed, leaving it as is for the moment...
I'll add a small descriptive paragraph later to avoid the dry code excerpt

docs/INSTALL_es.adoc
silopolis: > Should be also "run in place" in spanish @smoe Creación de LinuxCNC para "ejecutar en el lugar" ?

Done

hansu: Wasn't sure wheather to translate "build" or not. In German it would sound strange if we would translate it.
But doesn't matter at this point, a Spanish speaking person would understand that.

silopolis: Agree that "Run-In-Place" is kind of a proper name.
At least is it referenced everywhere, in any language, as RIP. Which is why, while translating it in "full text" I added the "RIP" mention. But it doesn't really make sense without expliciting it :/

smoe:
"in situ" comes to mind (https://en.wikipedia.org/wiki/In_situ), if that is what is meant, but then again, we are writing this documentation for earthly people not those bioscientists.

hansu:
> "in situ" comes to mind (https://en.wikipedia.org/wiki/In_situ), if that is what is meant, but then again, we are writing this documentation for earthly people not those bioscientists.

Hmm but that is latin

smoe:
I wasn't completely serious about this one. Your solution is just fine.

PR #1734, opened 2022-05-07
docs: remove index_XX.tmpl

hansu:
@petterreinholdtsen  These can also be removed right?


petterreinholdtsen:
[Hans]
> @petterreinholdtsen  this can also be removed right?

Yes, those are generated from index.tmpl using po4a.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
ok, thanks

PR #1735, opened 2022-05-07
MDRO initial version

rgbond:
Mdro is a DRO GUI meant for manual machines with DRO scales. It allows the user to drive the machine manually but still use the DRO in the same fashion as a traditional machinist's DRO.


silopolis:
Hi,

Like the idea.
Do you know TouchDRO ? Always found it inspiring

A documentation file to add to docs/src/gui/ with a screen capture or two
would be a nice addition.

hansu:
I don't really understand the usage of this.
How can a user manually control the machine? Only by hardware inputs configured in the appropriate HAL file?
And if I start the mdro with loadusr additionally to a running linuxcnc instance - why does the DRO shows always zeroes?

rgbond:
>  From: Hans ***@***.*** 
>  Sent: Sunday, May 08, 2022 2:46 AM
>  I don't really understand the usage of this.
>  How can a user manually control the machine? Only by hardware inputs configured in the appropriate HAL file?

HI Hans,
My mill is a Bridgeport clone. It came with handles on all of the axes. When I converted it to CNC I added servos, pulleys and belts
to each axis but I left the handles. I can still use the machine with the servos turned off if I enable estop and turn the handles by hand. I do
this fairly often for simple jobs. This program lets me use the DRO scales on each axis in that mode.

> And if I start the mdro with loadusr additionally to a running linuxcnc instance - why does the DRO shows always zeroes?

Did you hook the pins mdro.axis.n to the appropriate DRO inputs? I use it this way in my axis sim setup:

loadusr -W /home/rgb/bin/mdro.py -p 22 XYZ
net x-pos-fb  => mdro.0
net y-pos-fb  => mdro.1
net z-pos-fb  => mdro.2

Bob

hansu:
Thanks for your explanation.

>Did you hook the pins mdro.axis.n to the appropriate DRO inputs? I use it this way in my axis sim setup:

No I didn't know I had to do this. So a bit more comprehensive documentation would be nice like @silopolis mentioned.

rgbond:
I pushed a new version of the code. It has a new scheme to allow better options to specified in .ini files as well as the command line.  I updated the manual page, added some more comments as Hans suggested, and added an mdro.adoc along with the appropriate makefile changes.

hansu:
Thanks for the nice documentation!

Some suggestions to make it more intuitive
image

rgbond:
That looks nice. How did you encode the two non-ascii characters? I have to admit to knowing almost nothing about Unicode.

 

Bob

hansu:
>That looks nice. How did you encode the two non-ascii characters? I have to admit to knowing almost nothing about Unicode.

either just
                `('C','←','⏎'))`
                or
                `('C','\u2190','\u23CE'))`

rene-dev:
you dont need the from future import, and sys version check. in mdro.py

rgbond:
Thanks for the change. I pushed a new version of mdro.py with it and a new mdro.png screen capture. I would prefer keeping the import and the version check for now - I have systems here that are still running python 2... But I can pull it if it is a deal breaker.

hansu:
Looks good to me 👍. Maybe the buttons for the coordinate systems could be upper case.

Some other opinions about this PR? Shall we merge it?

andypugh:
Is "←" backspace / delete? If so, then I think that U232B - ⌫ is a little more different from the adjacent enter glyph?

hansu:
>Is "←" backspace / delete? 

Yes

>If so, then I think that U232B - ⌫ is a little more different from the adjacent enter glyph?

Yeah that's a more suitable one 👍.

rgbond:
I changed the labels on the coordinate systems to upper case as Hans suggested and updated the backspace key to the glyph that Andy suggested. Also pushed a new screen image for mdro.adoc.

rgbond:
OK, I think this is ready to merge.

hansu:
I am not mistaken, the documentation of the MDRO doesn't appear in the docs.
Question is: Where to include it? I think would fit best in the section 'User Interfaces'. Do you guys agree with that?

silopolis:
Le ven. 29 juil. 2022 à 15:37, Hans ***@***.***> a écrit :

> I am not mistaken, the documentation of the MDRO doesn't appear in the
> docs.
> Question is: Where to include it? I think would fit best in the section
> 'User Interfaces'. Do you guys agree with that?
>

👍

rgbond:
I've been trying to figure out how to do this... I think the goal is to get mdro into html/index.html, right? I haven't figured out where lines like these:
        <li><a href="gui/axis.html">Axis, Keyboard GUI</a></li>
come from.  I added a line to 
docs/src/Master_Documentation.adoc
but that didn't seem to change index.html. Some hints would be appreciated.
Bob

phillc54:
There are some hints on adding a doc [here](https://github.com/LinuxCNC/linuxcnc/blob/master/docs/README.adoc)

hansu:
It was more a question where you want me to add this but you can do this as well If you want.
Master_Documentation.adoc is for the PDF - it should be added there as well.
For the HTML only a links has do be added to the index.tmpl. The HTML file is already being generated: http://linuxcnc.org/docs/devel/html/gui/mdro.html

rgbond:
Thank you for the help! I misunderstood your email. It will go much faster if you change the docs. I think User Interface is the right section.

hansu:
Ok I added it. Should be available on the website in some time.

rgbond:
Did I miss a make file? I verified that the docs are generated… User Interfaces sounds right to me.

 

Bob

rgbond:
>From: Jérémie Tarot ***@***.*** 
>Sent: Sunday, May 08, 2022 1:27 AM
>Like the idea.
>Do you know TouchDRO ? Always found it inspiring

Hi Jérémie,
 I haven’t seen TouchDRO. I’ll look it up.

>A documentation file to add to docs/src/gui/ with a screen capture or two
>would be a nice addition.

I’ll add that. Thanks for the suggestion.
Bob

rgbond:
Hans, do you know how to tell if the Mdro command is in the distribution? I tried looking at the 2.9 package and didn’t see it, but I have no confidence was looking at the right place.

Thanks
Bob

hansu:
I guess you can have look if you have that folder

`/usr/lib/pythonX.XX/dist-packages/mdro`

Not 100% sure if it is supposed to be exactly there, but somewhere in the file system...

PR #1736, opened 2022-05-08
docs: gstat.adoc small indentation fix

silopolis:


hansu:
That indentation is not really needed but we can add it for better style.

silopolis:
Le dim. 8 mai 2022 à 21:32, Hans ***@***.***> a écrit :

> That indentation is not really needed but we can add it for better style.
>
True, but po4a reports them and I like it silent 😉
Plus, it's not bad for readability

PR #1740, opened 2022-05-15
Format fix

bkmgit:
Use ASCIIDOC format as specified at https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#images and https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#links


silopolis:
Thanks !

We're on the edge of a big merge on the docs front from the docs-devel branch, but I see that the first link is still broken there :/
Could you please retarget this to docs-devel branch so that @hansu can include it ?

bkmgit:
Thanks. Closing in favor of https://github.com/LinuxCNC/linuxcnc/pull/1741

PR #1741, opened 2022-05-15
Format fix

bkmgit:
Use ASCIIDOC format as specified at https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#images and https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#links

Suggestion made in https://github.com/LinuxCNC/linuxcnc/pull/1740

### DIFF
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -32,8 +32,9 @@ problem, all of the source files have been converted to .dxf files,
 and added to Git. The .dxf files can be edited with multiple CAD packages,
 and serve as the 'source' files for future changes to the drawings.
 They can be converted from .dxf to postscript (not .eps) using QCAD,
-which is available for many systems
-(https://repology.org/project/qcad/versions)[![Packaging status](https://repology.org/badge/tiny-repos/qcad.svg)].
+which is available 
+image:https://repology.org/badge/tiny-repos/qcad.svg[title="Badge of available systems"] 
+and also specified on https://repology.org/project/qcad/versions[this page].
 
 Unfortunately, EasyCAD (and AutoCAD) support a number of entities
 that QCAD does not import properly. Including some that were used
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -32,7 +32,7 @@ problem, all of the source files have been converted to .dxf files,
 and added to Git. The .dxf files can be edited with multiple CAD packages,
 and serve as the 'source' files for future changes to the drawings.
 They can be converted from .dxf to postscript (not .eps) using QCAD,
-which is available for many systems image:https://repology.org/badge/tiny-repos/qcad.svg[link=https://repology.org/project/qcad/].
+which is available for many systems, see link:https://repology.org/project/qcad/[here].
 
 Unfortunately, EasyCAD (and AutoCAD) support a number of entities
 that QCAD does not import properly. Including some that were used
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -32,9 +32,7 @@ problem, all of the source files have been converted to .dxf files,
 and added to Git. The .dxf files can be edited with multiple CAD packages,
 and serve as the 'source' files for future changes to the drawings.
 They can be converted from .dxf to postscript (not .eps) using QCAD,
-which is available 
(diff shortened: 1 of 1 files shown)


bkmgit:
Failing tests seem unrelated to this pull request, translations seem to have been removed/modified elsewhere.

docs/README.adoc
```diff
@@ -32,8 +32,9 @@ problem, all of the source files have been converted to .dxf files,
 and added to Git. The .dxf files can be edited with multiple CAD packages,
 and serve as the 'source' files for future changes to the drawings.
 They can be converted from .dxf to postscript (not .eps) using QCAD,
-which is available for many systems
-(https://repology.org/project/qcad/versions)[![Packaging status](https://repology.org/badge/tiny-repos/qcad.svg)].
+which is available 
+image:https://repology.org/badge/tiny-repos/qcad.svg[title="Badge of available systems"] 
+and also specified on https://repology.org/project/qcad/versions[this page].
```
hansu, line 37: I think it was more intended like this:
```suggestion
which is available for many systems image:https://repology.org/badge/tiny-repos/qcad.svg[link=https://repology.org/project/qcad/].
```

hansu:
@silopolis  @smoe @petterreinholdtsen An idea why this breaks our docs build?

smoe:
Is anything still building? The "make: command: command not found" I interpret to refer as a failure to execute something analogous to
```
$ command -v ls
/usr/bin/ls
```
My immediate thought was that the build server has changed somewhere somehow?!?

petterreinholdtsen:
[Hans]
> @silopolis  @smoe @petterreinholdtsen An idea why this breaks our docs
> build?

My guess would be that the .adoc-images-stamp target in
docs/src/Submakefile fail to properly process this image: link, but only
for the reflowed adoc editions generated by po4a.

A quick test indicate that the parsing seem to work:

  % echo 'image:https://repology.org/badge/tiny-repos/qcad.svg[link=https://repology.org/project/qcad/].' |grep -E ^image:[^[:space:]]  | sed -E "s/image:+([^[]+)\[/\nimage:\1\n/g" | grep image: | cut -d: -f2-
  https://repology.org/badge/tiny-repos/qcad.svg
  %

But I suspect trying to copy a non-existing image around on disk is
going to cause problems, and suspect it is better to use local copies of
images to be able to generate PDFs from all adoc files.

-- 
Happy hacking
Petter Reinholdtsen

docs/README.adoc
```diff
@@ -32,8 +32,7 @@ problem, all of the source files have been converted to .dxf files,
 and added to Git. The .dxf files can be edited with multiple CAD packages,
 and serve as the 'source' files for future changes to the drawings.
 They can be converted from .dxf to postscript (not .eps) using QCAD,
-which is available for many systems
-(https://repology.org/project/qcad/versions)[![Packaging status](https://repology.org/badge/tiny-repos/qcad.svg)].
+which is available for many systems image:https://repology.org/badge/tiny-repos/qcad.svg[link=https://repology.org/project/qcad/].
```
hansu, line 35: So a quick an ugly fix would be:
```suggestion
which is available for many systems, see link:https://repology.org/project/qcad/[here].
```

petterreinholdtsen:
[Hans]
> Sad that we can't embed images which are not present in the
> repository. Especially for this image it makes absolutely no sense to
> have this image static in the repo 😄

Personally I see it as an advantage that the documentation is self
contained and do not leak information (aka requests to fetch images)
externally.  It also make sure the documentation passes the 'deserted
island' test of free software, which is a good thing both for being
future proof and ensure it is also readable by those without Internet
connectivity.

As far as I can tell, this was the first try to introduce a non-local
image into the documentation.  It made me happy to note it was shut down
right away. :)

-- 
Happy hacking
Petter Reinholdtsen

hansu:
> Personally I see it as an advantage that the documentation is self contained and do not leak information (aka requests to fetch images) externally. It also make sure the documentation passes the 'deserted island' test of free software, which is a good thing both for being future proof and ensure it is also readable by those without Internet connectivity.

Interesting point of view which also makes sense 👍.

PR #1743, opened 2022-05-17
qt_vismach.py - Fix random error when this code return a float result

annhan:


c-morley:
Thank you. Nice catch

PR #1744, opened 2022-05-17
English man pages in _es subfolder

smoe:
The first commit removes what is simply redundant. The second commit moves English-worded .1 files from the _es subfolder to the English one where that original file is not existing. There is a chance that the respective man page is obsolete, but deletions are easy.


hansu:
Can you rebase your branch to the HEAD of docs-devel (for a better understanding)? Thanks!

smoe:
Redundant.

PR #1745, opened 2022-05-17
Set of changes for Docs+po4a for CI

smoe:
The CI for Debian package building was removed. Documentation only builds _en.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,6 @@ jobs:
         eatmydata ./autogen.sh
         eatmydata ./configure --with-realtime=uspace --disable-check-runtime-deps
         eatmydata make -O -j$((1+$(nproc))) default pycheck V=1
-        # Note that the package build covers html docs
         ../scripts/rip-environment runtests -p
     
   htmldocs:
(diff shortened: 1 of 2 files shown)


smoe:
@SebKuzminsky , I removed the Debian package building from the CI since this implies the complete documentation. Is that fine with you? At least as a temporary solution to shorten build times?

Admittedly I do not know how we will all be thinking about this when _fr and _es reach a 100% translation rate. The documentation itself is just more complete now, it did not just become any slower. Just with such a large fraction being directly passed from English to _fr and _es, it does not really make sense to wait for that. Still, even when this gets more reasonable since _fr and _es provide more content, we still do not want to always build this when we prepare an update to a PR.

smoe:
Hm. At least this was quick :o/ The links to HTML man pages are not found, which happen not to be built.

smoe:
Anybody with a clue?

petterreinholdtsen:
I had a look, and do not understand what this change is supposed to solve.  Can you explain in more detail?

smoe:
Hi Petter,

Thank you for having a look. It was meant to separate the _en docs (and dependencies) from the other languages' docs and then build only the _en ones, i.e. without po4a involved.

Also, the investment into the CI should be reduced. The Debian package for instance we may want to build less often.

Best,
Steffen

SebKuzminsky:
Hi Steffen, thanks for working on this problem.

I'm quite fond of building debs in CI, because that's the primary way we get updated executables to existing users, and finding out early (ie, before merging a PR) if that's about to break is a good time saver.  So I would rather accept long build times, and maintain the current deb-building validation.

I would rather look for ways to improve build time - for example, we seem to currently not be getting as much parallelism as we could.

smoe:
Hi @SebKuzminsky ,

Thank you for your comment. I see two/three kinds of motivations to reduce CI build time. 
1. the developer may be interested to learn if build problem is truly fixed.
2. quicker iteration between master and our developers - like when the authors of two forks are synchronizing their work.
3. there is a cap on the amount of CPU-time available per project, right?

Exploiting parallelism in our builds would help building the documentation immensely, which is about as intrinsically parallel as it gets. That would address 1+2.  But what about the billing. Isn't that of concern with an increasing number of languages?

What I would like to see is that the building of the Debian packaging does not build the documentation redundantly. An immediate fix could be to only build the architecture-dependent Debian packages. That would provide the .debs to our users and not build the docs again. The documentation could then best tested separately by only building the architecture-independent packages. IIRC then I already initiated that separation for the Debian package that is currently in the archive, not sure if those changes to debian/rules and debian/configure made it into the archive.

SebKuzminsky:
I agree, reducing build time (both in CI and "at home") is very valuable.  However that goal is in tension with the goal of having thorough validation of the quality of the repo.

Billing is not a big concern.  Github has never sent us a bill for CI time, and the buildbot is all dedicated donated hardware running at my house (and I don't bill the project for my electricity or my time).

Building the arch-indep debs (ie docs) "redundantly" (ie on all supported platforms) does feel somewhat like a waste of time, but it also provides valuable coverage - for example, it can detect problems with the build tools available on supported platforms, like the too-old po4a on Buster.
  
Maybe it would make sense to sort the translated languages into two categories: the ones we release to users (because they're complete enough) and the ones we don't release (because they're not complete enough).  The released languages would be built every time, and would get a `linuxcnc-doc-${LANG}` deb in `debian/control`, while the un-released languages would only get built sometimes (eg, maybe only one one platform), and would not get a deb in `debian/control`.

We could cause a build failure if a released language drops below the "complete enough to release" threshold.  And maybe another build failure if an un-released language climbs above the threshold, to spur us to promote that language to "released" and give it its own deb?

smoe:
Ok. This PR can then be closed, but I suggest to leave it open for another while to collect opions/ideas/needs/whatever from the community for another few "extended minutes".

smoe:
@SebKuzminsky did some recent work on the Debian packaging that we should address prior to this.

PR #1758, opened 2022-06-08
Enable translations for more manual pages

petterreinholdtsen:
Some manual pages uses roff macros and other features not handled by
po4a.  Make more manual pages translatable by removing the use of such
macros where it is trivial to do.


hansu:
What did these removed lines do?
```
.de TQ
.br
.ns
.TP \\$1
..
```

petterreinholdtsen:
[Hans]
> What did these removed lines do?
> ```
> .de TQ
> .br
> .ns
> .TP \\$1
> ..
> ```

They defined a .TQ macro to avoid having to write ".br\n.ns\n.TP".  Some
times unused, when used I expanded the macro and dropped the home made
.TQ statement.
-- 
Happy hacking
Petter Reinholdtsen

hansu:
I think you can just remove the macro declaration and leave the rest as it is because `.TQ` is (now) part of groff_man, maybe it wasn't 15 years ago when this inline-macro was inserted.

petterreinholdtsen:
[Hans ***@***.***> writes:

> I think you can just remove the macro declaration and leave the rest
> as it is because `.TQ` is (now) part of groff_man, maybe it wasn't 15
> years ago when this inline-macro was inserted.

If you can verify this, part of 186284aad7a4785a78349c589756308f7594bfb0
should be reverted too, as it also dropped the use of .TQ.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
>If you can verify this, part of https://github.com/LinuxCNC/linuxcnc/commit/186284aad7a4785a78349c589756308f7594bfb0
should be reverted too, as it also dropped the use of .TQ.

Yes I verified this on my Debian 10 system.
Furthermore, this macro was introduced 15 years ago in https://git.savannah.gnu.org/cgit/groff.git/commit/?id=259929625b21595951ed3ef5dba9aaaea359b464

To be sure that the macro an-ext.tmac is used, you can look in the man page of groff_man:

>       .TQ    Set an additional tag for a paragraph tagged with .TP.
>              The pending output line is broken.  The tag on the input
>              line following this macro and subsequent lines are handled
>              as with .TP.
>
>              This macro is a GNU extension not defined on systems
>              running AT&T, Plan 9, or Solaris troff; see an-ext.tmac in
>              section “Files” below.

hansu:
Is the image in e.g. `man9/toggle2nist.9` a problem for po4a?

petterreinholdtsen:
[Hans]
> Is the image in e.g. `man9/toggle2nist.9` a problem for po4a?

po4a object to the .PSPIC statement, yes.

Of course, 'man' do not handle it either. :)

See https://github.com/LinuxCNC/linuxcnc/pull/1314 for an idea how to
improve that situation.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[Hans]
> Are the current changes also needed to convert the man pages to
> asciidoc?

This set of patches is only to make more man page text available in
weblate.  I do not know if it affect conversion to asciidoc, but see no
reason why it should.  Conversion of images to plantuml and man pages to
asciidoc is better discussed in other issues and pull requests.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
Ok as it is some kind of temporarily fix, we can go for it (and it is compliant to https://github.com/LinuxCNC/linuxcnc/commit/186284aad7a4785a78349c589756308f7594bfb0).

PR #1759, opened 2022-06-10
Changes to documentation that came up while translating.

smoe:
I have translated a whopping 1% of the documentation to German and came across these issues.   @silopolis , I would love to get your comments, obviously, but most changes are not about style (except me prefering that table of G-codes) over something verbatim..

### DIFF
--- a/docs/help/tklinuxcnc.adoc
+++ b/docs/help/tklinuxcnc.adoc
@@ -78,7 +78,7 @@ machine.
   *Manual mode*;;
     Allows the operator to jog and home the axes, and control
     the coolant and spindle. +
-    Function key F3 puts the controller into manual mode.
+    Function key F3 puts the controller into Manual mode.
   *Auto mode*;;
     Allows the operator to load programs into the machine, run
     them, and pause/step/resume them. +
@@ -90,7 +90,7 @@ machine.
 *MIST ON/OFF, FLOOD ON/OFF*::
   Turns mist or flood coolant on or off. +
   Function key F7 toggles the state of the mist coolant. +
-  Function key F8 toggles the state of the coolant coolant. +
+  Function key F8 toggles the state of the flood coolant. +
   These buttons and function keys only work in Manual mode.
 
 *SPINDLE FORWARD/REVERSE/OFF*::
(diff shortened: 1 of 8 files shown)


SebKuzminsky (changes requested):
Thanks for these cleanups, I have two minor nitpicks.

smoe:
There were are few more "manual mode"s that I also changed to "Manual mode".

docs/help/tklinuxcnc.adoc
```diff
@@ -276,84 +276,86 @@ Manufacturing Sciences (NCMS) on the Next Generation Controller (NGC)
...
-G89                boring cycle, with dwell, feed out
-G90                non-incremental programming (the usual)
-G91                incremental programming
-G93                inverse time feed mode
-G94                feed per minute mode (the usual)
-G98,G99            set canned cycle return level
-----
+[cols="1,1"]
+|===
+|G00 X Y Z         | rapid feed |
```
hansu, line 281: Good idea to have this as table, but:
For all lines I think the last "|" is too much. At least according to the docs and my preview plug-in is also confused with this.
```suggestion
|G00 X Y Z         | rapid feed
```

And also for the following table.

hansu (changes requested):

silopolis:
Here they are.
Looks good to me 👌👍

docs/help/tklinuxcnc.adoc
smoe: Fixed - and also installed a previewer.

smoe:
You should not push --force when you have just accepted changes in the github web interface :o/
Thank you all!

smoe:
@petterreinholdtsen - these will eventually trigger a first iteration on changes that need to be run against weblate. Besides not wanting to grow my diff too much, I really wanted to see this circle closed. I have made changes to a table that may be difficult. I have not changed the "FIXME"s in the documentation that were introduced for the syncing between languages when the same text appeared in different contexts.

silopolis:
Le sam. 11 juin 2022 à 09:12, Hans ***@***.***> a écrit :

> Merged #1759 <https://github.com/LinuxCNC/linuxcnc/pull/1759> into master.
>

Drum roll!...

PR #1760, opened 2022-06-11
Add option to disable building translated docs

hansu:
This adds an option to the configure script to disable the building of the translated documentation:

`--disable-build-documentation-translation`

It's a bit long, but I want to have it similar to:

`--enable-build-documentation=html,pdf`

and wanted to have the option to add later something like:

`--enable-build-documentation-translation=es,fr`

Other ideas for the naming are welcome.

@petterreinholdtsen do you agree with the way it is realized?

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1136,12 +1136,36 @@ install with "sudo apt-get install w3c-linkchecker"])
     fi
 fi
 
-if ( test "$BUILD_DOCS" = "yes" ) ; then
+AC_ARG_ENABLE(build-documentation-translation,
+    AS_HELP_STRING(
+        [--disable-build-documentation-translation],
+        [Disable building documention in other languages than English.]
+    ),
+    [
+    case "$enableval" in
+	Y*|y*)
+	    BUILD_DOCS_TRANSLATED_EXPLICIT=yes ;;
+	no)
+	    BUILD_DOCS_TRANSLATED_EXPLICIT=no ;;
+	*)
+	    BUILD_DOCS_TRANSLATED_EXPLICIT=$enableval ;;
+	esac
+    ],
+    [
+        BUILD_DOCS_TRANSLATED_EXPLICIT=not_set
+    ])
+
+if ( test "$BUILD_DOCS" = "yes" && test "$BUILD_DOCS_TRANSLATED_EXPLICIT" != "no") ; then
     AC_PATH_PROG(PO4A, po4a, "none")
     BUILD_DOCS_TRANSLATED=yes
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
[Hans]
> @petterreinholdtsen do you agree with the way it is realized?

Look like it should work, but I did not test it.

It occur to me that using --enable-build-documentation-translation
explicitly should give an error if po4a is unavailable, while without
stating anything it should only be a warning.  Not sure if such feature
is worth the extra effort.  Perhaps making translations an option should
turn the po4a check into a fatal one?  It allow the failing po4a issue
to be solved with a configure option, which was not possible before and
the reason I made missing/incorrect po4a into a warning.  It could be
done by replacing AC_MSG_WARN with AC_MSG_ERROR.

You seem to have mixed in a independent typo fix in this commit.
Perhaps it should be separate pull requests?

-- 
Happy hacking
Petter Reinholdtsen

SebKuzminsky:
This is great, and I really like the direction you're thinking, to select what translation languages to produce via a configure argument.

hansu:
I tested it on a system with appropriate po4a and one without.

>It occur to me that using --enable-build-documentation-translation
explicitly should give an error if po4a is unavailable, while without
stating anything it should only be a warning. 

I changed that. Did you have this behaviour in mind?:

Following gives now an error with too old po4a:
```
./configure --with-realtime=uspace --enable-build-documentation=y --enable-build-documentation-translation
./configure --with-realtime=uspace --enable-build-documentation=y --enable-build-documentation-translation=es
```
following a warning:
```
./configure --with-realtime=uspace --enable-build-documentation=y 
```
following nothing:
```
./configure --with-realtime=uspace --enable-build-documentation=y --disable-build-documentation-translation
./configure --with-realtime=uspace                                --disable-build-documentation-translation
./configure --with-realtime=uspace                                --enable-build-documentation-translation=yes
```


>You seem to have mixed in a independent typo fix in this commit.
Perhaps it should be separate pull requests?

I didn't want do create an extra commit for changing a typo of only one character, but you are right, this would be better.

petterreinholdtsen:
[Hans]
> I changed that. Did you have this behaviour in mind?:

Yes.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
Ok. 
And `--enable-build-documentation-translation` is not too long?

petterreinholdtsen:
[Hans]
> Ok. 
> And `--enable-build-documentation-translation` is not too long?

I do not care. :)
-- 
Happy hacking
Petter Reinholdtsen

PR #1765, opened 2022-06-15
Update cpu_info.c for Raspberry revision 1.5

elovalvo:
Update cpu_info.c for Raspberry revision 1.5


petterreinholdtsen:
Perhaps its commit message should state how it is different from #1766?

hansu:
>Perhaps its commit message should state how it is different from https://github.com/LinuxCNC/linuxcnc/pull/1766?

Probably it's the same, one for 2.8 and one for master. But in this case we should only merge that one for 2.8 and merge 2.8 into master.

PR #1766, opened 2022-06-15
 Update cpu_info.c for Raspberry revision 1.5

elovalvo:
Update cpu_info.c for Raspberry revision 1.5


petterreinholdtsen:
Perhaps this should be dropped now that 1765 is merged into 2.8?

PR #1775, opened 2022-06-25
Switch to libtk-img to tkpng

jepler:
This was added with tk8.6 which is already a requirement per debian/configure.

This should resolve debian bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1012789


jepler:
Once this has been merged, can someone who's a debian uploader cut a new debian package, please and thank you?

jepler:
hmmm this may be silly.  png support is actually in the Tk core now (yay!) , since 8.6 which is in stretch (oldoldstable).

PR #1782, opened 2022-06-30
Update lathe_g70_71_demo.ngc

samcoinc:
Because of order of operations - the F word was set before the units change.  Putting Feed rate on the next line fixed it.


jepler (approved):
Thank you!

PR #1783, opened 2022-07-01
Fix soft limit calculation when using G43.1

d2inventory:
As discussed in this forum thread:
https://forum.linuxcnc.org/20-g-code/44244-g43-1-dynamic-tool-length-measurement-program-exceeds-machine-maximum-on-axis?start=10#227198

When using G43.1 Tool Length Offset, Axis will wrongly display a warning about exceeding the soft limits.
This PR adds the tool offset to the limits calculation so this warning wont be displayed anymore if not needed (It still will if it actually exceeds the limit).


c-morley:
Thank you

andypugh:
Should this be carried back to 2.8?

PR #1784, opened 2022-07-01
docs: More changes triggered while translating

smoe:


smoe:
This completes src/config - or at least those parts that I had felt half-way comfortable to attempt to find a translation for.

jepler (approved):
Thanks! I didn't review it all but it looks like good improvements of consistency.

PR #1786, opened 2022-07-02
Improve io(1) and iov2(1) manual pages

petterreinholdtsen:
Remove duplicate entries under "Additional IO v2 pins" in iov2
and add missing periods at the end of sentences in both io(1) and iov2(1)
to make sure the identical strings stay the same to save translators
duplicate work.


jepler (approved):
Thanks!

PR #1792, opened 2022-07-03
Make statement about 'Qtvcp only' easier to understand in ini-config.adoc

petterreinholdtsen:
For the casual reader, the original appear to refer to some unknown
acronym.  Making the statement a complete sentece make it easier to
understand.

### DIFF
--- a/docs/src/config/ini-config.adoc
+++ b/docs/src/config/ini-config.adoc
@@ -272,26 +272,26 @@ User Manual.
 * 'MIN_SPINDLE_0_OVERRIDE = 0.5' - The minimum spindle override the user may select.
   0.5 means 50% of the programmed spindle speed.  (This is used to
   set the minimum spindle speed).
-  On multi spindle machine there will be entries for each spindle number. QtVCP only.
+  On multi spindle machine there will be entries for each spindle number. Only used by the QtVCP based user interfaces.
 * 'MAX_SPINDLE_OVERRIDE = 1.0' - The maximum spindle override the user may
   select. 1.0 means 100% of the programmed spindle speed.
 * 'MAX_SPINDLE_0_OVERRIDE = 1.0' - The maximum feed override the user may select.
   1.2 means 120% of the programmed feed rate.
-  On multi spindle machine there will be entries for each spindle number. QtVCP only
+  On multi spindle machine there will be entries for each spindle number. Only used by the QtVCP based user interfaces.
 * 'DEFAULT_SPINDLE_SPEED = 100' - The default spindle RPM when the spindle
   is started in manual mode. if this setting is not present, this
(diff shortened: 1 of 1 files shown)


c-morley:
'based screens' would probably be better then using 'frontend'

petterreinholdtsen:
[c-morley]
> 'based screens' would probably be better then using 'frontend'

The setting for frontend/screen/display/interface is described like
this:

  * 'DISPLAY = axis' - The name of the user interface to use. Valid
    options may include: 'axis', 'touchy', 'gmoccapy', 'gscreen',
    'tklinuxcnc', 'qtvcp'

If not using frontend, I would believe it is better to switch to a term
already used for this setting, ie 'display' or 'user interface'.  But as
you find the use of frontend confusing, I have changed it to 'user
interface' to match other references to DISPLAY settings in the same
document.

I further changed it to drop the keyword part, and end up with "Only
used by the Qtvcp user interface."

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
user interface is good - yes consistency is a great idea.
Technically Qtvcp is not a screen, it's infrastructure for a screen. QtDragon, QtAxis, QtTouchy etc are the screens, that is why i suggested 'Qtvcp based' but maybe that's not important.

silopolis:
Le lun. 4 juil. 2022 à 14:41, c-morley ***@***.***> a écrit :

> user interface is good - yes consistency is a great idea.
> Technically Qtvcp is not a screen, it's infrastructure for a screen.
> QtDragon, QtAxis, QtTouchy etc are the screens,
>
I always found "screen" to be misleading as in my mind a VCP/UI can present
several ones to the user.
For me QtDragon, or Probe Basic in another land, are really VCPs...

that is why i suggested 'Qtvcp based' but maybe that's not important.
>
... And QtVCP or QtPyVCP are really frameworks, or toolkits to stick to GUI
dev lingo, and I believe making the distinction is important.

petterreinholdtsen:
I did not know that Qtvcp was a framework used by several user
interfaces, and have changed the proposed text to use the phrase 'Only
used by the Qtvcp based user interfaces.'

-- 
Happy hacking
Petter Reinholdtsen

silopolis:
Le lun. 4 juil. 2022 à 22:12, petterreinholdtsen ***@***.***>
a écrit :

>
> I did not know that Qtvcp was a framework used by several user
> interfaces, and have changed the proposed text to use the phrase 'Only
> used by the Qtvcp based user interfaces.'


👌
Maybe just QtVCP for consistent casing

petterreinholdtsen:
[Jérémie Tarot]
> Maybe just QtVCP for consistent casing

Good point.  Updated.

-- 
Happy hacking
Petter Reinholdtsen

PR #1800, opened 2022-07-05
Add link to the source of the Ziegler-Nichols method in the PID tuning chapter

petterreinholdtsen:


c-morley:
Has anybody used this method for linuxcnc? My understanding was it was not really suited for highly dynamic systems.

petterreinholdtsen:
[c-morley]
> Has anybody used this method for linuxcnc? My understanding was it was
> not really suited for highly dynamic systems.

I have no idea.  It is described in the text, so I thought it best to
refer to the source for those that want to check it out.  I assume it
would not be mentioned if it is unsuited for LinuxCNC.

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
I think it was mentioned because it was difficult (at the time) to find information on PID tuning.
The link is a good idea of course, I just think the method was not really useful - but I've not tried it.

PR #1814, opened 2022-07-09
Added po4a option for man pages to handle 18 more

petterreinholdtsen:


c-morley:
This is meaningless to me - do you want me to merge?

petterreinholdtsen:
[c-morley]
> This is meaningless to me - do you want me to merge?

Yes.

We had to exclude some man pages from the po4a translation system
because po4a were unable to parse them.  This make po4a handle more of
the manual pages, and thus make more man pages translatable.

-- 
Happy hacking
Petter Reinholdtsen

PR #1820, opened 2022-07-11
New component at_pid based on original at_pid and pid components.

petterreinholdtsen:
Started with a copy of pid.c, and copied relevant code from
at_pid.c and inserted it into relevant locations.  This way
at_pidv2 become compatible with the current pid component, and
those that want to test autotuning can insert it in-place
without modifying any pins and signals in their HAL files.

The original at_pid code was commited by Peter G. Vavaroutsos,
so insert his name as author and also copy the module name over.

Manual page at_pidv2(9) is based on pid(9) with the comment about
the auto-tuning code not being updated since 2011 added.


petterreinholdtsen:
Node, this component is not yet tested.  I believe it is equivalent to the at_pid module, only with the updates done to pid.c included.  If we decide to include it, it should replace at_pid.c, which will not serve any purpose once v2 is available.

PetteriAimonen:
For me the `at_pid` component never produced any useful results, even when I tried to simulate trivial response with `lowpass` filter. I wonder if it might have completely broken over the years somehow?

petterreinholdtsen:
[Petteri Aimonen]
> For me the `at_pid` component never produced any useful results, even
> when I tried to simulate trivial response with `lowpass` filter. I
> wonder if it might have completely broken over the years somehow?

I wonder if it ever worked, but I do not know pid tuning enough to tell
if the problem is my machine, my knowledge, the software or something
completely different.

But in the hope that at_pid do a better job than me, I wanted to see if
I could get it limping along.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
While trying to track down reports on the use of at_pid, I came across a few old mailing list post.  
http://emc.sourceforge.net/NIST-archive/msg00476.html contain useful references and reference to a tcl/tk tool I have not tracked down, and http://emc.sourceforge.net/NIST-archive/msg04655.html tells of using Zeigler-Nichols closed loop tuning with EMC.

Note, it would be perfectly possible to join pid.c and at_pidv2.c into one source file and enable the autotuning only with a compile time or run time option.  Is it a good idea?

c-morley:
auto tune as a runtime option sure sounds tempting...

petterreinholdtsen:
[c-morley]
 > auto tune as a runtime option sure sounds tempting...

 How is a decision made regarding this?

 The code is trivial to integrate into the existing pid component, and is
 not activated before the tune-start pin is set high, so it would be run
 time enabled by default.  I protected the autotuning code in the new
 at_pid.c with #ifdefs to make it easier to identify.

c-morley:
Ask your question on the dev mail list. If you don't get a good reason not too (or get nothing at all) go ahead.

petterreinholdtsen:
[c-morley]
> Ask your question on the dev mail list. If you don't get a good reason
> not too (or get nothing at all) go ahead.

OK.  As the dev mailing list subscription page (and in fact the entire
linuxcnc.org site) is blocking access from Tor users, and thus blocks me
from using it, I leave that part to someone else.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[c-morley]
> auto tune as a runtime option sure sounds tempting...

How is a decision made regarding this?

The code is trivial to integrate into the existing pid component, and is
not activated before the tune-start pin is set high, so it would be run
time enabled by default.  I protected the autotuning code in the new
at_pid.c with #ifdefs to make it easier to identify.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
Are you able to access the developer mailing list at the moment? Or is that still blocked?

PR #1822, opened 2022-07-11
Mention in classicladder(9) how to specify file name used.

petterreinholdtsen:


c-morley:
I think if you want to add this info, you should be more specific that classicladder (which loads the ladder program) is a different program then classicladder_rt (which sets up the realtime part)

petterreinholdtsen:
[c-morley]
> I think if you want to add this info, you should be more specific that
> classicladder (which loads the ladder program) is a different program
> then classicladder_rt (which sets up the realtime part)

Aha.  Then I guess I do not really understand what kind of information
is missing from the manual page, and suspect it is best to let someone
else phrase the missing details.  I just lifted the info from an
example, after trying to use the manual page to figure out where I would
find the ladder description.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
I tried extending the description a bit more.  Please let me know if I
got it right.
-- 
Happy hacking
Petter Reinholdtsen

hansu:
I think its good. Only the XXX should be replaced. And maybe we should create a man page for classicladder_rt as well and link to this?

petterreinholdtsen:
[Hans Unzner]
> I think its good. Only the XXX should be replaced. And maybe we should
> create a man page for classicladder_rt as well and link to this?

I would love to know what XXX should be replaced with. :)
-- 
Happy hacking
Petter Reinholdtsen

c-morley:
AFAIK (with a small bit of looking too) there is no default filename.
loading the classiladder gui without a filename gives you a blank namless ladder program.

Also is 'non-realtime' seem ok? Something like 'usertime' gives more of a clue that it must be loaded with loadusr no?

hansu:
> Also is 'non-realtime' seem ok? Something like 'usertime' gives more of a clue that it must be loaded with loadusr no?

I think 'userspace' is the word which is used for that.

docs/man/man9/classicladder.9
```diff
@@ -6,7 +6,18 @@ classicladder \- realtime software plc based on ladder logic
...
+
 .SH DESCRIPTION
+This component consist of a realtime program and a non-realtime
+program.  The non-real time part load the programmable ladder
+description, while the real time part provide the pins.  The default
+file name is XXX and this can be changed using an argument to the
+non-realtime part when it is loaded with loadusr.  By default the
+non-realtime part provide a graphical visualisation of the load
+ladder, wihch can be disabled using the --nogui option to the
+non-realtime part.
```
hansu, line 19:
```suggestion
This component consist of a realtime program and a userspace
program.  The userspace part loads the programmable ladder
description, while the real time part provides the pins.  The file 
name of the configuration can be changed using an argument to the 
userspace part.  By default the userspace part provides a graphical 
visualisation of the loaded ladder, which can be disabled using the 
'--nogui' option to the userspace part.
```

c-morley:
yes of course you are right- userspace is better and more common.

After a bit of testing -
loadusr classicladder will display what ever program had been previously loaded 

so doing this one after another:
loadrt classicladder_rt -loads realtime code
loadusr classicladder - displays blank ladder
loadusr classicladder program.cpl - loads specific ladder program
loadusr classicladder -displays previous ' program.cpl' ladder program

petterreinholdtsen:
[Hans Unzner]
> I think 'userspace' is the word which is used for that.

According to the meeting 2022-08-08, the 'user space' term is no longer
reflecting realities, so I went with non-realtime.

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
I think whether it's technically correct or not is unimportant, it signifies there is a difference and gives a clue what that difference is.

petterreinholdtsen:
I updated the text but kept non-realtime to reflect the views expressed
during the last video chat on userspace vs. non-realtime.

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
it's a terrible term.

Are you sure they weren't talking of the type of realtime?  'hard real time' vrs 'userspace realtime' ? 
There is still a difference between real time and user space.

petterreinholdtsen:
[c-morley]
> Are you sure they weren't talking of the type of realtime?  'hard real
> time' vrs 'userspace realtime' ?  There is still a difference between
> real time and user space.

The way I understand it is that with the recent setup, both real-time
and non-realtime code is running in userspace, so 'userspace' do not
really express a relevant difference any more.

As far as I understand it, there are two classes of real time code in
linuxcnc, one with strict time requirements and another with softer time
requirements, as well as the non-realtime code.  In earlier times the
real time code was a kernel module.  With the uspace approach it is all
in userspace.

Feel free to provide a better text for the classicladder(9) manual page.
My initial goal was to mention how to specify the file name used.  The
rest of the changes are guesswork triggered by your comments and
suggestions.

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
You are correct about the realrime code. there used to be kernel space real time (RTAI) and user space realtime
( PREEMPT_RT) and there were other options and shims etc
This is all in the context of realtime
Then there is userspace - regular old programs that are per-empted by the real time code.

HAL loads each differently 'loadrt' for realtime code (userpace based or kernel space based) and 'loadusr' for regular programs.

So there is still 'realtime' and 'userspace' programs. and even if there wasn't, HAL requires that they be loaded differently.

So to give a clue, it make sense to use terms that reflect this.

We were working on this and you flatly refused - so sorry if 'terrible term' was too pointed, I didn't mean it to infer you are doing terrible work or anything like that. It's just you can pick a better word - like userspace.

petterreinholdtsen:
[c-morley]
> So there is still 'realtime' and 'userspace' programs. and even if
> there wasn't, HAL requires that they be loaded differently.

The term 'userspace' is a well known term in operating system lingo, and
it is opposed to 'kernel space' which is the code running with
privileges inside the kernel.  Using 'userspace' as if it is opposed to
'realtime' is thus now misleading, as the 'realtime' code in linuxcnc is
running in 'userspace'.

I understand that 'realtime' used to also be 'kernel space' in linuxcnc,
but given that it is no longer (always) the case, I believe it is
misleading the reader to keep using terms as if 'realtime' is not also
'userspace'.

The cause is simply that I find the arguments from Sebastian and Hans
presented at the meeting at the beginning of this month more convincing
than your accurate description on how the terms have been used
historically in the linuxcnc proejct.  I understand what you are saying,
I just do not believe it is accurate any more.

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
Wow ya I guess we will not agree on this.

Then if you want to be really correct should you not use the term 'userspace realtime' and 'userspace'
Userspace realtime is not the same as userspace - ask Seb if you don't believe me.
And if I am wrong then we can change HALcmd to only use 'load'

I think user docs should lead people to the right answer as quickly and easily as possible and hopefully one can start to infer information from one part to another.
If you have to be a realtime expert to know the difference then the difference doesn't matter in this context.
We are not making these docs for people developing the realtime system.

My point was it leads people to understand that loadusr is for userpace programs while loadrt would be for the realtime program - which would also be useful to know in the whole project.

Anyways I have beaten the horse enough and thank you for explaining your position.

hansu:
We can have something like this:

"The non-realtime part (often called 'userspace' in the LinuxCNC documentation) ..."

So we would have the correctness and the connection loadusr etc. :wink:

petterreinholdtsen:
[Hans Unzner]
> We can have something like this:
>
> "The non-realtime part (often called 'userspace' in the LinuxCNC
> documentation) ..."
>
> So we would have the correctness and the connection loadusr
> etc. :wink:

Good idea.  Updated patch.

-- 
Happy hacking
Petter Reinholdtsen

PR #1825, opened 2022-07-11
Correct build rules for threads component.

petterreinholdtsen:
Seem to have a cut-n-paste error, when compared to all other
components.


SebKuzminsky:
Good catch!

PR #1830, opened 2022-07-12
Gmoccapy: Pass numbers to user tabs to allow numeric entry

andypugh:
Especially for the lathe macros.

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -3224,6 +3224,13 @@ class gmoccapy(object):
         # tooledit page is active, so keys must go through
         if self.widgets.ntb_preview.get_current_page() == 2:
             return
+            
+        if (self.widgets.ntb_preview.get_current_page() == 4 and
+            keyname in ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
+            'comma', 'period', 'BackSpace', 'Return']):
+            #user Tab, pass numbers through
+            return False
+            
 
         # take care of different key handling for lathe operation
         if self.lathe_mode:


andypugh:
I am not sure how to safely detect if the user tab is one that should get key events. This commit works for the lathemacro sim config, though, which was my main aim.

c-morley:
I wonder if it's a problem that would be best set in the INI, when specifying the loading of a panel?
Could add a optional line ALLOW_KEYPRESSES = True or specify which ones get through?

hansu:
I think it is fine in general. There are no other key actions that can be affected as far as i can see.
Sure, the safest would be to check if the focus is in one of the entry boxes, but that might be a lot of code for a small thing. 

I only noticed two minor things:

- When in manual mode (sure, it makes not much sense, but it can be started from that), the numbers of the keypad are not 'forwarded' to the user tab.
- Sure that you need the enter key there? When pressing it twice, it moves to the MDI entry when in MDI mode.

src/emc/usr_intf/gmoccapy/gmoccapy.py
```diff
@@ -3224,6 +3224,13 @@ def on_key_event(self, widget, event, signal):
         # tooledit page is active, so keys must go through
         if self.widgets.ntb_preview.get_current_page() == 2:
             return
+            
+        if (self.widgets.ntb_preview.get_current_page() == 4 and
+            keyname in ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
+            'comma', 'period', 'BackSpace', 'Return']):
+            #user Tab, pass numbers through
+            return False
+
```
hansu, line 3233: Code style only ;-)
```suggestion
            
        # user Tab, pass numbers through
        if self.widgets.ntb_preview.get_current_page() == 4 and
            keyname in ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
            'comma', 'period', 'BackSpace', 'Return']:
            return False
```

smoe: Should 'comma', 'period', 'BackSpace', 'Return' not possibly be a constant of some class such that typos/API changes would be found at compile time? Google could not find it for me. I like the code style change.

hansu: These constants are defined and derived from https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/gdkkeysyms.h

>The names are the same as those in the gdk/gdkkeysyms.h header file but without the leading “GDK_KEY_”.

\[https://lazka.github.io/pgi-docs/#Gdk-3.0/functions.html#Gdk.keyval_name ]

smoe: Thank you for that pointer. Stared at it for a while but nothing surfaced that I could suggest to make this any better. Also checked https://lazka.github.io/pgi-docs/#Gtk-3.0/enums.html and https://lazka.github.io/pgi-docs/#Gtk-3.constants.html .

andypugh:
It would be better to add a property to each tab (or have a separate array) based on INI values as suggested by Chris. 
Looking at the INI every keypess feels like too much overhead, but I doubt that was the suggestion. 
I think we should probably put this in now, to get the demo config working, and then try to think up a better plan later. This is a bit of an uncommon use-case.

PR #1831, opened 2022-07-13
Introduce 'dist' make target to generate tarball ready to build

petterreinholdtsen:
With this tarball, it should be possible to build and install
the source using "cd src; ./configure ; make; make install".


petterreinholdtsen:
Using the changes in #1835, the same tarball can be used across all the build variations.

PR #1836, opened 2022-07-15
Corrected debian/copyright CC license

petterreinholdtsen:
Correctly indented a line that was missing it. Also renamed it to
commonly used acronym CC-BY instead of just CC, and removed some
strange text fragment that was appended to the license text.  See
https://creativecommons.org/licenses/by/3.0/legalcode for the full
text where this fragment is not present.

### DIFF
--- a/debian/copyright
+++ b/debian/copyright
@@ -408,7 +408,7 @@ License: LGPL-2.1+
 
 Files: share/qtvcp/images/widgets/writer/icons
 Copyright: 2013 Yusuke Kamiyamane
-License: CC-3.0
+License: CC-BY-3.0
 
 Files: unit_tests/catch.hpp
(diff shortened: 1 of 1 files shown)


jepler (approved):
Thanks! Is there a tool to check this file that we should run during CI?

petterreinholdtsen:
[Jeff Epler]
> Thanks! Is there a tool to check this file that we should run during
> CI?

This issue started with an error reported by lintian, which is run by
the CI system already.  So it is more a question of someone looking at
the issues, not the lack of checking.

You can also check out
<URL: https://wiki.debian.org/CopyrightReviewTools >  for more extensive
options.

-- 
Happy hacking
Petter Reinholdtsen

PR #1837, opened 2022-07-15
Switch github CI jobs to use po4a version 0.67.

petterreinholdtsen:


jepler:
I'm not against it, but what's the motivation for using a super-new version of po4a?

silopolis:
Le sam. 16 juil. 2022 à 14:25, Jeff Epler ***@***.***> a
écrit :

> I'm not against it, but what's the motivation for using a super-new
> version of po4a?
>

Mainly the fact that our extended and intensive usage has risen several
issues that have been fixed by the good folks of the po4a team... Including
a nasty one about linebreaks that almost garbaged our work on po4a
migration.

0.67 will very soon be available in testing so this won't be a real issue
for long

jepler:
How will it affect building documentation with releases from linuxcnc.org that target "older" releases such as debian stable?

petterreinholdtsen:
[Jeff Epler]
> How will it affect building documentation with releases from
> linuxcnc.org that target "older" releases such as debian stable?

Not much, as the version in stable is already too old to use with
asciidoc.

-- 
Happy hacking
Petter Reinholdtsen

SebKuzminsky:
I built po4a 0.66 for buster, it's in the deb archive on wlo and it's being used to build the docs in the buildbot.  Backporting build-dependencies like this is not new to us, but it's also something we generally try to avoid, because of all the chaos it introduces.  In this case i think the benefit of working translation infrastructure makes it worth it.

jepler:
@SebKuzminsky that's good but it looks like now a higher version is needed again

petterreinholdtsen:
[Jeff Epler]
> @SebKuzminsky that's good but it looks like now a higher version is
> needed again

This time it is to fix
<URL: https://github.com/mquinson/po4a/pull/199 >,
<URL: https://github.com/mquinson/po4a/issues/339 >,
<URL: https://github.com/mquinson/po4a/issues/361 >,
<URL: bugs.debian.org/710678 >, and a few other issues reported directly
via IRC to the po4a developer.  All of these discovered while working on
the LinuxCNC documentation and fixed on our request.  We might discover
more issues as we work our way through the translations, of course, but
it seem to be stabilizing now.

We can of course instead rewrite the asciidoc and roff files in LinuxCNC
instad, and drop all use of " +\n" for hard linebreak and roff macros
like .de and .if, but I doubt it is the way we want to move.

-- 
Happy hacking
Petter Reinholdtsen

silopolis:
Le lun. 18 juil. 2022 à 06:58, petterreinholdtsen ***@***.***>
a écrit :

> We can of course instead rewrite the asciidoc and roff files in LinuxCNC
> instad, and drop all use of " +\n" for hard linebreak and roff macros
> like .de and .if, but I doubt it is the way we want to move.



This is surely NOT!
I'd personally take it quite badly...

So much for a package coming to testing in a matter of days and people
wanting to run a tesla engine on a post war chassis...

jepler:
@SebKuzminsky will you be able to update po4a on the buildbot a second time? Otherwise, it sounds like our generated docs will be badly broken.

SebKuzminsky:
I'll update the deb archive at wlo and the buildbot mid next week some time.
Thanks @petterreinholdtsen and @silopolis for all your work on our translations and translation infrastructure, and for reporting the issues you find to po4a upstream!

silopolis:
Le ven. 29 juil. 2022 à 16:35, Sebastian Kuzminsky ***@***.***>
a écrit :

> I'll update the deb archive at wlo and the buildbot mid next week some
> time.
>

👍

Thanks @petterreinholdtsen <https://github.com/petterreinholdtsen> and
> @silopolis <https://github.com/silopolis> for all your work on our
> translations and translation infrastructure, and for reporting the issues
> you find to po4a upstream!
>

Pleasure to be part of this with you guys 🙏

silopolis:
Le sam. 16 juil. 2022 à 15:58, Jeff Epler ***@***.***> a
écrit :

> How will it affect building documentation with releases from linuxcnc.org
> that target "older" releases such as debian stable?
>

This will imply either backporting po4a in Debian, or just download and
install po4a package from testing directly on stable as it runs perfectly
like this.

A download link can be provided as a hint at dependency checking time.

Note that this is already the case as we need 0.66+ since po4a migration,
which is only available in testing and sid.

Anyway, just another reason to consider master targeted at testing and
future stable...

PR #1839, opened 2022-07-15
WIP: Only change spindle speed info and flush queue when speed changes

### LINKED ISSUE #1813: Don't flush queue when spindle speed is set but stays same

 1. Generate G-code using [OpenBuilds CAM](https://cam.openbuilds.com/)
 2. Set `G64` path blending mode
 3. Run the file and observe slow motion

The generated G-code will be like this:

    G1 F1000 X100.000 Y100.000 Z-2.000 S10000
    G1 F1000 X100.100 Y100.010 Z-2.000 S10000
    G1 F1000 X100.200 Y100.020 Z-2.000 S10000
    G1 F1000 X100.300 Y100.040 Z-2.000 S10000
    G1 F1000 X100.400 Y100.060 Z-2.000 S10000

Segments should be blended together instead of stopping many times per millimeter.

Machine feeds very slowly and `linuxcnctop` shows that queue length is always 0 or 1.

LinuxCNC 2.8.2.11.g6a3d0a434 using the official distribution, with AXIS gui.

## Cause of bug

This happens because OpenBuilds CAM puts `S10000` on every line, which rs274 then converts to `SET_SPINDLE_SPEED` command.  

Currently `SET_SPINDLE_SPEED` unconditionally calls `flush_segments()`:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/task/emccanon.cc#L1918-L1926

I suggest adding a comparison against currently set speed first, and skip the function if there is no change.
I can pull request the change if wanted.

petterreinholdtsen:
At the moment repeated G1 with identical S# setting will cause the
spindle movement to come to a halt between each G1.  This only make
sense when the spindle speed change, so add a test to avoid flushing
the work queue if the spindle speed do not change.

Also fixes incorrect indentation in the function.

Fixes #1813.

### DIFF
--- a/src/emc/task/emccanon.cc
+++ b/src/emc/task/emccanon.cc
@@ -1917,27 +1917,36 @@ void START_SPINDLE_COUNTERCLOCKWISE(int s, int wait_for_atspeed)
 
 void SET_SPINDLE_SPEED(int s, double r)
 {
-    // speed is in RPMs everywhere
+    // speed r is in RPMs everywhere
 
-	canon.spindle[s].speed = fabs(r); // interp will never send negative anyway ...
+    // Only update spindle speed and flush segments if the speed
+    // changed, to allow identical speed to be set on consecutive
+    // blocks (issue #1813).
+    // FIXME Figure out how this interact with the calls triggered by css_maximum
+    double newspeed = fabs(r); // interp will never send negative anyway ...
+    if(newspeed == canon.spindle[s].speed)
+        return;
 
     EMC_SPINDLE_SPEED emc_spindle_speed_msg;
 
     flush_segments();
 
+    canon.spindle[s].speed = newspeed;
     emc_spindle_speed_msg.spindle = s;
     if(canon.spindle[s].css_maximum) {
-		if(canon.lengthUnits == CANON_UNITS_INCHES){
-			canon.spindle[s].css_factor = 12 / (2 * M_PI) * canon.spindle[s].speed * TO_EXT_LEN(25.4);
-		} else {
-			canon.spindle[s].css_factor = 1000 / (2 * M_PI) * canon.spindle[s].speed * TO_EXT_LEN(1);
-		}
-		emc_spindle_speed_msg.speed =  canon.spindle[s].dir * canon.spindle[s].css_maximum;
-		emc_spindle_speed_msg.factor =  canon.spindle[s].dir * canon.spindle[s].css_factor;
-		emc_spindle_speed_msg.xoffset = TO_EXT_LEN(canon.g5xOffset.x + canon.g92Offset.x + canon.toolOffset.tran.x);
+	if(canon.lengthUnits == CANON_UNITS_INCHES){
+	    canon.spindle[s].css_factor = 12 / (2 * M_PI) * canon.spindle[s].speed * TO_EXT_LEN(25.4);
 	} else {
-        emc_spindle_speed_msg.speed = canon.spindle[s].dir * canon.spindle[s].speed;
-		//   canon.css_numerator = 0; FIXME: Do we need this?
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
Note, this code is not tested.  I notice the SET_SPINDLE_SPEED() method is called several times after a test if css_maximum is set, but it is unclear to me why.  Perhaps it need to do something also when speed is unchanged?

jepler:
Two things come to mind while thinking briefly about this PR.

First, as to constant surface speed: imagine issuing "S100" then later "T1 M6 G43" and "S100". Because the tool X offset has changed, the CSS-related values need to be updated. However, I think after this PR they would not be.

Second, this relies on the internal idea of spindle speed matching the external idea. Testing should make sure that e.g., if a program ends with M2, or is aborted, or etc., while S100 is in effect according to the interpreter's internal state, the next "S100" is actually issued. Similarly, what if S100 is issued via MDI, then the spindle stopped via manual control, then "S100" is issued again...

I'm not against it but it needs to be merged only after thoughtful testing.

petterreinholdtsen:
[Jeff Epler]
> I'm not against it but it needs to be merged only after thoughtful
> testing.

Perhaps you can help come up with some test code we can place in tests/
to validate correct behaviour, to ensure this patch do not introduce new
issues.

I do not know the linuxcnc code well enough to make a qualified guess
about the effect of this patch, just wanted to see if I could help with
what at the surface appeared to be adding a simple if-test. :)

Perhaps I should create a separate patch with the indentation
corrections, while we work together on a better patch?

-- 
Happy hacking
Petter Reinholdtsen

jepler:
Working on an automated test is beyond my time capacity for now, sorry.

PetteriAimonen:
One approach would be to make `drop_segments()` (and thus also `flush_segments()`) reset some variable that is used only for holding the previously set spindle speeds. This way any queue flushing action such as abort or tool change will ensure next spindle command is issued in full.

Reading the code, it appears that the concerns of @jepler are valid for the current pull request. The tool changes will automatically call `SET_SPINDLE_SPEED` if CSS is active, but the call would get dropped.

petterreinholdtsen:
[Petteri Aimonen]
> Reading the code, it appears that the concerns of @jepler are valid
> for the current pull request.

I just made a naive approach to get the discusion started, feel free
create another and better pull request, and I can close this one.

--
Happy hacking
Petter Reinholdtsen

PetteriAimonen:
@petterreinholdtsen What you made is exactly what I would have made also :) So thanks, it certainly brought up the nuances that I didn't know about.

I'm also not that familiar with LinuxCNC internals yet so hard to say what approach would make most sense.

SebKuzminsky:
@petterreinholdtsen, any update on this one?

andypugh:
I think that perhaps the _interpreter_ could choose to ignore repeated identical S-words, simply not queueing redundant commands. 

However LinuxCNC docs state that after any S-command the system will wait for spindle-at-speed. And that includes being prepared to stop if it is not. 

In this case the G-code is somewhat eccentric, but this is exactly how you would code things if you _wanted_ to wait for spindle-at-speed on some sort of special purpose machine where that was not guaranteed. 

So, I think in this case LinuxCNC is doing exactly what the G-code is asking it to do, in the context of the documented behaviour.

PetteriAimonen:
@andypugh That explanation makes sense. Ignoring repeated spindle speed setting could perhaps be added as a `[RS274NGC]` ini file option?

andypugh:
I am going to close this pull request for now. If you want to re-submit with the same behaviour as an INI-file option then we can look at it again (And it would probably then be acceptable) 

Though really the fix belongs in whatever broken CAM package is doing this...

PR #1840, opened 2022-07-15
Get tests/motion/g0/test.sh working again

petterreinholdtsen:
Correct program name from emc to linuxcnc, axis names, python code and comments for python 3.

Based in input from Jeff Epler.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,5 +87,5 @@ jobs:
         eatmydata debian/configure
         eatmydata debuild -us -uc
         sudo apt-get install ../*.deb
-        eatmydata ./scripts/runtests -p tests/
+        ./scripts/runtests -p tests/
         eatmydata lintian --info --display-info --pedantic --display-experimental ../*.deb
--- a/scripts/runtests.in
+++ b/scripts/runtests.in
@@ -141,6 +141,7 @@ run_tests () {
 	    fi
 	fi
 	NUM=$(($NUM+1))
+	export TEST_DIR=$(readlink -f $testdir)
 	echo "Running test: $testdir" 1>&2
         if test -n "$SYSTEM_BUILD"; then
             # Tell `halcompile` where to install comps
--- a/tests/motion/g0/checkresult
+++ b/tests/motion/g0/checkresult
@@ -18,7 +18,7 @@ servo_period_ns = float(Popen(
         "-sec",
         "EMCMOT",
         "-ini",
-        "%s/tests/motion/g0/motion-test.ini" % os.getenv("EMC2_HOME")
+        "%s/motion-test.ini" % os.getenv("TEST_DIR")
     ],
     stdout=PIPE
 ).communicate()[0])
@@ -32,9 +32,9 @@ max_acceleration_ips2 = float(Popen(
         "-var",
         "MAX_ACCELERATION",
         "-sec",
-        "AXIS_0",
+        "AXIS_X",
         "-ini",
-        "%s/tests/motion/g0/motion-test.ini" % os.getenv("EMC2_HOME")
+        "%s/motion-test.ini" % os.getenv("TEST_DIR")
     ],
     stdout=PIPE
 ).communicate()[0])
@@ -50,9 +50,9 @@ max_velocity_ips = float(Popen(
         "-var",
         "MAX_VELOCITY",
         "-sec",
-        "AXIS_0",
+        "AXIS_X",
         "-ini",
-        "%s/tests/motion/g0/motion-test.ini" % os.getenv("EMC2_HOME")
+        "%s/motion-test.ini" % os.getenv("TEST_DIR")
     ],
     stdout=PIPE
 ).communicate()[0])
@@ -63,7 +63,9 @@ max_velocity_ipc = max_velocity_ips / cycles_per_second
 # read the samples file and turn it into an array, where index i is an
 # array of the floats found on line i of the samples file
 samples_filename = sys.argv[1] + ".halsamples"
-lines = file(samples_filename).readlines()
+with open(samples_filename) as f:
+    lines = f.readlines()
+
 samples = []
 for line in lines:
     s = [ float(x) for x in line.split() ]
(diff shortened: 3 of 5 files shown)


jepler:
This test has been disabled for a loooonnnnggg time 
```
diff --git a/tests/motion/g0/checkresult b/tests/motion/g0/checkresult
index 8136da7fcc..01744f761a 100755
--- a/tests/motion/g0/checkresult
+++ b/tests/motion/g0/checkresult
@@ -32,7 +32,7 @@ max_acceleration_ips2 = float(Popen(
         "-var",
         "MAX_ACCELERATION",
         "-sec",
-        "AXIS_0",
+        "AXIS_X",
         "-ini",
         "%s/tests/motion/g0/motion-test.ini" % os.getenv("EMC2_HOME")
     ],
@@ -50,7 +50,7 @@ max_velocity_ips = float(Popen(
         "-var",
         "MAX_VELOCITY",
         "-sec",
-        "AXIS_0",
+        "AXIS_X",
         "-ini",
         "%s/tests/motion/g0/motion-test.ini" % os.getenv("EMC2_HOME")
     ],
@@ -63,7 +63,9 @@ max_velocity_ipc = max_velocity_ips / cycles_per_second
 # read the samples file and turn it into an array, where index i is an
 # array of the floats found on line i of the samples file
 samples_filename = sys.argv[1] + ".halsamples"
-lines = file(samples_filename).readlines()
+with open(samples_filename) as f:
+    lines = f.readlines()
+
 samples = []
 for line in lines:
     s = [ float(x) for x in line.split() ]
diff --git a/tests/motion/g0/skip b/tests/motion/g0/skip
deleted file mode 100755
index ecdbef95dd..0000000000
--- a/tests/motion/g0/skip
+++ /dev/null
@@ -1,2 +0,0 @@
-#!/bin/sh
-exit 1
diff --git a/tests/motion/g0/test.sh b/tests/motion/g0/test.sh
index 95cbd1aa5c..06d34c69d7 100755
--- a/tests/motion/g0/test.sh
+++ b/tests/motion/g0/test.sh
@@ -3,9 +3,9 @@
 #export PATH=$PATH:$EMC2_HOME/tests/helpers
 #source $EMC2_HOME/tests/helpers/test-functions.sh
 
-emc motion-test.ini &
+linuxcnc motion-test.ini &
 
-# let emc come up
+# let linuxcnc come up
 sleep 4 
 
 (
@@ -26,20 +26,20 @@ sleep 4
     echo set home 1
     echo set home 2
 
-    # give emc a second to home
+    # give linuxcnc a second to home
     sleep 1.0
 
     echo set mode mdi
     echo set mdi g0x1
 
-    # give emc a half second to move
+    # give linuxcnc a half second to move
     sleep 0.5
 
     echo shutdown
 ) | nc localhost 5007
 
 
-# wait for emc to finish
+# wait for linuxcnc to finish
 wait
 
 exit 0
```
these additional changes make the test pass locally. However, the very old commit message that added the skip file remark that it failed on the buildbot.  This was before 2.5.0 was released.

jepler:
hmmm .. is the EMC2_HOME (??) variable not set when running tests from the packaging process?
```
2022-07-16T09:18:06.9301637Z Traceback (most recent call last):
2022-07-16T09:18:06.9302347Z   File "tests/motion/g0/checkresult", line 13, in <module>
2022-07-16T09:18:06.9304005Z     servo_period_ns = float(Popen(
2022-07-16T09:18:06.9304718Z ValueError: could not convert string to float: ''
2022-07-16T09:18:06.9553604Z *** tests/motion/g0: XFAIL: checkresult exited with 1
```

Perhaps the script should be changed to find the inifile as relative to the python special variable `__file__` (which would be the checkresult script itself).

EMC2_HOME: Legacy environment variable, still set by rip-environment so not available when packaged. This is what this script depends on
LINUXCNC_HOME: Current environment variable, but only set during the linuxcnc wrapper script execution so I think it wouldn't work while running the checkresult script

petterreinholdtsen:
[Jeff Epler]
> hmmm .. is the EMC2_HOME (??) variable not set when running tests from
> the packaging process?

scripts/runtest.in set EMC2_HOME when running tests, also for
non-run-in-place setups.

> Perhaps the script should be changed to find the inifile as relative
> to the python special variable `__file__` (which would be the
> checkresult script itself).

As the error come from trying to run inivar, perhaps the problem is that
the built package is not installed, and thus inivar is not available in
PATH?

The test worked when I ran it locally, so it must be something with the
CI build environment.
--
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[Petter Reinholdtsen]
> scripts/runtest.in set EMC2_HOME when running tests, also for
> non-run-in-place setups.

The problem is that it is set to /usr/.

> As the error come from trying to run inivar, perhaps the problem is
> that the built package is not installed, and thus inivar is not
> available in PATH?

Nver mind, I was able to get it working and pushed a fix, sadly
involving a fix for runtests.in too.

> The test worked when I ran it locally, so it must be something with
> the CI build environment.

I must thave remembered incorrect, as this no longer worked when I
tested locally.  Problem reproduced and fixed.

-- 
Happy hacking
Petter Reinholdtsen

jepler:
Should be fine to merge when green, though if it fails on buildbot we'll probably have to put back the skip file

petterreinholdtsen:
Strange, suddenly tests/halcompile/userspace-count-names fail in rip-and-test, while not for package.  Anyone got any idea what is going on here?

jepler:
sooo halcompile turns out to depend on USER_MODULE_DIR
```
src/hal/utils/halcompile.g:BINDIR = os.getenv('USER_MODULE_DIR', None) or os.path.join(BASE, "bin")
```
with this change, the userspace module used by this test gets installed to a location not on PATH.  Way down in the raw error logs:
```
2022-07-16T16:46:13.9045685Z         both.hal:1: execv(userspace_count_names): No such file or directory
2022-07-16T16:46:13.9046391Z         both.hal:1: waitpid failed userspace_count_names userspace_count_names
2022-07-16T16:46:13.9046905Z         both.hal:1: userspace_count_names exited without becoming ready
2022-07-16T16:46:13.9047596Z         Note: Using POSIX non-realtime
2022-07-16T16:46:13.9047988Z         + for HAL in *.hal
```

sorry this is such a tangle!

petterreinholdtsen:
[Jeff Epler]
> sooo halcompile turns out to depend on USER_MODULE_DIR

Ah.  So perhaps these tests can get rid of the sudo stuff to install the
components in root controlled area...

Anyway, I will change runtest.in to pass a different variable not used
by halcompile, to avoid the problem.

I suspect runtest was intended to pass USER_MODULE_DIR on to test
scripts, but for some strange reason it did not work.
-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[Jeff Epler] ***@***.***> writes:
> these additional changes make the test pass locally.

Very good.  I extended the scope of this pull request to get the test
working again and updated the pull request.

> However, the very old commit message that added the skip file remark
> that it failed on the buildbot.  This was before 2.5.0 was released.

Disabled 2011-06-04, I guess it is time to get it back online. :)

And in the process I learned how to run a test locally, using

  EMC2_HOME=`pwd`/../../.. ./checkresult result

Perhaps how to write and run tests should be documented somewhere in
docs/?  I searched for 'checkresult' in all of docs/, without any hits.

-- 
Happy hacking
Petter Reinholdtsen

PR #1842, opened 2022-07-16
Gmoccapy: fix attribute error

TRothfelder:
On starting gmoccapy, I get an AttributeError because elem in line 725 is of type int which has no attribute called upper.
I get the error with the buildbot version for debian buster and confirmed it against the master branch.

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -722,7 +722,7 @@ class gmoccapy(object):
                 size=_DEFAULT_BB_SIZE,
                 image_name=f"img_ref_{elem}"
             )
-            btn.set_property("tooltip-text", _("Press to home {0} {1}").format(name_prefix_sg, elem))
+            btn.set_property("tooltip-text", _("Press to home {0} {1}").format(name_prefix_sg, str(elem).upper()))
             btn.connect("clicked", self._on_btn_home_clicked)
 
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
It look like the value should be a axis name, not an integer.  Perhaps instead of dropping upper(), a different variable name should be used instead?

jepler:
Is it possible that numbers 0/1/2 are occurring for nontrivial kinematics machines while letters X/Y/Z are occurring for trivial kinematics machines?  If so, maybe `str(elem).upper()`?

Since there's some confusion about this, maybe mention the specific steps to reproduce the problem so someone else can test it as well.

TRothfelder:
It might be a axis name or a joint index, depending on "self.trivial_kinematics".
However, I tracked it down to this commit: https://github.com/LinuxCNC/linuxcnc/commit/fc49ac84c7fd8d5039d19f69b0cccd6dd4ebd03c
The commit added upper().

If I understand the code in the right way, the line gets executed for both, the joints and the axes. So, in my case gmoccapy started and since the joints weren´t referenced, the line got executed for the joints. In my case: 2 joints for the y axis. Therefore elem should be an integer. However, if self.trivial_kinematics is true, then "elem" might be a string.
Better way might be: str(elem).upper()

jepler:
I'm glad we came to the same conclusion. Once you update this PR with that change I think it'll be good to merge. Thank you.

TRothfelder:
I commited the changes.

jepler (approved):
Thanks!

PR #1843, opened 2022-07-16
interpeter: allow comments that print params to set the float format

c-morley:
now &d,%f or %.xf (x being a number) is interpreted to be a float format
for params.

I've tested, looking for corner cases.
Could someone review my code please?


petterreinholdtsen:
What about test code in tests/ exercising this new feature?

c-morley:
What? that's what real developers do... :)
Right now I'd be happy if someone looked at the code and said it wasn't stupid looking :)

c-morley:
merged

PR #1848, opened 2022-07-19
Have autogen.sh copy in some files

jepler:
Debian policy calls for these files not to be duplicated. However, we need to do it in a portable way (e.g., works on RPM-based systems so /usr/share/misc/config.guess is not OK; does not touch network, so doing curl/wget of the canonical config.guess file is not OK)

While there should be an obvious way to copy these files, there's .. not. This is the best I could come up with, using files from automake without actually invoking it.

While I did not actually BUILD on an rpm-based system, I checked the content of the automake rpm of fedora 36 and it looks like it should work.

I'd rather avoid the new build-time dependency on automake, but "automake --print-libdir" was the only 'portable' way I found to print the location of a directory containing config.guess, config.sub and install-sh.

Alternative to #1625
Closes #1628


petterreinholdtsen:
Did you consider automake -ac ?

petterreinholdtsen:
After checking the alternative sources to find config.* and install-sh in Debian, three obvious packages are available, libtool, autotools-dev and automake.  Of these three, the former and the latter seem equally out of place for linuxcnc, and autotools-dev is depricated.  From this I conclude that your approach of copying them from automake is the best of these options.  The only alternative I can think of is to fetch it from the source, ie wget from https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess and https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub .  No idea where the latest edition of install-sh originates.

dwrobel:
> checked the content of the automake rpm of fedora 36 and it looks like it should work.

FYI, I did a [build](https://copr.fedorainfracloud.org/coprs/dwrobel/linuxcnc/build/4756128/) with that change and it looks that it works on fedora-36 while it broke on fedora-35 [log](https://download.copr.fedorainfracloud.org/results/dwrobel/linuxcnc/fedora-35-x86_64/04756128-linuxcnc/builder-live.log.gz) and CentOS-9 [log](https://download.copr.fedorainfracloud.org/results/dwrobel/linuxcnc/fedora-35-x86_64/04756128-linuxcnc/builder-live.log.gz):
```
configure: error: cannot find install-sh, install.sh, or shtool in "." "./.." "./../.."
```
The thing which affected it was that I'm not using `autogen.sh` but `autoreconf` to bootstrap sources.

Despite of that - it works while using:
 - autoconf 2.71-2.fc36
 - automake 1.16.5-7.fc36

but doesn't for the following:
 - autoconf   2.69-37.fc35
 - automake 1.16.2-5.fc35

Switching to use `autogen.sh` to boostrap fixed the issue. But I noticed the following differences:

1) Using `autoreconf` with `-s` allowed one to not install `install-sh` but generate symlinks _(this is not possible when using `autogen.sh`)_
```
$ rpm -qf /usr/bin/autoreconf
autoconf-2.71-2.fc36.noarch
$ autoreconf -sfi .
$ ls -ls install-sh
4 lrwxrwxrwx 1 dw dw 40 Aug 23 18:44 install-sh -> /usr/share/autoconf/build-aux/install-sh
```
2) `autogen-sh` uses `automake --print-libdir` to find `install-sh` and that produces on fedora-36:
```
$ automake --print-libdir
/usr/share/automake-1.16
```

while _(as you might noticed above)_ `autoreconf -sfi` symlinked a file from a different location:

```
$ ls -ls install-sh
4 lrwxrwxrwx 1 dw dw 40 Aug 23 18:44 install-sh -> /usr/share/autoconf/build-aux/install-sh
```

Those two `install-sh` comes from two different packages:
```
$ rpm -qf /usr/share/automake-1.16/install-sh 
automake-1.16.5-7.fc36.noarch

[dw@fedora build-aux]$ rpm -qf /usr/share/autoconf/build-aux/install-sh 
autoconf-2.71-2.fc36.noarch
```
Happily, they are almost identical:
```
$ diff -u /usr/share/automake-1.16/install-sh  /usr/share/autoconf/build-aux/install-sh 
--- /usr/share/automake-1.16/install-sh	2020-11-14 04:47:59.000000000 +0100
+++ /usr/share/autoconf/build-aux/install-sh	2020-12-14 14:19:12.000000000 +0100
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/sh
 # install - install a program, script, or datafile
 
 scriptversion=2020-11-14.01; # UTC
```

Summing up, if one wants to use `autoreconf` to boostrap sources then `autoconf-2.71` or newer needs to be used.

PR #1852, opened 2022-07-20
AXIS: use separate comp for pyvcp panel

phillc54:
Currently if a .axisrc or USER_COMMAND_FILE is configured to add hal 
pins via user_hal_pins, then the pins are not accessible if a pyvcp 
panel is also configured.


c-morley:
Is it true that AXIS can only have one pyvcp panel? So not possible AXIS will try to make two pyvcp components? ?From memory I think this is true and not a problem.

phillc54:
AFAIK pyvcp panels and gladvcp panels are singular but gladevcp embedded tabs may be multiple.

PR #1864, opened 2022-07-24
Re-enable tests/motion-logger/startup-gcode-abort and make it more robust

petterreinholdtsen:
Instead of using 'sleep(1)' to guess if linuxcnc is ready to take
instructions from the UI, look for the 'motion.in-position' pin
to show up in hal and abort just after this.  This make sure the
abort happen earlier in the startup process, and hopefully ensure
that the test succeed every time.


c-morley:
Could you look at the conflict? 
I'll merge after.

petterreinholdtsen:
[c-morley]
> Could you look at the conflict? 
> I'll merge after.

Done.

-- 
Happy hacking
Petter Reinholdtsen

PR #1865, opened 2022-07-24
Re-enable tests/abort/g64 as it seem to be working now.

petterreinholdtsen:


c-morley:
I wonder if the buildbot will choke on this - the probable reason it was disabled.
I'll merge and I guess we will see :)

PR #1866, opened 2022-07-24
Docs stupid simple stuff

smoe:
Series of smallish changes that were motivated by translation work.


docs/src/examples/spindle.adoc
```diff
@@ -129,7 +128,7 @@ the man pages, just say 'man limit2' or 'man near' in a terminal.
 
 [source,{hal}]
 ----
-# load real time a limit2 and a near with names so it is easier to follow
+# instantiate the real time modules limit2 and near with names so it is easier to follow their connections
```
hansu, line 131: Noy sure if 'instantiate' is the right word here.

smoe: I liked it. It is used http://linuxcnc.org/docs/stable/html/gui/qtvcp_libraries.html in the documentation already. Same context? I mean, you want to give that hal component a name to refer to it as an individual and trace its connections, "instantiate" is a better fit than "load".

hansu: But the corresponding command is named `loadrt` ;-)
And the context in [qtvcp_libraries.html](http://linuxcnc.org/docs/stable/html/gui/qtvcp_libraries.html) is python code, not HAL commands.
Sorry for bothering you with my comments as the first action when I'm back home :smile:

smoe: The "load" everyone can read in the line below. I still think to make a proper comment. And I still like it.

c-morley: We usually use 'load' for HAL components.
Instantiate is commonly used in python.

A typical machinist type user would probably be confused by instantiate.

silopolis:
Le mer. 27 juil. 2022 à 05:17, c-morley ***@***.***> a écrit :

> We usually use 'load' for HAL components.
> Instantiate is commonly used in python.
>
+1 for consistency
Also looks more appropriate in configuration context vs coding

petterreinholdtsen:
The proposed set of patches is clearly an improvement of the documentation, getting rid of both typos, inconsistencies and incomprehensible language.  I hope it is not held back over a single word, which obviously is not to be confused in the text, as the two next lines the sentence refer to use 'loadrt'.  Perhaps the patch can be merged into master and the instanciate->load change commited separately?

docs/src/drivers/pluto-p.adoc
```diff
@@ -41,23 +41,23 @@ http://wiki.linuxcnc.org/cgi-bin/wiki.pl?LinuxCNC_Supported_Hardware[LinuxCNC Su
...
 * The index input is triggered on the rising edge. Initial testing has
   shown that the QZx inputs are particularly noise sensitive, due to
-  being polled every 25ns. Digital filtering has been added to filter
-  pulses shorter than 175ns (seven polling times). Additional external
+  being polled every 25 ns. Digital filtering has been added to filter
+  pulses shorter than 175 ns (seven polling times). Additional external
   filtering on all input pins, such as a Schmitt buffer or inverter, RC
   filter, or differential receiver (if applicable) is recommended.
-* The IN1…IN7 pins have 22-ohm series resistors to their associated FPGA
+* The IN1…IN7 pins have 22Ω series resistors to their associated FPGA
```
jepler, line 60: Should there be a space before Ω here?

docs/src/examples/spindle.adoc
```diff
@@ -78,9 +77,9 @@ net spindle-enable spindle.0.on => parport.0.pin-14-out
 
 == Spindle Direction(((Spindle Direction Example)))
 
-If you have direction control of your spindle the HAL pins
-spindle.N.forward and spindle.N.reverse are controlled by M3
-and M4. Spindle speed 'Sn' must be set to a positive non-zero value for
+If you have direction control of your spindle, then the HAL pins
+'spindle.N.forward' and 'spindle.N.reverse' are controlled by the G-codes
```
jepler, line 81: Historically, I think we usually typeset pin names in bold, and sometimes showed the variable parts in italics. At least, this is what `halcompile`-generated manpages do.

jepler (changes requested):

docs/src/drivers/pluto-p.adoc
smoe: Yes, will add. https://tex.stackexchange.com/questions/20962/should-i-put-a-space-between-a-number-and-its-unit

I just wish that asciidoc could be a bit more expressive with the kerning. Style guide?

docs/src/examples/spindle.adoc
smoe: I typically left all the quotes as they came with the English. No idea why I added them here, likely because how these were presented before and after. There are quite some occurences also of backticks in other files we should somehow revisit.

silopolis:
Le jeu. 28 juil. 2022 à 18:00, Steffen Möller ***@***.***> a
écrit :

> Yes, will add.
> https://tex.stackexchange.com/questions/20962/should-i-put-a-space-between-a-number-and-its-unit
>
> I just wish that asciidoc could be a bit more expressive with the kerning.
> Style guide?
>

Will add together with quoting 👍

smoe:
Hello,
I just rebased on master and used the opportunity to squash those five commits into one. Addressed what I understood to address, i.e. instantiate -> load and the blank to the ohm symbol (which for some reason does not work here in this github box on chromium). I have not removed the quotes since these are about everywhere else, too, and we should have something as a style guide that is followed by a respective documentation-wide action.

docs/src/examples/spindle.adoc
smoe: I suggest to leave those quotes in until the style guide has an opinion.

hansu:
Regarding the spaces between value and unit:
What I know about that, is that a thin/narrow space is the best option. I found that: https://en.wikipedia.org/wiki/Thin_space
At least with my preview plugin it works. But maybe that's more something for the style guide?

smoe:
This space should be &#8201; according to https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/ - I like that. Should be possible to automate that transition, say by some some sed -i 's/ mm\>/&#8201;mm/g' or so. @silopolis, seems like that style guide will be a bit longer :)

smoe:
In above sed expression GitHad has already substituted that half-space expression (&amp;#8201;), i.e., seems to work :)

SebKuzminsky:
Thanks for all these docs cleanups!

silopolis:
Le jeu. 28 juil. 2022 à 18:02, Steffen Möller ***@***.***> a
écrit :

> +'spindle.N.forward' and 'spindle.N.reverse' are controlled by the G-codes
>
> I typically left all the quotes as they came with the English. No idea why
> I added them here, likely because how these were presented before and
> after. There are quite some occurences also of backticks in other files we
> should somehow revisit.
>

Personal style guide proposal would be surrounded by backticks as it really
is inline code (rendered as monospaced font)
But I've nothing against overloading the monospaced rendering with bold and
italic as @jepler suggested

PR #1868, opened 2022-07-26
Adjusted at_pid documentation and logic

petterreinholdtsen:
Made sure output is set to bias, not zero, when tuning cycle ends.  Added
reference to paper describing the tuning method implemented.  Reinserted
code to supress some messages while running, similar to the pid component.
Adjusted man page text to better explain how to use auto tuning.

### DIFF
--- a/src/hal/components/at_pid.c
+++ b/src/hal/components/at_pid.c
@@ -154,6 +154,8 @@ RTAPI_MP_ARRAY_STRING(names, MAX_CHAN,"at_pid names");
 static int debug = 0;		/* flag to export optional params */
 RTAPI_MP_INT(debug, "enables optional params");
 
+#define NAME "AT_PID"
+
 #define AUTO_TUNER 1
 #ifdef AUTO_TUNER
 #include "rtapi_math.h"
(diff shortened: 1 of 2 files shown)


src/hal/components/at_pid.c
```diff
@@ -966,11 +968,11 @@ static int export_pid(hal_pid_t * addr, char * prefix)
 	hal_export_funct(buf, calc_pid, addr, 1, 0, comp_id);
     if (retval != 0) {
 	rtapi_print_msg(RTAPI_MSG_ERR,
-	    "AT_PID: ERROR: do_pid_calcs funct export failed\n");
+	    NAME ": ERROR: do_pid_calcs funct export failed\n");
```
c-morley, line 971: This looks odd - Why NAME?
And the quote looks in the wrong spot?

petterreinholdtsen:
[c-morley]
>> @@ -966,11 +968,11 @@ static int export_pid(hal_pid_t * addr, char * prefix)
>  	hal_export_funct(buf, calc_pid, addr, 1, 0, comp_id);
>      if (retval != 0) {
>  	rtapi_print_msg(RTAPI_MSG_ERR,
> -	    "AT_PID: ERROR: do_pid_calcs funct export failed\n");
> +	    NAME ": ERROR: do_pid_calcs funct export failed\n");
>
> This looks odd - Why NAME?
> And the quote looks in the wrong spot?

It is to ensure the same string is used in all print lines, and to make
the difference to pid.c more easy to handle.  In time I hope to get a
similar change into pid.c.  The quote character is in the right spot.
In C, constant strings can be concatinated using "first" "second"
"third" sequences, and NAME is defined to "AT_PID", so the line actually
reads

  "AT_PID" ": ERROR: do_pid_calcs funct export failed\n");

-- 
Happy hacking
Petter Reinholdtsen

Issue #1832, opened 2022-07-13
PNCconf - add Mesa - specify "logical number"  


zz912:
I am trying to add another tab to PNCconf.
1) I can't find the definition of "logic number" anywhere.
File /home/zdenek/linuxcnc/linuxcnc-2.8/src/emc/usr_intf/pncconf/private_data.py:
![private_data-py](https://user-images.githubusercontent.com/96618597/178674576-4a3ff6af-30f9-48a8-b6fd-8f1fd9bd3340.png)
I dont know how correctly logic number use.

2) Perhaps the parameters of the 7i96 card are incorrectly defined. I didn't really understand what PCW wanted to tell me.
https://forum.linuxcnc.org/27-driver-boards/46356-can-i-use-7i96s-with-lcnc-2-8?start=10#247126
https://forum.linuxcnc.org/27-driver-boards/46356-can-i-use-7i96s-with-lcnc-2-8?start=10#247183
I dont know what is SB number

silopolis:
Le mer. 13 juil. 2022 à 09:22, zz912 ***@***.***> a écrit :

> I am trying to add another tab to PNCconf.
>

Don't know what's your project but do you know @jetthornton 's MesaCT?

https://github.com/jethornton/mesact

I'd have a look before investing time...

zz912:
I am trying to add 7i96s to LCNC 2.8. If I managed to do it, only the extended input options of the InM Input Module would not work.

JT Configurator is not part of LCNC 2.8 and cannot configure for 7i96s+LCNC 2.8

c-morley:
The number represents whether the GPIOI can be changed at run time or not. 0 = run time changes allowed, 100 means no runtime changes.

here is an example of the numbers use in the code:
https://github.com/LinuxCNC/linuxcnc/blob/2.8/src/emc/usr_intf/pncconf/pncconf.py#L2960

zz912:
What "run time" mean?

I thought that:
number<100 
means some automatic mod
If it was an automatic mode. Is this mod related to the -"max GPIOI" parameter?

number>100 
means some automatic mod

I found that when GPIOI was number 0, I could not change it to GPIOO or to S.STEPB. If I changed the number to 100, the change appeared in PNCconf.

However, I don't like doing something where I don't know the essence of the problem. That's why I asked this question.

but my theory not coresponded with example:
[...]

PR #1875, opened 2022-07-27
Pncconf - ADDING_A_MESA_CARD.md - add Logic number

zz912:
It should be solution for this Issue:
https://github.com/LinuxCNC/linuxcnc/issues/1832

I'm new to Github, so I don't know if I should make this change for version 2.9 after that.


silopolis:
This file should also be translated to AsciiDoc

PR #1876, opened 2022-07-27
Add simple autopkgtest check for the Debian package.

petterreinholdtsen:


petterreinholdtsen:
I am not sure it will work as it stands, as tests seem to assume they got write access to the directory with the source, which I believe is not the case when the tests are running.

If we get this test working, linuxcnc will use much less time to migrate from unstable to testing in Debian.

silopolis:
Le mer. 27 juil. 2022 à 16:48, petterreinholdtsen ***@***.***>
a écrit :

> If we get this test working, linuxcnc will use much less time to migrate
> from unstable to testing in Debian.
>
Very much worth trying then 👍

petterreinholdtsen:
I rewrote runtests to write into $TMPDIR instad of the source directory, and suspect this should work now.

smoe:
Debian performs tests on reverse dependencies when libraries etc are updated in the system. It is in our very interest to have some basic functionality, say a couple of HAL scripts to then be informed when something breaks when a dependency of ours is updated.

PR #1877, opened 2022-07-28
Replace 'custom' debian/copyright labels with more descriptive ones.

petterreinholdtsen:
This avoid a lintian warning.  The ISC license is described in
https://www.isc.org/licenses/ , while the GNU-All-Permissive
license is described in
https://www.gnu.org/prep/maintain/html_node/License-Notices-for-Other-Files.html .


jepler (approved):
I did not verify this is the correct license description, hopefully lintian does

PR #1879, opened 2022-07-28
Merge auto tune support into pid HAL component

petterreinholdtsen:
Instead of having a separate PID component with auto tuning support,
it is easier for users to have only one PID component where auto
tuning is a runtime option for those that want to use it.

Drop the at_pid HAL component and move its implementation and
documentation into the pid HAL component.  The code is marked
using #ifdef AUTO_TUNING to make it easy to identify in the code,
and trivial to compile out for those that want to.

There is some questions on how useful and efficient the implemented
auto tuning method is, and while I am sure it can be improved, it
did work for my machie and provided useful Pgain, Igain and Dgain
values for my X and Y axis.  As it is a runtime option, those
that do not want automatic tuning can simply ignore this feature.


petterreinholdtsen:
See also #1820.

hansu:
I wonder if we should keep the at_pid man page, only consisting of a note that it's now integrated in pid. But as the at_pid seems to have never really worked, this may can be skipped?

petterreinholdtsen:
[Hans]
> I wonder if we should keep the at_pid man page, only consisting of a
> note that it's now integrated in pid. But as the at_pid seems to have
> never really worked, this may can be skipped?

It would be easy to do.  I assume the idea is to give those that know
about at_pid a pointer where to find the feature going forward.  Note, I
believe at_pid did work.  It just was not pin compatible with pid until
I polished it a bit.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
>I assume the idea is to give those that know
about at_pid a pointer where to find the feature going forward.

Exactly.

c-morley:
I'll wait a bit before merging in case someone has a good reason not to. But i think at least the premise, is a good idea.

PR #1882, opened 2022-07-29
Copyright typo's and minor adjustment

havardAasen:
Handles some minor issues in GPL-3+ and LGPL-2 section

### DIFF
--- a/debian/copyright
+++ b/debian/copyright
@@ -472,19 +472,19 @@ License: GPL-2+
 
 License: LGPL-2
  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


petterreinholdtsen:
Note, your LGPL change do not match the header in src/hal/hal_lib.c.

havardAasen:
I believe this is something that should be addressed in `src/hal/hal_lib.c`. This file supposedly uses LGPL-2 but the copyright notice is taken from LGPL-2.1.
Or did you mean that this commit should fix that as well?

Would have been easier if we had used [SPDX IDs](https://spdx.dev/ids/).

PR #1888, opened 2022-07-31
Phillc54/pncconf

phillc54:
PnCconf:
Add some error messages.
Fix 7i96s stepgen polarity.
Add a 7i98 firmware.
Add some connector names.


petterreinholdtsen:
It is unclear from your commit message why this feature is needed, ie what problem it solve.  Some of the strings that seem to be user facing seem to lack the _() marker to ensure the strings are translatable.  I failed to understand why passing some command line info on to the probe code is a hack, perhaps the 'hack' message could elaborate a bit on this?

c-morley (approved):
Looks good. As noted, for translation the translatable text messages should have _() around them.

c-morley:
I believe the 'hack' is because it's inferred data rather then directly stated. ie PCI is a final option only because it has no other way to set it - process of elimination.

PR #1889, opened 2022-07-31
AXIS: Add [DISPLAY] PREVIEW_TIMEOUT setting.

PetteriAimonen:
Previously AXIS would freeze when loading a G-code file that takes very
long to execute and/or never terminates. This commit adds an optional
timeout that stops the loading of preview and shows a notice.

Retains old behavior if the .ini file setting is not specified.

--

I've been hitting this problem especially with NativeCAM.
If one sets some parameters such as step down to 0 by mistake, AXIS just freezes.

Example .ngc file with infinite loop for testing:

    o100 do
    o100 while [1 LT 2]

Example .ini configuration:

    [DISPLAY]
    PREVIEW_TIMEOUT = 5.0

### DIFF
--- a/docs/src/config/ini-config.adoc
+++ b/docs/src/config/ini-config.adoc
@@ -305,6 +305,7 @@ User Manual.
   If the update time is not set right the screen can become unresponsive or very jerky.
   A value of 100ms (0.1 seconds) is a common setting though a range of 50 - 200ms (.05 - .2 seconds) may be useable.
   An under powered CPU may see improvement with a longer setting. Usually the default is fine.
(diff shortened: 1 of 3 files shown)


c-morley:
Could you add the description to the AXIS docs ?

PetteriAimonen:
Sure, but I'm not sure at what point. I already added to the ini file 
document that has the other DISPLAY setting descriptions.

c-morley:
probably here would work:
http://linuxcnc.org/docs/devel/html/gui/axis.html#_ini_settings

PetteriAimonen:
Thanks, I didn't notice that section before because I was looking at the stable and not devel branch of online docs.

Added to AXIS doc and made the items a list. This is how it looks now:

image

c-morley:
perfect thanks!

PR #1890, opened 2022-08-02
More changes motivated while translating.

smoe:
The atom editor I was using has removed the final blanks in tables that @petterreinholdtsen once introduced because of a po4a inconvenience - was that ok?

### DIFF
--- a/docs/src/getting-started/running-linuxcnc.adoc
+++ b/docs/src/getting-started/running-linuxcnc.adoc
@@ -152,7 +152,7 @@ configuration specific information in a text file named after
 a specific configuration (inifilename.txt).  That latter two
 files are displayed when using the Configuration Selector.
 
-The supplied sample configurations may specify HALFILEs
+The supplied sample configurations may specify HAL files
 (filename.hal) in the configuration INI file that are not present in the
 copied directory because they are found in the system
 HAL file library.  These files can be copied to the user
--- a/docs/src/getting-started/running-linuxcnc.adoc
+++ b/docs/src/getting-started/running-linuxcnc.adoc
@@ -14,7 +14,7 @@ or select it in the 'Applications -> CNC' menu.
 == Configuration Launcher(((Configuration Launcher)))
 
 When starting LinuxCNC (from the CNC menu or from the command line without
-specifying an .ini file) the Configuration Selector dialog starts.
+specifying an INI file) the Configuration Selector dialog starts.
 
 The Configuration Selector dialog allows the user to pick one of their
 existing configurations (My Configurations) or select a new one (from the Sample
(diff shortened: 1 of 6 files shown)


petterreinholdtsen:
[Steffen Möller]
> The atom editor I was using has removed the final blanks in tables
> that @petterreinholdtsen once introduced because of a po4a
> inconvenience - was that ok?

If we do this, we should enforce a newer version of po4a in
src/configure.ac, ie adjust this code to check for 0.66, not 0.62:

        V=$(po4a --version| awk '/po4a version/ {print $3}')
        if dpkg --compare-versions 0.62 gt "$V"; then

Perhaps time to move to 0.67 to get support for asciidoc forced newlines
too?

-- 
Happy hacking
Petter Reinholdtsen

smoe:
I am all in for a move to 0.67. After all, it seems fair to presume that at some point someone else will run into this, too. I have undone the changes to the tables for this commit. But sooner or later this will break if we do not migrate. Please go for it.

petterreinholdtsen:
[Steffen Möller]
> I am all in for a move to 0.67. [...]  Please go for it.

I agree.  Perhaps you can make a patch for it, to gather opinions?

--
Happy hacking
Petter Reinholdtsen

smoe:
I just saw that 0.66 is already used for the CI.

docs/src/getting-started/running-linuxcnc.adoc
```diff
@@ -140,30 +140,30 @@ available in other GUIs as well.
...
+include one or more INI file (filename.ini) for related
 configurations and a tool table file (toolfilename.tbl).
-Additionally, resources may include halfiles (filename.hal,
+Additionally, resources may include HAL files (filename.hal,
 filename.tcl), a README file for describing the directory, and
 configuration specific information in a text file named after
 a specific configuration (inifilename.txt).  That latter two
 files are displayed when using the Configuration Selector.
 
 The supplied sample configurations may specify HALFILEs
```
hansu, line 155:
```suggestion
The supplied sample configurations may specify HAL files
```

smoe: I was not sure about this one, thought "HALFILE" was a configuration parameter and should not be changed.

hansu: Yes it is more or less but how it was originally written it was not really clear.
We could extend it to "The supplied sample configurations may specify HAL files as parameter `HALFILE=`in the [HAL] section". But I think that would be too much here.

silopolis:
Le mer. 3 août 2022 à 11:18, Hans ***@***.***> a écrit :

>
> Yes it is more or less but how it was originally written it was not really
> clear.
> We could extend it to "The supplied sample configurations may specify HAL
> files as parameter HALFILE=in the [HAL] section". But I think that would
> be too much here.
>

Never too much explicit!
Me likes it 👍

PR #1897, opened 2022-08-07
LCNC 2.8 - Pncconf - adding 7i96s

zz912:
I know that the stable version only serves to remove the bug.

I would like to make an exception here. Reason:
I'm not a fan of the LPT port. I find a used 7i96 card + a stable version of LCNC to be the easiest way to reliably use LCNC.

When I started with LinuxCNC and the 7i96 card, its support was also only in the master version. I know that changing a few strings in the repositories is a very simple matter for good Linux users. When I started with Linux, it was hell for me.

I try to keep the KISS rule. Keep It Simply Stupid.
This should be the way to go for complete Linux beginners:
Buy a 7i96s card => install a stable version of LCNC => use Pncconf configurator to set LCNC => MILLING!!!
Believe me, some beginners are happy to be able to count the number of pulses/units.

If there was no shortage of chips for 7i96 cards, I would not have asked for this exception.

This configuration must not be merged into the master (2.9) branch.

I posted this configuration here but there was no response.
https://forum.linuxcnc.org/27-driver-boards/46356-can-i-use-7i96s-with-lcnc-2-8?start=20


phillc54:
A problem that I see with this is that if/when the user upgrades to 2.9 then their config will be broken as the i/o pin names are different.

zz912:
Every upgrade requires a configuration change, this is not a new thing. For example, moving from version 2.7 to 2.8 required HAL modifications for the multispindle "spindle. => spindle.0."

hansu:
Not necessarily. In 2.9 in most cases features were added, so many configs run under 2.9 without any change. But the step from 2.7 to 2.8 is maybe not the best example as there was the big change were axis and joints were treated separately. For that an update script was provided that did the changes for you.

phillc54:
I see that there is a now [PR](https://github.com/LinuxCNC/linuxcnc/pull/1899) for updating hostmot2 in 2.8. 
When that gets merged then it would be possible to modify PnCconf to fully support the 7i96s.

zz912:
PR for upgrade hostmot2 was merged:
https://github.com/LinuxCNC/linuxcnc/pull/1899
, so my solution for 7i96S is not needed.

PR #1900, opened 2022-08-07
Add reset pins for halui Rapid Override, Feed Override and Spindle Overrides

russellgower:
Add reset pins for halui Rapid Override, Feed Override and Spindle Overrides

### DIFF
--- a/docs/man/man1/halui.1
+++ b/docs/man/man1/halui.1
@@ -145,6 +145,9 @@ pin to enable direct spindle override value input
 .B halui.spindle.N.override.increase\fR bit in \fR
 pin for increasing the SO (+=scale)
 .TP
+.B halui.spindle.N.override.reset\fR bit in \fR
+pin for resetting the scale SO value (scale=1.0)
+.TP
 .B halui.spindle.N.override.scale\fR float in \fR
 pin for setting the scale of counts for SO
 .TP
@@ -508,6 +511,9 @@ pin to enable direct value feed override input
 .B halui.feed\-override.increase\fR bit in \fR
 pin for increasing the FO (+=scale)
 .TP
+.B halui.feed\-override.reset\fR bit in \fR
+pin for resetting the FO (scale=1.0)
+.TP
 .B halui.feed\-override.scale\fR float in \fR
 pin for setting the scale on changing the FO
 .TP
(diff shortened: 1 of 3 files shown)


petterreinholdtsen:
This commit should also update docs/man/man1/halui.1.

Would be nice if the commit message explained why it is needed.

pippin88:
I am keen for this function.

Being able to quickly reset overrides would be very helpful.

Some GUIs implement their own pins to add this function - better to have the function in HALUI, not GUI dependent.

petterreinholdtsen:
[pippin88]
> Some GUIs implement their own pins to add this function - better to
> have the function in HALUI, not GUI dependent.

Which GUIs implement their own pins for the reset function?  Do they
need to be patched to be compatible with this new halui functionallity?

-- 
Happy hacking
Petter Reinholdtsen

russellgower:
I've added the missing man page update and have amended the commit message to give a bit of context. 

I found my git fu somewhat lacking when I tried to work out how to do that, but I think I got there in the end - hopefully I haven't messed it up to badly.

petterreinholdtsen:
[russellgower]
> I've added the missing man page update and have amended the commit
> message to give a bit of context.

Look good.  I notice the first line of the commit message is a bit log
and has been cut.  Perhaps rephrase it to 'Add reset pins for halui
Rapid, Feed and Spindle Overrides' to make it fit within the github (and
git?) constraints?  Just nitpick.

> I found my git fu somewhat lacking when I tried to work out how to do
> that, but I think I got there in the end - hopefully I haven't messed
> it up to badly.

Look good to me.

You mentioned existing GUIs alrady having their own reset pins?  Which
are they?  No need to change them to use these new pins?

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
The HALUI program is just another GUI without the G. It has no higher priority or significance then the GUI you are using.
Now if your GUI is resetting something like an encoder count to 0 - you really should modify the GUI. 

Unfortunately there is no mechanism for linuxcnc's GUI/UIs to communicate to each others directly, nor any real protocol of how things would be preferred to work.

The encoder 'windup' problem sounds like a GMoccapy bug.

zz912:
The HALUI program is just another GUI without the G. It has no higher priority or significance then the GUI you are using.
Now if your GUI is resetting something like an encoder count to 0 - you really should modify the GUI.

Unfortunately there is no mechanism for linuxcnc's GUI/UIs to communicate to each others directly, nor any real protocol of how things would be preferred to work.

The encoder 'windup' problem sounds like a GMoccapy bug.

zz912:
I thought halui was superior to all GUIs.

I guess I'm describing my problem wrong. Gmoccapy override is intended for a potentiometer with mechanical stops that has a maximum and a minimum. I use the Quadratic encoder, which can take values from -infinity to +infinity.

If halui is not superior to Gmoccapy please ignore my previous post.

c-morley:
Looking at the GMoccapy docs if you are using an encoder then you are probably using the wrong Gmoccapy pins. there are different pins for encoders vrs potentiometers.

c-morley:
back to the code -  are these reset pins or preset pins?

PR #1901, opened 2022-08-08
Update remap.adoc

ZincBoy:
Both the "M/m" and "G/g" argspec words are not currently supported by interp_remap.cc. These should be removed from the argspec description as they are not passed to the called remapping code. Either that or the interp_remap.cc should be fixed.

### DIFF
--- a/docs/src/remap/remap.adoc
+++ b/docs/src/remap/remap.adoc
@@ -416,7 +416,7 @@ any extra parameters present.
 Note that RS274NGC rules still apply - for instance you may use axis
 words (eg X,Y,Z) only in the context of a G-code.
 
-`ABCDEFGHIJKMPQRSTUVWXYZ`::
+`ABCDEFHIJKPQRSTUVWXYZ`::
   Defines a required word parameter: an uppercase letter specifies that
   the corresponding word *must*
   be present in the current block. The word`s value will be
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
Given that both G and M are handled specially in src/emc/rs274ngc/interp_remap.cc already, what do you mean by 'interp_remap.cc should be fixed'?  To me it look like G and M can not be parameters to G and M methods, which make it unclear to me what the alternative to removing these from the list of possible parameters would be.

ZincBoy:
I was unsure if this was an issue with interp_remap.cc or not.  It would make sense that G and M words are not supported as argument parameters. Some controls (HAAS) do allow M to be used in G remaps.  Not having M available makes porting functions require changing parameter codes and requires additional post-processor updates.  I think the right thing to do is update the docs to remove G and M from argspec and be done with it.

petterreinholdtsen:
[Alex Lait]
> I was unsure if this was an issue with interp_remap.cc or not.  It
> would make sense that G and M words are not supported as argument
> parameters. Some controls (HAAS) do allow M to be used in G remaps.
> Not having M available makes porting functions require changing
> parameter codes and requires additional post-processor updates.

Perhaps it is an idea to be compatible with HAAS?

> I think the right thing to do is update the docs to remove G and M
> from argspec and be done with it.

Why?

-- 
Happy hacking
Petter Reinholdtsen

ZincBoy:
After thinking about it more, both G and M should be disallowed in argspec.  The interpreter allows M and G codes on the same line and would have no way of determining if the M or G was a parameter or a new code.  I don't think we want to change such a fundamental behavior for a minor compatibility difference.

L is one word that you would think should work but does not.  It is correctly documented as not working but I can't think of why that might be.  It look like it should work in interp_remap.cc but I have confirmed it does not.  If I find something in interp_remap.cc I will submit a patch there.

PR #1902, opened 2022-08-08
Use xdg-open consistently in deb packages and fix lintian overrides

petterreinholdtsen:
Instead of using both xdg-open and x-www-browser depending on
package, use only one of them.  Picked xdg-open after consulting

Drop irrelevant overrides and insert correct ones refering to xdg-open.


silopolis:
Hey who's got commit rightz?! 👍💪😎

PR #1903, opened 2022-08-08
Added docs build rules for German (de) HTML and PDF.

petterreinholdtsen:


hansu:
Why not merge these two PRs?

petterreinholdtsen:
The failing tests are due to bugs in the de.po files that have gone undetected until now.

petterreinholdtsen:
I added a commit to fix the german issues.  These should be fixed in Weblate instead, and be removed from this pull request before it is merged into master, to avoid git conflicts on Weblate.

hansu:
>msgid "images/encoder-counts-math.png"
>msgstr "images/encoder-zählt-math.png"

Haha that's why I mentioned some time ago to limit the scope of what is going into the documentation.pot...

>I added a commit to fix the german issues. These should be fixed in Weblate instead, and be removed from this pull request before it is merged into master, to avoid git conflicts on Weblate.

So why haven't you just done it in Weblate?

hansu:
I corrected them in Weblate so you can remove that fix here and check if it builds.

petterreinholdtsen:
I decided to ask @smoe to train the German translators instead of fixing all my self, hoping to avoid such issues in the future.

smoe:
Builds. Thank you both.

hansu:
I think it is more caused by using the automatic translation suggestions. If you look through them, everything looks fine, but it's easy to overlook that the filename is wrongly translated as well.

petterreinholdtsen:
[Hans]
> I corrected them in Weblate so you can remove that fix here and check
> if it builds.

I'll do that as soon as the Weblate fixes make it into master. :)

-- 
Happy hacking
Petter Reinholdtsen

hansu:
>I'll do that as soon as the Weblate fixes make it into master. :)

Ah, you are right :)

petterreinholdtsen:
After merging from Weblate, it look like there is still one build breaking error in de.po that need to be fixed.

hansu:
> After merging from Weblate, it look like there is still one build breaking error in de.po that need to be fixed.

Which one? All builds looks green ...

smoe:
Fixed.

petterreinholdtsen:
Pulled in the latest changes from Weblate, and now the German build work.

PR #1904, opened 2022-08-08
Fix typos

luzpaz:
Found via `codespell -q 3 -S *.po,*.pot,*.ts,./.git/logs,./share,./docs/man/es,./configs/attic,*_fr.*,*_es.*,README_es -L ans,ba,bulle,componentes,doubleclick,dout,dum,fo,halp,ihs,inout,parm,parms,ro,ser,te,ue,wille,wonte`


SebKuzminsky:
Thanks @luzpaz !  It'd be cool to have something like this as one of the automated tests, to catch new typos as we make them.

PR #1906, opened 2022-08-08
Fix source typo missed in a previous commit

luzpaz:
This fix didn't make it in to f69f00b58eb


petterreinholdtsen:
As this changed the name of HAL network, I compared it with the content of configs/sim/qtvcp_screens/woodpecker/woodpecker_postgui.hal to verify the typo was only used once.

PR #1909, opened 2022-08-09
Clarify argspec parameter behavior

ZincBoy:
Clarifying the argspec parameter behavior.  I think the axis words not being available unless enabled is actually a bug. However fixing this is not likely to be easy so I think we should document the current behavior.


jepler:
@ZincBoy this PR was created as a draft. Is it ready for review?

ZincBoy:
I have completed my testing and I didn't find anything else to include at this time.  I believe this is ready to get included.

jepler:
THanks

Issue #1873, opened 2022-07-27
Gmoccapy - [RS274NGC] - FEATURES


zz912:
I integrated probe for autotoolmeasurement:
I followed the instructions:
http://linuxcnc.org/docs/stable/html/gui/gmoccapy.html#gmoccapy:auto-tool-measurement

`[RS274NGC]`
`# Enables the reading of INI and HAL values from gcode`
`FEATURES=12`
``
`# is the sub, with is called when a error during tool change happens, not needed on every machine configuration`
`ON_ABORT_COMMAND=O <on_abort> call`
``
`# The remap code`
`REMAP=M6  modalgroup=6 prolog=change_prolog ngc=change epilog=change_epilog`

I wanted to see what the FEATURES parameter means.
I found:
http://linuxcnc.org/docs/stable/html/config/ini-config.html#gcode:ini-features
`[NOTE] The above six options were controlled by the FEATURES bitmask in versions of LinuxCNC prior to 2.8. This INI tag will no longer work.`

ZincBoy:
The FEATURES=12 line in the documentation should be replaced with:

INI_VARS = 1
HAL_PIN_VARS = 1

However, the default behavior is for both of those to be set to 1 (enabled) so they are redundant.  The FEATURES=12 can be completely removed without effect unless the INI has explicitly set INI_VARS or HAL_PIN_VARS to 0.

hansu:
See also here for more details: https://github.com/LinuxCNC/linuxcnc/pull/1915

hansu:
See updated docs here: http://linuxcnc.org/docs/html/config/ini-config.html

PR #1911, opened 2022-08-09
Update gmoccapy.adoc

ZincBoy:
Related to:
https://github.com/LinuxCNC/linuxcnc/issues/1873

### DIFF
--- a/docs/src/gui/gmoccapy.adoc
+++ b/docs/src/gui/gmoccapy.adoc
@@ -1037,8 +1037,9 @@ Modify your INI File to include the following:
 [source,{ini}]
 ----
 [RS274NGC]
-# Enables the reading of INI and HAL values from G-code
-FEATURES=12
+# Make sure INI_VARS and HAL_PIN_VARS are not set to 0. They are set to 1 by default.
+# LinuxCNC versions below 2.8 require the following line to be uncommented:
+# FEATURES = 12
 
 # is the sub, with is called when a error during tool change happens, not needed on every machine configuration
 ON_ABORT_COMMAND=O <on_abort> call
--- a/docs/src/gui/gmoccapy.adoc
+++ b/docs/src/gui/gmoccapy.adoc
@@ -1037,16 +1037,15 @@ Modify your INI File to include the following:
 [source,{ini}]
 ----
 [RS274NGC]
-# Make sure INI_VARS and HAL_PIN_VARS are not set to 0. They are set to 1 by default.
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
When do you expect this draft to be ready to go into master?

docs/src/gui/gmoccapy.adoc
```diff
@@ -1037,8 +1037,9 @@ Modify your INI File to include the following:
 [source,{ini}]
 ----
 [RS274NGC]
-# Enables the reading of INI and HAL values from G-code
-FEATURES=12
+# Make sure INI_VARS and HAL_PIN_VARS are not set to 0. They are set to 1 by default.
+# LinuxCNC versions below 2.8 require the following line to be uncommented:
+# FEATURES = 12
```
hansu, line 1042: I would word it a bit differently to make it more clear

```suggestion
# These two values doesn't have to be set explicitly because their default value is 1. Only make sure they are not overwritten in this case.
# In LinuxCNC versions prior to 2.8 these settings were set with the FEATURES bitmask (FEATURES = 12)
INI_VARS = 1
HAL_PIN_VARS = 1
```

And I would suggest to target this PR to 2.8 and merge it up to master.

hansu: Or maybe remove these lines completely and add your lines as a note below the code block:
```
[NOTE]
Make sure INI_VARS and HAL_PIN_VARS are not set to 0. They are set to 1 by default. 
```
The hint for versions < 2.8 I would remove, as it is more confusing than helpful.

docs/src/gui/gmoccapy.adoc
```diff
@@ -1037,14 +1037,14 @@ Modify your INI File to include the following:
...
-FEATURES=12
-
 # is the sub, with is called when a error during tool change happens, not needed on every machine configuration
 ON_ABORT_COMMAND=O <on_abort> call
 
 # The remap code
 REMAP=M6  modalgroup=6 prolog=change_prolog ngc=change epilog=change_epilog
+
+[NOTE]
+Make sure INI_VARS and HAL_PIN_VARS are not set to 0. They are set to 1 by default.
```
hansu, line 1047: I guess having `[NOTE]` inside the code block doesn't show it as a note...

ZincBoy: Yes, I figured out that there was a preview tab I should have used :)  I added an updated commit.

PR #1934, opened 2022-08-16
Delete README_es.md

smoe:
I do not think we should have other languages singled out. The README.adoc should by some magic be inviting already.


jepler:
Is the target branch (dh_install) of this PR correct?

smoe:
> Is the target branch (dh_install) of this PR correct?

I messed that up. Wanted to help @SebKuzminsky with the Debian packaging and lost my orientation.

jepler:
Ok, I tried changing the base branch but it still looked wrong. It may be better too close this PR and open a fresh one

smoe:
Maybe @SebKuzminsky accepts it and it goes in via that route - trivial and not urgent. But yes, can also create another PR.

PR #1936, opened 2022-08-18
Make sure HAL_FIELDS in halui.cc is complete

petterreinholdtsen:
A trailing space after a backslash make all parts following
ro_reset fail to be part of the define.

Discovered by compiling with clang.


jepler (approved):
Thank you!

PR #1937, opened 2022-08-18
Made sure 'to_python(unsigned u)' only compare with LONG_MAX when needed

petterreinholdtsen:
As LONG_MAX is larger than any value representable by unsigned (int)
on most platforms, the test in the method do not make sense on these.
Only compare with LONG_MAX on platforms where sizeof(int) == sizeof(long)
and skip that code in the more common case where sizeof(int) < sizeof(long).

Get rid of warning discovered with clang.

### DIFF
--- a/src/hal/halmodule.cc
+++ b/src/hal/halmodule.cc
@@ -39,8 +39,7 @@ PyObject *to_python(bool b) {
     return PyBool_FromLong(b);
 }
 
-PyObject *to_python(unsigned u) {
-    if(u < LONG_MAX) return PyLong_FromLong(u);
+PyObject *to_python(unsigned int u) {
     return PyLong_FromUnsignedLong(u);
 }
 


jepler:
I think the correct fix is to simply always `return PyLong_FromUnsignedLong(u);`, dropping the conditional call to `_FromLong` entirely.  In python2, the C API differentiated between "regular" and "long" ints, so it used to read
```c
PyObject *to_python(unsigned u) {
    if(u < LONG_MAX) return PyInt_FromLong(u);
    return PyLong_FromUnsignedLong(u);
 }
```

As for the claim about range of the two integral types .. I thought it's quite possible for an `unsigned` to hold a value that cannot be stored in `long`, on platforms where `sizeof(unsigned) == sizeof(long)`. Think of all classic 32-bit ILP32 systems... am I missing something about your statement?

petterreinholdtsen:
[Jeff Epler]
> I think the correct fix is to simply always `return
> PyLong_FromUnsignedLong(u);`, dropping the conditional call to
> `_FromLong` entirely.

It has so far, at least.  Perhaps a better fix, but see below.

> As for the claim about range of the two integral types .. I thought
> it's quite possible for an `unsigned` to hold a value that cannot be
> stored in `long`, on platforms where `sizeof(unsigned) ==
> sizeof(long)`. Think of all classic 32-bit ILP32 systems... am I
> missing something about your statement?

In the sizeof(int) == sizeof(long), INT_MAX would be identical to
LONG_MAX and PyInt_FromLong() should be used.

Perhaps better to protect the code with

  #if INT_MAX == LONG_MAX

do document that this is only useful on those platforms?

I updated the patch to do this instead.
-- 
Happy hacking
Petter Reinholdtsen

jepler:
Backing up, we have an `unsigned int` (written here just as `unsigned`) value that we need to convert to a Python integer.

The related APIs (in Python 3.9) that might help implement this are
```
PyAPI_FUNC(PyObject *) PyLong_FromLong(long);
PyAPI_FUNC(PyObject *) PyLong_FromUnsignedLong(unsigned long);
```

Isn't it always the case that any `unsigned int` can be converted to `unsigned long` without loss, (C99 draft standard 6.2.5.8 states that "the range of values of the type with smaller integer conversion rank is a subrange of the values of the other type." while 6.3.1.1 states that the rank of `int` is less than `long`, with the rank of `unsigned T` being equal to the rank of `signed T`)?

And if so, can't the use of `FromUnsignedLong` just be unconditional?

petterreinholdtsen:
[Jeff Epler]
> Isn't it always the case that any `unsigned int` can be converted to
> `unsigned long` without loss, (C99 draft standard 6.2.5.8 states that
> "the range of values of the type with smaller integer conversion rank
> is a subrange of the values of the other type." while 6.3.1.1 states
> that the rank of `int` is less than `long`, with the rank of `unsigned
> T` being equal to the rank of `signed T`)?

As far as I know, this can always be done.
>
> And if so, can't the use of `FromUnsignedLong` just be unconditional?

I assumed, based on the current code, that the split was intentional and
tried to keep it.  Your comments make it sound like the intention were
never there.  My guess, never having writtent a python module in C, is
that both long and unsigned long will work just fine in python.

-- 
Happy hacking
Petter Reinholdtsen

jepler (approved):
Thanks!  I'm sorry I had trouble communicating, confusing the history of why it was this way with the modification I was suggesting

PR #1938, opened 2022-08-18
Make sure get_speed_override() return value

petterreinholdtsen:
The method do not return the override value, but always true.  It
need information on which spindle to return value for, similar
to set_speed_override, to give a sensible value.

Fixes this warning from building with clang:

emc/rs274ngc/interpmodule.cc:310:26: warning: address of array 'interp._setup.speed_override' will always evaluate to 'true' [-Wpointer-bool-conversion]
    return interp._setup.speed_override;
    ~~~~~~ ~~~~~~~~~~~~~~^~~~~~~~~~~~~~


jepler (approved):
thank you!

PR #1939, opened 2022-08-19
Make sure C++ delete is operating on correct class

petterreinholdtsen:
The breq->_nml pointer is of type void*, while C++ delete expect a
object pointer.  The object assigned to _nml is of class NML, so
cast to (NML*) before deleting it.

Discovered by building with clang.

### DIFF
--- a/src/libnml/buffer/rem_msg.hh
+++ b/src/libnml/buffer/rem_msg.hh
@@ -23,6 +23,7 @@
 #define REM_MSG_HH
 
 
+class NML;
 class CMS_DIAGNOSTICS_INFO;
 struct CMS_HEADER;
 
@@ -82,7 +83,7 @@ struct REMOTE_BLOCKING_READ_REQUEST:public REMOTE_CMS_REQUEST {
     /* to this client */
     long timeout_millis;	/* Milliseconds for blocking_timeout or -1 to 
 				   wait forever */
-    void *_nml;
+    NML *_nml;
     void *_data;
     void *_reply;
 };
--- a/src/libnml/nml/nml_srv.cc
+++ b/src/libnml/nml/nml_srv.cc
@@ -70,7 +70,7 @@ NML_SERVER::NML_SERVER(NML * _nml, int _set_to_master):CMS_SERVER()
 	    } else {
 		rcs_print_error
 		    ("NML_SERVER:(ERROR) ProcessType was REMOTE.\n");
-		_nml = (NML *) NULL;
+		_nml = NULL;
 	    }
 	} else {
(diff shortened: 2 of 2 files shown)


jepler:
Why not declare the field as having the right type instead of `void*`?
```diff
From cafb077ececed7c9ffb2232c2855ec72fe296d2c Mon Sep 17 00:00:00 2001
From: Jeff Epler <jepler@gmail.com>
Date: Fri, 19 Aug 2022 09:35:00 -0500
Subject: [PATCH] Correctly declare field's type

This is an alternative to #1939 which adds a cast at the site where
`breq->_nml` is deleted. Both allow the right destructor to be called
at runtime.
---
 src/libnml/buffer/rem_msg.hh | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

diff --git a/src/libnml/buffer/rem_msg.hh b/src/libnml/buffer/rem_msg.hh
index 8a739076ff..f0c4e0c599 100644
--- a/src/libnml/buffer/rem_msg.hh
+++ b/src/libnml/buffer/rem_msg.hh
@@ -23,6 +23,7 @@
 #define REM_MSG_HH
 
 
+class NML;
 class CMS_DIAGNOSTICS_INFO;
 struct CMS_HEADER;
 
@@ -82,7 +83,7 @@ struct REMOTE_BLOCKING_READ_REQUEST:public REMOTE_CMS_REQUEST {
     /* to this client */
     long timeout_millis;	/* Milliseconds for blocking_timeout or -1 to 
 				   wait forever */
-    void *_nml;
+    NML *_nml;
     void *_data;
     void *_reply;
 };
-- 
2.30.2

```

petterreinholdtsen:
[Jeff Epler]
> Why not declare the field as having the right type instead of `void*`?

No idea why the type is 'void*'.  I assumed it was used for different
types and could not be fixed to one type.  If your patch work, I suggest
we use it.

--
Happy hacking
Petter Reinholdtsen

jepler:
I was relying on your analysis that the pointer actually stored in a REMOTE_BLOCKING_READ_REQUEST was always an NML object.

jepler (approved):
Thanks!

petterreinholdtsen:
I rewrote the patch to change void* to NML*, and it seem to build just
fine.
-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[Jeff Epler]
> I was relying on your analysis that the pointer actually stored in a
> REMOTE_BLOCKING_READ_REQUEST was always an NML object.

OK.  It is in the function doing the delete.  I do not know where else
it is used.

-- 
Happy hacking
Petter Reinholdtsen

PR #1940, opened 2022-08-19
Remove deprecated axis and deprecated_axes fields

petterreinholdtsen:
These were flagged as deprecated in 2015, and 7 years seem like long
enough time for people to prepare for its removal.


src/emc/task/taskmodule.cc
```diff
@@ -330,12 +325,10 @@ BOOST_PYTHON_MODULE(emctask) {
 	;
 
 #pragma GCC diagnostic push
```
jepler, line 327: can we be rid of the diagnostic push and (below) pop too? They exist only to avoid a compoiler diagnostic when the deprecated field was referred to.

jepler:
```
          File "/home/runner/work/linuxcnc/linuxcnc/lib/python/linuxcnc_util.py", line 47, in wait_for_linuxcnc_startup
            or (self.status.axes == 0) \
        AttributeError: 'linuxcnc.stat' object has no attribute 'axes'
```
looks like there at least one more use of the property, from python code.
```
lib/python/linuxcnc_util.py:                or (self.status.axes == 0) \
```


as well as a doc mention to remove:
```
docs/src/code/code-notes.adoc:`status.motion.traj.axes` (deprecated)::
```

jepler (changes requested):

petterreinholdtsen:
I corrected the python code and updated the documentation.  Once all build tests pass, I'll rebase and squash the patch.  I decided to keep documentation on the field, mentioning its removal version.  Is it better to remove all traces of it?

jepler:
doc change looks good, feel free to ping after you've squashed it to your satisfaction

petterreinholdtsen:
[Jeff Epler]
> doc change looks good, feel free to ping after you've squashed it to
> your satisfaction

Done.

-- 
Happy hacking
Petter Reinholdtsen

jepler:
Thanks!

PR #1945, opened 2022-08-20
Dropped unneeded {} around initializer list

petterreinholdtsen:
Discovered while building with clang.


jepler (approved):
It would be helpful if in the future you can post an example of the particular diagnostic as printed by the compiler.

I assume this is
```
<source>:2:25: warning: braces around scalar initializer [-Wbraced-scalar-init]
```

PR #1953, opened 2022-08-21
Bump the C/C++ language standards version

jepler:
This is cumulative with #1278 and just bumps our C & C++ compiler standards to gnu11 for C files and gnu++17 for C++ files. It should be popped out of draft once that PR Is merged and this one is rebased.

The main reason to consider NOT doing this is that it could lead to accidental inclusion of C language features that are not supported by the kernel's C dialect -- Kernel 5.18 is the first release which also switched to the gnu11 standard. This would affect building for kmod+rtai, something I don't think sees buildbot testing these days anyway.


petterreinholdtsen:
Did you base this pull request on the wrong branch?  Seem to contain commits also in pull request #1278.

jepler:
I updated this and also pushed it to a branch in linuxcnc/linuxcnc named jepler/bump-gcc-std so we can see how it affects all buildbot target systems

jepler:
This was also 'green' on the buildbot

PR #1958, opened 2022-08-22
Switch to python3 for fanuc_200f.py

dwrobel:
Also adds executable bit.


jepler (approved):
Thanks!

PR #1959, opened 2022-08-22
Fix use-after-free in sendkeys.c

dwrobel:
Fixes the following warnings reported by gcc 12.1.1:

 hal/user_comps/sendkeys.c: In function ‘init’:
 hal/user_comps/sendkeys.c:110:25: warning: pointer ‘codes’ may be used after ‘realloc’ [-Wuse-after-free]
   110 |                         free(codes);
       |                         ^~~~~~~~~~~
 hal/user_comps/sendkeys.c:107:26: note: call to ‘realloc’ here
   107 |                     v1 = realloc(codes, (index + 1) * sizeof(int));
       |                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 hal/user_comps/sendkeys.c:111:25: warning: pointer ‘pins’ may be used after ‘realloc’ [-Wuse-after-free]
   111 |                         free(pins);
       |                         ^~~~~~~~~~
  hal/user_comps/sendkeys.c:108:26: note: call to ‘realloc’ here
   108 |                     v2 = realloc(pins, (index + 1) * sizeof(int));
       |                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/src/hal/user_comps/sendkeys.c
+++ b/src/hal/user_comps/sendkeys.c
@@ -107,8 +107,8 @@ int init(int argc, char* argv[]){
                     v1 = realloc(codes, (index + 1) * sizeof(int));
                     v2 = realloc(pins, (index + 1) * sizeof(int));
                     if (!v1 || !v2) {
-                        free(codes);
-                        free(pins);
+                        free(v1 ? v1 : codes);
+                        free(v2 ? v2 : pins);
                         rtapi_print_msg(RTAPI_MSG_ERR, "sendkeys.N.keycode error\n");
                         return -ENOMEM;
                     } else {


src/hal/user_comps/sendkeys.c
```diff
@@ -107,8 +107,12 @@ int init(int argc, char* argv[]){
                     v1 = realloc(codes, (index + 1) * sizeof(int));
                     v2 = realloc(pins, (index + 1) * sizeof(int));
                     if (!v1 || !v2) {
-                        free(codes);
-                        free(pins);
+                        if (!v1) {
+                            free(codes);
```
petterreinholdtsen, line 111: A good defensive coding strategy is to assign NULL to any pointer that has been freed.  Perhaps a good idea to do here too?  I know the variables are local and there is a return a few lines below.

andypugh:
With the changes, if the realloc of pins fails then codes is never freed?

petterreinholdtsen:
[andypugh]
> With the changes, if the realloc of pins fails then codes is never
> freed?

Yeah, it probably need something like this:

if (v1) {
   free(v1);
   v1=NULL;
} else {
   free(codes);
   codes=NULL;
}

and similar for v2/pins.
-- 
Happy hacking
Petter Reinholdtsen

andypugh:
No, I think that your code is completely wrong? 

If the realloc fails for either v1 or v2 we want to free memory and exit with an error. 
My original code does exactly that. 

All the proposed solutions _don't_ do that, though they might make the compiler happier. 

Breaking the intent of the code to silence the compiler seems like the wrong way round?

petterreinholdtsen:
[andypugh]
> If the realloc fails for either v1 or v2 we want to free memory and
> exit with an error.  My original code does exactly that.
>
> All the proposed solutions _don't_ do that, though they might make the
> compiler happier.

Which part is not done by the proposed patches?  As far as I can tell
they still free the memory and exit with an error.  The discussion is
how to ensure the correct and complete set of memory is released.

> Breaking the intent of the code to silence the compiler seems like the
> wrong way round?

I guess the intent might be better expressed with a code comment, if the
current proposals is breaking the intent.  At least I have no idea what
you are talking about.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
My point is that if _either_ fail then we should free _both_ and exit. 

So by your logic are you saying that we need:

''''
...
int flag = FALSE;
if (v1){
    free(v1);
    v1 = NULL;
} else {
    free(codes);
    free(pins);
    codes = NULL;
    pins = NULL;
    flag = TRUE;
}
if (v2){
    free (v2);
    v2 = NULL;
} else {
    free (pins);
    free(codes);
    pins = NULL;
    codes = NULL;
    flag = TRUE;
}
if (flag){
    free(flag);
    flag = NULL;
    rtapi_print_msg(RTAPI_MSG_ERR, "sendkeys.N.keycode error\n");
    return -ENOMEM;
}
''''

I think that my problem is that I am not really seeing what is wrong with the original code. A variable "may" be being used unitialized. But it isn't, is it?

dwrobel:
>I think that my problem is that I am not really seeing what is wrong with the original code.
```
              v1 = realloc(codes, (index + 1) * sizeof(int));
              v2 = realloc(pins, (index + 1) * sizeof(int));
              if (!v1 || !v2) {
                    free(codes); // if v1 != NULL then `codes` might be invalid when `realloc` had to move content to `v1`
                    free(pins);  // if v2 != NULL then `pins` might be invalid when `realloc` had to move content to `v2`
               }
```

andypugh:
How about if we just drop v1 and v2? 

pins = realloc(pins...
codes = realloc(codes...
if (! pins || ! codes){
     free / clear / report error

andypugh:
> [andypugh]
> How about if we just drop v1 and v2?
> Then you get a leak if one or both of the reallocs fail.

Even if we detect the realloc fail and free / error out?

petterreinholdtsen:
[andypugh]
> How about if we just drop v1 and v2?

[Petter Reinholdtsen]
> Then you get a leak if one or both of the reallocs fail.

[andypugh]
> Even if we detect the realloc fail and free / error out?

Probably not.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
I suspect something like this is needed:

diff --git a/src/hal/user_comps/sendkeys.c b/src/hal/user_comps/sendkeys.c
index ac73c28cc6..215b00c736 100644
--- a/src/hal/user_comps/sendkeys.c
+++ b/src/hal/user_comps/sendkeys.c
@@ -107,8 +107,8 @@ int init(int argc, char* argv[]){
                     v1 = realloc(codes, (index + 1) * sizeof(int));
                     v2 = realloc(pins, (index + 1) * sizeof(int));
                     if (!v1 || !v2) {
-                        free(codes);
-                        free(pins);
+                        if (v1) free(v1); else free(codes);
+                        if (v2) free(v2); else free(pins);
                         rtapi_print_msg(RTAPI_MSG_ERR, "sendkeys.N.keycode error\n");
                         return -ENOMEM;
                     } else {
-- 
Happy hacking
Petter Reinholdtsen

dwrobel:
> I suspect something like this is needed:
```
>-                        free(codes);
>-                        free(pins);
>+                        if (v1) free(v1); else free(codes);
>+                        if (v2) free(v2); else free(pins);
```

IMHO the current version of the PR has the same logic and is a little more compact.

petterreinholdtsen:
[Damian Wrobel]
> IMHO the current version of the PR has the same logic and is a little
> more compact.

I agree.  I saw it after I sent in my sketch.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
Note, your proposed code is not the version I had in mind.  The flag
variable do not seem to be needed.

[andypugh]
> I think that my problem is that I am not really seeing what is wrong
> with the original code. A variable "may" be being used
> unitialized. But it isn't, is it?

The problem is there.  realloc(3) will tell you that the input pointer
will be invalid after a new pointer is returned, so free() on either
codes and pins both might be operating on an invalid pointer if one of
the realloc() calls succeeded.

-- 
Happy hacking
Petter Reinholdtsen

PR #1963, opened 2022-08-23
Simplify debian/configure by moving constant values to where they are used

petterreinholdtsen:
This reduce the work done by debian/configure and make it easier to see
exactly which settings are configured.


smoe:
Maybe I was not clear enough in my previous comment: I think this is a fine contribution and if I am not missing something then this should be accepted.

andypugh:
This seems to commit us to never supporting kernel-mode RTAI again. 
Are we sure we want to do that?

SebKuzminsky:
I am opposed to this PR, at least for now.  I still maintain hope that we can reintroduce kernel-mode RTAI support in the future.

If @jepler's uspace lxrt work proves to be as performant as kernel-mode RTAI on a wide and representative range of hardware, I may soften my stance.

petterreinholdtsen:
OK.  It was just a stopgap change before updating to Debian profile
based build like proposed in #1835.  When using profiles, we drop
debian/configure completely.

I believe it is a waste to have lots of unused variables in
debian/configure, and believe that given the fact that everything is
visible in the git history, those fixing RTAI can reintroduce these
variables any any time in the future, but se no point in using more time
on this cleanup.

petterreinholdtsen:
[Sebastian Kuzminsky]
> I am opposed to this PR, at least for now.  I still maintain hope that
> we can reintroduce kernel-mode RTAI support in the future.

OK.  It was just a stopgap change before updating to Debian profile
based build like proposed in #1835.  When using profiles, we drop
debian/configure completely.

I believe it is a waste to have lots of unused variables in
debian/configure, and believe that given the fact that everything is
visible in the git history, those fixing RTAI can reintroduce these
variables any any time in the future, but se no point in using more time
on this cleanup.

-- 
Happy hacking
Petter Reinholdtsen

PR #1970, opened 2022-08-28
Wurzel docs small changes

smoe:
Mostly adding dots and some case-sensitivity.

@hansu, I had a weird conflict with your changes on the GMOCCAPY dialog pins. Please cross-check that I have addressed that correctly.


hansu:
> @hansu, I had a weird conflict with your changes on the GMOCCAPY dialog pins. Please cross-check that I have addressed that correctly.

Looks good :)

PR #1976, opened 2022-08-30
pncconf: add inm and outm support plus 7i96s card

phillc54:
Add pncconf support for 7i96s for the outm_for_2.8 branch


andypugh:
Thanks for this. Sorry for the delay.

PR #1530, opened 2022-01-24
Added first skeleton of a coding style for asciidoc

smoe:
First shot to address https://github.com/LinuxCNC/linuxcnc/issues/1529

I did not want to separate it from the coding instructions since we regular code and documentation share weblate, some common features like tabs/indenting and the documentation is also a regular part of the source tree.
I forgot about @silopolis' very valid points on the integration of the glossary.

PR #1977, opened 2022-08-30
docs: intensive reformatting of gstate.adoc

smoe:
Quite an intense reformatting
 - removed explicit bold face in descriptions
 - arguments became part of left side of :: separator, other than return values
 - fixed indentation, removed empty lines
 - consistency in spelling

### DIFF
--- a/docs/src/gui/gstat.adoc
+++ b/docs/src/gui/gstat.adoc
@@ -6,7 +6,7 @@
 
 == Intro
 
-GStat is a python class used to send messages from LinuxCNC to other python programs.
+GStat is a Python class used to send messages from LinuxCNC to other Python programs.
 It uses GObject to deliver messages, making it easy to listen for specific information.
 This is referred to as event-driven programming, which is more efficient then every program
 polling LinuxCNC at the same time.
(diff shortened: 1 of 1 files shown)


hansu:
I think we should finish the Style Guide before doing such reformatting.

petterreinholdtsen:
[Hans Unzner]
> I think we should finish the Style Guide before doing such
> reformatting.

I suspect such reformatting is a good exercise for knowing what to write
in the style guide.

But I would very much like to see the style guide move forward.  If
someone know how to get there, please update pull request #1530.
-- 
Happy hacking
Petter Reinholdtsen

hansu:
>I suspect such reformatting is a good exercise for knowing what to write
in the style guide.

First I like the current formatting of the GStat messages overview.  And second I wouldn't change the formatting of such a central element just so spontaneous.

And it doesn't make sense to change something before you have a recipe how to change it.

I am also holding back some changes until we have decided how we should format pin lists and lists like this.

hansu:
I gave it a try and build it. Here the comparison:
Current version:
Screenshot_2022-08-30_19-51-30

This PR:
Screenshot_2022-08-30_19-51-41


--> So IMHO it's a change to the worse. Less clear and consumes more vertical space.

smoe:
You are right. This is worse. I am not sure why I change the tailing "-" behind those (return..)s. I'll reintroduce the "-"s. I typically just have problems with too many "+"s.

I am with Petter that we should know what to work for, style-guide wise.

smoe:
Should look good again. What got me triggered was that the () syntax is used inconsistently for return values and arguments. And then I just did not like all those "*..*"s when the asciidoc "::" syntax does just that. Still, I hope that we come up with something better.

smoe:
Undid almost all my changes, left a few simplistic ones.

hansu:
I thought we finish first the style guide for this, but ok, let's merge this first.

Issue #1954, opened 2022-08-22
incorrect manpage grouping in html doc index


jepler:
I noticed that rather than being grouped under the headings "API calls" everything has been moved down from the correct sections to "API calls: general". http://linuxcnc.org/docs/devel/html/#man_3hal

hansu:
I found the line causing this: https://github.com/LinuxCNC/linuxcnc/commit/168289d3fed81377f218e98048bcb30be5d3d4b5#r82354641.
Maybe @andypugh knows best how to fix this while keeping his improvement.

andypugh:
Sorry, can't really help. If I ever understood this it was only for long enough to do what I wanted to do.

rmu75:
To me it looks like API stuff is proberly grouped in the documentation section. Is there (still) something to do?

PR #1980, opened 2022-08-31
Fix generating HTML index of manpages

jepler:
This restores the 3hal, 3rtapi, and 3hm2 pages to their proper locations

Closes: #1954 

@hansu thanks for getting me in the neighborhood of the cause

### DIFF
--- a/docs/src/Submakefile
+++ b/docs/src/Submakefile
@@ -478,36 +478,35 @@ SE := \" style=\"-moz-column-width: 25ex; \
 #     * the manpage section title
 #
 ADD_HTML_MANPAGES = \
-	@echo "Adding manpages: man$(strip $(1)): $(strip $(2))"; \
-	echo "$(SA)$(strip $(1))$(SB)$(strip $(1))$(SC)$(strip $(2))</a></p>" >> objects/index.incl; \
-	echo "$(SD)$(strip $(1))$(SE)" >> objects/index.incl; \
+	echo "Adding manpages: man$(strip $(2)): $(strip $(3))"; \
+	echo "$(SA)$(strip $(2))$(SB)$(strip $(2))$(SC)$(strip $(3))</a></p>" >> objects/index.incl; \
+	echo "$(SD)$(strip $(2))$(SE)" >> objects/index.incl; \
 	echo "<ul>" >> objects/index.incl; \
-	for HTML_FILE in $(filter $(DOC_DIR)/html/man/man$(strip $(1))/%.html, $(MAN_HTML_TARGETS)); do \
+	for HTML_FILE in $(filter $(DOC_DIR)/html/man/man$(strip $(1))/%.$(strip $(2)).html, $(MAN_HTML_TARGETS)); do \
 		BASENAME=$$(basename $$HTML_FILE .html); \
 		echo "<li><a href=\"$${HTML_FILE\#$(DOC_DIR)/html/}\">$${BASENAME%.*}</a></li>"; \
 	done >> objects/index.incl; \
-	echo "</ul></div>" >> objects/index.incl
+	echo "</ul></div>" >> objects/index.incl; \
 
 
 objects/index.incl: $(GENERATED_MANPAGES) objects/var-MAN_HTML_TARGETS $(DOC_SRCDIR)/Submakefile
 	rm -f $@
-	$(call ADD_HTML_MANPAGES, 1, Commands and userspace components)
-	$(call ADD_HTML_MANPAGES, 9, Realtime components and kernel modules)
-	$(call ADD_HTML_MANPAGES, 3hal, API calls: HAL)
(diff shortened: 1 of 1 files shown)


docs/src/Submakefile
```diff
@@ -478,36 +478,35 @@ SE := \" style=\"-moz-column-width: 25ex; \
...
-	for F in $$(find $(DOC_DIR)/man/? -type f); do \
+	for F in $$(find $(DOC_DIR)/man/man* -type f); do \
 		B=$$(basename $$F); \
 		if ! grep -q $$B $@; then \
 			echo stray manpage not added to index: $$B; \
 			FAIL=1; \
 		fi; \
 	done; \
-	if [ $$FAIL -ne 0 ]; then false; fi
+	if [ $$FAIL -ne 0 ]; then exit 1; fi
```
petterreinholdtsen, line 509: This change look good to me, and I suspect it is working as intended.  But I wonder if the for loop perhaps would be more specific if it used $(DOC_DIR)/man/man? instead of $(DOC_DIR)/man/man*, and just did 'exit 1' inside the for loop instead of the FAIL=1 stuff.  Why do you want to process all man page paths before exiting?

hansu: @petterreinholdtsen your example seems to be accidentally formatted like Latex or math or so. Can you fix this to make your comment more clear?

petterreinholdtsen: Sorry, did not notice this.  Fixed.

hansu: @jepler what do you think? Would be nice to get this merged.

jepler: Suppose you mess up 37 things. Without FAIL= you have to make 37 times and fix one thing each time. However, please do whatever you care to do.

jepler: Anyway I didn't introduce FAIL= I just corrected its use so that it causes the overall make process to fail.

SebKuzminsky:
This problem affects 2.8 and newer.  Please rebase this to 2.8 and let's merge it.

hansu:
I would cherry-pick Jeffs commit and push it to 2.8 and then merge it up to 2.9 and so on.
Because I don't really want to force push Jeffs branch and rebasing to 2.8 is a lot more work than just cherry-picking it.
Everyone ok with this? 
I did that on my fork and it works (https://github.com/hansu/linuxcnc/commit/252cbadc4f987d27f974100279853d35cba9423c).

c-morley:
Is there going to be another 2.8 release?



Sent from my Galaxy



From: Hans Unzner ***@***.***>
Date: 2023-06-19 11:16 a.m. (GMT-08:00)
Cc: Subscribed ***@***.***>
Subject: Re: [LinuxCNC/linuxcnc] Fix generating HTML index of manpages (PR #1980)


I would cherry-pick Jeffs commit and push it to 2.8 and then merge it up to 2.9 and so on.
Because I don't really want to force push Jeffs branch and rebasing to 2.8 is a lot more work than just cherry-picking it.
Everyone ok with this?
I did that on my fork and it works ***@***.***<https://github.com/hansu/linuxcnc/commit/252cbadc4f987d27f974100279853d35cba9423c>).

SebKuzminsky:
> I would cherry-pick Jeffs commit and push it to 2.8 and then merge it up to 2.9 and so on.
> Because I don't really want to force push Jeffs branch and rebasing to 2.8 is a lot more work than just cherry-picking it.
> Everyone ok with this?
> I did that on my fork and it works ([hansu@252cbad](https://github.com/hansu/linuxcnc/commit/252cbadc4f987d27f974100279853d35cba9423c)).

Thanks for picking this up @hansu, I like what you did and I approve of getting this fix into 2.8 and 2.9 as well, via your method.

Tangentially, rebasing is exactly the same amount of work as cherry-picking: rebase is really only a wrapper around the cherry-pick primitive, with a few extra optional features thrown in.

hansu:
>Is there going to be another 2.8 release?

I don't know, but I think it would be good if we make a final release for 2.8 when 2.9 is released.

>Tangentially, rebasing is exactly the same amount of work as cherry-picking: rebase is really only a wrapper around the cherry-pick primitive, with a few extra optional features thrown in.

But rebasing takes also care about files of commits in between, which produces conflicts in files that I don't want to touch.

hansu:
Fixed with 252cbadc4f987d27f974100279853d35cba9423c and with the subsequent upward-merges to 2.9 and master.

SebKuzminsky:
But all the commits in this branch (there's only one) should be included in the rebase, so it'd be super straightforward.

If it was a longer branch and some of its commits should be skipped, "interactive" rebase makes that straightforward too

hansu:
>But all the commits in this branch (there's only one) should be included in the rebase, so it'd be super straightforward.

I don't think we want it super straightforward in this case, because then I have to deal with merge conflicts in 
VERSION, changelog, tc.c, tp.c ... which I don't want to touch at all

PR #1981, opened 2022-08-31
Docs HAL twopass

smoe:
Small bits.

### DIFF
--- a/docs/src/ladder/classic-ladder.adoc
+++ b/docs/src/ladder/classic-ladder.adoc
@@ -605,20 +605,21 @@ digit number eg. %Q3, or %B123. Q coils control HAL out pins, e.g. if
 %Q15 is energized then HAL pin classicladder.0.out-15 will be true. B
 coils are internal coils used to control program flow.
 
-* 'N.O. COIL' - (a relay coil.) When coil is energized it's N.O. contact
+* 'N.O. COIL' - (a relay coil.) When coil is energized its N.O. contact
   will be closed (on, true, etc)
 * 'N.C. COIL' - (a relay coil that inverses its contacts.) When coil is
-  energized it"s N.O. contact will be open (off, false, etc)
+  energized then its N.C. contact will be open (off, false, etc)
 * 'SET COIL' - (a relay coil with latching contacts) When coil is energized
-  it's N.O. contact will be latched closed.
+  then its N.O. contact will be latched closed.
 * 'RESET COIL' - (a relay coil with latching contacts) When coil is
-  energized It's N.0. contact will be latched open.
-* 'JUMP COIL' - (a 'goto' coil) when coil is energized ladder program jumps
-  to a rung (in the CURRENT section) -jump points are designated by a
-  rung label. (Add rung labels in the section display, top left label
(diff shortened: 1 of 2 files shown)


docs/src/ladder/classic-ladder.adoc
```diff
@@ -605,20 +605,21 @@ digit number eg. %Q3, or %B123. Q coils control HAL out pins, e.g. if
 %Q15 is energized then HAL pin classicladder.0.out-15 will be true. B
 coils are internal coils used to control program flow.
 
-* 'N.O. COIL' - (a relay coil.) When coil is energized it's N.O. contact
+* 'N.O. COIL' - (a relay coil.) When coil is energized its N.O. contact
```
hansu, line 608: Isn't the original one the correct one? It is --> it's ?

tlight9: "it's" would be the correct use

petterreinholdtsen:
What about expanding "it's" to "it is", thus getting rid of the doubt as
well as making the text easier to read for those that are unsure about
these contractions.

-- 
Happy hacking
Petter Reinholdtsen

smoe:
I had read it as "When the coil is energized , then its normally open (N.O.) contact is closed and current can pass." ? The "it is" variant I consider unfortunate since in my understanding when the coil is energized, it is no longer open and the sentence would then be confusing.

hansu:
> I had read it as "When the coil is energized , then its normally open (N.O.) contact is closed and current can pass." ? The "it is" variant I consider unfortunate since in my understanding when the coil is energized, it is no longer open and the sentence would then be confusing.

I agree. But still a bit misleading as I would understand "N.O. contact" as a noun. I like your suggestion "its normally open (N.O.) contact"

smoe:
The normally open contact was meant as a noun as in "The coil has a contact that is normally open." I only made that minimal change to avoid discussions - which almost worked :) A mere comma before the "its" (pronoun) may help the parsing. Shall I add that? I can also craft something de novo.

hansu:
Haha yes, sure. I was thinking of "its normally open contact", so its contact, which is normally open 😄

smoe:
Somehow  I sense we need a word for something that is worse than [https://en.wikipedia.org/wiki/Law_of_triviality](bike-shedding). We have
 * some sub-optimal wording in English
 * many non-native speakers that attempt to improve it
 * nobody ultimately feeling responsible
 * everybody feeling sufficiently competent to write those sentences

just that other than bike-shedding it makes quite a difference if we get this wrong or if the reader understands this the wrong way.

hansu:
Yes the bike-shed effect describes is quite well :D

c-morley:
classickader: 'It's N.O.' was intended as possessive:
 When the coils is energized, the (N)ormally (O)pen contact of the that coil, switches from open to closed.

c-morley:
'current will pass' i think would be better as 'signal will pass'
While current makes sense in a real relay, not so in a software relay.
You could argue it makes sense in context of the visual representation, but then I would add quotes around it to signify that it isn't really current in the real world.

PR #1984, opened 2022-09-01
Translation-motivated changes on touchy.adoc

smoe:
Typical ones for consistency.

A few axis -> axes changes of gmoccapy sneaked in.

### DIFF
--- a/docs/src/gui/touchy.adoc
+++ b/docs/src/gui/touchy.adoc
@@ -43,23 +43,23 @@ For more information on HAL files and the net command see the
 Touchy has several output pins that are meant to be connected to the
 motion controller to control wheel jogging:
 
-- 'touchy.jog.wheel.increment', which is to be connected to the 'axis.N.jog-scale' pin of each axis N.
-- 'touchy.jog.wheel.N', which is to be connected to 'axis.N.jog-enable' for each axis N.
+- 'touchy.jog.wheel.increment', which is to be connected to the 'axis._N_.jog-scale' pin of each axis _N_.
+- 'touchy.jog.wheel._N_', which is to be connected to 'axis._N_.jog-enable' for each axis _N_.
 
 [NOTE]
 'N' represents the axis number 0-8.
 
 - In addition to being connected to 'touchy.wheel-counts', the wheel counts
-  should also be connected to 'axis.N.jog-counts' for
-  each axis N.  If you use HAL component 'ilowpass' to smooth wheel jogging, be
-  sure to smooth only 'axis.N.jog-counts' and not 'touchy.wheel-counts'.
+  should also be connected to 'axis._N_.jog-counts' for
+  each axis _N_.  If you use HAL component 'ilowpass' to smooth wheel jogging, be
+  sure to smooth only 'axis._N_.jog-counts' and not 'touchy.wheel-counts'.
 
 .Required controls
 
-- Abort button (momentary contact) connected to the HAL pin 'touchy.abort'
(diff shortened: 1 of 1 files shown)


docs/src/gui/touchy.adoc
```diff
@@ -50,16 +50,16 @@ motion controller to control wheel jogging:
 'N' represents the axis number 0-8.
 
 - In addition to being connected to 'touchy.wheel-counts', the wheel counts
-  should also be connected to 'axis.N.jog-counts' for
-  each axis N.  If you use HAL component 'ilowpass' to smooth wheel jogging, be
-  sure to smooth only 'axis.N.jog-counts' and not 'touchy.wheel-counts'.
+  should also be connected to 'axis._n_.jog-counts' for
```
hansu, line 53: In the pin descriptions of the man pages, the pins are always (as far as I can see) formatted like this:

axis._N_.jog-counts

Maybe we should do it here as well for consistency.
(Another point for the Style Guide...)

smoe:
The capital N was new for me. @silopolis , please add this to your list of things to decide about.

The italic presentation is perceived well?

docs/src/gui/touchy.adoc
smoe: You are right. Have converted back to upper-case Ns, slanted.

hansu:
Seems like we already agreed on a capital italic N :D

smoe:
> Seems like we already agreed on a capital italic N :D

Other Ns are also capitalized, other variables like loopnum etc are in lowercase. We still need the style guide, but not for this PR to get accepted.

hansu:
Yes. And I don't really like the sentence end point for those short listitems. But only my opinion.

smoe:
We can also discuss those endpoints. What typicall happens is that the automated translation adds a period and then I have to go for an extra click and extra latency to either dismiss it or remove that dot. And typically some list item that is not the last one already ends with a period, so it is inconsistent and I get triggered.

@silopolis , not (or barely) negotiable for the style guide are:

* The itemization/enumeration does not disturb the reading but supports it.
* Without the leading dots the paragraph is still readable and gramatically correct.
* If a colon introduces the itemization/enumeration then the first line starts with a capital.
* If periods, commas, and/or conjuctions are used to terminate the item depends on the nature of what shall be communicated.

There is little motivation on my side to allow for an enumeration/itemization that has only a single item. We have quite a few of those in the documentation that I have not addressed but I was very much tempted.

silopolis:
Le ven. 2 sept. 2022 à 15:59, Steffen Möller ***@***.***> a
écrit :

> We can also discuss those endpoints. What typicall happens is that the
> automated translation adds a period and then I have to go for an extra
> click and extra latency to either dismiss it or remove that dot. And
> typically some list item that is not the last one already ends with a
> period, so it is inconsistent and I get triggered.
>
> @silopolis <https://github.com/silopolis> , not (or barely) negotiable
> for the style guide are:
>
>    - The itemization/enumeration does not disturb the reading but
>    supports it.
>    - Without the leading dots the paragraph is still readable and
>    gramatically correct.
>    - If a colon introduces the itemization/enumeration then the first
>    line starts with a capital.
>    - If periods, commas, and/or conjuctions are used to terminate the
>    item depends on the nature of what shall be communicated.
>
> There is little motivation on my side to allow for an
> enumeration/itemization that has only a single item. We have quite a few of
> those in the documentation that I have not addressed but I was very much
> tempted.
>

Agreed on all items 👍

PR #1983, opened 2022-08-31
Translations update from Hosted Weblate

weblate:
Translations update from [Hosted Weblate](https://hosted.weblate.org) for [LinuxCNC/LinuxCNC Documentation](https://hosted.weblate.org/projects/linuxcnc/linuxcnc-docs/).

Current translation status:

Weblate translation status

PR #1989, opened 2022-09-02
Maybe for the style guide again - Docs qtdragon2

smoe:
These "+"s do not show up in po4a-generated files, so admittedly I am in part working around po4a-associated difficulties. But here, I just think they are too much.

### DIFF
--- a/docs/src/gui/pyvcp-examples.adoc
+++ b/docs/src/gui/pyvcp-examples.adoc
@@ -15,10 +15,10 @@ To create a PyVCP panel to use with the AXIS interface that is
 attached to the right of AXIS you need to do the following basic
 things.
 
-* Create an .xml file that contains your panel description and put it in
+* Create an XML file that contains your panel description and put it in
   your config directory.
-* Add the PyVCP entry to the [DISPLAY] section of the ini file with your .xml file name.
-* Add the POSTGUI_HALFILE entry to the [HAL] section of the ini file
+* Add the PyVCP entry to the [DISPLAY] section of the INI file with your XML file name.
+* Add the POSTGUI_HALFILE entry to the [HAL] section of the INI file
   with the name of your postgui HAL file name.
 * Add the links to HAL pins for your panel in the postgui.hal file to
   'connect' your PyVCP panel to LinuxCNC.
@@ -28,9 +28,9 @@ things.
 To create floating PyVCP panels that can be used with any interface
 you need to do the following basic things.
 
-* Create an .xml file that contains your panel description and put it in
+* Create an XML file that contains your panel description and put it in
   your config directory.
-* Add a loadusr line to your .hal file to load each panel.
+* Add a loadusr line to your HAL file to load each panel.
 * Add the links to HAL pins for your panel in the postgui.hal file to
   'connect' your PyVCP panel to LinuxCNC.
 
@@ -47,9 +47,9 @@ The -Wn makes HAL 'Wait for name' to be loaded before proceeding.
 
 The pyvcp -c makes PyVCP name the panel.
 
-The HAL pins from panel1.xml will be named btnpanel.<pin name>
+The HAL pins from panel1.xml will be named btnpanel.<_pin name_>.
 
-The HAL pins from panel2.xml will be named sppanel.<pin name>
+The HAL pins from panel2.xml will be named sppanel.<_pin name_>.
 
 Make sure the loadusr line is before any nets that make use of the
 PyVCP pins.
@@ -304,7 +304,7 @@ The following figure shows what a complete panel might look like.
 
 image::images/ptest-final.png["Port Tester Complete Image"]
 
-To add the rest of the parallel port pins just modify the .xml and .hal files.
+To add the rest of the parallel port pins just modify the XML and HAL files.
 
 To show the pins after running the HAL script use the following
 command at the halcmd prompt:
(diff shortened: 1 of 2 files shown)


smoe:
Darn, these pyvcp examples should have been in another PR. Anyway...

docs/src/gui/pyvcp-examples.adoc
```diff
@@ -47,9 +47,9 @@ The -Wn makes HAL 'Wait for name' to be loaded before proceeding.
 
 The pyvcp -c makes PyVCP name the panel.
 
-The HAL pins from panel1.xml will be named btnpanel.<pin name>
+The HAL pins from panel1.xml will be named btnpanel.<_pin name_>.
```
hansu, line 50: This doesn't work. Use `btnpanel._<pin name>_` instead.

silopolis:
Le ven. 2 sept. 2022 à 12:34, Steffen Möller ***@***.***> a
écrit :

> These "+"s do not show up in po4a-generated files, so admittedly I am in
> part working around po4a-associated difficulties. But here, I just think
> they are too much.
>

Wasn't the forced line return issue fixed in po4a 0.67?

silopolis:
Le ven. 2 sept. 2022 à 17:35, Steffen Möller ***@***.***> a
écrit :

> Not sure. We now have linebreaks in weblates that we did not have before.
> But we lost the '+'s.
>

As I hardly see how they could be reintroduced afterward, this means the
translated adoc files must be broke and that the issue is omehow still
there :/

Maybe this time it's a Weblate problem? 🤔

petterreinholdtsen:
[Steffen Möller]
> Not sure. We now have linebreaks in weblates that we did not have before.
> But we lost the '+'s.

[Jérémie Tarot]
> As I hardly see how they could be reintroduced afterward, this means the
> translated adoc files must be broke and that the issue is omehow still
> there :/
>
> Maybe this time it's a Weblate problem? 🤔

Nah, it is simply an issue of expectations.  Let me try again to explain
how this (correctly) work:

If the adoc file have "first line +\nsecond line", the POT file get
"first line\nsecond line", the PO file traslate to
"første linje\nandre linje", and the generated translated adoc file get
"første linje +\nandre linje".  Both adoc files are correct, and the
translators do not have to worry about placing SPACE PLUS NEWLINE in the
correct sequence, and only have to place the NEWLINE in the correct spot.

See
<URL: https://github.com/mquinson/po4a/commit/afe0544cfaad8d5d4ac3e6caf0529b69300d4aef >
and previos related patch alongside the automated test to validate the
behaviour.

-- 
Happy hacking
Petter Reinholdtsen

smoe:
As a translator I prefer not to do anything but rely on the automated translation (that button click already annoys me) and do a quick confirmation that everything is translated as it should be ("ah, not that port -> harbor miss-translation again"). RSI and time are prohibiting everything else - no manual typing if at all possible. 
Now, when there are "verbatim" blocks, then
```
this
that
```
gets translated to just that. But the same appearing in another context, i.e. a regular paragraph or a note, means something different, I would have to manually enter the ' +' to transform
```
this
that
```
to
```
this +
that
```
since weblate does not see the +.

For most of our texts this is not problematic since there are no explicit linebreaks and for some texts there are far too many, so I am not too unhappy to see them negatively selected or you get an indication from reading the text where the + should go. But then again - argh!

I am pretty confident that the elimination of the "+" is a bug. Petter thinks differently. And I can live with what we have for now. There is another bug of po4a that removes the last character of table fields, as in a confusion of chomp with chop, but it is not perfectly reproducivle as it seems and I have not had the motivation to chase that up either, which should have a higher priority.

smoe:
I just read @petterreinholdtsen's comment again. I see a loss of semantics if the + goes away. And I do not want to manually do anything. The '+' I found to be treated well by the automated translators, so I wish it would remain in. The newline to consider special white space (if I interpret this right) sounds inconvenient to me. So, for now, I think I will just continue avoiding the '+'s.

petterreinholdtsen:
[Steffen Möller]
> I just read @petterreinholdtsen's comment again. I see a loss of
> semantics if the + goes away. And I do not want to manually do
> anything. The '+' I found to be treated well by the automated
> translators, so I wish it would remain in. The newline to consider
> special white space (if I interpret this right) sounds inconvenient to
> me. So, for now, I think I will just continue avoiding the '+'s.

I suspect either I misunderstand you, or you misunderstand me, as what
you write do not make sense to me.  The newline in adoc strings and a
lot of other formats in the PO is meaningful.  The plus sign (alone) is
not.

The current German translation of the documenation is not buildable (see
#1983), because of incorrect handling of newline.  Weblate have
automatic checks for this, and the German translation currently have 160
strings reported with "Failing check: Mismatching line breaks" in <URL:
https://hosted.weblate.org/projects/linuxcnc/linuxcnc-docs/de/ >,
alongside alot of other failing checks.  There are lots of other failing
checks too.  I strongly suggest to fix every one of them, either by
rewriting the string or by flagging them as ignored, to avoid new
problems sneaking into the translation.

-- 
Happy hacking
Petter Reinholdtsen

docs/src/gui/pyvcp-examples.adoc
hansu: \*dont forget me\*

silopolis:
#1530 Style guide

docs/src/gui/pyvcp-examples.adoc
hansu: Ok then not italic...

PR #1990, opened 2022-09-02
docs: first look at remap.adoc

smoe:
Needs to be merged prior to continue investing into weblate.

### DIFF
--- a/docs/src/remap/remap.adoc
+++ b/docs/src/remap/remap.adoc
@@ -14,8 +14,8 @@
 
 By 'remapping codes' we mean one of the following:
 
-. define the semantics of new - that is, currently unallocated - M- or G-codes
-. redefine the semantics of a - currently limited - set of existing codes.
+. Define the semantics of new - that is, currently unallocated - M- or G-codes
+. Redefine the semantics of a - currently limited - set of existing codes.
 
 === Why would you want to extend the RS274NGC Interpreter?
 
@@ -24,7 +24,7 @@ interpreter is fixed and cannot be extended by configuration options.
 
 In particular, some of these codes implement a fixed sequence of steps
 to be executed. While some of these, like M6, can be moderately
-configured by activating or skipping some of these steps through ini
+configured by activating or skipping some of these steps through INI
 file options, overall the behavior is fairly rigid. So - if your
 are happy with this situation, then this manual section is not for you.
 
@@ -38,7 +38,7 @@ might make their way into the main LinuxCNC distribution, the result of
 this approach is a hodge-podge of special-case solutions.
 
 A good example for this deficiency is tool change support in LinuxCNC:
-while random tool changers are well supported, it is next to impossible
+While random tool changers are well supported, it is next to impossible
 to reasonably define a configuration for a manual-tool change machine
 with, for example, an automatic tool length offset switch being
 visited after a tool change, and offsets set accordingly. Also, while a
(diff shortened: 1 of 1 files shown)


docs/src/remap/remap.adoc
```diff
@@ -2678,15 +2674,15 @@ thing.
...
+RANDOM_TOOLCHANGER INI option was set, but at the end of the process
 'toolTable[0]' reflects the tool currently in the spindle.
 
 When that has happened:
 
-. iocontrol signals task to go ahead
-. task tells the interpreter to execute a 'synch()' operation, to see
-  what has changed
-. the interpreter 'synch()' pulls all information from the world model
+. Iocontrol signals task to go ahead.
```
hansu, line 2682: I/O control

smoe: I think this refers to an iocontrol HAL module. Maybe `iocontrol` and I should just not capialize despite the colon?

smoe: same for `task`.

petterreinholdtsen:
[Steffen Möller]
> I think this refers to an iocontrol HAL module. Maybe `iocontrol` and
> I should just not capialize despite the colon?

What about being explicit, in other words write 'The iocontrol component
signals task to go ahead'?  Then we do not have to guess if it refer to
the iocontrol component.

-- 
Happy hacking
Petter Reinholdtsen

docs/src/remap/remap.adoc
hansu: Sure, I only had the title of https://linuxcnc.org/docs/html/config/iov2.html in mind which says "I/O control".
But yeah, just a suggestion...

hansu:
>What about being explicit, in other words write 'The iocontrol component
signals task to go ahead'? 

:+1: 
Or format it italic oder as code

c-morley:
Why would you remove the information about synch()?
 Is is referring to the interpeter's synch() method (earlier in document:)

The interpreter’s synch() method is called by task and does just that - read all the world model actual values which are relevant for further execution.

When synch() is run, I dare says is an important detail.

smoe:
> Why would you remove the information about synch()? Is is referring to the interpeter's synch() method (earlier in document:)
> 
> The interpreter’s synch() method is called by task and does just that - read all the world model actual values which are relevant for further execution.
> 
> When synch() is run, I dare says is an important detail.

Thank you for spotting that. Synch is mentioned twice in the paragraph, the first time without the ()s, and I did not get that this was important when the autotranslators have just completed it to synchronization which I then also clarified for the English. So - many thanks again. Well done.

I'll do some more reading so that I understand this myself and then come up with something that is autotranslator-proof. ;) Please feel free to beat me to it.

c-morley:
ok, ya that totally fooled me :)
Sorry for the noise.

docs/src/remap/remap.adoc
```diff
@@ -2343,15 +2343,15 @@ FIXME Write missing information
 
 FIXME Write missing information
 
-=== A diagram: task, interp, iocontrol, UI (??)
+=== A diagram: `task`, `interp`, `iocontrol`, `UI` (??)
```
hansu, line 2346: Not sure if we should have monospaced font in headings. Maybe something for the style guide ;-)

silopolis:
Le dim. 4 sept. 2022 à 21:17, Hans Unzner ***@***.***> a
écrit :

>
> In docs/src/remap/remap.adoc
> <https://github.com/LinuxCNC/linuxcnc/pull/1990#discussion_r962357557>:
>
> > @@ -2343,15 +2343,15 @@ FIXME Write missing information
>
>  FIXME Write missing information
>
> -=== A diagram: task, interp, iocontrol, UI (??)
> +=== A diagram: `task`, `interp`, `iocontrol`, `UI` (??)
>
> Not sure if we should have monospaced font in headings. Maybe something
> for the style guide ;-)
>
I believe we do and quite a bunch are coming...

Maybe we could take advantage of GH features and tag messages related to
style guide with PR id #1530 ?!

docs/src/remap/remap.adoc
hansu: #1530

PR #1993, opened 2022-09-03
docs: triggered by translations

smoe:
Nothing fancy.

### DIFF
--- a/docs/src/gui/halui.adoc
+++ b/docs/src/gui/halui.adoc
@@ -1,5 +1,6 @@
 :lang: en
 :toc:
+:example-caption: Example
 
 [[cha:hal-user-interface]]
 = HAL User Interface
@@ -54,20 +55,21 @@ When halui starts it will read the `MDI_COMMAND` fields in the INI and
 export pins from 00 to the number of `MDI_COMMAND` 's found in the INI, up
 to a maximum of 64 commands. These pins can be connected like any
 HAL pins. A common method is to use buttons provided by virtual
-control panels.
+control panels like shown in the example <<code:Example-HAL-file-connections, Example for MDI_COMMAND connections>>.
 
-.Example HAL file connections
+
+[[code:Example-HAL-file-connections]]
+.Example for MDI_COMMAND connections
+====
+.HAL file
 [source,{hal}]
 ----
-net quill-up       <= pyvcp.quillup
-net quill-up       => halui.mdi-command-00
-net reference-pos  <= pyvcp.referencepos
-net reference-pos  => halui.mdi-command-01
-net call-mysub     <= pyvcp.callmysub
-net call-mysub     => halui.mdi-command-02
+net quill-up      halui.mdi-command-00 <= pyvcp.quillup
+net reference-pos halui.mdi-command-01 <= pyvcp.referencepos
+net call-mysub    halui.mdi-command-02 <= pyvcp.callmysub
 ----
-
 .Nets connecting the `halui.mdi-command-NN` pins provided by halui.
+[source,sh]
 ----
 $ halcmd show pin halui.mdi
 Component Pins:
@@ -77,6 +79,7 @@ Owner  Type  Dir Value  Name
    10  bit   IN  FALSE  halui.mdi-command-02 <== call-mysub
     ...
 ----
+====
 
 When a halui MDI pin is set (pulsed) true, halui will send the MDI
 command defined in the INI.
@@ -92,6 +95,7 @@ is included in the distribution.
 
 All halui pins are also documented in the halui man page:
 
+[source,sh]
 ----
 $ man halui
 ----
--- a/docs/src/gui/halui.adoc
+++ b/docs/src/gui/halui.adoc
@@ -14,8 +14,7 @@
 
 Halui is a HAL based user interface for LinuxCNC, it connects HAL pins to
 NML commands. Most of the functionality (buttons, indicators etc.) that
-is provided by a traditional GUI (AXIS, GMoccapy, QtDragon, etc.), is provided by HAL
-pins in Halui.
+is provided by a traditional GUI (AXIS, GMOCCAPY, QtDragon, etc.), is provided by HAL pins in Halui.
 
 The easiest way to add halui is to add the following to the [HAL]
 section of the INI file:
@@ -55,8 +54,9 @@ When halui starts it will read the `MDI_COMMAND` fields in the INI and
 export pins from 00 to the number of `MDI_COMMAND` 's found in the INI, up
 to a maximum of 64 commands. These pins can be connected like any
 HAL pins. A common method is to use buttons provided by virtual
-control panels. Example HAL file connections:
+control panels.
 
+.Example HAL file connections:
 [source,{hal}]
 ----
 net quill-up       <= pyvcp.quillup
@@ -67,9 +67,7 @@ net call-mysub     <= pyvcp.callmysub
 net call-mysub     => halui.mdi-command-02
 ----
 
-These nets connect the `halui.mdi-command-NN` pins
-provided by halui:
-
+.These nets connect the `halui.mdi-command-NN` pins provided by halui:
 ----
 $ halcmd show pin halui.mdi
 Component Pins:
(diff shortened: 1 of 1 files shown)


docs/src/gui/halui.adoc
```diff
@@ -55,8 +54,9 @@ When halui starts it will read the `MDI_COMMAND` fields in the INI and
 export pins from 00 to the number of `MDI_COMMAND` 's found in the INI, up
 to a maximum of 64 commands. These pins can be connected like any
 HAL pins. A common method is to use buttons provided by virtual
-control panels. Example HAL file connections:
+control panels.
 
+.Example HAL file connections:
```
hansu, line 59:
```suggestion
.Example HAL file connections
```

docs/src/gui/halui.adoc
```diff
@@ -67,9 +67,7 @@ net call-mysub     <= pyvcp.callmysub
 net call-mysub     => halui.mdi-command-02
 ----
 
-These nets connect the `halui.mdi-command-NN` pins
-provided by halui:
-
+.These nets connect the `halui.mdi-command-NN` pins provided by halui:
```
hansu, line 70:
```suggestion
These nets connect the `halui.mdi-command-NN` pins provided by halui:
```
This wording doesn't fit as a caption

smoe: It should be "Nets connecting the `halui.mdi-command-NN` pins provided by halui."

hansu: I mean we don't have to force using captions for small code snippets that are integrated in the text flow ...

smoe: You use the caption when the image is not deeply embedded in the text.  I had introduced those captions for the many [underline] instructions that basically created their own header styles only to then have just that single image as a content. That image could be floating anywhere. So I prefer to have a good caption for it and focus on the main parts of the article.

I am not completely sure what the context of this block now is about. But anything that has the character of a reference   or that is an example that is not discussed in the main text, well, that to me is something that should get captions and just live somewhere happily.

What makes me shy away from captions again is that asciidoc cannot reference those items properly. It is not LaTeX after all. But here, if it does not look bad, just go for it, I suggest. That "Table xx"-prefix can be switched off if you find that disturbing - I do not like those too much, I must admit.

hansu: If not deeply embedded in the text, the example/image/table/etc. should have a caption, I agree,  but then you should reference it in the describing text as well.

Like

>These pins can be connected like any HAL pins. A common method is to use buttons provided by virtual control panels like shown in Example 1.
>
> Example 1: Example HAL file connections
>```
>net quill-up       <= pyvcp.quillup
>net quill-up       => halui.mdi-command-00
>net reference-pos  <= pyvcp.referencepos
>net reference-pos  => halui.mdi-command-01
>net call-mysub     <= pyvcp.callmysub
>net call-mysub     => halui.mdi-command-02
>```

But this is not possible with our Asciidoc version. Up from 1.5.6 it should be possible maybe (https://github.com/asciidoctor/asciidoctor/issues/858)

Currently we can have it only like this:

image


Otherwise I would prefer:

>These pins can be connected like any HAL pins. A common method is to use buttons provided by virtual control panels like:
>
>```
>net quill-up       <= pyvcp.quillup
>net quill-up       => halui.mdi-command-00
>net reference-pos  <= pyvcp.referencepos
>net reference-pos  => halui.mdi-command-01
>net call-mysub     <= pyvcp.callmysub
>net call-mysub     => halui.mdi-command-02
>```

smoe: I think we are in sync.

How to proceed? Yet another build? I would otherwise just ask you to amend to your liking at commit.

smoe: Push, I mean.

smoe: Commit and Push.

hansu: So do you agree to my workaround?

hansu: What about this?
image

docs/src/gui/halui.adoc
```diff
@@ -165,28 +163,28 @@ that is set in the [TRAJ] section of the INI file.
...
-* 'halui.joint.N.has-fault' (bit out) - status pin telling that joint N has a fault
-* 'halui.joint.N.home' (bit in) - pin for homing joint N
-* 'halui.joint.N.is-homed' (bit out) - status pin telling that joint N is homed
-* 'halui.joint.N.on-hard-max-limit' (bit out) - status pin telling that joint N  is  on  the  positive  hardware limit
-* 'halui.joint.N.on-hard-min-limit' (bit out) - status  pin  telling  that  joint  N is on the negative hardware limit
-* 'halui.joint.N.on-soft-max-limit' (bit out) - status pin telling that joint N  is  on  the  positive  software limit
-* 'halui.joint.N.on-soft-min-limit' (bit out) - status  pin  telling  that  joint  N is on the negative software limit
-* 'halui.joint.N.override-limits' (bit out) - status pin telling that joint N's limits are  temporarily  overridden
-* 'halui.joint.N.unhome' (bit in) - pin for unhoming joint N
+* 'halui.joint._N_.select' (bit in) - pin for selecting joint _N_
```
hansu, line 177: Furthermore this changes does absolutely no effect because `' '` already makes it italic.

smoe: Nope. https://docs.asciidoctor.org/asciidoc/latest/text/italic/

smoe: 'not italic', __italic_\_

smoe: And yes, I would like to remove 90% of the "'"s in the documentation.

smoe: Uh, seems like I just learned something - https://linuxcnc.org/docs/devel/html/gui/halui.html shows these lines as italic, indeed. Hm.

hansu: So you are not building the docs to verify your changes?

smoe: I look at them in the previewer.  The .N. -> ._N_. change was semi-automated.

smoe: There is something else to it. The _N\_ indicates that this N is variable, since being slanted is often seen to indicate such variability. I would not want to take those semantics away, even if invisible. If the outer highlighting would be bold, this would have worked just fine. Also, slanting something within something slanted could "undo" the slanting, which then depends on some future asciidoc rendering. So, I suggest to keep those inner _N\_s.

silopolis:
Le mar. 6 sept. 2022 à 12:13, Steffen Möller ***@***.***> a
écrit :

> Nope. https://docs.asciidoctor.org/asciidoc/latest/text/italic/
>

Beware that we don't use asciidoctor per se so there may be differences in
implementation and rendering

BTW, as asciidoctor has become the reference implementation supporting the
development of the asciidoc standard to be released, opportunity to
consider switching to it could be discussed...

hansu:
> Le mar. 6 sept. 2022 à 12:13, Steffen Möller ***@***.***> a écrit :
> Nope. https://docs.asciidoctor.org/asciidoc/latest/text/italic/
> Beware that we don't use asciidoctor per se so there may be differences in implementation and rendering BTW, as asciidoctor has become the reference implementation supporting the development of the asciidoc standard to be released, opportunity to consider switching to it could be discussed...

As we are already using most of the Asciidoctor syntax, it is worth a thought:
https://docs.asciidoctor.org/asciidoctor/latest/migrate/asciidoc-py/

hansu:
It's more a long term thing...

silopolis:
Le mar. 6 sept. 2022 à 13:42, Steffen Möller ***@***.***> a
écrit :

> There is something else to it. The _N_ indicates that this N is variable,
> since being slanted is often seen to indicate such variability. I would not
> want to take those semantics away, even if invisible.
>

Markup semantics first, style next... I like that! 👍👌

silopolis:
Le lun. 5 sept. 2022 à 15:56, Hans Unzner ***@***.***> a
écrit :

>
> In docs/src/gui/halui.adoc
> <https://github.com/LinuxCNC/linuxcnc/pull/1993#discussion_r962934249>:
>
> > @@ -67,9 +67,7 @@ net call-mysub     <= pyvcp.callmysub
>  net call-mysub     => halui.mdi-command-02
>  ----
>
> -These nets connect the `halui.mdi-command-NN` pins
> -provided by halui:
> -
> +.These nets connect the `halui.mdi-command-NN` pins provided by halui:
>
> I mean we don't have to force using captions for small code snippets that
> are integrated in the text flow ...
>
We could, except for trivial ones of course. This would sustain overall
quality by offering a short description, just as for images, tables...
Later, this could allow building appendices tables of those special content
blocks

PR #1995, opened 2022-09-03
docs: translation-motivated changes

smoe:
Nothing too special, but I had one fight with one of those man pages that sneaked into the main documentation (which I think is wrong in the first place). I like how this looks but we need to find another way to integrate those in the main documentation.


petterreinholdtsen:
The build error is this: asciidoc: ERROR: tools.adoc: line 244: illegal style name: width="90%"

smoe:
Rebased to master and removed the grid+frame table parameters that likely caused the build failure - need to investigate, this once worked IIRC.

PR #2002, opened 2022-09-05
hal: fix typo in variable name

luzpaz:
substitute 'threshold' for 'threshhold'


andypugh:
I propose that as an interim measure we support both pin names, and document that the pin name will change in a future release. (Basically when there are enough other changes to make a new INI version and update to update_ini worthwhile)

PR #2003, opened 2022-09-06
docs: fixed two weblate warnings

smoe:
Do not think this fixes the weblate error, but the warnings from https://github.com/LinuxCNC/linuxcnc/runs/8171898278?check_suite_focus=true should disappear.

Also introduced two descriptions to introduce clarity / remove unneeded subsections.


petterreinholdtsen:
Can you fix the conflict making it possible to merge this?  The change look good to me.

smoe:
@petterreinholdtsen - rebased to master and conflict resolved.

PR #2005, opened 2022-09-06
docs: man asciidoc - series of small changes

smoe:
Found an "ini" via weblate and then gave the asciidoc files in src/man/man1 a review.

### DIFF
--- a/docs/man/man1/gmoccapy.1
+++ b/docs/man/man1/gmoccapy.1
@@ -27,14 +27,14 @@
 gmoccapy \- TOUCHY LinuxCNC Graphical User Interface
 .SH SYNOPSIS
 .B gmoccapy
-\fI\-ini\fR \fIINIFILE\fR
+\fI\-ini\fR \fI<INI file>\fR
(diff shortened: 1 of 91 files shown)


smoe:
I have now manually screened all man pages for the typical culprits - ini>INI, hal>HAL, etc. This should render these ready for weblate without too many extra iterations.

SebKuzminsky:
I admit I only spot-checked the changes here, but I trust you got it all right.  Thanks!

docs/man/man1/gmoccapy.1
```diff
@@ -27,14 +27,14 @@
 gmoccapy \- TOUCHY LinuxCNC Graphical User Interface
 .SH SYNOPSIS
 .B gmoccapy
-\fI\-ini\fR \fIINIFILE\fR
+\fI\-ini\fR \fI<INI file>\fR
```
hansu, line 30: Still some space for optimization left :D 
I have no idea why this is shown like this:
grafik
(also in the other occurrences)

PR #2007, opened 2022-09-06
CI harder

SebKuzminsky:
This PR improves our Github CI:
+ Build packages and test the packaged executables on Debian Unstable and Debian Testing, not on Ubuntu 20.04.
+ Build architecture-dependent and architecture-independent packages separately (like debian's build infrastructure does).

Testing on debian unstable revealed a build system bug, which is also fixed in this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
         eatmydata make -O -j$((1+$(nproc))) docs
         # Note that the package build covers html docs
 
-  package:
+  package-arch:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


petterreinholdtsen:
If all tests passes, I believe this should be good to go.  Personally, I would use 'apt' instead of 'apt-get'.  I notice the running of the runtests scripts is changed to run as a non-root user.  While it seem like a good idea, it is unclear from the commit headers exactly why this is done.

petterreinholdtsen (approved):

SebKuzminsky:
> I notice the running of the runtests scripts is changed to run as a non-root user. While it seem like a good idea, it is unclear from the commit headers exactly why this is done.

The tests were run as a regular non-root user before too.  It's a bit unclear from the workflow file, but this PR switches the execution environment from github's "run as a user in an Ubuntu 20.04 VM" to "run as root in a bunch of different docker containers".  LinuxCNC *really* doesn't like you to run it as root, so the acrobatics to avoid being root in docker are necessary.

jepler:
Yay, thank you

PR #2008, opened 2022-09-06
GStat linuxcnc->LinuxCNC and other smaller bits.

smoe:
Is there another open PR of mine that addresses GStat? We need to get the layout (mostly) right before I start with its translations. This has loads of definitions again, as known from qtvcp-*.adoc . Where are we with it?


smoe:
Yes, it was https://github.com/LinuxCNC/linuxcnc/pull/1977 - but here everything should be truly straight-forward - I just cannot do all those linuxnc -> LinuxCNC changes in weblate that easily.

PR #2009, opened 2022-09-07
docs: Weblate punctuation checks

smoe:
Went through the "any checks" warnings of weblate, which turned out to be mostly changes to the punctuation that have not been committed.

### DIFF
--- a/docs/man/man9/setsserial.9
+++ b/docs/man/man9/setsserial.9
@@ -54,7 +54,7 @@ value.
 The next line invokes setsserial. This is run in a slightly strange way in order
 to have kernel-level access to a live Hostmot2 config. It is basically a 
 HAL module that always fails to load. This may lead to error messages being
-printed to the halcmd prompt. These can often be ignored. 
+printed to the `halcmd:` prompt. These can often be ignored. 
 All the real feedback is via the dmesg command. It is suggested to have a second
 terminal window open to run dmesg after each command. 
 
--- a/docs/src/getting-started/updating-linuxcnc.adoc
+++ b/docs/src/getting-started/updating-linuxcnc.adoc
@@ -997,7 +997,7 @@ ShuttleXpress, so the old driver name is no longer accurate.
 === linuxcncrsh
 
 "Home All" is now supported with the set home subcommand
-by using -1 for the joint number
+by using -1 for the joint number.
 
 The jogging commands have been altered to accommodate both joint (free)
 and teleop (world) jogging.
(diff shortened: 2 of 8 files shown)


petterreinholdtsen:
This one look good to me.  The only think I am unsure about is the &thinsp; stuff, not sure what the conclusion was there.  I assume it will not cause any problems.

hansu:
[Petter]
You mean the `&thinsp;` stuff?

I am also unsure because these characters generate a lot of warnings. Not sure how we could get rid of warnings like this:
```
../docs/html/tooldatabase/tooldatabase.html:195: parser error : Entity 'thinsp' not defined
to the 6&thinsp;mm endmill with the fewest operating hours.  Then, when a
```

smoe:
How does this affect the PDF? I happily stop introducing those special spaces for now. And, obviously, they are easily sed-ed out should the need arise.

hansu:
It's no problem with the PDF's.

I did some investigation.
First, the warnings can be avoided by using `&#8201;` instead iof `&thinsp;`.
But this doesn't promote the readability of the asciidoc sources. 

Changing  `&thinsp;` to `&#8201;` would be the easiest fix to avoid the warnings. Better to find the root of it, but I have no idea where these warnings come from. If I run only asciidoc to format them as HTMl, I don't get theses warnings.

Second, I did a comparison about different spaces between value and unit (just for fun)

1. No space (no option, I know)
2. Hair space `&#8202;`
3. Thin Space `&#8201;`
4. Normal space

(PDF)
grafik

smoe:
So, for slim numbers like "1" we would want to use half and for wide numbers it would be thin?!? Easy.

hansu:
> So, for slim numbers like "1" we would want to use half and for wide numbers it would be thin?!? Easy.

Yes, we could do that.
But I have another proposal. I think it's the best solution to insert the unicode character as unicode into the source files.
E.g. just copy and paste it from a site like [this](https://www.compart.com/en/unicode/U+2009).
It has two advantages:
- better readable sources
- no warnings

hansu:
>But I have another proposal. I think it's the best solution to insert the unicode character as unicode into the source files.

@smoe any objections regarding this?

smoe:
> > But I have another proposal. I think it's the best solution to insert the unicode character as unicode into the source files.
> 
> @smoe any objections regarding this?

I think I would just want to ignore those number-unit-distances for the time being and just finish those translations. I can live with a few warnings. And if those warnings are wrong then I prefer to fix what is wrong over investing extra work to circumvent the problem.

hansu:
42 warnings which can be eliminated with a simple a replace action, a work wich is quicker done than discussing about it.
But it was more meant if you have objections to insert unicode characters in this way. But seems not. So I go for it?

smoe:
Go for it. That would be done in a separate branch? I would just go and rebase this one to master then.

hansu:
>Go for it. That would be done in a separate branch? I would just go and rebase this one to master then.

Maybe I'll do it after the merge.
But weird that it fails.

PR #2010, opened 2022-09-07
CI more dists

SebKuzminsky:
This extends the "package-arch" and "package-indep" CI jobs to test on Buster and Bullseye (in addition to Testing and Unstable as they currently do).

In order to do this, I added the deb archive at linuxcnc.org to the sources list, so that our backported po4a 0.67 is available on Buster and Bullseye.


.github/workflows/ci.yml
```diff
@@ -1,4 +1,5 @@
 name: Build CI
+# vim: sw=2
```
petterreinholdtsen, line 2: This vim line look like it better stay on the first line outside the YAML structure.  Are you sure it have to be on line two?

.github/workflows/ci.yml
```diff
@@ -65,7 +66,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        image: ["debian:testing", "debian:unstable"]
+        image: ["debian:buster", "debian:bullseye", "debian:testing", "debian:unstable"]
```
petterreinholdtsen, line 69: It seem better to me to use 'oldstable', 'stable', 'testing' and 'unstable' instead of 'buster', 'bullseye', 'testing', 'unstable', to dynamically change when Debian make new releases.  No idea if it work, but recommend testing those names if you have not done so already.

.github/workflows/ci.yml
```diff
@@ -83,7 +84,18 @@ jobs:
...
+
+        # Add linuxcnc.org deb archive to apt sources list.
+        apt-get --yes --quiet install gpg software-properties-common
+        gpg --keyserver hkps://keyserver.ubuntu.com --recv-key 3CB9FD148F374FEF
+        gpg --output /etc/apt/trusted.gpg.d/linuxcnc-deb-archive.gpg --export 3CB9FD148F374FEF
+        DIST=$(echo ${{matrix.image}} | cut -d : -f 2)
+        add-apt-repository "deb http://linuxcnc.org $DIST base"
+        apt-get --quiet update
+
+        # Install stuff needed to turn a linuxcnc working tree into a debian source package.
```
petterreinholdtsen, line 96: This is equivalent to adding random stuff from the internet and allow anyone with the ability to hijack TCP connections to run anything they want on the CI test host.  Perhaps only enable it for oldstable and stable, and not for testing and unstable, to at least reduce the exposure as much as possible?  Or perhaps provide the GPG key alongside the CI script, to start the chain of trust within the git repository instead of having to trust the hkps server protocol and connection?

petterreinholdtsen (changes requested):
The change I want to have to approve this is avoid downloading the GPG key from the Internet and including it in the git repo instead.

.github/workflows/ci.yml
SebKuzminsky: It can be anywhere.

.github/workflows/ci.yml
SebKuzminsky: That's a good idea, I'll include the pubkey in the repo.

.github/workflows/ci.yml
smoe: Admittedly, that change of names that we cannot control is why I prefer to have the distributions named explicitly. So, I would also s/testing/bookworm/.

petterreinholdtsen:
[Steffen Möller]
> Admittedly, that change of names that we cannot control is why I
> prefer to have the distributions named explicitly. So, I would also
> s/testing/bookworm/.

Why?

To me the point is to test the upload dist, the next stable, the current
stable and the previous stable version, not some arbitrary versions that
will be uninteresting in a year or two.

Unrelated to this, it occured to me that the code to install a new
version of po4a is not needed in testing and unstable any more, so a
change similar to this should be implemented there:

diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index 5d8ff9d96e..3cbcf839d7 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,8 +26,6 @@ jobs:
       run: |
         ./scripts/travis-install-build-deps.sh
         sudo apt-get install -y eatmydata
-        curl -O https://snapshot.debian.org/archive/debian/20220716T154603Z/pool/main/p/po4a/po4a_0.67-2_all.deb
-        sudo apt install ./po4a_0.67-2_all.deb
         cd src
         eatmydata ./autogen.sh
         eatmydata ./configure --with-realtime=uspace --disable-check-runtime-deps

For those environments the travis-install-build-deps.sh will install a
new enough version.  With the extra repo, I suspect it will also work
for stable, oldstable and older ubuntu versions.

-- 
Happy hacking
Petter Reinholdtsen

smoe:
> [Steffen Möller]
> Admittedly, that change of names that we cannot control is why I prefer to have the distributions named explicitly. So, I would also s/testing/bookworm/.
> Why? To me the point is to test the upload dist, the next stable, the current stable and the previous stable version, not some arbitrary versions that will be uninteresting in a year or two.

My preference is to keep testing LinuxCNC in the environments we accept or aspire to be compatible with LinuxCNC environments. Just because Debian decides to call something stable that we have not adapted the LinuxCNC code for, in my mind should not trigger any action on our side. To stay relevant we should then manually add another distribution name. Also, our users should not be surprised when the distribution of Debian changes but they still run their buster USB stick.

What we have now (stable name explicitly but testing left as testing) is inconsistent. Either way is better than what we have now. Maybe I should also add that all these many words typed may not be worth the difference, but then again, any user who runs into an unexpected incompatibility when apt-get updating will type many more words :)

SebKuzminsky:
I'm expecting 0808e8b to pass, if so it's ready to merge IMO.

PR #2018, opened 2022-09-10
docs: worked over weblate inconsistencies

smoe:
Went through the many "marked for edit"-entries in Weblate. Most of those Weblate changes were reverse-translations that have now been confirmed in their assignment to the translation. But for a good part there were unexpected
differences between English and the translation that required changes on either side.

### DIFF
--- a/docs/src/config/pncconf.adoc
+++ b/docs/src/config/pncconf.adoc
@@ -13,7 +13,7 @@ parallel port driven systems).
 
 PnCconf is still in a development stage (Beta) so there are some bugs and
 lacking features.
-Please report bugs and suggestions to the LinuxCNC forum page or mail-list.
+Please report bugs and suggestions to the LinuxCNC forum page or mailing list.
 
 There are two trains of thought when using PnCconf:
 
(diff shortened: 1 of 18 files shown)


docs/src/config/pncconf.adoc
```diff
@@ -36,17 +36,17 @@ details.
 
 == Step by Step Instructions
 
-.PnCConf Splash
-image::images/pncconf-splash.png["PnCConf Configuration Wizard"]
+.PnCconf Splash
```
hansu, line 39: Does anyone has an idea for what "PnC" stands for? I guess  @jepler can answer this...

jepler: > Does anyone has an idea for what "PnC" stands for? I guess  @jepler can answer this...

I didn't originate this part of linuxcnc

hansu: As it is based on Stepconf, I assumed you originated this as well. My bad. A look into the sources gives me @c-morley as possible candidate for this question :slightly_smiling_face:

c-morley: I'm guilty (of course hugely based on Jeff's way-better-coded Stepconf at the time)
Point aNd Click

hansu: Thanks, great to have this clarified :slightly_smiling_face:

PR #1896, opened 2022-08-06
Duplicate Strings gmoccapy.pot and linuxcnc.pot

zz912:
The same Strings were generated to gmoccapy.pot and linuxcnc.pot
I inspired here:
https://github.com/LinuxCNC/linuxcnc/pull/542/files

This FIX needs to be done only for branche 2.8. It already exists in the master version.

PR #542, opened 2018-12-14
Modify the Submakefile that extracts the POT file.

Solitarily:
The translation of "gmoccapy" does not need to be included in "linuxcnc.pot". So I commented it out. The translation of "gladevcp" is not valid in "gmoccapy.pot". It should be included in "linuxcnc.pot".

PR #2020, opened 2022-09-11
Revert "Duplicate Strings gmoccapy.pot and linuxcnc.pot"

gmoccapy:
Reverts LinuxCNC/linuxcnc#1896

It is very frustrating to see that you now splited the gmoccapy translation in two pot files.

gmoccapy.glade is the main screen from gmoccapy! It is no widget it is the main screen!!
All the effort I have done to translate the GUI in different languages has become now much more complicated. I explained all at the beginning of this discussion why I did all the thinks.

I agree that it should not be in both submake files, but IT BELOGNGS TO GMOCCAPY not to linuxcnc.

As I am still out of time I hope someone will revert this change, otherwise I will do that as soon as I have time.

Norbert

### DIFF
--- a/src/emc/usr_intf/gmoccapy/Submakefile
+++ b/src/emc/usr_intf/gmoccapy/Submakefile
@@ -3,6 +3,9 @@ GMOCCAPY_MODULES = dialogs getiniinfo notification player preferences widgets
 PYTARGETS += ../bin/gmoccapy ../lib/python/gmoccapy/__init__.py $(patsubst %,../lib/python/gmoccapy/%.py,$(GMOCCAPY_MODULES)) \
 	../share/gmoccapy/gmoccapy.glade 
 
+PYI18NSRCS += emc/usr_intf/gmoccapy/gmoccapy.py $(patsubst %,emc/usr_intf/gmoccapy/%.py,$(GMOCCAPY_MODULES)) \
+	emc/usr_intf/gmoccapy/gmoccapy.glade 
+
 ../lib/python/gmoccapy/__init__.py:
 	@mkdir -p ../lib/python/gmoccapy
 	@touch $@


hansu:
It's the same change as in https://github.com/LinuxCNC/linuxcnc/pull/542/

You wrote finally in [this comment](https://github.com/LinuxCNC/linuxcnc/pull/542/#issuecomment-841332678):

>I just checked again, may be it can be this way, I do not see any difference after doing a make clean and build fresh between master and 2.8.
>
>So ok, may be merged to 2.8 too.

gmoccapy:
And also I mentioned:

> Why I did include some glade files in my Makefile was, that that ones did not have valid translations and are used within gmoccapy.

gmoccapy:
Most files have now a valid translation, because they have been translated together with gmoccpay.

So the widget files can be in linuxcnc, but gmoccapy.glade need to be in gmoccapy.pot file.

Norbert

hansu:
>So the widget files can be in linuxcnc, but gmoccapy.glade need to be in gmoccapy.pot file.

gmoccapy.glade is still in gmoccapy.pot. The diff in https://github.com/LinuxCNC/linuxcnc/pull/1896#issuecomment-1242692841 shows that gmoccapy.glade, gmoccapy.py, gmoccapy/dialogs.py and so on are only removed in linuxcnc.pot.

zz912:
Mr. Norbert,

I deleted "PYI18NSRCS"

PYI18NSRCS  GOES TO LINUXCNC
look at:
https://github.com/LinuxCNC/linuxcnc/blob/e15761456940b34210b2121afa5c30d140acee3f/src/po/Submakefile#L110

gmoccapy:
Sorry for the inconvinience! I took the time to check all the changes again and you are right, the translation of gmoccapy is still individuell from the huge linuxcnc pot file.

I need to take more time to double-check my comments and bet your pardon!

Norbert

PR #2021, opened 2022-09-11
autoconf: new ax_boost_base from autoconf-archive

SebKuzminsky:
This commit updates src/m4/ax_boost_base.m4 to the version in autoconf-archive v2022.09.03.  This fixes a bug with boost library path detection on armhf.


jepler (commented):
Seems ok, didn't test

PR #2030, opened 2022-09-12
Added new component reset from MachineKit.

petterreinholdtsen:
This component is one of several currently only in MachineKit.  It is simply copied from the https://github.com/machinekit/machinekit-hal repo and a line with 'author' information was added.

### DIFF
--- a/src/hal/components/reset.comp
+++ b/src/hal/components/reset.comp
@@ -36,7 +36,7 @@
  *
  ******************************************************************************/
 
-component reset "Resets a IO signal";
+component reset "Resets an IO signal";
 pin in  bit     trigger             "Trigger input";
 pin io  u32     out_u32      = 0    "Unsigned 32 bit integer output value";
 pin in  u32     reset_u32    = 0    "Unsigned 32 bit integer reset value";
@@ -52,6 +52,11 @@ pin in  bit     falling       = false "Trigger on falling edge";
 function _  fp "Update the output value";
 description """
 Component to reset IO signals.
+
+This function works like a conditional sets - it is fed with a float
+and/or bit/s32/u32 pins that are I/O, but are save the value only if
+the "trigger" pin is set. The values assigned to those signals are
+passed via the input pins reset_float/s32/u32/bit.
 """;
 license "GPL";
 author "Alexander Rössler";
--- a/src/hal/components/reset.comp
+++ b/src/hal/components/reset.comp
@@ -1,0 +1,72 @@
+/******************************************************************************
+ *
+ * Copyright (C) 2015 Alexander Rössler
+ *
+ *
+ * This module resets a IO signal
+ *
+ ******************************************************************************
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+ * 02110-1301, USA.
+ *
+ * THE AUTHORS OF THIS PROGRAM ACCEPT ABSOLUTELY NO LIABILITY FOR
+ * ANY HARM OR LOSS RESULTING FROM ITS USE.  IT IS _EXTREMELY_ UNWISE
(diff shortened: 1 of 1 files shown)


hansu:
What is the purpose of this component? Maybe a bit more description would be nice.

petterreinholdtsen:
[Hans Unzner]
> What is the purpose of this component? Maybe a bit more description
> would be nice.

I have no idea. :)

I just wanted to draw attention to the components present in MachineKit
but not in LinuxCNC, and see if it is a point to bring the two sets of
components closer together.

As those who join us on IRC already know, I notice there are 37
components in machinekit that are not in linuxcnc, 104 'v2' components
that might be in linuxcnc already, and 42 components in linuxcnc that
are not in machinekit.  I did not try to port all of them, just picked
one seemingly generic one at random and tried to add it to LinuxCNC.
 
-- 
Happy hacking
Petter Reinholdtsen

src/hal/components/reset.comp
```diff
@@ -0,0 +1,72 @@
...
+ * machinery must be designed to comply with local and national safety
+ * codes, and the authors of this software can not, and do not, take
+ * any responsibility for such compliance.
+ *
+ * This code is part of the Machinekit HAL project.  For more
+ * information, go to https://github.com/machinekit.
+ *
+ ******************************************************************************/
+
+component reset "Resets a IO signal";
```
smoe, line 39: n

src/hal/components/reset.comp
```diff
@@ -0,0 +1,72 @@
...
+pin io  float   out_float    = 0.0  "Float output value";
+pin in  float   reset_float  = 0.0  "Float reset value";
+pin io  bit     out_bit      = false  "Bit integer output value";
+pin in  bit     reset_bit    = false  "Bit reset value";
+pin in  bit     retriggerable = true  "Allow additional edges to reset";
+pin in  bit     rising        = true  "Trigger on rising edge";
+pin in  bit     falling       = false "Trigger on falling edge";
+function _  fp "Update the output value";
+description """
+Component to reset IO signals.
```
smoe, line 54: Adding a bit more of a description - This function works like a conditional sets - it is fed with a float and/or s/u 32 pins that are I/O, but are written to only if the "reset_bit" is set. The values assigned to those signals are passed via the input pins reset_float/s32/u32.

hansu:
>Thank you for this, Petter. That seems like a good initiative. @hansu , this sounds like something for the next OpenMike session to discuss: Adoption of MachineKit components for LinuxCNC 3.0?

You mean LinuxCNC 2.10 😉. 
Yeah we can discuss that.

c-morley:
Out of curiosity, what's the interest of bringing in machinekit's components?

petterreinholdtsen:
[c-morley]
> Out of curiosity, what's the interest of bringing in machinekit's
> components?

My interest is to make sure LinuxCNC have as many of the nice features
of MachineKit available.

I just created <URL: https://github.com/LinuxCNC/linuxcnc/pull/2034 >
to add 64 bit types in HAL, allowing more MachineKit components to be
integrated into LinuxCNC.

-- 
Happy hacking
Petter Reinholdtsen

silopolis:
Le mar. 13 sept. 2022 à 16:22, petterreinholdtsen ***@***.***>
a écrit :

> [c-morley]
> > Out of curiosity, what's the interest of bringing in machinekit's
> > components?
>
> My interest is to make sure LinuxCNC have as many of the nice features
> of MachineKit available.



Me likes that! :)

hansu:
In general I agree if the regarding component is of interest. In this case I have no idea about the benefit of this component.

cakeslob:
Hey guys, since theres some questions as to the purpose of this comp, I though I would test out some AI stuff with a friend. 
Heres what it generated based on the link to the comp file. Weather or not this is accurate is up for question, but it looks legit so let me know what you think. 

#####################################

This code is for a component called "reset" which is used in the Machinekit HAL. The purpose of the component is to provide a way to reset the machine (or system) in a controlled manner. The component has two input pins (reset_in and enable_in) and one output pin (reset_out).When the reset_in pin is asserted (high), the component will assert the reset_out pin. This will cause the machine to reset.The enable_in pin can be used to enable or disable the component. If the enable_in pin is asserted (high), the component is enabled. If the enable_in pin is deasserted (low), the component is disabled.The component also has a parameter called "polarity" which can be used to control the polarity of the reset_out pin. If the polarity is set to 0 (the default), the reset_out pin will be active-low. This means that the machine will reset when the reset_out pin is asserted (low). If the polarity is set to 1, the reset_out pin will be active-high. This means that the machine will reset when the reset_out pin is deasserted (high).


#############################################################

andypugh:
I have no idea what use it is, I have certainly never needed this, but I have merged it anyway.

hansu:
In my opinion the name is a bit misleading. It feels more like a set than a reset.
So from the name I wouldn't expect that I can set the signal to a specific value on a trigger condition.
As far as I understood this, the output is zero until the trigger input becomes active and then it is the output is set to the value of the reset-input.
It sounds more like a sample & hold (https://en.wikipedia.org/wiki/Sample_and_hold) function.

PR #2031, opened 2022-09-12
Update hm2_eth.c

sundtek:
close socket when executing close_board(), shutdown is not enough.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/hm2_eth.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_eth.c
@@ -746,19 +746,23 @@ static int init_board(hm2_eth_t *board, const char *board_ip) {
 }
 
 static int close_board(hm2_eth_t *board) {
-
+    int ret;
     board->llio.reset(&board->llio);
 
     if(use_iptables()) clear_iptables();
 
     if(board->req.arp_flags & ATF_PERM) {
-        int ret = ioctl_siocdarp(board);
+        ret = ioctl_siocdarp(board);
         if(ret < 0) perror("ioctl SIOCDARP");
     }
-    int ret = shutdown(board->sockfd, SHUT_RDWR);
-    if (ret < 0)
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
[sundtek]
> close socket when executing close_board(), shutdown is not enough.

Is there a way to observe the problem caused by not calling close()?
Can we add a test to detect the problem?

Also, are you sure shutdown() is still needed when using close()?
In <URL: https://stackoverflow.com/questions/48208236/tcp-close-vs-shutdown-in-linux-os >

I both read that shutdown() and close() is needed, and that close() will
close both reading and writing as I believe shutdown(board->sockfd, SHUT_RDWR)
is doing, and thus I am unsure.  As far as I can tell, it do not harm to
do both, as there is no writing or reading between shutdown() and close here.

-- 
Happy hacking
Petter Reinholdtsen

sundtek:
as far as I understand shutdown only closes the connection -- but not the socket.

I only spotted the issue when reading through the sources, I do not have a mesa card yet.

petterreinholdtsen:
[sundtek]
> as far as I understand shutdown only closes the connection -- but not
> the socket.

I agree.  My main question is if shutdown() is still needed if close()
is called just after it.

Note, I am not objecting to this change.

-- 
Happy hacking
Petter Reinholdtsen

sundtek:
Considering that this is a UDP socket (connection less) shutdown is not needed at all but doesn't hurt either.

jepler:
Thanks for your contribution! I don't see the harm to this change. (Though the question of whether to eliminate shutdown should be resolved and it would be nice if it could be tested on hardware)

However.. 

It's not particularly important since we rely on rtapi_app exiting to clean up all resources like file descriptors.

pcw-mesa:
I tested this change (start/close LinuxCNC with an Ethernet card about 20 times) 
and it has no obvious effects on behavior

jepler:
Thank you for submitting & testing!

PR #2034, opened 2022-09-13
Extend HAL to handle 64bit types and add new abs_s64 component

petterreinholdtsen:
The abs_s64 component is fetched from MachineKit, and the 64 bit type change is inspired by the u64 and s64 type available in MachineKit.


src/hal/halmodule.cc
```diff
@@ -74,15 +79,57 @@ bool from_python(PyObject *o, double *d) {
...
+bool from_python(PyObject *o, hal_s32_t *i) {
     PyObject *tmp = 0;
     long long l;
     tmp = PyLong_Check(o) ? o : PyNumber_Long(o);
     if(!tmp) goto fail;
 
     l = PyLong_AsLongLong(tmp);
     if(PyErr_Occurred()) goto fail;
-    if(l < 0 || l != (uint32_t)l) {
+    if(l != (hal_s32_t)l) {
```
smoe, line 111: Should this not be hal_u32_t ?

smoe: It should not :) On https://docs.python.org/3/c-api/long.html, it is long long PyLong_AsLongLong([PyObject](https://docs.python.org/3/c-api/structures.html#c.PyObject) *obj), so there is no particular need to cast this down, right? Anyway ...

hansu:
The question is more if ever a use case popped up where s/uint32 was to small ...

smoe:
No, that is not the question. The question is if HAL stands true to its name and abstracts the hardware we are having. And today's hardware frequently features native 64bit computes. And I can think of use cases - easily.

hansu:
But then we should replace the 32 bit data type generally by 64 bit and change all components accordingly. Or have a 32 bit and 64 bit variant of all components... (just  tie in with your little provocation :D)

smoe:
I do not think so since HAL may communicate with various sorts of microcontrollers (or custom CPUs on FPGAs?) and they may .. donno .. maybe even only feature anything down to 8 bits? Or only [4 bits](https://de.wikipedia.org/wiki/Nibble)?

Another thought is that for a link to robotics and whatever AI may assist them (or assist LinuxCNC), we just do no know what we may need to efficiently communicate. I would hence prepare for all the common basic data types. I do not see that this costs us anything. Does it?

May I point to https://en.wikipedia.org/wiki/Quadruple-precision_floating-point_format ?  I am tempted.

andypugh:
I am curious about what Machinekit considers 64 bit types useful for. 
Though it is worth noting that the stepgens and encoder counters store counts internally in 64 bit buffers. 
However, they update those from smaller data types, either single bits or, in the case of Mesa cards single bytes (smart-serial MPG) or (from memory) a 16-bit buffer for the full encoders. It is just assumed that a delta > 50% of the data width is a wrap. 
Smart-serial supports a very wide range of data types, signed and unsigned ints from 2 to 96 bits, 8 and 16 bit floats (as well as the more usual HAL data types)

petterreinholdtsen:
[Andy Pugh]
> I am curious about what Machinekit considers 64 bit types useful for.

Perhaps you can find the answer to this in the git repositories on
<URL: https://github.com/machinekit> or their web site
<URL: http://machinekit.io/ >?  I have not searched a lot for it.  It
se like a useful extention given that many computers these days are 64
bit natively, so I did not try to locate its use case.
-- 
Happy hacking
Petter Reinholdtsen

hansu:
It seems that every build on master fails after this PR was merged:

```
warning: unknown warning option '-Werror=frame-larger-than=2560'; did you mean '-Werror=frame-larger-than='? [-Wunknown-warning-option]
hal/components/conv_float_u64.comp:14:5: error: integer literal is too large to be represented in any integer type
#if 18446744073709551616u != 0
    ^
1 warning and 1 error generated.
```

petterreinholdtsen:
[Hans Unzner]
> It seems that every build on master fails after this PR was merged:

It sure look like it.  Clang refuse the code with an error, while gcc
just present a warning.

I'll see what I can do.

-- 
Happy hacking
Petter Reinholdtsen

PR #2035, opened 2022-09-13
Added new component led_dim from MachineKit.

petterreinholdtsen:
This component is one of several currently only in MachineKit.  It is simply copied from the https://github.com/machinekit/machinekit-hal repo.


hansu:
This sounds like an enrichment 👍

silopolis:
Let the machine pimping contest begin! 🤩😉

PR #2046, opened 2022-09-17
docs: iterating through weblate edits

smoe:
One gets the impression that we are getting closer to an end.


docs/src/drivers/mb2hal.adoc
```diff
@@ -57,24 +57,22 @@ Set `HAL_MODULE_NAME=mymodule` (default `HAL_MODULE_NAME=mb2hal`)
 `[MB2HAL_INIT]`
 
 [options="header",cols="14,8,13,67"]
-|=======================
+|===
 |Value | Type | Required | Description
-|`INIT_DEBUG`| Integer | No | Debug level of init and INI file parsing. +
+l|INIT_DEBUG | Integer | No | Debug level of init and INI file parsing. +
```
hansu, line 62: These changes I don't like. Looked better before (left):
grafik

hansu (changes requested):

smoe:
Would need to be l -> m, then. Updated.

hansu:
@smoe a merge conflict...

PR #2047, opened 2022-09-18
docs: po4a-fixed tables in hal/components

smoe:
Encountered a series of missing terminal "e" etc in weblate. po4a is depending on a blank prior to the | in tables (which apparently is formally correct to do) and otherwise interprets letters directly preceeding the | as an operator.


hansu:
I cannot find any missing "e"'s in the diff. However, it looks good, thanks :)

smoe:
The terminal "e", "s", "m","l" ... are missing in what po4a produces, so I see them missing in weblate.

hansu:
Ah ok.

PR #2049, opened 2022-09-18
docs: improve docs regarding SSERIAL

hansu:
Would be nice if someone could double-check these very few lines i changed in hostmot2.9. Thanks!

### DIFF
--- a/docs/man/man9/hostmot2.9
+++ b/docs/man/man9/hostmot2.9
@@ -209,25 +209,26 @@ only two pins (step and dir) but the Hostmot2 stepgen can drive up to 8 output
 pins for specialised applications (depending on firmware). This parameter
 applies to all stepgen instances. Unused, masked pins will be available as GPIO.
 .TP
-\fBsserial_port_N (N = 0 .. 3)\fR [optional, default: 00000000 for all ports]
+\fBsserial_port_\fIN\fR (\fIN\fR = 0 .. 3)\fR [optional, default: 00000000 for all ports]
 Up to 32 Smart Serial devices can be connected to a Mesa Anything IO board
 depending on the firmware used and the number of physical connections on the
-board. These are arranged in 1-4 ports of 1 to 8 channels.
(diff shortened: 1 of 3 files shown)


docs/man/man9/hostmot2.9
```diff
@@ -209,25 +209,26 @@ only two pins (step and dir) but the Hostmot2 stepgen can drive up to 8 output
...
 Up to 32 Smart Serial devices can be connected to a Mesa Anything IO board
 depending on the firmware used and the number of physical connections on the
-board. These are arranged in 1-4 ports of 1 to 8 channels.
+board. These are arranged in 1-4 ports (N) of 1 to 8 channels.
  Some Smart Serial (SSLBP) cards offer more than one load-time configuration,
 for example all inputs, or all outputs, or offering additional analogue input on
 some digital pins.
- To set the modes for port 0 use, for example \fBsserial_port_0=0120xxxx\fR
- A "0" in the string sets the corresponding port to mode 0, 1 to mode 1, and so
+ To set the modes for port 0, use for example \fBsserial_port_0=0120xxxx\fR
```
andypugh, line 219: I think the the comma is an elipsis, so it was wrong before and the correction is wrong too. 

To set the mode for port 0 use, for example, ....

hansu: Or maybe completely without commas?

smoe: To me, the bug is that these pins use an underscore instead of a dot (sserial.port.0 is what I would have expected) :o/. I would reorder that sentence, i.e. bring the pin to the beginning of the sentence, like 
```
E.g., the assignment of pin as in "\fBsserial_port_0=0120xxxx\fR" would set the modes of port 0..
```

hansu: [Steffen]
No that's correct - it's a config string, not a pin.
Or keep it short:
`E.g., \fBsserial_port_0=0120xxxx\fR would set the modes for port 0.`

PR #2050, opened 2022-09-19
build mesa_uart on uspace

hansu:
Was there a specific reason why `mesa_uart.comp `was not added to the build when building for uspace?


smoe (approved):
@jepler , @andypugh , you may be closest to confirm that this is indeed reasonable to add to uspace. If I get/guess this right (have not encountered this in the documentation, yet) then this is what allows me to use the mesa cards to talk to various microcontrollers, right? I want/need that also in uspace :)

jepler (approved):
It builds so it must be right.

PR #2051, opened 2022-09-19
Update links in documentation

hansu:
Is everyone fine with that if all links in the docs which are pointing to 
http://linuxcnc.org/docs/html/* are changed to 
http://linuxcnc.org/docs/2.9/html/* ?
(This is mostly the case in links in the man pages.)

This allows us to reference always the correct version of the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ for such compliance.
 [Weblate]: https://hosted.weblate.org/projects/linuxcnc/
 [Website]: https://linuxcnc.org/
 
-[Ｄｏｃｕｍｅｎｔａｔｉｏｎ]: http://linuxcnc.org/docs/stable/html/
-[Ｉｎｓｔａｌｌ]: http://linuxcnc.org/docs/stable/html/getting-started/getting-linuxcnc.html
-[Ｂｕｉｌｄ]: http://linuxcnc.org/docs/stable/html/code/building-linuxcnc.html
+[Ｄｏｃｕｍｅｎｔａｔｉｏｎ]: http://linuxcnc.org/docs/2.9/html/
+[Ｉｎｓｔａｌｌ]: http://linuxcnc.org/docs/2.9/html/getting-started/getting-linuxcnc.html
+[Ｂｕｉｌｄ]: http://linuxcnc.org/docs/2.9/html/code/building-linuxcnc.html
 [License]: COPYING
--- a/README_es.md
+++ b/README_es.md
@@ -22,4 +22,4 @@ La traduccion al español de este software queda adherida a este descargo y a la
 
 Consulte el archivo 'docs/src/code/building-linuxcnc.txt' para obtener información sobre cómo construir y ejecutar el software, o vea:
 
-http://linuxcnc.org/docs/devel/html/code/building-linuxcnc.html
+http://linuxcnc.org/docs/2.9/html/code/building-linuxcnc.html
--- a/configs/by_interface/mesa/hm2-servo/hm2-servo.hal
+++ b/configs/by_interface/mesa/hm2-servo/hm2-servo.hal
@@ -15,7 +15,7 @@
 # sudo ln -s $HOME/emc2-sandbox/src/hal/drivers/mesa-hostmot2/firmware /lib/firmware/hm2
 #
 # See also:
-# <http://www.linuxcnc.org/docs/devel/html/man/man9/hostmot2.9.html#config%20modparam>
+# <http://www.linuxcnc.org/docs/2.9/html/man/man9/hostmot2.9.html#config%20modparam>
 # and http://wiki.linuxcnc.org/cgi-bin/emcinfo.pl?HostMot2
 #
 # #####################################################################
--- a/configs/by_interface/mesa/hm2-stepper/hm2-stepper.hal
+++ b/configs/by_interface/mesa/hm2-stepper/hm2-stepper.hal
@@ -15,7 +15,7 @@
 # sudo ln -s $HOME/emc2-sandbox/src/hal/drivers/mesa-hostmot2/firmware /lib/firmware/hm2
 #
 # See also:
-# <http://www.linuxcnc.org/docs/devel/html/man/man9/hostmot2.9.html#config%20modparam>
+# <http://www.linuxcnc.org/docs/2.9/html/man/man9/hostmot2.9.html#config%20modparam>
 # and http://wiki.linuxcnc.org/cgi-bin/emcinfo.pl?HostMot2
 #
 # #####################################################################
--- a/configs/by_interface/mesa/plasma-5i20/plasma-5i20.hal
+++ b/configs/by_interface/mesa/plasma-5i20/plasma-5i20.hal
@@ -4,7 +4,7 @@
 # sudo ln -s $HOME/emc2-sandbox/src/hal/drivers/mesa-hostmot2/firmware /lib/firmware/hm2
 #
 # See also:
-# <http://www.linuxcnc.org/docs/devel/html/man/man9/hostmot2.9.html#config%20modparam>
+# <http://www.linuxcnc.org/docs/2.9/html/man/man9/hostmot2.9.html#config%20modparam>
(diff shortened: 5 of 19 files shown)


petterreinholdtsen:
[Hans Unzner]
> Is everyone fine with that if all links in the docs which are pointing to 
> http://linuxcnc.org/docs/html/* are changed to 
> http://linuxcnc.org/docs/2.9/html/* ?
> (This is mostly the case in links in the man pages.)

I note that it is currently a mix of stable/devel/"" links with no
apparent system.  My gut feeling is that instead of updating every URL
once per release, it would be better to link to the 'stable' version
everywhere.

> This allows us to reference always the correct version of the docs.

Only right now.  After 2.9 is released, it will be wrong  again. :)

Also, Do we really want to maintain _es files separately?  We could
maintain them using po4a instead.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
>I note that it is currently a mix of stable/devel/"" links with no
apparent system. 

Now, but after this patch all links will be /2.9/.

>it would be better to link to the 'stable' version
everywhere.

But then you switch between versions when viewing the documentation. Not very helpful in case something important has changed and you don't pay attention on the URL.

>Only right now.  After 2.9 is released, it will be wrong  again. :)

Of course they have to be updated then like the version number. We can add a variable in asciidoc to change this only once e.g...

silopolis:
Le lun. 19 sept. 2022 à 18:12, Hans Unzner ***@***.***> a
écrit :

> Is everyone fine with that if all links in the docs which are pointing to
> http://linuxcnc.org/docs/html/* are changed to
> http://linuxcnc.org/docs/2.9/html/* ?
> (This is mostly the case in links in the man pages.)
>
> This allows us to reference always the correct version of the docs.
>

Already suggested it a couple of times but... The best solution for this
would be an AsciiDoc attribute referencing the branch name and maybe
another one for the docs base URL.
Something like:

```
:branch: {sys: git rev-parse --abbrev-ref HEAD}
:wlo-url: https://www.linuxcnc.org
:docs-base: {wlo-url}/docs/
:docs-base-html: {docs-base}/{branch}/{lang}/html/
:docs-base-pdf: {docs-base}/{branch}/{lang}/pdf/

```

Links could then be shortened to {docs-base-html}/* in the body of text.

The order and position of branch and lang attributes in URL pattern is just
an example.

Dynamic, no more links tweaking and release chore, and future proof.

hansu:
Yes that was what I meant with "variable" :+1: 
So one more reason to go on with #1376 ;-)

silopolis:
Le lun. 19 sept. 2022 à 22:25, Hans Unzner ***@***.***> a
écrit :

> Yes that was what I meant with "variable" 👍
>

oh ! Sorry, skimmed over it 🖐

So one more reason to go on with #1376
> <https://github.com/LinuxCNC/linuxcnc/pull/1376> ;-)
>

The list is getting long ;)

smoe:
> Also, Do we really want to maintain _es files separately? We could maintain them using po4a instead.

That is INSTALL_es and README_es, right? The English INSTALL.adoc is considerably lacking. Should this not just be a pointer to the appropriate chapter in the documentation? I opt to remove them both.

The README should also be rewritten. The best of it is the notion of .po files :)

hansu:
>Also, Do we really want to maintain _es files separately?  We could
maintain them using po4a instead.

I only updated the files which are in the repository...

petterreinholdtsen:
There are merge conflicts here.

I came across another version specific link today:

```
diff --git a/docs/src/man/man1/xhc-whb04b-6.1.adoc b/docs/src/man/man1/xhc-whb04b-6.1.adoc
index d2bf580cc3..1de11aab54 100644
--- a/docs/src/man/man1/xhc-whb04b-6.1.adoc
+++ b/docs/src/man/man1/xhc-whb04b-6.1.adoc
@@ -401,7 +401,7 @@ MDI_COMMAND=(debug,macro16)
 
 Exercise caution if using copy and paste of this example code from the online web docs.
 Certain characters are incompatibly encoded by the web site (minus becomes em-dash).
-It is safer to copy and paste from https://raw.githubusercontent.com/LinuxCNC/linuxcnc/2.8/src/hal/user_comps/xhc-whb04b-6/example-configuration.md[].
+It is safer to copy and paste from https://raw.githubusercontent.com/LinuxCNC/linuxcnc/master/src/hal/user_comps/xhc-whb04b-6/example-configuration.md[] .
 
 ----
 #
```

hansu:
[Jérémie]
>Already suggested it a couple of times but... The best solution for this
would be an AsciiDoc attribute referencing the branch name and maybe
another one for the docs base URL.

The majority of affected files are man pages. So the Asciidoc solution would only cover a few.
But I have a proposal. Not perfect, but it works.
The URL is displayed in the man page like
`http://linuxcnc.org/docs/devel/html/getting-started/about-linuxcnc.html <>` instead of 
`<http://linuxcnc.org/docs/devel/html/getting-started/about-linuxcnc.html>`.
But in the HTML it's correct. As we plan to convert the man pages to AsciiDoc I wouldn't put too much work in here.

It's an disadvantage for comp-files that they need the include file to show the URL properly (I am thinking when users exchange those files). But I found only two occurrences in comp files, so maybe it's better to skip those.

Just let me know if you guys think it's an good idea to do that for all links in man page.

petterreinholdtsen:
[Hans Unzner]
> 674a5d860044659737a5477c9b3f2e605a2bd65d  proposal how to unify links
> in man pages

Is the intention to use the .SO trick in all man pages using
linuxcnc.org links?

Did you succeed in finding a way to do something similar for adoc files?

My gut feelig is that it would be better to use either the /devel/ or
the /stable/ links, not specific version numbers, as the former have a
longer life span than fixed version numbers.  If I had to choose, I
would pick /stable/, I suspect, but am not sure.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
[Petter]
> Is the intention to use the .SO trick in all man pages using linuxcnc.org links?

Yes.

>Did you succeed in finding a way to do something similar for adoc files? 

That part is easy. Just like Jérémie suggested in https://github.com/LinuxCNC/linuxcnc/pull/2051#issuecomment-1251465592 or similar.

>My gut feelig is that it would be better to use either the /devel/ or the /stable/ links, not specific version numbers, as the former have a longer life span than fixed version numbers. If I had to choose, I would pick /stable/, I suspect, but am not sure.

I don't agree, because the "stable" and "devel" does change from release to release. If we change all links in 2.8 to stable, they will soon point to 2.9. And 2.9 is currently between "stable" and "devel".

Maybe we can derive this part of the links from the branch name. 
So  `http://linuxcnc.org/docs/<branch-name>/html/getting-started/about-linuxcnc.html` for example.

But then make needs to be run before the links can be displayed properly.

hansu:
[Petter]
>Why is this a problem?  If we assume the documentation become better and
more complete, while still documenting how behavour change from version
to version, it would always be good to read the latest edition.

Assume some point in the future where 2.9 is just released but you are still using 2.8 and you see no reason to upgrade to 2.9 (or just haven't done it yet or maybe even don't know about the new version). You click on a link in a man page that points you to a docs page like linuxcnc.org/docs/stable/html/... at take the information on that page at face value. But the component might have changed in 2.9 so you are wasting your time trying to figure out why it doesn't work for your 2.8 system.

petterreinholdtsen:
[Hans Unzner]
> But the component might have changed in 2.9 so you are wasting your
> time trying to figure out why it doesn't work for your 2.8 system.

A good way to handle this is to explain in the documentation how the
behavoiur have changed and when.  This is good in any case to help those
upgrading from earlier setups, even if they are running the latest
version.  If there is description of newer features not found in the old
version, it might even motivate to upgrade to a more recent LinuxCNC
version.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
[Petter]
>Did you succeed in finding a way to do something similar for adoc files?

For adoc it's is normally not necessary as we use anchors for inter-adoc links. There might be only very few exceptions where we link to a version specific URL which is not generated from adoc.

PR #2057, opened 2022-09-21
gmoccapy: fix error on creating file when RS274NGC_STARTUP_CODE is not set

hansu:
Reported here: https://forum.linuxcnc.org/gmoccapy/46944-error-when-creating-new-program

@gmoccapy are you ok with this simplification?


gmoccapy:
Thanks for fixing this

hansu:
I take this as a yes :)

PR #2058, opened 2022-09-21
doc: hal, intro - attempt for an easier first presentation (smoe:docs_hal_intro)

smoe:
Helpful?

### DIFF
--- a/docs/src/hal/intro.adoc
+++ b/docs/src/hal/intro.adoc
@@ -79,16 +79,16 @@ But multiple interfaces have been provided that allow HAL to be manipulated
 but none of these interfaces are 'HAL itself'.
 
 HAL itself is not a program, it consists of one or more lists of loaded programs (the components) that are periodically executed (in strict sequence), and an area of shared-memory that these components use to interchange data. 
-
 The main HAL script runs only once at machine startup, setting up the realtime threads and the shared-memory locations, loading the components and setting up the data links between them (the "signals" and "pins")
 
-In principle multiple machines could share a common HAL to allow them to inter-operate, however the current implementation of LinuxCNC is limited to a single interpreter and a single Task module. Currently this is almost always a G-code interpreter and "milltask" (which was found to also work well for lathes and adequately for robots) but these modules are selectable at load-time. 
+In principle multiple machines could share a common HAL to allow them to inter-operate, however the current implementation of LinuxCNC is limited to a single interpreter and a single Task module.
+Currently this is almost always a G-code interpreter and "milltask" (which was found to also work well for lathes and adequately for robots) but these modules are selectable at load-time. 
 With an increasing interest in the control of multiple cooperating machines, to overcome this limitation is likely one of the prime steps for the future development of LinuxCNC to address.
 It is a bit tricky though and the community is still organizing its thoughts on this.
 
 HAL lies at the core of LinuxCNC and is used and/or extended by all the parts of LinuxCNC, which includes the GUIs.
 The G-code (or alternative language) interpreter knows how to interpret the G-code and translates it into machine operations by triggering signals in HAL.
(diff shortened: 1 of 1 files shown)


silopolis:
Le jeu. 22 sept. 2022 à 01:13, Steffen Möller ***@***.***> a
écrit :

> Helpful?
>

Any improvement to the introduction of this fundamental part of the system
is definitely helpful! 👍

On first read I see some improvement... But some parts not that much. Will
review when not on mobile...

smoe:
This seems to build again. @silopolis, I rewrote most of my initial attempt and consider this much better already.

hansu:
I don't agree 100% to that proposal. Some parts I think they not really belong here or are not helpful like the paragraph "Reuse of Components and New Developments" which I would remove. 
But other parts like "Pins and Signals" are a good improvement.

>Technically, the HAL is a regular program running in the background. It is a special in that it can communicate really fast with other progams.

This sentence is not really informative

>The HAL is the very root of LinuxCNC.

One could also understand the interpeter (canon) as the root...


**I think one of our more experienced developers should review that.**

smoe:
> I don't agree 100% to that proposal. Some parts I think they not really belong here or are not helpful like the paragraph "Reuse of Components and New Developments" which I would remove.

I had already forgotten that I wrote that. Was meant to explain some vertical integration, I presume.

> But other parts like "Pins and Signals" are a good improvement.
 
> > Technically, the HAL is a regular program running in the background. It is a special in that it can communicate really fast with other progams.
> 
> This sentence is not really informative

It was meant to guide noobs (like me) to distinguish between
 * the middleware (as which I expect HAL to identify itself) and the
 * many components that contribute the knowledge on pieces of CNC equipment and the 
 * integration (what components listens to what other component) of it all in the many HAL files.

I am not completely competent that my declaration of HAL as a [blackboard system](https://en.wikipedia.org/wiki/Blackboard_system) is technically completely sound, but that imagination _very much_ helped me to approach it.

> 
> > The HAL is the very root of LinuxCNC.
> 
> One could also understand the interpeter (canon) as the root...

Fantastic! When it comes to understanding G-code - yes. From how I perceived it, the G-code is just something that keeps triggering the system but it is HAL that keeps all components together. So, that introduction should put HAL in its place and a ref to canon is indeed needed.

> **I think one of our more experienced developers should review that.**

Please!!!

Thank you @hansu, already!

smoe:
Changed PR to 2.9 :)  Thanks!

smoe:
So, here we are. I have basically rewritten the top half in a way that I could understand what HAL is doing, and this is what I feel like an intro could be. The bottom half that is not a glossary is bit like an integrators tutorial and I would like to move that text elsewhere. I am also not a complete fan of the part on ClassicLadder - the many different ways to implement extensions to HAL or to just interfere with it via halcmd should be introduced, instead.

docs/src/hal/intro.adoc
```diff
@@ -1,7 +1,6 @@
 :lang: en
 :toc:
 
-[[cha:hal-introduction]]
```
hansu, line 4: I guess this causes the build failure.

hansu:
I think it is quite much to read for an introduction. But I don't know enough to give a good short explanation/summary of this.

smoe:
It is admittedly more than a "and here comes HAL"-kind of intro. After reading that introduction, the reader should know what HAL is and how it fits into what LinuxCNC is doing.

I agree that this file can be shortened a bit, but not necessarily on the parts I had added :)

andypugh:
I have made a pull-request to the source branch of this PR with some suggested changes.

smoe:
Thank you tons, @andypugh. Very nice. Merged and rebased on 2.9. I suggest to squash this all prior to a merge.

hansu:
Thanks Andy for proof reading and pushing that forward :+1:

PR #2064, opened 2022-09-23
Changed to regular format for man page cross references.

petterreinholdtsen:
Man page viewers like khelp are able to convert hostmot2(9) into a link to the hostmot2 manual page, while they do not understand 'man hostmot2' or similar notation.  Thus, converted references to the regular fomat used in manual pages.


hansu:
Never heard of khelp, but nice to have them consistent :+1:.

PR #2068, opened 2022-09-25
Gmoccapy sample configs

hansu:
- Added a sample config for gmoccapy demonstrating use of embedded pyngcgui. 
- Removed some duplicated sample configs. The question here is which to keep - those with "gmoccapy_" prefix or theose without? @gmoccapy added those without prefix more recently. Was the plan to have them all without prefix?


hansu:
I'll merge this now, the files can be renamed later if a different style is preferred.

gmoccapy:
I do not agree in deleting some files. IMHO the config with the simulated hardware button show easy the connections and also that glade panel can bei placed on fixed positions. And the user messages config shows how to use the Hal pin based user messages. This config is the best starting point for users after reading the documentation.

So there should be better phone config more.

Norbert

hansu:
I only deleted duplicates.
E,g, we had `gmoccapy_sim_hardware_button.ini` and `sim_hardware_button.ini`. I only deleted the last one.

But as I said we can have them all without the `gmoccapy_` prefix as well. But I chose with because there have been more with than without.

PR #2070, opened 2022-09-28
docs: misc (smoe:docs_misc_10)

smoe:
Series of smaller changes and fixes.

### DIFF
--- a/docs/src/hal/tools.adoc
+++ b/docs/src/hal/tools.adoc
@@ -93,8 +93,8 @@ Halshow (<<cha:halshow,complete usage description>>)
 can be started from the command line to show details for selected
 components, pins, parameters, signals, functions, and threads of a running HAL.
 The WATCH tab provides a continuous display of selected pin, parameters, and signal items.
-The "File" menu provides buttons to 1) save the watch items to
-a watch list and to load and existing watch list.
+The File menu provides buttons to save the watch items to
+a watch list and to load an existing watch list.
 The watch list items can also be loaded automatically on startup.
 For command line usage:
 
@@ -111,8 +111,6 @@ Notes:
   Create watchfile in halshow using: 'File/Save Watch List'.
   LinuxCNC must be running for standalone usage.
 ----
-
-LinuxCNC must be running for standalone usage.
 
 .Halshow Watch Tab
 image::images/halshow-4.png["Halshow Watch Tab",align="center"]
@@ -151,7 +149,7 @@ operating instructions should be located here eventually. For now,
 refer to section  <<sec:tutorial-halscope>> in the tutorial chapter,
 which explains the basics.
 
-The halscope "File" menu selector provides buttons to save a configuration
+The halscope File menu selector provides buttons to save a configuration
 or open a previously saved configuration.  When halscope is terminated,
 the last configuration is saved in a file named autosave.halscope.
 
--- a/docs/src/hal/tools.adoc
+++ b/docs/src/hal/tools.adoc
@@ -93,7 +93,7 @@ Halshow (<<cha:halshow,complete usage description>>)
 can be started from the command line to show details for selected
 components, pins, parameters, signals, functions, and threads of a running HAL.
 The WATCH tab provides a continuous display of selected pin, parameters, and signal items.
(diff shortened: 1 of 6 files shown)


docs/src/hal/tools.adoc
```diff
@@ -151,7 +151,7 @@ operating instructions should be located here eventually. For now,
 refer to section  <<sec:tutorial-halscope>> in the tutorial chapter,
 which explains the basics.
 
-The halscope File menu selector provides buttons to save a configuration
+The halscope "File" menu selector provides buttons to save a configuration
```
hansu, line 154: I think this would be more common:
```suggestion
The halscope file menu selector provides buttons to save a configuration
```

docs/src/hal/tools.adoc
```diff
@@ -305,9 +305,11 @@ APP = halreport /tmp/halreport.txt
...
 The function addf-ordering can be important for servo loops where
 the sequence of the functions computed at each servo period is
-important.  Typically, the order is: read input pins, do the
-motion command-handler and motion-controller functions, perform
-pid calculations, and finally write output pins.
+important.  Typically, the order is:
+. read input pins,
+. do the motion command-handler and motion-controller functions,
+. perform pid calculations, and finally
+. write output pins.
```
hansu, line 312: Do you had a list in mind here?

smoe: An enumerated list, yes.

docs/src/hal/tools.adoc
smoe: I saw it but then thought that the author should have that freedom and the original is formally correct.

smoe:
Quickly resolved the conflicts.

docs/src/hal/tools.adoc
hansu: But the output is no enumerated list unfortunately.

docs/src/hal/tools.adoc
hansu: But `"File" menu` looks a bit weird and never seen it like this.

smoe: I do not want to change it, but won't intervene if you are insisting on changing it. IMHO this only distracts from other problems that sentence has.

docs/src/hal/tools.adoc
smoe: Blank line was missing. Thanks! Fixed in my latest push.

docs/src/hal/tools.adoc
hansu: I would say F/file menu or menu "File"

smoe: Ah, I did not recall to have introduced those quotes. Hm. Right.

smoe:
Removed the quotes embracing "File" , in two occasions :) Also remove a redundant sentence.

docs/src/hal/tools.adoc
smoe: Wikipedia has it as 'File menu' (without any quotes), so I went for that.  Thank you!

hansu: :laughing: :see_no_evil:

hansu:
Thanks!

docs/src/hal/tools.adoc
silopolis: What about

```suggestion
The `halscope` 's `menu:File[Save configuration...] menu entry allows to save a configuration
```

smoe: I think that is overdoing it.

silopolis: Having a language allowing us to semantically markup elements specific to software UIs is a valuable strength IMHO.
Menus (and navigation within) as well as buttons need a defined formatting anyway. RH uses AsciiDoc special macros and puts any UI element in bold, whereas GitLab uses the ``*`Menu > Entry...`*`` pattern for navigation.

Halscope is either an application name or a command name. In the latter case it should be inline code blocked. In the former, we have to decide if we put some kind of emphase on applications and products names. Finally we have to devise (a) rule(s) whether to use one or the other depending on context.

PR #2072, opened 2022-09-29
Fix GitHub CI

SebKuzminsky:
This makes our Github Actions CI work again, by (sadly) not trying to build on Debian Unstable/Sid.

These are the bugs in Debian Unstable preventing us from building:
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1020893
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1008735

Not sure what the best way forward here is.  It's really important that we test our build on Debian Unstable (since that's what our "official" debian packages must target), but I really don't like having our CI totally stop for a week or whatever when Sid breaks.

As far as I know there's no way to mark a build as "failed, but let's keep going anyway":  https://docs.github.com/en/actions/creating-actions/setting-exit-codes-for-actions#about-exit-codes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        image: ["debian:buster", "debian:bullseye", "debian:bookworm", "debian:sid"]
+        image: ["debian:buster", "debian:bullseye", "debian:bookworm"]
     container:
       image: ${{ matrix.image }}
       # IPC_OWNER is needed for shmget IPC_CREAT
@@ -140,7 +140,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        image: ["debian:buster", "debian:bullseye", "debian:bookworm", "debian:sid"]
+        image: ["debian:buster", "debian:bullseye", "debian:bookworm"]
     container:
       image: ${{ matrix.image }}
       # IPC_OWNER is needed for shmget IPC_CREAT
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,10 +100,10 @@ jobs:
       run: |
         set -e
         set -x
-        apt-get --yes --quiet install --no-install-recommends gpg software-properties-common
+        apt-get --yes --quiet install --no-install-recommends gpg
         gpg --homedir="${PWD}/gnupg" --output /etc/apt/trusted.gpg.d/linuxcnc-deb-archive.gpg --export 3CB9FD148F374FEF
         DIST=$(echo ${{matrix.image}} | cut -d : -f 2)
-        add-apt-repository "deb http://linuxcnc.org $DIST base"
+        echo "deb http://linuxcnc.org $DIST base" >> /etc/apt/sources.list
         apt-get --quiet update
 
     - name: Build architecture-specific Debian packages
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -100,10 +100,10 @@ jobs:
       run: |
         set -e
         set -x
-        apt-get --yes --quiet install --no-install-recommends gpg software-properties-common
+        apt-get --yes --quiet install --no-install-recommends gpg
         gpg --homedir="${PWD}/gnupg" --output /etc/apt/trusted.gpg.d/linuxcnc-deb-archive.gpg --export 3CB9FD148F374FEF
         DIST=$(echo ${{matrix.image}} | cut -d : -f 2)
-        add-apt-repository "deb http://linuxcnc.org $DIST base"
+        echo "deb http://linuxcnc.org $DIST base" >> /etc/apt/sources.list
```
petterreinholdtsen, line 106: What about skipping the entire 'add extra repo' part on bookworm and unstable, is something like this:
```
case "$DIST" in
  bookwork|sid)
    ;;
  *)
     add_linuxcnc_org_repo
    ;;
esac
```

jepler:
An individual step can be permitted to fail depending on an expression. I ran the following in my actions playground:
```yaml
  build:
    strategy:
      matrix:
        exit-status:
        - 0
        - 1
        fail-ok:
        - false
        - true
    runs-on: ubuntu-latest
    steps:
    - run: "exit ${{matrix.exit-status}}"
      continue-on-error: ${{matrix.fail-ok}}
```

Visually [the results](https://github.com/jepler/actions-playground/actions/runs/3151367879) are a bit contradictory but they are doing their best to communicate what's going on
image
Of course if nobody goes and looks, nobody will notice so it's just wasted cycles.

Therefore you could write a `continue-on-error` expression that tests whatever elements of the `matrix` that you want to test, OR you could explicitly put it in the matrix. I think it would look like this:
```yaml
      matrix:
        image: ["debian:buster", "debian:bullseye", "debian:bookworm"]
        fail-ok: false
        include:
          - image: "debian:sid"  
          - fail-ok: true   
```

Ah, and here's github's own documentation, they use a slightly different way of wording it and apply "continue-on-error" to the whole workflow instead of a specific step:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-preventing-a-specific-failing-matrix-job-from-failing-a-workflow-run
```yaml
runs-on: ${{ matrix.os }}
continue-on-error: ${{ matrix.experimental }}
strategy:
  fail-fast: false
  matrix:
    node: [13, 14]
    os: [macos-latest, ubuntu-latest]
    experimental: [false]
    include:
      - node: 15
        os: ubuntu-latest
        experimental: true
```

smoe:
@petterreinholdtsen - I suggest you accept @SebKuzminsky 's patch and we see how we can reintroduce sid a bit later again?

petterreinholdtsen:
[Steffen Möller]
> @petterreinholdtsen - I suggest you accept @SebKuzminsky 's patch and
> we see how we can reintroduce sid a bit later again?

I am not convinced it is a good way forward.  I created 
<URL: https://github.com/LinuxCNC/linuxcnc/commits/fix-github-ci-pere >
with a different approach without the exit code matrix approach, instead
adding a ugly hack in d/configure to ensure it will build if we upload
to unstable right now.  It seem like a more useful target than only
fixing the github CI build.

I welcome feedback from @SebKuzminsky on this approach.  If he agree it
is a better way forward, I suggest it is merged into master.

-- 
Happy hacking
Petter Reinholdtsen

SebKuzminsky:
Merged https://github.com/LinuxCNC/linuxcnc/pull/2077 instead.

PR #2075, opened 2022-10-01
Gmoccapy: use icon theme for notifications and remove images folder

hansu:
In https://github.com/LinuxCNC/linuxcnc/pull/1548 the use of icon themes for gmoccapy was introduced.
Now I adapted the notification class to be able to use icon themes as well.
Therefore the images in the images folder are not needed any longer so i removed this folder.

@PeterMue:
1. Can you please have a look at the changes if I have removed everything that belongs to the images in the images folder?
2. I added three icons to the icon themes. For material only some examples as placeholder. I think it's better if you add them in https://github.com/PeterMue/gmoccapy-icon-themes

Furthermore I just copied the `_load_symbolic_from_icon_theme` in notifications.py from gmoccapy. I guess there is an elegant solution to re-use this function from gmoccapy without creating an extra class of file or something like that.


PeterMue:
I'll have a look the next couple days maybe but I can't promise so.

src/emc/usr_intf/gmoccapy/gmoccapy.py
```diff
@@ -146,12 +146,12 @@ def excepthook(exc_type, exc_obj, exc_tb):
 def _find_valid_icon_themes():
     valid_icon_themes = [
         # path, name
-        (None, "none")
```
PeterMue, line 149: Removing the "None" option leads to a probably unwanted problem: When having any icon theme (even no linux-cnc compatible - like a generic desktop icon theme) in the `USER_ICON_THEME_DIR` ("~/.icons") and no specific preference set, the first icon theme from `USER_ICON_THEME_DIR` gets picked up and most likely is incompatible (missing icons).

I can think of two ways to address this:
1. Reverse the order of directories to first load from the ICON_THEME_DIR
2. Assign a fixed default theme (classic for example)

src/emc/usr_intf/gmoccapy/gmoccapy.py
```diff
@@ -2024,9 +2024,8 @@ def _init_icon_themes(self):
                 print(f"Warning: preferred icon-theme '{icon_theme_preference}' not found; switching to 'default'.")
                 icon_theme_choice.set_active(0)
 
-        # switch theme if not default
```
PeterMue, line 2027: Same as already mentioned: i think we have to define a "default"

PeterMue (changes requested):
When removing the images folder we should also remove the references to these files in the glade file.
There are some image objects defined with a pixbuf reference to images in this folder.
I think it's save to just keep the image object in the glade file without any "picture" defined, so something like `sed '/pixbuf.*images\/.*\.[png|gif|svg]/g' -i gmoccapy.glade` should do the trick.

PeterMue:
> Furthermore I just copied the _load_symbolic_from_icon_theme in notifications.py from gmoccapy. I guess there is an elegant solution to re-use this function from gmoccapy without creating an extra class of file or something like that.

Well, i think the elegant solution would be to move this in a separate class but that applies to many other functions is this monster "gmoccapy" class as well 😆 

To just remove the duplication we could define the method dynamically on the instance like so:

```python
    # inside notification.py

    def _load_symbolic_from_icon_theme(self, icon_name, size, style = None ):
        # defined dynamically from gmoccapy
        pass
```

and then define a reference to the existing method when using the Notification class:

```python
        # inside gmoccapy.py

        self.notification = notification.Notification()  # Our own message system
        self.notification.load_symbolic_from_icon_theme = MethodType(self.load_symbolic_from_icon_theme, self.notification)
```

Doesn't make the whole thing less complex but keeps the duplications low... 

What do you think @hansu ?

hansu:
>When removing the images folder we should also remove the references to these files in the glade file.

:+1:

> I think it's save to just keep the image object in the glade file without any "picture" defined, so something like `sed '/pixbuf.*images\/.*\.[png|gif|svg]/d' -i gmoccapy.glade` should do the trick.

Isn't it mandatory to keep the image objects as they will be filled with the images from the icon theme?

src/emc/usr_intf/gmoccapy/gmoccapy.py
hansu: I would go for option 2

hansu:
> > Furthermore I just copied the _load_symbolic_from_icon_theme in notifications.py from gmoccapy. I guess there is an elegant solution to re-use this function from gmoccapy without creating an extra class of file or something like that.
> 
> Well, i think the elegant solution would be to move this in a separate class but that applies to many other functions is this monster "gmoccapy" class as well laughing
> 
> To just remove the duplication we could define the method dynamically on the instance like so:
> 
> ```python
>     # inside notification.py
> 
>     def load_symbolic_from_icon_theme(self, icon_name, size, style = None ):
>         # defined dynamically from gmoccapy
>         pass
> ```
> 
> and then define a reference to the existing method when using the Notification class:
> 
> ```python
>         # inside gmoccapy.py
> 
>         self.notification = notification.Notification()  # Our own message system
>         self.notification.load_symbolic_from_icon_theme = MethodType(self.load_symbolic_from_icon_theme, self.notification)
> ```
> 
> Doesn't make the whole thing less complex but keeps the duplications low...
> 
> What do you think @hansu ?

Yeah at least better than my approach with the duplicated function.
Or what if we create a class for the icon theme related stuff inside gmoccapy.py?
Or should we take the step and create a class for that in its own file as kind of a role model?

hansu:
However, thanks for your review!

PeterMue:
> > When removing the images folder we should also remove the references to these files in the glade file.
> 
> 👍
> 
> > I think it's save to just keep the image object in the glade file without any "picture" defined, so something like `sed '/pixbuf.*images\/.*\.[png|gif|svg]/d' -i gmoccapy.glade` should do the trick.
> 
> Isn't it mandatory to keep the image objects as they will be filled with the images from the icon theme?

Yes you're right, the parent object "GtkImage" must remain but the "pixbuf" property can be omitted.
e.g. for the img_auto:
```xml
  <object class="GtkImage" id="img_auto">
    <property name="visible">True</property>
    <property name="can_focus">False</property>
    <property name="pixbuf">images/auto_mode.png</property>
  </object>
```
Just drop the line `<property name="pixbuf">images/auto_mode.png</property>` should be ok.

hansu:
@PeterMue 
1. pixbuf property removed
2. Do you agree with the setting of the default icon theme?
3. I created a class for some icon theme functions, but it cannot be found by the notification class. Do you think this is generally possible to do it this way?
4. Does it have a special reason, that you append the theme_dir here but never use it? https://github.com/LinuxCNC/linuxcnc/blob/e8e07554801cb1b2fc885521edae0a2c5d459e52/src/emc/usr_intf/gmoccapy/gmoccapy.py#L155

PeterMue:
@hansu 

> 3. I created a class for some icon theme functions, but it cannot be found by the notification class. Do you think this is generally possible to do it this way?

I've moved the icon related stuff to it's own file see https://github.com/hansu/linuxcnc/pull/8

> 2. Do you agree with the setting of the default icon theme?

No, i think there's a bug right now. The selected theme does not correspond with the selected element in the dropdown.
I'll have a look at this in a few minutes.

> 4. Does it have a special reason, that you append the theme_dir here but never use it?

No i don't think so. But it would not surprise me if i added this with the intention to be handy later on - and forgot about it.

hansu:
>I've moved the icon related stuff to it's own file see https://github.com/hansu/linuxcnc/pull/8

Thanks!


>> Does it have a special reason, that you append the theme_dir here but never use it?
>
>No i don't think so. But it would not surprise me if i added this with the intention to be handy later on - and forgot about it.

Then it's maybe better to remove it there. it confused me a bit, because I thought the path was taken from there.

PeterMue:
Ah 💡 wait a second. I was just about to remove the path but then i came across the model in the glade file.
The zero-index of each entry is the unique key for the model, thats the reason why i chose the path as this is pretty uniq (instead of the name).
So we should keep this.

(Although it's probably not visible to the user which theme in the combobox belongs to which folder, it's at least not crashing anything)

PR #2077, opened 2022-10-02
Add workaround for broken lsb_release in current Debian Sid/Bookworm

SebKuzminsky:
No longer add linuxcnc.org APT repo for unstable and bookworm builds and add override to debian/configure and update-dch-from-git to get sid/bookworms building until <URL: https://bugs.debian.org/1008735 > is fixed.


SebKuzminsky:
Thanks @petterreinholdtsen !

PR #2080, opened 2022-10-04
Tooledit: Correct data type and validate input for orientation + warn if out of range

### LINKED ISSUE #1773: Gmoccapy: gremlin crashes on wrong value for orientation in tool table
If you enter in a tool table of lathe configuration a value for the orientation that exceeds the range of glcanon's `lathe_shapes` (0-9), and try to change to this tool, you get following error and gremlin is dead until you restart gmoccapy:
```
  File "/home/cnc/linuxcnc/linuxcnc-dev-master/lib/python/gremlin.py", line 275, in expose
    else: self.redraw_ortho()
  File "/home/cnc/linuxcnc/linuxcnc-dev-master/lib/python/rs274/glcanon.py", line 356, in inner
    return f(self, *args, **kw)
  File "/home/cnc/linuxcnc/linuxcnc-dev-master/lib/python/rs274/glcanon.py", line 649, in redraw_ortho
    self.redraw()
  File "/home/cnc/linuxcnc/linuxcnc-dev-master/lib/python/rs274/glcanon.py", line 1368, in redraw
    self.cache_tool(current_tool)
  File "/home/cnc/linuxcnc/linuxcnc-dev-master/lib/python/rs274/glcanon.py", line 1468, in cache_tool
    self.lathetool(current_tool)
  File "/home/cnc/linuxcnc/linuxcnc-dev-master/lib/python/rs274/glcanon.py", line 1799, in lathetool
    dx, dy = self.lathe_shapes[orientation]
```

For 2.8 you only get a an empty dialog (see https://forum.linuxcnc.org/gmoccapy/46182-radius-compensation-lathe)

One solution would be to catch this case inside glcanon: https://github.com/LinuxCNC/linuxcnc/blob/master/lib/python/rs274/glcanon.py#L1792-L1800
But how then to pass this to gmoccapy?

Another solution would be to validate the input of the tool table, like it is done in Axis.

hansu:
I corrected the data type to int and changed the tool table to allow only input from 0-9.
When a wrong value is in the tool table (orientation in degree) a warning is popping up and saving will not be allowed until a valid value is entered.
I am not sure if this is a bit too overdone to create a dialog for this rare case.
I also could throw an exception that would be shown in the general exception dialog...

Resolves #1773

### DIFF
--- a/lib/python/gladevcp/tooledit_gtk.glade
+++ b/lib/python/gladevcp/tooledit_gtk.glade
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <interface>
-  <requires lib="gtk+" version="2.18"/>
+  <requires lib="gtk+" version="3.0"/>
   <!-- interface-naming-policy project-wide -->
   <object class="GtkListStore" id="liststore1">
     <columns>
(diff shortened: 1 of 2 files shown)


c-morley (commented):
Been a while since being in this code, but I think your changes look good and are a great idea.
Only one thing came to mind, that probably would be answered if I studied the full code a bit but...

line 297 was removed:
liststore = self.model
Is liststore defined somewhere else?

hansu:
Thanks for having a look.

I moved  `liststore = self.model` to the beginning of the function (line 321), because i needed it for the "pre check"

PR #2081, opened 2022-10-04
Minor fixes

SebKuzminsky:
A couple of minor fixes that presented themselves recently, nothing too controversial I think.


jepler (approved):
thanks

PR #2084, opened 2022-10-08
docs. use ASCII art for images in man pages

hansu:
@petterreinholdtsen and I came to the conclusion to not use images in man pages. The only two images of timing diagrams existing in the man pages can be represented in ASCII. 

I think this can close #1314.


docs/man/images/Images_info.adoc
```diff
@@ -0,0 +1,54 @@
+# Images in man pages
+
+Images in man pages are not recommended. It's better to use a ASCII representation (ASCII art) if possible.
+
+Images can be added in the man pages for representation as HTML and PDF
+The images have to be in *.ps-format and should be placed only in this folder.
+They can be included using `.PSPIC -L ../man/images/image.ps`.
+The PNG needed for the HTML view is automatically created on generating LinuxCNC_Manual_Pages.pdf
```
petterreinholdtsen, line 8: Why is this section here, with a recipe on how to do it, if it is not recommended?

petterreinholdtsen (approved):
The component man pages look good to me, but the Image_info.adoc file confuses me.  I believe this is fine to be merged, though.

docs/man/images/Images_info.adoc
hansu: It was meant as additional information, but if it's confusing, I'll remove it.

petterreinholdtsen:
[Hans Unzner]
> It was meant as additional information, but if it's confusing, I'll
> remove it.

I read it as instructions to do something we do not want done, which
seem rather counter productive to me.   Not confusing at such, but if
someone really want to do it, I guess they can figure out how on their
own. :)
-- 
Happy hacking
Petter Reinholdtsen

PR #2090, opened 2022-10-12
Update ini-homing.adoc

satiowadahc:
This font seems to change between OS/browser. Enclosing in a code block should standardize on monospace fonts.

### DIFF
--- a/docs/src/config/ini-homing.adoc
+++ b/docs/src/config/ini-homing.adoc
@@ -380,7 +380,7 @@ Axis gui, then the following INI entries for that joint are needed.
 HOME_SEARCH_VEL = 0
 HOME_LATCH_VEL = 0
 HOME_USE_INDEX = NO
-HOME equals to HOME_OFFSET
+HOME_OFFSET = 0 (Or the home position offset)
 HOME_SEQUENCE = 0 (or other valid sequence number)
 ```
 
--- a/docs/src/config/ini-homing.adoc
+++ b/docs/src/config/ini-homing.adoc
@@ -380,7 +380,7 @@ Axis gui, then the following INI entries for that joint are needed.
 HOME_SEARCH_VEL = 0
 HOME_LATCH_VEL = 0
 HOME_USE_INDEX = NO
-HOME_OFFSET = 0 (Or the home position offset)
+HOME_OFFSET = 0 (Or the home position offset (HOME))
 HOME_SEQUENCE = 0 (or other valid sequence number)
 ----
 
--- a/docs/src/config/ini-homing.adoc
+++ b/docs/src/config/ini-homing.adoc
@@ -376,11 +376,13 @@ home position like a rotary joint and you want that joint to home at
 the current position when the "Home All" button is pressed in the
 Axis gui, then the following INI entries for that joint are needed.
 
-. HOME_SEARCH_VEL = 0
-. HOME_LATCH_VEL = 0
-. HOME_USE_INDEX = NO
-. HOME equals to HOME_OFFSET
-. HOME_SEQUENCE = 0 (or other valid sequence number)
+```
+HOME_SEARCH_VEL = 0
+HOME_LATCH_VEL = 0
+HOME_USE_INDEX = NO
+HOME equals to HOME_OFFSET
+HOME_SEQUENCE = 0 (or other valid sequence number)
+```
 
 [NOTE]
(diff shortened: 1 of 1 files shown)


hansu:
Good idea to have it as code block, but then it should meet the INI file syntax.

docs/src/config/ini-homing.adoc
```diff
@@ -376,11 +376,13 @@ home position like a rotary joint and you want that joint to home at
...
-. HOME_USE_INDEX = NO
-. HOME equals to HOME_OFFSET
-. HOME_SEQUENCE = 0 (or other valid sequence number)
+```
+HOME_SEARCH_VEL = 0
+HOME_LATCH_VEL = 0
+HOME_USE_INDEX = NO
+HOME_OFFSET = 0 (Or the home position offset)
+HOME_SEQUENCE = 0 (or other valid sequence number)
+```
```
hansu, line 385:
```suggestion
----
HOME_SEARCH_VEL = 0
HOME_LATCH_VEL = 0
HOME_USE_INDEX = NO
HOME_OFFSET = 0 (Or the home position offset)
HOME_SEQUENCE = 0 (or other valid sequence number)
----
```

Sorry I missed that the first time - this is the correct syntax for Asciidoc.

c-morley:
HOME_OFFSET = 0 (Or the home position offset)

That doesn't seem clear.

My read from docs is:
 HOME_OFFSET must be equal  to the HOME setting
or said another way:
 HOME_OFFSET and  HOME must be set to the same value

HOME is not an position offset, it's just a position.

hansu:
Maybe in this way:?

    HOME_SEARCH_VEL = 0
    HOME_LATCH_VEL = 0
    HOME_USE_INDEX = NO
    HOME = <home_position>        
    HOME_OFFSET = <home_position>   # Defaults to 0
    HOME_SEQUENCE = 0               # Or other valid sequence number

andypugh:
HOME_OFFSET and HOME don't have to be the same. 

HOME_OFFSET is where the home switch is relative to the machine origin (ie, the zero position). 

HOME is simply where the joint moves to at the end of the homing sequence.

c-morley:
In the context of the paragraph, I believe they are instructing how to home and set home position without the joint moving, in which case I think its right - home offset and home position must be the same.

satiowadahc:
I'm indifferent to the content, issue was the resulting text that changed due to different browsers/OS systems that made it illegible between `o/O/0`

hansu:
[Andy]
>HOME is simply where the joint moves to at the end of the homing sequence.

Yeah but under "Immediate Homing" I would understand no machine movement at all while homing.


I think valid combinations to use without home switches are either
(as far as I could test in the simulation)
- HOME_OFFSET = HOME

or
- HOME_OFFSET = 0 and HOME = <some_value>



Where the first one homes the joint at the current position and **sets** the position to HOME and the latter homes at the current position and **moves** the joint to HOME.

andypugh:
I am thinking that I should have gone to the trouble of looking in the base documents in search of context rather than guessing what the question was based only on this commit.

satiowadahc:
rn_image_picker_lib_temp_53a1f9a4-98e0-45ad-b024-e6ded2ed4c56

Viewing with firefox on windows 10.

docs/src/config/ini-homing.adoc
```diff
@@ -376,11 +376,13 @@ home position like a rotary joint and you want that joint to home at
...
-. HOME_SEARCH_VEL = 0
-. HOME_LATCH_VEL = 0
-. HOME_USE_INDEX = NO
-. HOME equals to HOME_OFFSET
-. HOME_SEQUENCE = 0 (or other valid sequence number)
+----
+HOME_SEARCH_VEL = 0
+HOME_LATCH_VEL = 0
+HOME_USE_INDEX = NO
+HOME_OFFSET = 0 (Or the home position offset)
```
hansu, line 383:
```suggestion
HOME_OFFSET = 0 (Or the home position offset (HOME))
```

hansu (commented):
So I think then it's fine. @c-morley do you agree?

And @satiowadahc the commits can be squashed together.

satiowadahc:
> So I think then it's fine. @c-morley do you agree? Maybe this little change.
> 
> And @satiowadahc the commits can be squashed together.

Does github not have an option to squash on pr?

hansu:
> Does github not have an option to squash on pr?

It has, but it's disabled for this repository. So you have to do it manually.

PR #2094, opened 2022-10-14
Update m-code.adoc

satiowadahc:


docs/src/gcode/m-code.adoc
```diff
@@ -428,7 +428,7 @@ M67 E- Q-
 ----
 
 * 'M67' - set an analog output synchronized with motion.
-* 'E-' - output number ranging from 0 to 3.
+* 'E-' - output number ranging from 0 to 3 (Adjustable from motmod argument num_aio)
```
smoe, line 431: A "." went missing, but I suggest to accept the patch nonetheless and fix the syntax later.  Maybe another sentence somewhere where to read more on motmod and num_aio would be good, too.

PR #2096, opened 2022-10-15
Mb2hal: add version number for backward compatibilty

hansu:
Some time ago I added some functionality to MB2HAL and therefore changed some pin names (https://github.com/LinuxCNC/linuxcnc/pull/1030, https://linuxcnc.org/docs/2.9/html/drivers/mb2hal.html#_pins).
So this might break a configuration working with 2.8.
To avoid this, this PR adds a version number to the MB2HAL INI file. If this is not set, MB2HAL behaves like before.
If set so 1.1, the new features are "unlocked" and the changed pin names are applied.

f25ed1307a1c083a6cb0a6d5a8227bb011c19f07 is technically not needed to ensure backward compatibility, but as these functions belong to 1.1, they should be available only there.

### DIFF
--- a/docs/man/man1/mb2hal.1
+++ b/docs/man/man1/mb2hal.1
@@ -103,9 +103,9 @@ m = HAL_TX_NAME or transaction number if not set, n = element number (NELEMENTS)
 .br
 Example: 
 .br
-mb2hal.00.01.<type> (transaction=00, second register=01 (00 is the first one))
+mb2hal.00.01.int (TRANSACTION_00, second register)
 .br
-mb2hal.TxName.01.<type> (HAL_TX_NAME=TxName, second register=01 (00 is the first one))
+mb2hal.readStatus.01.bit (HAL_TX_NAME=readStatus, first bit)
 
 .SH AUTHOR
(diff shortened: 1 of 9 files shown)


petterreinholdtsen:
[Hans Unzner]
> Some time ago I added some functionality to MB2HAL and therefore
> changed some pin names
> (https://github.com/LinuxCNC/linuxcnc/pull/1030,
> https://linuxcnc.org/docs/2.9/html/drivers/mb2hal.html#_pins).

It seem like pull request #1030 had a proposal to make sure the old pin
names kept working.  Were this change never implemented?  It would
ensure there is a migration time for existing setup.  It seem like a
better approach than the 'flag dag' versioning approach in this pull
request, as it would allow new and old setups to work in parallel.
--
Happy hacking
Petter Reinholdtsen

hansu:
> It seem like pull request [#1030](https://github.com/LinuxCNC/linuxcnc/pull/1030) had a proposal to make sure the old pin
names kept working.  Were this change never implemented?

Not really. It was more a question if I should include these changes or not.

PR #2097, opened 2022-10-15
presence of rtai modules does not indicate realtime is running

### LINKED ISSUE #2095: RTAI: LinuxCNC keeps saying to run `halrun -U`

 1. Compile LinuxCNC against RTAI
 2. Run latency-histogram, latency-test or LinuxCNC (I only used run-in-place for this test)
 3. Try to re-run any of those programs

Application would start back up

RTAI is already running, run `halrun -U` which doesn't work. I have to manually run `rmmod rtai_sched rtai_hal rtai_math`

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`): Debian 11/Bullseye
 * I am using this kernel version (shown by `uname -a`): 4.19.2XX
 * I am running ...
   * [X] A binary I built myself
 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`): master checkouts from the last few weeks
 * I am using this user interface (GUI) (e.g., AXIS, Touchy, gmoccapy, etc): Any
 * I am using this interface hardware vendor and chipset (e.g., parallel port, ethernet port, FPGA card): N/A

jepler:
I hope this Closes #2095 though there may be other instances of incorrect checking for modules matching "rtai" being loaded.

### DIFF
--- a/scripts/latency-histogram
+++ b/scripts/latency-histogram
@@ -298,12 +298,6 @@ proc load_packages {} {
   }
 
   if [info exists ::LH(rtai)] {
-    if [catch {exec lsmod | grep rtai} msg] {
-      # puts ok_to_start_rtai
-    } else {
-      popup "RTAI is already running, (try: \$ halrun -U)"
-      exit 1
-    }
     exec $::LH(realtime) start &
     progress "Delay for realtime startup"
     after 1000 ;# wait to load Hal package
--- a/scripts/latency-plot
+++ b/scripts/latency-plot
@@ -173,15 +173,6 @@ proc init_hal_timedelta {} {
 } ;# init_hal_timedelta
 
 proc check {} {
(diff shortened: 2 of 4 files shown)


jepler:
I may need more testing or I may be wrong about the ability to "safely" leave rtai_sched loaded when using kmod+rtai. I can repeatedly start linuxcnc with axis (well, 4 times) but >50% of the time starting latency-histogram again hangs.

If you want to revert the part where we unload this for rtai, I won't stand in the way. It'll continue preventing us from testing kmod+rtai on buildbot, but I believe uspace+rtai is the way forward.

jepler:
For reasons discussed in https://github.com/LinuxCNC/linuxcnc/issues/2095#issuecomment-1279843168 I've also added a reversion of the module unloading for kmod+rtai situations.

Issue #2098, opened 2022-10-15
uspace+rtai does not provide rate monotonic scheduling, apparently


jepler:
In a VM with 4.19.195-rtai-amd64 configured for uspace+rtai I get occasional (less than 1 in 50) failures in threads.0, the test which is intended to reveal lapses in rate-monotonic-scheduling
```line 937: got 0, expected 10 or 1```

andypugh:
Can you define "rate monotonic scheduling"? 

I know what the words mean, but not what them mean in that combination.

jepler:
For correctness, we depend on the *fast* thread never being interrupted by the *slow* thread. This is important to just two components, but without it they will work erratically: the software stepgen & software encoder.

andypugh:
Thanks

PR #2099, opened 2022-10-16
uspace+rtai: fix priority

jepler:
rtai priority numbers run opposite to pthread priority numbers.

After this change, no threads.0 failures for 1000 tests. Before, failure probability was at least 1 in 50 tests.

Closes: #2098


SebKuzminsky:
I haven't tested it, but this looks good to me.

PR #2102, opened 2022-10-20
docs: reformatted pin names

smoe:
Some typos, loads of eye candy to sync with style sheet (pin names should be monospaced), and helping the translation with punctuation and more semantic line breaks.


hansu (changes requested):
Unfortunately our asciidoc to man page converter doesn't support the monospace notation.
It renders like this:
grafik

So please remove this formatting.

PR #2104, opened 2022-10-21
docs: + -> , corrections (smoe:docs_misc_6)

smoe:


docs/src/gui/gladevcp.adoc
```diff
@@ -1044,13 +1044,13 @@ Insensitized children will not respond to input.
...
 
-<Panel_basename>.<widgetname>:: FIXME pin name
+`_<Panel_basename>_._<widgetname>_`:: FIXME pin name
   If the pin is low then child widgets are visible which is the default state.
 
 `hal_table` and `hal_hbox` have one HAL BIT input pin which controls if their child widgets are sensitive or not. +
 These widgets's pin name uses the convention:
 
-<Panel_basename>.<widgetname>:: FIXME pin name
+`_<Panel_basename>_._<widgetname>_`:: FIXME pin name
```
hansu, line 1053: I am pretty sure this renders as
`_<Panel_basename>_._<widgetname>_`

hansu (changes requested):

smoe:
asciidoctor tag? I do not recall about how many of these are in there. Since there is a FIXME anyway, if there are not too many others, I suggest to just leave it.

hansu:
> asciidoctor tag? I do not recall about how many of these are in there. Since there is a FIXME anyway, if there are not too many others, I suggest to just leave it.

Maybe this is the chance to figure out the correct pin names? ;-)

docs/src/gui/gladevcp.adoc
smoe: Done.

smoe:
> > asciidoctor tag? I do not recall about how many of these are in there. Since there is a FIXME anyway, if there are not too many others, I suggest to just leave it.
> 
> Maybe this is the chance to figure out the correct pin names? ;-)

I would feel uneasy about it. Someone who knows what is going on should complete that.

hansu:
>I would feel uneasy about it. Someone who knows what is going on should complete that.

Ok, done :heavy_check_mark:

PR #2105, opened 2022-10-21
docs: add some info on tuning Preempt-RT for latency

SebKuzminsky:
Tuning the kernel for latency is an important step that we currently don't talk about at all in the docs.  I'm not sure this is the best place for it, it may belong somewhere in the "Integrator's Manual", I'm open to suggestions here.

### DIFF
--- a/docs/src/install/latency-test.adoc
+++ b/docs/src/install/latency-test.adoc
@@ -4,11 +4,9 @@
 [[cha:latency-testing]]
 = Latency Testing(((Latency Testing)))
 
-[[sec:latency-test]]
-== Latency Test(((Latency Test)))
 
-This test is the first test that should be performed on a PC
-to see if it is able to drive a CNC machine.
+[[sec:what-is-latency]]
+== What is latency?
 
 Latency is how long it takes the PC to stop what it is doing and respond
 to an external request, such as running one of LinuxCNC's periodic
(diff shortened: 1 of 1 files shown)


hansu:
Nice :+1: 
I think it fits well in the RT Kernel subsection, but I wouldn't expect to find it in the System Requirements section.
Maybe just add a link in http://linuxcnc.org/docs/html/install/latency-test.html?

SebKuzminsky:
Good point @hansu, I agree.  I moved this new info to the "Latency Test" section as you suggested.

The "Latency Test" document seems slightly misplaced though, it's the only file in `docs/src/install`.  At some point (not as part of this PR) we should maybe move that file to `docs/src/integrator`.  And at the same time maybe rename it to just "Latency", since it covers not just testing now.

silopolis:
Le dim. 23 oct. 2022 à 17:20, Sebastian Kuzminsky ***@***.***>
a écrit :

> Good point @hansu <https://github.com/hansu>, I agree. I moved this new
> info to the "Latency Test" section as you suggested.
>
> The "Latency Test" document seems slightly misplaced though, it's the only
> file in docs/src/install. At some point (not as part of this PR) we
> should maybe move that file to docs/src/integrator. And at the same time
> maybe rename it to just "Latency", since it covers not just testing now
>

All installation, configuration and administration docs should be moved to
integrator guide.

User docs should only hold operator and cnc programmer targeted content

Issue #2106, opened 2022-10-22
RTAI: No modules built against 5.4 kernel


NTULINUX:
## Here are the steps I follow to reproduce the issue:

 1. Build RTAI with 5.4 kernel (this is still in progress for other people to use, it's a bit of a hack right now)
 2. Build LinuxCNC against said kernel via `make`

## This is what I expected to happen:

LinuxCNC kernel modules compile or at least attempt to compile against the 5.4 kernel

## This is what happened instead:

Everything compiles except the kernel modules. First the error was:

```
Reading 204/204 dependency files
Done reading dependencies
MAKEFLAGS="ik -j4 --jobserver-auth=3,4" \
make KBUILD_EXTRA_SYMBOLS=/usr/realtime/modules/Module.symvers -C /home/ntu/linux-5.4.y SUBDIRS=`pwd` CC=gcc V=0 modules
make[1]: Entering directory '/home/ntu/linux-5.4.y'
 CALL    scripts/atomic/check-atomics.sh
 CALL    scripts/checksyscalls.sh
 CHK     include/generated/compile.h
 Building modules, stage 2.
 MODPOST 397 modules
make[1]: Leaving directory '/home/ntu/linux-5.4.y'
cp Module.symvers *.ko ../rtlib/
cp: cannot stat 'Module.symvers': No such file or directory
cp: cannot stat '*.ko': No such file or directory
make: [Makefile:540: modules] Error 1 (ignored)
You now need to run 'sudo make setuid' in order to run in place.
```

Then with this change:

```
diff --git a/src/Makefile b/src/Makefile
index 96eba38334..374b4bf35d 100644
--- a/src/Makefile
+++ b/src/Makefile
@@ -538,7 +538,7 @@ ifeq ($(BUILD_SYS),kbuild)
modules:
       MAKEFLAGS="$(filter-out --warn-undefined-variables,$(MAKEFLAGS))" \
       $(PYTHON) modsilent.py $(MAKE) KBUILD_EXTRA_SYMBOLS=$(moduledir)/Module.symvers -C $(KERNELDIR) SUBDIRS=`pwd` CC=$(CC) V=$(BUILD_VERBOSE) modules
-       -cp Module.symvers *$(MODULE_EXT) ../rtlib/
+       -cp $(moduledir)/Module.symvers $(moduledir)/*$(MODULE_EXT) ../rtlib/
endif
```

The problem is:

```
Reading 204/204 dependency files
Done reading dependencies
[...]

PR #2108, opened 2022-10-23
RTAI: Fix build with kernel 5.4

NTULINUX:
Note: Always better to use $() instead of ``

See:

https://www.shellcheck.net/wiki/SC2006


NTULINUX:
Fixes https://github.com/LinuxCNC/linuxcnc/issues/2106

andypugh:
Had to do this on my phone, my Macs all don’t show the buttons for merging due to old browsers (at a guess)

PR #2112, opened 2022-10-24
Changed udev rules to only grant access using TAG+="uaccess"

petterreinholdtsen:
This grant access to users logged in locally via the console.

Remove access using read or write file permissions for everyone or the plugdev group.

### DIFF
--- a/debian/extras/lib/udev/rules.d/99-shuttle.rules
+++ b/debian/extras/lib/udev/rules.d/99-shuttle.rules
@@ -1,11 +1,11 @@
 # ShuttleXpress USB jog pendant
 # model "S-XPRS"
-SUBSYSTEM=="hidraw", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0020", MODE="0444"
+SUBSYSTEM=="hidraw", ATTRS{idVendor}=="0b33", ATTRS{idProduct}=="0020", MODE="0400", TAG+="uaccess"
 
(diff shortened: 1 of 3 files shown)


smoe:
Would prefer to assess the effects on our community in a 2.10 branch and have this backported if successful. Remote execution may be a thing.

andypugh:
Do the devices in question operate via device events? Are those forwarded by ssh (etc) to HAL? 
I think that it would be an odd thing to expect to work, and anyone doing that would be at liberty to edit the udev rules to make it work (if it can)

petterreinholdtsen:
[andypugh]
> Do the devices in question operate via device events? Are those
> forwarded by ssh (etc) to HAL?

No.  The device events from the kernel to udev, which are affected by
the udev rules, are only used to create the device files in /dev/.

> I think that it would be an odd thing to expect to work, and anyone
> doing that would be at liberty to edit the udev rules to make it work
> (if it can)

The change make sure 'console users', ie users logged in on the machine
using local terminals or X get access, but not 'remote users' ie users
logged inn via ssh, telnet, rdf or similar.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
I am struggling to think of a problem scenario. These are devices that pass events in to HAL (via USB HID) so need to be plugged in to the controller. What could anyone possibly do with remote access to them? They only do anything with _physical_ access to twirl the knobs

PR #2115, opened 2022-10-29
Fixed typo in gmoccapy.adoc (directo->direcory)

petterreinholdtsen:
I am not sure if I understood what is being expressed here, so not sure the correction is correct.


hansu:
Good catch! It should actually be "one directory level up"
It seems i cut away the whole column from the 11th character.
Do you also want to fix the rest of the column? It should be:

```
 go to home directory
 one directory level up

 move selection left
 move selection right
 jump to directory as set in settings

 select / ENTER

 back
```

petterreinholdtsen:
[Hans Unzner]
> Good catch! It should actually be "one directory level up"
> It seems i cut away the whole column from the 11th character.
> Do you also want to fix the rest of the column? It should be:

I believe I fixed them all.

-- 
Happy hacking
Petter Reinholdtsen

PR #2119, opened 2022-10-30
Added a div2.comp for dividing a float

Roguish000:
by a float to get a float.

### DIFF
--- a/src/hal/components/invert.comp
+++ b/src/hal/components/invert.comp
@@ -25,7 +25,7 @@ param rw float deadband "The \\fBout\\fR will be zero if \\fBin\\fR is between -
 function _;
 license "GPL";
 author "Stephen Wille Padnos";
-see_also " invert, div2 "
+see_also " invert, div2 ";
 ;;
 
 #include <rtapi_math.h>
--- a/src/hal/components/invert.comp
+++ b/src/hal/components/invert.comp
@@ -25,6 +25,7 @@ param rw float deadband "The \\fBout\\fR will be zero if \\fBin\\fR is between -
 function _;
 license "GPL";
 author "Stephen Wille Padnos";
+see_also " invert, div2 "
 ;;
 
 #include <rtapi_math.h>
--- a/src/hal/components/invert.comp
+++ b/src/hal/components/invert.comp
@@ -25,7 +25,7 @@ param rw float deadband "The \\fBout\\fR will be zero if \\fBin\\fR is between -
 function _;
 license "GPL";
 author "Stephen Wille Padnos";
-see_also " invert, div2 ";
+see_also " invert(9), div2(9) ";
 ;;
 
 #include <rtapi_math.h>
--- a/src/hal/components/mult2.comp
+++ b/src/hal/components/mult2.comp
@@ -5,6 +5,8 @@ pin out float out "out = in0 * in1";
 function _;
 license "GPL";
 author "John Kasunich";
+see_also " invert, div2 ";
+
 ;;
 FUNCTION(_) {
     out = in0 * in1;
--- a/src/hal/components/mult2.comp
+++ b/src/hal/components/mult2.comp
@@ -5,7 +5,7 @@ pin out float out "out = in0 * in1";
 function _;
 license "GPL";
 author "John Kasunich";
-see_also " invert, div2 ";
+see_also " invert(9), div2(9) ";
 
 ;;
 FUNCTION(_) {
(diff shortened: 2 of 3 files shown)


andypugh:
This has always been possible with invert and mult, though not obvious or elegant. 
I believe that your divide-by-zero handling could be more sophisticated, however. If you look at the "invert" component that outputs a zero if the denominator is too close to zero. (which does not make an mathematical sense, but is possibly less troublesome)

Firstly, your component will output the error message every thread cycle whilst the input is zero, which can cause problems all of its own. It really should output the error once, then reset, then (possibly) output it again as the input value changes. 

Also, NaN "pollutes" everything it touches, and can itself cause irrecoverable errors. Outputting a zero (as invert does) is mathematically wrong, but easier to recover from. 

Another option might be to simply not update the output if the denominator is zero. 

Possibly different handling it required if the numerator is _also_ zero.

Roguish000:
andypugh,  not sure if you recall, but you actually assisted me with this comp very late one night a year or so ago.
yes, the error checking could probably more eloquent, but I am not a coder... it's my best effort.  Also, everyone should know not to divide by zero....
Please let it go and get it in for the next build round.
I can certainly continue to work on it...........

andypugh:
The alternative to NAN is + / - Inf (NAN is only really 0/0) 
Printing the error message every thread cycle is probably a bigger issue.

petterreinholdtsen:
[andypugh]
> The alternative to NAN is + / - Inf (NAN is only really 0/0)

It is an alternative, but not a very good one, as the answer
mathamatically is undefined, not +- infinity.  I suspect modelling it
after the invert component is probably a better idea, allowing the two
components to be compatible.

> Printing the error message every thread cycle is probably a bigger
> issue.

Absolutely.  We were not aware of this issue.  I'm helping Roguish to
create a new and improved patch.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
OK, I see clamping to a maximum magnitude as you now have is more mathematically satisfying. Though I might have allowed a smaller divisor, maybe 1e-12. And also the error message has "DEVISOR" rather than "DIVISOR"

src/hal/components/invert.comp
```diff
@@ -25,6 +25,7 @@ param rw float deadband "The \\fBout\\fR will be zero if \\fBin\\fR is between -
 function _;
 license "GPL";
 author "Stephen Wille Padnos";
+see_also " invert, div2 ";
```
petterreinholdtsen, line 28: This should be see_also " invert(9), div2(9) ";

Also, perhaps sort the list alphabetically?

src/hal/components/mult2.comp
```diff
@@ -5,6 +5,8 @@ pin out float out "out = in0 * in1";
 function _;
 license "GPL";
 author "John Kasunich";
+see_also " invert, div2 ";
```
petterreinholdtsen, line 8: This should be see_also " invert(9), div2(9) ";

c-morley: I'm curious, what's with the (9) part?

Also might I gently suggest that if we over ask for little changes that in the big picture don't really matter, we risk discouraging further pull requests. I've done this myself and thought I'd mention it.
I might just point out the preferred way - for future pull requests.

The NaN and endlesserror messages was the import part to fix (and thanks for doing it)

petterreinholdtsen:
[c-morley]
>> @@ -5,6 +5,8 @@ pin out float out "out = in0 * in1";
>  function _;
>  license "GPL";
>  author "John Kasunich";
> +see_also " invert, div2 ";
>
> I'm curious, what's with the (9) part?

It make sure the text show up as invert(9) in the manual page, allowing
man page parsers like khelp to recognize the string as a reference to
another manual page and convert the string to a link going directly to
the invert(9) manual page.  It is important for the usability of the
manual page cross referencing.

See for example the tac(1) link on
<URL: https://www.man7.org/linux/man-pages/man1/cat.1.html >.  It would
not be a link if the text was only 'tac' without the (1).

-- 
Happy hacking
Petter Reinholdtsen

Roguish000:
to all.    I modified the deadband value to 1e-12, fixed the spelling, and add the (9) to the see_also values in both mult2 and invert comps.
Anything else????

c-morley:
> [c-morley]
> > @@ -5,6 +5,8 @@ pin out float out "out = in0 * in1"; function _; license "GPL"; author "John Kasunich"; +see_also " invert, div2 "; I'm curious, what's with the (9) part?
> It make sure the text show up as invert(9) in the manual page, allowing man page parsers like khelp to recognize the string as a reference to another manual page and convert the string to a link going directly to the invert(9) manual page. It is important for the usability of the manual page cross referencing. See for example the tac(1) link on <URL: https://www.man7.org/linux/man-pages/man1/cat.1.html >. It would not be a link if the text was only 'tac' without the (1).
> […](#)
> -- Happy hacking Petter Reinholdtsen

Oh cool. Had no idea.

andypugh:
Thanks

PR #2127, opened 2022-11-06
packaging: mark linuxcnc-doc-de Multi-Arch: Foreign

SebKuzminsky:
This makes it match the other docs packages.  "Multi-Arch: Foreign" lets this architecture-independent binary package satisfy dependencies of binary packages of other architectures (i.e. in a multi-arch system).


petterreinholdtsen (approved):
Yes, this is needed.

PR #2128, opened 2022-11-06
qtvcp: add spacer to center buttons in cam_align

hansu:
Center buttons and labels to fit also in small frames.

Before:
image

After:
Bildschirmfoto_2022-11-06_19-38-29


hansu:
Just figured out that this a problem in gmoccapy which also occurs on the offsetpage e.g.

c-morley:
Might this be a good regardless of gmoccapy problems?

hansu:
But the left aligned version has also its advantage. If you have a very small camera window, the left buttons would be always visible, if you have all centred they might be cut.

PR #2138, opened 2022-11-10
Only build hal_speaker driver on x86, and adjust to work for uspace and RTAI

### LINKED ISSUE #2136: Commit #2e2d75c breaks arm64 builds
As the title says, commit [#2e2d75c](https://github.com/LinuxCNC/linuxcnc/commit/2e2d75c8860fa65487aee78c080304f7d77e3987) breaks arm64 builds with the following output in terminal during building:

`Compiling realtime hal/drivers/hal_speaker.c
hal/drivers/hal_speaker.c:76:10: fatal error: sys/io.h: No such file or directory
   76 | #include <sys/io.h>
      |          ^~~~~~~~~~
compilation terminated.
make: *** [Makefile:1227: objects/rthal/drivers/hal_speaker.o] Error 1
`

petterreinholdtsen:
The ioperm() code is only needed in uspace mode, and the <sys/io.h> file is most useful in user space.

Fixes #2136.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -63,6 +63,13 @@ TRIVIAL_BUILD=no
 endif
 endif
 
+# Used to enable x86 only HAL parts
+ifeq ($(shell uname -m),i686)
+X86ARCH = true
+endif
+ifeq ($(shell uname -m),x86_64)
+X86ARCH = true
+endif
 
 # Beautify output
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


andypugh:
The test for RTAPI_RTAI seems to fail with an RTAI build. 

And I think that the test around ioperm is backwards? 

It might be simpler just to not #define FASTIO so that section is skipped and rtapi_inb and outb are used. 
I suspect that the few microseconds saved are immaterial with this driver.

petterreinholdtsen:
[andypugh]
> The test for RTAPI_RTAI seems to fail with an RTAI build.

Any idea why?

> And I think that the test around ioperm is backwards?

Yes.  Fixed.

> It might be simpler just to not #define FASTIO so that section is
> skipped and rtapi_inb and outb are used.  I suspect that the few
> microseconds saved are immaterial with this driver.

I tried to only define FASTIO when RTAPI_RTAI is defined.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
OK, it took a while to figure it out, but the RTAPI_RTAI is not defined, nor is anything else. 
The answer is to add, at the beginning:

#include "config.h"     /* environment flags */

petterreinholdtsen:
[andypugh]
> OK, it took a while to figure it out, but the RTAPI_RTAI is not
> defined, nor is anything else.

Thanks.  Patch updated.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
Looks good here master / RTAI:

Runtest: 246 tests run, 246 successful, 0 failed + 0 expected, 2 skipped

phillc54:
The same result here for master on arm64

PR #2145, opened 2022-11-17
Fix typos in source

luzpaz:
Fixes various source typos


petterreinholdtsen:
Look good to me.  Had to check out the halui overide->override change, and apparently it is an old typo that has been fixed a while back.

hansu:
Looks good to me too, thank you.

PR #2146, opened 2022-11-17
docs: smallish changes+fixes while translating (smoe:docs_misc_mac_4)

smoe:
Just to keep the patch small.

### DIFF
--- a/docs/man/man9/sserial.9
+++ b/docs/man/man9/sserial.9
@@ -7,22 +7,21 @@ hostmot2 - Smart Serial LinuxCNC HAL driver for the Mesa Electronics HostMot2
 Smart-Serial remote cards
 .SH SYNOPSIS
 The Mesa Smart-Serial interface is a 2.5Mbs proprietary interface between the
-Mesa Anything-IO cards and a range of subsidiary devices termed "smart-serial 
-remotes".
+Mesa Anything-IO cards and a range of subsidiary devices termed "smart-serial remotes".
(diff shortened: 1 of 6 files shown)


docs/man/man9/sserial.9
```diff
@@ -414,7 +413,7 @@ higher maximum update rates.
 
 (float out) .7i70.0.0.analogN (modes 1 and 2 only) Analogue input values. 
 
-(float out) .7i70.0.0.fieldvoltage (mode 2 only) Field voltage monitoring pin. 
+(float out) .7______________________i70.0.0.fieldvoltage (mode 2 only) Field voltage monitoring pin.
```
SebKuzminsky, line 416: This looks wrong.

SebKuzminsky (changes requested):
Much of this is just changing line breaks.  Is this useful is some way I don't understand?

smoe:
> Much of this is just changing line breaks. Is this useful is some way I don't understand?

These line breaks are affecting the automated translations by some double blanks that are introduced (or by some other magic that I fail to induce). This is particularly unfortunate when text is indented or when some highlighting by _ or * is in the way. I am introducing what I call "semantic line breaks", which is meant to improve (or at least not disturb) the automated interpretation of the documentation.

sebastiankuzminsky:
> These line breaks are affecting the automated translations by some double blanks that are introduced (or by some other magic that I fail to induce).

Ok, I trust you on this one :-)

Thanks for taking such good care of the docs!

PR #1657, opened 2022-03-09
Update ClassicLadder to 0.8.10

havardAasen:
This is an update of ClassicLadder, to version 0.8.10, released in late 2010. Still quite old and only GTK 2. The most important change is that all drawing is done with Cairo, that's needed for the change to GTK 3, it also includes a new toolbar and the possibility to use the mouse scroll, that feature is backported from 0.9.113.

I also removed the `--config` flag, this has been marked as deprecated at least since LinuxCNC 2.6.

The upstream version includes a log feature, which I haven't enabled (all relevant code is commented out), it should be easy enough to enable it if we decide to.

I've looked at the latest release, 0.9.113, this release works with both GTK 2 and 3, but a lot has happened in the last ten years and it's not so easy to juggle the things we need and what we don't want, still, it would be nice if we managed to include part of it.

This pr is marked as a draft for three reasons:
- Update documentation
- More testing, especially with modbus (haven't been able to test this myself
- Upstream also included a new variable which I'm uncertain of, `VAR_SYSTEM` which partly collides with our `VAR_ERROR_BIT`.  The changelog entry for 0.8.5 mentions:
> New system variables %Sxxx and %SWxxx : %S0 is the classic 1Hz flash, always usefull...
> And %SW0, %SW1 & %SW2 for BCD time, date & day of week.
> %S10 is set per the modbus master if error of communication (EMC principle), in the future I've planned

PR #2156, opened 2022-11-27
Added new tests/classicladder/estop/ check.

petterreinholdtsen:
This tests the estop behavour in classicladder, pressing an external estop button, checking estop activates, and then reset estop and ensure it clears.

### DIFF
--- a/tests/classicladder/estop/checkresult
+++ b/tests/classicladder/estop/checkresult
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+result=$1
+
+if grep -qi error: $result ; then
+    echo "found string 'error' in $result"
+    exit 1
+fi
+
+exit 0
--- a/tests/classicladder/estop/classicladder-estop-ui.py
+++ b/tests/classicladder/estop/classicladder-estop-ui.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import time
+
+import linuxcnc
+import linuxcnc_util
+import hal
+
+def wait_for_hal_pin(name, value, timeout=10):
+    start_time = time.time()
+    while (time.time() - start_time) < timeout:
+        if hal.get_value(name) == value:
+            return
+        time.sleep(0.1)
+    raise RuntimeError("error: hal pin %s didn't get to %s after %.3f seconds" % (name, value, timeout))
+
+
+# unbuffer stdout
+sys.stdout = os.fdopen(sys.stdout.fileno(), 'w')
+
+retval = 0
+
+h = hal.component("python-ui")
+h.newpin("ext-estop-button", hal.HAL_BIT, hal.HAL_OUT)
+h.ready() # mark the component as 'ready'
+
+hal.connect('python-ui.ext-estop-button', 'ext-estop')
+
+c = linuxcnc.command()
+s = linuxcnc.stat()
+e = linuxcnc.error_channel()
+
+l = linuxcnc_util.LinuxCNC(command=c, status=s, error=e)
+
+c.state(linuxcnc.STATE_ESTOP_RESET)
+c.state(linuxcnc.STATE_ON)
+c.wait_complete()
+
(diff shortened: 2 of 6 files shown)


petterreinholdtsen:
The test seem to be working, but I am not confident I am actually testing classicladder behaviour or just standard LinuxCNC behaviour.

c-morley:
If you just want to test classicladder, why not got run it alone in a HAL script rather then running it with linuxcnc?
But i would think the most important things to test would be quite difficult - the GUI.
Do you know what you function would like to test?

petterreinholdtsen:
[c-morley]
> If you just want to test classicladder, why not got run it alone in a
> HAL script rather then running it with linuxcnc?

Mostly because I do not know how to do that.  Also, I suspect using the
python DISPLAY approach allow for more complex test logic.

> But i would think the most important things to test would be quite
> difficult - the GUI.

Perhaps you can try to come up with a test for this, and put it next to
the estop test in tests/classicladder/?

> Do you know what you function would like to test?

My goal is to make sure classicladder have at least some tests executed
automatically to gain some confidence that
<URL: https://github.com/LinuxCNC/linuxcnc/pull/1657 > do not break
everything.  I believe the upgrade have waited long enough.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
Trying to figure out if the test actually check the classicladder implementation, I removed the --nogui option in classicladder-estop.hal and modified the plc program to see if it would break the test.  First I changed some 'normally open' and 'normally closed' inputs to their opposite, without much luck, then then I removed some of the virtual wiring in the PLC, and this would indeed break the test.  This make me believe the test is actually exercising part of classicladder and thus testing what I intend it to test.

This make it believe it is ready to go into master, where there are currently not one test for classicladder in tests/.

c-morley:
I though a HAL only test might be easier, but if you are successful with what you have that's great.
I don't know how to write GUI tests and only really understand the premise of our existing tests.

Testing the basic bit input to output is a fine start on successful testing.
I think I was was just thinking out loud that the parts that are mostly likely to have bugs are the hardest to write a test for.
some things come to mind to test:
-float value  to s32 value (I now wish I rounded them instead)
-function checks on the fancy elements
-modbus
-reasonable timer accuracy
-above related to servo period
-make sure adjusting the number of elements doesn't overlap others.


anyways as for the new classicladder, we should probably just push it into master now that there is a 2.9 branch - no test better them having more people using it.

petterreinholdtsen:
[c-morley]
> I though a HAL only test might be easier, but if you are successful
> with what you have that's great.

I am at least somewhat successfull, but I am kind of flying blind here
with very superficial knowledge of classicladder while having fairly
good control over the test mechanics.

> I don't know how to write GUI tests and only really understand the
> premise of our existing tests.

It is possible, at least.  Feel free to check out the test-gui branch in
my github clone of linuxcnc for a sketch and example.

> Testing the basic bit input to output is a fine start on successful
> testing.

> I think I was was just thinking out loud that the parts that are
> mostly likely to have bugs are the hardest to write a test for.
> some things come to mind to test:
> -float value  to s32 value (I now wish I rounded them instead)
> -function checks on the fancy elements
> -modbus
> -reasonable timer accuracy
> -above related to servo period
> -make sure adjusting the number of elements doesn't overlap others.

Sound useful, but I must admin I do not understand what you are talking
about.  I so far just barely understand classicladder stuff, and have
not come across any of the concepts you list here. :)

> anyways as for the new classicladder, we should probably just push it
> into master now that there is a 2.9 branch - no test better them
> having more people using it.

Well, I believe automated unit tests are a lot better than throwing it
at surprised users, but I agree.  We should get it into master now and
write tests for any problems we discover, to ensure future upgrades
become easer and we can have confidence that classicladder functionally
stay inteact as we move forward.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
Lets put it in and see how it goes.

PR #2157, opened 2022-11-27
matrixkins: Add calibrated 3-axis kinematics HAL component

PetteriAimonen:
This component allows calibrating away small axis alignment errors in 3-axis cartesian machines.

The component itself should be compatible with most machines.
Dual-axis gantry machine support might be worth adding, but I don't have such a machine myself.

I've documented a generic method that could be used to measure the calibration constants, but it may need adjustment based on machine types and tools available.

### DIFF
--- a/docs/man/man9/kins.9
+++ b/docs/man/man9/kins.9
@@ -22,6 +22,8 @@ For additional information, see the Documents 'Advanced Topics':
 .PP
 .B loadrt lineardeltakins (see separate manpage)
 .PP
+.B loadrt matrixkins
+.PP
 .B loadrt maxkins
 .PP
 .B loadrt pentakins
@@ -320,6 +322,10 @@ https://w.wiki/NcY
 .TP
 .B genserkins.D\-\fIN
 Parameters describing the \fIN\fRth joint's geometry.
+
+.SS matrixkins \- Calibrated kinematics for 3-axis cartesian machines
+Similar to trivkins, but allows calibrating out small imperfections in axis
+alignment. See matrixkins(9) man page for detailed instructions.
 
 .SS maxkins \- 5-axis kinematics example
 Kinematics for Chris Radek's tabletop 5 axis mill named 'max' with tilting
--- a/src/hal/components/matrixkins.comp
+++ b/src/hal/components/matrixkins.comp
@@ -1,0 +1,320 @@
+//   This is a component for LinuxCNC HAL
+//   Copyright 2022 Petteri Aimonen <jpa at git.mail.kapsi.fi>
+//
+//   This program is free software; you can redistribute it and/or
+//   modify it under the terms of version 2 of the GNU General
+//   Public License as published by the Free Software Foundation.
+//
+//   This program is distributed in the hope that it will be useful,
+//   but WITHOUT ANY WARRANTY; without even the implied warranty of
+//   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+//   GNU General Public License for more details.
+//
+//   You should have received a copy of the GNU General Public License
+//   along with this program; if not, write to the Free Software
+//   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+component matrixkins "Calibrated kinematics for 3-axis machines";
+
+description
+"""
+The matrixkins component implements custom kinematics for 3-axis
+Cartesian machines that allows compensating minor alignment inaccuracies
+in software.
+
+\\fBKINEMATICS MODEL:\\fR
+
+By default identity matrix is used, which is equal to trivial kinematics:
+
+.nf
+| X_joint |   | 1 0 0 |   | X_axis |
+| Y_joint | = | 0 1 0 | * | Y_axis |
+| Z_joint |   | 0 0 1 |   | Z_axis |
+.fi
+
+Adjusting the calibration matrix allows compensating out many
+mechanical issues, including:
+
+  1. Scale error of each axis.
+  2. Perpendicularity between each pair of axes.
+  3. Parallelism between spindle rotational axis and Z movement.
+  4. Perpendicularity between spindle rotational axis and X/Y movement.
+
+The matrix coefficients are set by parameters C_xx .. C_zz.
+For 3 axis machine, the equations become:
+
+.nf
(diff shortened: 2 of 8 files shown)


petterreinholdtsen:
[Petteri Aimonen]
> This component allows calibrating away small axis alignment errors in
> 3-axis cartesian machines.

This look useful, but I do not know LinuxCNC enough to know for sure.

One thing I do notice, and suspect would be nice to add, is a block with
links from the manual page to related manual pages.  The one I have in
mind is the kins(9) man page, which can get a cross reference if you add
a block like this:

see_also """
\\fBkins\\fR(9).
""";

I also suspect the kins(9) man page in docs/man/man9/kins.9 should get a
cross reference to matrixkins(9).

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
A small tweak that might make it easier to work with might be to rename the parameters to

Xx, Xy Xz
Yx, Yy, Yz
Zx, Zy, Zz 

(or some similar scheme)

petterreinholdtsen:
If you need help updating the patch, ping me on IRC (#linuxcnc or #linuxcnc-devel).

petterreinholdtsen:
[Petter Reinholdtsen]
> One thing I do notice, and suspect would be nice to add,

Another thing that would be very nice to include, making sure the
feature keep working in the future, is a script in tests/ validating the
correctness of matrixkins.  For example setting some well known
coordinates and looking at its output.

-- 
Happy hacking
Petter Reinholdtsen

PetteriAimonen:
I agree on cross-refs, and adding a test is a good idea also - I have to study how the tests are supposed to work.

Regarding `c0 .. c8` vs. `Xx .. Zz`, I actually had it that way at first.
But when trying to think about the equations:

     X_joint = Xx * X_axis + Xy * Y_axis + Xz * Z_axis

there were just too many repetitions of the same symbol for me.
I don't have strong feelings either way, but in my own use I found it easier to just have `c0..c8` for the constants and reserve `x y z` symbols for the positions.

petterreinholdtsen:
[Petteri Aimonen]
> I agree on cross-refs, and adding a test is a good idea also - I have
> to study how the tests are supposed to work.

The tests are executed using scripts/runtests, and there are several
possible approaches, test.sh, test and test.hal .  The first is
executed, the second is executed with bash, and the last is executed
with halrun.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
> there were just too many repetitions of the same symbol for me.

I would leave the internal representation of the variables the same, but you can choose the HAL pin names freely. 

````
res += hal_pin_float_newf(HAL_IN, &(haldata->c[0]), comp_id,"matrixkins.Xx", i);
res += hal_pin_float_newf(HAL_IN, &(haldata->c[1]), comp_id,"matrixkins.Xy", i);
...
res += hal_pin_float_newf(HAL_IN, &(haldata->c[8]), comp_id,"matrixkins.Zz", i);
````

That way the use of the pins can be made very clear in HAL to help the user and the underlying code can remain clear too. 

I think that you could also consider not using the haldata struct and omitting the stage where you copy the array into a,b...i and assign the HAL pins to the a,b...i variables directly which might save some CPU time (but not enough to worry about) 

None of this would get in the way of the module being accepted, these are just enhancements not demands. 
Ditto the tests, tests could follow later.

PetteriAimonen:
I'm open to renaming, but I'm really not fond of the `Xx..Zz` naming scheme.
I see the logic and it was the obvious choice, but having tried it, it felt more confusing than useful.

> I think that you could also consider not using the haldata struct to omit the stage where you copy the array into a,b...i and assign the HAL pins to the a,b...i variables directly which might save some CPU time (but not enough to worry about)

Could you elaborate on this? I think there may be something about HAL components that I don't fully understand currently, I've mostly been studying from the existing ones.

My understanding is that `hal_pin_float_newf()` always allocates new storage for the underlying `float`, and I must store the pointer to it somewhere. Do you mean that instead of having a static struct called `haldata`, I would just have static variables called `a, ... i`, which would be the pointers? The [hal_pin_new](http://linuxcnc.org/docs/html/man/man3/hal_pin_s32_new.3hal.html) manual says that the memory must be allocated by `hal_malloc()` and I'm unsure whether the static variables are?

In any case, then the code would have a lot of `*e * *i - *f * *h` which is a bit hard to read.

---

Regarding the array, initially I wanted to make this more generic, to be able to set the number of joints and axes freely.
I guess for that purpose a `Xx..Zz` naming scheme would expand nicely to `Xa` etc.
But then it would require quite slow and elaborate matrix inversions, compared to the closed form solution for 3x3.
So it made more sense to keep this one simple.

andypugh:
I have mainly used .comp recently and forgot the hal_malloc step. You could hal_malloc each variable individually but that would soon become cumbersome. 
Maybe consider using HAL parameters instead of pins. It is unlikely that the correction values would need to be part of a HAL net.  (When starting this I thought that params could be linked to normal module variables but the docs say otherwise. I am no-longer clear quite what their point is. ) 

I am not set on the pin -naming scheme either, but I feel that the pin names should be clear when reading / editing the HAL file without having to reference the manpage. 

A set of #defines might be a way to keep the code clean without copying data to working variables.

PetteriAimonen:
I'm working on the tests now, and good thing I am. Looks like there is a few updates in the newest master that I had failed to take into account (I'm still running 2.8 stable on the actual machine).

PetteriAimonen:
Ok, updated now:

* Added testcase
* Changed pins to parameters and renamed as `C_xx` .. `C_zz`.
* Added documentation cross-references
* Added `KINS_NOT_SWITCHABLE` to make it actually work with newest master.
* Squashed and rebased to latest master

One remaining thing I wonder about is kinematics type.
I have it currently set as `KINEMATICS_BOTH`.
I guess because the X / Y / Z are almost exactly the same even before homing & activating compensation, `KINEMATICS_IDENTITY` could make sense also.
Any thoughts?

dngarrett:
Please look at recent updates to userkins in b818171f3ba0a90144
and incorporate use of hal_set_unready()

Also, as recommened in userkins man page:

 7) To use a different component name, rename the file
    (example mykins.comp) and change all instances of
    'userkins' to 'mykins'


So change all instances of 'userkins' to 'matrixkins'
userkins_setup --> matrixkins_setup

andypugh:
Just a note for next time, you could have kept the array for the coefficients and only changed the HAL pin names, they don't  need to match.

PR #2160, opened 2022-11-27
Docs29 mac 2

smoe:
Couple of changes while translating against branch 2.9

### DIFF
--- a/docs/man/man1/xhc-hb04.1
+++ b/docs/man/man1/xhc-hb04.1
@@ -3,21 +3,17 @@
 xhc-hb04 \- User-space HAL component for the xhc-hb04 pendant.
 
 .SH DESCRIPTION
-The xhc-hb04 component supports a common USB pendant that provides a
-number of pushbuttons, a manual pulse generator (mpg or jog wheel),
-and a selector switch for the wheel.
+The xhc-hb04 component supports a common USB pendant that provides a number of pushbuttons,
+a manual pulse generator (mpg or jog wheel), and a selector switch for the wheel.
 .PP
-There are at least two hardware versions -- one with 16 buttons and
-a more common one with 18 buttons.  The information herein is based
-on the 18 button device with a USB Vendor:Product code of 10CE:EB70.
+There are at least two hardware versions -- one with 16 buttons and a more common one with 18 buttons.
+The information herein is based on the 18 button device with a USB Vendor:Product code of 10CE:EB70.
 .PP
-In addition to buttons, the pendant provides an LCD display for
-the current stepsize multiplier (from a set of available integer
-values), position (absolute and relative, labeled MC and WC
-respectively), feedrate (override percent and value in units per
-minute), and spindle speed (override percent and value in
-revolutions per minute (RPM)).  The display is managed by a rotary
-switch that selects one of four axes for wheel positioning,
+In addition to buttons, the pendant provides an LCD display for the current stepsize multiplier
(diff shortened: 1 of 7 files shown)


docs/man/man1/xhc-hb04.1
```diff
@@ -108,91 +100,83 @@ Example:
...
-inactive (sleeping) message.
+(bit out) \fIxhc\-hb04.sleeping\fR True when the driver receives a pendant inactive (sleeping) message.
 .TP
 (bit out) \fIxhc\-hb04.jog.enable\-off\fR True when the pendant rotary
 selector switch is in the OFF position or when the pendant is sleeping.
 .TP
-(bit out) \fIxhc\-hb04.enable\-[xyza]\fR True when the pendant rotary
-selector switch is in the [xyza] position and not sleeping.
+(bit out) \fIxhc\-hb04.enable\-[xyza]\fR
+True when the pendant rotary selector switch is in the [xyza] position and not sleeping.
```
hansu, line 151: Why are these line break changes necessary?

smoe: I got a sentence starting with "selector switch is in the [xyza] position and not sleeping." to translate in weblate. All those line breaks are about keeping those parts of a paragraph on a single that that make no sense if read without context. Also the highlighting is often broken into multiple llines, which is not compatible with the auto-translators.

docs/man/man1/xhc-hb04.1
```diff
@@ -108,91 +100,83 @@ Example:
...
+A 1 pulse on this pin changes the stepsize to the next lower stepsize in the stepsize sequence specified in the xhc\-hb04 (loadusr) command.
 
 .SH Input Pins (to the pendant LCD display)
 .TP
-(float in) \fIxhc\-hb04.[xyza].pos\-absolute\fR Absolute position display.
-(typically connect to: halui.axis.N.pos\-feedback). The LCD display
-for pos\-absolute is fixed format with a sign, 4 number digits and 3
-fraction digits (+XXXX.XXX), require: \-9999.999 <= value <= 9999.999.
+(float in) \fIxhc\-hb04.[xyza].pos\-absolute\fR
+Absolute position display.  (typically connect to: halui.axis.N.pos\-feedback).
```
hansu, line 112:
```suggestion
(float in) \fIxhc\-hb04.[xyza].pos\-absolute\fR Absolute position display. 
(typically connect to: halui.axis.N.pos\-feedback).
```

hansu (commented):
The line breaks are not really consistent. 
For some, the description starts the line after the pin name, for some in the same line like line 132, 135, 145.
Please go through these line breaks again. Thanks!

docs/man/man1/xhc-hb04.1
hansu: ok

docs/man/man1/xhc-hb04.1
smoe: I just checked, the line breaks indeed have an effect on how the text is rendered. IMHO the pin names are too difficult to read when the description starts right afterward, so I decided for the line break after the pin. The current version is in 2.9 is inconsistent on this already.

smoe:
Moved all pin descriptions to second line. I suggest to revisit this situation at some point when we address all those man pages.

smoe:
@hansu resolved conflicts

hansu:
Thanks

PR #2164, opened 2022-11-28
Added rtapi_ioperm() declaration for hal_speaker

petterreinholdtsen:
This fixes the following compiler warning:

hal/drivers/hal_speaker.c: In function ‘rtapi_app_main’: hal/drivers/hal_speaker.c:164:9: warning: implicit declaration of function ‘ioperm’ [-Wimplicit-function-declaration]
  164 |     if (ioperm(SPEAKER_PORT, 1, 1) < 0) {
      |         ^~~~~~


SebKuzminsky:
The Realtime API (RTAPI) is meant to be an abstraction layer that makes it so all our realtime code can run on all the available realtime kernels.

Functionality like starting threads, allocating shared memory segments, or accessing I/O ports is done differently on different realtime kernels and platforms, but RTAPI provides an API that hides all those differences.  So yes, i think this new function belongs in RTAPI.

Please also add it to the rtapi_io manpage.

petterreinholdtsen:
[Sebastian Kuzminsky]
> Please also add it to the rtapi_io manpage.

OK, man page updated.

--
Happy hacking
Petter Reinholdtsen

PR #2168, opened 2022-11-29
Curtdutt/raster

curtdutt:
Adds raster component along with accompanying RasterProgrammer python component.

Teaches halcompiler to understand port pins. Adds sim laser config to show how vector control and raster control can be integrated.


petterreinholdtsen:
[Curtis Dutton]

> Or will that happen when this pull request gets merged?

It will not happen when the pull request is merged.  I tend to squash
commits into one in my pull requests, unless the request consist of
distingt logically separate changes that should be remembered in the git
history (like some importing code and modifying it, in case an updated
import need to be done in the future).

-- 
Happy hacking
Petter Reinholdtsen

curtdutt:
Alright. For this purpose they can be squashed into one commit. I'll submit
a revised pull request.

PR #2171, opened 2022-11-30
Adds laser raster and vector control components.

curtdutt:
laserpower.comp - Scales laser output based upon velocity raster.comp - Controls laser power based upon preprogrammed bitmap data.

Includes associated sim configuraion to demonstrate laser component integration.

Includes basic test for raster component.

Teaches halcompiler about port pins.
     Adjusts pin component structure member names to prevent conflicts with preprocessor macros.

### DIFF
--- a/configs/sim/axis/laser/laser.hal
+++ b/configs/sim/axis/laser/laser.hal
@@ -1,0 +1,79 @@
+loadrt conv_s32_u32 names=laser.motion.type-conv
+loadrt bitslice names=laser.motion.type-slice personality=2
+loadrt limit1 names=laser.power.min.limit,laser.power.max.limit,laser.power.cmd.limit
+loadrt laserpower names=laser.control
+loadrt raster names=laser.raster
+loadrt not names=motion.enabled-not
+
+addf laser.power.min.limit                      servo-thread
+addf laser.power.max.limit                      servo-thread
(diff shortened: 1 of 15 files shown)


andypugh:
Thanks for this

petterreinholdtsen:
The test code call 'realtime start', which uses a binary not present in
Debian.  Also, the test leave behind a background process when it fail,
causing the test suite to fail.  I'm working on a patch to solve the
latter.

What is the 'realtime start' call supposed to do?  It do not seem to be
needed when I test it.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
os.system("realtime start") shouldn't be necessary anyway, the HAL file has the required "start" command. 

However LinuxCNC does have a "realtime" script, so it should work: https://github.com/LinuxCNC/linuxcnc/blob/master/scripts/realtime.in

petterreinholdtsen:
[andypugh]
> os.system("realtime start") shouldn't be necessary anyway, the HAL
> file has the required "start" command.

Then it should be removed.

> However LinuxCNC does have a "realtime" script, so it should work:
> https://github.com/LinuxCNC/linuxcnc/blob/master/scripts/realtime.in

It is not found when runtest is executed on github, which is the reason
I noticed it.

-- 
Happy hacking
Petter Reinholdtsen

PR #2172, opened 2022-12-01
Fix obsolete link to documentation in image-to-gcode(1).

petterreinholdtsen:


SebKuzminsky:
This should go in 2.8.

petterreinholdtsen:
[Sebastian Kuzminsky]
> This should go in 2.8.

OK, moved to 2.8.

I just prefer to focus on future releases instead of old ones. :)

-- 
Happy hacking
Petter Reinholdtsen

SebKuzminsky:
> I just prefer to focus on future releases instead of old ones. :)

That's understandable.  But 2.8 isn't exactly "old", it's the current stable release, and it still deserves our support and bug fixes.

PR #2177, opened 2022-12-02
Add  HOME_IS_SHARED = 0 to sim axis.ini for all joint

alkabal:
Missing 
HOME_IS_SHARED = 1  for joint 1.


alkabal:
If this change is unwanted feel free to close it.

I have do some change mostly because i have problem with my settings trying to create a sim config.

This patch does not correct it bt it seem more logic to me.

The real problem i have i using : 

[JOINT_0]
HOME_OFFSET = -1

HOME_SEARCH_VEL = -12
HOME_LATCH_VEL = -2

The hal file using comp never find the home.

alkabal:
Thinking a little more about the proposed change, i think too many sim file need to be edited for remove the shared home switch.

andypugh:
It is annoying when the sims get stuck on a home switch. Perhaps the fix would be to simply change simulated-home.hal so that it doesn't matter if the home switches are shared.

PR #2178, opened 2022-12-03
Reintroduce translation links.

petterreinholdtsen:
This make translations easier to find on linuxcnc.org.

### DIFF
--- a/docs/src/index.tmpl
+++ b/docs/src/index.tmpl
@@ -88,6 +88,28 @@ function setup_page(){
   <img src="linuxcnc-logo-chips.png" alt="LinuxCNC Logo" width="175"/>
 </div>
 
+@TRANSLATIONS@
+<p>Translations:
+<a href="cs/">Czech</a>
+<a href="da/">Dansk</a>
+<a href="de/">Deutch</a>
+<a href="es/">Espa&ntilde;ol</a>
+<a href="fr/">Francés</a>
+<a href="hu/">Hungarian</a>
+<a href="it/">Italian</a>
+<a href="ka/">Georgian</a>
+<a href="nb/">Norsk bokm&aring;l</a>
+<a href="pt/">Portuguese</a>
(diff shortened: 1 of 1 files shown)


hansu:
Why not for the 2.9 docs page as well?

petterreinholdtsen:
[Hans Unzner]
> Why not for the 2.9 docs page as well?

Because I want to get it working in master first.  At the moment the
links only work from the english edition to the translated editions, but
not from the translations.  Not sure how to fix it properly.

Besides, we translate for master, not 2.9, so proof reading 2.9
translations is not really useful.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[Sebastian Kuzminsky]
> Hmm, really?  All the translation infrastructure is in 2.9, why
> wouldn't that be the main target of translation updates?

Really.  We translate for branch master, check weblate setup.  If we try
to do both, we will get a lot of merge conflicts and if we only do 2.9,
we can not improve the translation for 2.10 in any sensible way at the
same time, as long as the merge forward approach is used.

-- 
Happy hacking
Petter Reinholdtsen

SebKuzminsky:
I had assumed we'd keep translating 2.9 and merge up to master/2.10 as usual.  Any new translatable strings in master/2.10 would be translated there.

Merges from 2.9 to master/2.10 might have conflicts  but those would be resolved as with any merge.  Does po4a not support this workflow, or make it unusually cumbersome somehow?

I'm happy to defer to you on the topic of po4a best practices, i'm only trying to learn why the normal way of doing things isn't appropriate here.

petterreinholdtsen:
[Sebastian Kuzminsky]
> So we apparently need some kind of smart way to deal with po files,
> even if we don't make any translations in 2.9...  Do you have a
> suggestion here?

Yes.  Do not touch the PO files in 2.9.

-- 
Happy hacking
Petter Reinholdtsen

PR #2184, opened 2022-12-05
Removal of _es files from source tree.

smoe:
In some deeper theory these files should be redundant with what is provided via weblate.

There are quite a few *_es* files in the archive in the configs/ hierarchy. Are those also covered by our weblate initiative?  Don't think so. Should they? It would rather be another weblate subproject, right? Our attempt to save the previous translations for a quick transition has yet not worked out as hoped. I would hence be reluctant to change too much.


petterreinholdtsen:
[Steffen Möller]
> In some deeper theory these files should be redundant with what is
> provided via weblate.

The only reason these were not part of the po4a/weblate setup is that I
were not aware of documentation outside docs/ when I wrote it.  To add
it to the existing setup, they english files need to be added to
docs/po4a.cfg.  To make a different POT file for them, to make it
possible to have a separate weblate subproject, a guess a new po4a.cfg
only for these files need to be created and the build system adjusted to
handle it.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
I guess the lack of feedback indicate that no0ne really care about the
_es files and it is OK to just remove them?

In any case, if we want to move them into the translation framework, the
files need to be listed in docs/po4a.cfg, for example like this:

  [type: Text_def] ../configs/sim/axis/README $lang:../configs/sim/axis/README_$lang

I suspect it is better to set up a different POT file for the files in
configs, and translate them separately.  Perhaps using configs/po4a.cfg
and configs/po/ to handle these translations?

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[Steffen Möller]
> Would that be a separate section in weblate?

Yes.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
We had an enthusiastic Spanish translator who translated files as he found them (in some cases replacing the English version, which was annoying). 
The files I was expecting to see in the list here were things like rtfaults.txt
But the files in the commit appear to be in the docs hierarchy rather than in the source tree as the commit title suggests?

smoe:
All in docs/, "source tree" was redundantly referring to what is in github to distinguish it from what we generate with po4a. That said, I happily extend that effort into what is in src/ - or maybe better keep that for another PR.

smoe:
I found https://github.com/LinuxCNC/linuxcnc/commit/ab788522d4a75958cea46b43d928131c59cbb8e3 reintroducing rtfaults.txt but fail to find it in the archive.  The file docs/README.md is a bit too long, such that I would make it a part of the regular documentation.

For the other small READMEs in configs, well, ... if the files are helpful then yes, let's have them in weblate and swamp the configs directory after the installation with them.
.

SebKuzminsky:
I'm not sure how translations of these files would be helpful.  The translations would be generated at build time, so the translated top-level README would not be available on the github project page, and the docs/INSTALL and docs/README would only be visible to users who installed linuxcnc-doc-es.  That probably makes the translated docs/INSTALL file not useful, but maybe the docs/README would have some tiny bit of utility?

In my opinion translating these files is not worth the hassle, and I approve of this PR to remove them.

PR #2186, opened 2022-12-06
Added longer timeout and better error reporting in tests/rs274ngc-startup/

petterreinholdtsen:


petterreinholdtsen:
I suspect the timeout change is not really needed, as the change was to try to avoid a problem cauesd by new 'ps' behaviour (#2196).  Closing it because of this.

PR #2187, opened 2022-12-06
docs29: reformatting xhc man page pins to monospaced font

smoe:
Redid an earlier patch to get the font right that denotes signals/pins.


hansu:
Haven't looked how the man page looks like, but the HTML version still doesn't render correctly: http://linuxcnc.org/docs/2.9/html/man/man1/xhc-whb04b-6.1.html#Input/Output%20Signals

PR #2189, opened 2022-12-07
Rewrote raster test to clean up on failures.

petterreinholdtsen:
It should always call 'halrun -U' to stop HAL, even when something go wrong.

Improving pull request #2171.

### DIFF
--- a/tests/raster/test
+++ b/tests/raster/test
@@ -5,51 +5,21 @@ from struct import *
 import os
 from raster import *
 
-os.system('realtime start')
-
-c = component("test")
-
-
 theta = 0.005
 timeout = 0.25
-
-reset = c.pinNew("reset", halType.BIT, pinDir.OUT)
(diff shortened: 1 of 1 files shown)


andypugh:
Any idea why Sid builds are failing (not just on this  one)

petterreinholdtsen:
[andypugh]
> Any idea why Sid builds are failing (not just on this  one)

No.  All I can say so far is that it is not eatmydata related.  Seem to
be something leaving a hal process behind, but no idea what, why or why
it only happen on sid.

-- 
Happy hacking
Petter Reinholdtsen

SebKuzminsky:
I'm also reproducing the test failures here on Debian Unstable (without eatmydata).  It's repeatably the same few specific tests that fail.

Note that tests should not have to run `halrun -U` themselves to clean up at the end.

PR #2144, opened 2022-11-17
Docs misc mac 3

smoe:

PR #1991, opened 2022-09-02
docs/src/plasma: translation-triggered consistency checks (smoe:docs_plasma)

smoe:
The typical initial run to avoid later syntactical changes in weblate.

PR #2190, opened 2022-12-08
changes against 2.9 motivated by translations (Docs misc 10)

smoe:
While going through the weblate entries requesting changes.


SebKuzminsky:
I'm a bit confused about how to handle translations now that we have a 2.9 and master branch that both use gettext...

@petterreinholdtsen pointed out that .po files are difficult to merge using git, which as I understand it means we should stop doing translations in 2.9.  Specifically, we should not commit to any .po files in 2.9.

Any time we change a .adoc file (one of the English ones in the git repo, I mean), the matching translations in the .po file become outdated and the translated versions of that .adoc become more and more English.

I think that means we should avoid translation-driven changes to the .adoc files, like what are in this PR.  Changes like this will result in *less* translation coverage in 2.9.  I think translation-driven changes should go in master, where they can *improve* translations instead of hurting them.

Do I have that right?

petterreinholdtsen:
[Sebastian Kuzminsky]
> Do I have that right?

Yes, I believe so.  But I believe improvement to the documentaiton would
be worth including in release branches, even at the cost of reducing the
translation coverage, if we can make sure new releases are published
regularly, for example every third month.  This would ensure
translations at most use three months to reach our users.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
For general improvements of the documentation it's without question that they should go into 2.9 as well IMHO.
If we make translation optimizing changes (like line breaks etc.) in 2.9:
- it would would decrease the translation coverage
- but it will help avoiding merge conflicts, when merging up from 2.9 to master
So we have to pick one goal I guess.

smoe:
That "misc 10" patch was waiting on my disc for a while, did not want to have too many open PRs which then conflict with each other, so I delayed a few. It was really nice to see that the version in 2.9 was almost always superior to what that older patch suggested.

Those line breaks are essential for the translators - especially so when the paragraph is indented. As long as the documentation is not featuring something that is only available in 2.10, I think we should just make the documentation 2.9 better and merge with master, i.e. have master and 2.9 mostly identical. The weblate-provided .po files from master I expect can then make some good use as .po files for 2.9. My major concern here is for all the other languages, for which the translations are not yet existing but expected to arrive, eventually.

c-morley:
But 2.9 and 2.10 will not stay the same forever.



Sent from my Galaxy



From: Steffen Möller ***@***.***>
Date: 2022-12-08 11:27 a.m. (GMT-08:00)
Cc: Subscribed ***@***.***>
Subject: Re: [LinuxCNC/linuxcnc] changes against 2.9 motivated by translations (Docs misc 10) (PR #2190)


That "misc 10" patch was waiting on my disc for a while, did not want to have too many open PRs which then conflict with each other, so I delayed a few. It was really nice to see that the version in 2.9 was almost always superior to what that older patch suggested.

Those line breaks are essential for the translators - especially so when the paragraph is indented. As long as the documentation is not featuring something that is only available in 2.10, I think we should just make the documentation 2.9 better and merge with master, i.e. have master and 2.9 mostly identical. The weblate-provided .po files from master I expect can then make some good use as .po files for 2.9. My major concern here is for all the other languages, for which the translations are not yet existing but expected to arrive, eventually.

smoe:
I do not think we have an answer yet about how we want to update translations for updates of the 2.9 release. This is also why I think we should have everything as similar as possibly between 2.9 and master until we release.  And then ... donno.

hansu:
Does the buildbot actually build packages for the translated docs for 2.9? Correct me if I am wrong, but I don't think so.
And for the website the 2.9 translated docs are not shown either.
So based on that, I wouldn't focus too much on the translation coverage in 2.9.

SebKuzminsky:
> Does the buildbot actually build packages for the translated docs for 2.9?

Yes, for zh-cn, es, de, and fr (but currently only for Buster):  http://buildbot.linuxcnc.org/dists/buster/2.9-rtpreempt/binary-amd64/?C=M;O=D

But I agree with you, translations should go in master (aka 2.10), not 2.9, because of the difficulty with merging po files.

c-morley:
> I agree we should focus on master for translations, as long as we make
new releases from master every few months.
>
Linuxcnc releases every few months is not remotely realistic.  Once a year would be a good and difficult goal.

petterreinholdtsen:
[c-morley]
> Linuxcnc releases every few months is not remotely realistic.  Once a
> year would be a good and difficult goal.

I believe we should make it realistic, and that it is realistic to get
there with some work. :)

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
Why?



Sent from my Galaxy



From: petterreinholdtsen ***@***.***>
Date: 2022-12-11 11:02 a.m. (GMT-08:00)
Cc: c-morley ***@***.***>, Comment ***@***.***>
Subject: Re: [LinuxCNC/linuxcnc] changes against 2.9 motivated by translations (Docs misc 10) (PR #2190)


[c-morley]
> Linuxcnc releases every few months is not remotely realistic. Once a
> year would be a good and difficult goal.

I believe we should make it realistic, and that it is realistic to get
there with some work. :)

--
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[c-morley]
> Why?

Not quite sure which aspect the 'why' refer to, but will try to adress
both.  I believe we should do releases every few months because it
reduce the time delay from a new feature is implemented until it is
available in a new release, and I believe it is realistic to get to a
point where we can release very few months by doing more test driven
development and by automating the release process to ensure we know if
the code is relasable at any time.

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
Quick small releases don't allow real world testing - or more realistically, the users are the testers.
Writing a test for everything is a waste of developers time, which we have precious little. ( not to say tests in general are a waste of time -  some docs on how to write good tests would be helpful)
It means large projects must stay out of master till they are 'finished' - could debate if this is good or bad for sure.
We spend more effort on releases rather then development. 
You say we can improve the release process by automation and I'm not sure what you mean by that - we already use the buildbot to build packages often. Please expand on what you mean. To my thinking really releasing means we have stopped development and put in extra effort to clean up and stabilize the packages for production/end user use.  Can't really automate that. 

Now releasing every 12 -18 months seems to be more inline with the way the project works - balancing pure development work with bug fixes and releasing. 

Certainly time between releasing bug fixes in released branches could be easily improved.

Now saying all that, I recognize that our very long time between releases encourages people to use master as production code. I think that again is really a reflection of the lack of the specialized qualified people and a lack of a crystalized goal. 
So I don't think releasing every 3 months is doable nor a good idea to strive for. 
Certainly basing other policy on that (at least currently unrealistic ) idea also seems worth talking about.
But I do agree that improving the time between releases and maybe setting a firm target goal of when to release would be an excellent idea.

petterreinholdtsen:
[c-morley]
> Quick small releases don't allow real world testing - or more
> realistically, the users are the testers.

This is where test driven development come in, to stop depending on
users as the testers.

> Writing a test for everything is a waste of developers time, which we
> have precious little. ( not to say tests in general are a waste of
> time -  some docs on how to write good tests would be helpful)

Are there no docs on that at the moment?  If not, I guess I should try
to write some.

I do not believe writing tests for everything is a waste of time, but we
are far from that at the moment, so I believe writing more tests is very
valuable use of developers time.  Once we get closer to full testing,
adding more tests will gain little advantage, and perhaps we come to a
point where writing more tests will be a waste of time.

Just writing a simple test for classicladder, like I did in
tests/classicladder/, has already demonstrated that classicladder
crashes on arm, which is very valuable to detect without waiting for a
user to run into it.  I am sure there are more low hanging fruits like
this that can increase our confidence in the linuxcnc correctness.

> It means large projects must stay out of master till they are
> 'finished' - could debate if this is good or bad for sure.

At least they would have to ensure they include tests to verify their
operation before going into master.

> We spend more effort on releases rather then development.

Sad to hear.  I believe a good way to avoid it is to have more tests,
and automate the task involved to get to a point where we always know if
master is fit for release and can make a new release with the press of a
button.

> Please expand on what you mean. To my thinking really releasing means
> we have stopped development and put in extra effort to clean up and
> stabilize the packages for production/end user use.  Can't really
> automate that.

I mean that we should regularly build and automatically test the ISOs to
ensure they are working, we should do the same with the various guis and
other parts that should be working with LinuxCNC.  If we have enough
tests in place, we will know if the code is ready for release or not,
and if it is not we have to fix it, and if it is we can release quickly
without a long stabilization period.

It might take a while to get there, but with a culture of writing tests
and fixing their failures, we can pull it off.

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
If the predication of entry of code into linuxcnc is a full suit of tests, I predict you will severely limit the number of and size of code that actually gets in. There is lots of code produced by non professional programmers. So either we don't include their code or we need a dev to make the tests.

There is a difference between encouraging more test and making policy that requires more tests.

automating ISOs seems a great idea. Acomplaint I have is we release an ISO  and it often has bad bugs in it but we seldom get a new ISO. So new users always have to deal with the bugs untill they find out to update.
But functional ISOs are a small part of releasing.

Maybe we fundamentally see the linuxcnc project developing differently?

petterreinholdtsen:
[c-morley]
> If the predication of entry of code into linuxcnc is a full suit of
> tests, I predict you will severely limit the number of and size of
> code that actually gets in. There is lots of code produced by non
> professional programmers. So either we don't include their code or we
> need a dev to make the tests.

A first start might be to document how to write tests, I've started on a
document like that in
<ULR: https://github.com/LinuxCNC/linuxcnc/pull/2203/ >.

> There is a difference between encouraging more test and making policy
> that requires more tests.

Absolutely.  But the first step is to establish a culture were tests are
seen as a valuable and vital part of the process.  I believe we are on
the right track there, and that we will succeed if we all help build
this culture.

> Maybe we fundamentally see the linuxcnc project developing
> differently?

Could be.

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
Tests are good. Relying on tests to catch everything means you need to know everything that can go wrong. Some things like components and gcode lend themselves to test. Somethings like configuration programs, are so complicated that only real word testing is going to show the non obvious.
So time to 'shake down' linuxcnc is required on every release. I just think you are under estimating this or overestimating dev energy. maybe I'm wrong, maybe ten more energeric devs will join.

I'll just point out that the start of this conversation, was that you suggested a policy of focusing on master only for translation only if we release every few months.

So that is current policy based on, at least at the moment, unrealistic expectations.
These expectations IMHO are not going to change very quickly if at all. 

The merging problems was exactly the problem I discussed on the mail list and I just can't see anyway around that, without incurring a high barrier to being a dev, but we will cross that bridge when we get to it.

I'm really not trying to be a negative nilly - just a second look at consequences on policy.
Thank you for explaining your position and thank you for your contributions to linuxcnc!

hansu:
>But I agree with you, translations should go in master (aka 2.10), not 2.9, because of the difficulty with merging po files.

The target of the translations that are done with Weblate will not be changed. I was talking about the changes in the docs sources.

With "So based on that, I wouldn't focus too much on the translation coverage in 2.9." I was meaning that I would prefer merging Steffens "translation motivated changes" into 2.9 instead of master. 
The po files stay the same while the docs change --> reduction of translation coverage.

hansu:
Hmm no really an agreement whether to merge Steffens changes to 2.9 or not.  But I would like to merge them in the next time. So if nobody contradicts, I will merge them as they are now targeted (#1991, #2190, #2144, #2207 and #2208)

smoe:
The translation-effort has two sides. Important for 2.9 is the improvement of the English. The other side are the translations, and French and Spanish are practically non-existant because of the changes already made. Until someone iterates through all the suggestions in weblate, we basically do not have them.

From my point of view, the only chance to get those translations to Fr and Es back is to keep the 2.9 texts mostly identical to what happens in master. Hence, editing in 2.9 instead of master is what saves the translations, not the problem. We just need a way to copy those files back to 2.9.

smoe:
Eliminated conflict, rebase.

smoe:
> So if nobody contradicts, I will merge them

@hansu - yes, please!

docs/src/gui/gmoccapy.adoc
```diff
@@ -539,40 +539,40 @@ For a three axes XYZ machine the HAL pins will react as shown in the following t
...
+m|gmoccapy.h-button.button-0 | delete MDI history    |             | edit offsets
+m|gmoccapy.h-button.button-1 |                       | home all    | touch X
+m|gmoccapy.h-button.button-2 |                       |             | touch Y
+m|gmoccapy.h-button.button-3 |                       | home x      | touch Z
+m|gmoccapy.h-button.button-4 | open classic ladder   | home y      |
+m|gmoccapy.h-button.button-5 | open HAL scope        | home z      |
+m|gmoccapy.h-button.button-6 | open HAL status       |             | zero G92
+m|gmoccapy.h-button.button-7 | open HAL meter        |             |
+m|gmoccapy.h-button.button-8 | open HAL calibration  | unhome all  | set selected
+m|gmoccapy.h-button.button-9 | open HAL show         | back        | back
```
hansu, line 551: Good idea. Can you add this for _Table 1_ as well?

docs/src/gui/gmoccapy.adoc
```diff
@@ -1260,11 +1221,11 @@ image::images/gmoccapy_settings_hardware.png["Hardware settings",align="left"]
...
-The main reason for this was my own test to solve this through HAL connections, resulting in a very
-complex HAL file. Imagine a user having an MPG Wheel with 100 ipr and he wants to slow down the max.
-vel. from 14000 to 2000 mm/min, that needs 12000 impulses, resulting in 120 turns of the wheel!
-Or an other user having a MPG Wheel with 500 ipr and he wants to set the spindle override witch has
-limits from 50 to 120 % so he goes from min to max within 70 impulses, meaning not even 1/4 turn.
+The main reason for this was my own test to solve this through HAL connections, resulting in a very complex HAL file.
+Imagine a user having an MPG Wheel with 100 ipr and he wants to slow down the max.  vel. from 14000 to 2000 mm/min,
+that needs 12000 impulses, resulting in 120 turns of the wheel!
+Or an other user having a MPG Wheel with 500 ipr and he wants to set the spindle override
+witch has limits from 50 to 120 % so he goes from min to max within 70 impulses, meaning not even 1/4 turn.
```
hansu, line 1228:
```suggestion
which has limits from 50 to 120 % so he goes from min to max within 70 impulses, meaning not even 1/4 turn.
```

smoe:
@hansu - done as requested (and a bit more, likely redundant with the other PRs, but they are conflicting already, anyway).

PR #2196, opened 2022-12-08
fix zombie detection problem with procps 4.0 (sid)

SebKuzminsky:
This makes our launcher scripts work with both procps 4 (sid) and procps 3 (bookworm and older).


petterreinholdtsen:
I've reported this issue to Debian as https://bugs.debian.org/1025790 .

PR #2198, opened 2022-12-10
Drop pastie.org and add 0bin.net to list of suggested services

petterreinholdtsen:
No need to use a http only service like pastie.org after letsencrypt came around.  The 0bin.net service is end-to-end encrypted and a good alternative.


andypugh:
I would suggest adding paste.debian.net too.

petterreinholdtsen:
[Andy Pugh]
> I would suggest adding paste.debian.net too.

Done.

-- 
Happy hacking
Petter Reinholdtsen

PR #2199, opened 2022-12-10
Trim list of suggested Debian distributions to use

petterreinholdtsen:
Drop Jessie and Wheezy from alternative distribution list and document when Wheezy, Buster and Ubunty Precise was end of life.

### DIFF
--- a/docs/src/getting-started/getting-linuxcnc.adoc
+++ b/docs/src/getting-started/getting-linuxcnc.adoc
@@ -266,17 +266,25 @@ run on any 64-bit x86 system)
 
 [options="header"]
 |===
-| Distribution   | Architecture  | Kernel     | Package name    | Typical use
-| Debian Buster  | amd64 & i386  | Stock      | linuxcnc-uspace | simulation only
-| Debian Buster  | amd64 & armhf | preemp-rt  | linuxcnc-uspace | machine control & simulation
-| Debian Buster  | amd64         | RTAI       | linuxcnc        | machine control (known issues)
-| Debian Jessie  | amd64 & i386  | Stock      | linuxcnc-uspace | simulation only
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
I would prefer to switch the page to only mention Bullseye and Bookworm, but our package set do not allow it. No idea if anyone is porting to newer Ubuntu.

SebKuzminsky:
I think we should leave Wheezy in the list for now, since that's the newest distro we're building RTAI debs for.

petterreinholdtsen:
[Sebastian Kuzminsky]
> I think we should leave Wheezy in the list for now, since that's the
> newest distro we're building RTAI debs for.

Any use of Wheezy is a gaping security problem.

-- 
Happy hacking
Petter Reinholdtsen

SebKuzminsky:
*Any* use?  I think many (maybe even most) LinuxCNC machine tools are behind firewalls and are not used for web browsing etc.  Some are probably fully airgapped, not connected to a network at all.  It's maybe not *as* secure as running a recent kernel on a recent distro, but it may be acceptable to some users.

There is some hardware where no alternative kernel works as well as RTAI.

These two observations together tell me we should not stop recommending Wheezy yet.

Maybe we can get Buster RTAI debs going on the new buildbot; *that* would be a good time to stop promoting Wheezy.

andypugh:
On Sat, 10 Dec 2022 at 17:27, Sebastian Kuzminsky ***@***.***>
wrote:

> I think we should leave Wheezy in the list for now, since that's the
> newest distro we're building RTAI debs for.
>
Not entirely true, there are hand-crafted  Buster RTAI debs (and
matching kernel) on wlo.

-- 
atp

PR #2209, opened 2022-12-16
Change HAL pin type for hy_gt_vfd.fault-info from float to u32

### LINKED ISSUE #2206: HY_GT_VFD fault info register
The HAL component HY_GT_VFD declares a HAL pin called hy_gt_vfd.fault-info as a FLOAT output. However, that register (0x5000) is a set of BIT values and should therefore be declared as a U32.  Then it will be possible for python to convert it to a hex value for display in a GUI.

petterreinholdtsen:
Restructure code to handle two types of HAL pin types for modbus register.

Fixes #2206.

### DIFF
--- a/docs/src/man/man1/hy_gt_vfd.1.adoc
+++ b/docs/src/man/man1/hy_gt_vfd.1.adoc
@@ -259,8 +259,15 @@ PC.03 = 1::
 
 *hy_gt_vfd.fault-info* (float, out)::
 
-        The VFD's fault info register.
+        The VFD's fault info register in floating point
+        representation.  This is kept for backwards compatibility with
+        existing setups and will be removed in the future.
 
+*hy_gt_vfd.fault-info-code* (u32, out)::
+
+        The VFD's fault code register value.  Introduced in LinuxCNC
+        version 2.10.  0x00 if no fault is detected, see GT Series
+        Inverter Manual page 87 for list of fault codes.
 
 == ISSUES
 
--- a/src/hal/user_comps/hy_gt_vfd.c
+++ b/src/hal/user_comps/hy_gt_vfd.c
@@ -64,9 +64,18 @@ typedef struct {
     // Multiply the uint16 value read by this multiplier to get the human-
     // readable value.
     float multiplier;
-
-    hal_float_t *hal_pin;
+    int type;
+    union {
+        hal_float_t *hal_pin_float;
+        hal_u32_t *hal_pin_u32;
+    };
 } modbus_register_t;
+
+typedef enum {
+    REGISTER_UNKNOWN = 0,
+    REGISTER_FLOAT = 1,
+    REGISTER_U32 = 2,
+} register_type_t;
 
 int num_modbus_registers = 0;
 modbus_register_t *modbus_register;
@@ -246,7 +255,14 @@ int read_modbus_register(modbus_t *mb, modbus_register_t *reg) {
         hy_modbus_sleep();
         r = modbus_read_registers(mb, reg->address, 1, &data);
         if (r == 1) {
-            *reg->hal_pin = data * reg->multiplier;
+            switch (reg->type) {
+            case REGISTER_FLOAT:
+                *reg->hal_pin_float = data * reg->multiplier;
+                ;;
+            case REGISTER_U32:
+                *reg->hal_pin_u32 = data;
+                ;;
+            }
             return 0;
         }
         fprintf(stderr, "%s: error reading %s (register 0x%04x): %s\n", __func__, reg->name, reg->address, modbus_strerror(errno));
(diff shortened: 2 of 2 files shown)


petterreinholdtsen:
I would like someone who know this component a little better to see if this is a wanted change and if the approach taken is a good one.

andypugh:
Looks reasonable, though possibly the input_terminal should be int too. (but then I suspect that output_terminal isn't really handled at all in any useful way)

SebKuzminsky:
Looks good to me.  I no longer have access to this VFD to test it, I'm interested to see if it works for @persei802.

I also don't have access to the manual, but I wonder if the fault-info register is a bitmap of different errors; if so it might make sense to break it out into a bunch of bit pins identifying the different errors.

persei802:
That is correct. There are 26 bits indicating a fault condition. In my GUI, I convert this register to a hex value so it's easy to determine which bits are set. The manual doesn't actually explain what each of these bits mean, except that 0x00 is no fault. In my personal opinion, it is not useful to break them out into individual HAL pins.
Jim

docs/src/man/man1/hy_gt_vfd.1.adoc
```diff
@@ -257,9 +257,11 @@ PC.03 = 1::
 
         The VFD's external counter register.
 
-*hy_gt_vfd.fault-info* (float, out)::
+*hy_gt_vfd.fault-info* (u32, out)::
 
-        The VFD's fault info register.
+        The VFD's fault info register.  The type of this register was
+        changed from float to u32 in version 2.10.  The commit was
+        done 2022-12-17.
```
SebKuzminsky, line 264: This kind of user-facing changelog information belongs in [Updating LinuxCNC](http://linuxcnc.org/docs/devel/html/getting-started/updating-linuxcnc.html), not in the manpage.

SebKuzminsky (changes requested):
Looks good, except for the nitpick about changelog location.

c-morley:
Could this be done with s32 rather then u32 or do you need to extra range? 
u32 is awkward to use in many components ie classicladder.

andypugh:
I think U32 is the right choice if the intention is to break it out into bits. It matches the bitslice component. 
It's a bitmap, so signed does not make any sense (even if it is only 16 bits)

c-morley:
Then maybe have both pins? HAL pins are cheap.

petterreinholdtsen:
[Sebastian Kuzminsky]
> @SebKuzminsky requested changes on this pull request.
>
> Looks good, except for the nitpick about changelog location.
>
>> +        The VFD's fault info register.  The type of this register was
> +        changed from float to u32 in version 2.10.  The commit was
> +        done 2022-12-17.
>
> This kind of user-facing changelog information belongs in [Updating
> LinuxCNC](http://linuxcnc.org/docs/devel/html/getting-started/updating-linuxcnc.html),
> not in the manpage.

The idea is to make sure those using the fault-info value today who
experience unexpected behaviour when they upgrade to 2.10, should be
able to find a clue why in the man page of the failing module.

I find it valuable when information about when a feature was introduced
is included in manual pages, like it is for example in several linux
syscall man pages and debhelper(7).

I have nothing against listing it in
getting-started/updating-linuxcnc.adoc.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[c-morley]
> Then maybe have both pins? HAL pins are cheap.

I left the old fault-info float pin alone, and introduced a new
fault-info-bits pin instead, to stay backwards compatible while also
providing a u32 with easier interpretation of the error bits.  Updated
documentation to match.

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
Ahh I see, that works too.

persei802:
I misspoke earlier. This register has 26 fault codes, not bits. It is a 16 bit register. Anyway, I ran Petter's modified code and it works fine. I have no further issues with this component.

docs/src/man/man1/hy_gt_vfd.1.adoc
SebKuzminsky: The patch was reworked to not break backward compatibility, so no note is needed in `Updating LinuxCNC`.

petterreinholdtsen:
[Jim]
> I misspoke earlier. This register has 26 fault codes, not bits. It is
> a 16 bit register. Anyway, I ran Petter's modified code and it works
> fine. I have no further issues with this component.

Aha.  Adjusted documentation to drop the 'bit' claim.  Where is the list
of fault codes documnented?

-- 
Happy hacking
Petter Reinholdtsen

persei802:
They are in the GT Series Inverter Manual.
fault_codes

petterreinholdtsen:
[Jim]
> They are in the GT Series Inverter Manual.

Thank you.  Added this reference to the man page, and renamed the new
u32 pin from fault-info-bits to fault-info-code, as it is not a bit
field.

I believe this is ready for merging.
-- 
Happy hacking
Petter Reinholdtsen

Issue #2212, opened 2022-12-18
Clang 15 error with LinuxCNC


NTULINUX:
## Here are the steps I follow to reproduce the issue:

 1. Build LinuxCNC with Clang
 2. make
 3. wait

## This is what I expected to happen:

Successful build (worked a few months ago, maybe something changed with Clang 15?)

## This is what happened instead:

error: unknown warning option `-Wno-stringop-truncation`; did you mean `-Wno-string-concatenation`? [-Werror,-Wunknown-warning-option]

## It worked properly before this:

LinuxCNC from roughly 4 months ago with Clang 14

## Information about my hardware and software:

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`): Custom Gentoo
 * I am using this kernel version (shown by `uname -a`): 6.1
 * I am running ...
   * [ ] A binary version from linuxcnc.org (including buildbot.linuxcnc.org)
   * [X] A binary I built myself
   * [ ] A binary version from some other source besides linuxcnc.org
 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`): 6fc2d5beab2cf6cc14793f7b34ed616eb9e64987
 * I am using this user interface (GUI) (e.g., AXIS, Touchy, gmoccapy, etc): N/A
 * I am using this interface hardware vendor and chipset (e.g., parallel port, ethernet port, FPGA card): N/A

NTULINUX:
Rats... another one just popped up:

Compiling realtime objects/hal/components/sim_home_switch.c
hal/components/reset.comp:76Compiling realtime objects/hal/components/sim_matrix_kb.c
:5: error: void function '_' should not returnCompiling realtime objects/hal/components/sim_parport.c
 a value [-Wreturn-type]
    return 0;

Not sure how to properly fix this one..

NTULINUX:
I don't know if -Wno-unknown-warning-option is the right fix here, feel free to amend or make suggestions!

petterreinholdtsen:
The correct fix for reset is to drop the return statement, I believe.

petterreinholdtsen:
[NTU]
> I don't know if -Wno-unknown-warning-option is the right fix here,
> feel free to amend or make suggestions!

I suspect a better fix is to move the GCC only options to
src/configure.ac and only activate activate it for gcc.

-- 
Happy hacking
Petter Reinholdtsen

NTULINUX:
> [NTU]
[...]

PR #2213, opened 2022-12-18
Fix build with Clang

NTULINUX:


petterreinholdtsen:
Related to  issue #2212.

petterreinholdtsen:
I suspect using -Wno-unknown-warning-option might be the wrong approach.  It would be better to move -Wno-stringop-truncation to a place where it is only active with gcc, or perhaps drop the flag alltogether, as it hides questionable code.  See https://developers.redhat.com/blog/2018/05/24/detecting-string-truncation-with-gcc-8 for some info on the issues.

petterreinholdtsen:
Asked @SebKuzminsky about the return statement, and he said it should go away.  Thus I went with my patch in #2214  and not this patch.

PR #2234, opened 2023-01-01
NML: include jjogmode when logging EMC_JOG commands

SebKuzminsky:
I'm really not a fan of CMS/RCS/NML, but replacing it with something modern, standard, and easier to use will be a huge task.

### DIFF
--- a/src/emc/nml_intf/emc.cc
+++ b/src/emc/nml_intf/emc.cc
@@ -980,6 +980,7 @@ void EMC_JOG_ABS::update(CMS * cms)
     EMC_JOG_CMD_MSG::update(cms);
     cms->update(pos);
     cms->update(vel);
+    cms->update(jjogmode);
 
 }
 
@@ -990,6 +991,7 @@ void EMC_JOG_STOP::update(CMS * cms)
 {
 
     EMC_JOG_CMD_MSG::update(cms);
+    cms->update(jjogmode);
 
 }
 
@@ -1314,6 +1316,7 @@ void EMC_JOG_CONT::update(CMS * cms)
 
     EMC_JOG_CMD_MSG::update(cms);
     cms->update(vel);
+    cms->update(jjogmode);
 
 }
 
@@ -2117,6 +2120,7 @@ void EMC_JOG_INCR::update(CMS * cms)
     EMC_JOG_CMD_MSG::update(cms);
     cms->update(incr);
     cms->update(vel);
+    cms->update(jjogmode);
 
 }
 


petterreinholdtsen:
Which issue is this solving / changing?  What about a new tests/ case demonstrating the problem?

SebKuzminsky:
A new field was added to these four NML messages in fd985dba, but the code that renders the messages to a text string was not updated.  This PR updates the text-rendering code to show the new field that's been in the messages since 2015.

These text strings are used only for debug logging, to help humans understand the flow of commands through the motion controller.  This PR does not affect the motion controller's behavior in any other way.

Here's a demo of the change.

Before the current PR (on a freshly started LinuxCNC instance, just out of E-stop and powered on):
~~~
Issuing EMC_JOG_CONT --          (  +124,+40,   +16,    +0,0.283333,)
Issuing EMC_JOG_STOP --          (  +136,+32,   +17,    +0,)
Issuing EMC_JOG_INCR --          (  +125,+48,   +18,    +1,0.100000,0.283333,)
Issuing EMC_TRAJ_SET_TELEOP_ENABLE --    (  +230,+104,   +19,    +0,)
Issuing EMC_JOINT_HOME --        (  +123,+24,   +20,    -1,)
Issuing EMC_JOG_INCR --          (  +125,+48,   +21,    +0,0.100000,0.283333,)
Issuing EMC_JOG_CONT --          (  +124,+40,   +22,    +1,-0.283333,)
Issuing EMC_JOG_STOP --          (  +136,+32,   +23,    +1,)
~~~

Note that the jogs before and after homing look indisinguishable.

And after this PR (again with a freshly powered on LinuxCNC):
~~~
Issuing EMC_JOG_CONT --          (  +124,+40,   +16,    +0,0.283333,    +1,)
Issuing EMC_JOG_STOP --          (  +136,+32,   +17,    +0,    +1,)
Issuing EMC_JOG_INCR --          (  +125,+48,   +18,    +1,1.000000,0.283333,    +1,)
Issuing EMC_TRAJ_SET_TELEOP_ENABLE --    (  +230,+104,   +19,    +0,)
Issuing EMC_JOINT_HOME --        (  +123,+24,   +20,    -1,)
Issuing EMC_JOG_INCR --          (  +125,+48,   +21,    +0,1.000000,-0.283333,    +0,)
Issuing EMC_JOG_CONT --          (  +124,+40,   +22,    +1,0.283333,    +0,)
Issuing EMC_JOG_STOP --          (  +136,+32,   +23,    +1,    +0,)
~~~

Now we see the extra field at the end of the JOG messages.  1 for joint-mode jogs (Free trajector planner) and 0 for axis-mode jogs (Teleop trajectory planner).

(Homing must  happen in Free mode, which is why the UI disabled Teleop before requesting homing. Task automatically switches from Free to Teleop after successfully homing, which is why we don't see the UI requesting that change.)

SebKuzminsky:
As a tangent, at some point we should discuss replacing CMS/RCS/NML (the messaging system used between the non-realtime parts of LinuxCNC, ie between UIs and Task and between IO and Task, but not between Motion and Task) with a more standard, modern system.  Something like protobuf or one of the many alternatives.

The use of shared memory for message passing between the non-realtime parts of LinuxCNC is not important to me at all, and probably hurts us in these day of fast and well-understood loopback networking

PR #2240, opened 2023-01-03
task: fail if the user's requested interp can't be loaded

SebKuzminsky:
The INI file may override which interpreter Task uses, via [TASK]INTERPRETER.

Before this commit, if the user's specified interpreter failed to load, Task would log a warning but then fall back to using the default interp and continue.

After this commit, if the user's specified interpreter fails to load, Task refuses to start.

### DIFF
--- a/src/emc/task/emctask.cc
+++ b/src/emc/task/emctask.cc
@@ -443,6 +443,10 @@ int emcTaskPlanInit()
 	if((inistring = inifile.Find("INTERPRETER", "TASK"))) {
 	    pinterp = interp_from_shlib(inistring);
 	    fprintf(stderr, "interp_from_shlib() -> %p\n", pinterp);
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
[Sebastian Kuzminsky]
> After this commit, if the user&#39;s specified interpreter fails to
> load, Task refuses to start.

This sound like something to mention in the release notes, as it would
cause existing setups to suddenly stop working.  I realise those setups
are currently broken, but it would be nice to explain in the
documentation that this might happen and how to solve it.

--
Happy hacking
Petter Reinholdtsen

SebKuzminsky:
I've reworked the pluggable interpreter part of Task a tiny bit more as part of fixing canterp, I'm closing this PR in favor of a coming canterp PR.

PR #2241, opened 2023-01-03
docs: semantic line breaks, bit of syntax. (smoe/docs29_mac_11)

smoe:


docs/src/plasma/plasma-cnc-primer.adoc
```diff
@@ -510,8 +528,8 @@ This is more appropriate for the 1 kHz servo thread and also allows more time
 
 There is a lot of confusion around how to decode the THCAD output so lets consider the Mesa 7I76e and the THCAD-10 for a moment with the following hypothetical calibration data:
 
-* Full scale 928 kHz (1/32 29000 Hz)
-* 0 volt 121.6 kHz (1/32 3800 Hz)
+* Full scale 928 kHz (1/32 29 kHz = 29000 Hz)
+* 0 V 121.6 kHz (1/32 3.8 kHz = 3800 Hz)
```
hansu, line 532: This is misleading (and mathematically wrong).
Either 29 kHz or 29000 Hz. I think our users are tough enough to know that.
```suggestion
* Full scale ≙ 928 kHz (1/32 29 kHz)
* 0 V ≙ 121.6 kHz (1/32 3.8 kHz)
```

hansu (changes requested):

smoe:
I have some good confidence that I have not introduced anything that was mathematically wrong, just the whole paragraph is so unfortunate. My interpretation is that we are supposed to learn that the value for 1V change is HzPerVolt = (Hz~max~ - Hz~min~) / (V~max~ - V~min~), which is not stated anywhere. And the reverse calculation for the 5V example would then be Hz~min~ + 5*HzPerVolt - I did not look at the value they came up with. If this is not how it is supposed to work then I think I would want to have another scale that does work linearly :)

smoe:
Resolved conflict (which implied a rebase on 2.9).

hansu:
> I have some good confidence that I have not introduced anything that was mathematically wrong, just the whole paragraph is so unfortunate. 

I mean this:

`Full scale 928 kHz (1/32 29 kHz = 29000 Hz)`

and this would be interpreted as

`Full scale 928 kHz (1/32 * 29 kHz = 29000 Hz)`

But `1/32 * 29 kHz ≠ 29000 Hz`



So just drop the "= 29000 Hz"

smoe:
How about
`Full scale 928 kHz ( * 1/32 = 29 kHz = 29000 Hz)`

hansu:
> How about `Full scale 928 kHz ( * 1/32 = 29 kHz = 29000 Hz)`

:+1: 
That is at least correct even though I have no idea about the relevance of the 1/32 value.

smoe:
The original is already bad and I suggest to just accept my change to then jointly improve it afterwards.

docs/src/plasma/plasma-cnc-primer.adoc
```diff
@@ -510,8 +528,8 @@ This is more appropriate for the 1 kHz servo thread and also allows more time
 
 There is a lot of confusion around how to decode the THCAD output so lets consider the Mesa 7I76e and the THCAD-10 for a moment with the following hypothetical calibration data:
 
-* Full scale 928 kHz (1/32 29000 Hz)
-* 0 volt 121.6 kHz (1/32 3800 Hz)
+* Full scale ≙ 928 kHz (1/32 29 kHz)
+* 0 V ≙ 121.6 kHz (1/32 3.8 kHz)
```
hansu, line 532:
```suggestion
* Full scale ≙ 928 kHz (928 kHz/32 = 29 kHz)
* 0 V ≙ 121.6 kHz (121.6 kHz/32 = 3.8 kHz)
```

hansu:
> The original is already bad and I suggest to just accept my change to then jointly improve it afterwards.

Yes, then please add that or my most recent suggestion.
My first suggestion was unfortunately wrong :innocent:

smoe:
You suggestions was lovely. Committed.

hansu (approved):

hansu:
> You suggestions was lovely. Committed.

Thanks

I'll merge some of your PR's tomorrow when I can merge them up directly after to avoid those conflicts we had.

PR #2242, opened 2023-01-03
Stricter pluggable interp

SebKuzminsky:
If the user specifies an invalid `[TASK]INTERPRETER`, error out and refuse to start.

Before this PR, Task would warn but then fall back to the default G-code interpreter.


SebKuzminsky:
Ah yeah,

> Also has a minor change in the semantics of a non-absolute path

Oh yeah, it now looks in `/usr/lib/linunxcnc` instead of `/usr/lib/emc2`.

SebKuzminsky:
As you might have guessed, fixing canterp is what led me on this sidetrack...  #2244

PR #2243, opened 2023-01-03
docs: changes triggered while translating (docs_fixing_xfc_man_page)

smoe:


docs/src/plasma/qtplasmac.adoc
```diff
@@ -901,22 +901,22 @@ For more information, see the <<qt_pm_comms,PowerMax Communications>> section.
...
 [width="100%",cols="4,16",options="header"]
 |===
-|Name  |Description
-|FLOAT |Indicates that the float switch is activated.
-|OHMIC |Indicates that the probe has sensed the material.
-|BREAK |Indicates that the torch breakaway sensor is activated.
+|Name   |Description
+m|FLOAT |Indicates that the float switch is activated.
+m|OHMIC |Indicates that the probe has sensed the material.
+m|BREAK |Indicates that the torch breakaway sensor is activated.
```
hansu, line 907: IMHO the monospace font does not really fit here. The names in the left column are just the button names.

Also for the following "m"s.
Furthermore a description could be added what the tables represent and the table header could be named more specific. Like "button name instead if just "name".

smoe:
@hansu - But those ugly all-caps names have the role of an identifier, so I think having them monospaced is ok. Anyway - I think we both want to do something more important. Let's agree to disagree on this one and I'll close the PR.

hansu:
They are just all upper case in the GUI. And that doesn't make them to a command or similar.
grafik


But your changes to the other files are good and I would like to keep them...

smoe:
That QtVCP screenshot looks nice, not "ugly" as I had associated those all-caps - sorry for that. I may be easier to prepare a new branch with just those other changes, but I am swamped, don't want to touch it.

PR #2249, opened 2023-01-06
include the new xyzab-tdr gui in linuxcnc.deb

SebKuzminsky:


petterreinholdtsen:
Look good to me.

One thing that occured to me is that these binaries are not really ment to be run directly by users, and should probably be moved somewhere under /usr/lib/ instead, and this location should be placed in PATH in the linuxcnc wrapper script.  Perhaps /usr/lib/linuxcnc/bin/ could be a sensible location.

SebKuzminsky:
> One thing that occured to me is that these binaries are not really ment to be run directly by users, and should probably be moved somewhere under /usr/lib/ instead, and this location should be placed in PATH in the linuxcnc wrapper script. Perhaps /usr/lib/linuxcnc/bin/ could be a sensible location.

I like this suggestion.

PR #2250, opened 2023-01-07
Fix `merge -s ours`

SebKuzminsky:
This is my attempt at a partial fix for the recent mistaken `git merge -s ours` (1d836df8).

This PR cherry-picks most of the commits from 2.9 between the most recent good merge (d788b67f) and the new, bad merge (1d836df8).  I used `git cherry-pick -x` so that the new commit message includes the SHA of the original commit from 2.9, to help future archaeologists.

Some of the 2.9 commits were already in master (thanks to Chris's 624647fb and Dewey's 3d3bdb72).

I had problems with nearly all the commits that touch the docs: 1c6e3cb5, 245ea867, f2b9835e, ec02e63f, 9d57d998, 28878842, 3df87ab3, 0da277b2, f9d19a4d, 802f497f, 1dfa6e42, 59d94f66, f52eaf62, 2cd49cc3, 2cc3cffd, a9472ef6, a4c68b0d, e404459a.

I skipped those, and would like someone who knows the docs better to handle them.

I also didn't deal with the pncconf conflict in 4de6a4da - it *looked* like a different patch was already in master; Chris, can you take a look at these?

PR #2251, opened 2023-01-07
Alternate bad merge fix

jepler:
Hi.

I wanted to make a different attempt at fixing the merge problem introduced at 1d836df as a check on @SebKuzminsky's work (thanks @SebKuzminsky)

Here's what I did:
 * Start a new branch at `1d836df^`, the commit before the problematic one
 * Merge origin/2.9 into this commit. For each doc conflict (there were conflicts in 4 files), take the one that looks "more semantic" in its line breaking
 * Attempt to cherry-pick each non-merge commit after 1d836df (only one generated a non-empty commit)
 * "git merge -s origin/master" to subsume the 'other side of history' without taking its changes (ironic, I know)

Here are the differences from seb's patch (- side) to my patch (+ side): https://gist.github.com/a3f9e75449b23a3817d91f4fdd83244f

As far as I can tell the differences do not touch any functionality, only documentation and some help text.

Probably the differences arise because Seb skipped merging commits that touched docs and caused conflicts, while I resolved the conflicts (albeit somewhat arbitrarily).


SebKuzminsky:
This seems like a better solution than #2250.

PR #2253, opened 2023-01-08
New HAL component for publishing status to MQTT broker

petterreinholdtsen:
Include runtest script.

The code to include machine name in MQTT message is not working.

### DIFF
--- a/debian/linuxcnc.install.in
+++ b/debian/linuxcnc.install.in
@@ -57,6 +57,7 @@ usr/bin/melfagui
 usr/bin/milltask
 usr/bin/millturngui
 usr/bin/mitsub_vfd
+usr/bin/mqtt-publisher
 usr/bin/z_level_compensation
 usr/bin/monitor-xhc-hb04
 usr/bin/motion-logger
@@ -201,6 +202,7 @@ usr/share/man/man1/mitsub_vfd.1
 usr/share/man/man1/monitor-xhc-hb04.1
 usr/share/man/man1/motion-logger.1
 usr/share/man/man1/moveoff_gui.1
+usr/share/man/man1/mqtt-publisher.1
 usr/share/man/man1/ngcgui.1
 usr/share/man/man1/panelui.1
 usr/share/man/man1/pi500_vfd.1
--- a/docs/src/man/man1/mqtt-publisher.1.adoc
+++ b/docs/src/man/man1/mqtt-publisher.1.adoc
@@ -1,0 +1,128 @@
+= mqtt-publisher(1)
+
+== NAME
+mqtt-publisher - send HAL pin data to MQTT broker periodically
+
+== SYNOPSIS
+
+*loadusr -W mqtt-publisher [options] keys*=pin1[,pin2...]
+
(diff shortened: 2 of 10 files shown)


SebKuzminsky:
This PR probably should update the dependencies of linuxcnc-uspace.deb to include the paho client library.

SebKuzminsky:
I pushed "rebuild" on the Action with the failed download from Ubuntu's servers and it passed.

Some comments on the code:

* The `names` argument to components is by convention used to name the component itself, not pins it makes.  Perhaps something like `--input-pins` would be less surprising?  <http://linuxcnc.org/docs/devel/html/hal/comp.html#_instance_creation>
* Speaking of input pins, some way to make pins of other data types might be useful.  For example, the [`sampler`](http://linuxcnc.org/docs/devel/html/man/man9/sampler.9.html) component does this.
* You should probably move parsing of the INI file out of the main loop, and cache the value.  
* `usage()` doesn't match `options`.

andypugh:
>  I did consider something like names=x_s32,working_bit or names=x[s32],working[bit], but would prefer to follow some
> already existing convention. Not sure if such convertion exist.

Pick your convention....
GladeVCP used hal.pin-f and hal.pin-s
https://linuxcnc.org/docs/html/gui/gladevcp.html#_widget_and_hal_pin_naming

mux_generic uses hal.pin-s32. hald.pin-u32, hal.pin-float etc:
https://linuxcnc.org/docs/stable/html/man/man9/mux_generic.9.html

Smart-serial devices leave you to guess.

petterreinholdtsen:
[andypugh]
> Pick your convention....

Thank you for the pointers.

I had not really considered including the type in the pin name, but that
might indeed be a good idea.

> GladeVCP used hal.pin-f and hal.pin-s
> https://linuxcnc.org/docs/html/gui/gladevcp.html#_widget_and_hal_pin_naming
>
> mux_generic uses hal.pin-s32. hald.pin-u32, hal.pin-float etc:
> https://linuxcnc.org/docs/stable/html/man/man9/mux_generic.9.html

I prefer full names over the short version, I must admit.

> Smart-serial devices leave you to guess.

I take it you talk about sserial(9) here, and the fact that the pin
names do not include any information about their type.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
> I take it you talk about sserial(9) here, and the fact that the pin names do not include any information about their type.
> […](#)

Yes. I am responsible for the sserial situation (and mux-generic too) so was scathing about sserial. It's also worth noting that I am not even consistent in the own code.

https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/drivers/mesa-hostmot2/sserial.h#L132
Is one way to handle mutable pin types. I have forgotten why I did not do the same thing with the HAL pins, it might be something to do with the pre-programmed a-bit-like-DMA mode that is supported in hm2 but not currently implemented.

petterreinholdtsen:
I dropped the controversial and non-working parts (ie the default set of published values).  I believe it is ready to be merged into 2.9 now.

Issue #2204, opened 2022-12-12
Halcompile does not include modbus headers..............


Roguish000:
Ok.  running on both a real machine, and a VM, of debian 12 bookworm, with the latest and greatest versions of linuxcnc and everything associated, including a RIP build, all master braches.

Testing on an **unmodified** existing user comp.
I tried to use halcompile to create a very slightly version of the vfdb_vfd.c comp.  Tried all kinds of things.
Finally was able to read the halcompile message better.    See the image.
Note the -I in the line.  '-I/home/noel/comp -I/usr/include -I/usr/include/linuxcnc'
But the modbus header files are in  /usr/include/modbus
So they are not found.
I copied all 4 modbus header files into /usr/include/linuxcnc.....

and voila.  it worked.  well not really.  got past that error...............now more errors.  see the second, longer image.
 still not sure, not a programmer. 
but this is an existing and working user comp.
shouldn't it compile correctly?

A little help please......

![Screenshot at 2022-12-11 16-06-19](https://user-images.githubusercontent.com/42012033/206937322-80ae7a07-9e81-427c-8f8b-b4269c9f78c2.png)

![Screenshot at 2022-12-11 16-16-59](https://user-images.githubusercontent.com/42012033/206937927-dba3af9f-20aa-4084-8aff-54d7a7e10ef3.png)

petterreinholdtsen:
[Roguish]
> A little help please......

I had a look, and to me it look like your component need some extra -I
to the compile step and -l arguments to the linker step.  I had a look
at the source of halcompile to try to figure out how to do so, and the
trick seem to be two options, extra_compile_args and extra_link_args
that can be added to a .comp file, for example like this:

  option extra_compile_args "-I/usr/include/modbus";
  option extra_link_args "-lm -lmodbus";

The former should fix your problem including <modbus.h>, and the latter
should fix the linking issue with round and modbus_* methods.  Not sure
where the iniFindInt method come from, so no library suggestion there.

It was not obvious to me how to activate similar options for .c based
components, so I am not sure if this solve your problem.

In any case I suspect the should be documented in the halcompile(1)
manual page, which where I failed to find them.  They are documented in
<URL: http://linuxcnc.org/docs/devel/html/hal/comp.html#_options >.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
There might be a way to pass the flag via a 

#pragma GCC ... 

Line in the source file. 

[...]

PR #2254, opened 2023-01-09
halcompile: Allow definition of compiler and linker flags

andypugh:
Currently halcompile allows a .comp file to define: option extra_compile_args
option extra_link_args
This works for .comp files but can not be used for .c files. See https://github.com/LinuxCNC/linuxcnc/issues/2204


petterreinholdtsen:
[andypugh]
> Currently halcompile allows a .comp file to define: option
> extra_compile_args option extra_link_args This works for .comp files
> but can not be used for .c files. See
> https://github.com/LinuxCNC/linuxcnc/issues/2204

Good idea.  

What about a test case in tests/ to make sure this keep working?

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
OK, test added. I chose to use the "readline" library as the external reference. I hope that isn't a bad choice.

andypugh:
This has become a total mess. I am going to start again.

PR #2255, opened 2023-01-09
Merge 2.8 into 2.9

SebKuzminsky:
This is a manual (i.e. not via a direct PR on github) merge of 2.8 into 2.9.  The automatic merge fails because `docs/src/remap/remap.txt` was renamed to `docs/src/remap/remap.adoc`.


hansu:
I think this is a very good workflow for merging up if merge conflicts are existing. So everyone can see if they are resolved correctly before it gets merged.

c-morley:
How about getting in the habit of merging your work right after committing it?
Our problem last weekend started with doc work that didn't get merged and had conflicts.
That meant I couldn't merge my pncconf work (which is very hard for anyone else to merge)
By the time Dewey got there there was a stack of merge work.

It's easiest for everyone if the person who caused the merge conflicts fixes them ASAP, so they only have to deal with their own code.

Do yuo know if it is possible to set up github to test merges-that would be helpful too?

hansu:
>How about getting in the habit of merging your work right after committing it?

Yeah that sounds reasonable.

>Do yuo know if it is possible to set up github to test merges-that would be helpful too?

What exactly do you have in mind? Github tells you when there is a conflict...

c-morley:
isn't that test for feature branch to branch?
I'm suggesting if we could get github to test 'up merges' after a feature branch is commit to,

ie:
if my work is committed to 2.9, then github or (buildbot I suppose) tests merging 2.9 to master.

hansu:
But what if that test fails? This shouldn't be a blocker to get it merged. More a warning that there will be merge conflicts after this merge which has to be fixed immediately.

c-morley:
Yes I was thinking a warning not a block -  but it should be dealt with ASAP.

c-morley:
I mean it would be easier not to merge at all . but it seems that is a bridge too far for everyone :)
This is the next best I can currently think of.

hansu:
I think conflicts can't be avoided in every case. So I would go for a warning.

PR #2261, opened 2023-01-11
merge remap fixes up to master

SebKuzminsky:
These fixes originally went into 2.8, and I just merged 2.8 into 2.9.  This PR brings those fixes up to master.


configs/sim/axis/remap/getting-started/python/remap.py
```diff
@@ -45,7 +45,7 @@ def involute(self, **words):
         self.set_errormsg("feedrate > 0 required")
         return INTERP_ERROR
 
-    if equal(self.speed,0.0):
+    if equal(self.speed[0], 0.0):
```
smoe, line 48: This is quite a difference. Well-spotted.

PR #2265, opened 2023-01-12
Gmoccapy suggests installing onboard

smoe:


hansu:
How does this suggestion look like?

smoe:
> I think this is a good idea 👍. How does this suggestion look like?

```
./src/emc/usr_intf/gmoccapy/gmoccapy.py:                print (_("**** try sudo apt-get install onboard"))
./src/emc/usr_intf/gmoccapy/gmoccapy.py:            print (_("**** try sudo apt-get install onboard"))
./docs/src/plasma/qtplasmac.adoc:sudo apt install onboard
```

hansu:
No, I mean will the installation stop at some point and a question like "do you want to install 'onboard'?" will wait for the user input, or ...?

smoe:
No, that does not happen. I just meant to point out that when the program code that onboard should be installed, so should likely also the .deb . If we had a separate .deb for GMOCCAPY (like linuxcnc-gmoccapy) then I would make the onboard dependency a clear recommendation.

hansu:
Still don't know what this addition exactly does, but makes sense having this as suggestion.

hansu (approved):

smoe:
I do not think this has any effect, except to increase the awareness for that dependency of LinuxCNC on the onboard package.

PR #2269, opened 2023-01-15
mb2hal: fix pin type change for version 1.000

sensille:
With the introduction of version 1.1 one pin changed type for version 1.0. I think this was by accident. With this fix old config files stay compatible.

### DIFF
--- a/src/hal/user_comps/mb2hal/mb2hal_hal.c
+++ b/src/hal/user_comps/mb2hal/mb2hal_hal.c
@@ -177,13 +177,13 @@ retCode create_each_mb_tx_hal_pins(mb_tx_t *mb_tx)
         case mbtx_06_WRITE_SINGLE_REGISTER:
         case mbtx_16_WRITE_MULTIPLE_REGISTERS:
             if (gbl.version < 1001) {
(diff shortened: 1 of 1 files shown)


hansu:
Good catch, thank you!
But this fix should also go into 2.9, so please rebase and change the target branch.

petterreinholdtsen:
[Arne Jansen]
> With the introduction of version 1.1 one pin changed type for version
> 1.0. I think this was by accident. With this fix old config files stay
> compatible.

Is there a good way to test the behaviour of the different versions
using a script in tests/, to ensure this keep working and and unexpected
changes are detected in the future?

See <URL: http://linuxcnc.org/docs/devel/html/code/writing-tests.html >
for some clues on how to write tests.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
> Is there a good way to test the behaviour of the different versions using a script in tests/, to ensure this keep working and and unexpected changes are detected in the future?

That is already on my to-do list... I'll try to create a test today.

sensille:
force push wasn't the best idea it seems, still fighting with github ...

hansu:
> force push wasn't the best idea it seems, still fighting with github ...

I shouldn't be more than
```
(git checkout mb2hal_type_fix)
git rebase upstream/2.9
git push -f
```
and then changing the target branch in Github.

PR #2271, opened 2023-01-15
hostmot2 driver patch to make litehm2 work

sensille:
This patch enables the litehm2 FPGA port (https://github.com/sensille/litehm2) to work with linuxcnc.

This enables basic support for the board. More support like port naming might follow in the future, I'd like to build some kind of auto-discovery there. But for now it is sufficient to fully use the board.

### DIFF
--- a/docs/man/man9/hm2_eth.9
+++ b/docs/man/man9/hm2_eth.9
@@ -40,7 +40,8 @@ As shipped, the board address is 192.168.1.121.
 .SH DESCRIPTION
 
 hm2_eth is a device driver that interfaces Mesa's ethernet based Anything I/O boards (with the HostMot2 firmware) to the LinuxCNC HAL.
-The supported boards are: 7I76E, 7I80DB, 7I80HD, 7I92, 7I93, 7I94, 7I95, 7I96, 7I96S, 7I97, 7I98.
+The supported boards are: 7I76E, 7I80DB, 7I80HD, 7I92, 7I93, 7I94, 7I95, 7I96, 7I96S, 7I97, 7I98. It also supports boards with the
+litehm2 firmware (https://github.com/sensille/litehm2).
 The board must have its firmware loaded on the board by the mesaflash(1) program.
 
 hm2_eth is only available when LinuxCNC is configured with "uspace" realtime.
--- a/docs/man/man9/hostmot2.9
+++ b/docs/man/man9/hostmot2.9
@@ -28,7 +28,7 @@ The identifier consists of the last 4 digits of the board serial number, which i
 This will make configs less portable, but does mean that boards can be re-connected less carefully.
 
 .SH DESCRIPTION
-hostmot2 is a device driver that interfaces the Mesa HostMot2 firmware to the LinuxCNC HAL.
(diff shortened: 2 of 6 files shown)


petterreinholdtsen:
[Arne Jansen]
> This patch enables the litehm2 FPGA port
> (https://github.com/sensille/litehm2) to work with linuxcnc.

Should the litehm2 board be mentioned in the documentation for hostmot2?

Is there a way to test if this work as intended using a script in
tests/?
--
Happy hacking
Petter Reinholdtsen

sensille:
> Should the litehm2 board be mentioned in the documentation for hostmot2?

That would be nice, maybe here, in the introduction? 
http://linuxcnc.org/docs/2.9/html/drivers/hostmot2.html
Or is there a place for a more general overview of linuxcnc, listing the available hardware options?
Can you recommend a place?

> Is there a way to test if this work as intended using a script in tests/? -- Happy hacking Petter Reinholdtsen

That could be hard. It would need a fake mesa card replying via network, maybe running on localhost? But that would take some significant effort. Do you think it is worth it?

hansu:
Unfortunately we don't have a really good overview page which hardware is supported.
But I agree, some documentation for that should be added.
Furthermore I think this better fits in master, as 2.9 should only receive bugfixes and documentation updates.

petterreinholdtsen:
[Arne Jansen]
> Can you recommend a place?

I do not really know hostmot2 enough to have an opinion here.  I just
checked the content of docs/ to see if litehm2 were mentioned, and could
not find anything.

> That could be hard. It would need a fake mesa card replying via
> network, maybe running on localhost? But that would take some
> significant effort. Do you think it is worth it?

No idea.  Sound like a lot of work with many pitfalls.

-- 
Happy hacking
Petter Reinholdtsen

sensille:
> Unfortunately we don't have a really good overview page which hardware is supported. But I agree, some documentation for that should be added. Furthermore I think this better fits in master, as 2.9 should only receive bugfixes and documentation updates.

I can re-target the patch against master if you deem that to be more appropiate, but I'm unsure what to do about the documentation. Maybe I can wait a few days to see if someone comes up with a good suggestion.

sensille:
Force-push, rebased against master. As discussed in IRC I also added a bit of documentation. Let me know if this goes in the right direction.

andypugh:
> Unfortunately we don't have a really good overview page which hardware is supported.

Not in the docs, but a good starting point is this Wiki page: 
http://wiki.linuxcnc.org/cgi-bin/wiki.pl?LinuxCNC_Supported_Hardware

andypugh:
Has this patch been tested to make sure that it doesn't break Mesa card function?

sensille:
I don't have a Mesa card, so I can't test it. But the new functionality is only used if the card identifies as "litehm2".

sensille:
Maybe someone with access to mesa hardware can give this patch a quick test?

andypugh:
I tried the patch with a Mesa 7i96. It took some effort to find a working set of hardware, but I did load the driver and see it successfully read the position of an encoder. So it is probably all OK.

SebKuzminsky:
Thanks @sensille !

sensille:
Thanks for testing and merging!

PR #2272, opened 2023-01-15
document linuxcnc command line options

hansu:
Documented all command line option of the linuxcnc script in the man page as well as in the help text of the script itself.

I am not sure what the option `-r` does. 

It sets `RUNTESTS=yes` and for my understanding this will result that stdout and stderr are not being written to these files:
```
if [ -z $RUNTESTS ]; then
if ! tty -s; then
    exec 2>> $DEBUG_FILE
    exec >> $PRINT_FILE
fi
fi
```
But when I tested that, it makes no difference.

### DIFF
--- a/docs/src/man/man1/linuxcnc.1.in
+++ b/docs/src/man/man1/linuxcnc.1.in
@@ -28,7 +28,8 @@
 linuxcnc \- LinuxCNC (The Enhanced Machine Controller)
 .SH SYNOPSIS
 .B linuxcnc
-[\fI-v\fR] [\fI-d\fR] [\fIINI file\fR]
+[\fI-h\fR] [\fI-v\fR] [\fI-d\fR] [\fI-r\fR] [\fI-l\fR] [\fI-k\fR] [\fI-t <tpmodulename> [parameters]\fR]
+[\fI-m <homemodulename> [parameters]\fR] [\fI-H <dirname>\fR] [\fIINI file\fR]
 .SH DESCRIPTION
 \fBlinuxcnc\fR is used to start LinuxCNC (The Enhanced Machine Controller). It
 starts the realtime system and then initializes a number of LinuxCNC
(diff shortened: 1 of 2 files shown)


petterreinholdtsen:
[Hans Unzner]
> But when I tested that, it makes no difference.

Look like it only make a difference when not running from a terminal.
Did you try on the command line?  Perhaps try with '< /dev/null'?
-- 
Happy hacking
Petter Reinholdtsen

hansu:
`tty -s` gives nothing, so the inversion of that should do 
```
    exec 2>> $DEBUG_FILE
    exec >> $PRINT_FILE
```

if  `$RUNTESTS` is empty.

petterreinholdtsen:
[Hans Unzner]
> `tty -s` gives nothing, so the inversion of that should do

Not quite sure what you mean here.  Perhaps this example make my comment
easier to understand:

  % if tty -s ; then echo tty; else echo no tty; fi;if tty -s ; then echo \
    tty; else echo no tty; fi < /dev/null
  tty
  no tty
  %

-- 
Happy hacking
Petter Reinholdtsen

hansu:
Ok yes I got it. 
So do you think my description of the "-r" flag is correct?

petterreinholdtsen:
[Hans Unzner]
> Ok yes I got it. 
> So do you think my description of the "-r" flag is correct?

Not quite.  Instead of your proposal:

  "For running tests: Doesn't print version number and config dir and
  doesn't write stdout and stderr to files"

I would try something along this line.

  Disable redirection of stdout and stderr to ~/linuxcnc_print.txt and
  ~/linuxcnc_debug.txt when stdin is not a tty.  Used when running
  linuxcnc tests non-interactively.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
>I would try something along this line.
>
>  Disable redirection of stdout and stderr to ~/linuxcnc_print.txt and
  ~/linuxcnc_debug.txt when stdin is not a tty.  Used when running
  linuxcnc tests non-interactively.
  
  Thanks

PR #2280, opened 2023-01-17
Added man page cross reference between spindle(9) and motion(9).

petterreinholdtsen:
The motion component have some spindle pins, so it would be useful to guide readers between them.


hansu:
Could be into 2.9

petterreinholdtsen:
[Hans Unzner]
> Could be into 2.9

Sorry, my mistake.  Was targeting 2.9, but forgot to set the branch when
creating the github PR.  Fixed now.

--
Happy hacking
Petter Reinholdtsen

PR #2282, opened 2023-01-17
Merge mb2hal test from 2.8

hansu:
Test 1/2 should fail because in 2.9 there is an additional line in the debug output.

Not sure whats the best workflow to fix that.


petterreinholdtsen:
[Hans Unzner]
> Test 1/2 should fail because in 2.9 there is an additional line in the
> debug output.
>
> Not sure whats the best workflow to fix that.

I suspect adding a commit on top of the merge from 2.9 to fix the test
is a good way to handle it.  This way the patch from 2.8 is recorded as
merged into 2.9, and the test keep working.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
I think I would prefer the workflow which Sebastian did in https://github.com/LinuxCNC/linuxcnc/pull/2255:
1. merge 2.8 into a temporary branch
2. add the fix for the test
3. merge the temporary branch via Github

petterreinholdtsen:
[Hans Unzner]
> I think I would prefer the workflow which Sebastian did in
> https://github.com/LinuxCNC/linuxcnc/pull/2255:
> 1. merge 2.8 into a temporary branch
> 2. add the fix for the test
> 3. merge the temporary branch via Github

I thought that was what I described, as this change is in a PR already (
in other words qtemporary branch based on 2.9) and the missing steps are
2 or 3.

--
Happy hacking
Petter Reinholdtsen

hansu:
>I thought that was what I described, as this change is in a PR already 

I though you meant to put a commit on top of 2.9 after this PR has been merged.
But good, that we agree on that :).

hansu:
I had to create a new PR for that: https://github.com/LinuxCNC/linuxcnc/pull/2283

PR #2285, opened 2023-01-18
Qtdragon keyboard jograte switch

sensille:
CMorley recently added the feature to change the jog rate via keyboard shortcut in qtdragon. I'd like to propose an improvement to it by making the scale exponential.

The original thread is here:

https://forum.linuxcnc.org/qtvcp/47962-keyboard-jogging-in-qtdragon

CMorley also added the feature that the mode automatically switches to manual on keyboard jogging, as it does with onscreen-jogging. It was added to qtdragon. This PR also contains a commit that adds the same code to qtdragon/hd.

### DIFF
--- a/lib/python/qtvcp/qt_action.py
+++ b/lib/python/qtvcp/qt_action.py
@@ -1,4 +1,5 @@
 import os
+import math
 import subprocess
 from time import sleep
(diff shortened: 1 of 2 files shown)


c-morley:
Thank you for the code. It has been incorporated now.

PR #2288, opened 2023-01-18
Hm2 bspi fixes

SebKuzminsky:
This PR makes the Mesa 7i65 board inputs work right for us.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/bspi.c
+++ b/src/hal/drivers/mesa-hostmot2/bspi.c
@@ -148,6 +148,8 @@ int hm2_tram_add_bspi_frame(char *name, int chan, rtapi_u32 **wbuff, rtapi_u32 *
         return -1;
     }
     if (rbuff != NULL){
+        // Reading from addr[0] instead of addr[chan] is intentional
+        // here - all the channels share one receive FIFO.
         r = hm2_register_tram_read_region(hm2,hm2->bspi.instance[i].addr[0], sizeof(rtapi_u32),rbuff);
         if (r < 0) {
             HM2_ERR( "Failed to add TRAM read entry for %s\n", name);
--- a/src/hal/drivers/mesa-hostmot2/bspi.c
+++ b/src/hal/drivers/mesa-hostmot2/bspi.c
@@ -179,6 +179,26 @@ int hm2_allocate_bspi_tram(char* name)
     return 0;
 }
 
+EXPORT_SYMBOL_GPL(hm2_bspi_clear_fifo);
+int hm2_bspi_clear_fifo(char * name)
+{
+    hostmot2_t * hm2;
+    int i, r;
+
+    i = hm2_get_bspi(&hm2, name);
+    if (i < 0){
+        HM2_ERR_NO_LL("Can not find BSPI instance %s.\n", name);
+        return -1;
+    }
+    rtapi_u32 zero = 0;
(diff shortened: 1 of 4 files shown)


andypugh:
What prompted this? I wasn't aware of an issue.

SebKuzminsky:
Sorry, this change was a bit rushed, and as it turns out it may have been in error...

@jepler and @cradek and I are building a new machine, using a 7i80 and a 7i65.  It's probably also worth mentioning that we're using a custom 7i80 firmware from @pcw-mesa.

We were observing bogus results when reading the SPI peripherals on the 7i65:
* The "misc input" digital input pins on the CPLD would flip randomly, even when the physical input pins were explicitly driven high or low.
* The voltages we presented to the SPI ADC inputs (e.g. by connecting an analog input pin to ground or +5) would appear on the incorrect HAL pin.  For example, we'd drive AIN0 to +5, and the voltage would show up on the `mesa-7i65.0.analogue.07.in` pin instead of `mesa-7i65.0.analogue.00.in`.
* One or another of the analog input HAL pins would read a perfectly steady 6.25V, which is the voltage you'd get if you interpreted the CPLD input SPI frame as an ADC value.  Driving the misc input pins high or low would change this "voltage" in the expected ways.
* The mapping between physical analog input pins and HAL pins (and the analog input pin in HAL that appeared to carry the "CPLD Misc Input" frame) would change over time.       

We (well, Jeff, who's the brains of our group) observed the following things:
* In EXTRA_SETUP(), the ADCs are initialized by writing to BSPI channel 3: https://github.com/LinuxCNC/linuxcnc/blob/2.8/src/hal/drivers/mesa_7i65.comp#L195-L201
* BSPI channel 3 has Echo enabled: https://github.com/LinuxCNC/linuxcnc/blob/2.8/src/hal/drivers/mesa_7i65.comp#L173-L174
* ... Which means that the writes to configure the ADCs also perform reads, the results of which will be queued in the receive FIFO.
* In general, there is no synchronization or guarantee when reading from the BSPI FIFO.  If there's old stuff in the BSPI receive FIFO, or if reading from it underflows, the driver wouldn't know. 


From this we we formed the theory that there was unexpected stuff in the BSPI receive FIFO, *before* the values the driver expects based on the TRAM packet.  Those junk frames in the receive FIFO would get read out before the actual frames that the driver expects.  The old junk frames in the FIFO would still get interpreted according to the TRAM structure, and parsed into analog and digital hal pins, leading to the corruption we were seeing.

We inserted the "write anything to the BSPI FIFO Counter register in order to clear the FIFO" change, and it appeared to fix the problem.  So we merged to 2.8 and merged up to 2.9, where we needed it.

But then we rebooted and the problem came back, so we must be mistaken about something here...

We'd love extra eyeballs on this.  Are the 7i65 BSPI inputs (ADC and CPLD digital inputs) working right for you @andypugh?  Were they working before our change, for example on 2.8.4?

andypugh:
I do have a 7i65, but I am busy each evening until the weekend, then off to the Alps to slide down them. So I don't know when I will be able to get to it. 

This thread might warrant further study: https://sourceforge.net/p/emc/mailman/message/29739585/

pcw-mesa:
One improvement I can see is is for the read function to check if the expected amount of data is in the RX FIFO
and if not (due to dropped packets etc) clear the FIFO, use the previous cycles read data and clear the FIFO
on the next write function.

PR #2293, opened 2023-01-21
docs: move halmodule + change title

hansu:
Edit: 
Some restructuring and putting the subsection [Helpful functions](https://linuxcnc.org/docs/2.9/html/hal/halmodule.html#_helpful_functions) into a new section named "HAL Python Module".

Here a bit visualized:

image


andypugh:
I am not sure about this. I always saw this as the users page version of the halcompile docs.

c-morley:
How about
top one - HAL python module
bottom one - linuxcnc python module

hansu:
Hmm maybe the document should be split and the chapter "[6. Helpful Functions](http://linuxcnc.org/docs/2.9/html/hal/halmodule.html#_helpful_functions)"  transferred to a new document like "HAL python module" ?

c-morley:
I guess whats more important - that when learning about say HAL you see there is a python module for making HAL components or
finding out linuxcnc uses python alot and here are all the modules.
I personally think the first is better. Though a python chapter with a small description that links might be useful too.

Maybe we should poll our users?
There are other subjects I find a little strange or buried (like gcode programming: other codes)
But I'm a developer so probably looking for different things.

hansu:
I think the stuff described in that chapter is more for advanced users which want to create own components/VCPs.
When learning about HAL or setting up a machine it's more about connecting present HAL pins or modifying them.

c-morley:
Should move HALUI up then.
it would be interesting to know the amount of people that use linuxcnc with out using 'advanced' features.  I bet it's pretty low.

I think we should poll about the manual on the forum.

silopolis:
Agreed 👍
Just, I still don't understand why move it in the 'UI Programming' section..
Actually, I would split out the Glade_VCP and Qt_VCP of the document and
move them to the respective VCPs docs, with backlinks.
Rename the file as you suggest, but move it to the 'Developer Information'
section

hansu:
>Should move HALUI up then.

I agree.

What about this now?

I also thought "Filter Programs" fits better in the config section. 
Not sure where "Vismach Virtual Machines" fits best. Maybe GUI? Or config?

SebKuzminsky:
What's the status of this PR?

hansu:
My first suggestion didn't find acceptance and my last version / comment didn't receive any feedback.

hansu:
Preview: https://hansu.github.io/linuxcnc-doc/f9ff23f4042b58b2c0ab06d81319626d0d890f6e/html/index.html

hansu:

PR #2294, opened 2023-01-21
Reduced range of conv_float_u64 component to avoid clang build error

petterreinholdtsen:
Clang error out when using 18446744073709551616u:

  error: integer literal is too large to be represented in any integer type

Reduce range by one as a workaround.


petterreinholdtsen:
Follow up for #2034.

hansu:
Looks good, thanks!

PR #2295, opened 2023-01-22
docs: add qtvcp options in man page

hansu:
Adds the recently added options to the docs and add the options to the man page.


c-morley:
Thank you!

PR #2296, opened 2023-01-23
Halcompile permissions

SebKuzminsky:
This fixes the permissions of `.so` uspace realtime components built by halcompile.  For some reason they come out of gcc executable, which doesn't hurt anything but still isn't right.  This just chmods them after building to turn off the `x` bit.

Also adds checks to existing halcompile tests that verify that "userspace" (non-realtime) components are executable (they are) and realtime components are not executable (they aren't, after this PR).


smoe:
The best-possible fix would probably be to find out why the exact reason why that executable flag was set. I only found https://github.com/LinuxCNC/linuxcnc/blob/e929156e312b567877019034e578f3f848007fb5/src/hal/utils/halcompile.g#L1233 .
Either way, I approve this change.

PR #2298, opened 2023-01-24
Mb2hal followup #2278

sensille:
This is a followup to #2278 where we decided to only add checks where needed and check against the version instead. #2278 also missed a place where a crash still could happen.


hansu:
This makes the code better understandable in adding checks only where needed and makes it more clear why they are needed. Thanks Arne!

PR #2299, opened 2023-01-25
Fix canterp.so installation

dwrobel:
Fixes installation of canterp.so (introduced in
c54043254a432d214b296183545c5779d9e1b6cd).

Building rpm packages fails as follows:
    error: Installed (but unpackaged) file(s) found:
        /usr/lib64/linuxcnc

This is a result of mismatch between
  - creating directories using: $(DESTDIR)/lib/linuxcnc

  - installing canterp.so using: $(FILE) ../lib/linuxcnc/canterp.so $(DESTDIR)$(libdir)/linuxcnc

It manifests on systems where "${libdir}" differs from "lib".

Note: this does not fixes loading canterp.so which uses hardcoded path[1].

[1] https://github.com/LinuxCNC/linuxcnc/commit/1f733ebaf9065f5c2304c34168964f8c9605b0dc#diff-e4daea21a6d335db374a201d7386d70ef7ba440ee0d87aa462c87a051be68cffR39

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -666,7 +666,7 @@ install-docs:
 install-dirs:
 	$(DIR) $(DESTDIR)$(EMC2_RTLIB_DIR) \
 		$(DESTDIR)$(sysconfdir)/linuxcnc $(DESTDIR)$(bindir) $(DESTDIR)$(libdir) \
-		$(DESTDIR)/lib/linuxcnc $(DESTDIR)$(includedir)/linuxcnc \
+		$(DESTDIR)/$(libdir)/linuxcnc $(DESTDIR)$(includedir)/linuxcnc \
 		$(DESTDIR)$(docsdir) $(DESTDIR)$(ncfilesdir) \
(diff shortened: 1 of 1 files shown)


SebKuzminsky:
Thanks for the fix!

dwrobel:
>Thanks for the fix!

```
$ git blame Makefile | grep lib\/linuxcnc\ 
1c9e07e187b src/Makefile (CMorley                 2023-03-31 18:33:23 -0700  670) 		$(DESTDIR)/lib/linuxcnc $(DESTDIR)$(includedir)/linuxcnc \
```
I'm not sure what was the reason, but I'm observing that this PR was silently reverted by https://github.com/LinuxCNC/linuxcnc/commit/1c9e07e187b951be568e0cc73564091f417ca6d1.

Was there something wrong with my patch?
Should I re-submit it, or maybe something was wrong with the revert?
BTW, I didn't check what else got reverted.

c-morley:
Master was accidentally merged to 2.9so we reverted it.
I'm sure more things got reverted too as I'm no git expert.
please submit your patch again.

Issue #2125, opened 2022-11-03
Halscope crash bug


pcw-mesa:
Halscope in master has crashed multiple times when selecting the # of samples pane.
What is bad is that you must exit LinuxCNC to be able to run halscope again as you get
this error if you try to re-launch halscope from LinuxCNC:

HAL: ERROR: duplicate component name 'halscope'
SCOPE: ERROR: hal_init() failed

SebKuzminsky:
I just spent a bunch of time in Halscope (on Debian Bullseye and Debian Bookworm), including changing the number of samples, and did not notice any problems.  What platform did you see this on?

pcw-mesa:
X86, Mint 20.3
The times I have seen this happen are when halscope has been running for a long time (days)
I

pcw-mesa:
Thanks, that will help.
BTW here's a backtrace, not sure if it of any use:

#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:65
#1  0x00007f930508ee95 in __vfprintf_internal
    (s=s@entry=0x7ffe1d515be0, format=format@entry=0x560f98f565d9 "%s\n%s", ap=ap@entry=0x7ffe1d515d60, mode_flags=mode_flags@entry=2) at vfprintf-internal.c:1688
#2  0x00007f93050a211a in __vsnprintf_internal (string=0x7ffe1d535330 "", maxlen=<optimized out>, 
    maxlen@entry=94, format=0x560f98f565d9 "%s\n%s", args=args@entry=0x7ffe1d515d60, mode_flags=mode_flags@entry=2)
    at vsnprintf.c:114
#3  0x00007f9305143fd1 in ___snprintf_chk
    (s=s@entry=0x7ffe1d535330 "", maxlen=maxlen@entry=94, flag=flag@entry=1, slen=slen@entry=94, format=format@entry=0x560f98f565d9 "%s\n%s") at snprintf_chk.c:38
#4  0x0000560f98f5205d in snprintf (__fmt=0x560f98f565d9 "%s\n%s", __n=94, __s=0x7ffe1d535330 "")
    at /usr/include/x86_64-linux-gnu/bits/stdio2.h:67
#5  draw_waveform (chan_num=<optimized out>, highlight=highlight@entry=1) at hal/utils/scope_disp.c:894
#6  0x0000560f98f52a82 in refresh_display () at hal/utils/scope_disp.c:258
[...]

PR #2302, opened 2023-01-26
halscope: Add some missing "channel valid" checks

jepler:
There's also some confusion about whether "selected" channel of 0 or -1 denotes "none/invalid".

One of these may have been the cause of a crash @cradek saw in halscope on arm64.

### DIFF
--- a/src/hal/utils/scope_disp.c
+++ b/src/hal/utils/scope_disp.c
@@ -233,12 +233,13 @@ void refresh_display(void)
 	    draw_baseline(n + 1, FALSE);
 	}
     }
-    if (vert->chan_enabled[vert->selected - 1]) {
+    if (vert->selected > 0 && vert->chan_enabled[vert->selected - 1]) {
         draw_baseline(vert->selected, TRUE);
     }
 
     /* Draw trigger line */
-    if (vert->chan_enabled[ctrl_shm->trig_chan - 1]) {
+    int trig_chan = ctrl_shm->trig_chan - 1 ;
+    if (trig_chan >= 0 && trig_chan < 16 && vert->chan_enabled[trig_chan]) {
         draw_triggerline(ctrl_shm->trig_chan,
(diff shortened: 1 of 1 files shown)


src/hal/utils/scope_disp.c
```diff
@@ -556,12 +557,12 @@ static double snap(int y) {
 }
 
 static void left_drag(int dy, int y, GdkModifierType state) {
+    if(vert->selected <= 0) return;
+
```
cradek, line 561: This doesn't build -- needs to be moved back down.

jepler: weird, I thought we switched to a modern standard that supported interspersing declarations & other statements. I guess that's in master only, not 2.9.

It has to be before one of the array indexes of `vert->selected`, otherwise there's still undefined behavior (even though it's probably OK on all normal computers)

jepler: oh, oops, OK. :blush:

SebKuzminsky:
Is the crash reproducible?  Does this fix it?

cradek:
These changes fix the crash for me.  I can reproduce it reliably on the pi400 by turning on (aka leaving at the default setting) compositing in xfce, have linuxcnc running, delete autosave.halscope, unloadrt scope_rt, and then starting halscope.  It flashes both windows briefly, and then crashes.

(I accidentally discovered that turning off compositing worked around the problem.)

I don't think it has been reproduced on any other machine.  Jeff failed to reproduce on his amd64.

samcoinc:
Pcw had mentioned that he was seeing halscope crash.   Where was that..

samcoinc:
this is what I was thinking of,,

https://github.com/LinuxCNC/linuxcnc/issues/2125#issuecomment-1302214135

cradek:
That sure seems like it might be the same problem.

PR #2303, opened 2023-01-26
Execute LOG commands at runtime

d2inventory:
Log commands in G-Code are executed during the interpreted phase.

This changes it so the commands are executed when the program is running.
Like this the timestamps represent the real time and not the time when the G-Code-file was read and interpreted.

### DIFF
--- a/src/emc/nml_intf/emc.hh
+++ b/src/emc/nml_intf/emc.hh
@@ -334,6 +334,9 @@ extern int emcOperatorText(int id, const char *fmt, ...) __attribute__((format(p
 // print note to operator
 extern int emcOperatorDisplay(int id, const char *fmt, ...) __attribute__((format(printf,2,3)));
 
+// log text to file
+extern int emcLogCmd(int id, const char *text);
+
 // implementation functions for EMC_AXIS types
 
 extern int emcAxisSetUnits(int axis, double units);
--- a/src/emc/nml_intf/emc_nml.hh
+++ b/src/emc/nml_intf/emc_nml.hh
@@ -49,6 +49,12 @@ class EMC_OPERATOR_ERROR:public RCS_CMD_MSG {
     char error[LINELEN];
 };
 
+
+#define EMC_OPERATOR_TEXT_ID_TEXT       0
+#define EMC_OPERATOR_TEXT_ID_LOGOPEN    1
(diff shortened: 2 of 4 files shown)


petterreinholdtsen:
[d2inventory]
> Log commands in G-Code are executed during the interpreted phase.
>
> This changes it so the commands are executed when the program is running.
> Like this the timestamps represent the real time and not the time when
> the G-Code-file was read and interpreted.

Interesting change.  Why do you want the timestamps to be more accurate?
Which issue did you run into with the delayed timestamps.  I ask only
out of curiosity, and wondering if the MQTT component I am working on
might need more accurate timestamping too.

-- 
Happy hacking
Petter Reinholdtsen

d2inventory:
This is not about accuracy of timestamps.

When you load a .ngc file the interpreter will translate the text into executable objects (more or less lol)
At this stage the log commands are executed and written into the file.

Later on when I run this program the log commands are ignored.

So for example if I run a program and write a log message whenever I change the tool to see how fast I am I don't get any logs. This PR changes this, so the logs are written when the program is executed rather than just when the file is read from the hard disk.

andypugh:
I think I very much agree with this change in principle, though I won't be able to test it for a few days. 
I do wonder if it works properly in kernel-space realtime? I assume that your testing has been in userspace / preempt-rt?

d2inventory:
Your assumption is very much right andy :D

hansu:
Hmm but what if you want to debug an erroneous G-code file with help of these log commands. The errors you may want to identify occur while the interpreter goes initially through the file and not while actually "executing" the the G-code.

If someone other shares my concerns, maybe we can add another type of log message?
If not, just forget about it ;-)

d2inventory:
@hansu this is indeed true, I didn't think about that.

Would doing both be a solution?
The file shouldn't be overwritten until the program is executed. So works for debugging with the interpreter.
If someone uses just append the entries will be doubled, but I guess that's an issue the User can fix themself.

d2inventory:
This change breaks stuff like smartprobe.ngc so I'm going to close it, especially because I can't reproduce the issue I had about a year ago when I wrote this to fix it.

andypugh:
I still think that logging actions probably belong in the queue, though in the case of probing (which is a queue-buster) this matters less than it would otherwise.

PR #2305, opened 2023-01-26
Make manual tool change more verbose

d2inventory:
hal_manualtoolchange.py accepts the ini-file path as an argument in order to read the tooltable-file.

This allows to dialog to display tool diameter and comments.

https://i.imgur.com/MTFFabB.png

### DIFF
--- a/share/axis/tcl/dialog.tcl
+++ b/share/axis/tcl/dialog.tcl
@@ -129,7 +129,7 @@ proc nf_dialog {w title text image default args} {
     # database for -wraplength and -font so that they can be
     # overridden by the caller).
 
-    option add *Dialog.msg.wrapLength 3i widgetDefault
+    option add *Dialog.msg.wrapLength 7i widgetDefault
     if {![string compare $tcl_platform(platform) "macintosh"]} {
 	option add *Dialog.msg.font system widgetDefault
     } else {
--- a/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py
+++ b/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py
@@ -59,11 +59,17 @@ def get_tool_change_message(n):
         inidir = os.path.dirname(inipath)
         inifile = linuxcnc.ini(inipath)
         tooltable_file = inifile.find("EMCIO", "TOOL_TABLE")
+        tooltable_dbline = inifile.find("EMCIO", "DB_PROGRAM")
         machine_units = inifile.find("TRAJ", "LINEAR_UNITS")
+
+        # check if a db_program is in use for tool changes
+        if tooltable_dbline:
+            return _("Insert tool %d and click continue when ready") % n
+
         # make sure we get an absolute path to the tool table
-        if (tooltable_file != ""):
-            if (not os.path.isabs(tooltable_file)):
-                tooltable_file = os.path.join(inidir, tooltable_file)
+        if not os.path.isabs(tooltable_file):
+            tooltable_file = os.path.join(inidir, tooltable_file)
+
         # load the tool table file
         tool_info = get_tool_info(tooltable_file, n)
         diameter = ("%f" % tool_info["diameter"]).rstrip("0").rstrip(".")
@@ -75,8 +81,8 @@ def get_tool_change_message(n):
         })
     except Exception as error:
         # old style message with just tool number and the error message
-        return "".join((_("Insert tool %d and click continue when ready") % n,
-                        _("\n\nError: %s") % error))
+        return "".join(_("Insert tool %d and click continue when ready") % n,
+                        _("\n\nError: %s") % error)
 
 def get_tool_info(file, n):
     with open(file, "r") as f:
@@ -91,9 +97,9 @@ def parse_line(line):
     parts = line.partition(";")
     data["comment"] = parts[2]
     for token in parts[0].upper().split(" "):
-        if (token.startswith("T")):
+        if token.startswith("T"):
             data["number"] = int(token[1:])
-        if (token.startswith("D")):
+        if token.startswith("D"):
             data["diameter"] = float(token[1:])
     return data
 
--- a/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py
+++ b/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py
@@ -35,7 +35,7 @@ def stop_polling_hal_in_background():
 
 def do_change(n):
     if n:
-        message = _("Insert tool %d and click continue when ready") % n
+        message = get_tool_change_message(n)
     else:
         message = _("Remove the tool and click continue when ready")
     app.wm_withdraw()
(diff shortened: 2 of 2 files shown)


dngarrett:
For generality, hal_manualtoolchange should handle
ini files that use:

   [EMCIO]DB_PROGRAM = prog_name

In such cases, [EMCIO]TOOL_TABLE is not required
(and is ignored if included) so no tool data will be
available from a file.   hal_manualtoolchange
could use old style Exception message for this case also.


Ref:
http://linuxcnc.org/docs/devel/html/tooldatabase/tooldatabase.html

> Tool data is conventionally described by a tool
> table file specified by an inifile setting:
> [EMCIO]TOOL_TABLE=tooltable_filename. A tool table
> file consists of a text line for each available
> tool describing the tool’s parameters, see Tool
> Table Format.
>
> The tool database interface provides an
> alternative method for obtaining tool data via a
> separate program that manages a database of tools.
> ... 
> Note INI file settings for [EMCIO]TOOL_TABLE are
> ignored when a db_program is specified.

Ref sim configs:
https://github.com/LinuxCNC/linuxcnc/tree/master/configs/sim/axis/db_demo

d2inventory:
I don't think the overlap of manual tool change and using a tool DB is very high but ofc this shouldn't prevent me from including it too.

The issue I have right now is that I can't get your db_demo to work with manual tool change.

I tried to remove the -no_use_hal_manualtoolchange option from this line in base.inc:

`HALFILE = LIB:basic_sim.tcl -no_use_hal_manualtoolchange`

This results in HAL errors, which I tried to trivially solve by commenting the offending lines out in db_postgui.hal which ofc didn't work :smile: 

Could you tell me how to get it to work using the manual tool changer so I can easily test it? Thx.

dngarrett:
The configs/sim/db_demo/*ini do not use hal_manualtoolchange
because they configure a primitive toolchanger in hal

However, in general, using hal_manualtoolchange could be
used in conjunction with a [EMCIO]DB_PROGRAM
#================================================================================
A simple demonstration of using a DB_PROGRAM with
hal_manualtoolchange can be made with by editing 
 configs/sim/axis/axis.ini to create configs/sim/axis/axis_mod.ini:
as follows:

$ diff axis.ini axis_mod.ini
103c103,104
< TOOL_TABLE = sim.tbl
---
> #TOOL_TABLE = sim.tbl
> DB_PROGRAM = db_demo/db_nonran.py

1) linuxcnc axis_mod.ini
2) F1,F2,Ctrl-Home
3) MDI: t10m6
Results in message from hal_manualtoolchange:
Error: expected str, bytes or os.PathLike object, not NoneType

#================================================================================
The root problem i noticed is:
+ tooltable_file = inifile.find("EMCIO", "TOOL_TABLE")
may return with tooltable_file=None
which is not handled at:

+        if (tooltable_file != ""):
+            if (not os.path.isabs(tooltable_file)):
+                tooltable_file = os.path.join(inidir, tooltable_file)

Resulting in exception:
Error: expected str, bytes or os.PathLike object, not NoneType
#================================================================================
A candidate solution:
diff --git a/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py b/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py
index 0acf7ae22b..5dd6ef95d1 100644
--- a/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py
+++ b/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py
@@ -59,6 +59,9 @@ def get_tool_change_message(n):
         inidir = os.path.dirname(inipath)
         inifile = linuxcnc.ini(inipath)
         tooltable_file = inifile.find("EMCIO", "TOOL_TABLE")
+        if tooltable_file is None:
+            # old style message with just tool number and the error message
+            return "".join((_("Insert tool %d and click continue when ready") % n))
         machine_units = inifile.find("TRAJ", "LINEAR_UNITS")
         # make sure we get an absolute path to the tool table
         if (tooltable_file != ""):

dngarrett:
If both [EMCIO]TOOL_TABLE and EMCIO]DB_PROGRAM are
specified, DB_PROGRAM is used and TOOL_TABLE is ignored

To avoid parsing an uncommented [EMCIO]TOOL_TABLE item
when [EMCIO]DB_PROGRAM is specified, the checks used to
supply an old style message (with just tool number)
should test inclusion  of DB_PROGRAM also

```
diff --git a/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py b/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py
index 0acf7ae22b..2eb1b8f96d 100644
--- a/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py
+++ b/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py
@@ -59,6 +59,10 @@ def get_tool_change_message(n):
         inidir = os.path.dirname(inipath)
         inifile = linuxcnc.ini(inipath)
         tooltable_file = inifile.find("EMCIO", "TOOL_TABLE")
+        db_program = inifile.find("EMCIO","DB_PROGRAM")
+        if tooltable_file is None or db_program is not None:
+            # old style message with just tool number and the error message
+            return "".join((_("Insert tool %d and click continue when ready") % n))
         machine_units = inifile.find("TRAJ", "LINEAR_UNITS")
         # make sure we get an absolute path to the tool table
         if (tooltable_file != ""):
```

d2inventory:
I tried to make it work with DB_PROGRAM but I constantly got errors and it didn't work at all.

db_ran.py always gave me an error about tool 15 not being available or something like this.
Maybe it's not a good idea to launch it twice?

Anyway I changed it so that if there is a DB_PROGRAM option set it will fall back to the less verbose version, kinda like you suggested.

I also changed the dialog width a big so it's not so very cramped.

before: https://i.imgur.com/H1vUTRU.png
now: https://i.imgur.com/23pBmfF.png

dngarrett:
Commit 088be31097 incorporates the tool table comment field as part of the tooldata structure and adds a python interface.

This would allow hal_manualtoolchange to retrieve tool comments without reading/parsing an ini file and would work with both [EMCIO]TOOL_TABLE and [EMCIO]DB_PROGRAM

The comment retained is limited to 40 characters as it is staically allocate for all (1000!) tools.

Usage example (using configs/sim/axis/axis.ini):

>>> import linuxcnc as l; s=l.stat(); s.poll(); s.toolinfo(1)
{'toolno': 1, 'pocketno': 1, ... , 'comment': '1/8 end mill '}

Perhaps you could consider a revision to use the new toolinfo() provision.

rene-dev:
good idea, but there is a better way of getting that info. I will look into that.

d2inventory:
I updated it as suggested using linuxcnc.stat() to get the comment. This limits the comment to 39 characters though :-/

andypugh:
Let's see how it goes. Sorry for the delay.

d2inventory:
Hi @andypugh 

How come these changes are not in the newest release?

https://github.com/LinuxCNC/linuxcnc/blob/v2.9.1/src/emc/usr_intf/axis/scripts/hal_manualtoolchange.py

andypugh:
> How come these changes are not in the newest release?

The pull-request was to "master" not to 2.9, so that is where it went.

d2inventory:
Oh, ok. I was confused because my other pull request is in the newest release but this one isn't.

jamwaffles:
Sorry to necro an old PR but I sat down to implement this very feature this morning and saw it was already in the code, which led me here. I can build lcnc from source, but it would be great to see this PR in the next release as I've been frustrated many a time trying to remember which tool number was which. What needs to be done to get this feature in the next built release, and can I help in any way?

hansu:
New features will be only added in the next major release (2.10 is the next major release for us) so it will take some time if you see this in an official release unfortunately.
But you can build your own version or install the development version if you want to use this feature.

PR #2310, opened 2023-01-27
Adding function for carousel HAL component

zz912:
Hello,

I tried use component carousel for encoder without trigger signal: https://forum.linuxcnc.org/24-hal-components/47845-carousel-dont-work-w-o-trigger It did not work. 

So we have a problem when moving from position 9 to 8 and from position 3 to 2. Why is this so:
9 = 2^3+0+0+2^0
8 = 2^3+0+0+0

3 = 0+0+2^1+2^0
2 = 0+0+2^1+0

When I want the revolver, let it turn from position 9 to 8 or from position 3 to 2. Then the engine starts and immediately stops. A situation occurs when, due to manufacturing tolerances, bit ^0 (2^0=1) turns on one millisecond before the rest of the bit, and the program thinks it is already in the position where it should be (but physically it is not) and turns off the rotation relay.

I solved it. I made a modification to the carousel component. The original solution worked in such a way that moving from the first position to the second position was only about checking whether it is already in the correct position. I have several phases. 

The first phase (case 0 and case 1) is the same as it was. ATC is started and the direction of rotation is evaluated.

The second phase (case 102) to leave the first position. I check if the position is already 0 and I don't care how the read position changes. 

In the third phase (case 103), I just check that the position is zero. 

In the fourth phase (case 104), I check that the position has changed to anything else and i make decision, how to continue. Than I am waiting user defined short time (several miliseconds). Than I read code position.  If position is not correct I continue move. If position is correct I continue to next phase. 

In the fifth phase (case 105) I waiting user defined time. It is long time for stabilize mechanical oscilation of toolchanger. Than I check if the position is correct, than I make decision, how to continue. If position is not correct I repeat cycle . If position is correct I continue to next phase.

I needed to make sure that the ATC motor still wouldn't spin when the encoder failed. The Brother TC-211 uses a motor that is not designed for continuous operation. That's why I made the "error_time" parameter. If the ATC cycle length is longer than this parameter, the ATC motor will shut down and the program will hang. The default value of this parameter is 3600s = 1 hour. I don't expect anyone to have a longer cycle ATC. I have this value set to 20 seconds myself.

I would like to ask for help.
1) I would need someone to test these changes on an ATC without an encoder with only indexing. I don't have such an ATC myself, so I can't verify if my changes messed something up.
2) I'm just an amateur programmer, so I need comments about the C language so I don't break any conventions. 
3) I need advice on what else to do in the LinuxuCNC project so as not to break the rules. I guess I'll have to edit the manual? And what else?

### DIFF
--- a/src/hal/components/carousel.comp
+++ b/src/hal/components/carousel.comp
@@ -138,6 +138,12 @@ param r  signed state = 0 "Current component state";
 param r  bit    homing  = 0 "Shows that homing is in progress. Only used for index mode";
 param r  float  timer "Shows the value of the internal timer";
 param r signed motor-dir "Internal tag for search direction";
+param r  float  timer-cycle "Shows the cycle time";
+param rw float read-time = 0 """Time in seconds for reading Gray, Straight Binary or BCD code after some position is founded. 
+Value for encoder with trigger signal can be 0. Recomended value for encoder without trigger signal can be 0.02""";
+param rw float check-time = 0 """Time in seconds to wait to stop mechanical oscilation of toolchanger in a position.
+Recomended value for unreliable ATC positioning can be 1. Only used in Gray, Straight Binary or BCD code.""";
+param rw float error-time = 3600 "Time in seconds to stop motor in error case";
 
 option count_function;
 option extra_setup;
@@ -291,6 +297,18 @@ FUNCTION(_){
     if (unhome && ! old_unhome) homed = 0;
     old_unhome = unhome;
 
+    // measuring cycle time
+    if (state > 0){
+    timer_cycle += fperiod;}
+    if (state == 9){
+    timer_cycle = 0;}
+
+    // stop motor in error case    
+    if (timer_cycle > error_time){    
+    motor_fwd = 0;
+    motor_rev = 0;
+	}
+
     switch (state){
     case 0: // waiting at start
         motor_dir = 0;
@@ -364,7 +382,39 @@ FUNCTION(_){
             motor_vel = fwd_dc;
             motor_dir = 1;
         }
-        state = 2;
+ 
+        if ((inst_code == 'G' || inst_code == 'B' || inst_code == 'D') ){
+		state = 102;
+		} else {
+		state = 2;
+		}
+   case 102: // leaving position    
+        if ((current_position != 0) && enable) return;
+        state = 103;
+        break;
+   case 103: // finded middle position (0 position)    
+        if ((current_position == 0) && enable) return;
+        state = 104;
+        timer = read_time;
+        break;
+   case 104: // finded some position  
+        if ((current_position != mod_pocket) && enable) return;
+        timer -= fperiod;
+        if (timer > 0) return;
+        motor_fwd = 0;
+        motor_rev = 0;
+        state = 105;
+        timer = check_time;
+        break;
+   case 105: // pause for check correction position
+		timer -= fperiod;
+        if (timer > 0) return;
+        if ((current_position != mod_pocket) && enable){
+		state = 0;
+		} else {
+		state = 2;
+		}	 
+        break;                      
     case 2: // moving
         if ((current_position != mod_pocket) && enable) return;
         if (rev_pulse > 0){


andypugh:
The problem of incorrect bit patterns appearing during motion is the reason that the "strobe" input bit exists.
(If your encoder does not offer this, then it is no help)
Gray-code encoding is another solution, as that is designed so that only one bit changes at any one time. 
It sounds a little like the changes you have made will work with your system, but possibly not with some others?

zz912:
> The problem of incorrect bit patterns appearing during motion is the reason that the "strobe" input bit exists.
(If your encoder does not offer this, then it is no help)

My encoder does not offer this. I have Brother TC-211 with original ATC and with original encoder. Therefore, I think people who want to retrofit these machines would appreciate it. The Brother TC-211 is a popular machine for the garage.

> Gray-code encoding is another solution, as that is designed so that only one bit changes at any one time.

Gray-code dont solve problem. I will try explain it.

ATC is in position  2 - in grey 0011
you want position 1 - in grey 0001

In ideal world dont exist any problem.
Timeline:
You are in position 2
0   ms     0011
5   ms     0011
10 ms    0011
15 ms    0011
20 ms    0011
25 ms   0000
30 ms   0000
35ms    0000
.........
.........
1065 ms  0000
1070 ms  0000
1075 ms  0001
You are in position 1


In real world:
Timeline:
You are in position 2
0   ms     0011
5   ms     0011
10 ms    0011
15 ms    0011
20 ms    0011
25 ms   0001
30 ms   0000
35ms    0000
Now you have  problem. In time 25ms Carousel.comp turn off motor and send signal ready to LCNC.  You are between positions 1 and 2.
.......
120 ms the spindle will destroy the ATC  (In case you have a machine with the same ATC type as Brother TC-211. Gray code didn't help you.)

> It sounds a little like the changes you have made will work with your system, but possibly not with some others?

No. I think the changes I made will not affect other ATCs. The problem is that I don't know, I just think. Therefore, I would like to ask someone to test it. I am careful. ATC is not a cheap device.

andypugh:
Would debouncing the encoder bits solve the problem? (initially in HAL using the "debounce" component, but I am open to the idea of building it in to the carousel ,comp) 
Rather than your scheme of checking for zero between pockets (which won't suit some changers) we could just set a debounce time for which the detected pocket needs to be stable before "trusting" it.

zz912:
I experiented with debounce:
https://forum.linuxcnc.org/24-hal-components/47845-carousel-dont-work-w-o-trigger#261009
Maybe my use of debounce was wrong, but it didn't make any improvement.

Would it be acceptable for you to turn on the passage through zero with a parameter?
`param rw bit go_throw_zero = 0 "The position change must pass through the zero position. Suitable for unreliable encoders. Only used in Gray, Straight Binary or BCD code.";`

Rather than integrating debounce into carousel.comp, I'd just make a HAL usage example. I think that the integration of one functional HAL component into another functional HAL component is against the main principle of LinuxCNC KISS (keep it simple stupid).

I have long wondered if there can be an encoder (BCD, Binary or Gray) with position 0 and with some "jump" from one position to another position. I didn't invent anything that could be used in practice. However, I'd rather opened this draft instead of a Pull Request to discuss it.

andypugh:
Can you try this updated version of carousel? It just waits a programmable number of thread cycles before it chooses to "believe" the position feedback (a new parameter, called "debounce")

https://github.com/LinuxCNC/linuxcnc/blob/andypugh/carousel_tweaks/src/hal/components/carousel.comp

andypugh:
It's just a warning, it will still work. (oddly, I don't see that warning) 
(But I have fixed it)

zz912:
The warning disappeared.

Debounce probably doesn't work.

I can set debounce to values 0, 3, 10, 50, 50 , 60000.
None of these values have any effect on the carousel functionality.

andypugh:
That's disappointing. Without any hardware to test with it was a bit of a gamble, but I thought it ought to work. 
Can you pull the latest update that I just made (which makes deb into a parameter) and halscope the actual position and the deb parameter during a jog? 

I understand that you put some effort into your changes, but I think that they over-complicate things, and that this idea (if it was competently executed) would be more generally applicable. Mainly because only a sub-set of feedback mechanisms can ever go to zero between positions.

zz912:
I think this change will help. The bug was that deb was originally declared in a function. Therefore, logically, they could only charge debounce or zero values. It can not be middle 0 and debounce. In every reading function deb = 0.

I'll test this change once I'm at the CNC and give feedback.

I don't think my solution is wrong, but I don't mind if we use your solution.

andypugh:
Doh! I know what was wrong now. I am out of practice. 

````
int deb
````
needed to be:
````
static int deb
````

zz912:
I added one word "static" and it start work good. You can celebrate.

zz912:
Hi Andy,

I experimented with the debounce value and crashed ATC. I'm mad at myself. It's a friend's machine and he has to solve it. This is a cracked zinc alloy casting. I did several non-standard things at once and in manual mode I ran the spindle to the ATC intermediate position.

I forgot my modified carousel.comp wasn't there. My carousel.com has "param rw float check-time". As soon as carousel.comp thinks it is in the right position, it waits for example 1s and checks again if it is in the right position. If not it goes to the correct position again. That's why I also made the "error-time" parameter to stop the ATC engine after some time in a non-standard situation.

I know this is my own fault, but re-checking the position after some time would prevent an accident, for example, if the contacts on the contactors controlling the ATC motor were welded.

I wanted to ask if you are willing to integrate the check-time and error-time parameters into your carousel.comp component?

andypugh:
Are you saying that the debounce in my version doesn't always work and that crashed the machine? Or was something configured wrong?

It does sound like perhaps the tool change routine should do some extra checks before moving the arm. Maybe some HAL logic to stop everything if carousel is "ready" and "pocket-number" != "current-position" ?

zz912:
"Are you saying that the debounce in my version doesn't always work and that crashed the machine? Or was something configured wrong?"

In the situation when I crashed ATC, I had set debounce=0. As I wrote above, the accident with ATC is only my fault.
On the other hand, the debounce value is not TRUE or FALSE, so it must be set. Some positions work with debounce = min. 0 and some positions work with debounce min. 15.

"It does sound like perhaps the tool change routine should do some extra checks before moving the arm. Maybe some HAL logic to stop everything if carousel is "ready" and "pocket-number" != "current-position" ?"

This is what I tried to create. I tried HAL logic STOP after 1 second. That's why I set debounce=0. After that I wanted to make home references. Then bang. I'm an Idiot.
The extra check is in my carousel.comp. You must not make an extra check immediately. You have to wait for inertial forces. The parameter check-time=1 second is used for this.

zz912:
I think, this part is not seafety for heavy ATC. You assume that when you stop the motor, the ATC stops immediately and you set the HAL pin ready=1.

line 391:
            motor_fwd = 0;
            motor_rev = 0;
            motor_vel = hold_dc;
            active = 0;
            if (enable) ready = 1;
            state = 9;

I thought about what I wanted. I would like 2 debounce parameters: 
debounce_run = 20ms
- it is same as debounce now
- to read the code while turning
- advantage is that it does not delay the ATC cycle
- disadvantage is that the value must not be too long, because the code would not be read in time, and it must not be too short, so as not to misread the code. Furthermore, the correct value depends on the speed of ATC movements.

debounce_stop = 500ms
- to read the code while turning off the motor
- advantage is that it is 100% sure that there will be no ATC crash
- disadvantage is that it delays the overall ATC cycle time


The question is what should happen if debounce_stop fails.
I prefer to repeat the cycle of searching for a position. But someone else might prefer, for example, only that the ready signal does not change from 0 to 1.
I would be happy to modify the code for everyone, but we need to agree if you would like it this way and how.

andypugh:
Your proposal looks like it would be quite simple to add, using the existing decel-time input. 
I think this would need special handling of the case where decel-time is nonzero and align-dc is zero where a check was made that the carousel was definitely in-position. 
With your sensor setup (no strobe pin) the option to use a nonzero align-dc would not work (it would get stuck in state 7)

I really should write some actual toolchanger docs, though I would feel like a bit of a fraud as I have never owned a toolchanger, and have only ever configured one (simple) example of one.

zz912:
Hi Andy.

> Your proposal looks like it would be quite simple to add, using the existing decel-time input.
> I think this would need special handling of the case where decel-time is nonzero and align-dc is zero where a check was made that the carousel was definitely in-position.
> With your sensor setup (no strobe pin) the option to use a nonzero align-dc would not work (it would get stuck in state 7)

How does it look like? I have a problem that the machine retrofit that a friend and I are doing is coming to an end. Once the machine is ready, it won't be so easy to test changes in the carousel HAL component. I was lucky that a friend managed to fix the ATC. The ATC revolver part is currently taken off. Therefore, we can now test carousel.comp modifications without restrictions and without the risk of another crash. As soon as the machine is assembled again, my friend will probably not allow me to do any experiments. (If necessary, we can unscrew the revolver part again, but that will be complicated.)

Your debounce modification of the carousel component improved ATC behavior. The action of the revolver is smooth. My modifications made the carousel component reliable, but there were occasional situations where the motor would turn off and on at some problematic positions.

I would like to know if I should modify the carousel component for myself, or are you interested in finishing the modification of the officially released carousel.comp component together, so that it is safe to use for ATC encoders without a trigger signal.

Based on the previous point, I need to start working on the complete ATC cycle. I will easily adapt it to the behavior of the carousel component you modified.


> I really should write some actual toolchanger docs, though I would feel like a bit of a fraud as I have never owned a toolchanger, and have only ever configured one (simple) example of one.

I don't understand why you should feel like a fraud? Your carousel component is so complex that you would need about 20 types of ATC to test everything. And no one can ask that of you.

Feel free to write some documentation:
1) At least users will know what you intended.
2) If you have stupid things or some imperfections, the users will point it out to you and you will be able to redo them.

andypugh:
> On 12 Feb 2023, at 12:27, zz912 ***@***.***> wrote:
> 
> I would like to know if I should modify the carousel component for myself, or are you interested in finishing the modification of the officially released carousel.comp component together, so that it is safe to use for ATC encoders without a trigger signal.

I want to add a wait then check to the end if the routine, so it doesn’t set “ready” if the carousel is not settled in position.

zz912:
> I want to add a wait then check to the end if the routine, so it doesn’t set “ready” if the carousel is not settled in position.

It is good that the "ready" signal is not activated.

Will I get any information that there has been a failure to position? For example, a signal, or a combination of signals or something? I will need to resolve this situation.

zz912:
My solution of using an encoder w/o trigger signal was replaced by Andy's solution using debounce. That's why I'm closing this PR.

andypugh:
> Your proposal looks like it would be quite simple to add, using the existing decel-time input. I think this would need special handling of the case where decel-time is nonzero and align-dc is zero where a check was made that the carousel was definitely in-position.

I have made this change. If decel_time is non-zero then the system waits for the specified time and then does a final check for the carousel position. 
It only sets "ready" if the final position matches the request. Otherwise it prints an error.

zz912:
Thank you.

PR #2311, opened 2023-01-27
2.9: Several cosmetic and minor bugfixes for touchy

cradek:
For the filechooser fix, I confirmed that loading a gcode file in another UI does cause touchy to update the screen correctly.  The rest are cosmetic changes.


andypugh:
Ref the error messages, it would be nice if there was a way to see the previous error message. If you get two at a time then you only see the last. (which is often a consequence of the first, and so not particularly diagnostic)

cradek:
Thanks, yes I could imagine a scrolling list on an additional tab to solve that problem.  But these changes don't make that problem better or worse; they do improve the way it works currently.

PR #2313, opened 2023-01-29
Fix up several UI problems

jepler:
This:
 * Removes the import and use of GLUT. GLUT is never used to create a UI window by Axis or Gremlin.
 * fixes the gremlin commandline program so it starts again
 * fixes exceptions printed on the terminal while running configs/sim/touchy/gladevcp/touchy.ini which may have also affected correct behavior
 * comments on why it's necessary to make direct GLX calls in gremlin, which is obstensibly a Gtk program
 * remove some commented out code

I don't know how these fixes intersect (if at all) with the current problems running axis & gremlin-based UIs on unstable or testing.

### DIFF
--- a/src/emc/usr_intf/gremlin/gremlin.py
+++ b/src/emc/usr_intf/gremlin/gremlin.py
@@ -92,6 +92,10 @@ class StatCanon(rs274.glcanon.GLCanon, rs274.interpret.StatMixin):
 
 
 
+# Gtk is not capable of creating a "legacy" or "compatibility" context, which necessitates
+# descending to the GLX API layer to create the required context. This can only be removed
+# someday when the core drawing routines of Gremlin/AXIS are upgraded to modern OpenGL style,
+# a large undertaking.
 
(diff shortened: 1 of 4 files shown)


src/emc/usr_intf/gremlin/gremlin.py
```diff
@@ -115,12 +115,6 @@ def __init__(self, inifile):
     
         self.xwindow_id = None
 
-        #self.set_has_alpha(True)
-        #'set_has_stencil_buffer',
-        glutInit()
```
SebKuzminsky, line 120: Do you want to delete line 49 too?

jepler: Argh, yes

jepler: Updated

SebKuzminsky:
The build failure is my fault, #2315 should fix it.

c-morley:
Did you guys try Qtvcp? - is this something that applies to it ? The openGL code for all three Axis, Gladevcp and Qtvcp are very similar.

PR #2315, opened 2023-01-29
Fix halcompile tests

SebKuzminsky:
Fix some tests that I updated earlier, so they work in the way our Github actions tests them.


jepler (approved):
Looks plausible, thanks!

PR #2317, opened 2023-01-31
message: Update HAL component docs to match behavior

PetteriAimonen:
The code is implemented as edge=1 meaning rising edge, docs said the opposite.
This commit changes the documentation only.

@sheetcam @andypugh Do either of you want to take a look?

Might make sense to merge this to 2.8 branch also.


andypugh:
Thanks. Looks right to me.

PR #2318, opened 2023-01-31
Update xyzc.ini

looping40:
Typo error fixed in comment header

### DIFF
--- a/configs/sim/axis/geometry/xyzc.ini
+++ b/configs/sim/axis/geometry/xyzc.ini
@@ -1,5 +1,5 @@
 # minimal sim config to demonstrate '!' in [DISPLAY]GEOMETRY
-# showing rotations respecting g4x,g92 offsets
+# showing rotations respecting g5x,g92 offsets
 [EMC]
 VERSION = 1.1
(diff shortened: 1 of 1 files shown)


sebastiankuzminsky:
Does this also respect offsets from [G41 and G42](http://linuxcnc.org/docs/2.9/html/gcode/g-code.html#gcode:g41-g42)?  Maybe the comment should be expanded, not changed?

I haven't looked into it enough to have an opinion, only enough to have a question...

looping40:
From the configs/sim/axis/geometry/README : 
The '!' specifies rotations respecting G5x,G92 offsets

As So I assumed it was a typo error. 
May be we should ask dngarrett for confirmation

dngarrett:
This fix applies to 2.9 so i incorporated in that
branch -- thanks.

There have not been any users of the GEOMETRY '!'
provision (although i have used it on an XZC
machine for several years).  With some testing:
1) fixed missing units conversions for mm systems
2) i noted it does not work as wanted for more
   general systems so added doc note marking
   usage as 'Experimental'

Thanks for trying it, maybe i can work more  on it later

PR #2322, opened 2023-02-04
docs_hal_revisited

smoe:
One of those local branches that had not yet seen the light. Nothing urgent or special.

### DIFF
--- a/docs/src/hal/comp.adoc
+++ b/docs/src/hal/comp.adoc
@@ -12,16 +12,17 @@
 
 == Introduction
 
-Writing a HAL component can be a tedious process, most of it in setup
-calls to 'rtapi_' and 'hal_' functions and associated error checking.
+This section introduces to the compilation HAL components, i.e. the addition of some machinists' knowledge on how to deal with the machine.
+It should be noted that such components do not neccessarily deal with the hardware directly.
+They often do, but not necessarily, e.g. there could be a component to convert between imperial and metric scales,
+so this section does not require to dive into the interaction with hardware.
+
+Writing a HAL component can be a tedious process, most of it in setup calls to 'rtapi_' and 'hal_' functions and associated error checking.
 'halcompile' will write all this code for you, automatically.
+Compiling a HAL component is also much easier when using 'halcompile', whether the component is part of the LinuxCNC source tree, or outside it.
 
-Compiling a HAL component is also much easier when using 'halcompile',
-whether the component is part of the LinuxCNC source tree, or outside it.
-
-For instance, when coded in C, a simple component such as "ddt" is around 80
-lines of code. The equivalent component is very short when written using the
-'halcompile' preprocessor:
+For instance, when coded in C, a simple component such as "ddt" is around 80 lines of code.
+The equivalent component is very short when written using the 'halcompile' preprocessor:
 
 [[code:simple-comp-example]]
 .Simple Component Example
@@ -40,26 +41,26 @@ old = tmp;
 
 == Installing
 
-To compile a component, if a packaged version of LinuxCNC is used, development packages
-have to be installed using either Synaptic from the main menu 'System -> Administration -> Synaptic package manager'
-or running one of the following commands in a terminal window:
+To compile a component, if a packaged version of LinuxCNC is used,
+development packages have to be installed using either Synaptic from the main menu 'System -> Administration -> Synaptic package manager'
+or by running one of the following commands in a terminal window:
 
-.Development packages installation
+.Installation of Development packages for LinuxCNC
 ----
 sudo apt install linuxcnc-dev
 # or
 sudo apt install linuxcnc-uspace-dev
 ----
 
-Another method is using the Synaptic package manager, from the
-Applications menu, to install the `linuxcnc-dev` or `linuxcnc-uspace-dev`
-packages.
+Another method is using the Synaptic package manager, from the Applications menu, to install the `linuxcnc-dev` or `linuxcnc-uspace-dev` packages.
 
 == Using a Component
 
-Components need to be loaded and added to a thread before it can be used.
+Components need to be loaded and added to a thread before it can be employed.
+The provided functionality can then be invoked directly and repeatedly by one of the threads
+or it is called by other components that have their own respective triggers.
 
-.Example
+.Example HAL script for installing a component (ddt) and executing it every millisecond.
 ----
 loadrt threads name1=servo-thread period1=1000000
 loadrt ddt
@@ -69,8 +70,7 @@ addf ddt.0 servo-thread
 More information on 'loadrt' and 'addf' can be found in the
 <<cha:basic-hal-reference,HAL Basics>>.
 
-To test your component you can follow the examples in the
-<<cha:hal-tutorial,HAL Tutorial>>.
+To test your component you can follow the examples in the <<cha:hal-tutorial,HAL Tutorial>>.
 
 == Definitions
 
@@ -82,10 +82,8 @@ To test your component you can follow the examples in the
   functions, and data) but behave independently when their pins, parameters, and data have different values.
 
 * 'singleton' - It is possible for a component to be a "singleton", in which case exactly one instance is created.
-  It seldom makes sense to write a 'singleton' component,
-  unless there can literally only be a single object of that kind in the system
-  (for instance, a component whose purpose is to provide a pin with the current UNIX time,
-  or a hardware driver for the internal PC speaker)
+  It seldom makes sense to write a 'singleton' component, unless there can literally only be a single object of that kind in the system
+  (for instance, a component whose purpose is to provide a pin with the current UNIX time, or a hardware driver for the internal PC speaker).
 
 == Instance creation
 
@@ -103,8 +101,7 @@ This can be useful in components that need the timing information.
 
 == Syntax
 
-A '.comp' file consists of a number of declarations, followed by ';;' on a line of its own,
-followed by C code implementing the module's functions.
+A '.comp' file consists of a number of declarations, followed by ';;' on a line of its own, followed by C code implementing the module's functions.
 
 Declarations include:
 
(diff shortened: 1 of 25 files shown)


hansu:
conflicts here

smoe:
conflicts addressed.

docs/src/hal/comp.adoc
```diff
@@ -40,26 +41,26 @@ old = tmp;
...
 
-To compile a component, if a packaged version of LinuxCNC is used, development packages
-have to be installed using either Synaptic from the main menu 'System -> Administration -> Synaptic package manager'
-or running one of the following commands in a terminal window:
+To compile a component, if a packaged version of LinuxCNC is used,
+development packages have to be installed using either Synaptic from the main menu 'System -> Administration -> Synaptic package manager'
+or by running one of the following commands in a terminal window:
 
-.Development packages installation
+.Installation of Development packages for LinuxCNC
```
hansu, line 48:
```suggestion
.Installation of Development packages of LinuxCNC
```

docs/src/hal/halshow.adoc
```diff
@@ -48,9 +44,7 @@ Notes:
        LinuxCNC must be running for standalone usage.
 ----
 
-Example to limit number of decimal points for floats
-and use a file named my.halshow in the current directory:
-
+.Example to limit number of decimal points for floats and use a file named my.halshow in the current directory:
```
hansu, line 47: Having this all as a caption makes the code snippet a bit isolated (not referenced in text).

docs/src/hal/halshow.adoc
```diff
@@ -345,8 +320,7 @@ Back to the editor where we enter the following command:
 linksp Xpos-cmd ddt.0.in
 ----
 
-Now if we look at the Xpos-cmd signal using the tree node we'll see
-what we've done:
+Now if we look at the `Xpos-cmd` signal using the tree node we'll see what we've done:
```
hansu, line 323:
```suggestion
Now if we look at the `xpos-cmd` signal using the tree node we'll see what we've done:
```

docs/src/hal/haltcl.adoc
```diff
@@ -251,11 +230,11 @@ puts [show sig *vel]
 puts [show sig *acc]
 ----
 
-== Haltcl Interactive
+== `haltcl`` Interactive
```
hansu, line 233:
```suggestion
== Haltcl Interactive
```
Please no monospace fonts in headlines

docs/src/hal/rtcomps.adoc
```diff
@@ -87,32 +92,32 @@ On the step type and control type selected.
 [[sec:stepgen-parameters]]
 === Parameters(((HAL stepgen parameters)))
 
-* '(float) stepgen.<chan>.position-scale' - Steps per position unit.
+* '(float) stepgen._<chan>_.position-scale' - Steps per position unit.
```
hansu, line 95: Have you checked if that renders correctly?

docs/src/hal/rtcomps.adoc
```diff
@@ -283,45 +285,45 @@ The PWM generator supports three different 'output types'.
 
 Each PWM generator will have the following pins:
 
-* '(float) pwmgen.<chan>.value' - Command value, in arbitrary units.
+* '(float) pwmgen._<chan>_.value' - Command value, in arbitrary units.
```
hansu, line 288: (same here)

docs/src/hal/rtcomps.adoc
```diff
@@ -766,51 +768,45 @@ All five outputs have the same frequency, amplitude, and offset.
...
+== `lut5`(((lut5)))
 
-The lut5 component is a 5 input logic component based on a look up table.
+The `lut5` component is a 5 input logic component based on a look up table.
 
-* 'lut5' does not require a floating point thread.
+* `lut5` does not require a floating point thread.
 
-.Loading lut5
+.Loading `lut5``
```
hansu, line 794:
```suggestion
.Loading `lut5`
```

docs/src/hal/tools.adoc
```diff
@@ -141,25 +127,21 @@ pin+joint.1.pos-hard-limit
...
-token-name pairs may appear as single or multiple lines.
+When loading a watchfile with the 'File/Load Watch List' menu item, the token-name pairs may appear as single or multiple lines.
 Blank lines and lines beginning with a # character are ignored.
 
 [[sec:halscope]]
 == Halscope
 
-Halscope is an 'oscilloscope' for the HAL. It lets you capture the
-value of pins, signals, and parameters as a function of time.
+Halscope is an 'oscilloscope' for the HAL. It lets you capture the value of pins, signals, and parameters as a function of time.
```
hansu, line 136:
```suggestion
`Halscope` is an 'oscilloscope' for the HAL. It lets you capture the value of pins, signals, and parameters as a function of time.
```

docs/src/hal/twopass.adoc
```diff
@@ -12,16 +12,13 @@
 
 == TWOPASS
 
-LinuxCNC 2.5 supports TWOPASS processing of LinuxCNC configuration files
-that can help in the modularization and readability of LinuxCNC files.
-(LinuxCNC files are specified in an LinuxCNC INI file in the LinuxCNC stanza
-as [LinuxCNC]HALFILE=filename).
+This section describes an option to use names not numbers to distinguish multiple occurences of a single module in a HAL file.
```
hansu, line 15: You can use names without using TWOPASS which doesn't help loading the module twice. So this wording could be a bit misleading.

docs/src/hal/twopass.adoc
```diff
@@ -40,23 +36,18 @@ loadrt and2 names=aa,ab,ac
...
-and their names since when you add (or remove) a component, you
-must find and update the single loadrt directive applicable to
-the component.
-
-TWOPASS processing is enabled by including an INI file parameter in
-the [HAL] section:
+It can be a maintenance problem to keep track of the components and their names, since when you add (or remove) a component,
+you must find and update the single loadrt directive applicable to the component.
 
+.TWOPASS processing is enabled by including an INI file parameter in the [HAL] section, where "anystring" can be any non-null string.
```
hansu, line 42:
```suggestion
TWOPASS processing is enabled by including an INI file parameter in the [HAL] section, where "anystring" can be any non-null string:

.Enabling TWOPASS processing
```

docs/src/plasma/plasma-cnc-primer.adoc
```diff
@@ -431,7 +431,7 @@ LinuxCNC support this via `M67`/`M68` for analog commands and `M62`-`M65` for di
 How this is implemented is totally arbitrary.
 Lets look at how the LinuxCNC QtPlasmaC configuration does this:
 
-.Select Material Settings In QtPlasmaC And Use The Feedrate For That Material
+.Select Material Settings In QtPlasmaC And Use The Feedrate For That Material.
```
hansu, line 434:
```suggestion
.Select Material Settings in QtPlasmaC and Use the Feedrate for that Material.
```

docs/src/hal/rtcomps.adoc
smoe: I have not, admittedly.

docs/src/hal/tools.adoc
smoe: `halscope`?

hansu: :+1:

docs/src/hal/rtcomps.adoc
smoe: Have now checked in vs code - is fine, would not expect anything else because of the period appearing frequently at the end of sentences and thus needing a low operator precedence.

docs/src/hal/halshow.adoc
smoe: changed back

docs/src/hal/twopass.adoc
smoe: That is how I understood TWOPASS. I addressed everything except for this one now, I think. Could you please correct what I came up with? I doubt that it was better before - otherwise I likely would have understood it better :)

docs/src/hal/halshow.adoc
smoe: Reversed.

docs/src/hal/tools.adoc
smoe: done

docs/src/hal/twopass.adoc
hansu: Maybe something like this:

This section describes an option to have multiple load-commands for multiple instances of the same component at different positions in the file or among different files.
Internally, this requires to read the HAL file twice, hence the name TWOPASS.
Supported since LinuxCNC version 2.5, the TWOPASS processing of LinuxCNC configuration files helps with their modularization and readability.
To recall, LinuxCNC configuration files are specified in a LinuxCNC INI file as `[HAL]HALFILE=filename`.

hansu: @smoe Have you noticed this?

smoe:
@hansu - adopted

smoe:
While reading through that paragraph, I still have not understood if it is ok to use a component "a" in anotherfile.hal if that component was only loaded in a.hal. Also, it is not clear, if two files (filea.hal and fileb.hal) can both load a component and name that "a", and if you need twopass for that.

hansu:
Without TWOPASS you can use a component in several HAL files, but the component has to be loaded in the first HAL file.
As that is a bit inconvenient, TWOPASS was introduced. So you can load "component a" in a.HAL and the load the same component in b.HAL (but with a different instance name).

> Also, it is not clear, if two files (filea.hal and fileb.hal) can both load a component and name that "a", and if you need twopass for that.

Yes with TWOPASS, but not with the same name (how should HAL distinguish between them?)

smoe:
> > Also, it is not clear, if two files (filea.hal and fileb.hal) can both load a component and name that "a", and if you need twopass for that.
> 
> Yes with TWOPASS, but not with the same name (how should HAL distinguish between them?)

You could just open a new namespace with every file you open, i.e. you mask the prior declaration.

Any such "corollaries" I would explicitly list, together with a smallish explanation why these are consequences from the prior described design. I am just not the right person to do that - and would also like to have this peer reviewed by some more people.

I'll just go and fix the conflict, will also add a "FIXME" if you don't mind, asking for further clarifications.

hansu:
>You could just open a new namespace with every file you open, i.e. you mask the prior declaration.

And then have pin names like `<namespace>.<original_pin name>`? Would be more confusing I think ...

> I am just not the right person to do that - and would also like to have this peer reviewed by some more people.

👍

smoe:
Resolved conflicts, squashed a few commits that looked redundant in their description.

hansu:
I just checked the changes and I wonder if something went wrong when force-pushed to resolve the conflicts?
Because some changes I thought you already had adressed? Those I marked again as unresolved.

smoe:
> I just checked the changes and I wonder if something went wrong when force-pushed to resolve the conflicts? Because some changes I thought you already had adressed? Those I marked again as unresolved.

Either that or I have once again locally "fixed" those undoings while performing other changes. To avoid hickups, I just checked that there are now conflicts have just accepted everything via the website.

hansu:
Thanks

PR #2323, opened 2023-02-04
gmoccapy -add log levels to gmoccapy and getiniinfo

### LINKED ISSUE #2273: Configuration warnings by sim.gmoccapy
Hello, Just to make a good impression on new users (and convey trust for those transitioning to the new release) we should avoid warnings on stdout/stderr. I came across

```
**** GMOCCAPY INFO ****
**** Entering make macro button
found 15 Macros
Macro go_to_position x-pos y-pos z-pos has image link
Image = /home/moeller/linuxcnc/configs/sim.gmoccapy/macros/images/goto_x_y_z.png
get button with image
Macro i_am_lost has image link
Image = /home/moeller/linuxcnc/configs/sim.gmoccapy/macros/images/i_am_lost.png
get button with image
Macro macro_8 has image link
Image = /home/moeller/linuxcnc/configs/sim.gmoccapy/macros/images/macro_8.png
get button with image
**** GMOCCAPY INFO ****
```
Why is this shown? Maybe a debug level wrongly set? The files are existing:
```
-rw-r--r-- 1 moeller moeller  24520 26. Dez 18:38 /home/moeller/linuxcnc/configs/sim.gmoccapy/macros/images/goto_x_y_z.png
-rw-r--r-- 1 moeller moeller  14074 26. Dez 18:38 /home/moeller/linuxcnc/configs/sim.gmoccapy/macros/images/i_am_lost.png
-rw-r--r-- 1 moeller moeller 222151 26. Dez 18:38 /home/moeller/linuxcnc/configs/sim.gmoccapy/macros/images/macro_8.png
```
I could open these files in gimp (macro_8 is very cute). So, maybe just not print this?

```
**** GMOCCAPY INFO ****
**** Place in table
/usr/bin/gmoccapy:1565: DeprecationWarning: Gtk.Table.resize is deprecated
  self.widgets.tbl_DRO.resize(rows, cols)
**** GMOCCAPY INFO ****
```
https://docs.gtk.org/gtk3/method.Table.resize.html mentions that this is deprecated since 3.4 and GtkGrid would resize automatically. Such dynamics may be interesting - after all, reactive web pages are a thing, too. No idea how will this works for us, though.

```
**** GMOCCAPY INFO ****
**** Gremlin Einrichtung gestartet ****
Found an error!

  File "/usr/lib/python3/dist-packages/gladevcp/hal_gremlin.py", line 164, in do_set_property
    if self.lathe_option:
       ^^^^^^^^^^^^^^^^^
AttributeError: 'HAL_Gremlin' object has no attribute 'lathe_option'
```
The default "sim.gmoccapy" is a 3-axis mill and that is why that lathe_option is not present?!?
Directly underneath it states the following, which may be related (even though it does not look like it):
```
not enough values to unpack (expected 2, got 1)
'btn_sel_next'
'btn_sel_prev'
'btn_select'
'btn_home'
'btn_dir_up'
'btn_jump_to'
not enough values to unpack (expected 2, got 1)
not enough values to unpack (expected 2, got 1)
'IconFileSelection' object has no attribute 'on_btn_jump_to'
not enough values to unpack (expected 2, got 1)
not enough values to unpack (expected 2, got 1)

(gmoccapy:101189): Gtk-CRITICAL **: 22:43:43.955: gtk_tree_view_scroll_to_cell: assertion 'tree_view->priv->tree != NULL' failed

(gmoccapy:101189): Gtk-CRITICAL **: 22:43:44.047: gtk_entry_set_text: assertion 'text != NULL' failed
Attribute error <GParamString 'icon-theme-path'> and <class 'str'> sans 10
Attribute error <GParamString 'icon-theme-name'> and <class 'str'> sans 10
0  =  /usr/bin/gmoccapy
1  =  -ini
2  =  /home/moeller/linuxcnc/configs/sim.gmoccapy/gmoccapy.ini
Entry = trivkins
Entry = coordinates=xyz
found the following coordinates xyz
```
Harmless:
```
**** virtual keyboard program found : <onboard>
14:09:30.842 WARNING Config: mousetweaks GSettings schema not found, mousetweaks integration disabled.
/usr/bin/gmoccapy:3318: DeprecationWarning: Gtk.FontButton.get_font_name is deprecated
  self.notification.set_property('font', self.widgets.fontbutton_popup.get_font_name())
/usr/bin/gmoccapy:381: DeprecationWarning: Gtk.FontButton.set_font_name is deprecated
  self.widgets.fontbutton_popup.set_font_name(self.prefs.getpref("message_font", "sans 10", str))
/usr/bin/gmoccapy:476: DeprecationWarning: Gtk.ColorButton.set_color is deprecated
  self.widgets.abs_colorbutton.set_color(Gdk.color_parse(self.abs_color))
```
which has been deprecated with version 3.4 (according to https://athenajc.gitbooks.io/python-gtk-3-api/content/gtk-group/gtkcolorbutton.html) and we are requesting 3.11+ as a dependency.

Similarly:
```
**** GMOCCAPY INFO ****
**** Entering make touch button
/usr/bin/gmoccapy:972: DeprecationWarning: Gtk.Widget.override_background_color is deprecated
  btn.override_background_color(Gtk.StateFlags.ACTIVE, Gdk.RGBA(1.0, 1.0, 0.0, 1.0))
**** GMOCCAPY INFO ****
```
Maybe just give default settings?
```
**** GMOCCAPY GETINIINFO **** 
[KINS] KINESTYPE is trivkins
**** GMOCCAPY GETINIINFO **** 
No MIN_ANGULAR_VELOCITY entry found in [DISPLAY] of INI file
Using default value of 0.1 degree / min
**** GMOCCAPY GETINIINFO **** 
No DEFAULT_ANGULAR_VELOCITY entry found in [DISPLAY] of INI file
Using default value of 360 degree / min
**** GMOCCAPY GETINIINFO **** 
No MAX_RAPID_OVERRIDE entry found in [DISPLAY] of INI file 
 Default settings 100 % applied!
**** GMOCCAPY GETINIINFO **** 
```

c-morley:
This will allow the user to control the many debug messages sent to the terminal. 
It uses qtvcp's logging library so no reinventing the wheel. 
Other files should probably be converted to use logging too. 

Resolves #2273

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -40,7 +40,6 @@ import hal                 # base hal class to react to hal signals
 import hal_glib            # needed to make our own hal pins
 import sys                 # handle system calls
 import os                  # needed to get the paths and directories
-import gladevcp.makepins   # needed for the dialog"s calculator widget
 import atexit              # needed to register child's to be closed on closing the GUI
 import subprocess          # to launch onboard and other processes
 import tempfile            # needed only if the user click new in edit mode to open a new empty file
@@ -48,13 +47,9 @@ import linuxcnc            # to get our own error system
 import locale              # for setting the language of the GUI
 import gettext             # to extract the strings to be translated
 from collections import OrderedDict # needed for proper jog button arrangement
-
-from gladevcp.gladebuilder import GladeBuilder
-
-from gladevcp.combi_dro import Combi_DRO  # we will need it to make the DRO
-
 from time import strftime  # needed for the clock in the GUI
 #from Gtk._Gtk import main_quit
+
 
 # Throws up a dialog with debug info when an error is encountered
 def excepthook(exc_type, exc_obj, exc_tb):
@@ -66,7 +61,7 @@ def excepthook(exc_type, exc_obj, exc_tb):
         w = None
     lines = traceback.format_exception(exc_type, exc_obj, exc_tb)
     message ="Found an error!\nThe following information may be useful in troubleshooting:\n\n" + "".join(lines)
-    print(message)
+    LOG.error(message)
     m = Gtk.MessageDialog(parent = w,
                           modal = True ,
                           destroy_with_parent = True,
@@ -110,13 +105,6 @@ BASE = os.path.abspath(os.path.join(os.path.dirname(sys.argv[0]), ".."))
 LIBDIR = os.path.join(BASE, "lib", "python")
 sys.path.insert(0, LIBDIR)
 
-# as now we know the libdir path we can import our own modules
-from gmoccapy import widgets       # a class to handle the widgets
-from gmoccapy import notification  # this is the module we use for our error handling
-from gmoccapy import preferences   # this handles the preferences
-from gmoccapy import getiniinfo    # this handles the INI File reading so checking is done in that module
-from gmoccapy import dialogs       # this takes the code of all our dialogs
-from gmoccapy import icon_theme_helper
 
 _AUDIO_AVAILABLE = False
 try:
@@ -258,25 +246,22 @@ class gmoccapy(object):
         self.user_mode = False
         self.logofile = None
         for index, arg in enumerate(argv):
-            print(index, " = ", arg)
+            LOG.debug(index, " = ", arg)
             if arg == "-user_mode":
                 self.user_mode = True
                 self.widgets.tbtn_setup.set_sensitive(False)
-                message = _("**** GMOCCAPY INI Entry ****")
-                message += "\n" + _("user mode selected")
-                print (message)
+                message = "\n" + _("user mode selected")
+                LOG.debug(message)
             if arg == "-logo":
                 self.logofile = str(argv[ index + 1 ])
-                message = _("**** GMOCCAPY INI Entry ****")
-                message += "\n" + _("logo entry found = {0}").format(self.logofile)
-                print (message)
+                message = "\n" + _("logo entry found = {0}").format(self.logofile)
+                LOG.debug(message)
                 self.logofile = self.logofile.strip("\"\'")
                 if not os.path.isfile(self.logofile):
                     self.logofile = None
-                    message = _("**** GMOCCAPY INI Entry Error ****")
-                    message += "\n" + _("Logofile entry found, but could not be converted to path.")
+                    message = "\n" + _("Logofile entry found, but could not be converted to path.")
                     message += "\n" + _("The file path should not contain any spaces")
-                    print(message)
+                    LOG.warning(message)
 
         # check if the user want a Logo (given as command line argument)
         if self.logofile:
@@ -466,7 +451,7 @@ class gmoccapy(object):
         LANGDIR = os.path.join(BASE, "share", "Gtksourceview-2.0", "language-specs")
         file_path = os.path.join(LANGDIR, "gcode.lang")
         if os.path.isfile(file_path):
-            print("**** GMOCCAPY INFO: Gcode.lang found ****")
+            LOG.info("GMOCCAPY INFO: Gcode.lang found")
             self.widgets.gcode_view.set_language("gcode", LANGDIR)
 
         # set the user colors and digits of the DRO
@@ -524,7 +509,7 @@ class gmoccapy(object):
                 exec(compile(open(rcfile, "rb").read(), rcfile, 'exec'))
             except:
                 tb = traceback.format_exc()
-                print(tb, file=sys.stderr)
+                LOG.error(tb, file=sys.stderr)
                 self.notification.add_message(_("Error in ") + rcfile + "\n" \
                     + _("Please check the console output."), ALERT_ICON)
 
@@ -545,7 +530,7 @@ class gmoccapy(object):
                 style_context.add_provider_for_screen(screen, provider_custom, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION)
             except:
                 tb = traceback.format_exc()
-                print(tb, file=sys.stderr)
+                LOG.error(tb, file=sys.stderr)
                 self.notification.add_message(_("Error in ") + css_file + "\n" \
                     + _("Please check the console output."), ALERT_ICON)
 
@@ -642,9 +627,8 @@ class gmoccapy(object):
 ###############################################################################
 
     def _make_DRO(self):
-        print("**** GMOCCAPY INFO ****")
-        print("**** Entering make_DRO")
-        print("axis_list = {0}".format(self.axis_list))
+        LOG.debug("Entering make_DRO")
+        LOG.debug("axis_list = {0}".format(self.axis_list))
 
         # we build one DRO for each axis
         self.dro_dic = {}
@@ -681,8 +665,7 @@ class gmoccapy(object):
         return list(self.joint_axis_dic.keys())[list(self.joint_axis_dic.values()).index(value)]
 
     def _make_ref_axis_button(self):
-        print("**** GMOCCAPY INFO ****")
-        print("**** Entering make ref axis button")
+        LOG.debug("Entering make ref axis button")
 
         # check if we need axis or joint homing button
         if self.trivial_kinematics:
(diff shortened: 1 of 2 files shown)


hansu:
Really great Chris, thank you! :+1: :+1: :+1: 
I also had in mind to use qtvcp's logger for gmoccapy, but I just didn't find the time to implement it.
Can you target the PR to this branch: https://github.com/LinuxCNC/linuxcnc/tree/gmoccapy-3-4-2 ?
Then I can add info to the change log and docs and merge to 2.9.

All those removed imports - those are really not necessary?

src/emc/usr_intf/gmoccapy/gmoccapy.py
```diff
@@ -2067,7 +2029,7 @@ def _init_icon_themes(self):
                 selected_index = [icon_theme[1] for icon_theme in valid_icon_themes].index(icon_theme_preference)
                 icon_theme_choice.set_active(selected_index)
             except ValueError:
-                print(f"Warning: preferred icon-theme '{icon_theme_preference}' not found; switching to 'default'.")
+                LOG.warning(f"Warning: preferred icon-theme '{icon_theme_preference}' not found; switching to 'default'.")
```
hansu, line 2032:
```suggestion
                LOG.warning(f"Preferred icon-theme '{icon_theme_preference}' not found; switching to 'default'.")
```

src/emc/usr_intf/gmoccapy/gmoccapy.py
```diff
@@ -5783,12 +5735,58 @@ def _make_hal_pins(self):
...
+    from qtvcp import logger
+    LOG = logger.initBaseLogger('GMoccapy', log_file=None, log_level=logger.WARNING)
+
+    # we set the log level early so the imported modules get the right level
+    if '-d' in sys.argv:
+        # Log level defaults to INFO, so set lower if in debug mode
+        logger.setGlobalLevel(logger.DEBUG)
+        LOG.debug('DEBUGGING logging on')
+    elif '-i' in sys.argv:
+        # Log level defaults to INFO, so set lower if in debug mode
```
hansu, line 5758:
```suggestion
        # Log level defaults to WARNING, so set lower if in debug mode
```
defaults to WARNING if I get it right?

What about a flag for CRITICAL?

hansu (commented):
Great job in assigning the log levels! I agree in almost most of the cases, however, I have still some suggestions.

And the correct order of log levels is DEBUG, INFO, VERBOSE, WARNING, ERROR, CRITICAL, is that right?

And they can be only set like this, right?
```
[DISPLAY]
DISPLAY = gmoccapy -q
```

Maybe it would be good to have a LOG_LEVEL in the DISPLAY section as well.

c-morley:
> Great job in assigning the log levels! I agree in almost most of the cases, however, I have still some suggestions.
> 
> And the correct order of log levels is DEBUG, INFO, VERBOSE, WARNING, ERROR, CRITICAL, is that right?
> 
> And they can be only set like this, right?
> 
> ```
> [DISPLAY]
> DISPLAY = gmoccapy -q
> ```
> 
> Maybe it would be good to have a LOG_LEVEL in the DISPLAY section as well.

verbose, debug, info, warning, error, critical
unfortunately at the moment verbose works but does not say 'verbose' it says 'debug' and is the same color as debug.

LOG_LEVEL is more difficult because the log levels must be set before loading some libraries and before instantiating gmoccapy . Not impossible -just have to add code to parse the INI.
 In the current suggestion it is consistent with qtvcp which is nice too.

i don;t know how to change the base branch - could you instruct me?

src/emc/usr_intf/gmoccapy/gmoccapy.py
c-morley: yes i se the comments are wrong - copied from qtvcp and i changed the default level recently.
I'll need to fix them on qtvcp too. thanks

my opinion:
critical - in my mind if you have critical errors you should pretty much be logging then shutting down.
Having a critical log setting would block errors from showing. can't think of any reason you would want to hide errors.

src/emc/usr_intf/gmoccapy/gmoccapy.py
```diff
@@ -5783,12 +5735,58 @@ def _make_hal_pins(self):
...
+    from qtvcp import logger
+    LOG = logger.initBaseLogger('Gmoccapy', log_file=None, log_level=logger.WARNING)
+
+    # we set the log level early so the imported modules get the right level
+    if '-d' in sys.argv:
+        # Log level defaults to WARNING, so set lower if in debug mode
+        logger.setGlobalLevel(logger.DEBUG)
+        LOG.debug('DEBUGGING logging on')
+    elif '-i' in sys.argv:
+        # Log level defaults to WARNING, so set lower if in debug mode
```
hansu, line 5758:
```suggestion
        # Log level defaults to WARNING, so set lower if in info mode
```

hansu:
> LOG_LEVEL is more difficult because the log levels must be set before loading some libraries and before instantiating gmoccapy . Not impossible -just have to add code to parse the INI. In the current suggestion it is consistent with qtvcp which is nice too.

Yeah but it's fine so. No problem if documented properly.
 
> i don;t know how to change the base branch - could you instruct me?

You'll find the edit button for that at the top of this page:
grafik

c-morley:
Base has been changed, thanks !

c-morley:
I could squash together if you would prefer.

hansu:
> I could squash together if you would prefer.

Yeah would be nice. Have you any thoughts on my other comments?

c-morley:
I thin I have answered all comments - if I have missed one please ask me again or point them out.
If you are referring to some code changes, say changing from error to warning or such - I have no strong opinion - I just set them to what seemed reasonable in a 10 second look.
Feel free to change them to what ever you feel is best.
I'll squish the branch in a couple minutes.

c-morley:
That is the integer representation of the global  log level.
It is in qtvcp/logger.py
I think we should remove it. Opinion?

hansu:
Then it feels a bit redundant, because it says first "INFO logging on" and then this. (Didn't remember if what it prints for default - WARNING).
And using "Base" confused me a bit as well as switching to an integer representation.
So yes can be removed.


Another thing came into my mind while I had a look at the multi-line log messages. It would be nice if the output can be look like this:
```
[Gmoccapy][INFO] A long long long long long long long
                 long long message that goes over several lines.
```
Currently, it's only possible either so
```
[Gmoccapy][INFO] A long long long long long long long
long long message that goes over several lines.
```
or so
```
[Gmoccapy][INFO] A long long long long long long long
[Gmoccapy][INFO] long long message that goes over several lines.
```
First would be no problem if there were no other messages. For the second it's hard so see how many errors/warning there are actually.
(I mean yes you can add the correct number of spaces - the logging domain woul probably not change so that would work, but inconvenient and a bit ugly 🙃)
I thought about something like this, so a function "newline" that adds as many spaces as the last prefix needed:
```
LOG.info(" A long long long long long long long")
LOG.newline("long long message that goes over several lines.")
```

Dou you think that is a reasonable feature?

c-morley:
for LOG.newline()to be useful one would need LOG.info_newline() plus all the other log levels.
I am still having trouble with getting extra log levels to work properly.
you could use:
LOG.info("A long long long long long long long line\n        and the rest of the long line line.")

c-morley:
should be 8 spaces between \n and the 'and'

hansu:
>for LOG.newline()to be useful one would need LOG.info_newline() plus all the other log levels.

I thought about a static class variable that "remembers" the last used log level.

>  you could use:
> LOG.info("A long long long long long long long line\n and the rest of the long line line.")
> should be 8 spaces between \n and the 'and'

Yeah that's what I was meaning in the sentence inside the parantheses.

hansu:
>for LOG.newline()to be useful one would need LOG.info_newline() plus all the other log levels.

I thought about a static class variable that "remembers" the last used log level.

>  you could use:
> LOG.info("A long long long long long long long line\n and the rest of the long line line.")
> should be 8 spaces between \n and the 'and'

Yeah that's what I was meaning in the sentence inside the parantheses.

hansu:
Let's skip this. This seems to be too much work for a bit improvement in readability. Because in the log file there has to be added number of spaces...

PR #2328, opened 2023-02-08
work around changed behavior in glBitmap

SebKuzminsky:
Until "recently" (maybe some time late 2022), this code used to work. Now it fails on some platforms.

¯\_(ツ)_/¯

Jeff suggested passing in a 1x1-pixel transparent bitmap instead of a 0x0-pixel Null bitmap, and this seems to work.


SebKuzminsky:
This lets Axis start on my Bookworm/Wayland machine that used to reliably reproduce the OpenGL "out of memory" error with `glBitmap`.

jepler (approved):
It feels like we're working around somebody else's bug, but if this fixes it, let's ship it

PR #2333, opened 2023-02-10
iconview: add logging + improve error handling

hansu:
This adds logging to iconview.py and fixes some error messages.

### DIFF
--- a/lib/python/gladevcp/iconview.py
+++ b/lib/python/gladevcp/iconview.py
@@ -39,6 +39,13 @@ from gi.repository.GdkPixbuf import Pixbuf
 import os
 import mimetypes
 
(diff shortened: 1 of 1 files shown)


c-morley:
I just recenly fpund out that exceptions such as:
except Exception as e:

are best logged something like this:
LOG.exception(e)

That get printed at the error log level, but with better exception information

hansu:
Yes I like it. Gives more information and is nice short to write. Already updated ;-)

c-morley:
Yes I will use in the future too. I got verbose logging working and in the process saw this in the docs.
I think this logging work in gmoccapy you are doing is a real step of goodness!

Issue #2277, opened 2023-01-16
gladevcp XML/PNG files are not in debs


c-morley:
User on the forum is having problems getting glade to work on an instaled version.
Seems the requires hal_python.xml  plus all the widget image files are not installed on the system.

/usr/lib/python3/dist-packages/gladevcp  is missing the .xml and .png files.

These are found in the source at:
lib/python/gladevcp

affects 2.9 and master for sure. Probably older too.

c-morley:
I should say that installing the XML file and setting the envromental variable allowed me to load widgets.
It's been awhile but maybe the XML can go somewhere else - rings a bell....

c-morley:
ok here we go.
the XML and images are in:
lib/share/glade3
and it seems they now need to be in
 usr/lib/glade

Haven't got the images to work yet...

c-morley:
You can also set the catalog from the Preference dialog - should add that to docs.

rodw-au:
What version of Debian?
What version of Linuxcnc?
URL to forum post?

It got harder now with 3 versions of linuxcnc and 3 versions of Debian in play.

c-morley:
1) no idea - I was using Mint-My guess is the source is wrong.
2) 2.9 and master (I was using 2.9)
3) The forum page led me to the issue - the fellow in the end was using RIP and forgot to run the init script.

blazini36:
> ok here we go. the XML and images are in: lib/share/glade3 and it seems they now need to be in usr/lib/glade
> 
> Haven't got the images to work yet...

Images are an issue here too. I've gotten gladeVCP to work in a couple of different setups and both spit out this
`HdyCarousel	needs an icon named 'widget-libhandy-carousel'`
for every widget

smoe:
I just started gmoccpy on Debian arm64 (February 8, 2003, official package) and run in that problem, too:
./lib/python/gladevcp/pi.png
is not installed, expected as
/usr/lib/python3/dist-packages/gladevcp/pi.png

The linuxcnc.install.in instructions seem fine (they just install everything in usr/lib/python3), have not tried it, but
```
--- a/src/Makefile
+++ b/src/Makefile
@@ -800,6 +800,7 @@ install-python: install-dirs
        $(FILE) ../lib/python/qtvcp/plugins/*.py $(DESTDIR)$(SITEPY)/qtvcp/plugins
[...]

PR #2334, opened 2023-02-11
Also installing PNGs for gladevcp

smoe:
Also adopted brace expansion to cover multiple file types in other lines. This addresses the PNGs problem in https://github.com/LinuxCNC/linuxcnc/issues/2277 - I just did not know where to send the XML files to - also within Python?

Whoever merges this please add another line for the xml files with a different directory or add ",xml" to the list in the braces, please.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -18,6 +18,7 @@
 #     install-docs: Install all enabled documentation into $(DESTDIR).
 #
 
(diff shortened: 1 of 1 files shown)


hansu:
This looks good on first sight, but let me test it when I am back from vacation.

hansu:
Thanks!

PR #2336, opened 2023-02-11
gmoccapy/notification: remove attribute errors

hansu:
The problem here was that `Gtk.Window.__init__() `sets all `__gproperties__` to their default values.
But the properties `icon_theme_path` and `icon_theme_name` belongs to a Gtk.IconTheme object that is created after that:

```python
class Notification(Gtk.Window):
    ...

    def __init__(self):
        Gtk.Window.__init__(self)
        ...
        self.icon_theme = Gtk.IconTheme()
```
So I just moved   `Gtk.Window.__init__(self)` down. Not sure if this is the best solution.
Because I don't know if those properties gets always initialized  automatically when the Window object is instantiated.

Maybe @PeterMue has an idea?

### DIFF
--- a/src/emc/usr_intf/gmoccapy/notification.py
+++ b/src/emc/usr_intf/gmoccapy/notification.py
@@ -63,9 +63,9 @@ class Notification(Gtk.Window):
            'use_frames' : (GObject.TYPE_BOOLEAN, 'Use Frames for messages', 'You can separate the messages using frames, but you will need more space',
                     True, GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT),
             'icon_theme_path' : (GObject.TYPE_STRING, 'Icon theme lookup path', 'Pathes where to look for icon themes',
-                      "sans 10", GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT),
+                      "", GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT),
             'icon_theme_name' : (GObject.TYPE_STRING, 'Icon theme name', 'Name set in gmoccapy preferences',
-                      "sans 10", GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT),
+                      "classic", GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT),
                       }
     __gproperties = __gproperties__
 
@@ -76,8 +76,6 @@ class Notification(Gtk.Window):
 
     # build the main gui
     def __init__(self):
-        Gtk.Window.__init__(self)
-        self.connect('destroy', lambda*w:Gtk.main_quit())
         self.messages = []
         self.popup = Gtk.Window(type = Gtk.WindowType.POPUP)
         self.vbox = Gtk.VBox()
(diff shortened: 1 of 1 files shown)


hansu:
@c-morley Any thoughts on this?

c-morley:
The error comes from Gtk.IconTheme setting properties that Gtk.window already set to default?
So you invoke gtk.window after gtk.Icontheme so window over writes Icon_theme to default?
Do I have that right?
You can set gproperties directly if you want with __gproperties__ ={} , but I'm not sure if that is the problem you are trying to work around.

No opinion so far other then, hey if it works.... 'don't let the pursuit of perfection get in the way of good enough'

though instead of
Gtk.Window.__init__(self)

isn't
super().__init__(self)
preferred ?

Though that shouldn't change behavior.

hansu:
No the problem is that `self.icon_theme` is set to a default value in  `Gtk.Window.__init__(self)` but at this point it doesn't know yet that it is a Gtk.IconTheme() because that is assigned later.

>No opinion so far other then, hey if it works.... 'don't let the pursuit of perfection get in the way of good enough'

Alright :smile: 

>isn't
super().init(self)
preferred ?

That sounds like a good style, I will change that :+1: .

PeterMue:
I think your solution to move the super call after the property initialization is perfectly fine.
I was first thinking of the `__init__` method like a custructor call in other languages such as c++ or java but python treats it more like a generic init method that _may_ be called after object creation. So moving the call to the super init method at the end of (or wherever it makes sense) the child's init method seems to be a absolutely valid solution in my eyes.

Thinking of an alternative we might could do a lazy initialization inside the `do_set_propety` method like first do a falsy-check (or whatever the "wrong" default value is) and initialize the icon_theme object if necessary first. But I can't see any benefit of this approach in this case.

Agree with @c-morley calling the super initializer without type info is more concise `super().__init__(self)`

hansu:
Thanks :slightly_smiling_face:

hansu:
Changing this to `super().__init__(self)` gives following error:

```
  File "/home/cnc/linuxcnc/linuxcnc-2.9/bin/gmoccapy", line 5771, in <module>
    app = gmoccapy(sys.argv)
          ^^^^^^^^^^^^^^^^^^
  File "/home/cnc/linuxcnc/linuxcnc-2.9/bin/gmoccapy", line 205, in __init__
    self.notification = notification.Notification()  # Our own message system
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/cnc/linuxcnc/linuxcnc-2.9/lib/python/gmoccapy/notification.py", line 98, in __init__
    super().__init__(self)
  File "/usr/lib/python3/dist-packages/gi/overrides/Gtk.py", line 522, in __init__
    _window_init(self, *args, **kwargs)
  File "/usr/lib/python3/dist-packages/gi/overrides/__init__.py", line 319, in new_init
    return super_init_func(self, **new_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: could not convert value for property `type' from Notification to GtkWindowType
 (gmoccapy:62)
```

Any ideas?

c-morley:
Both of these seem to work for me:                

super(Notification, self).__init__()

super().__init__()

c-morley:
They are unusual signature - not using self as an argument.
Seems weird.

hansu:
Yeah with `super().__init__()` withouf self as argument it works :).

Feels only a bit weird that  `Gtk.Window.__init__(self)` is replaced by `super().__init__()` .

PR #2337, opened 2023-02-11
Merge 2.9 into master

hansu:
I updated the pot files in 2.9 and I am thinking that was not a good idea, because is does not really have an effect and only 
creates merge conflicts.

I think it is best to revert that. Any objections? 
@petterreinholdtsen ?

### DIFF
--- a/src/po/gmoccapy/gmoccapy.pot
+++ b/src/po/gmoccapy/gmoccapy.pot
@@ -7,8 +7,8 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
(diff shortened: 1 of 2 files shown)


petterreinholdtsen:
[Hans Unzner]
> I updated the pot files in 2.9 and I am thinking that was not a good
> idea, because is does not really have an effect and only creates merge
> conflicts.

That is correct.

> I think it is best to revert that. Any objections?
> @petterreinholdtsen ?

It might cause merge conflicts anyway with the merge forward approach
used, but probably best to revert anyway.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
Okay, reverted.

Issue #1486, opened 2021-12-31
qtplasma does not work on Wayland


dwrobel:
`qtplasma` does not seem to work on the latest Gnome 41.2 / Wayland 1.19 environment available on Fedora 35 _(tested on https://github.com/LinuxCNC/linuxcnc/commit/14a02d19b05caf06e6fccd0865df81440760bbac)_. Note, that `axis` works fine on the same environment.

Here is the screenshot on Wayland after an attempt to load `metric_circles.ngs` file:

![Screenshot from 2021-12-31 10-46-43](https://user-images.githubusercontent.com/228873/147816501-63bcab2e-0716-42df-8f86-722c6168b8f2.png)

Please note that the `Select` button has been pressed, but the screen hadn't been refreshed _(there is still a file selector dialog visible)_. Shortly after that `Gnome` starts to complain that `qtplasma` is not responding:

![Screenshot from 2021-12-31 10-46-49](https://user-images.githubusercontent.com/228873/147816590-71c7f6b2-a58a-4ce1-8816-6a248d320e8b.png)

For the reference the same `metric_circles.ngs`  loaded correctly on the `X11` session:
![Screenshot from 2021-12-31 10-44-02](https://user-images.githubusercontent.com/228873/147816657-d6c83231-9c6d-48a3-8829-2f0ff337aae3.png)

dwrobel:
One more screenshot from Wayland, taken directly after launching `qtplasma` without pressing any buttons:
![Screenshot from 2021-12-31 11-22-00](https://user-images.githubusercontent.com/228873/147818231-c13b81b4-cc26-4270-923b-d9144f342dc2.png)

Please note that the white rectangle part of the screen does not seem to be refreshed at all and it looks that in `qtplasma` this part is completely transparent as the text _dw@t630:~/Pictures_ comes from the Gnome terminal title bar window which was beneath `qtplasma` when I launched it.

c-morley:
Could you try running gcode_graphics directly in a terminal:  python3 gcode_graphics.py
Also interesting if any gladevcp screen grahics display works with wayland.
They use alot of common code.

dwrobel:
>Could you try running gcode_graphics directly in a terminal: python3 gcode_graphics.py

Please find attached a screen recording from running it on a `Wayland` session:
```
[dw@t630 linuxcnc]$ loginctl 
SESSION  UID USER SEAT  TTY 
     16 1000 dw   seat0 tty2

[...]

PR #2340, opened 2023-02-11
qtvcp: workaround for transparent preview on wayland

phillc54:
Tested on bookworm amd64 and arm64.

No ill effects noticed on amd64 buster and bullseye.

Issue reported [here](https://github.com/LinuxCNC/linuxcnc/issues/1486)

### DIFF
--- a/src/emc/usr_intf/qtvcp/qtvcp.py
+++ b/src/emc/usr_intf/qtvcp/qtvcp.py
@@ -8,6 +8,10 @@ import hal
 import signal
 import subprocess
 
+# workaround for transparent preview on Wayland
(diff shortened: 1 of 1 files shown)


phillc54:
On second thoughts it is probably better to add this to the linuxcnc script so we end up with:

```
    export PYOPENGL_PLATFORM="x11"
    export GDK_BACKEND="x11"
    export QT_QPA_PLATFORM="xcb"

```

SebKuzminsky:
Yes, please put this in `scripts/linuxcnc.in`, in the same place as the setting of `PYOPENGL_PLATFORM`.

PR #2348, opened 2023-02-16
Ohmic and Spotting Fixes

goglie:
Miscellaneous fixes:
-Fixed behavior issues on ohmic detection. The ohmic contact was triggered inadvertently (poorly formulated and faulty equation.
-Fixed a "Spotting" issue. If the source did not have time to do the pointing or was in default, the process continued as if nothing had happened instead of retrying.


phillc54:
Could you tidy this up by splitting it into two separate PRs with single commits please. There is also some trailing whitespace the should be removed.

hansu:
And commit messages in English would be nice.

petterreinholdtsen:
If you need help navigating git to adjust this pull request, I would be
happy to assist on IRC, #linuxcnc-devel on libera.chat

-- 
Happy hacking
Petter Reinholdtsen

snowgoer540:
I don't know what specific "problems" you are trying to solve so I am just going from what I can surmise from the code / commit comments (via google translate), I have the following comments (broken down by file):

**plasmac.comp** - your changes make the spotting time / spotting threshold check start _after_ an arc ok is received.  This essentially will turn spotting into piercing and breaks the current functionality.  I am confused if you are intending to pierce the center of a hole, or use the spotting feature.

Judging from your commit comment, it seems you are trying to use the spotting feature, but the machine is moving on without attempting to spot at the location (meaning it _appears_ to skip the M3).  This can happen when the **Spotting Threshold** is set to 0v (on the PARAMETERS tab).  I recently ran into the same thing.  Raising it to even 1v causes the machine to hold the M3 command long enough for the plasma power source to respond.  Then the torch fires, the Spotting timer starts only after the voltage rises past the threshold.  I did change the default Spotting Threshold to 1v in commit 22249d4 on Jan 21, 2023.  However any config made prior to this change would still be set to the previous 0v default.

I advise to revert your code changes and try changing your threshold above 0v and see if that helps.  It takes a bit of trial and error to get a nice spot.  For me on a Hypertherm this is 5v, 65mS.  

**qtplasmac_gcode.py** - the end result of this seems to add a space at the end of the file.  I'm not sure that was the intent.  Assuming the intent was to add the start of processing for a spotting_min_diameter variable from the post processor, then my initial reaction is that if the post processor is producing a spotting_min_diameter variable, then the post processor should just be coded to export the spots in a manner that conforms with the QtPlasmaC user guide.  Otherwise, I need more information.

**ohmic.comp** - this commit should be split out (as Phill said).  While I agree that this component has issues, I think that adding a 100mS "timer" to the component breaks the spirit of THCAD ohmic sensing.  I am not sure if there are people who use this component or not (I have to imagine there are), but I would suggest you write a custom component for your own use vs making this specific change.

snowgoer540:
Closing this pull request due to all of the reasons in the previous comment.

If you wish to make changes and discuss, you can do so and re-open the pull request and we will re-evaluate at that time.

PR #2350, opened 2023-02-18
    gmoccapy -audio_player: switch to python3 gstreamer

phillc54:


hansu:
I tried a quick build, but that failed at Gst.init(). But on Bullseye. I can try in some days with Bookworm.

phillc54:
Hmm, I did do this work on bullseye and I haven't tried bookworm.
I also did some similar stuff today for Gscreen and it worked ok as well.
I will do some more testing tomorrow.

hansu:
Do I need to install some package?

c-morley:
Gmaoccapy supports logginf now.
I suggest adding error logging to the import - similar to what qtvcp does.
If the import fails print a suggestion of what package to get.

phillc54:
@hansu 
I tried it on buster and got an error with the Gst.init() line so I changed it ti Gst.init(None) and it fixed for both amd64 and arm64, I have pushed the change.
It worked as it was on my bullseye and bookworm amd64 and arm64 VM's and this change appers to show no issues on them.

@c-morley 
That seems a good idea but I am not sure about the gi.repository stuff, is that all included normally?

c-morley:
From what I can see:
-There is no reason to import gst in gmoccapp.py - it's importedin the player.py file
-if you move:
`_AUDIO_AVAILABLE = False
try:
    from .gmoccapy import player        # a class to handle sounds
    _AUDIO_AVAILABLE = True
except:
    pass
`
to where the rest of 'from gmoccapy inport...' are.
then add this to player.py:
`from qtvcp import logger
# Force the log level for this module
# LOG.setLevel(logger.INFO) # One of DEBUG, INFO, WARNING, ERROR, CRITICAL
`

and finally to player.py
`try:
    gi.require_version('Gst', '1.0')
    from gi.repository import Gst
except:
    LOG.warning('Gst module missing, - is package python3-gst installed?')`


or something close to along those lines

c-morley:
Wow this code editor is as bad as linuxcnc forum... check the raw text for the code I quoted.

c-morley:
also not sure why the period in
```
from .gmoccapy import player 
```
but you don't need it in my tests

phillc54:
Thanks Chris, I will have a play with it...

> also not sure why the period in
> 
> ```
> from .gmoccapy import player 
> ```
> 
> but you don't need it in my tests

Yes, I thought that I had removed that as it gave errors on one of my VMs.
All the different distro behaviour is getting to be a PITA...

I have also done a few things in Gscreen so I will send them your way to have a look at.

c-morley:
The many versions of linuxcnc is going to be a nightmare for us.
oh poor gscreen... thanks for that.

src/emc/usr_intf/gmoccapy/player.py
```diff
@@ -24,26 +24,36 @@
...
 import gi
-from gi.repository import Gtk as gtk
-from gi.repository import GObject as gobject
-import gst
+from gi.repository import GLib
+
+from qtvcp import logger
+LOG = logger.getLogger(__name__)
+# Force the log level for this module
+LOG.setLevel(logger.INFO) # One of DEBUG, INFO, WARNING, ERROR, CRITICAL# attempt to setup audio
```
c-morley, line 32: 'LOG.setLevel'  is usually commented out.
It is used to force this file to a different log level then gmoccapy proper. Usually when debugging.

phillc54: Fixed.

hansu (approved):
Looks good for me. @c-morley also good for you?

c-morley:
All Good.

hansu:
Thanks!

PR #2355, opened 2023-02-21
gscreen: add setting of log levels, fix Vte terminal

c-morley:
draft for gscreen logging. I think others are doing gscreen work so don't want to create merge problems.
Can be used as a guide or I can do this later.

### DIFF
--- a/src/emc/usr_intf/gscreen/gscreen.py
+++ b/src/emc/usr_intf/gscreen/gscreen.py
@@ -44,27 +44,16 @@ for num,temp in enumerate(sys.argv):
         if temp == '-h' or temp == '--help' or len(sys.argv) == 1:
             _print_help()
 
-# Set up the base logger
-#   We have do do this before importing other modules because on import
-#   they set up their own loggers as children of the base logger.
-from qtvcp import logger
-LOG = logger.initBaseLogger('GScreen', log_file=None, log_level=logger.INFO)
-
 import gi
 gi.require_version("Gtk","3.0")
 gi.require_version("Gdk","3.0")
 gi.require_version('PangoCairo', '1.0')
 from gi.repository import Gtk,Gdk,GObject,Pango,PangoCairo,cairo,GLib
 from gi.repository import Pango as pango
-try:
-    from gi.repository import Vte as vte
-except:
-    LOG.error("**** WARNING GSCREEN: could not import vte terminal - is package installed?")
 
 import hal
 import errno
-import gladevcp.makepins
-from gladevcp.gladebuilder import GladeBuilder
+
 #import pango
 import traceback
 import atexit
@@ -144,10 +133,6 @@ try:
 except:
     pass
 import linuxcnc
-from gscreen import emc_interface
-from gscreen import mdi
-from gscreen import preferences
-from gscreen import keybindings
 
 # this is for hiding the pointer when using a touch screen
 #pixmap = gdk.Pixmap(None, 1, 1, 1)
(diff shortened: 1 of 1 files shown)


phillc54:
I applied this patch to my current gscreen branch and there were only a few conflicts.
One conflict was the Vte terminal as I had done that as well but with spawn_async rather than spawn_sync.

I changed the "Gmoccapy"  in the following line to "Gscreen"
`LOG = logger.initBaseLogger('Gmoccapy', log_file=None, log_level=logger.WARNING)`

I was getting an error with `LOG.exception()` so I changed it to `LOG.error(e)`

I moved all the logging initialization up near the top so it would be available for the audio and desktop notification setup.

I added several LOG commands and it all seems to work well.

c-morley:
If you move the logging initialization to the top then you can't set the logging level on the gladevcp libraries - they will always be at warning level.
spawn_asynch is the newest code, I noticed that too late :)
What was the error with log.exception ? I started using it in qtvcp, it formats errors nicer with better info - has the newest libraries changed again? Thanks for feed back.

phillc54:
> If you move the logging initialization to the top then you can't set the logging level on the gladevcp libraries - they will always be at warning level.

So should we move the notify and audio test down the bottom below logging?

> spawn_asynch is the newest code, I noticed that too late :)

Yeah, I had almost completed sync and then noticed the change.

> What was the error with log.exception ? I started using it in qtvcp, it formats errors nicer with better info - has the newest libraries changed again?

Oh, it was just missing the argument 'e'

c-morley:
> > If you move the logging initialization to the top then you can't set the logging level on the gladevcp libraries - they will always be at warning level.
> 
> So should we move the notify and audio test down the bottom below logging?

yes 
> 
> > spawn_asynch is the newest code, I noticed that too late :)
> 
> Yeah, I had almost completed sync and then noticed the change.
> 
> > What was the error with log.exception ? I started using it in qtvcp, it formats errors nicer with better info - has the newest libraries changed again?
> 
> Oh, it was just missing the argument 'e'

lol of course! Nice catch.
It's supposed to be used only with try - exception code. 
It prints at the ERROR log level but prints more exception-like details.
The levels are (in order)  VERBOSE, DEBUG, INFO, WARNING, ERROR (and exception) and finally CRITICAL

phillc54:
The more I look at logging the more I like it...

I just tried by adding all six levels of logging in the __init__ function of gladevcp/hal_mdihistory.py and the global level setting seems to work whethe the logging setup code is at the top or the bottom of gscreen.py.

c-morley:
Maybe I need to look at your code to understand.

The loglevel set by code such as
logger.setGlobalLevel(logger.INFO)
must be set before any file that uses logging is imported

at the same time that  etGlobalLevel line must be called after the sys args can be read (to see the -i,-d etc)

Yes logging is really nice and can improve performance - printing a lot is expensive.

I'm off for sleep.. happy coding.

phillc54:
Attached is what I tested with. This goes on top of the current 2.9 branch.
I tested with sims/gscreen/gscreen ini using -v then -d  then -i then -q and the global log level seems to work.

I also noticed that gladevcp/iconview.py has no comment in front of LOG.setlevel.
 
[gscreen_logging_test.patch](https://github.com/LinuxCNC/linuxcnc/files/10790616/gscreen_logging_test.patch)

c-morley:
Well Phill that is great!
I never thought sys.arg was valid at that time.
That is way nicer way to lay it out.
It's amazing what one can do - when one doesn't know they can't!

Chris
________________________________

phillc54:
> It's amazing what one can do - when one doesn't know they can't!

I find that out often and mostly accidently.

c-morley:
Closing this do to Phill's pull request supersedes it.

PR #2356, opened 2023-02-21
gscreen: assorted updates

phillc54:


src/emc/usr_intf/gscreen/gscreen.py
```diff
@@ -53,10 +53,11 @@ def _print_help():
...
-gi.require_version('PangoCairo', '1.0')
-from gi.repository import Gtk,Gdk,GObject,Pango,PangoCairo,cairo,GLib
+from gi.repository import Gtk, Gdk, GLib
 from gi.repository import Pango as pango
+
 try:
+    gi.require_version('Vte', '2.91')
     from gi.repository import Vte as vte
 except:
     LOG.error("**** WARNING GSCREEN: could not import vte terminal - is package installed?")
```
c-morley: this  should be a warning only. Gscreen screens can run without Vte.
I would remove the ****** WARNING GSCREEN.

src/emc/usr_intf/gscreen/gscreen.py
```diff
@@ -98,15 +98,14 @@ def _print_help():
...
 try:
-    import pygst
-    pygst.require("0.10")
-    import gst
+    gi.require_version('Gst', '1.0')
+    from gi.repository import Gst
     _AUDIO_AVAILABLE = True
     print("**** GSCREEN INFO: audio available!")
 except:
+    _AUDIO_AVAILABLE = False
```
c-morley, line 135: LOG.debug fir audio available
log warning for missing library

src/emc/usr_intf/gscreen/gscreen.py
```diff
@@ -2570,7 +2570,7 @@ def on_theme_choice_changed(self, widget):
         """
         active_iter = widget.get_active_iter()
         if active_iter:
-            print("texst",widget.get_model()[active_iter][0])
+#            print("test",widget.get_model()[active_iter][0])
```
c-morley: probably removable

src/emc/usr_intf/gscreen/gscreen.py
```diff
@@ -164,7 +164,7 @@ def _print_help():
 def dbg(str):
     global gscreen_debug
     if not gscreen_debug: return
-    print(str)
+    print("**** GSCREEN DEBUG:", str)
```
c-morley: using LOG is better then '**** GSCREEN DEBUG 'etc

c-morley:
I put my 2 cents in but I think you already were thinking the same things.
That's a real good spring cleaning!

phillc54:
I think it is almost good enough, the GUI is almost done as well as the CSS for the default theme.

c-morley:
I just tried it out - that looks a lot more functional. I think it should go in 2.9 though.

phillc54:
Oops, I thought that I had set it for 2.9, just another mistake :)
I will change it shortly.

phillc54:
I think this should be ok to go now.
I will take a break and come back later to add a bit of polish.

phillc54:
I think that I may have broken something to do with exceptions.
I tried to start SilverDragon and it crashes because it cannot find the colorbuttons.
I will see what I can do tomorrow.

phillc54:
> I think that I may have broken something to do with exceptions.

That was a false alarm, it was caused by issues in SilverDragon.

c-morley:
Should we push this in now Phill?

phillc54:
I found an issue with sounds that I am working on at the moment.

phillc54:
I am not sure if hard coding the sounds directory was a good idea...

c-morley:
I'm sorry I missed this. I think for now hard coding is fine. It does seem in different linux flavours the sounds are in different places. In Qtvcp there is a function for searching known locations for this reason.

If you are happy, could you fix the conflict and merge?

phillc54:
Yep, I can do that. I probably need to test it all against the gladevcp changes as well.

Off Topic: Any hints on why an xid cannot be found for the hal_gremlin widget when using Glade. This is the line (I think) in gremlin.py:
            self.xwindow_id = GdkX11.X11Window.get_xid(widget.get_window())

c-morley:
You are using :
self.xwindow_id
in hal_gremlin?
Do you get 'None' or an error?

phillc54:
It gives an int of zero.
It is in the reshape function of gremlin.py

c-morley:
The window_id is not available till later in the process - probably after first realization.
check it in hal_gremlin's  'fileloaded' function - you should get the id.

phillc54:
Thanks, I will have a look tomorrow.
Would you prefer all these gscreen commits squashed?

c-morley:
I think squashing would be good.

phillc54:
I fixed the conflict and changed sounds so they are not hard coded. I cannot test in Mint so I left the commits separate so you can easily see what I did. If you are happy with it I will squash and push.

c-morley:
Looks good to me. Thank you.

PR #2359, opened 2023-02-23
sserial: fix float parameters

rmu75:
hm2_sserial_get_param_value asusmes float parameters to have same length as HAL float type. this change converts parameter from actual indicated length or warns if length is not supported, similar to process values.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/sserial.c
+++ b/src/hal/drivers/mesa-hostmot2/sserial.c
@@ -385,8 +385,21 @@ int hm2_sserial_get_param_value(hostmot2_t *hm2,
         case LBP_ENCODER:
             break; // Hard to imagine an encoder not in Process data
         case LBP_FLOAT:
-            r = hm2_sserial_get_bytes(hm2, chan, (void*)&(p->float_written),
-                                      g->ParmAddr, g->DataLength/8);
+            {
+                char buf[g->DataLength/8];
+                r = hm2_sserial_get_bytes(hm2, chan, &buf[0], g->ParmAddr, g->DataLength/8);
+                if (g->DataLength == sizeof(float) * 8) {
+                    float temp;
+                    memcpy((void*)&temp, &buf[0], sizeof(float));
+                    p->float_written = temp;
+                } else if (g->DataLength == sizeof(double) * 8) {
+                    double temp;
+                    memcpy((void*)&temp, &buf[0], sizeof(double));
+                    p->float_written = temp;
+                } else {
+                    HM2_ERR("sserial get param value: LBP_FLOAT of bit-length %i not handled\n", g->DataLength);
+                }
+            }
             if (set_hal) p->float_param = p->float_written;
             HM2_DBG("LBP_FLOAT %f %f \n", p->float_param, p->float_written);
             break;


andypugh:
I am a bit confused by this at first, as I was sure that sserial supports all sorts of different floats in the parameters section, including 8-bit minifloats. 
https://github.com/LinuxCNC/linuxcnc/blob/2.8/src/hal/drivers/mesa-hostmot2/sserial.c#L1588

And there is code similar to yours in a number of places:
https://github.com/LinuxCNC/linuxcnc/blob/2.8/src/hal/drivers/mesa-hostmot2/sserial.c#L1889
https://github.com/LinuxCNC/linuxcnc/blob/2.8/src/hal/drivers/mesa-hostmot2/sserial.c#L1661

Are you certain that your code is needed? (and if it is, why not support 16 and 8 bit floats too?)

rmu75:
With this code, my use case works. PCW said in IRC that there are probably no (other) users of float parameters at the moment.

andypugh:
I don't think that there needs to be a C/C++ datatype for the smaller floats. The internal storage can be double (or hal_float_t) 
The reason for the minifloats is that there was only 96 bits in a sserial packet. The trick is to convert those smaller bit patterns into a normal float.
I didn't actually do this, but that was the plan. I think it might be as simple as filling out the sign, mantissa and exponent as separate bit patterns, but I would need to look into that.

rmu75:
It could certainly be done with a packed bitfield struct, but IMO adding 8 and 16 bit floats would be orthogonal to making float parameters working.

andypugh:
> It could certainly be done with a packed bitfield struct, but IMO adding 8 and 16 bit floats would be orthogonal to making float parameters working.

You are right, of course.
I have tried looking at numbers in single and double precision format, and the bit fields do appear to the the same (just truncated) between the types.

However, digressions aside, are you convinced this is needed and works?

rmu75:
You can't just truncate 4 bytes from float64 to get a float32, you would need some rounding and exponent in float32 has only 8 bits vs. 11 bits in double.

I will check again tomorrow on another system with and without patch, but I'm pretty sure that verbatim copying of sserial-remote-provided bytes into the hal_float_t in this function is incorrect and conversion like it is done with process data here https://github.com/LinuxCNC/linuxcnc/blob/2.8/src/hal/drivers/mesa-hostmot2/sserial.c#L2116 is needed.

LBP_NONVOL_FLOAT is not handled, that remains to be done.

andypugh:
> You can't just truncate 4 bytes from float64 to get a float32, you would need some rounding and exponent in float32 has only 8 bits vs. 11 bits in double.

Indeed not. What I mean is that you can recalculate the exponent based on the different bitsize and offset, then put that in the exponent bits, and then put the mantissa directly into those bits, accepting that it is shorter and less accurate.

rene-dev:
stmbl uses floats for process data and parameters, but not for NONVOL.

rmu75:
> stmbl uses floats for process data and parameters, but not for NONVOL.

at first glance STMBL has one float parameter, `scale`, and that has a default value of 0 (I didn't find any initialization), in that case this bug doesn't bite because 0 is just a bunch of 0x00s in float32 and float64. stmbl.hal example file sets the parameter, after that it is never read back.

I don't have a working STMBL to test myself, but i suspect if you initialize scale in sserial.c to e.g. "10", it will read back as 5.398241e-315 and setting to 0 from linuxcnc will probably not work.

rmu75:
Original code reads variable number of bytes (specified in PD) into hal_float_t variable:
`r = hm2_sserial_get_bytes(hm2, chan, (void*)&(p->float_written),
                                      g->ParmAddr, g->DataLength/8);`
and then immediately interprets those variable bytes as hal_float_t:
`HM2_DBG("LBP_FLOAT %f %f \n", p->float_param, p->float_written);`

rmu75:
another machine, linuxcnc 2.9-pre, run in place, without patch.

from halshow:
```
Parameters:
    36  float RW   5.164076e-315  hm2_7i97.0.rsio.0.0.ftest
    36  s32   RW              1  hm2_7i97.0.rsio.0.0.stest
    36  float RW   5.418964e-315  hm2_7i97.0.rsio.0.0.tlh
    36  float RW   5.398241e-315  hm2_7i97.0.rsio.0.0.tll
    36  u32   RW     0xFFFFFFFF  hm2_7i97.0.rsio.0.0.utest
```
tlh and tll are 32 bit floats and should be 14.0 resp. 10.0. 

according to https://www.omnicalculator.com/other/floating-point, 10.0 is 0x41200000, 0x0000000041200000 is 5.398241248e-315; 14 is 0x41600000, 5.41896386e-315 is 0x0000000041600000.

andypugh:
OK, I am convinced. I want to make a more generic function for the compact floats at some point, but I don't know when I will get to it.

rmu75:
thank you.

About the minifloats, I didn't find any standard format. Seems number of bits and exponent offset are application specific. For 16 bit floats, there is at least a standard interchange format, but it seems a bunch of different formats like truncated 32bit float are in use too.

But I could certainly help extracting the float conversion stuff into generic functions.

PR #2363, opened 2023-02-26
Update plasma-cnc-primer.adoc

JimmyRigIt:
Ohmic touchoff documentation fix - The current documentation is out of date and wont compile for qtplasmac 2.9
Below is the fix


smoe:
@hansu - you know such things. The documentation should likely find a way to state how it worked for 2.8, too.

hansu:
I have no idea what the `db_ohmic` component is neither where it comes from.

phillc54:
It is just the [dbounce](http://linuxcnc.org/docs/devel/html/man/man9/dbounce.9.html) component loaded with the name db_ohmic. Originally plasmac configs used the old [debounce](http://linuxcnc.org/docs/devel/html/man/man9/debounce.9.html) component which does not use names= . It was replaced with the dbounce component in all new plasmac based configs in September 2020.

Issue #2368, opened 2023-03-01
LCNC 2.9 - hexagui.py doesnt work


zz912:
I am trying make own Vismach. I inspired in examples. I found that hexagui.py doesnt work.
Model is loaded, but I cant move or rotate with it.

If I comment:
hexagui.py line: 148
 strut = Track([strut],base_joint_coords, plat_joint_coords, world_coords)

It can be loaded and I can move or rotate with it. But struts are not in correct position 
![hexagui](https://user-images.githubusercontent.com/96618597/222131710-c9f7a82b-f247-48e3-a3a1-cb710f2bf3f9.png)
hexagui.py is the only example that uses "Track". So I don't know if the error is in hexagui.py (bad using Track) or in vismach.py (bad definition Track)

zdenek@CNC:~/linuxcnc/linuxcnc-2.9/src$ halrun
halcmd: loadusr -W hexagui
halcmd: vismach.py: Track: why am i here? world is not in the scene yet
vismach.py: Track: why am i here? world is not in the scene yet
vismach.py: Track: why am i here? world is not in the scene yet
vismach.py: Track: why am i here? world is not in the scene yet
vismach.py: Track: why am i here? world is not in the scene yet
vismach.py: Track: why am i here? world is not in the scene yet
Exception in Tkinter callback
Traceback (most recent call last):
  File "/usr/lib/python3.7/tkinter/__init__.py", line 1705, in __call__
    return self.func(*args)
  File "/home/zdenek/linuxcnc/linuxcnc-2.9-Tripteron/lib/python/rs274/OpenGLTk.py", line 361, in tkMap
    self.tkExpose()
  File "/home/zdenek/linuxcnc/linuxcnc-2.9-Tripteron/lib/python/rs274/OpenGLTk.py", line 374, in tkExpose
    self.tkRedraw()
  File "/home/zdenek/linuxcnc/linuxcnc-2.9-Tripteron/lib/python/rs274/OpenGLTk.py", line 351, in tkRedraw
    self.redraw()
  File "/home/zdenek/linuxcnc/linuxcnc-2.9-Tripteron/lib/python/vismach.py", line 805, in redraw
    self.model.traverse()
  File "/home/zdenek/linuxcnc/linuxcnc-2.9-Tripteron/lib/python/vismach.py", line 41, in traverse
    p.traverse()
[...]

PR #2369, opened 2023-03-01
vismach.py update Track map_coords() indexing

dngarrett:
omission occuring at commit eb62efcaedfe4ce314e
issue #2368


SebKuzminsky:
I haven't tested this but I trust your work.

PR #2370, opened 2023-03-01
Raspberry Pi HM2 updates

SebKuzminsky:
This adds a sample config for a 7i90 connected to a Raspberry Pi via SPI.

Also makes an hm2_rpspi error message a little more helpful.


petterreinholdtsen:
Look good to me.

PR #2372, opened 2023-03-02
Merge 2.9 into master - conflicts

hansu:
The other conflics are easy to solve but here I am not sure which one is the correct wording:

`If present, local UI files` or `If present, local UI/QSS/python files`

I guess @c-morley should know that.

```
<<<<<<< HEAD
//FIXME Does that applies to .ui files only or handler.py and qss too ?
//FIXME How does overriding/inheritance actually works ?

If present, local UI files in the configuration folder will be loaded instead of the stock UI files.

Local UI files allow you to use your customized designs rather than the default screens.

QtVCP will look for a folder named _<screen_name>_ (in the launched configuration folder that holds the INI file).
=======
If present, local UI/QSS/python files in the configuration folder will be loaded instead of the stock UI files.

Local UI/QSS/python files allow you to use your customized designs rather than the default screens.

QtVCP will look for a folder named <screen_name> (in the launched configuration folder that holds the INI file).
>>>>>>> 2.9
```

### DIFF
--- a/configs/by_interface/mesa/hm2-servo/7i93.ini
+++ b/configs/by_interface/mesa/hm2-servo/7i93.ini
@@ -57,7 +57,7 @@ CYCLE_TIME =            0.010
 # files, and then to execute any individual HALCMD commands.
 # list of hal config files to run through halcmd
 # files are executed in the order in which they appear
-HALFILE = hm2-servo.hal
+HALFILE = hm2-servo-eth.hal
 # list of halcmd commands to execute
 # commands are executed in the order in which they appear
 #HALCMD =               save neta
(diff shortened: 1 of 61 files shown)


c-morley:
https://github.com/LinuxCNC/linuxcnc/commit/25e07f808af318ce5783e73105add4fe312b81b2#r102843546

hansu:
@smoe: https://github.com/LinuxCNC/linuxcnc/commit/25e07f808af318ce5783e73105add4fe312b81b2#r102859688

smoe:
What do you want me to do?

hansu:
> What do you want me to do?

Not more as you already did. I only wanted to make sure that you get notified by that comment.

hansu:
Thanks

PR #2375, opened 2023-03-03
mb2hal_HOWTO.ini - added some info to comments

zz912:
1) Added information:
#The address is a decimal number
2) Added example for control spindle driver.

### DIFF
--- a/src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini
+++ b/src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini
@@ -94,6 +94,7 @@ SERIAL_DELAY_MS=10
 MB_SLAVE_ID=1
 
 #REQUIRED: The first element address.
+#The address is a decimal number
 FIRST_ELEMENT=0
 
 #REQUIRED unless PIN_NAMES is specified: The number of elements.
@@ -133,6 +134,16 @@ PIN_NAMES=cycle_start,stop,feed_hold
 #     mb2hal.m.n.int     (input)
 #     NELEMENTS needs to be 1 or PIN_NAMES must contain just one name.
 #     Both pin values are added and limited to 65535 (UINT16_MAX). Normally use one and let the other open (read as 0).
+#     Usage tip:   If you need to change individual bits in a register, use the MUX4 function.
+#                  # setup Modbus code 
+#                  # for control spindle driver
+#                  setp mux4.0.in0 1    # bit1-0 (01B:Stop)
+#                  setp mux4.0.in1 18   # bit1-0 (10B:Run) & bit5-4 (01B:FWD)
+#                  setp mux4.0.in2 34   # bit1-0 (10B:Run) & bit5-4 (10B:REV)
+#                  # RUN FOWARD, RUN REVERSE, STOP
+#                  net spindle-fwd 	spindle.0.forward => mux4.0.sel0
+#                  net spindle-rev 	spindle.0.reverse => mux4.0.sel1
+#                  net spindle-command	mb2hal.m.n.float <=  mux4.0.out
 # fnct_15_write_multiple_coils:
 #     mb2hal.m.n.bit     (input)
 # fnct_16_write_multiple_registers:


src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini
```diff
@@ -94,6 +94,7 @@ SERIAL_DELAY_MS=10
 MB_SLAVE_ID=1
 
 #REQUIRED: The first element address.
+#The address is a decimal number
```
hansu, line 97: Numbers including a comma is not allowed here afaik
```suggestion
#The address is an integer number
```

src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini
```diff
@@ -133,6 +134,16 @@ PIN_NAMES=cycle_start,stop,feed_hold
...
+#     Usage tip:   If you need to change individual bits in a register, use the MUX4 function.
+#                  # setup Modbus code 
+#                  # for control spindle driver
+#                  setp mux4.0.in0 1    # bit1-0 (01B:Stop)
+#                  setp mux4.0.in1 18   # bit1-0 (10B:Run) & bit5-4 (01B:FWD)
+#                  setp mux4.0.in2 34   # bit1-0 (10B:Run) & bit5-4 (10B:REV)
+#                  # RUN FOWARD, RUN REVERSE, STOP
+#                  net spindle-fwd 	spindle.0.forward => mux4.0.sel0
+#                  net spindle-rev 	spindle.0.reverse => mux4.0.sel1
+#                  net spindle-command	mb2hal.m.n.float <=  mux4.0.out
```
hansu, line 146: First, the description is not correct. If you want to change individual bits you can use e.g. [`bitwise`](https://linuxcnc.org/docs/2.9/html/man/man9/bitwise.9.html)
What you are doing is you assign some specific values to one output signal depending on some bits. And the MUX output type is float, so not the best type if you want to have a bit-correct output.
And I see no relation here to mb2hal. This is more general.

hansu (changes requested):

src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini
zz912: "Integer" will not help mb2hal begginer user.

This is typical page from driver manual:
Delta_C2000-manual_modbus
Integer can be sometimes similar hexadecimal number.
I want make help, that user has to transform 2000H => 8192D, and then he must write only "8192"
This will help too:
https://www.binaryhexconverter.com/hex-to-decimal-converter
I am spend many time for find out it.


A fantastic thing would be if mb2hal could process xxxxH and convert it to a decimal number by itself. But this is beyond my programming capabilities.

src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini
zz912: 1) If I want use mb2hal for control spindle driver. I need change only bits in one register.
Delta_C2000-manual_modbus

In ideal world producers of drivers should use for command START, STOP, etc, "fnct_05_write_single_coil(05 = 0x05) " It should be fantastic. But we dont live in Ideal world, so I invent some solution.

2) I am not opposed to a better solution for controlling Modbus drivers using mb2hal. I don't know how the HAL module bitwise would help me. 
Currently, the HAL component classicladder or the HAL driver designed for a specific driver is used to control the Modbus driver. Using the classiclader component to control the spindle driver feels like using a machine gun on a sparrow. Writing a HAL component for every driver that exists is not the solution.

3) This combination of mb2hal and MUX already works for me in two cases.

4) I agree that FLOAT type MUX.OUT is not ideal.

src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini
zz912: My first encounter with Modbus was in LCNC using the mb2hal component. It took me a lot of time before I broke the first communication. There was a lot I had to study. One of the things that I lost a lot of time on was that the address should be converted from hexadecimal to decimal. Once I know it, I'll just convert it.

Can we write:
#The address is decimal system
or
#The address must be converted from hexagonal to decimal system
#you can use: https://www.binaryhexconverter.com/hex-to-decimal-converter
?

What do you suggest?

src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini
hansu: >     4. I agree that FLOAT type MUX.OUT is not ideal.

I think `mux_generic` is better here.

However, this fits better somewhere in the driver section or in HAL howto than here.

zz912: I think Modbus is most commonly used on CNC machines to control spindle drivers. That's why I put it here. If mb2hal was a component in a 3D printer project, I wouldn't put it there. If you don't agree with the example, I would like to provide information here that it is possible to use the ModBus function6 to control individual bits in the register.

Or we can generalize the example more.

Something similar is listed by the manufacturer of Delta drivers for input control:
Delta_C2000-control_inputs


When I tried to get the first ModBus spindle driver running. I tried changing individual bits in the registry with fnct_05_write_single_coil, which of course didn't work.

mux_generic is better.  In the example, it would be enough to rewrite mux4 as mux-gen

src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini
hansu: `#REQUIRED: The first element address (decimal integer)`

src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini
hansu: I suggest to write a short section in the [mb2hal doc page](http://linuxcnc.org/docs/2.9/html/drivers/mb2hal.html) or in the Hardware Examples section and add a note/reference in the sample INI file. But I don't want to bloat that INI file with general information.
A general howto would be nice how to set up linuxcnc with mb2hal to control VFDs.

src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini
zz912: #REQUIRED: The first element address (decimal integer)
I agree with it. Could you change it? Git and Github is not my strong point.

src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini
zz912: So, this part we will not added here. What should I do now?

hansu: This is just my opinion. But if nobody else jumps in here it will probably not be added.
You can write a short tutorial if you think this is helpful for other users.

SebKuzminsky:
I've only skimmed this discussion, sorry...  My take is:
* Integers never have decimal points in them.
* Integers can be written in decimal or in hex.
* Integers written in hex begin with "0x", ie "0x1b3c".
* Writing hex integers with a trailing "H", ie "1b3cH" is also somewhat customary, but it's not how our parsers will ever work.  Use "0x".
If one follows the above conventions, does the INI parser do the right thing?

zz912:
To: SebKuzminsky 
Although I started this PR, I'm already lost in the redo https://github.com/LinuxCNC/linuxcnc/pull/2380
Therefore, I will ask HnsU for a reaction.

If it was possible to make the modification as described by SebKuzminsky, I would ask you to modify the line:
https://github.com/LinuxCNC/linuxcnc/blob/a2b692dcc62fd9e33b597e366ec0f64e76912072/src/hal/user_comps/mb2hal/mb2hal_HOWTO.ini#L96

hansu:
The info has now been added. I will re-open #2380 when I have time to provide an appropriate solution only accepting "0x" as hex.

PR #2377, opened 2023-03-03
Adding Limit_Axis Component

satiowadahc:
This is a handy little component for changing axis limits dynamically. 

I have one machine with a multi-axis head that I prevent the head from spinning while Z is above the gantry so it doesn't hit itself. On the same machine X is prevented from running into the tool holder unless the motion.digital-output pin connected to it is high (while doing a tool change).

Any feed back is appreciated particularly on naming pins. min/max limit is the output for the given min/max range input. Sticky was a minor performance optimization to not check other range's if the current one is still active.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -934,6 +934,8 @@ obj-$(CONFIG_MATRIX_KB) += matrix_kb.o
 matrix_kb-objs := hal/components/matrix_kb.o $(MATHSTUB)
 obj-$(CONFIG_MUX_GENERIC) += mux_generic.o
 mux_generic-objs := hal/components/mux_generic.o $(MATHSTUB)
+obj-m += limit_axis.o
+limit_axis-objs := hal/components/limit_axis.o $(MATHSTUB)
 obj-$(CONFIG_PWMGEN) += pwmgen.o
 pwmgen-objs := hal/components/pwmgen.o $(MATHSTUB)
 obj-$(CONFIG_SIGGEN) += siggen.o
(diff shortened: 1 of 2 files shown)


satiowadahc:
Also if documentation is needed whats the current preferred way?

andypugh:
I am curious why you did this the "hard" way rather than writing it as a .comp ? 
As a .comp it would be self-documenting. 
Otherwise, you could add a new manpage in asciidoc format in 
/docs/src/man)/man9/
Or in troff format in 
/docs/man/man9

But, before dong that, I am curious about its usefulness. I am assuming that it drives the ini pins? As far as I know these are only read at the start of the G-code, so running the component in a realtime thread might not be particularly useful, it could (probably) be a userspace component that only runs when needed.

satiowadahc:
I could try to convert it to comp. Just haven't looked at the syntax.

Yes it drives ini limits. It works for position limits. Velocity, accel don't update. Following error does update, but I can't think of a use case for that.

andypugh:
I don't think that there is any point converting to comp now that it isn't a .comp.

petterreinholdtsen:
[andypugh]
> I don't think that there is any point converting to comp now that it
> isn't a .comp.

The small advantage is that the documentation and code is closer
together in a single file, increasing the chance of the two staying in
sync. :)

-- 
Happy hacking
Petter Reinholdtsen

satiowadahc:
I guess the personality arg might work to dynamically create pins? Maybe I'll give it a shot tonight.

andypugh:
It's not elegant (and leads to a slightly confusing auto-generated doc) but you can use an extra_setup to set the "personality" and thus control the pin creation. 
An example is http://linuxcnc.org/docs/stable/html/man/man9/bldc.9.html which mentions "personality" in the pin list, but it isn't a user-settable item. Instead it uses module parameters to generate it it in the extra_setup:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/components/bldc.comp#L851

andypugh:
Is this PR still current?

andypugh:
I converted the component to a .comp to generate docs and to allow names= etc. 
a4a8ce1a09744503949b390b800df0562137800e
5f4f074e30dcbca816b66e76a49e8dc9ca36d81b

PR #2380, opened 2023-03-04
inifile: parse numbers also as hex

hansu:
`boost::lexical_cast ` doesn't support hex numbers. So if the type is int, this tries now to parse with `std::stoi` to parse also numbers in hex format.

### DIFF
--- a/src/libnml/inifile/inifile.hh
+++ b/src/libnml/inifile/inifile.hh
@@ -91,8 +91,18 @@ public:
         try {
             *result = boost::lexical_cast<T>(tmp);
         } catch (boost::bad_lexical_cast &) {
-            ThrowException(ERR_CONVERSION);
-            return(ERR_CONVERSION);
+            // If boost::lexical_cast fails, try std::stoi to parse also numbers in hex format
+            if(std::is_same<T, int>::value) {
+                try {
+                    *result = std::stoi(tmp, nullptr, 16);
+                } catch (...) {
+                    ThrowException(ERR_CONVERSION);
+                    return(ERR_CONVERSION);
+                }
+            } else {
+                ThrowException(ERR_CONVERSION);
+                return(ERR_CONVERSION);
+            }
         }
 
         return(ERR_NONE);


petterreinholdtsen:
[Hans Unzner]
> `boost::lexical_cast ` doesn&#39;t support hex numbers. So if the type
> is int, this tries now to parse with `std::stoi` to parse also numbers
> in hex format.

I would imagine such behavial change would require updated also to some
documentation?  Should it not depend on hex markers like 0x before
trying to parse as hex, instead of only trying it for numbers that can
not be parsed as integers?  What if the user want to parse '123' as hex?

Which number types is the documentaiton claiming is supported in the INI
file?  Why is hex support needed now?

References: <URL: https://cplusplus.com/reference/string/stoi/ > and
<URL: https://cplusplus.com/reference/cstdlib/strtol/ >.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
Yes of course the docs need to be changed according to this. But I don't know if this will be accepted or not.
Because this is not the most elegant way of adding support for this, but changing boost::lexical_cast to std::stoi and std:stof and so on seems a bit overdone.

This change is motivated by https://github.com/LinuxCNC/linuxcnc/pull/2375 - modbus addresses of VDFs are sometimes noted in hexadecimal so this makes life in this case a bit easier. Not needed but a nice to have.

petterreinholdtsen:
[Hans Unzner]
> Yes of course the docs need to be changed according to this. But I
> don't know if this will be accepted or not.

> Because this is not the most elegant way of adding support for this,
> but changing boost::lexical_cast to std::stoi and std:stof and so on
> seems a bit overdone.

Not sure if I understand you here.

I would switching to stoi() with base=0 would be a better approach, to
get a well defined number parser.

--
Happy hacking
Petter Reinholdtsen

hansu:
As far as I understand it, the method `Find` here is generic for all types of `*result` except `std::string`.

So switching generally from boost::lexical_cast to std::stoi with base=0 does not include floats.

https://github.com/LinuxCNC/linuxcnc/blob/efdfde6d70eb4e2a208b93bf7a9ac0dea6e5a64c/src/libnml/inifile/inifile.hh#L83-L109

smoe:
You may want to cross-check this stackoverflow thread https://stackoverflow.com/questions/74302381/boostlexical-cast-can-convert-hex-inside-string-to-int
IIRC a concern was that (much like how we are using chmod) the "0x" prefix is not given for the modbus config, so any default conversion does not know how to read it.

hansu:
Sorry, I don't get what you mean.

smoe:
> Sorry, I don't get what you mean.

Do you write
```
HEXVAR=0x42
```
or 
```
HEXVAR=42
```
to assign a value of (decimal) 66 to a variable that expects a hexadecimal value?

petterreinholdtsen:
[Hans Unzner]
> With this change you can write
>
> `MB_SLAVE_ID = 16`
>
> or
>
> `MB_SLAVE_ID = 0x10`

Yes, but it would be quite unclear to the untrained reader of '16' is
hex or decimal, given that 1a is hex.  Number sequences like would break
the priciple of least surprise if parsed using this method:

 MY_IDS = 19, 1a, 1b, 1c, 1d, 1e, 1f, 20

If I understand correctly, it would be desimal 19, 26, 27, 28, 29, 30,
31, 20, which is very contrary to what I expect.  I believe somethink
like this would lead to less surprises:

 MY_IDS = 0x19, 0x1a, 0x1b, 0x1c, 0x1d, 0x1e, 0x1f, 0x20
 
-- 
Happy hacking
Petter Reinholdtsen

smoe:
@rene-dev What would be the right way to design it? HAL knows all INI files and every module asks HAL? I am not sure if this is a thing, but I am somewhat worried that a value of an INI file could be changed by setp and a later loaded module is not aware of that change, parsing the same INI file to then stick to the unchanged default value?!?

smoe:
> [Petter]
> 
> > Yes, but it would be quite unclear to the untrained reader of '16' is
> > hex or decimal, given that 1a is hex.
> 
> Yeah without "0x" it will be treated as decimal. A bit stupid, that "1a" is allowed without leading "0x".

Should there be a warning?

> [Steffen]
> 
> > What would be the right way to design it? HAL knows all INI files and every module asks HAL?
> 
> The LinuxCNC core looks into the INI file to set up the motion stuff and the GUIs looks for parameters to set up the controls and also GUI-customizing parameters which the "core" doesn't even know about.

I just want to be understood with my concern that the GUI may be asked to read an INI file that the HAL file has already read and you then do not know if that value that is read was not changed. For instance some max speed is of concern for everyone.

hansu:
>Should there be a warning?

It could be changed that trying to parse HEX values without leading "0x" throws an error.

>I just want to be understood with my concern that the GUI may be asked to read an INI file that the HAL file has already read and you then do not know if that value that is read was not changed. For instance some max speed is of concern for everyone.

Then it's a design error of the external program reading the INI file...

andypugh:
Is this in or out? I am not sure that the discussion reached consensus. 
I am generally in favour of acceptinf 0xNNN as a hex variable, less so in accepting 1a (etc) as hex if dec fails, as dec will only fail part of the time if hex is intended, and then you get the wrong number. 

But perhaps I scanned this too quickly and this has been addressed?

hansu:
Yeah that's the consensus. Currently it also accepts e.g. 1a as hex so there is some more work needed. Currently I prioritize other topics, so this might take some time.

SebKuzminsky:
It looks like this PR changes the NML INI parser so that if it fails to parse a value as a decimal int, it tries again as if though the value had a "0x" prefix (ie, parse it as a hex value), even if it doesn't.  This is not correct behavior, according to the discussion/consensus above.

hansu:
> It looks like this PR changes the NML INI parser so that if it fails to parse a value as a decimal int, it tries again as if though the value had a "0x" prefix (ie, parse it as a hex value), even if it doesn't. This is not correct behavior, according to the discussion/consensus above.

Yes that's correct and that's the reason why this branch wasn't merged yet. I should better have marked it as draft.
That's why I write "there is some more work needed."

PR #2381, opened 2023-03-05
gladevcp: replace some deprecated widgets

phillc54:


hansu:
Great, this was next on my agenda. I am currently converting deprecated widgets in gmoccapy. See https://github.com/hansu/linuxcnc/tree/gmoccapy-gtk3-wip (to avoid double work).

I also did an attempt to convert offetpage.glade to Gtk3. But just saving with Glade >= 3.22 broke some function in that widget. Did you encounter something like this too?

phillc54:
I saved all my glade work at Gtk V3.24.
I think that there were a few color issues that I needed to fix. One was odd in that Gmoccapy returned a string color and Gscreen returned a Gdk color. You can see that at L213 and L221 of the code.
I did some work today on the gladevcp version of versa-probe but I need to find some documentation on how it is supposed to be added to a config.
This PR adds a couple of deprecation messages to Gmoccapy at startup but they shouldn't be too difficult to fix.

lib/python/gladevcp/offsetpage.glade
```diff
@@ -38,252 +38,257 @@
...
+        <col id="6">0</col>
+        <col id="7">0</col>
+        <col id="8">0</col>
+        <col id="9">0</col>
         <col id="10">True</col>
         <col id="11">False</col>
-        <col id="12" translatable="yes"></col>
-        <col id="13" translatable="no">black</col>
+        <col id="12">0</col>
+        <col id="13">black</col>
```
hansu, line 53: I think this is the key why it didn't worked for me when I converted it. Have you done these changes manually?

hansu:
The `SpeedControl` seems not to work properly:

grafik

phillc54:
Yes, the offsetpage glade file needed to be done manually to fix those issues, now it saves fine directly from glade.
The speedcontrol should be ok now.

hansu:
Sorry, still a bit tuning needed :innocent: 

grafik

phillc54:
That can be fixed in gmoccapy.glade by setting the top and bottom margins.


Screenshot 2023-03-07 at 9 58 36 am

phillc54:
That works for me. I guess it depends on whether you look at it from a gladevcp perspective or a gmoccapy perspective. I am not sure about the name change for the container though as I don't know why it was an instance variable originally.

hansu:
What depends on the point of view?
I guess a local variable for the container would work here as well.
And I think the name change is fine - even if the name stays the same, the container type is different, so the type change  would already break "external access".
And it's weird to have a self.table which is actually a grid or a box.

phillc54:
> What depends on the point of view?

Whether your change is required, it works well for all situations I have seen except for the ugly (IMHO) layout in one position in gmoccappy.

> I guess a local variable for the container would work here as well.
> And I think the name change is fine - even if the name stays the same, the container type is different, so the type change would already break "external access".

Without knowing if any users reference it directly and for what reason then I am not sure.

> And it's weird to have a self.table which is actually a grid or a box.

Agreed, but it was just to help prevent broken configs in case some users are referencing it.


I have probably done the same in some of the other widgets so you may need to change them as well.

hansu:
>Whether your change is required, it works well for all situations I have seen except for the ugly (IMHO) layout in one position in gmoccappy.

But it doesn't hurt to add those seven lines. And you don't know in how many other custom VCPs/GUIs this widget is used and if they need to adapt too.

hansu:
>Edit: I have noticed a couple of other deprecations as well, I will get to them in the next few days.

I guess you are now done with them?

phillc54:
Yes, I think that is enough for the time being.

hansu:
I did a bit testing and found that my suggestion did not really work with the Box version you have chosen.
The vexpand has some weird effects in gmoccapy especially in the tool table page.
So please remove the expand properties again. Sorry.
And a rebase to 2.9 would be nice. Thanks.

``` patch

diff --git a/lib/python/gladevcp/speedcontrol.py b/lib/python/gladevcp/speedcontrol.py
index 69c8d59c1e..cdb60777fb 100755
--- a/lib/python/gladevcp/speedcontrol.py
+++ b/lib/python/gladevcp/speedcontrol.py
@@ -140,12 +140,8 @@ class SpeedControl(Gtk.Box, _HalSpeedControlBase):
 
 
         self.draw.set_valign(Gtk.Align.CENTER)
-        self.draw.set_hexpand(True)
-        self.draw.set_vexpand(True)
         self.btn_plus.set_valign(Gtk.Align.CENTER)
-        self.btn_plus.set_vexpand(True)
         self.btn_minus.set_valign(Gtk.Align.CENTER)
-        self.btn_minus.set_vexpand(True)
```

phillc54:
I guess that is because LinuxCNC is not running?

c-morley:
looks like someone changed it to 'lathe_view_option' in gremlin.
hal_gremlin looks for the INI in some crazy ways there.
I would let gremlin take care of that and add an error trap to it, to create a dummy Stat() for running in glade. See qt5_graphics for ideas.

phillc54:
I just added:
```
        if not ini_filename:
            self.lathe_option = False
            self.initialised = False
            return

``` 
above this line in the __init__ function:
`       inifile = linuxcnc.ini(ini_filename)
`
and it seems to "fix" it for loading in Glade.

c-morley:
Do you still see a graphics display in GLade with that?

phillc54:
I am guessing not. :(.
It looks the same as it was with the errors above.
It seems there is a bit more to do here, I will have a look at your suggestion.
Screenshot 2023-03-14 at 2 09 54 pm

phillc54:
Updated speedcontrol, hid the error message in gremlin and rebased to 2.9.

I will work on gremlin showing properly in Glade as a separte commit. There is also work to do on hal_pyngcgui and hal_gremlin_plus.

hansu:
Great, thanks!

PR #2384, opened 2023-03-05
Gmoccapy: convert deprecated widgets to Gtk >= 3.2

hansu:
This is only for information what I am currently working on to avoid double work.


hansu:
Replaced by #2461

PR #2391, opened 2023-03-14
docs: rename hal_parport.1 to hal_parport.9, because it's a realtime …

SebKuzminsky:
…component

Update po4a.cfg and hal/components to match.


hansu:
We had that discussion about userspace vs. non-realtime. Does anyone remember where that was and what was the point at the end?
Changing everything from userspace to non-realtime should also sync that page: http://linuxcnc.org/docs/2.9/html/hal/components.html

SebKuzminsky:
@petterreinholdtsen said in IRC this looks ok to him.

SebKuzminsky:
This merge also updates `components.adoc` to reflect the new reality - thanks for checking!

The `README.realtime-nomenclature` file was not meant to be part of this PR, and I reverted it immediately.

PR #2395, opened 2023-03-16
gladevcp: fix glade catalog paths for package installations

phillc54:


phillc54:
I think something went wrong here...

I will close and try again.

PR #2396, opened 2023-03-16
gladevcp: fix glade catalog paths for package installations

phillc54:


hansu:
Great! Also nice the fix regarding the icons. Thanks!

hansu:
I think it was a good step to have the gladevcp-icons in a separate folder. But I think they will fit better in `share/glade/icons/` instead of `lib/python/gladevcp/glade-icons/`  as the target  is `/usr/share/glade/`. Or do you think they are then "too far away" from the other gladevcp files?

PR #2397, opened 2023-03-16
docs: add info about Gstat message 'graphics-gcode-properties'

hansu:
this adds information about recently added functionality in https://github.com/linuxcnc/linuxcnc/commit/97f8cb071c1a7976b9c27cadf9b99303aba19c1d.

@c-morley Any idea what "x_zero_rxy" and so on expresses?


c-morley:
@snowgoer540 : help me out please:
This has to do with getting the outer limits on a rotated-on-Z program

snowgoer540:
@c-morley, sure.  

min_extents and max_extents are calculated based on drawing a bounding box around the previewed shape.  If you applied an XY rotation via G10 offset, the bounding box size will change, as it is still the largest box around the shape.   On our plasma screens, we have the ability to "frame" the area to be cut.  Previously we used the min and max extents to calculate the coordinates for the framing move, but if you had a G10 rotation applied (for a crooked sheet of material, for example), it could be very inaccurate.

I added min_extents_zero_rxy and max_extents_zero_rxy to solve this.  The result ends up being a bounding box calculated as if the preview was being displayed at zero rotation about XY.  That allows us to rotate and translate the new bounding box by the XY rotational offset and about the active offset's origin to get the framing coordinates (commit 5941fee adds this updated framing method to QtPlasmaC).

At any rate, other screens can now use this info and take this approach to "frame" the area to be cut on a machine (laser cutter, router table, etc) if they so desire.

Maybe that's more than you were looking for... looking at commit 6c73748 may be helpful.

Let me know if you need more info.

hansu:
Thanks for clarifying.
Is it like this (assuming the rectangle in gray is a rotated coordinate system) or the other way around?
bild

hansu:
But `z_zero_rxy` would be always the same as `z` and is only there for constancy right?

snowgoer540:
Yes, that's correct.  

Consistency, and leaving it allowed the use of pre-existing functions without hassle.

snowgoer540:
The images you added for the docs look really nice!

PR #2403, opened 2023-03-20
Docs correct description of linunxcnc.command.feedrate

hansu:
As far as I observed, this sets the feedrate override and not the absolute feedrate. That would also match the `linuxcnc.stat.feedrate` message.


hansu:
This line 
https://github.com/LinuxCNC/linuxcnc/blob/ec2dc5fe7d1b55bb43c5f96b1e2f31e3d61d2fa8/src/emc/usr_intf/axis/extensions/emcmodule.cc#L456 and so on confirms my observation.

PR #2406, opened 2023-03-23
gladevcp: add font properties to drowidget

phillc54:


docs/src/gui/gladevcp.adoc
```diff
@@ -576,6 +575,17 @@ window. For example, selecting a 'HAL_LED' widget, this will show that
 a 'HAL_LED' is derived from a  'GtkWidget', which in turn is derived
 from a 'GtkObject', and eventually a 'GObject'.
 
+Full class hierarchy can be seen by invoking the GtkInspector while in
+the Glade GUI by selecting a widget then pressing Control-Shift-I.
+If the Inspector doesn't open then it can be enabled from a terminal by
```
hansu, line 580: That's a really good hint, nice to have it mentioned here!

docs/src/gui/gladevcp.adoc
```diff
@@ -1509,26 +1523,40 @@ Reference Type::
...
+  Specify the size of the font between 8 and 96. Defaults to 40
+Font Weight::
+  Specify the weight of the font. Select from lighter, normal, bold, or bolder.
+  Defaults to bold
+Unhomed Color::
+  The text color when unhomed pecified as a Gdk.RGBA color. Defaults to red,
+  Gdk.RGBA(red=1.000000, green=0.000000, blue=0.000000, alpha=1.000000)
+Homed Color::
+  The text color when homed pecified as a Gdk.RGBA color.   Defaults to green,
+  Gdk.RGBA(red=0.000000, green=0.501961, blue=0.000000, alpha=1.000000)
```
hansu, line 1542: Wouldn't it be easier to use if we have here the "real" property names here like `font_family` and so on?
I know it's not really consequently used here, but e.g. in Combi_DRO.

lib/python/gladevcp/drowidget.py
```diff
@@ -41,46 +52,72 @@ class HAL_DRO(Gtk.Label):
...
+        'homed_color' : (Gdk.RGBA.__gtype__, 'Homed Color', 'The color of the DRO text when homed',
+                    GObject.ParamFlags.READWRITE),
     }
     __gproperties = __gproperties__
 
     def __init__(self, *a, **kw):
+        self.css = Gtk.CssProvider()
+        self.css_text = f"""label {{
+    font-family: sans;
+    font-size: 40px;
```
hansu, line 82: This feels a bit big as default value.

hansu (commented):
Looks good to me, only some little comments. And I guess the failed build has nothing to do with these changes...

lib/python/gladevcp/drowidget.py
phillc54: The inspector may also be opened at run time so I should mention that as well.

I am not quite sure what you mean with the property names, I named them so they would be consistant with all the other properties. The CSS names are the standard Gtk3 property names.

What font-size do you suggest?

hansu:
>The inspector may also be opened at run time so I should mention that as well.

I understood it already this way.

>I am not quite sure what you mean with the property names, I named them so they would be consistant with all the other properties.

I meant it like this:

```diff
-Font Family::
+font_family::
  Specify the font family e.g. mono. Defaults to sans.
  If the font does not exist then the current system font will be used.
```
Because this is the name one have to know if you want to use this property.

>What font-size do you suggest?

Maybe something between 20 and 30? But if you think 40 is a good default value, then it's also fine for me :wink:

phillc54:
I see what you mean about the doc properties. There were some others that were similar so I did them as well.

Font size default is 26

hansu:
Thanks. I am only a bit unsure whether to merge it now or not because all the builds on the build bot seem so fail currently...

PR #2409, opened 2023-03-27
While iterating through translations (smoe:docs29_misc_17)

smoe:


src/hal/components/spindle.comp
```diff
@@ -111,11 +111,13 @@ Set the minimum deceleration. If you do not have a spindle encoder this is in RP
...
 .TP
-\\fBspindle.\\fP\\fIN\\fB.speed-tolerance.\\fPx float in Tolerance for 'at-speed' signal (in RPM).
+\\fBspindle.\\fP\\fIN\\fB.speed-tolerance.\\fPx float in
+Tolerance for 'at-speed' signal (in RPM).
 Actual spindle speeds within this amount of the commanded speed will cause the at-speed signal to go TRUE.
 
 .TP
-\\fBspindle.\\fP\\fIN\\fB.zero-tolerance.\\fPx float in Tolerance for 'zero-speed' signal (in RPM).
+\\fBspindle.\\fP\\fIN\\fB.zero-tolerance.\\fPx float in
+Tolerance for 'zero-speed' signal (in RPM).
```
SebKuzminsky, line 120: Why are you reintroducing the newlines here?  I thought po4a preferred no newline in paragraphs.

SebKuzminsky (approved):

SebKuzminsky:
The build failure is independent of this PR and will be fixed by #2422.

src/hal/components/spindle.comp
smoe: Somehow the rendering got lost. "float in" is the type/direction, "Tolerance for.." the description, and it appears all together in a weird way that was somehow difficult to deconvolute by eye.

docs/src/hal/parallel-port.adoc
```diff
@@ -189,48 +190,48 @@ loadrt hal_parport cfg="0x378 0xc000"
...
+so if you like to 'get under the hood' and re-arrange things, be sure to check these values before you start LinuxCNC.
 
 == Pins
 
-* 'parport.<p>.pin-<n>-out' (bit) Drives a physical output pin.
-* 'parport.<p>.pin-<n>-in' (bit) Tracks a physical input pin.
-* 'parport.<p>.pin-<n>-in-not' (bit) Tracks a physical input pin, but inverted.
+* `parport.<p>.pin-`__<n>__`-out` (bit) Drives a physical output pin.
+* `parport.<p>.pin-`__<n>__`-in` (bit) Tracks a physical input pin.
+* `parport.<p>.pin-`__<n>__`-in-not` (bit) Tracks a physical input pin, but inverted.
```
hansu, line 199: All the formatting like this is not compatible with our AsciiDoc version (asciidoc-py). It produces wrong and misleading output.

Sorry that I didn't review this PR earlier, but I was a bit unmotivated and a bit tired of always pointng out the same error.

@smoe can you prepare a PR to revert the ``` `.*`__<.*>__`.*` ``` changes?
It is in parallel-port.adoc and rtcomps.adoc - 68 occurences now.

Nevertheless you did a great job on the German translation. I was positively surprised when I picked a page for a German guy from the forum. And you also got some positiv feedback there: https://forum.linuxcnc.org/42-deutsch/48623-linux-cnc-in-oberoesterreich?start=30#268417 😉

hansu: The only thing that is working is something like `+parport._<p>_.pin-_<n>_-out+` but not compatible with future AsciiDoc versions, so I wouldn't recommend that.

hansu: Or just leave it as it is to avoid double work for some time in the future when we hopefully have migrated to AsciiDoc?

Issue #2386, opened 2023-03-08
command data safety between milltask and motion


devinor:
The send commond code is as follows. In line 95, copy the struct into sharemem
https://github.com/LinuxCNC/linuxcnc/blob/27546d2192c157a021ee4eed48e617aaabadc34d/src/emc/motion/usrmotintf.cc#L74-L117

The recive command code is as follows. Determine the data in the shared memory, if head is not equal tail, means has new command.

motion has just executed to 393 lines of code, because there is no new message, so head == tail, continue to execute, if at this time milltask executes to usrmotintf.cc::95 lines of code and assigns a value, will it cause an error in command.c::397 lines of code?

Motion may think that new data has come, and it is complete data, which will cause the following command to execute incorrectly

https://github.com/LinuxCNC/linuxcnc/blob/348b0d196d9fad662472589cc9882985732796d1/src/emc/motion/command.c#L381-L420

SebKuzminsky:
> The recive command code is as follows. Determine the data in the shared memory, if head is not equal tail, means has new command.

No, that's not what the `head` & `tail` mean.  Each command has a unique `head` and `tail`, and they're the same.  If `head` and `tail` are equal it means a single, complete command is in the `emcmotCommand` buffer.  If Motion sees `head` and `tail` with different values, it means Motion checked while Task was busy updating the command.  In this case Motion will not process the partial command, and will instead wait, and check back later to see if Task has finished copying it.

I think the communication between Task and Motion is safe, at least from the bug you're chasing here.

Here's how I think this code flows:
1. Task calls `usrmotWriteEmcmotCommand` with a new `emcmot_command_t` in `c`.
2. `usrmotWriteEmcmotCommand` gives the new command a unique `head` and matching `tail`, and a unique `commandNum` (usrmotintf.c lines 85-86).
3. `usrmotWriteEmcmotCommand` copies the new command into the memory shared with Motion (usrmotintf.c line 95).
4. `usrmotWriteEmcmotCommand` polls the status buffer (also shared with Motion), waiting for this new command's `commandNum` to be acknowledged by Motion (usrmotintf.c lines 100-102).

Meanwhile, Motion does this:
1. Make sure there's a complete, consistent command in the shared buffer (command.c line 393).
2. Make sure it's a new command, i.e. one we haven't handled and acknowledged already (command.c line 397).
3. Begin but _do not finish_ acknowledging the new command (command.c line 404).
[...]

PR #2412, opened 2023-03-30
fix command safety between milltask and motion

devinor:
fix bug #2386


c-morley:
Can you explain a consequence of this bug?

SebKuzminsky:
This patch duplicates the check on line 397.  The additional check is not needed, see comments in #2386.

PR #2415, opened 2023-03-31
Merge 2.9 into master

hansu:
A conflict in drowidget.py which I am not sure how to resolve correctly.

@phillc54, @c-morley  can you check how this can be resolved best?

```
<<<<<<< HEAD

        self.css_text = """
                        .background  {background-color: #000000;}
                        .labelcolor  {color: #FF0000;}
                        """

        self.css = Gtk.CssProvider()
        self.css.load_from_data(bytes(self.css_text, 'utf-8'))
        self.get_style_context().add_provider(self.css,Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION)
        self.get_style_context().add_class('background')
        self.get_style_context().add_class('labelcolor')

        GLib.timeout_add(100, self.periodic)

=======
>>>>>>> 2.9
```


phillc54:
I am not sure that I did this correctly but I think it is now OK

phillc54:
Actually after getting an email saying that I merged master into 2.9 I am sure I have done something wrong but I am not sure what?

c-morley:
Yep - want me to revert it? I think I figured out how to do it.

phillc54:
Yes please, I am afraid that I may break it more...

c-morley:
2.9 should be reverted now - please check

c-morley:
for drowidget, which version is the latest 2.9 or master?

phillc54:
Thank Chris, It seems good now, I owe you a beer. :)

2.9 drowidget is the last I worked on.

c-morley:
ok I merged 2.9 to master - cross my fingers - I think we are ok.

phillc54:
Thanks again, I thought that I would be clever and fix the conflict from this web page but it didn't quite go to plan. :(

c-morley:
Glad I could help.

hansu:
Thank you guys.
Only weird that the merge commit isnt shown here https://github.com/LinuxCNC/linuxcnc/commits/master

c-morley:
I have no idea why.

phillc54:
It shows up here, under "Commits on Mar 26, 2023"

SebKuzminsky:
We now have all of master's history in 2.9 (even though the bogus merge commit was later reverted).  This should *never* happen, and it's hugely problematic IMO.

I was away from my computer this past week or I would have maybe noticed and tried to intervene sooner.

To fix history now we'd have to:
* Roll 2.9 back to c9ef3d85a8 (the last commit before b3e6b494 "Merge branch 'master' into 2.9" on Apr 1), and force-push
* Roll master back to 7ba312e3d8 (the last commit before the the bogus merge in 2.9 got merged back into master), and force-push
* Identify and re-apply all the correct changes in 2.9 and master that will have been lost in the roll-backs
* Finally merge the fixed 2.9 into the fixed master.

This is the second time in the past 3 months that our history has gotten messed up by an erroneous merge.  Here is the discussion we had last time:  https://sourceforge.net/p/emc/mailman/emc-developers/thread/tpc52q%24oro%241%40ciao.gmane.io/#msg37758545

@jepler and I advocated then for enabling github's "branch protection" feature on our important long-lived branches (2.8, 2.9, and master).  This would mean that all changes to these branches would go through the PR process, so at least two people would have to make a mistake before history got corrupted, instead of just one like now.

The proposal was rejected then, with the (totally fair) objection that it would introduce too much overhead in our development process, but I raise the same proposal again.  Fixing these kinds of problems in "emergency mode", after the fact, is worse than checking our incoming changes by hand an having slower PRs as a result.

PR #2419, opened 2023-04-01
axis: copy gcode to temporary file before opening (fixes #2418)

### LINKED ISSUE #2418: "Run from here" reloads the gcode file but doesn't update the display

 1. Start a job
 2. Edit the coordinates in the gcode file (e.g. add 100 to all the Y values) but do not reload the file in the Axis gui
 3. Hit the emergency stop
 4. Take the machine out of emergency stop
 5. Use "run from here" to resume your job
 6. The machine cuts the updated coordinates in the gcode file, even though the old coordinates are still shown in the gcode display

I expected the machine to continue with the coordinates shown in the gcode display (i.e. the line of gcode that I actually clicked "run from here" on).

The machine rapid'd over 100mm and broke my tool.

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`): Debian 10 buster
 * I am using this kernel version (shown by `uname -a`): `Linux linuxcnc 4.19.0-17-rt-amd64 #1 SMP PREEMPT RT Debian 4.19.194-2 (2021-06-21) x86_64 GNU/Linux`
 * I am running ...
   * [X] A binary version from linuxcnc.org (including buildbot.linuxcnc.org)
   * [ ] A binary I built myself
 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`): 1:2.8.2.11.g6a3d0a434
 * I am using this user interface (GUI) (e.g., AXIS, Touchy, gmoccapy, etc): AXIS
 * I am using this interface hardware vendor and chipset (e.g., parallel port, ethernet port, FPGA card): Parallel port

jes:
The issue here is that if you open a gcode file, and then change the file on disk, you'll think your changes are completely inert because they don't show up in the GUI, but actually LinuxCNC has its own separate file handle behind-the-scenes, and your machine can do scary and unpredictable things.

With this fix, the code being executed by LinuxCNC always matches what is displayed in the AXIS GUI even if you subsequently change the file.

### DIFF
--- a/src/emc/usr_intf/axis/scripts/axis.py
+++ b/src/emc/usr_intf/axis/scripts/axis.py
@@ -1183,7 +1183,7 @@ def open_file_guts(f, filtered=False, addrecent=True):
         loaded_file = f
         program_filter = get_filter(f)
         if program_filter:
-            tempfile = os.path.join(tempdir, os.path.basename(f))
+            tempfile = os.path.join(tempdir, "filtered-" + os.path.basename(f))
             exitcode, stderr = filter_program(program_filter, f, tempfile)
             if exitcode:
                 root_window.tk.call("nf_dialog", (".error", "-ext", stderr),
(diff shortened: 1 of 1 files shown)


jes:
Since the "filter program" stuff already copied to a temporary file named by the loaded filename, I made that copy to "filtered-"+f instead. Maybe there's a better approach, I'm not sure whether the filename passed to LinuxCNC actually matters or not (e.g. will a user doing more complex LinuxCNC usage be surprised that their filtered file has got "filtered-" prepended? Can they even find out?).

jes:
I think the CI failure is unrelated so I'm going to ignore it, please let me know if that's not right.

SebKuzminsky:
This PR needs to be rebased onto the fixed master branch.  Something like `git fetch && git rebase --onto origin/master 85305d78b77c06c855a658aef39de2ab91ee3ee2` should do it.  You may want to wait until https://github.com/LinuxCNC/linuxcnc/pull/2422 (or something like it) gets merged up to master, to fix the CI build problem.

jes:
OK, no problem. Are you happy with the change in principle then?

jes:
Thanks!

PR #2421, opened 2023-04-05
Halscope channel selection

hansu:


smoe:
Have not checked the correctness of the code but the behavior is like what I would have expected halscope to work :)

hansu:
> Have not checked the correctness of the code but the behavior is like what I would have expected halscope to work :)

You had to turn of the channel with a separate button in the frame "Verical". See http://linuxcnc.org/docs/2.9/html/hal/tutorial.html#_more_channels

smoe:
> > Have not checked the correctness of the code but the behavior is like what I would have expected halscope to work :)
> 
> You had to turn of the channel with a separate button in the frame "Verical". See http://linuxcnc.org/docs/2.9/html/hal/tutorial.html#_more_channels

I prefer it this way.

src/hal/utils/scope_vert.c
```diff
@@ -803,62 +806,67 @@ static void chan_sel_button(GtkWidget * widget, gpointer gdata)
...
+            }
+        }
+        vert->chan_enabled[chan_num - 1] = 1;
+        /* make chan_num the selected channel */
+        vert->selected = chan_num;
+        channel_changed();
+    } else if (vert->selected == chan_num) {
+    /* a click on an already active channel turns it off */
+        set_channel_off(chan_num);
+        ignore_click = 0;
```
SebKuzminsky, line 854: The whitespace changes at the beginnings of these lines (and elsewhere in this branch) make the diff look larger than it really is.  It's not a big deal, but please try to avoid spurious whitespace changes like these.

The original code (before this branch) had a random mix of spaces and tabs, which is not ideal.  Our coding documentation, scattered and incomplete as it is, is silent on the spaces/tabs issue but requests to [avoid whitespace changes where possible](http://linuxcnc.org/docs/devel/html/code/contributing-to-linuxcnc.html#_follow_the_style_of_the_surrounding_code).

I fetched your branch and looked at it with `git log -p --ignore-space-change` to make it easier to see the meaningful changes.

src/hal/utils/scope_vert.c
```diff
@@ -647,7 +647,7 @@ static void init_vert_info_window(void)
     /* turn off the channel if button is clicked */
     g_signal_connect(button, "clicked",
 	G_CALLBACK(channel_off_button), NULL);
-    gtk_widget_show(button);
+    // gtk_widget_show(button);
```
SebKuzminsky, line 650: Probably better to just delete this line, and the ones above it where the "Chan Off" button is created.

SebKuzminsky (changes requested):
I like this new behavior, thanks @hansu.

I vaguely remember hearing that the halscope GUI was based on some particular real oscilloscope that the author liked, but this new behavior feels normal and modern to me.

hansu:
>The whitespace changes at the beginnings of these lines (and elsewhere in this branch) make the diff look larger than it really is. It's not a big deal, but please try to avoid spurious whitespace changes like these.

My strategy was to fix the whitespaces (mix of tabs and spaces, wrong indentation) at the locations where I made changes.
Because the code was sometimes a bit hard to read because of the wrong indentation.
I can try to reduce the whitespace changes a bit, but I don't really want to keep the wrong indentation in the functions I touched.

SebKuzminsky:
> My strategy was to fix the whitespaces (mix of tabs and spaces, wrong indentation) at the locations where I made changes.
> Because the code was sometimes a bit hard to read because of the wrong indentation.
> I can try to reduce the whitespace changes a bit, but I don't really want to keep the wrong indentation in the functions I touched.

I've also come across this problem in the LinuxCNC sources, and I think you're right to fix the indentation errors when you're "in the neighborhood", *particularly* when the indentation is misleading and confusing.

So I retract my complaint, and only ask that, in the future, when possible, please make white-space/indentation changes in a separate commit from the commit where you're actually changing the behavior of the code.

src/hal/utils/scope_vert.c
SebKuzminsky: It's probably better to have just one way to turn off a channel (by clicking on the enabled channel button).  If someone complains persuasively about the change we can always resurrect the old "Chan Off" button code out of git.

hansu: Okay, I removed the button code.

src/hal/utils/scope_vert.c
```diff
@@ -836,29 +836,29 @@ static void chan_sel_button(GtkWidget * widget, gpointer gdata)
...
+        if (chan->name == NULL) {
+        /* need to assign a source */
+            if (dialog_select_source(chan_num) != TRUE) {
+            /* user failed to assign a source */
+            /* force the button to pop back out */
+                ignore_click = 1;
+                gtk_toggle_button_set_active(GTK_TOGGLE_BUTTON(widget), FALSE);
+                return;
+            }
+        }
```
hansu, line 836: I am only a bit unsure about how to indent comments like these best.
IMHO they belong over the if statement as they describe the whole body, but I wanted to keep changing only whitespaces.
But this way it looks also a bit wrong.

But indenting this 'correctly' would associate it with the next if statement:
```diff
         if (chan->name == NULL) {
-        /* need to assign a source */
+            /* need to assign a source */
             if (dialog_select_source(chan_num) != TRUE) {
```

SebKuzminsky:
I just noticed that halscope lets me trigger on a channel that's disabled, that seems wrong:
Screenshot from 2023-04-06 15-30-54

This behavior is also present in 2.9 (without this branch), it's not something that was introduced here.

SebKuzminsky (approved):
Good stuff!

src/hal/utils/scope_vert.c
SebKuzminsky: It's a subjective question of aesthetics, but my preference is that comments in the body of an if statement should be indented like the rest of the contents of the body, and should explain what the situation is now that we've entered into the body of the "if".

If a comment explains what the "if" is checking, then it should go above the "if" (and should start on the same column as the "if"), and probably end with a question mark for clarity.

And yes, i agree that makes the situation above somewhat ambiguous: does the comment explain the context inside the body, or does it explain the check that the next "if" is about to perform?  You just have to read it and figure it out!

hansu: Okay it was the first one. I fixed that. But now is a new approval needed, sorry :D

hansu:
> I just noticed that halscope lets me trigger on a channel that's disabled, that seems wrong:
> 
> This behavior is also present in 2.9 (without this branch), it's not something that was introduced here.

No, I think this makes absolutely sense in some cases. Imagine you have a signal you want to trigger on but don't want to mess up your display with too many signals. This is equivalent to use the external trigger input of a HW oscilloscope.

hansu:
I also wonder if we can drop the highlighting of the selected signal graphs itself now.
The selection can now be easily seen on the channel buttons, so there is no need to do that on the graphs as well. 
Especially for the red one the highlighting changes the color so much that it might be mixed up with another channel.

With graph highlighting (current):
halscope_selected_color

Without graph highlighting:
halscope_selected_color_new


Furthermore I wonder why only the first channel has a baseline (the white one at the end of the second GIF) but not the other channels.

hansu:
@SebKuzminsky may I ask you for a further approval? The further indentation changes made a new approval required.

SebKuzminsky (approved):
This looks great!

PR #2424, opened 2023-04-06
Fix task motion race

SebKuzminsky:
This fixes a race condition in communications between Task and Motion, identified by @devinor in #2386.

It introduces a mutex to synchronize the command buffer between Task and Motion.

Task (which is non-realtime) takes the mutex to lock the buffer while it's writing the new command to the shared memory buffer.

Motion (which is realtime) uses trylock so it won't block if Task has the command buffer locked.  Instead it waits until the next invocation of `motion-command-handler`, when it checks again to see if Task is done copying and has unlocked the mutex.  I believe this "try again later" behavior is the same as the intent of the old racy `head`/`tail` synchronization code (which probably worked fine on single-core machines, back at the dawn of time, but falls apart on "modern" 1990s-era hardware).

### DIFF
--- a/src/emc/motion/command.c
+++ b/src/emc/motion/command.c
@@ -417,19 +417,25 @@ void emcmotCommandHandler(void *arg, long servo_period)
     int abort = 0;
     char* emsg = "";
 
+    emcmot_command_t c;
+
     /* check for split read */
     if (emcmotCommand->head != emcmotCommand->tail) {
 	emcmotDebug->split++;
 	return;			/* not really an error */
     }
-    if (emcmotCommand->commandNum != emcmotStatus->commandNumEcho) {
+
+    // Make a local copy of the command.
+    c = *emcmotCommand;
+
+    if (c.commandNum != emcmotStatus->commandNumEcho) {
 	/* increment head count-- we'll be modifying emcmotStatus */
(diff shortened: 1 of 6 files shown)


andypugh (approved):
I read the conversation between yourself and devinor. I am curious why such a potentially serious problem hasn't been seen to cause problems, but I also think that between you you have fixed it.

src/emc/motion/command.c
```diff
@@ -417,19 +417,26 @@ void emcmotCommandHandler(void *arg, long servo_period)
...
+
+    if (rtapi_mutex_try(&emcmotStruct->command_mutex) != 0) {
+        // Failed to take the mutex, because it is held by Task.
+        // This means Task is in the process of updating the command.
+        // Give up for now, and try again on the next invocation.
+        return;
     }
-    if (emcmotCommand->commandNum != emcmotStatus->commandNumEcho) {
+    // Make a local copy of the command and unlock the command buffer.
+    c = *emcmotCommand;
```
devinor, line 429: Hi SebKuzminsky. I have a question.
if we use mutex, Is memory copy necessary?

devinor:
Yes, code maintainability is also important.
I pull a new request #2436

devinor (approved):

PR #2428, opened 2023-04-08
Gladevcp: added missing icons + added symlinks for default icons

hansu:
This is kind of a follow up to #2396.
I created some icons which are missing, fixed some sizes and for those I found suitable icons from glade, I created symbolic links to them in the debian *.links file.
Furthermore I moved the icons to `share` to match the installation path (to be found easier in this repo).

This removes almost all XXX needs an icon named 'widget-xxxxx' when starting Glade when LinuxCNC is installed on Bookworm.


hansu:
Now also fixed for run-in-place. @phillc54 can you please check if it works for you as well?
In the last commit I added symbolic links to gtk icons. The question is if we should keep that or create the links in the rip-environment-script as they are only needed for rip. for the installation there is the .links file.

phillc54 (approved):
Nice...
It works great on my RIP install.

PR #2429, opened 2023-04-09
hm2_pktUART: Various bugfixes and additional queued reads and writes

andypugh:
Fix bugs in setup function (masking out important data)
Queued reads and writes are much better suited to operation with Ethernet and parallel port devices.
Also add the Rx and Tx status registers to the regular TRAM reads to be immediately available.
Document the changes and delete the old (too terse) docs.

(Note to reviewers: As far as I know almost no-one is using this at the moment,  but these changes are needed for the new Modbus driver)


andypugh:
Well, I got bored waiting for approval, and found that I can still push from the command line.

Issue #713, opened 2020-04-08
multi spindle discussion


c-morley:
Current behavior AFAIKT with a bit of testing.

M3/M4 will start all spindles as long as their respective S codes are already set, adding a $ with a number will start a specific spindle.
M5 stops all spindles - adding a $ with a number specifies which spindle to stop
After an s code has used a $ to specify a spindle, plain s codes after that will only control that specified spindle.
You cannot put a $ with a number with out another spindle control code -ie you can't preset a specific spindle to be modal in this way.

My thoughts - Mostly copied from dev email list...

Setting the s code to a particular spindle (with %) will from then on always controls that spindle with a plain s code. Not sure if that is on purpose but i wouldn't think so.

If you don't set the s code with $ then only spindle 0 will start and stop with m3 m4 m5.
meaning a program running on a single spindle machine runs the same on a dual spindle machine.
But that relying on not setting the $ code seems a bit risky considering linuxcnc can be remapped and use subroutines (that could set it) or maybe left over from a manual setting (I didn't try that)

I did find that using the python module to control the spindle (like all the python screens do) the stop command seems to assume spindle 0 so pressing stop button only stops spindle 0. Which may or may not be whats expected but there is no python command to stop all spindles, unless you added one I didn't check.

I still feel that with mcode and python module, specifying all spindles with say -1 is a better practice.
We should fix the python module to be able to specify all spindles for stopping.

I feel that with s code, it's probably a bug that it remembers the last specified spindle and continues to adjust it with a plain s code.
It would be good practice and consistent if specifying $-1 would set the spindle speed for all spindles.

andypugh:
Making S modal was deliberate, but on balance I think it was a mistake. 

It is probably also true that if someone really needed to set up a group of spindles and start / stop them together then they could be expected to read the manual to find out how. 

statetags and various Python interfaces do not, at the moment, report on more than the default spindle. The overlap between those using these interfaces and those using multiple spindles is probably small.

andypugh:
M19 stops all spindles regardless of $ number. Should it?

c-morley:
m19 - I wouldn't think so.
Yes I thought maybe it was modal so tried $1 on a line by it's self and that was not accepted. I agree I don't think that is the way to go.

Qtvcp now has support for multiple spindles if some one wants to add them.

I just think it's important to iron out the details before releasing so we don;t have to break behavior on the next release.

andypugh:
[...]

PR #2437, opened 2023-04-13
Allowing Default Spindle to be selected

satiowadahc:
Adds a setting to INI files to allow changes to M3,M4,M5

Feedback would be appreciated on if this should also affect M19, M51, G33, G74, G76, G84, G85, G86, G93-G97

Motivation: Mill with multiple spindles, our machinists did not like M5 didn't stop all spindles. I see the conversation here was to make a default value, #713. But this allows default behavior to be be changed. To comply with that setting the default should be 0.

### DIFF
--- a/docs/src/config/ini-config.adoc
+++ b/docs/src/config/ini-config.adoc
@@ -508,6 +508,7 @@ if __name__ == "__main__":
   Specifies a list of colon (:) separated directories for user defined functions.
   Directories are specified relative to the current directory for the INI file or as absolute paths.
   The list must contain no intervening whitespace.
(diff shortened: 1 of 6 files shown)


satiowadahc:
Now I submitted this, should the setting per code be available or is that too much configuration? IE adding a `DEFAULT_M3_M4_SPINDLE` `DEFAULT_M5_SPINDLE` `DEFAULT_M19_SPINDLE`...

satiowadahc:
https://github.com/LinuxCNC/linuxcnc/blame/master/src/emc/rs274ngc/interp_convert.cc#L4016

Defaults to 0

andypugh:
Yes, I would prefer it to default to "all" too. I wonder if it is too late to change it?

rene-dev:
closed by #2438

PR #2438, opened 2023-04-13
RS274: M5 default to all spindles.

satiowadahc:
Default M5 to stop all spindles. Replaces: #2437


satiowadahc:
@andypugh better?

rene-dev:

PR #2441, opened 2023-04-16
Remove gladevcp symlink creation

### LINKED ISSUE #2440: gladevcp icon symlinks are rejected by lintian
GladeVCP installs symlinks for its icons that point to icons that are not part of linuxcnc, but part of libgladeui-common (which is a dependency of linuxcnc).  This violates lintian:
* http://buildbot.linuxcnc.org/buildbot/builders/3304.dsc-buster-rtpreempt/builds/2827/steps/shell/logs/stdio
* http://buildbot2.highlab.com/buildbot/#/builders/6/builds/665/steps/16/logs/stdio

Because lintian fails, we do not publish debs.

I ask that the GladeVCP maintainers (@hansu et al) find and implement a solution for this.  The first thing that comes to mind is to just  usethe libgladeui-common icons directly, instead of creating symlinks to them and using the symlinks, but I defer to your judgement here.

hansu:
Fixes #2440

This removes the links from the .links file to not create symlinks with `dh_link`.
For some of them it worked to set alternative icons in the glade catalogue file, but not for all.
So for the other ones I created symlinks which are just copied by the makefile. 
If you think this also triggers the linter, I can simply drop the last commit.


hansu:
Hmm I have no idea why this happens:
`install: cannot stat '../share/glade/pixmaps/hicolor/16x16/actions/widget-gladevcp-hal_button.png': No such file or directory`

hansu:
@SebKuzminsky I think  it's good to go now.

SebKuzminsky:
The source package, which isn't being tested by our github CI but *is* being tested by the buildbot, still fails:

http://buildbot2.highlab.com/buildbot/#/builders/18/builds/696

```
E: linuxcnc source: source-contains-unsafe-symlink share/glade/pixmaps/hicolor/16x16/actions/widget-gladevcp-hal_button.png
N: 
N:    The source contains an unsafe symlink. If followed, the link will escape
N:    the source root. Note that all absolute symlinks are unconditionally
N:    considered "unsafe" in this case (unlike in binary packages).
N:    
N:    If it is a part of the package's testsuite, Lintian may have failed to
N:    recognise it as a test. In that case, please override the tag.
N:    
N:    Severity: serious, Certainty: possible
N:    
N:    Check: cruft, Type: source
```
(And many more files in the same directory also trigger this lintian source warning.)

hansu:
Okay, I didn't know that even the existence of those symbolic links leads to linter errors.
Let's give it a further try now ;-)

hansu:
Looks better now: http://buildbot2.highlab.com/buildbot/#/grid?branch=refs%2Fpull%2F2441%2Fmerge

SebKuzminsky:
Thanks for figuring this one out @hansu !

Issue #2435, opened 2023-04-12
Wrong axis direction with 5 axis kinematics, e.g. with xyzbc-trt kinematics


DerAndere1:
# The issue tracker is not a support forum

## Here are the steps I follow to reproduce the issue:

 1. See source code at https://github.com/LinuxCNC/linuxcnc/blob/76b870db31a541a4c4e6b1e1deb02724b5c51961/src/emc/kinematics/maxkins.c#L63-L94 and https://github.com/LinuxCNC/linuxcnc/blob/39809fc03e57e72db6c2bc8916958e38ce49fed5/src/emc/kinematics/trtfuncs.c#L297-L349
 3. Compare with documentation at https://linuxcnc.org/docs/html/gcode/machining-center.html#_rotational_axes
 4. Observe that 
 - with maxkins, axis B moves in wrong direction.
 - With xyzbc-trt kinematics, axes B and C move in the wrong direction. 
 - With xyzac-trt kinematics, the A axis is affected. 
 - With 5axiskins, the B axis is affected.

## This is what I expected to happen:

inverse kinematics should ensure that rotational axes move CCW relative to the work piece when moving in positive direction when viewed from positive. For a tilting/rotating table, this means the rotational joint of the table must move clockwise when viewed from positive. 

These are some expected joint positions calculated from given cartesian positions using fixed xyzbc-trt inverse kinematics:

|Cartesian position XYZBC| Joint position XYZBC|
|--- |--- |
|(1, 0, 2, 0, 0) | (1, 0, 2, 0, 0) |
|(1, 0, 2, 90, 0) | (-2, 0, 1, 90, 0) |
|(1, 0, 2, 0, 90) | (0, -1, 2, 0, 90) |

## This is what happened instead:

e.g. with xyzbc-trt kinematics, axes B and C move in wrong direction.

These are some results using current xyzbcKinematicsInverse() from trtfuncs.c:

|Cartesian position (X,Y,Z,B,C) | Joint position (X,Y,Z,B,C) |
|--- |--- |
|(1, 0, 2, 0, 0) | (1, 0, 2, 0, 0) |
|(1, 0, 2, 90, 0) | (2, 0, -1, 90, 0) |
|(1, 0, 2, 0, 90) | (0, 1, 2, 0, 90) |

## It worked properly before this:

Never

## Information about my hardware and software:

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`): None
[...]

PR #2442, opened 2023-04-17
Add note about axis direction and cite original source of 5-axis-kinematics docs

DerAndere1:
in 5-axis-kinematics docs:
-Add note that with the described kinematics, rotational axes move in the opposite direction of conventional kinematics. (https://github.com/LinuxCNC/linuxcnc/issues/2435)

- Cite the original source of the text

### DIFF
--- a/docs/src/motion/5-axis-kinematics.adoc
+++ b/docs/src/motion/5-axis-kinematics.adoc
@@ -20,6 +20,8 @@ Coordinated multi-axis CNC machine tools controlled with LinuxCNC, require a spe
 
 The kinematics components are given as well as vismach simulation models to demonstrate their behaviour on a computer screen. Examples of HAL file data are also given.
 
+Note that with these kinematics, the rotational axes move in the opposite direction of what the convention is. See section ["rotational axes"](https://linuxcnc.org/docs/html/gcode/machining-center.html#_rotational_axes) for details.
+
 == 5-Axis Machine Tool Configurations
 
 In this section we deal with the typical 5-axis milling or router machines with five joints or degrees-of-freedom which are controlled in coordinated moves.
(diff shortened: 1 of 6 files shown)


andypugh (commented):
Are you saying that _all_ the kinematics are reversed now? 
Have you referenced https://linuxcnc.org/docs/stable/html/motion/5-axis-kinematics.html ?
I am starting to think that perhaps the error is in the document that you are referencing, if there is an error. 
Have you considered "Clockwise or counterclockwise is from the point of view of the workpiece."

DerAndere1:
> Are you saying that all the kinematics are reversed now?
Yes, I currently believe that all 5 axis kinematics have certain axes inverted. See updated text in issue https://github.com/LinuxCNC/linuxcnc/issues/2435
> Have you referenced https://linuxcnc.org/docs/stable/html/motion/5-axis-kinematics.html ?
That document and its figures assume wrong axis directions.
> Have you considered "Clockwise or counterclockwise is from the point of view of the workpiece."
Yes this is what I consider conventional axis direction. But https://linuxcnc.org/docs/stable/html/motion/5-axis-kinematics.html and existing kinematics in LinuxCNC do not conform.

Here are youtube-videos showing LinuxCNC in action with different kinematics and their wrong axis directions:

5axiskins (wrong B direction): https://www.youtube.com/watch?v=65Wd6ZGoWys&themeRefresh=1
xyzac-trt (wrong A and C direction); https://youtu.be/BevZm46cTQw?t=347

Conventional axis direction (e.g. Haas Automation): https://youtu.be/HxPjH4v5iEg?t=213

SebKuzminsky:
Merged into 2.9 here: https://github.com/LinuxCNC/linuxcnc/commit/a2b692dcc62fd9e33b597e366ec0f64e76912072

Thanks @DerAndere1 !

PR #2443, opened 2023-04-17
Fix rotational axis direction with some 5 axis kinematics

DerAndere1:
Fixes some kinematics where the direction of rotational axes does not match the conventions (https://github.com/LinuxCNC/linuxcnc/issues/2435)

It was suggested that new kinematics modules should be created instead of fixing existing modules. I propose these fixes instead because I think it is safe to assume that users test their machines carefully upon updating LinuxCNC with 5 axis kinematics. It is however not safe to assume, that G codes written with TCP control are checked carefully before being run. The whole purpose of TCP control is that the CAM software can output "machine-independent" G-code. Therefore, behaviour of LinuxCNC kinematics should follow the conventions and their own documentation.


andypugh:
Maxkins has been in LinuxCNC for a very long time. It is hard to know how many people are using it, maybe not very many. 
But here it is, doing what the original author intended, 15 years ago: https://youtu.be/0jQy-zc5P6w
It's a really bad idea to break existing configs. But I fully support your other pull request to note that it behaves unexpectedly.

DerAndere1:
I am totally open for alternative solutions. This PR is more to document the required changes. At the moment I do not have much time and I am not a LinuxCNC user. So I leave this as an exercise for later. I assume the following steps need to be followed: 
- testing the changes in hardware 
- check for additional bugs (I am confused by the `dz = dz + dt` in trtfuncs.c. In my mind, tool length offset should be handled differently)
- register a new kinematics name, e.g. xyzac-trt2, xyzbc-trt2 and maxkins2 
- add new files

andypugh:
I think for most purposes the xxxx-trt kins are what people should be using. Maxkins was for a particular machine a long time ago when kinematics was new, and it was created mainly as a demonstration.

jepler:
I don't have an opinion but I think the 5 axis max has not existed for a long time, it was cradek 's modified desktop mill if memory serves

andypugh:
Have you tested this kinematics on a similar machine? 

I wonder if it might be a better idea to add a deprecation notice to the Maxkins manual page suggesting that the rotation directions may be incorrect, and suggesting the use of one of the alternative kinematics?

DerAndere1:
I adaptet the maxkins and xyzbc-trt kinematics and incorporated them in other Firmware (Marlin) for testing. The problem is that not only maxkins (xyzbc with swivel head and rotary table) but also xyzbc-trt inverse kinematics are affected. Tests of forward kinematics are pending. As long as there is no corrected version of these kinematics, I would not deprecate them.

andypugh:
What are you basing your opinion of the correct directions on? There is certainly at least one image on the internet on the subject which is incorrect. Looking from the negative side of an axis towards the positive side, it is my understanding that clockwise rotation is positive rotation.  
https://www.evl.uic.edu/ralph/508S98/coordinates.html
Furthermore I believe that CNC machines use a right-handed coordinate system.

DerAndere1:
your two remarks are correct.  The problem is that maxkins and 5axiskins do not respect this for the B axis. 

In addition, each rotary axis (as opposed to joint) defines the orientation of the tool when viewed from the perspective of the workpiece (lets call this "statement 3"). xyzac-trt and xyzbc-trt kinematics get it wrong for the A, B and C axis, probably because they did not account for statement 3. Statement 3 means that  `G1 C-90` should rotate the tool in CCW direction relative to the workpiece (when viewed from negative Z). To achieve this with a tilting rotary table, the table has to rotate in CW direction around a shaft that is parallel to the Z axis (when viewed from negative Z toward positive Z).

Source: NIST RS274 NGC Interpreter, Version 3. 

This was discussed in issue https://github.com/LinuxCNC/linuxcnc/issues/2435. More details including confirmation by reviewers and video proof can be found in the discussion at PR https://github.com/LinuxCNC/linuxcnc/pull/2442

andypugh:
The xyzac-trt and xyzbc-trt kins files were written by a professor of mechanical engineering  so I am reluctant to assume that they are wrong without knowing anything about the subject myself. 
@rudydupreez Are you still getting notifications from Github?

rudydupreez:
In multiaxis machining Gcode rotation directions can be confusing. They 
van be defined in work coordinates or tool coordinates.
For xyzac-trt and xyzbc-trt I have used the work coordinate system as a 
reference for gcode programming, with the right hand rule
  for positive rotations. This seems to be the most usual  way as can be 
seen in the Mastercam definitions. See following link:

mastercam.com/news/blog/what-are-the-axes-in-5-axis-machining

See also references in on page 502 of the Lcnc  doc for v2.9.3-92.

Hope it helps
Rudy du Preez


------ Original Message ------
From "andypugh" ***@***.***>
To "LinuxCNC/linuxcnc" ***@***.***>
Cc "rudydupreez" ***@***.***>; "Mention" 
***@***.***>
Date 2025/06/19 00:54:45
Subject Re: [LinuxCNC/linuxcnc] Fix rotational axis direction with some 
5 axis kinematics (PR #2443)

DerAndere1:
Thanks for the clarification Mr. Preez. Your kinematics contibutions are immensely valuable and have brought CNC machining and open source software to another level. 

Seeing that axis directions can be defined in two ways, I strongly recommend to add new kinematic models that use the second definition. I have listed some more information sources about axis directions in my latest comment at issue https://github.com/LinuxCNC/linuxcnc/issues/2435

andypugh:
Maybe the answer here is to add an extra HAL pin to the relevant kinematics files. Called something like "left-handed-coordinates" or similar. 
By default this would be set to -1 (probably). 
Setting it to 1 would reverse the rotations. 

Then
``` C
pos->tran.z = joints[2] + zb - zv + *(haldata->pivot_length);
```
(as you suggest) would become
``` C
[code]pos->tran.z = joints[2] + zb - (left_handed_coordinates *  zv) + *(haldata->pivot_length);
```

This would leave the existing behaviour unchanged for existing configurations (we _really_ don't want to break those) but allow the option to change the reference frame used to define the sense of the rotations. 

(it probably needs a better name for the pin)

I propose doing it this way rather than as a conditional as I think it is likely to be slightly more efficient.

rudydupreez:
I agree with Andy's proposal which is very simple to implement. That 
will allow all possibilities as required by machine, CAM program or 
post-processor. One could even have one config with a flag button on the 
screen to make the selection or with an mcode that can be handled in a
number of ways.

Rudy Du Preez

------ Original Message ------
From "andypugh" ***@***.***>
To "LinuxCNC/linuxcnc" ***@***.***>
Cc "rudydupreez" ***@***.***>; "Mention" 
***@***.***>
Date 2025/06/22 21:32:41
Subject Re: [LinuxCNC/linuxcnc] Fix rotational axis direction with some 
5 axis kinematics (PR #2443)

DerAndere1:
I will update this PR according to your suggestion. Thanks for the feedback.

BsAtHome:
> Maybe the answer here is to add an extra HAL pin to the relevant kinematics files. Called something like "left-handed-coordinates" or similar. By default this would be set to -1 (probably). Setting it to 1 would reverse the rotations.
> [code]pos->tran.z = joints[2] + zb - (left_handed_coordinates *  zv) + *(haldata->pivot_length);

It is a bad idea to (blindly) accept values from a pin in this calculation. It is a boolean choice and should be handled as such. I'd suggest to do:
```c
...
/* Somewhere pin created */
hal_bool_t *left_handed_coordinates;
err = hal_pin_bit_new("pin-name-here", HAL_IN, &left_handed_coordinates, comp_id);
...
real_t lhc = *left_handed_coordinates ? 1.0 : -1.0;
pos->tran.z = joints[2] + zb - (zv * lhc) + *(haldata->pivot_length)
```
This does not take much longer, from a performance aspect. It loads a constant based on the bool, effectively adding only few assembly instructions. The compiler is free to optimize.

The bool pin, when created, should automatically default to false resulting in a -1 multiplication without setting it manually, thus keeping status quo.

Issue #2416, opened 2023-04-01
OpenGL error when starting AXIS on X11


allistar:
# The issue tracker is not a support forum

## Here are the steps I follow to reproduce the issue:

 1. Build and run the latest linuxCNC on an X11 (xorg) machine
 2. Run LinuxCNC
 3.

## This is what I expected to happen:
The Axis GUI should start as expected.

## This is what happened instead:
This error is happening:

`Traceback (most recent call last):
File "/home/allistar/Projects/linuxcnc-test/bin/axis", line 3964, in <module>
    get_coordinate_font(vars.dro_large_font.get())
File "/home/allistar/Projects/linuxcnc-test/bin/axis", line 3842, in get_coordinate_font
    glnav.use_pango_font(coordinate_font, 0, 128)
File "/home/allistar/Projects/linuxcnc-test/lib/python/glnav.py", line 65, in use_pango_font
    glBitmap(1, 0, 0, 0, 0, h-d, bytearray([0]*4))
File "src/latebind.pyx", line 39, in OpenGL_accelerate.latebind.LateBind.__call__
File "src/wrapper.pyx", line 318, in OpenGL_accelerate.wrapper.Wrapper.__call__
File "src/wrapper.pyx", line 311, in OpenGL_accelerate.wrapper.Wrapper.__call__
File "/usr/lib/python3.11/site-packages/OpenGL/platform/baseplatform.py", line 415, in __call__
    return self( *args, **named )`

## It worked properly before this:
I suspect this breakage is caused by an update to a system package.  I know that reverting [this](https://github.com/LinuxCNC/linuxcnc/commit/710ce7e39dcbc584615771947bf26658808ce8e3) commit fixes it.

## Information about my hardware and software:

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`): Gentoo Linux 2.13
 * I am using this kernel version (shown by `uname -a`): Linux 6.1.19-rt-rt8 # 3 SMP PREEMPT_RT
 * I am running ...
   * [ ] A binary version from linuxcnc.org (including buildbot.linuxcnc.org)
   * [x] A binary I built myself
   * [ ] A binary version from some other source besides linuxcnc.org
 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`): 2.10.0-pre0-1093-g7ba312e3d8
 * I am using this user interface (GUI) (e.g., AXIS, Touchy, gmoccapy, etc): AXIS
[...]

PR #2346, opened 2023-02-15
tweak glBitmap so it works on Buster, Bullseye, and Bookworm

SebKuzminsky:
Sigh.

This fixes DRO corruption on Buster (X11) [pointed out](https://github.com/LinuxCNC/linuxcnc/commit/56ecf76e11c5ce9feb78d3cd830c135db83c09d7#commitcomment-99885313) by @snowgoer540, and it still works fine on Bullseye (Wayland) and Bookworm (Wayland).

PR #2446, opened 2023-04-23
glbitmap fix for #2416

rene-dev:


rene-dev:
https://www.khronos.org/opengl/wiki/Raster_Position_And_Clipping

rene-dev:
@allistar @dwrobel @smoe @jethornton @phillc54 @petterreinholdtsen @hansu @20after4 can you guys please test this?
I have no system that can reproduce it.

snowgoer540:
When @SebKuzminsky pushed commit ee2623b, that commit did cause issue on Buster (and possibly older) systems where there would be a non-transparent pixel in each character block that made the preview overlays look noisy. I tested and passed my findings to Seb and he pushed commit 710ce7e. I never saw the original issue (crash) Seb sought to fix on my Bullseye based VM. 

For what it's worth, I did take the time to apply the changes in this commit to my Buster based VM and it has no issues with noisy pixels this time.  So from a Buster standpoint, this change seems fine to me.  But I do wonder if this will cause issue for the original group of people (whoever they were)?  I guess that comes down to how similar None and ''.encode() are?

I don't have hardware that has crashes with this, so it's hard for me to test otherwise.

phillc54:
It works for me on Bullseye X11 and Bookworm X11

It crashes with the following on Bookworm Wayland
```
File "/usr/lib/python3/dist-packages/OpenGL/error.py", line 230, in glCheckError
    raise self._errorClass(
OpenGL.error.GLError: GLError(
    err = 1285,
    description = b'out of memory',
    baseOperation = glBitmap,
    pyArgs = (0, 0, 0, 0, 0, 16.0, None),
    cArgs = (0, 0, 0, 0, 0, 16.0, None),
    cArguments = (0, 0, 0, 0, 0, 16.0, None)
)

```

JayFoxRox:
I think this patch is an improvement and should be accepted.

>  I guess that comes down to how similar None and ''.encode() are?

The difference should be between a null pointer (~`NULL`) and a pointer to an empty buffer.
I couldn't find proper docs for PyOpenGL though, so there might be special cases for empty buffers.

The GL manpage specifically has a note about using `NULL` to advance the raster position.

> It crashes with the following on Bookworm Wayland

`GL_OUT_OF_MEMORY` in particular is a very weird error for this situation, because the `glBitmap` command should barely consume any memory the way it is written now.

So this sounds like a bug with your OpenGL implementation (or a lack thereof). What implementation are you using?

If it's not a bug in the driver, then the error checking might trigger for a previous command (example, if `glNewList()` does not call `glGetError()`, so the next command which uses `glGetError()` sees it).

Did you try to run it through [apitrace](https://github.com/apitrace/apitrace) or some other OpenGL debugger?

rene-dev:
I just ran apitrace, and the call looks correct, so None is the correct parameter.
the issue must be somewhere else.
`glBitmap(width = 0, height = 0, xorig = 0, yorig = 0, xmove = 16, ymove = -25, bitmap = NULL)`

phillc54:
I am not sure that I am doing this correctly. apitrace is v11.1+repack-1.1 installed with Synaptic.

The line now reads:
`apitrace trace $EMCDISPLAY -ini "$INIFILE" $EMCDISPLAYARGS $EXTRA_ARGS`

This is the result:
[python3.11.trace.zip](https://github.com/LinuxCNC/linuxcnc/files/11315658/python3.11.trace.zip)

rene-dev:
Can you open and run the trace file in qapitrace(apitrace-gui package), and see if it reports other errors?

phillc54:
These are the results of Trace > Replay
Screen Shot 2023-04-25 at 12 39 36 pm
Screen Shot 2023-04-25 at 12 40 01 pm

phillc54:
Sure.
[glxinfo.txt](https://github.com/LinuxCNC/linuxcnc/files/11316078/glxinfo.txt)

phillc54:
Would that be because all of the tests I did were on VM's? Although this machine should be identical to the one with X11.

phillc54:
If it is any help I just ran up the X11 VM which runs fine and ran glxinfo, attached is the diff between the X11 and Wayland VM's.
[glxinfo.diff.txt](https://github.com/LinuxCNC/linuxcnc/files/11316144/glxinfo.diff.txt)

phillc54:
OK, I made a mistake. I hadn't upgraded my X11 VM for the previous test and since upgrading it also gives the same error. I am now getting this error on four up to date Bookworm VM's, two amd64 and two arm64. Unfortunately I don't have a bare metal Bookworm to test on.

JayFoxRox:
> I don't know, if this is a bug in the mesa software render, because nobody nowadays uses these outdated functions, however it seems to be related to the following codepath in mesa:
> 
> [...]


Looks like a generic mesa bug or underspecification in OpenGL about:
- `glBitmap` with `NULL` (which is underspecified for PBOs, and just a note in a man-page, not in the specification)
- `glBitmap` being ignored based on raster-pos (which is unknown while recording a display-lists)
- `glBitmap` interaction with display-lists (side-effect of modifying raster-position)

I've created an issue about mesa in https://gitlab.freedesktop.org/mesa/mesa/-/issues/8904

> I think this patch is an improvement and should be accepted.

So I was wrong, due to it all being very vague in the OpenGL specification.

The mesa issue will probably affect many users for the next couple of years (until a potential fix is rolled out completly).

I now think a better solution would be to have a 0x0, 0x1/1x0 or 1x1 image with a non-zero *valid* buffer. It should avoid all previously mentioned issues (but you might want to read the spec again to make sure).
However, that solution (#2346) was used before and also didn't work (#2416).
Anyway, it might make sense to debug why those solutions didn't work, and attempt workarounds or fixes for those solutions.

rene-dev:
I implemented a workaround: https://github.com/LinuxCNC/linuxcnc/commit/fced097d2af3fc073d30bdc47523feca71481828

andypugh:
> I implemented a workaround: [fced097](https://github.com/LinuxCNC/linuxcnc/commit/fced097d2af3fc073d30bdc47523feca71481828)

So, should this PR be closed? And does https://github.com/LinuxCNC/linuxcnc/commit/fced097d2af3fc073d30bdc47523feca71481828 need to be cherry-picked into 2.9?

smoe:
I have no idea. A later version of mesa has a fix included for glBitmap https://gitlab.freedesktop.org/mesa/mesa/-/commit/bac1c88ea3e1f0966505a3bfb016c6b2e8cbb629 but shall we wait for that? Merged three months ago, https://gitlab.freedesktop.org/mesa/mesa/-/tags suggests that 23.1.3 or later are fine. On https://tracker.debian.org/pkg/mesa it states that current testing should no longer be affected.
Since we only get backported versions of LinuxCNC into the current stable release anyway, my idea would be to have that depend on a later version of the mesa libraries, which we could then backport, too.

I do not understand what the patch is doing. I suggest leaving it in with a comment to remove it when 23.1.3 or later has become a runtime requirement.

rene-dev:
The patch tries the other way in case the first way fails. It's a hack, but works.

snowgoer540:
> And does [fced097](https://github.com/LinuxCNC/linuxcnc/commit/fced097d2af3fc073d30bdc47523feca71481828) need to be cherry-picked into 2.9?

The code was the same in both 2.9 and master so it's my vote that we keep them consistent until 2.9 is released.

andypugh:
Apparently fixed by bde2cdc5051647e36435136adee2c13768b5f52b

PR #2451, opened 2023-04-26
Carousel.comp - added debounce for jog

zz912:
Hello,

I have machine with ATC encoder without trigger signal. In the past carousel.comp did not work well for my machine. Andy helped me and made debounce for positioning. https://github.com/LinuxCNC/linuxcnc/commit/f6e8d08c35be9dfb04476db0f9ac3a3094125f34

Now I am making jog ATC. I solved the encoder reading problem for jog just like Andy solved for positioning.


andypugh:
Thanks, seems reasonable.

PR #2457, opened 2023-04-29
Docs: add document "writing-tests" also to 2.9

hansu:
This document was only added to master but I think this should also go into 2.9.
I already spent too much time searching for it in 2.9 :see_no_evil:.
And yes I am aware that this will probably produce some merge conflicts, so merge to master after this has been merged but I think this is worth it.

### DIFF
--- a/docs/po4a.cfg
+++ b/docs/po4a.cfg
@@ -26,6 +26,7 @@
 [type: AsciiDoc_def] src/code/nml-messages.adoc $lang:src/$lang/code/nml-messages.adoc
 [type: AsciiDoc_def] src/code/rs274.adoc $lang:src/$lang/code/rs274.adoc
 [type: AsciiDoc_def] src/code/style-guide.adoc $lang:src/$lang/code/style-guide.adoc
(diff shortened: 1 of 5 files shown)


petterreinholdtsen:
The reason I only added it to master is that the majority of new developer should take place in master, and I thus assume the majority of new tests should be written for and committed to master.  Why are you looking for it in 2.9?

hansu:
So is the consensus to not have this in 2.9?

hansu:
Ok :+1: . Rebased and resolved merge conflicts.

hansu:
> I think adding new docs and new tests in 2.9 is a good idea. I support this PR.

So then merge it?

smoe:
Go for it. What it describes is already valid for the current release and we likely add tests to 2.8 and 2.9 over the next months, so we can steadily improve that description.

SebKuzminsky:
The version of `writing-tests.adoc` that was added to 2.9 by this PR is slightly different from the version in master.  It looks like a few bugfixes were applied to the doc in master, but not in 2.9.  I have backported the bugfixes to 2.9.

hansu:
Thanks!

PR #2458, opened 2023-04-30
mb2hal: wait for threads to stop before deallocating the memory they'…

SebKuzminsky:
…re using

This fixes a shutdown-time race condition in mb2hal between the main thread and the per-link threads.

The main thread connects to all the Modbus interfaces and allocates all HAL data structures, then starts a worker thread for each modbus link.

The main thread and the worker threads all watch a "time to shutdown" status variable, which is set by a shutdown signal handler.  When the "time to shutdown" variable gets set, the per-link threads all exit and the main thread closes down the modbus communications and deallocates the HAL memory.

But there was no synchronization between the shutdown of the main thread and the shutdown of the worker threads, so sometimes the worker threads were still accessing the stuff that the main thread was freeing, leading to segfaults.

This commit makes the main thread join the worker threads before deallocating anything.

### DIFF
--- a/src/hal/user_comps/mb2hal/mb2hal.c
+++ b/src/hal/user_comps/mb2hal/mb2hal.c
@@ -91,7 +91,6 @@ int main(int argc, char **argv)
 
     /* Each link has it's own thread */
     pthread_attr_init(&thrd_attr);
-    pthread_attr_setdetachstate(&thrd_attr, PTHREAD_CREATE_DETACHED);
     for (counter = 0; counter < gbl.tot_mb_links; counter++) {
         ret = pthread_create(&gbl.mb_links[counter].thrd, &thrd_attr, link_loop_and_logic, (void *) &gbl.mb_links[counter].mb_link_num);
         if (ret != 0) {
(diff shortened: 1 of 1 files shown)


SebKuzminsky:
I still see occasional failures depending on where in the execution of mb2hal the quit signal is delivered:
```
Running test: ../tests/mb2hal/mb2hal.1a
--- expected    2023-04-29 09:03:06.557730300 -0600
+++ result      2023-04-30 10:29:43.131633835 -0600
@@ -122,9 +122,9 @@
 mb2hal create_each_mb_tx_hal_pins DEBUG: mb_tx_num [4] pin_name [mb2hal.Modbus_fnct_16.02]
 mb2hal create_each_mb_tx_hal_pins DEBUG: mb_tx_num [4] pin_name [mb2hal.Modbus_fnct_16.03]
 mb2hal main OK: HAL components created OK
+mb2hal quit_signal DEBUG: signal [15] received
 mb2hal main OK: Link thread loop and logic 0 created OK
 mb2hal main OK: mb2hal is running
-mb2hal quit_signal DEBUG: signal [15] received
 mb2hal quit_cleanup DEBUG: started
 mb2hal quit_cleanup DEBUG: unloading HAL module [(ignored comp id)] ret[0]
 mb2hal quit_cleanup DEBUG: done OK
*** ../tests/mb2hal/mb2hal.1a: FAIL: checkresult exited with 1
```

This PR does not address that problem, only the segfault issue.

PR #2461, opened 2023-05-01
Gmoccapy 3.4.3

hansu:
To do:

~- [ ] Fix problem with override limit (https://forum.linuxcnc.org/gmoccapy/37676-ignore-limits-option-grayed-out?start=20#270162)~
- [x] Add setting to disable sound
- [x] Add Embedded Tab location next to DRO
- [ ] ~Add setting for width of virtual keyboard for widescreens~ --> Will be shifted to next release


hansu:
One question - is "Window decorated" clear to understand for those who have not read the Gtk API? Or should I better word it like "Show resp. hide title bar" ?

grafik

PR #2463, opened 2023-05-03
Typo in Create or Edit Section

YClayton:
Replace "over write" with "overwrite".


petterreinholdtsen:
Look good to me.  Should it go into an earlier edition or is the master branch a good target for it?

hansu:
The last one went also in master (https://github.com/LinuxCNC/linuxcnc/pull/2462).
I would also prefer in 2.9, but doesn't matter for that little change I think.

Issue #2467, opened 2023-05-04
race between EMC_AUX_ESTOP_OFF and EMC_TOOL_ABORT


rene-dev:
checkout master
comment out emcLubeOff() here: https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/task/emctask.cc#L346
run sim/axis
9 out of 10 times you cannot get the machine out of estop.
also, many tests fail as exiting estop doesnt work reliably, for example abort/feed-rate
if you replace emcLubeOff() with emcCoolantFloodOff() its working again.

rene-dev:
emcAuxEstopOff() sends EMC_AUX_ESTOP_OFF using forceCommand,
emcIoAbort() sends EMC_TOOL_ABORT using sendCommand.

adding a printf in io shows only EMC_AUX_ESTOP_OFF_TYPE or EMC_TOOL_ABORT_TYPE is ever received in io.
if only EMC_TOOL_ABORT_TYPE is reveived, the machine does not exit estop.

probably the best solution is to remove EMC_TOOL_ABORT_TYPE, as it is not called anyway if the machine does get out of estop.

rene-dev:
turns out both commands are force-queued.
just removing emcIoAbort doesnt fix it reliably.
https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/task/taskclass.cc#L167

c-morley:
Is it used in iov2?

rene-dev:
Both components do something with the message. It has nothing to do with v2

rene-dev:
probably also the cause of #139

PR #2471, opened 2023-05-05
Removing unused or broken parts of code.

rene-dev:
This branch is an attempt to clean up a bit.

1. removal of emctask module. it seems broken, and unused. google only shows 2 hits. the configs using it have been orphaned. there is more code, Im still working on this. let me know if anyone uses it.
2. removal of unused nml files. should not affect anybody.
3. removal of anything lube in IO and NML. the idea predates hal, and shows the madness of a data passing from hal over nml back to hal. If you are using lube pins with halui, just rewire them directly in hal. if you are using tklinuxcnc and lube pins, this does affect you(anyone?). this is part of a bigger PR Im working on cleaning up IO.


c-morley:
emctask I believe is to allow remap.
Unfortunately Michael left before finishing the docs:
http://linuxcnc.org/docs/2.9/html/remap/remap.html#_introduction_extending_task_execution
Do you think it would be difficult to fix? Seems a shame to lose the potential ability.

c-morley:
To allow remap of task - which I'd guess never got tests. Maybe it never got finished. I don't know for sure.
With documentation, it would surely get used.

SebKuzminsky:
I'm fairly ignorant when it comes to remap, can you educate me here - what does it mean to "remap task"?

rene-dev:
The module isnt imported anywhere, but in one orphaned configuration.  It has nothing to do with remaps.

c-morley:
How do you remap task then?

c-morley:
> I'm fairly ignorant when it comes to remap, can you educate me here - what does it mean to "remap task"?

http://linuxcnc.org/docs/2.9/html/remap/remap.html#_introduction_extending_task_execution
Well I can only extrapolate from this.

SebKuzminsky:
I would prefer to see this as three PRs that can be discussed individually.  But since it's one PR covering all three topics I'll try to address them all here.


# Remove emctask

I don't know what emctask is or what it does.  Or maybe I should say, what it was indended to do before it was abandoned mid-development?  I see that removing it doesn't seem to break the tests, so...  It's probably ok?  I think at the same time (i.e. in the same PR) we should remove the orphaned configs that use it.


# Remove unused NML files

Is this just old unused cruft?  Good riddance, thanks for removing it.


# Remove lube pin NML messages and remove lube pins from IO

Can you confirm my understanding of this one?

## Before this commit

You could send NML messages from the UI to io (via Task) that would request io to turn lube on/off, or you could poke `halui.lube.on` and `halui.lube.off` to request that halui send those NML messages for you.

In addition, Task would send these lube-control NML messages to io all by itself, to request "lube on" when we _come out_ of E-stop, and "lube off" when we _enter_ E-stop. 

The only(?) effect of io's "lube control" was to set the HAL pin `iocontrol.0.lube` true if the lube system should be on and false if it should be off.  It was up to the machine builder to connect that pin to some HAL circuitry that would run lube pump timers or whatever's appropriate for that machine.

## After this commit

Task and IO now have nothing to do with lube.  You can not control lube via NML, and you can not control it via halui.  The old `iocontrol.0.lube` pin no longer exists.

If you want to control lube on your machine you have to find or generate some other control signal to drive your existing machine-specific lube stuff, perhaps [`iocontrol.0.user-enable-out`](http://linuxcnc.org/docs/devel/html/man/man1/io.1.html#PINS).

## My comments

I am not sure this is a good idea.  At the very least this needs documentation for everyone who's used to the old way of interacting with lube.

c-morley:
This mentions importing task:
http://linuxcnc.org/docs/2.9/html/remap/remap.html#remap:how-tool-change-currently-works

c-morley:
importing emctask using .py hot comments seems to work.
What is broken about it ?

c-morley:
> How do you remap task then?

I guess extending is the right word rather then remapping

SebKuzminsky:
@c-morley thanks for that info.

I'm still missing the big picture here - do you what some use cases are for this feature?

SebKuzminsky:
> I think you could make a new/extend IO control using python.
> maybe like add a bed heater to IO control.

Wouldn't that need new NML messages, to send from Task to IO?

The bed heater on my 3d printer just uses [M68](http://linuxcnc.org/docs/devel/html/gcode/m-code.html#mcode:m68), and a pid component in HAL uses the matching `motion.analog-out-???` pin as its command input.

I think this is pretty similar to how @rene-dev is proposing (in this PR) to handle lube: no NML messages, no special handling in any part of the "core" of LinuxCNC.  Just a custom HAL circuit, getting a control signal from G-code or maybe some existing "here's my state" output pin from `motion` or `io` or something.

In the case of lube it's maybe just a bit signal connecting `iocontrol.0.user-enable-out` to a gpio output pin (out in the real world, that output pin turns on existing pump/timer hardware).

rene-dev:
> # Remove emctask
> I don't know what emctask is or what it does. Or maybe I should say, what it was indended to do before it was abandoned mid-development? I see that removing it doesn't seem to break the tests, so... It's probably ok? I think at the same time (i.e. in the same PR) we should remove the orphaned configs that use it.

the configs have been moved to orphaned 7 years ago. I doubt anyone is using it. google and the forum results in nothing.
https://github.com/LinuxCNC/linuxcnc/commit/49f30eebb2a72784e80769c736d928951fc0112a
 
> # Remove unused NML files
> Is this just old unused cruft? Good riddance, thanks for removing it.

yes, more to come.


> I am not sure this is a good idea. At the very least this needs documentation for everyone who's used to the old way of interacting with lube.
> 

By default, lube represents the enable state(also unreliable, due to #2467). this can be changed only by sending the nml command. only 2 bits of code do that, halui and tklinuxcnc.

halui kind of defeats the purpose, setting a hal pin via hal.
I doubt anyone uses the lube pin AND tklinuxcnc.

The history of the feature probably predates hal.

c-morley:
It was just an example idea - I am not saying it is the best example.
Think like io control2 - that could have been done in python.

Also you can query linuxcnc internals from a GUI, which is obscure but sometimes handy.
Why would we remove potential functionality?
If we documented it, I bet it would be used - python is a much lower bar then c.

c-morley:
There you go - looks it was used to make a rack style tool changer.
Just because someone put it in as an orphan, doesn't mean it was useless idea.

rene-dev:
@c-morley somebody started an iocontrol in python, didnt go anywhere.
Im not saying it needs to be removed, Im just saying its probably dead and will never be used.
I could not find an example anywhere.
most of the machine specific stuff can now be done in hal and remap.
a rack toolchanger is now implemented as a remap, as it requires the machine to be moved.

rene-dev:
@c-morley the rack example you mentioned doesnt actually use the emctask module, it uses a remap.

c-morley:
I'm confused - doesn't your pull request remove the task code?
You said it's broken - is it?
in what way?
you say it's not used - probably not - it's hardly documented and the only example was pretty much hidden.

How about we try to document it -it's probably useful.

c-morley:
Yes I see the rack part is the remap.
It uses task extending for io control.

c-morley:
Ok digging a little from wiki:
http://wiki.linuxcnc.org/cgi-bin/wiki.pl?RemappingStatus

"
Task Python plugin

There have been long discussions on the role of iocontrol. My take is that it is on the way out and should be replaced by Embedded Python as outlined in this branch (see the configs/sim/remap/iocontrol-removed config how that works).

However, this really makes only sense in the context of further work items, which I plan to work on (see below). Only then this feature starts making real sense. Since it is configurable and deactivated by default, it does no harm (ha!), which is why I suggest leaving it in.
"

So it's code to make things easier to change in the future.
So really iocontrol/v2 should be replaced with python to make future changes a lower bar to change.
Or at least that seems to be the idea.

rene-dev:
as suggested, Im splitting this up into 2 PRs. lets move the task related discussion there.

rene-dev:
any objections to whats left in this PR? only functional change is removal of lube pins.

hansu:
I never used the lube pins and so doesn't know about their functionality. I only wonder if there is an alternative when those pins are removed?

rene-dev:
see above explanation, basically they are useless.
https://github.com/LinuxCNC/linuxcnc/pull/2471#issuecomment-1536573018

c-morley:
Why do you sat they are useless? If you want to control a lube so it automacally goes on and off with machine-on plus is controllable from say halui, it's pretty handy. I bet it could be added to the python module then it becomes more useful as you could have timed lube, lube only on motion etc.

andypugh:
Yes, they just basically mirror "machine on"

Maybe add a bit to the update_ini script to automatically replace them?

c-morley:
except you can turn them off without turning the machine off via halui

c-morley:
So it was easy to add lube commands to the python module which makes it much more useful.
I think it needs a little love ie lube level should be more then a bool.
It's nice to have it in linuxcnc's status so it's available to the gui without connecting pins.
If it's not a big deal, I vote to keep it.

hansu:
Maybe we should improve it a bit to make it more useful instead of removing it?
I could image adding values to set a duty cycle and pulse the lubrication...

>removal of anything lube in IO and NML. the idea predates hal, and shows the madness of a data passing from hal over nml back to hal.

If your main motivation was the way how it's implemented - maybe a bit refactoring would be better?

rene-dev:
@hansu @c-morley look at this commit only. https://github.com/LinuxCNC/linuxcnc/pull/2471/commits/975fdf78ec238f3a689d96a35b747b9fb45b91a8
so you are saying all that code is reasonable just to change a hal pin?
yes its easy to que up a command, but adding some is a nightmare. they are fixed at compiletime, you cannot add more from python.
If I you want to add for example thru spindle coolant, you would probably use a motion output or mcode, but not add all this nml madness.
this is communication from the UI to hal, or from hal to hal. I dont see why you would involve nml at all.

do you use the pin? do you know anyone using the pin?
I did a quick research on the forum, and all of them use use cl or custom hal for lube,
as most lube systems need more than just an enable pin.
Thats what hal was invented for, like I said this predated hal where you would actually implement the custom behavior and driver in iocontrol. I dont think such behavior belongs in the linuxcnc core.

Its also broken. did you see the linked issue about the race condition?
there is a race condition which causes this to be unreliable.
has been like this for at least 13 years according to git.

hansu:
> so you are saying all that code is reasonable just to change a hal pin?

No, I support the clean-up in general. I only thought the lube pin could remain for compatibility.


>do you use the pin? do you know anyone using the pin?

No. My current machine doesn't have the ability for that. 

>I dont think such behavior belongs in the linuxcnc core.

Maybe, maybe not. Maybe the most machines run a more basic setup and don't need that.

petterreinholdtsen:
I believe @andypugh mentioned during the weekend that he is using the lube pin to control his work light, so the lube pin might be used quite a bit like this.  It provide a pin that show up as a on/off button in the guis. :)

andypugh:
> I believe @andypugh mentioned during the weekend that he is using the lube pin to control his work light

No, I use the mist pin. 
I haven't checked, but according to Rene the Lube pin goes on with machine-on and off with machine-off, and is therefore redundant. 

If this is the case, then is there an argument for adding a new filter to the update_ini script to switch the lube pin to machine-on?

hansu:
>I haven't checked, but according to Rene the Lube pin goes on with machine-on and off with machine-off, and is therefore redundant.

Except it can be turned off without turning off the machine

hansu:
I would say we should decide whether we want to provide pins for possible use cases or whether we should keep the provided list of pins low and force the user to do this in hal.
So I think it's a bit contradictory adding pins for a work light on one hand and remove pins for lubrication on the other.

andypugh:
> Except it can be turned off without turning off the machine

Ah, yes. 
halui.lube.off bit in - pin for stopping lube
halui.lube.on bit in - pin for starting lube

Maybe we could combine lube, coolant, worklight and several uncommitted pins into a single NML message? I think that there is a definite use-case for HAL pins that can be controlled by G-code, halui, python (and maybe other hal pins using the NIST on, off, is-on structure). Possibly with names definable (and over-ridable) in the INI. (Just thinking out loud here)

hansu:
>>    Except it can be turned off without turning off the machine
>
>Ah, yes.
>halui.lube.off bit in - pin for stopping lube
>halui.lube.on bit in - pin for starting lube

I was thinking more about

```
iocontrol.0.lube_level

(Bit, In) Should be driven FALSE when lubrication tank is empty.
```

which probably also switch off the lube pin.

andypugh:
> iocontrol.0.lube_level
> which probably also switch off the lube pin.

This sort of hard-coded behaviour would be an important consideration for the discussion I am trying to start in #2532. 
It might be as simple as not creating the lube-level pin if the lube pin is re-allocated in the INI.

c-morley:
> > I believe @andypugh mentioned during the weekend that he is using the lube pin to control his work light
> 
> No, I use the mist pin. I haven't checked, but according to Rene the Lube pin goes on with machine-on and off with machine-off, and is therefore redundant.
> 
> If this is the case, then is there an argument for adding a new filter to the update_ini script to switch the lube pin to machine-on?

But a small change in the linuxcnc python module would allow you to control the lube pin. If I had realized this I would have already done this for timed lubed control.

gmoccapy:
A Lube system can be realized completely with Hal or not?
I have made a python glade timer Modul counting the time machine on, machine moving and spindle time (can be found in the forum) that one can also be adapted to present Hal pin and realize a complete lube system.

So the ability is still given, the question is, if the way I described is the one a normal user may understand? And will a normal user use a lube system?

Many questions, IMHO just remove the old stuff and if users ask for such a feature integrate that feature with reliable and we'll documented code.

Norbert

PR #2484, opened 2023-05-11
Rodw getting linuxcnc

rodw-au:
Seb wanted to update this document pending the 2.9 release and  incorporate my learnings that have been discussed on the forum and in the developers email list.

A lot of new content is required for Bookworm and explain how to optimise performance.
I have removed references to RTAI and Precise and instead referred users to the V2.8 docs if they need more information about these. The document was very heavy on content for essentilly  obsolete instalation methods so it needed to go to make room for the new content.

We want users to be able to install Linuxcnc without wading though volumes of obseete content.

PREEMPT_RT has won the RT battle and is expected to reach the mainline Linux kernel code soon, possibly in the 6.5 release (current kernel mainline is 6.4)... Let's not focus on other RT methods that are not available from a Bookworm installation.

I also removed info on using Zsync as with more reliable bandwidth and verifying the image after download as this is not required today.

For installing .ISOs, Balena Etcher is recommended as it runs on Linux, Macintosh and Windows. It is by far the quickest and simplest tool I have used. This also reduces the content.

I have added FIX_ME in a few places where helo is required.

It would be great if somebody could test the Raspberry Pi instructions. Its been a long time since i have done it as suggested.

Please consider this as a draft and provide your feedback or suggest amendments.

If you could find a corner on a Linuxcnc server to host the 6.3 kernel and headers I now have in google drive, it would be a great help.

### DIFF
--- a/docs/src/getting-started/getting-linuxcnc.adoc
+++ b/docs/src/getting-started/getting-linuxcnc.adoc
@@ -11,10 +11,10 @@ adventurous.  If you have an existing install that you want to upgrade,
 go to the <<cha:updating-linuxcnc,Updating LinuxCNC>> section instead.
 
 NOTE: LinuxCNC requires a special kernel with real-time extensions.
-There are three possibilities here: preempt-rt, RTAI or Xenomai. In
-addition there are two versions of LinuxCNC which work with these kernels.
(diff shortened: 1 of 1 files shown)


rodw-au:
Closing this as I have more changes to make.

PR #2486, opened 2023-05-11
change interpl linked list to std::deque

rene-dev:


andypugh:
This looks simpler, and still appears to work. What's the motivation? Just being more standard?

SebKuzminsky (approved):

andypugh:
Good enough for me.

PR #2487, opened 2023-05-11
replace deprecated Py_SetProgramName with PyConfig_SetString

rene-dev:
https://peps.python.org/pep-0587/


rene-dev:
I dont know whats going on here, is the python in buster too old?

gmoccapy:
stuttgart meeting says can be merged when we drop support for buster.

PR #2493, opened 2023-05-12
hm2 oneshot: change logical or to bitwise or

rene-dev:
I could not find the register definition for this, but it looks like a bug to me.
is the intended behavior to mask 3 bits?
maybe @pcw-mesa or  @SebKuzminsky can confirm?


hansu:
If it's a bug it should go into 2.9 or maybe even in 2.8, I suppose.

pcw-mesa:
Yes that's a typo...
Note sure where the register map map should go

rene-dev:
let me rebase to 2.8

rene-dev:
2.8 doesnt have this componen so 2.9 it is.

PR #2494, opened 2023-05-12
Python3 gcode generators

### LINKED ISSUE #2491: update python G-code generators to python3, update [FILTER]PY to python3
While looking at @ozzyrob's #2485, i realized that our python G-code generator scripts were missed in the python2-to-python3 transition.  We should update all our sample configs to call out `[FILTER]PY = python3`, and update our python G-code generators to run in python3.

Here's the list of our python G-code generators:
```
$ find $(find . -name nc_files -type d) -name '*.py'
./nc_files/filtertest.py
./nc_files/filtertest-gui.py
./nc_files/flowsnake.py
./nc_files/holecircle.py
./nc_files/involute.py
./nc_files/remap_lib/python-stdglue/stdglue.py
```

SebKuzminsky:
This fixes #2491.

@ozzyrob, this PR includes your holecircle fix from #2485.

### DIFF
--- a/configs/by_interface/mesa/hm2-servo/3x20-small.ini
+++ b/configs/by_interface/mesa/hm2-servo/3x20-small.ini
@@ -33,7 +33,7 @@ PROGRAM_EXTENSION = .py Python Script
 png = image-to-gcode
 gif = image-to-gcode
 jpg = image-to-gcode
-py = python
+py = python3
 
 [RS274NGC]
 # File containing interpreter variables
--- a/configs/by_interface/mesa/hm2-servo/4i65.ini
+++ b/configs/by_interface/mesa/hm2-servo/4i65.ini
@@ -33,7 +33,7 @@ PROGRAM_EXTENSION = .py Python Script
 png = image-to-gcode
 gif = image-to-gcode
 jpg = image-to-gcode
-py = python
+py = python3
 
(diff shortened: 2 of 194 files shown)


SebKuzminsky:
@ozzyrob , the updated holecircle script prints some error messages when I use it.

I open a terminal, run `linuxcnc`, and run the script.  In the terminal where I ran LinuxCNC I get a bunch of back traces from holecircle, but the gcode loads and runs fine.

Do you have time to debug this?

ozzyrob:
Ok I had a look, and it would seem to happen when the input box has a "NULL" value. eg the value in the box is deleted to input a new value.

Tried this snippet
def update_ok(*args):
        result = True
        for i in vars.values():
            try:
                i.get()
            except ValueError:
                result = False
                break
            except FloatingPointError:
                result = False
                break
            **except (TypeError, tkinter.TclError):
                result = False
                break**  

Bold section was added

ozzyrob:
When changing the angle I got a trace back:

This snippet seemed to work:
    def update_preview(*args):
        prev.delete("all")
        try:
            count = vars['count'].get()
            th0 = vars['th0'].get()
            inc = vars['inc'].get()
        except ValueError: return
        **except (TypeError, tkinter.TclError): return**

Once again bold section was added

PR #2499, opened 2023-05-21
convert enums to enum class in task and interp

rene-dev:
enum class provides some advantages compared to a c enum:

1. doesnt clutter global namespace
2. its a real type, so can be type checked by the compiler(found 2 bugs this way)
3. wont silently cast to bool or int
4. makes core more readable where enum silently casts to bool or int
5. editor can autocomplete them
6. eventually better python integration

unfortunately you cant pull in the  enum namespace, thats a cpp20 feature, so you have to specify the namespace every time, or use every single item.
There are many more to convert.


SebKuzminsky:
This looks like a great cleanup, thanks @rene-dev

Issue #2483, opened 2023-05-11
address of array 'settings->speed_override' will always evaluate to 'true'


rene-dev:
when speed_override was changed to an array for multi spindle, use of speed_override in emc/rs274ngc/interp_write.cc has not been updated. clang correctly complains about this.

> emc/rs274ngc/interp_write.cc:165:19: warning: address of array 'settings->speed_override' will always evaluate to 'true' [-Wpointer-bool-conversion]
    if (settings->speed_override) emz[6] =  48;
    ~~  ~~~~~~~~~~^~~~~~~~~~~~~~
emc/rs274ngc/interp_write.cc:167:24: warning: address of array 'settings->speed_override' will always evaluate to 'true' [-Wpointer-bool-conversion]
  } else if (settings->speed_override) {
         ~~  ~~~~~~~~~~^~~~~~~~~~~~~~
emc/rs274ngc/interp_write.cc:318:53: warning: address of array 'settings->speed_override' will always evaluate to 'true' [-Wpointer-bool-conversion]
    state.flags[GM_FLAG_SPEED_OVERRIDE] = settings->speed_override;
                                        ~ ~~~~~~~~~~^~~~~~~~~~~~~~
3 warnings generated.

andypugh:
Addressed (poorly) by b0ad7136e8f046c276b17d1734cc1fc98ee30cd6

PR #2501, opened 2023-05-21
Issue #2483. Simplistic fix to suit 90%+ of use cases.

andypugh:
A possible fix to #2483 

This probably covers 90+% of installations, ie those with only one spindle. 
A better fix would require:
1) The settings struct to have a copy of the count of spindles, to iterate through them
2) The state-tags struct to have 8 spindle-on bits, 8 spindle-cw bits and 9 speed-override bits. 

Active M-codes could either attempt to report for each spindle, or just report that spindle override is active if it is active on _any_ spindle. 

Spindle-override has been reporting enabled even if disabled for years now, with no complaints. It is probably very unusual to use the M51 P0 command


SebKuzminsky (approved):
This looks like a step in the right direction to me.

andypugh:
I think that this PR is redundant, as a fix was already made in https://github.com/LinuxCNC/linuxcnc/commit/b0ad7136e8f046c276b17d1734cc1fc98ee30cd6

PR #2522, opened 2023-06-12
2.9 Updates to getting-linuxcnc

rodw-au:
Replaced #2496

### DIFF
--- a/docs/src/getting-started/getting-linuxcnc.adoc
+++ b/docs/src/getting-started/getting-linuxcnc.adoc
@@ -4,430 +4,528 @@
 [[cha:getting-linuxcnc]]
 = Getting LinuxCNC(((Getting LinuxCNC)))
 
-This section describes how to install LinuxCNC version 2.9 on a PC already running Debian Bookworm (Debian 12). Linuxcnc requires the PREEMPT_RT real time kernel to ensure the strict timing required for CNC operations are met. We will also cover  how to install Debian Bookworm optimised for Linuxcnc on the x86/AMD64 and ARM64 platforms for first time users. We will also cover some troubleshooting steps and common problems you may experience.
+This section describes the recommended way to download
+and make a fresh install of LinuxCNC.  There are also
+<<sec:_alternate_install_methods,Alternate Install Methods>> for the
+adventurous.  If you have an existing install that you want to upgrade,
+go to the <<cha:updating-linuxcnc,Updating LinuxCNC>> section instead.
 
-NOTE: The PREEMPT_RT kernel is a dependency of LinuxCNC so on most computer platforms, eg AMD64/x86, it will be installed with LinuxCNC. On some platforms such as the Raspberry Pi, it needs to be installed seperately. Some environments (eg. Armbian) may suport installation of Debian Bookworm but require special procedures to install PREEMPT_RT which are beyond the scope of this document.
+NOTE: LinuxCNC requires a special kernel with real-time extensions.
+There are three possibilities here: preempt-rt, RTAI or Xenomai. In
+addition there are two versions of LinuxCNC which work with these kernels.
+See the table below for details.
 
-There are other versions of LinuxCNC which work with Xenomai and RTAI kernels. See the table at the end of this document for details.
+Fresh installs of LinuxCNC are most easily created using the Live/Install
+Image.  This is a hybrid ISO filesystem image that can be written to a
+USB storage device or a DVD and used to boot a computer.  At boot time
+you will be given a choice of booting the "Live" system (to run LinuxCNC
+without making any permanent changes to your computer) or booting the
+Installer (to install LinuxCNC and its operating system onto your
+computer's hard drive).
 
-[[sec:installing_linuxcnc]]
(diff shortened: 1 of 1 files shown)


rodw-au:
@sebastiankuzminsky @hansu 
Rebased against 2.9 branch as requested

andypugh:
Discussed at the Norway meetup.

rene-dev:

dpslwk:
in the `Updating Linuxcnc on Debian Bookworm (X86 only)` section
the previous instructions for adding the custom linuxcnc repositories took the safe option of adding them into a serpate sources.list file namely `/etc/apt/sources.list.d/linuxcnc.list`

yet these new instructions now have users directly modifying the base list file `/etc/apt/sources.list`
(it even includes a waring to say messing with this is dangerous)

adding custom repositories via separate files under `/etc/apt/sources.list.d/sources.list.d/` is I believe the preferred way / best practice

rodw-au:
> in the `Updating Linuxcnc on Debian Bookworm (X86 only)` section the previous instructions for adding the custom linuxcnc repositories took the safe option of adding them into a serpate sources.list file namely `/etc/apt/sources.list.d/linuxcnc.list`
> 
> yet these new instructions now have users directly modifying the base list file `/etc/apt/sources.list` (it even includes a waring to say messing with this is dangerous)
> 
> adding custom repositories via separate files under `/etc/apt/sources.list.d/sources.list.d/` is I believe the preferred way / best practice

Thaats exactly what this line does
`sudo nano /etc/apt/preferences.d/99linuxcnc-uspace`
Ref: https://github.com/LinuxCNC/linuxcnc/blob/2.9/docs/src/getting-started/getting-linuxcnc.adoc?plain=1#L254

Are you referring to another section of the code?

rodw-au:
My mistake, you  are looking here:
https://github.com/LinuxCNC/linuxcnc/blob/2.9/docs/src/getting-started/getting-linuxcnc.adoc?plain=1#L295
I am no expert with apt. Can that be moved to /etc/apt/preferences.d/99linuxcnc-uspace?

dpslwk:
im talking about this line
https://github.com/LinuxCNC/linuxcnc/blob/c74e11b05a23d212bcb4bc0a6247c176fbd40817/docs/src/getting-started/getting-linuxcnc.adoc?plain=1#L295

I would have that line as
`sudo echo "deb http://buildbot2.highlab.com/debian/ bookworm 2.9-uspace" | sudo tee -a /etc/apt/sources.list.d/linuxcnc.list`

and the warning on this line about it is then not needed
https://github.com/LinuxCNC/linuxcnc/blob/c74e11b05a23d212bcb4bc0a6247c176fbd40817/docs/src/getting-started/getting-linuxcnc.adoc?plain=1#L287



I'm not so familiar with  `/etc/apt/preferences.d/` but I believe all that is OK

rodw-au:
Thanks.

From what I can see, files in preferences.d are parsed in alphabetical order, so If its spread over more than one file, I think we should be using 10linuxcnc.list or similar so it is loaded before the pinning priorities are set. It makes sense for any pinned repos to be actioned last so two files are required in case there are other files added. (I did observe files for Google Chrome and Teamviewer on one Bullseye PC I have.)

I could not find where the repositories are set in the linuxcnc repos for guidance. I thought they might have been under the live-build repo but could not find them. I have not installed Buster for years... Can anyone advise where to look (or share the file here)?

It would be nice if the additional files could be hosted on the linuxcnc domain so we could just use wget. Can this be done?

This is all new now its in a distro's repository :)

dpslwk:
no the line
`deb http://buildbot2.highlab.com/debian/ bookworm 2.9-uspace` 
needs to be in a file under `/etc/apt/sources.list.d/`(what im suggesting)  or added to `/etc/apt/sources.list` (as currently written)

it will not work in a file under `/etc/apt/preferences.d/`

you know what ill just file a quick PR

rodw-au:
Sorry, my mistake. I thought it was under the same folder. Makes sense now.

hansu:
@petterreinholdtsen I wonder whether it's correct that the list "Alternate Install Methods" is that short now.
You extended the list in https://github.com/LinuxCNC/linuxcnc/pull/2199/files and maybe those changes should be re-added?

rodw-au:
My trimming was done without knowledge of #2199 but is similar in the changes.
I left Bullseye off the list because there was no method to install it from repositories and no ISO for it.

Even with the new buildbot supporting it, I think that still makes sense because the 5.x kernels in Bullseye will be more likely to cause problems than Bookworm's 6.x kernels (I have tested every released kernel from 5.10 to 6.1x and even past that eg. 6.2 and 6.3).

We deliberately refer people to the 2.8 Docs if the user wants to use kernels that are over 10 years old. 
Adding them  back in is only going to add clutter for new users because of the number of gotchas a user could face installing linuxcnc from the Bookworm repos.

Make no mistake this was a major rewrite and it had to cull a lot of obsolete info to make room for the necessary new content.

If interested, I found this video about installing Bookworm which I thought was very well done over the weekend. I intend to share it on the forum.
https://youtu.be/gddlhr9ST9Y

PR #2525, opened 2023-06-17
interp: fix GE/LE operators to handle double-precision comparison

snowgoer540:
This commit does four things:
1. Changes the LE/GE operators to better handle double-precision  floating-point comparisons. When comparing two values as equal (or not  equal) the EQ and NE operators considered the floating-point precision  "issue" by comparing the difference between the two values to  TOLERANCE_EQUAL, GE and LE did not.
2. Simplify the difference code by using fabs()
3. Modify the TOLERANCE_EQUAL threshold to 1e-6.  Previously this was  set to .0001 which is very close to a reasonable difference between two  gcode coordinate values.
4. Update the documentation to reflect these change.

### DIFF
--- a/docs/src/gcode/overview.adoc
+++ b/docs/src/gcode/overview.adoc
@@ -704,11 +704,10 @@ false, and any non-zero number is equivalent to logical true.
 
 == Equality and floating-point values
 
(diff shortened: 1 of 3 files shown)


rene-dev:

PR #2526, opened 2023-06-17
Corrected pin names and formatting of motion(9) man page

petterreinholdtsen:
The pin names for acc-cmd and vel-cmd had an incorrect joint- prefix. Verified the correct name looking at src/emc/motion/motion.c.

In the process, fixed incorrect formatting of is-unlocked pin in manual page to match other pins.

### DIFF
--- a/docs/man/man9/motion.9
+++ b/docs/man/man9/motion.9
@@ -276,8 +276,8 @@ The velocity limit for the teleop planner
 and are subject to change or removal at any time.)
 
 .TP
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
[Rene Hopf]
> Merged #2526 into 2.8.

I suspect the change should propagate into 2.9 and master too.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
>I suspect the change should propagate into 2.9 and master too.

done

PR #2528, opened 2023-06-18
add optional p word to set peck distance/count

rene-dev:
adds optional p word to https://github.com/LinuxCNC/linuxcnc/pull/1860/

### DIFF
--- a/docs/src/gcode/g-code.adoc
+++ b/docs/src/gcode/g-code.adoc
@@ -1757,7 +1757,7 @@ It is an error if:
 
 [source,{ngc}]
 ----
-G73 X- Y- Z- R- Q- <L->
+G73 X- Y- Z- R- Q- P- <L->
 ----
 
 * 'R' - retract position along the Z axis.
@@ -1766,7 +1766,7 @@ G73 X- Y- Z- R- Q- <L->
 
 The 'G73' cycle is drilling or milling with chip breaking.
 This cycle takes a Q number which represents a 'delta' increment along
-the Z axis.
+the Z axis. Peck clearance can be specified by optional P number.
 
 * Preliminary motion.
   ** If the current Z position is below the R position, The Z axis does
@@ -2371,7 +2371,7 @@ seconds at the bottom of the hole.
 
 [source,{ngc}]
 ----
-G83 (X- Y- Z-) or (U- V- W-) R- L- Q-
+G83 (X- Y- Z-) or (U- V- W-) R- L- Q- P-
 ----
 
 The 'G83' cycle (often called peck drilling) is intended for deep
@@ -2381,7 +2381,8 @@ drilling in aluminum). This cycle takes a Q number which represents a
 'delta' increment along the Z-axis. The retract before final depth will
 always be to the 'retract' plane even if G98 is in effect. The final
 retract will honor the G98/99 in effect. G83 functions the same as G81
-with the addition of retracts during the drilling operation.
+with the addition of retracts during the drilling operation. Peck clearance
+can be specified by optional P number.
 
 * Preliminary motion, as described in the
   <<gcode:preliminary-motion,Preliminary and In-Between Motion>> section.
--- a/docs/src/config/ini-config.adoc
+++ b/docs/src/config/ini-config.adoc
@@ -541,7 +541,11 @@ The maximum number of `USER_M_PATH` directories is defined at compile time (typ:
   Allow to clear the G92 offset automatically when config start-up.
 * `DISABLE_FANUC_STYLE_SUB = 0` (Default: 0)
   If there is reason to disable Fanuc subroutines set it to 1.
-
+* 'PARAMETER_G73_PECK_CLEARANCE = .020' (default: Metric machine: 1mm, imperial machine: .050 inches)
+  Chip breaking back-off distance in machine units
+* 'PARAMETER_G83_PECK_CLEARANCE = .020' (default: Metric machine: 1mm, imperial machine: .050 inches)
+  Clearance distance from last feed depth when machine rapids back to bottom of hole, in machine units.
+ 
 [NOTE]
 ====
 The above six options were controlled by the `FEATURES` bitmask in versions of LinuxCNC prior to 2.8.
--- a/src/emc/rs274ngc/interp_cycles.cc
+++ b/src/emc/rs274ngc/interp_cycles.cc
@@ -154,9 +154,7 @@ int Interp::convert_cycle_g83(block_pointer block,
      Thanks to Billy Singleton for pointing it out... */
   CHKS((delta <= 0.0), NCE_NEGATIVE_OR_ZERO_Q_VALUE_USED);
 
-  rapid_delta = _setup.parameter_drill_cycle_chip_break_distance;
-  if (_setup.length_units == CANON_UNITS_MM)
-    rapid_delta = (rapid_delta * 25.4);
+  rapid_delta = _setup.parameter_g83_peck_clearence;;
 
   for (current_depth = (r - delta);
        current_depth > bottom_z; current_depth = (current_depth - delta)) {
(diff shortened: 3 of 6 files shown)


hansu:
Is it intended that this PR includes Chris' commits?

petterreinholdtsen:
[Rene Hopf]
> adds optional p word to https://github.com/LinuxCNC/linuxcnc/pull/1860/

Nice.

Hos is this related to other G-code dialects?

Note, If I am to believe an old Norwegian Fanuc manual, its G-code
dialect used Q, F and L with G73 and G83, but no P.  Do not see a point
in quoting Norwegian only texts here.  I did find this english page
related to Fanuc,
<URL: https://www.helmancnc.com/g83-peck-drilling-cycle-deep-hole-for-fanuc/ >:

  Z = Z-depth (feed to Z-depth starting from R plane)
  R = Position of the R plane
  Q = Depth of cut for each cutting feed (depth of each peck)
  F = Cutting feedrate
  K = Number of repeats (if required)

In <URL: https://gcodetutor.com/fanuc-training-course/g73-g83-drilling-cycle.html >
I find these:

  Z = Depth of hole
  I = Size of first cutting depth
  J = Amount of reduction of each depth of peck
  K = Minimum peck depth
  R = Retract value
  P = Dwell on last peck
  F = Feed rate

Simliar in
<URL: https://www.cnccookbook.com/g81-g73-g83-drill-peck-canned-cycle/ >:

  Z = Hole bottom
  R = Retract position in Z. Motions from initial Z to R are performed
      at rapids speeds. From R to hole bottom is done at feed speed.
  P = Dwell time at bottom of hole.
  Q = Depth to increase on each peck.
  F = Cutting feedrate
  L = Number of repeats

Also found <URL: https://www.machinistguides.com/g83-code/ >:

  F = Feedrate
  G98 = Return to initial plane
  G99 = Return to R plane
  R = Position of the R plane (clearance level)
  X = Hole position in X-Axis
  Y = Hole position in Y-Axis
  Z = Position of the Z Axis at the bottom of hole
  Q = Depth to increase on each depth
  P = Dwell time at the bottom of the hole
  K or L = Number of cycle repetitions

Is the peck clearence similar to R or K here, or is the P a new
parameter?  It seem to be in conflict with 'dwell time at the bottom' on
some machines.

-- 
Vennlig hilsen
Petter Reinholdtsen

andypugh:
My understanding of these commits is that it makes the following statements in the docs a bit more explicit:
"Rapid up a bit." (G73)
"Rapid move back down to the current hole bottom, backed off a bit." (G83)
In that you can now decide how big "a bit" is. 

In the Fanuc docs is says that this is a "machine parameter", and that is equivalent to what Chris' initial commit implemented.

petterreinholdtsen:
[Andy Pugh]
> My understanding of these commits is that it makes the following
> statements in the docs a bit more explicit:
> "Rapid up a bit." (G73)
> "Rapid move back down to the current hole bottom, backed off a bit."
> (G83) In that you can now decide how big "a bit" is.

Mine too.  Note in the 1980 edition of the Fanuc doc I understand it to
move completely out of the hole every time, so "a bit" is back to the
position at the start of the drilling.

> In the Fanuc docs is says that this is a "machine parameter", and that
> is equivalent to what Chris' initial commit implemented.

Yeah.  The issue I wanted to raise is more if P is a good word for this
parameter, as it seem to be used elsewhere with G83 to mean something
different, not if it should be a machine parameter.

-- 
Happy hacking
Petter Reinholdtsen

docs/src/gcode/g-code.adoc
```diff
@@ -1766,7 +1766,7 @@ G73 X- Y- Z- R- Q- <L->
 
 The 'G73' cycle is drilling or milling with chip breaking.
 This cycle takes a Q number which represents a 'delta' increment along
-the Z axis.
```
petterreinholdtsen, line 1769: I suspect the "Rapid up .010 of an inch or 0.254 mm." part below should be modified too.  The documented  0.1"/0.254mm do not seem to match the ini value default listed above.

petterreinholdtsen:
I noticed in <URL: https://yewtu.be/watch?v=XRNnWAUoXrk > that FreeCAD
have a 'Retract' length that can be set when specifying peck drilling.
Is this the same as the R word in LinuxCNC?  The comment in the video
made me believe it was the length retracted per peck, which seem to be
the same as the value discussed in this pull request.
-- 

Happy hacking
Petter Reinholdtsen

andypugh:
> I noticed in <URL: https://yewtu.be/watch?v=XRNnWAUoXrk > that FreeCAD have a 'Retract' length that can be set when specifying peck drilling. Is this the same as the R word in LinuxCNC? 

Probably. I think the only way to be sure would be to change it and see how the generated G-code differs. 
...
Amongst the ways to be sure include looking at the FreeCAD source code. 

But I would expect that it is R

andypugh:
Maybe it would be better to leave P available in case we want to add a pause at final depth. 
However, there are not many letters available to use and inside the block structure only some letters are a suitable type.

  double a_number;
  double b_number;
  double c_number;
  double d_number_float;
  int dollar_number;
  double e_number;
  double f_number;
  int h_number;
  double i_number;
  double j_number;
  double k_number;
  int l_number;
  int n_number;
  double p_number;
  double q_number;
  double r_number;
  double s_number;
  int t_number;
  double u_number;
  double v_number;
  double w_number;
  double x_number;
  double y_number;
  double z_number;

abcuvwxyz are unavailable. As are F, S. N, T. 
H, L, N, T are also the wrong data type. 
R, L, Q are already in use. 

which leaves D, E, I, J, K, P. 

D might have been a better choice? (I have to take the blame here, as @rene-dev  and I discussed this and it was I who decided that P was best)

petterreinholdtsen:
I agree that it seem best to leave P unused to allow for a dwell/pause
at the bottom of holes in the future.

I came across
<URL: https://www.haascnc.com/service/codes-settings.type=gcode.machine=mill.value=G73.html > and
<URL: https://www.haascnc.com/service/codes-settings.type=gcode.machine=mill.value=G83.html >
which describe the options used by Haas.

G73:

 F - Feedrate
 * I - First peck depth
 * J - Amount to reduce pecking depth for pass
 * K - Minimum peck depth (The control calculates the number of pecks)
 * L - Number of loops (Number of holes to drill) if G91 (Incremental Mode) is used
 * P - Pause at the bottom of the hole (in seconds)
 * Q - Peck Depth (always incremental)
 * R - Position of the R plane (Distance above part surface)
 * X - X-Axis location of hole
 * Y - Y-Axis location of hole
 Z - Position of the Z-Axis at the bottom of hole

G83:

  * E - Chip-clean RPM (Spindle reverses to remove chips after each cycle)
  F - Feedrate
  * I - Size of first peck depth
  * J - Amount to reduce peck depth each pass
  * K - Minimum depth of peck
  * L - Number of holes if G91 (Incremental Mode) is used, also G81 through G89.
  * P - Pause at end of last peck, in seconds (Dwell)
  * Q - Peck depth, always incremental
  * R - Position of the R plane (position above the part)
  * X - X-Axis location of hole
  * Y - Y-Axis location of hole
  Z - Position of the Z-Axis at the bottom of hole

* indicates optional

Could it be an idea to reuse Q instead of introducing a new parameter,
in other words retract the peck depth (back to the earlier stop point)
when pecking?

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
Via <URL: https://github.com/FreeCAD/FreeCAD/pull/7185 > I discovered
that FreeCAD calculate the default CHIPBREAKING_AMOUNT to Q * 0.05 if Q
is set.  Perhaps this is a better default for LinuxCNC too, instead of
0.05" / 1mm?

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[andypugh]
> I think it is .01" or 0.254mm by default. 1mm seems too much for a
> default. I often peck with a Q of 1mm, in that case the system would
> spend equal time cutting air as cutting metal.

OK.  I probably misunderstood the values of
_setup.parameter_g73_peck_clearance and
_setup.parameter_g83_peck_clearance of src/emc/rs274ngc/rs274ngc_pre.cc
in the proposed patch, ie .050 for inches and 1 for the non-inches case.

-- 
Happy hacking
Petter Reinholdtsen

gmoccapy:
stuttgart meeting says we(andy) should change the letter.

petterreinholdtsen:
[Norbert Schechner]
> stuttgart meeting says we(andy) should change the letter.

Which letter should we change to?

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
I think probably D. Unless anyone wants to propose another of the available and suitable pins:
D, E, I, J, K, P.

petterreinholdtsen:
[Andy Pugh]
> I think probably D. Unless anyone wants to propose another of the
> available and suitable pins: D, E, I, J, K, P.

Aha.  As it look like a7909cbba10d109f47e2980ec1716ee45e62f8b2 went into
master using P for this, it is unclear how this will be done.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
> Aha. As it look like [a7909cb](https://github.com/LinuxCNC/linuxcnc/commit/a7909cbba10d109f47e2980ec1716ee45e62f8b2) went into master using P for this

Well, that's confusing because I changed eveything to D and reworded the docs before pushing.

smoe:
Regards from our Video conference - @smoe volunteered to rebase this branch to the current master so we learn about the bits that are or are not yet merged in via some other route. Reminder: Andy prefers P -> **D**.

PR #2529, opened 2023-06-19
2.9 Update getting-linuxcnc.adoc: use sources.list.d

dpslwk:
as per my comments on #2522


rodw-au:
@SebKuzminsky It would be great if you could commit this change as this will give a simple way also support other Debian versions to access your new buildbot

SebKuzminsky:
Thanks folks!

PR #2531, opened 2023-06-19
Implement new standardized pin worklight.

petterreinholdtsen:


SebKuzminsky:
I think this would be better done entirely in HAL, without involving Task and IO.

hansu:
Anyone an idea why the build fails since the last merge to master?

petterreinholdtsen:
[c-morley]
> On advantage of using incontrovertible is that the process becomes
> scandalized.

Can you explain what you mean with the above sentence for us non-native
speakers?

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
Sorry...bad typing too.
One advantage of using nml messages with iocontrol is that then the HAL pins/behavior become standardized. Ie all GUIs can expect and  connect to standardized pins with standardized behavior. 

It's already painful for users to switch GUIs because pin names and behavior is often different.

petterreinholdtsen:
[Hans Unzner]
> Anyone an idea why the build fails since the last merge to master?

No idea.  Look like the build hosts on github are unable to download the
2 MiB po4a .deb from snapshots.debian.org.  The file is present and it
get 95% of the way before it fail.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[c-morley]
> One advantage of using nml messages with iocontrol is that then the
> HAL pins/behavior become standardized. Ie all GUIs can expect and
> connect to standardized pins with standardized behavior.

My goal with this patch is twofold.

 - Provide a standardized pin for work light, as several machines seem
   to have work lights hooked up to some pin but everyone hook it up
   differently.

 - Use the standardised pin in GUIs, at least Axis and Gmoccapy to
   provide a on/off switch for the work light alongside the flood
   coolent button.

I do not know if my approach is the best one, and am happy to test an
alternative approach.  I went with the current lube/flood/mist approach,
as it is standardized and handled by several GUIs.

> It's already painful for users to switch GUIs because pin names and
> behavior is often different.

Yeah.

-- 
Happy hacking
PEtter Reinholdtsen

hansu:
> I went with the current lube/flood/mist approach,
as it is standardized and handled by several GUIs.

But isn't it that what @rene-dev wants to remove/change in https://github.com/LinuxCNC/linuxcnc/pull/2471 ?

c-morley:
Yes Rene wants to remove it. It's one of the reasons I commented. I believe Rene major complaint is its a lot of code for such simple behavior, also a couple of the pins have potential to be much more useful but currently just follow the estate. It would be much simpler with HAL which of course he is right. But I am not sure having every GUI make their own pins is a good idea. Also makes it so only one ui can control these pins. 

I think it might be better to figure out how to lessen the code burden and still use NML.

And or maybe we should talk about the larger idea of io control.

hansu:
Can the light also be switched on when the machine is switched off (not enabled)?

andypugh:
Ref #2531: This seems like an immense amount of code just to turn the worklikght on and off. 
I think that this is mainly because pins of this class are controlled via NML messages and the realtime motion queue. 
There probably isn't a pressing reason to turn the worklight on and off in perfect synchronisation with the axis motion, which is what this achieves. 
(There is an argument that this is necessary for the coolant controls, though) 

iocontrol and this whole system of IO pin control is already gone from the master branch, so your changes would be relatively short-lived (and will cause some minor work resolving merge conflicts into master because of that) 

I don't want to merge this into the upcoming 2.9.1 release, which is intended to be a bugfix-only release for 2.9.0 (the major bug in the 2.9.0 release is that it didn't get released :-) 

I think this touches too much of the base code of the stable 2.9 branch, and isn't applicable to the development branch.

petterreinholdtsen:
Perhaps someone can provide some recommendation how to properly
standarize a pin for worklight?

I agree that it has a huge amount of code for a simple feature, but
found no obvious way to make it simpler.
-- 
Happy hacking
Petter Reinholdtsen

gmoccapy:
stuttgart meeting agrees this needs to wait until we figure out a more general way of defining pins. see https://github.com/LinuxCNC/linuxcnc/issues/2532

heeplr:
Would anyone following this please explain briefly, why the ```emcStatus``` buffer couldn't be made R/W and ```worklight``` added as another attribute of ```emcioStatus``` but without adding new message types?

I'm probably missing something but not issuing EMC_WORKLIGHT_* messages and just manipulating emcioStatus.worklight buffer and trigger ```EMC_IO_STAT``` updates seems to be less hassle.

If I understand this correctly, currently the worklight status can be
* read from ```emcStatus.worklight.on``` which is updated with ```worklight.update(cms)``` in ```EMC_IO_STAT::update()```,
* or with ```EMC_WORKLIGHT_STAT``` to read the state without getting the complete IO state (why would I want that, it's small?)
* and written by sending an ```EMC_WORKLIGHT_ON/OFF``` messages. 

Instead of just 
```
// read
state = emcioStatus.worklight.on;
// write
emcioStatus.worklight.on = state;
````

What am I missing here? Any quick hint in the right direction would be highly appreciated.

PR #2533, opened 2023-06-20
Corrected internal Axis widget name of Coolant label

petterreinholdtsen:
The label 'Coolant:' was misnamed lubel.  It is not related to lubing,
but to cooling, so rename to 'coolantl' instead.  Adjusted
Axis simulation configs to also simulate coolant to get the label
to show up during simulation.

### DIFF
--- a/configs/sim/axis/axis.ini
+++ b/configs/sim/axis/axis.ini
@@ -118,6 +118,7 @@ HALFILE = sim_spindle_encoder.hal
 HALFILE = axis_manualtoolchange.hal
 HALFILE = simulated_home.hal
 HALFILE = check_xyz_constraints.hal
+HALFILE = cooling.hal
 
 # list of halcmd commands to execute
 # commands are executed in the order in which they appear
--- a/configs/sim/axis/axis_mm.ini
+++ b/configs/sim/axis/axis_mm.ini
@@ -109,6 +109,7 @@ HALFILE = core_sim.hal
 HALFILE = sim_spindle_encoder.hal
 HALFILE = axis_manualtoolchange.hal
 HALFILE = simulated_home.hal
+HALFILE = cooling.hal
 
 # list of halcmd commands to execute
 # commands are executed in the order in which they appear
--- a/configs/sim/axis/cooling.hal
+++ b/configs/sim/axis/cooling.hal
@@ -1,0 +1,7 @@
+# Fake the existance of coolant options
+loadrt and2 count=3
+addf and2.0 servo-thread
+addf and2.1 servo-thread
+net flood iocontrol.0.coolant-flood => and2.0.in0
+net mist iocontrol.0.coolant-mist => and2.0.in1
+net lube iocontrol.0.lube => and2.1.in0
--- a/configs/sim/axis/historical_lathe.ini
+++ b/configs/sim/axis/historical_lathe.ini
@@ -45,6 +45,7 @@ HALFILE = core_sim.hal
 HALFILE = axis_manualtoolchange.hal
 HALFILE = simulated_home.hal
 HALFILE = lathe.hal
+HALFILE = cooling.hal
 POSTGUI_HALFILE = lathe_postgui.hal
 
 [TRAJ]
--- a/src/emc/usr_intf/axis/scripts/axis.py
+++ b/src/emc/usr_intf/axis/scripts/axis.py
@@ -1359,7 +1359,7 @@ widget_list=[
 
        ("ajogspeed", Entry, pane_top + ".ajogspeed"),
 
-       ("lubel", Label, tabs_manual + ".coolant"),
+       ("coolant", Label, tabs_manual + ".coolant"),
        ("flood", Checkbutton, tabs_manual + ".flood"),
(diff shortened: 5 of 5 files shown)


SebKuzminsky:
I have not tested this, but it looks like a good bug fix to me.

SebKuzminsky (approved):

hansu:
Is there still an indicator for coolant present or was that replaced now by the lube indicator?

andypugh:
Is it possible that it was intended to be a low lube level warning indicator? (Purely based on there being an "L" at the end)

petterreinholdtsen:
[Hans Unzner]
> Is there still an indicator for coolant present or was that replaced
> now by the lube indicator?

There are still two, one flood and one mist.  At least that is how I
understand the forget() lines in axis.py.

[Andy Pugh]
> Is it possible that it was intended to be a low lube level warning
> indicator?  (Purely based on there being an "L" at the end)

No idea, but that make sense to me.  Perhaps a better indicator to have
on the display than the 'lube' pin?

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
I don't know why it is called "lubel" but isn't it just the label that says "Coolant" to indicate what the checkboxes are for?

petterreinholdtsen:
[Andypugh]
> I don't know why it is called "lubel" but isn't it just the label that
> says "Coolant" to indicate what the checkboxes are for?

Perhaps.  The confusing part is that the same pins seem to already be
handled by the Flood and Mist pins.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
I think that the intent is to just not bother with the label if there are no coolant controls to label?

Or maybe I am completely misunderstanding the change?

hansu:
Please do NOT merge this.
I agree the name `lubel` seems to be wrong or at least misleading, but the functionality is correct.
`lubel` is the heading-like label for coolant and/or flood. 
Your change gives Mist and Flood a label named "lube:" which is controlled by the lube pin. This makes completely no sense.
See:
grafik

Original behaviour:
grafik

petterreinholdtsen:
[Hans Unzner]
> Please do NOT merge this.
> I agree the name `lubel` seems to be wrong or at least misleading, but
> the functionality is correct.

You are right.  I have completely misunderstood the code.  The 'Coolant'
string is simply a GUI header, not a checkbutton related to lube or
flood/mist.

I will rewrite to make the functionallity more clear.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
Btw, is there a simulation in configs/ that can be used to show the
Coolant label?

-- 
Happy hacking
Petter Reinholdtsen

hansu:
>Btw, is there a simulation in configs/ that can be used to show the
Coolant label?

Not as far as I know. You can use any axis sim config and connect one or both of the cooling pins.
I just added those dummy connections to make them show:

```
loadrt and2 count=3
addf and2.0 servo-thread
addf and2.1 servo-thread
net flood iocontrol.0.coolant-flood => and2.0.in0
net mist iocontrol.0.coolant-mist => and2.0.in1
net lube iocontrol.0.lube => and2.1.in0
```

andypugh:
> Btw, is there a simulation in configs/ that can be used to show the Coolant label?

I actually disagree with the decision to configure all (or at least most) of the  sim configs with a LIB:basic:hal.tcl rather than a HAL file that users can easily edit. 

One example config that  already has the pins connected and doesn#'t use the Tcl library HAL files is sim-axis-vismach-vmc_toolchange.

petterreinholdtsen:
[Andy Pugh]
> I actually disagree with the decision to configure all (or at least
> most) of the  sim configs with a LIB:basic:hal.tcl rather than a HAL
> file that users can easily edit.

I am sure this issue deserve its own issue or pull request to discuss
its merit, instead of doing so here in my pull request to try to improve
how to understand the 'lubel' tk widget.

I rewrote the patch to only rename the lubel name to coolantl, to reduce
the confusion.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
There appear to be no need to avoid calling the widget 'coolant', so I am preparing an updated patch doing this.

hansu:
I would not include the stupid quick and dirty workaround to display the cooling check boxes. This could be a bit misleading. Better would be to create a small pyVCP which has two leds showing the status mist and flood and connect that to the appropriate pins..

petterreinholdtsen:
[Hans Unzner]
> I would not include the stupid quick and dirty workaround to display
> the cooling check boxes. This could be a bit misleading. Better would
> be to create a small pyVCP which has two leds showing the status mist
> and flood and connect that to the appropriate pins..

Patches most welcome. :)

I included it to make it easier to see that the change did not break
anything.  Would be happy to move it into a separate pull request,
unless you beat me to it and create a better patch. :)

-- 
Happy hacking
Petter Reinholdtsen

hansu:
>Better would be to create a small pyVCP which has two leds showing the status mist and flood and connect that to the appropriate pins..

Not even this is needed like Andy mentioned. One can just use the sim/axis/vismach/VMC_toolchange sim config.

Issue #2505, opened 2023-05-24
Error in the documentation


Sigma1912:
This example line for a startup code is not valid and causes a config error on startup:
RS274NGC_STARTUP_CODE = G17 G20 G40 G49 G64 P0.001 G80 G90 G92 G94 G97 G98 

http://linuxcnc.org/docs/html/config/ini-config.html#sec:rs274ngc-section
http://linuxcnc.org/docs/devel/html/config/ini-config.html#sub:ini:sec:rs274ngc

The example should, probably, read:
RS274NGC_STARTUP_CODE = G17 G20 G40 G49 G64 P0.001 G80 G90 G92.1 G94 G97 G98

Note the G92.1 instead of G92 which expects axis words

Sigma1912:
Made a Pull Request #2535 regarding this

petterreinholdtsen:
Fixed in 17208cd636dc78c0dbfc9738cfa6558190c8d090 .

PR #2535, opened 2023-06-21
Fix example for RS274NGC_STARTUP_CODE in the documentation

Sigma1912:
Regarding issue #2505 

The current example line for a startup code is not valid and causes a config error on startup:
RS274NGC_STARTUP_CODE = G17 G20 G40 G49 G64 P0.001 G80 G90 G92 G94 G97 G98

http://linuxcnc.org/docs/html/config/ini-config.html#sec:rs274ngc-section
http://linuxcnc.org/docs/devel/html/config/ini-config.html#sub:ini:sec:rs274ngc

The example should, probably, read:
RS274NGC_STARTUP_CODE = G17 G20 G40 G49 G64 P0.001 G80 G90 G92.1 G94 G97 G98

Note the G92.1 instead of G92 which expects axis words

### DIFF
--- a/docs/man/man1/axis-remote.1
+++ b/docs/man/man1/axis-remote.1
@@ -66,7 +66,7 @@ None known at this time.
 .SH AUTHOR
 This man page written by Alex Joni, as part of the LinuxCNC project.
 .SH REPORTING BUGS
-Report bugs to alex_joni AT users DOT sourceforge DOT net
+Report bugs at https://github.com/LinuxCNC/linuxcnc/issues
 .SH COPYRIGHT
 Copyright \(co 2007 Alex Joni.
 .br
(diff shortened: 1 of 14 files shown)


andypugh:
I think that there is no doubt that you are correct, but that this change belongs in 2.9. Also, your PR seems to include a merge, so includes many more commits than you probably intended. 

Rather than ask you to re-create the commit, on a different branch, and with a clean history, I will make the change manually myself. 

For future reference, it is generally a good idea to "git pull --rebase" before creating a new commit. 
Also, I like to "git push --dry-run" first, to see if the commit looks like I expect. To be extra careful I will often copy the commit hash range that is printed by the dry-run (eg 123456...789abc) and then "git log -p  123456...789abc" to see, line by line, exactly what will be committed.

andypugh:
Fixed via https://github.com/LinuxCNC/linuxcnc/commit/17208cd636dc78c0dbfc9738cfa6558190c8d090

PR #2536, opened 2023-06-21
Corrected internal Axis widget name of Coolant label

petterreinholdtsen:
The label 'Coolant:' was misnamed lubel.  It is not related to lubing, but to cooling, so rename to 'coolantl' instead.  Adjusted Axis simulation configs to also simulate coolant to get the label to show up during simulation.


petterreinholdtsen:
Created by mistake based on wrong git repo.

Issue #2489, opened 2023-05-11
Gmoccapy - Combination M61 + G4


zz912:
I use RIP instalation Branche 2.9, sim gmoccapy

If I enter a combination of M61 and G4 commands in the command line, sometimes the G4 command is executed, sometimes not.

Watch the video:

https://github.com/LinuxCNC/linuxcnc/assets/96618597/b8a00f0e-d288-4b07-a1a1-aa126b025976

zz912:
1) If you cannot view the video, please download it and play it in VLC player. On some PCs the video plays, on some PCs the video says that it is broken.

2) I think this bug is the cause of these bugs:
https://github.com/LinuxCNC/linuxcnc/issues/2453
https://github.com/LinuxCNC/linuxcnc/issues/2448
I tried to look for the cause of these bugs and found that the execution of MDI_COMMANDS is random. Maybe this bug isn't the cause, but I can't look any further.

3) I managed to simulate the bug:
- on a physical PC, RIP installation, branche 2.9
- on Oracle VM VirtualBox, RIP installation, branche 2.9 (reduce CPU is irrelevant)
- on Oracle VM VirtualBox, Debian installation, branche 2.8.4 (CPU reduction is irrelevant)

4) I have a problem that HansU wants to help me solve the previous bug, but he can't simulate it. Therefore, I would like to ask for feedback if this error also appears.

zz912:
I made tests of this bug:
2.8.0~pre3, GUI AXIS, BUG - NO
2.8.0~pre3, GUI GMOCCAPY, BUG - YES
2.8.0~pre3, GUI GSCREEN, BUG - NO
2.8.0~pre3, GUI SILVERDRAGON, BUG - NO

2.7.15,  GUI AXIS, BUG - NO
2.7.15,  GUI GMOCCAPY, BUG - NO

zz912:
The cause of this problem is in these lines:
https://github.com/LinuxCNC/linuxcnc/blob/56883b969838b721c0b0ebe12b9c14abf91d640e/src/emc/usr_intf/gmoccapy/gmoccapy.py#L3540-L3544

As these lines cause more problem, I made a HAL pin which disable these lines.
https://github.com/LinuxCNC/linuxcnc/pull/2571

Here is the test result:
![MDI_commands](https://github.com/LinuxCNC/linuxcnc/assets/96618597/bccb05aa-12ff-4141-b35f-ee1427aedc57)

[...]

Issue #2448, opened 2023-04-24
Gmoccapy - Remap - random bug


zz912:
I have a bug on a real machine:
https://forum.linuxcnc.org/38-general-linuxcnc-questions/48870-atc-random-bug

The good news is that I am able to simulate this error on a virtual machine.

Simulate the problem:
1) Download and Buid branche 2.9
2) I followed this:
http://linuxcnc.org/docs/2.9/html/remap/remap.html#remap:upgrading-an-existing

cd /home/user/linuxcnc/linuxcnc-2.9/configs/sim/gmoccapy/python/
rm remap.py
rm toplevel.py 
rm stdglue.py
cp ../../../../nc_files/remap_lib/python-stdglue/stdglue.py .
echo 'from stdglue import *' >remap.py
echo 'import remap' >toplevel.py

3) edit /home/zdenek/linuxcnc/linuxcnc-2.9/configs/sim/gmoccapy/gmoccapy.ini
add:

[RS274NGC]
REMAP = M6  modalgroup=6  prolog=change_prolog ngc=atc epilog=change_epilog

[PYTHON]
TOPLEVEL=./python/toplevel.py
PATH_APPEND=./python

4) copy  ngc files to to folder /home/zdenek/linuxcnc/linuxcnc-2.9/configs/sim/gmoccapy/macros/
[atc.ngc.txt](https://github.com/LinuxCNC/linuxcnc/files/11311261/atc.ngc.txt)
[atc-down.ngc.txt](https://github.com/LinuxCNC/linuxcnc/files/11311262/atc-down.ngc.txt)
[atc-up.ngc.txt](https://github.com/LinuxCNC/linuxcnc/files/11311264/atc-up.ngc.txt)
remove .txt :-)

5) run linuxcnc gmoccapy.ini
6) home all axes
7) in MDI try:
M6 T1
M6 T2
M6 T3
....
![subrutine_error](https://user-images.githubusercontent.com/96618597/234020019-0d5fca5e-a049-4cd0-b7e9-49c6330dd0ed.png)

Please be patient. On a physical machine, the error appears approximately once in 10 attempts, but on a virtual PC, this error appears approximately once in 40 attempts.

I wanted to make your work easier and write a program that would call M6 commands. Unfortunately, this error does not appear here. It is necessary to write M6 commands in MDI.

I have already cheered many times that I have removed the error with some modification of Remap, but in the end this message always appears.

zz912:
[...]

PR #2552, opened 2023-07-01
Gmoccapy - added self.command.wait_complete() + self.stat.poll()

zz912:
Hello everybody,

I added self.command.wait complete() + self.stat.poll() to avoid race condition.

This PR will not make any change in normal use, but should help in non-standard situations and/or complicated LCNC integrations.

Here is an example of how LCNC behaves when wait_complete() is omitted:
https://forum.linuxcnc.org/gmoccapy/49335-self-command-mdi-without-self-command-wait-complete?start=10#273798

It is very important to note that not using wait complete() can cause not only a bug in the given part of the code, but can cause unpredictable behavior of the entire LCNC. In this example, the LCNC remains in AUTO mode, even though this mode is not called.

If it seems to you that I used wait_complete() or self.stat.poll() only randomly and in as many places as possible, that's not the case. For each wait_complete() or self.stat.poll() I was wondering if it is necessary and if it will cause a downstream problem.

I think you can find cases where waiting_complete() or self.stat.poll() is duplicated unnecessarily. I'm convinced that it's better to use wait_complete() or self.stat.poll() multiple times than to miss somewhere.

Gmoccapy is often mentioned in the LCNC manual as a template for developing custom GUIs or custom LCNC modules. That's why I also added wait_complete() or self.stat.poll() to every function where it was needed. If someone copies such a function into their project, it will contain wait_complete() or self.stat.poll().

I know self.stat.poll() is periodic in the function so it should still be current, but with this PullRequest I'm trying to eliminate random errors.

Zdenek

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -578,6 +578,7 @@ class gmoccapy(object):
         # the size and digits of the DRO
         # set default values according to the machine units
         digits = 3
+        self.stat.poll()
         if self.stat.linear_units != _MM:
             digits = 4
         self.dro_digits = self.prefs.getpref("dro_digits", digits, int)
@@ -1158,6 +1159,7 @@ class gmoccapy(object):
         self.widgets.vbtb_jog_incr.set_property("name","jog_incr_buttons")
 
     def _jog_increment_changed(self, widget,):
+        self.stat.poll()
         # first cancel any joints jogging
         JOGMODE = self._get_jog_mode()
         if self.stat.task_mode == linuxcnc.MODE_MANUAL:
@@ -1169,6 +1171,7 @@ class gmoccapy(object):
 
     def _on_btn_jog_pressed(self, widget, button_name, shift=False):
         LOG.debug("Jog Button pressed = {0}".format(button_name))
+        self.stat.poll()
 
         # only in manual mode we will allow jogging the axis at this development state
         # needed to avoid error on start up, machine not on
@@ -1216,6 +1219,8 @@ class gmoccapy(object):
 
     def _on_btn_jog_released(self, widget, button_name, shift=False):
         LOG.debug("Jog Button released = {0}".format(button_name))
+        
+        self.stat.poll()
         # only in manual mode we will allow jogging the axis at this development state
         if not self.stat.enabled or self.stat.task_mode != linuxcnc.MODE_MANUAL:
             return
@@ -1231,6 +1236,7 @@ class gmoccapy(object):
             self.command.jog(linuxcnc.JOG_STOP, JOGMODE, joint_no_or_axis_index)
 
     def _get_jog_mode(self):
+        self.stat.poll()
         # self.stat.motion_mode ==
         # 1 = Joint
         # 2 = MDI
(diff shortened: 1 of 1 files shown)


gmoccapy:
Please do not push!

Just adding stat.poll and wait complete to nearly every place in the code is not only a very bad coding style, but als unnessasary and sure not the solution to the described situation. Additionally as the problem can not be reproduced on "normal" machines.

I will try to review the code, but will need some time.

Norbert

zz912:
>Please do not push!

Sorry if I push anyone. I appreciate the work of all people from this community.
This coding style is the coding style of the desperate (me).
I retrofitted the machine to LCNC and can't finish it.
The solution would be to use the AXIS GUI, but I don't like that.
I will use these modifications on my machines for now. I just wanted to offer them to others.
It is up to you whether you use any of my suggestions or not.
You can easily throw away the entire PR. I will not mind.

> Additionally as the problem can not be reproduced on "normal" machines.

I can reproduce 3 random bug on "simulation"  machines.
https://github.com/LinuxCNC/linuxcnc/issues/2448
https://github.com/LinuxCNC/linuxcnc/issues/2489

If there are bugs in the simulation, I believe they will show up in "normal" machines as well.

This PR won't fix those bugs, but it might help.

c-morley:
'Please do not push!'  means Norbert does not want the changes 'pushed' to the repository. I don't think he means that you are being 'too pushy' about your continued investigation.

gmoccapy:
I did not want to offent anybody or critisise the ongoing investment!
Just I did not want the changes to be pushed in the way it is done actauly!

I agree, that adding a wait complite in many places would be a good way to test for a changed behavior, but adding poll all around will slow the reaction of the GUI down in a unnecessary way.

Unfortunately I do not have time to check that in detail. I will meet (hope in september) with some other LinuxCNC people for a weekend. I have a list of bugs in gmoccapy I want to investigate about and solve them if possible.

Norbert

zz912:
Thank you for your response. I would blame the problem of misunderstanding on Google translator. I think you understand my idea, but you want to do the implementation differently. That's why I'm closing this PR and I'll look forward to the implementation from you when you have time for it.

It's good that Norbert has a list of bugs. Unfortunately, it's a shame that it's not public. I wouldn't have to point out the mistakes he knows about.

gmoccapy:
As this is not solved, I do reopen it.
@zz912 
The "bug" and feature request list is public ;-)
Just see the issues list of this git repo.

Norbert

Sigma1912:
This may be clear to everybody else but I just had a bit of a lightbulb when researching this issue:
https://github.com/LinuxCNC/linuxcnc/issues/3129

It seems that adding  'self.command.wait_complete()'  might actually block the GStat module from polling and thus miss changes in the state which cam then lead to Gmoccapy (which relies on GStat messages for its event driven architecture) not receiving the expected messages.

I suppose this applies to any python script running (eg python remaps, vcp panels, handlers, etc)

Sigma1912:
given that the bugs stated in the opening post are marked closed can we close this as well?

zz912:
This Pull Request is within my capabilities at the time I made it. I was going to close it back then, but @gmoccapy reopened it, so I'm leaving it alone. I recommend closing it.

andypugh:
A question for @gmoccapy I think?

gmoccapy:
Ok, I will close this one now.

Norbert

PR #2553, opened 2023-07-01
Fix typos in src/ subdirectory

luzpaz:
Found via `codespell -q 3 -S *.po,*.pot,*.svg,*.ts,./.git/logs,./share,./docs/man/es,./configs/attic,*_fr.*,*_es.*,README_es,./src/hal/classicladder -L ans,ba,breal,bu,bulle,componentes,currenty,doubleclick,dout,dum,extint,faktor,fo,fpr,halp,ihs,inout,inport,mot,parm,parms,propertys,ro,seh,ser,smoe,te,tey,trough,ue,ure,wille,wont,wonte`

Follow-up to #2548

PR #2548, opened 2023-06-28
Fix typos in configs/ subdirectory

luzpaz:
Found via `codespell -q 3 -S *.po,*.pot,*.svg,*.ts,./.git/logs,./share,./docs/man/es,./configs/attic,./configs/sim/axis/orphans,*_fr.*,*_es.*,README_es,./src/hal/classicladder -L ans,aready,ba,bu,bulle,componentes,currenty,doubleclick,dout,dum,extint,faktor,fo,fpr,halp,ihs,inout,inport,parm,parms,ro,ser,te,tey,trough,ue,ure,wille,wont,wonte`

Edit: this PR is made against the 2.9 branch and supercedes #2545

PR #2560, opened 2023-07-02
Fix typos in src/ subdirectory

luzpaz:
Found via `codespell -q 3 -S *.po,*.pot,*.svg,*.ts,./.git/logs,./share,./docs/man/es,./configs/attic,*_fr.*,*_es.*,README_es,./src/hal/classicladder -L ans,ba,breal,bu,bulle,busses,componentes,currenty,doubleclick,dout,dum,extint,faktor,fo,fpr,halp,ihs,inout,inport,mot,parm,parms,propertys,ro,seh,ser,smoe,te,tey,trough,ue,ure,wille,wont,wonte`

Follow-up to #2548  
Supercedes #2553

### DIFF
--- a/src/hal/components/panelui.c
+++ b/src/hal/components/panelui.c
@@ -109,7 +109,7 @@ int nochange = 0x40;
 int rowshift;
 int ncols = 8;
 int nrows = 8;
-int rollover = 2; //maximuim number of simultaneous keys presses recognised
+int rollover = 2; //maximum number of simultaneous keys presses recognised
 int s = 0; //key press state argument to pass to python
 int r; //row argument passed to python
 int c; //column argument passed to python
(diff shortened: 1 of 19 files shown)


src/hal/components/panelui.c
```diff
@@ -109,7 +109,7 @@ int nochange = 0x40;
 int rowshift;
 int ncols = 8;
 int nrows = 8;
-int rollover = 2; //maximuim number of simultaneous keys presses recongnised
+int rollover = 2; //maximuim number of simultaneous keys presses recognised
```
hansu, line 112: I wonder if we should change that to the AE spelling like the majority of the documentation is written in?

hansu (approved):

PR #2562, opened 2023-07-02
Use G38.3 in qt_auto_probe_tool.ngc to be able to check probe status.

petterreinholdtsen:
There is a test after one of the G38.2 in question that look like this:

  O500 IF [#5070 EQ 0]
  G90
  O500 return [-3] ; indicate probe contact failure to epilog
  O500 endif

But the manual state that "If the probing operation failed, G38.2 and G38.4
will signal an error by posting an message on screen if the selected GUI
supports that. And by halting program execution.".  When I try to include a similar
test in my test program, the program is aborted and the test is never executed.
This make me believe the wrong G code is used.   Or is there something special
about the M6 remapping environment that make G38.2 behave differently?

Made sure both calls to G38.2 are changed to G38.2 with a test verifying the
status.

Also adjusted the handling of the measured height to take any coordinate system
offsets.

### DIFF
--- a/nc_files/remap-subroutines/qt_auto_probe_tool.ngc
+++ b/nc_files/remap-subroutines/qt_auto_probe_tool.ngc
@@ -2,6 +2,11 @@ o<qt_auto_probe_tool> sub
 ;(debug, in change tool_in_spindle=#<tool_in_spindle> current_pocket=#<current_pocket>)
 ;(debug, selected_tool=#<selected_tool> selected_pocket=#<selected_pocket>)
 
+; calculate Z offset value based on current work offset (#5220) and
+; G92/G52 offset if enabled (#5210)
+#<zworkoffset> = [#[5203 + #5220 * 20] + #5213 * #5210]
+
+
 #<tool> = #<selected_tool>
 #<pocket> = #<selected_pocket>
 ; we must execute this only in the milltask interpreter
@@ -61,22 +66,27 @@ O400 endif
 
 F #<_hal[qtversaprobe.searchvel]>
 G91
-G38.2 Z- #<_ini[VERSA_TOOLSETTER]MAXPROBE>
+G38.3 Z- #<_ini[VERSA_TOOLSETTER]MAXPROBE>
 G0 Z #<_hal[qtversaprobe.backoffdist]>
-
-;reprobe at probe speed
-F #<_hal[qtversaprobe.probevel]>
-G38.2 Z- [#<_hal[qtversaprobe.backoffdist]> *1.2]
 
 O500 if [#5070 EQ 0]
 G90
 O500 return [-3] ; indicate probe contact failure to epilog
 O500 endif
 
+;reprobe at probe speed
+F #<_hal[qtversaprobe.probevel]>
+G38.3 Z- [#<_hal[qtversaprobe.backoffdist]> *1.2]
+
+O600 if [#5070 EQ 0]
+G90
+O600 return [-3] ; indicate probe contact failure to epilog
+O600 endif
+
 G90
 G53 G0 Z[#<_ini[CHANGE_POSITION]Z>]
 
-#<touch_result> = #5063
+#<touch_result> = [#5063 + #<zworkoffset>]
 
 (DEBUG, %fProbe Height: #<_hal[qtversaprobe.probeheight]>)
 (DEBUG, %fBlock Height: #<_hal[qtversaprobe.blockheight]>)


c-morley:
These code changes seem good to me - have you tested the error process on a machine or sim?

petterreinholdtsen:
[c-morley]
> These code changes seem good to me - have you tested the error process
> on a machine or sim?

I could not get it to run in a simulator, and have not tested it on a
machine.  The setup in that directory did not work in the simulator with
or without my change, and I did not understand what the problem was.
-- 
Happy hacking
Petter Reinholdtsen

nc_files/remap-subroutines/qt_auto_probe_tool.ngc
```diff
@@ -61,22 +66,27 @@ O400 endif
...
+O600 if [#5070 EQ 0]
+G90
+O600 return [-3] ; indicate probe contact failure to epilog
+O600 endif
+
 G90
 G53 G0 Z[#<_ini[CHANGE_POSITION]Z>]
 
-#<touch_result> = #5063
+#<touch_result> = #5063 + #<zworkoffset>
```
c-morley, line 89: Need brackets around the addition:
#<touch_result> = [#5063 + #<zworkoffset>]

petterreinholdtsen:
[c-morley]
> Need brackets around the addition:

Ah, of course.  Thank you for noticing.  Fixed.

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
it seems there is also some needed changes in std_glue to get linuxcnc to notice the returned error numbers. Since it's a common library, I just don't know if it would break others relied on behavior.
Still checking that code change yet,

petterreinholdtsen:
[c-morley]
> it seems there is also some needed changes in std_glue to get linuxcnc
> to notice the returned error numbers. Since it's a common library, I
> just don't know if it would break others relied on behavior.

Which changes do you expect are needed?  Are we talking about
nc_files/remap_lib/python-stdglue/stdglue.py?  It seem to only check
return_value against 0.

Note, I did not add a new return value (-3), I only added another place
where the same return value was used.  Similar code was already in place
after the second G38.2.

> Still checking that code change yet,

Great to hear you are giving it a critical view. :)

Btw, Anyone know what the 'qt_' part of the qt_auto_probe_tool.ngc file
name mean?

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
QT because it was built for qtvcp's screens, though it should work for anyone.

gmoccapy:
stuttgart meeting says @c-morley should decide on this.

nc_files/remap-subroutines/qt_auto_probe_tool.ngc
petterreinholdtsen: Done.

VectorHasting:
I'm happy to test this, I'm also working on pull request #2706, and @c-morley flagged this request there and asked if I could test. I've been delayed, and I apologize.  

But I have questions: 

It looks like from the comment at the beginning that the intention was to _keep_ error codes... and I thought that was the difference between G38.2 and G38.3: that G38.2 returns an error code but G38.3 does not. The code has been changed **_to_** G38.3 at lines 69 and 79.  

But the comment at the beginning seems to include an ambiguous typo: 

 > Made sure both calls to G38.2 are changed to G38.2 with a test verifying the status.

Assuming what was meant is "change G38.2 to G38.3" I'm wondering if that will defeat getting an error message back??
Or, if the intention was to stop getting the error back, then why?

Those questions aside: how do I test this?

We are only talking about testing the QtDragon GUI, right? Or is this routine used in the other GUIs? (I've been using it on QtDragon)

My thought is to set up my machine so that the _ini[VERSA_TOOLSETTER]MAXPROBE is too small to get from the _ini[VERSA_TOOLSETTER]Z down to the probe surface. Then I would set myself up to issue a T2M6 command. That will call this remap code and I should get an error back to the GUI because it won't be able to get a probe signal while it's doing line 69 or line 79 (since the backoffdist will be heading up from the insufficiently low maxprobe).

Another thought: while I like very much the logical completeness of having the o600 section to check for the error after the reprobe... I nonetheless find myself wondering how that would actually every happen? There would have to be a probe failure of some kind between the success of line 69 and the start of line 79, otherwise 69 would have already failed. In fact, in order to test it, I think I would have to move the probe out of the way after the rapid probe and before the slow reprobe could get there. Hmmm. Not sure I want to try that, mine is bolted on an upright support. 

Which, btw, brings me to: I have already gotten this error loads of times when my maxprobe distance is insufficient to reach the tool-setter. So, again I'm wondering if this switch to G38.3 is what is intended here, since it's been working for me without this change. On the other-hand, I'm hesitant about my knowledge here, so I'm very ready to be set straight, and thank you in advance.

Please advise :-)

VectorHasting:
I've failed to post here for some time because I became sick and then disabled not long after my last post. I'm only posting now because this was merged (and also, I'm better). 

The good news is that this change is only to sample code which is practically un-usable as-is (because it does not return the tool to the original position when the M6 command was issued, see discussion on on pull request [#2706](https://github.com/LinuxCNC/linuxcnc/pull/2706) .

The bad news is it doesn't do what the poster intended: I have run the test, and shot a video, but failed to post it because I hadn't finished editing it because of my aforementioned health challenges. 

Ironically, I was about to make this post and post the video later today or tomorrow. I'm working on the video now and will post it forthwith, but thought I should post this before I fail to follow through. 

G38.2 **does** pass any error encountered during the probe movement back to the calling program, but G38.3 **does not**. 

I've verified that this code change in qt_auto_probe_tool.ngc does the opposite of what was intended: by using G38.3 in place of G38.2, it causes no error to be displayed (and sometimes these have been quite helpful) on the user interface. 

If possible, I would recommend reversing the merge. If it's not possible/practical, it's not a big deal since it shouldn't actually impact anyone's working code (because, again, as written, this is an unusable example snippit). 

I'll post the video on youtube and link to it here asap, and then also do the same for my changes on PR #2706 which _will_ make this remap code usable by returning the tool to the position when it was called, and that proposed change will also change back to using G38.2 instead of G38.3. 

Thanks, and sorry!

VectorHasting:
Here is the [long delayed video](https://youtu.be/UbcpHurjrGo), if anyone ever wants to watch testing for 19 min.

petterreinholdtsen:
Thank you for the very helpful video demostrating your experience, and
the qtdragon frontend I will have to test on my machine as it seem to
provide quite a lot of useful features.

Your video demonstrated two issues, one regarding the measuring result,
and one regarding error handling.  I have not yet looked at the
measuring result part yet, will look at it later.  I need to read up on
the topic again to remember why I believe the coordinate system offset
should be included in the tool measurement value.  I suspect I found an
example doing this to ensure consistent tool height measurements
independent of coordinate system offsets and coordinate system used.

Regarding the error handling, I believe the issue is caused by a
misunderstanding on your part on how the error reporting should be done.
This pull request changed the error handling to a consistent use of
programmatically reporting the error to the program calling the
qt_auto_probe_tool subroutine, instead of some errors being returned to
the program calling the qt_auto_probe_tool subroutine and two errors
causing a popup to show up and the program to be terminated.

Your issue seem to be that there is no longer a popup dialog and program
termination, which actually was the intended change to make the
subroutine work in a non-interactive setting and move the responsibility
for notifying the user to the caller of the subroutine.  Those using the
qt_auto_probe_tool subroutine always need to check the returned value
(notice the use of 'return [-N]' (with N being -1, -2, -3, 3 and 999)
sprinkled around the code, and handle any error appropriately, and can
not depend on popups and automatic termination of the program.  If you
want a popup from the subrouteine itself, all these error handling
blocks need to be changed and the subroutine become unusable in a
non-interactive setting.  How is your code calling the subroutine?  How
are you checking return values there, and why is it not poping up a
dialog when a error code is returned?

Note, I suspect returning 3 might be a typo and that it would be better
to return -3 to signal an error.

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
I suspect the returned value is supposed to be handled by the
prepare_epilog() method in nc_files/remap_lib/python-stdglue/stdglue.py:

def prepare_epilog(self, **words):
    try:
        if not self.value_returned:
            r = self.blocks[self.remap_level].executing_remap
            self.set_errormsg("the %s remap procedure %s did not return a value"
                             % (r.name,r.remap_ngc if r.remap_ngc else r.remap_py))
            return INTERP_ERROR
        if self.blocks[self.remap_level].builtin_used:
            #print "---------- T builtin recursion, nothing to do"
            return INTERP_OK
        else:
            if self.return_value > 0:
                self.selected_tool = int(self.params["tool"])
                self.selected_pocket = int(self.params["pocket"])
                emccanon.SELECT_TOOL(self.selected_tool)
                return INTERP_OK
            else:
                self.set_errormsg("T%d: aborted (return code %.1f)" % (int(self.params["tool"]),self.return_value))
                return INTERP_ERROR
    except Exception as e:
        self.set_errormsg("T%d/prepare_epilog: %s" % (tool,e))
        return INTERP_ERROR

I am not quite sure what will happen here, but I belive
self.value_returned will be false and self.return_value will be
negative.  As far as I can tell, it will return INTERP_ERROR after
setting an error message.  Is this not causing a popup?

Are you also using such epilog=prepare_epilog value in your M6
remapping?

-- 
Happy hacking
Petter Reinholdtsen

VectorHasting:
Dear @petterreinholdtsen,

thank you so much for this followup!

When I read the [G38 documentation](https://linuxcnc.org/docs/2.6/html/gcode/gcode.html#sec:G38-probe), it says "if the probing operation failed, G38.2 and G38.4 will signal an error by posting an message on screen if the selected GUI supports that." Obviously, this begs some questions. Is it saying G38.2 posts the message, or that it informs the GUI so it can post the message? Is it saying G38.3 and G38.5 don't send any message at all, or just that they don't put a message on screen? I believe what they mean from that sentence, and from the bullet points in that section, and from my testing, is that G38.3 does in fact send no error whatsoever. And that this is why the change to G38.3 is not what's intended. I'm guessing G38.3's use-case would be for an entirely interactive setting where an operator is looking straight at the system and just probing it. In that scenario, they won't want any error coming back because it's just more work to clear the screen because they saw the error right there in front of them. I dunno. 

What I do know, is what I did in the video. I'm using the QtDragon client, and I followed that interface's installation instructions [found here](https://linuxcnc.org/docs/devel/html/gui/qtdragon.html#_tool_measurement_pins). 

I know there's potential ambiguity about stdglue.py, so here's my background:

My .ini file line under the [RS274NGC] section is as follows: 

`REMAP=M6  modalgroup=6 prolog=change_prolog ngc=qt_auto_probe_tool         epilog=change_epilog`

(so yes, I use change_epilog)

I believe the path it uses to find stdglue.py is found in the [PYTHON] section, and mine is set as follows: 

`TOPLEVEL = /home/maker/linuxcnc/configs/SaberCNC_10/python/toplevel.py
PATH_APPEND=/home/maker/linuxcnc/configs/SaberCNC_10/python`

Which I believe means I'm resolving to the stdglue.py that was created by the pncconf wizard, which in my system is at: _~/linuxcnc/configs/SaberCNC_10/python_ .

I'm attaching my .ini file, and that stdglue.py file to this message, but I appended .txt to the filenames to allow them upload. 
[SaberCNC_10.ini.txt](https://github.com/user-attachments/files/16168829/SaberCNC_10.ini.txt)
[stdglue.py.txt](https://github.com/user-attachments/files/16168830/stdglue.py.txt)

As far as I can tell, the "prepare_epilog" method in my installed stdglue.py is exactly as you've described. 

And so then, as to your question: 
"I am not quite sure what will happen here, but I belive self.value_returned will be false and self.return_value will be negative.  As far as I can tell, it will return INTERP_ERROR after setting an error message.  Is this not causing a popup?" ...

... the answer is no, it is not sending a popup, nor is it showing in the status log.  
(This is consistent with the hypothesis that G38.3 actually is not sending any error code at all, just stopping on fail, as per the documentation cited above.)

Soooo, I kinda stand by my opinion that this pull request would be better reverted.

And yet! 
I think you raise a valid set of points regarding an overall philosophy on how to handle error messages. If this were a more controlled, less open-source context, I believe I would be on your side arguing (as I think you're arguing here) for a general philosophy that errors should be passed up and down the stack and handled as close to the 'user' as possible. And we would stipulate that a 'user' might be a human or some automated system that might be able to correct the error. 
(But this agreement is subject to the caveat of your assumption that G38.3 is sending some error code, but I don't think it is.) 

But given the realities of LinuxCNC, both as an open-source platform, and as a controller for physical machines that can maim, (and assuming that the documentation is wrong, and G38.3 is sending an error code, and that QtDragon is not handling the popup messages already, which I don't think is the case) I'd like to discuss your dual points: 

A) "Those using the qt_auto_probe_tool subroutine always need to check the returned value (notice the use of 'return [-N]' (with N being -1, -2, -3, 3 and 999) sprinkled around the code, and handle any error appropriately, and can not depend on popups and automatic termination of the program."

When I search LinuxCNC on GitHub for qt_auto_probe_tool.ngc, I only find two mentions in sims: woodpecker and qtdragon. To my mind, that means "those using" are woodpecker and qtdragon. 

Are you using woodpecker? I went to look for installation instructions, and google would only give me non-linuxcnc sites that are pretty old. On the 'stable' [LinuxCNC User Introduction](https://linuxcnc.org/docs/stable/html/user/user-intro.html#sec:graphical-user-interfaces) page, they no longer list woodpecker. 

So therefore, I kinda think there's effectively only one member of the "those using" class, and that is the community of QtDragon users. 

I'm using QtDragon, but I'm not a developer on it. It has developers who have [documented installing qt_auto_probe_tool.ngc](https://linuxcnc.org/docs/stable/html/gui/qtdragon.html) extensively. They even make an attempt to incorporate a workflow using it (which I hope to improve over on #2706 ) . 

Which I think brings me to your second point:
B) "If you want a popup from the subrouteine itself, all these error handling blocks need to be changed and the subroutine become unusable in a non-interactive setting."

I think you're saying that the error should only be handled by the QtDragon interface. (Again, I think QtDragon **is** handling it, but with G38.3 there isn't any error message to be handled. Nonetheless, assuming you're right, and using G38.2 means a popup is generated earlier in the stack, and G38.3 is sending the error, just not stopping the flow to do a popup, I'll continue). I guess I still come down feeling that since the folks over at QtDragon made use of this stub by publishing it and modifying their installation scripts to copy it into the configuration sub-directories, that by us changing it for a theoretical non-interactive setting, we have made it less-functional with the active QtDragon interface, and in so doing we're doing the community a greater disservice.

I suppose QtDragon developers could be informed that they're not handling errors from M6 remapping in the most correct manner (assuming G38.3 is generating an error), and that they should mount a change to their interface to look for returning error messages. But it seems they should have time to respond before having it thrust upon them.  

But in a larger sense, two questions (still assuming G38.3 is generating an error). First, does it even make sense to have a non-interactive qt-based linuxcnc? Isn't Qt primarily an interactive enabler? The "qt_" in qt_auto_tool_probe.ngc stands for "Qt" as in [Qt Software](https://en.wikipedia.org/wiki/Qt_(software)), and therefore, why should it be subjected to a requirement to handle remap code that is preparing for a theoretical "non-interactive setting." And second, what is a use-case for a non-interactive probe failure setting? If a probe move fails, I have a hard time conceiving of a usage where it would be desirable to have a "non-interactive" process continue without human intervention. 

I hope this is helpful, and if I've written anything that isn't, I mean no disrespect. I am super-grateful to everyone who is helping me and others in this community learn and do more.

All the best.

PR #2563, opened 2023-07-03
Fix typos in docs/ and misc. leftovers

luzpaz:
Found via `codespell -q 3 -S *.po,*.pot,*.svg,*.ts,./.git/logs,./share,./docs/man/es,./configs/attic,*_fr.*,*_es.*,README_es,./src/hal/classicladder -L ans,ba,breal,bu,bulle,busses,componentes,currenty,doubleclick,dout,dum,extint,faktor,fo,fpr,halp,ihs,inout,inport,mot,parm,parms,propertys,ro,seh,ser,smoe,te,tey,trough,ue,ure,wille,wont,wonte`

Follow-up to #2548

----

I made this a catch-all PR and added leftover typos. Hopefully that is OK to do. Also please review closely, I wasn't sure if the man pages were generated and therefore not the target for a fix.

### DIFF
--- a/configs/by_interface/parport/plasma-thc-sim/simulator.hal
+++ b/configs/by_interface/parport/plasma-thc-sim/simulator.hal
@@ -63,6 +63,6 @@ setp wcomp.3.max -10
 setp wcomp.3.min -15
 setp wcomp.4.max 15
 setp wcomp.4.min 10
-# ArcOK comarator, uses ignition timeout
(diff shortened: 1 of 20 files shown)


hansu:
>Also please review closely, I wasn't sure if the man pages were generated and therefore not the target for a fix.

The man pages, which are included as man page in the repo are not generated. Only those which are included in a comp file or those in the adoc format. So this is fine =)

PR #2566, opened 2023-07-04
Make it easier to figure out where the current G43 values are stored.

petterreinholdtsen:


docs/src/gcode/g-code.adoc
```diff
@@ -1392,6 +1392,9 @@ G43 <H->
 
 'G43 Hn' uses the offset for tool n.
 
+The active tool length compensation values are stored in the numbered
+parameters '5401-5409'.
```
SebKuzminsky, line 1396: I think there should be no newlines in this hunk, to support translations.

SebKuzminsky (changes requested):

petterreinholdtsen:
[Sebastian Kuzminsky]
> +The active tool length compensation values are stored in the numbered
> +parameters '5401-5409'.
>
> I think there should be no newlines in this hunk, to support
> translations.

The msgid generated by po4a by running 'make -C src pofiles' for this
string look like this:

msgid ""
"The active tool length compensation values are stored in the numbered "
"parameters '5401-5409'."

This support translations just fine.

-- 
Happy hacking
Petter Reinholdtsen

docs/src/gcode/g-code.adoc
hansu: @smoe what is your experience here?

smoe:
I have to paddle back from (a good part of) my past corrections - when the text is at the top level of the document, then po4a and weblate seem to indeed work fine and present this without a newline. Weblate is doing some maintenance at the moment, so I cannot verify. I suggest to accept as it is and if it fails then Petter is the one to communicate this to the po4a folks.

docs/src/gcode/g-code.adoc
zz912: > I think there should be no newlines in this hunk, to support translations.

My view of the matter is that. That translations should not hinder improvement.

PR #2567, opened 2023-07-04
Explain with example how to use G38.2 to measure tool height.

petterreinholdtsen:
Include warning about moving with G49 active.

### DIFF
--- a/docs/src/gcode/g-code.adoc
+++ b/docs/src/gcode/g-code.adoc
@@ -1249,6 +1249,24 @@ If the probing operation failed, G38.2 and G38.4 will signal an error
 by posting an message on screen if the selected GUI supports that.
 And by halting program execution.
 
+Here is an example formula to probe tool height with conversion from a
+local coordinate system Z offset to machine coordinates which is
+stored in the tool table.  The existing tool height compensation is
+first cancelled with G49 to avoid including it in the calculation of
+height, and the new height is loaded from the tool table.  The start
+position must be high enough above the tool height probe to compensate
+for the use of G49.
+
+.G38.2 Example
+[source,{ngc}]
+----
+G49
+G38.2 Z-100 F100
+#<zworkoffset> = [#[5203 + #5220 * 20] + #5213 * #5210]
+G10 L1 P#5400 Z#<zworkoffset> (set new tool offset)
+G43
+----
+
 A comment of the form '(PROBEOPEN filename.txt)' will open
 'filename.txt' and store the 9-number coordinate consisting of
 XYZABCUVW of each successful straight probe in it.


SebKuzminsky (changes requested):
I like the new information here but I think it should use semantic line breaks (even though the surrounding documentation does not (yet)).

SebKuzminsky:
This build failure which has been troubling us lately seems to be a bug in Debian Sid:
```
dpkg: error processing package python3-minimal (--configure):
 installed python3-minimal package post-installation script subprocess returned error exit status 1
Errors were encountered while processing:
 python3-minimal
```

petterreinholdtsen:
[Sebastian Kuzminsky]
> I like the new information here but I think it should use semantic
> line breaks (even though the surrounding documentation does not
> (yet)).

This is the generated msgid (using 'make -C src pofiles) for the new
strings:

#. type: Plain text
#: src/gcode/g-code.adoc:1259
msgid ""
"Here is an example formula to probe tool height with conversion from a local "
"coordinate system Z offset to machine coordinates which is stored in the "
"tool table.  The existing tool height compensation is first cancelled with "
"G49 to avoid including it in the calculation of height, and the new height "
"is loaded from the tool table.  The start position must be high enough above "
"the tool height probe to compensate for the use of G49."
msgstr ""

#. type: Block title
#: src/gcode/g-code.adoc:1260
#, no-wrap
msgid "G38.2 Example"
msgstr ""

#. type: delimited block -
#: src/gcode/g-code.adoc:1268
#, no-wrap
msgid ""
"G49\n"
"G38.2 Z-100 F100\n"
"#<zworkoffset> = [#[5203 + #5220 * 20] + #5213 * #5210]\n"
"G10 L1 P#5400 Z#<zworkoffset> (set new tool offset)\n"
"G43\n"
msgstr ""

Anything wrong with these that cause you to ask for removal of newlines?

-- 
Happy hacking
Petter Reinholdtsen

petterreinholdtsen:
[Sebastian Kuzminsky]
> This build failure which has been troubling us lately seems to be a
> bug in Debian Sid:

Probably <URL: https://bugs.debian.org/1040316 >.

-- 
Happy hacking
Petter Reinholdtsen

SebKuzminsky:
> Anything wrong with these that cause you to ask for removal of newlines?

I am channeling @smoe here, who I believe prefers no newlines within sentences (to improve translations).

For example: <https://github.com/LinuxCNC/linuxcnc/pull/2256#pullrequestreview-1244233541>

petterreinholdtsen:
[Sebastian Kuzminsky]
> I am channeling @smoe here, who I believe prefers no newlines within
> sentences (to improve translations).

And I am sure he have experienced po4a splitting strings unexpectedly in
some cases, causing him to believe the newline need to be removed
everywhere.  But po4a handle newline in paragraphs at least in this case
as it should, as can be seen when looking at docs/po/documentation.pot.

-- 
Happy hacking
Petter Reinholdtsen

smoe:
Newlines do not trigger a technical fault in po4a.  The automated translation within weblate will attempt to mimic the layout of the sentence and/or translate two halfs that are then nonsensical. Also, I prefer a line-orientation for diffs in github,  but that is admittedly not my prime motivation.

petterreinholdtsen:
[Steffen Möller]
> Newlines do not trigger a technical fault in po4a.  The automated
> translation within weblate will attempt to mimic the layout of the
> sentence and/or translate two halfs that are then nonsensical.

Not quite sure what you mean by this, as weblate operate on single msgid
strings, and the automated translation will be handed a complete msgid.

-- 
Vennlig hilsen
Petter Reinholdtsen

smoe:
You can typically kind of guess what the original meaning is but not always and it all takes time and energy from our translators.

petterreinholdtsen:
[Steffen Möller]
> You can typically kind of guess what the original meaning is but not
> always and it all takes time and energy from our translators.

Please provide an example with pointers to a msgid and the proposed
automated translation.

As you can see from my examples, a string like this:

  Here is an example formula to probe tool height with conversion from a
  local coordinate system Z offset to machine coordinates which is
  stored in the tool table.  The existing tool height compensation is
  first cancelled with G49 to avoid including it in the calculation of
  height, and the new height is loaded from the tool table.  The start
  position must be high enough above the tool height probe to compensate
  for the use of G49.

Will show up in the POT file as a msgid like this, note how there are no
newlines in this msgid:

  msgid ""
  "Here is an example formula to probe tool height with conversion from a local "
  "coordinate system Z offset to machine coordinates which is stored in the "
  "tool table.  The existing tool height compensation is first cancelled with "
  "G49 to avoid including it in the calculation of height, and the new height "
  "is loaded from the tool table.  The start position must be high enough above "
  "the tool height probe to compensate for the use of G49."

Which newlines are passed to the automatic translation in this case?
-- 
Happy hacking
Petter Reinholdtsen

smoe:
Next time I encounter the line-break-triggered translation problem I will send you a link. German was at 99% (now back to 97%), though, and for the remaining 50 untranslated entries those line breaks are not the problem - it is just me not knowing how to translate them.

I did not find your example in weblate but expect all the new lines to be forwarded to the autotranslators. If weblate learns to distinguish how newlines are treated in regular texts versus verbatim sections, then I tend to agree that we could get somewhere. I still like to have the git diffs on semantic units and your segregation of "tool table" into a third line triggers some "why?" in me.

petterreinholdtsen:
[Steffen Möller]
> Next time I encounter the line-break-triggered translation problem I
> will send you a link.

I recommend starting with 'make -C src pofiles' and looking at the
generated .pot file, as it is the source of everything Weblate work on.


> I did not find your example in weblate but expect all the new lines to
> be forwarded to the autotranslators.

It is not yet on Weblate, I created the POT file myself using the
command mentioned above based on this pull requests branch, to
demonstrate how po4a is parsing the text correctly.

> If weblate learns to distinguish how newlines are treated in regular
> texts versus verbatim sections, then I tend to agree that we could get
> somewhere. I still like to have the git diffs on semantic units and
> your segregation of "tool table" into a third line triggers some
> "why?" in me.

Weblate is not involved here.  It is po4a and the POT files that decide
what Weblate is working on.  Look at the POT content next time you are
in doubt the adoc file is properly parsed and handled by the translation
framework.

-- 
Happy hacking
Petter Reinholdtsen

smoe:
In weblate, the multi-line paragraphs are should presented as a single line with line breaks represented by double-blanks. That is correct. And those double-blanks should not matter when auto-translators do encounter them. This does not always happen, though, and then the translation fails.

You have a point that is somewhat painful for me to admit. Instead of forcing everything on single lines that could cause trouble, I should have investigated why the presentation with line breaks occurred - much like you suggested. Likely, it was some inconsistent indentation that I should have traced up, instead. Not completely unthinkable that I have fixed both the line breaks and the indentation at the same time.

So, please proceed as you want and I'll ring a bell when trouble arises. Just maybe still try to keep on the same line what belongs together, so I would prefer:

  msgid ""
  "Here is an example formula to probe tool height with conversion from a local "
  "coordinate system Z offset to machine coordinates which is stored in the tool table."
  "The existing tool height compensation is first cancelled with G49 to avoid including it in the calculation of height,"
  "and the new height is loaded from the tool table."
  "The start position must be high enough above the tool height probe to compensate for the use of G49."

smoe:
https://hosted.weblate.org/translate/linuxcnc/linuxcnc-docs/de/?q=state%3Aneeds-editing&offset=17. is one that I (?, likely) had simplified manually:

From
```
A file whose content is compared to the output from running the test
 scripts.  
If the test output is identical to the content of the
 expected file, the test succeeds.
```
to
```
A file whose content is compared to the output from running the test scripts.
If the test output is identical to the content of the expected file, the test succeeds.
```
which is still presented as two lines when it should be only one, right?

The text in https://github.com/LinuxCNC/linuxcnc/blob/master/docs/src/code/writing-tests.adoc?plain=1#L64 is
```
expected::
   A file whose content is compared to the output from running the test scripts.
   If the test output is identical to the content of the expected file, the test succeeds.
```

smoe:
https://hosted.weblate.org/translate/linuxcnc/linuxcnc-docs/de/?q=state%3Aneeds-editing&offset=19 is another one that was just changed (I presume again this was someone manually) from

```
If this file exist, a test failure is expected and does not cause
 runtests to return an exit code signaling an error.
```

to a single line (not shown). There was a blank in the start of the second line, so this is not too surprising.

petterreinholdtsen:
[Steffen Möller]
> So, please proceed as you want and I'll ring a bell when trouble
> arises.

Very good.  I guess this pull request might be ready to go in, then.

> Just maybe still try to keep on the same line what belongs together,
> so I would prefer:

Note, there are no newline in the msgid you presented here:

>   msgid ""
>   "Here is an example formula to probe tool height with conversion from a local "
>   "coordinate system Z offset to machine coordinates which is stored in the tool table."
>   "The existing tool height compensation is first cancelled with G49 to avoid including it in the calculation of height,"
>   "and the new height is loaded from the tool table."
>   "The start position must be high enough above the tool height probe to compensate for the use of G49."

Newlines are written as \n in msgid.  Similar to how C strings are
concatinated, the end result for the string above is as an example no
space between "table." and "The", resulting in this string fragment:
"tool table.The Existing", and not what I suspect you wanted, ie "tool
table.\nThe Existing".  Asciidoc paragraphs only get a single space
between each line from po4a, no newline involved.

-- 
Happy hacking
Petter Reinholdtsen

smoe:
Hm. I see. Sounds about right, except that it looked all weird in weblate. So, my "have it in one line"-approach solved the issue and with 2.9 ante portas this was fine enough for me. I can see though that we should find out why the weblate entries have more than single lines and maybe fix that.

petterreinholdtsen:
[Steffen Möller]
> https://hosted.weblate.org/translate/linuxcnc/linuxcnc-docs/de/?q=state%3Aneeds-editing&offset=17. is
> one that I (?, likely) had simplified manually:

This link lead me to this string: "*component_is_ready* ".  I guess
search and offset might not end up at the same string every time.  If
you use the chain link in the upper right corner of the edit field or
the string in the "Nearby string" tag, the link should be more stable.
I believe I found the string you talk about as
<URL: https://hosted.weblate.org/translate/linuxcnc/linuxcnc-docs/de/?checksum=8bf5fcddb2888465 >.

> which is still presented as two lines when it should be only one,
> right?

Yes.  Look like description lists get the unwanted newlines you talk
about.

I made a simple demonstration on how po4a handle various situations.  See
<URL: https://github.com/petterreinholdtsen/linuxcnc/blob/po4a-demo/po4a-failed-stringsplit/project.pot >
if you want to experiment.  Do not know Asciidoc enough to say if po4a
is doing a mistake.

-- 
Happy hacking
Petter Reinholdtsen

smoe:
I have reviewed quite another number of changes in weblate now and came to accept that the "all in one line" presentation indeed seems to work for regular "top-level" text. And it also works in "*" itemizations.

Problematic are
  * descriptions
  * other indented text, like in man pages
  
No idea if this has recently improved or if I am missing something. So, I was too strict with all the semantic line breaks. Anyway, I still think it is good not to blindly run everything through the "fmt" but have sentences spread on as few lines as possible. This should with help the interpretation of patches.

petterreinholdtsen:
@SebKuzminsky Do you still request changes, or can this patch be used as it is?

andypugh:
In the absence of objections, I am going to merge and close this

petterreinholdtsen:
@andypugh was it merged?  @SebKuzminsky If you do not still request changes, perhaps that flag can be removed?

Should this still aim for the 2.9 branch, or are we to late for this?

andypugh:
It missed the 2.9 release, but the online docs at www.linuxcnc.org are "live" so it will show up there.

Issue #2524, opened 2023-06-15
Gmoccapy - command.mdi() must be call in mode IDLE only


zz912:
This bug may be the cause of this bug:
https://github.com/LinuxCNC/linuxcnc/issues/2453
But I don't want to deal with this connection now.
I would like to fix this bug only to increase the reliability of Gmoccapy.

Here is the definition of the proper use of self.command.mdi() :
http://linuxcnc.org/docs/2.9/html/config/python-interface.html#_preparing_to_send_commands

Here is a call to self.command.mdi("G43) without checking that the mode is IDLE:
https://github.com/LinuxCNC/linuxcnc/blob/56883b969838b721c0b0ebe12b9c14abf91d640e/src/emc/usr_intf/gmoccapy/gmoccapy.py#L3540-L3544
The condition self.stat.task_mode != linuxcnc.MODE_AUTO does not solve the state when the MDI command was started using HALUI.

I could fix it like this:
```
self.stat.poll()
 if "G43" in self.active_gcodes and self.stat.task_mode != linuxcnc.MODE_AUTO and self.stat.interp_state == linuxcnc.INTERP_IDLE: 
     self.command.mode(linuxcnc.MODE_MDI) 
     self.command.wait_complete() 
     self.command.mdi("G43") 
     self.command.wait_complete() 
```
The problem is that I don't know why command G43 is called?
If the condition self.stat.interp_state == linuxcnc.INTERP_IDLE is not met, command G43 will not be executed.
Is it necessary to call command G43 then?

gmoccapy:
Tool handling is redesigned by ReneDev, that should solve this issue. may also be related to #2613
no work on tool handling, as it will be much better in short time

PR #2571, opened 2023-07-06
Gmoccapy - new HAL pin disable_automatic_G43

zz912:
Hello,

It is said that a bad programmer, when he cannot solve a problem, creates a parameter that the user must set. The bad programmer also declares it as an advantage that the user can set many things himself.

I didn't want to use this solution because it is not automatic.

One of the causes of this bug:
That will run this command:
https://github.com/LinuxCNC/linuxcnc/blob/56883b969838b721c0b0ebe12b9c14abf91d640e/src/emc/usr_intf/gmoccapy/gmoccapy.py#L3540-L3544
after tool change.

The bug https://github.com/LinuxCNC/linuxcnc/issues/2453 could be solved by having command.wait_complete() wait for mdi-commands by HALUI to complete. It would be fantastic because it would be automatic. The user would not have to worry about anything. Unfortunately, I was unable to create such a fix.

However, yesterday I realized that there may be more situations where it is necessary to prevent race conditions. For example, if I write my own python module. Therefore, I am deliberately not calling this Pull Request FIX 2453. 

This Pull Request can be used for FIX 2453, but I would like bug 2453 to be solved more elegantly in the future and the automatic command G43 to be turned off automatically.

If this Pull Request is approved, it would be necessary to modify the manual:
1) http://linuxcnc.org/docs/2.9/html/gui/gmoccapy.html#_tool_related_pins
disable_automatic_G43(bit IN) - disable automatic run command "G43" after toolchange
Automatic run command "G43" can cause run condition. So writing a program makes you responsible to disable automatic run command "G43" so that a race condition does not occur.

2) http://linuxcnc.org/docs/2.9/html/gui/gmoccapy.html#_known_problems
12.3 Warning "Must be in MDI mode....." after tool change
Gmoccapy will run the after tool change MDI command G43. This function is very useful, unfortunately it can cause a problem in certain cases when a race condition occurs and this message is raised:
Must be in MDI mode
Be careful, this bug is random. Just because it didn't show up right away doesn't mean it won't show up later. This can be caused by using:
1. HALUI - MDI commnads:
For example in INI file is:
```
[HALUI]
MDI_COMMAND = M61 Q1
MDI_COMMAND = M61 Q2
```
(If you are in MDI mode the bug will not appear.)
2. Your own program
For example:
```
self.command.mode(linuxcnc.MODE_MDI)
self.command.wait_complete()
self.command.mdi("M61 Q1")
self.command.wait_complete()
self.command.mode(MODE_MANUAL)
self.command.wait_complete()
```
Sometimes the automatic run command "G43" will win and then self.command.mode(MODE_MANUAL) will run. In this case, nothing happens.
Sometimes self.command.mode(MODE_MANUAL) wins and then an error occurs when trying to run G43.

I would like ask for requested a review from [gmoccapy](https://github.com/gmoccapy)

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -3537,11 +3537,14 @@ class gmoccapy(object):
             self.on_hal_status_interp_idle(None)
             return
 
+        # automatic run command "G43"
         if "G43" in self.active_gcodes and self.stat.task_mode != linuxcnc.MODE_AUTO:
-            self.command.mode(linuxcnc.MODE_MDI)
-            self.command.wait_complete()
-            self.command.mdi("G43")
-            self.command.wait_complete()
+            if not hal.get_value("gmoccapy.disable_automatic_G43"):
+                self.command.mode(linuxcnc.MODE_MDI)
+                self.command.wait_complete()
+                self.command.mdi("G43")
+                self.command.wait_complete()
+                LOG.debug("automatic run command G43 was executed ")
 
     def _set_enable_tooltips(self, value):
         LOG.debug("_set_enable_tooltips = {0}".format(value))
@@ -5685,6 +5688,10 @@ class gmoccapy(object):
         pin = self.halcomp.newpin('toolchange-change', hal.HAL_BIT, hal.HAL_IN)
         hal_glib.GPin(pin).connect('value_changed', self.on_tool_change)
         self.halcomp.newpin('toolchange-confirm', hal.HAL_BIT, hal.HAL_IN)
+
+        # make a pin to disable automatic run command "G43" after toolchange
+        # automatic run command "G43" can cause run condition
+        self.halcomp.newpin("disable_automatic_G43", hal.HAL_BIT, hal.HAL_IN)
 
         # make a pin to confirm a warning dialog
         self.halcomp.newpin('warning-confirm', hal.HAL_BIT, hal.HAL_IN)


andypugh:
Is gmoccapy doing something that can also be done in the INI file? 
See RETAIN_G43 here: https://linuxcnc.org/docs/stable/html/config/ini-config.html#gcode:ini-features

zz912:
> Is gmoccapy doing something that can also be done in the INI file? See RETAIN_G43 here: https://linuxcnc.org/docs/stable/html/config/ini-config.html#gcode:ini-features

I dont know if Gmoccapy duplicate funkcionality of RETAIN_G43.

gmoccapy:
I did not know the retain G43 INI entry, seems to be quiet new

Gmoccapy reloads the tool. On start up with g43
After each tool change (only from the GUI)  a G43 is also applied.

Disabling the reload tool on start is not a good idea, getting rid of the G43 might be worse to try.
I do not think a Hal pin to disable  that is a good idea weather, may be a check box on the  settings page

Need to think about that, hope to come together with others in September

Norbert

gmoccapy:
By the way

Gmoccapy only applies an G43 if it has been active before the change

Norbert

andypugh:
> I did not know the retain G43 INI entry, seems to be quiet new

It was controlled by the FEATURES bitmap in 2.7 and became an INI entry in 2.8

gmoccapy:
@zz912 
If I did get all your information right, the strange behavior does only appear if the Halui command do contain tool relate information, is that correct?

In the manual I think I did advice, that using halui and is not reccommended(otherwise I should add that)

Please allow me an other question: Why you need to use M61 Q2 command through Halui Commands? Just to simulate or a real application? I have integrated about 10 ATC through remap and Hal and did not need to use such commands.

If G43 part is not applied by gmoccapy, the message does not appear, right?

I do ask myself, if we disable that part in gmoccapy, wouldn't it be dangerous for any user after an update? I just do not look at G43 as I do know my very beautiful and reliable GUI does care for that.

Would it be a solution also to disable the part by an entry in the settings page?
Otherwise that would irritate new users.

I would like to understand, what commands and why you need to use on your real machine, as IMHO a VM do react different as a normal machine.

The friendly Norbert

zz912:
>  I do not think a Hal pin to disable that is a good idea weather, may be a check box on the settings page

The checkbox doesn't help me at all:
1) "automatic command G43" causes a bug. If this checkbox was turned on by the operator, it would cause LCNC to behave randomly, which is very dangerous. If "automatic command G43" is to be turned on and off manually, it should be hidden so that it can be changed by the integrator LCNC NOT operator .

2) Manually turning "automatic command G43" on and off does not solve my problem. I want to keep the default Gmoccapy functionality. I don't want each machine to behave differently with Gmoccapy. I need to make own cycle:
- turn off "automatic command G43"
- run my code (M61 QXX or M6 TXX)
- turn on "automatic command G43"

In the past I have asked what the "automatic command G43" is actually needed for.
https://github.com/LinuxCNC/linuxcnc/issues/2524
https://github.com/LinuxCNC/linuxcnc/issues/2453#issuecomment-1566212729
But no one responded.

If we knew the exact functionality of "automatic command G43", maybe "automatic command G43" could be removed from on_hal_status_tool_in_spindle_changed . That would solve everything.

If "automatic command G43" were only needed for these buttons:
toolchange_button
It would be enough to move "automatic command G43" from on_hal_status_tool_in_spindle_changed to these buttons.

> I have integrated about 10 ATC through remap and Hal and did not need to use such commands.

I am lucky man. I have everything special. :-)

I have a carousel tool change machine. I need to have two physical buttons on it:
number tool +
number tool -
After pressing the "number tool" button, the carousel will rotate by one position and M61 will be activated. These buttons are only active in MANUAL MODE. So I have to switch from MANUAL MODE to MDI MODE and back to MANUAL MODE. 
"automatic command G43" it spoils me. 
A friend has these buttons on 1983 machines controlled by Siemens. That's why he also wants them on the new LCNC machine. I think more people will want this functionality.

> If I did get all your information right, the strange behavior does only appear if the Halui command do contain tool relate information, is that correct?

It is true. Unfortunately, the situation is complicated. While researching this bug I found another bug:
https://forum.linuxcnc.org/38-general-linuxcnc-questions/49445-python-command-wait-complete-does-not-wait-for-c-halui-commands
Currently I don't believe in HALUI MDI-commands. I'm thinking of writing my own MDI-commands using a Python module.

> In the manual I think I did advice, that using halui and is not reccommended(otherwise I should add that)

I didn't find it there. I don't think it should be banned to use HALUI. We only know of one bug so far. On the previous machine I use HALUI to switch modes and it works without problems.

> If G43 part is not applied by gmoccapy, the message does not appear, right?

OK

> I do ask myself, if we disable that part in gmoccapy, wouldn't it be dangerous for any user after an update? I just do not look at G43 as I do know my very beautiful and reliable GUI does care for that.

1) It will be dangerous, that is why I want only disable HAL pin.
2) Gmoccapy is beautiful GUI, that is why I spent so much time for make it better.

zz912:
I tested HAL pin disable_automatic_G43 with issue #2489.
Here is the result:
https://github.com/LinuxCNC/linuxcnc/issues/2489#issuecomment-1637126681

gmoccapy:
@andypugh 
I will be on the LinuxCNC Meeting in Stuttgart and use all the time to optimize the gmoccapy code, that men's also this Isue.

Norbert

gmoccapy:
@zz912 
Even if my question is not related to the bug itself, please allow me to ask:
I still do not understand why you use M61 Qn? I understand that you do have two button to turn the carousel clockwise or counterclockwise. That should work without problem. But why do that button also set MDI Command M61Qn? This sets the too,l in spindle!
Why not set Tn without M6 as turning the carousel should only prepare the tool, but not change it. 

This is only to understand the behavior of your ATC as understanding helps to solve ;-)

Norbert

zz912:
@gmoccapy
On my machine, the change tool is required in two ways.

The first method is manual. This method is also important for servicing and changing the tool in the carousel magazine.

1. Sequence (physical UP button)
- orient spindle M19
- drive the spindle up

2. Sequence (physical button LEFT/RIGHT)
- position the carousel
- run M61 Qn

3. Sequence (physical button DOWN)
- drive spindle down

The second method is manual
Run M6 Tn
Sequence 1 then 2 then 3.

gmoccapy:
stuttgart meeting agrees the behavior of gcode should not depend on a HAL pin.

zz912:
I agree that this solution is not ideal. But at the time I solved it, this solution was the best I could come up with. Currently, I prefer to fix the RETAIN_G43 option in the INI file and delete this G-code, which I turned off with HAL Pin, from gmoccapy.py completely. What was the conclusion of the Studgart meeting?

PR #2575, opened 2023-07-10
Get rid of stdout lines from GUI drawing interpreter.

### LINKED ISSUE #2559: Loaded G-code program is executed on touch off in some kind of sandbox mode
I discovered a surprising feature of LinuxCNC I suspect is a bug.  At least it break the principle of least surprise for me.

In short, the loaded G-code program is executed in some kind of sand box mode when pressing the touch off button in the GUI.  I've tested this in Axis.  To reproduce, save this simple g-code program as test-touch-off.ngc:

```
(msg, Did you expect this?)
(print, The value of parameter 1000 is #1000)
(msg, The value of parameter 1000 is #1000)
G53 G0 Z0
G53 G0 X0 Y0
M2
```

Next, load it into your LinuxCNC instance (for example a simulator of axis_mm, but it work on a physical instance too).

Press the 'Home All' button, and then press 'Z' followed by 'Touch Off'.  Notice how the print and msg strings show up in stdout where linuxcnc was started and as notifications in the GUI.

The program was obviously executed at touch off, but the G0 instruction to move is ignored.  I also discovered that the numbered parameter have different content while running in the 'touch off sandbox', and M66 return bogus pin values.

What is going on here?

petterreinholdtsen:
Every time the GUI parse the G code program to draw the expected moves, any (print, message) line will output its message to stdout.  This is very confusing when debugging a program.  This patch rewrite the handling of (print, ...) to avoid this output.

Fixes #2559.

### DIFF
--- a/src/emc/canterp/canterp.cc
+++ b/src/emc/canterp/canterp.cc
@@ -102,6 +102,7 @@ public:
     void print_state_tag(StateTag const &tag);
     void set_loglevel(int level);
     void set_loop_on_main_m99(bool state);
+    FILE* get_stdout() { return NULL; };
     FILE *f;
     char filename[PATH_MAX];
 };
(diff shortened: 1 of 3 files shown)


SebKuzminsky:
This looks good and uncontroversial to me.

PR #2578, opened 2023-07-10
More tiny bits that I found while revisiting weblate (smoe:docs29_mac_103)

smoe:
Nothing special, I think. Most was on the removal of terminal blanks.
Concerning my "semantic line breaks" I need to back paddle a bit.
Weblate gets presented regular top-level paragraphs and the itemizations that start with "*"s just fine, i.e. all on a single line.

This "all in one line"-presentation fails when there are indentations, as they are found in man pages or the "::"-descriptions.

So, I was introducing far more semantic line breaks than I should have for which I want to apologize.

### DIFF
--- a/docs/man/man1/halshow.1
+++ b/docs/man/man1/halshow.1
@@ -67,7 +67,7 @@ LinuxCNC 2.9:
 None known at this time. 
 .PP
 .SH AUTHOR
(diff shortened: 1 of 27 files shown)


hansu:
Looks good to me :+1:

PR #2581, opened 2023-07-11
docs: missing colon and line breaks (smoe:docs_testwriting_linebreaks)

smoe:
Hello, the lines indented are causing trouble to the auto-translators somewhere between the .adoc and weblate. I have adjusted those to my liking and then (modestly) also introduced semantic line breaks elsewhere.
@petterreinholdtsen as the author and weblate thinkahead, I hope you can tolerate that, maybe even like it?

### DIFF
--- a/docs/src/code/writing-tests.adoc
+++ b/docs/src/code/writing-tests.adoc
@@ -3,25 +3,21 @@
 
 == The test framework
 
-The code base has unit and integration tests that can be executed
-automatically to ensure the program works as intended.  Such tests are
-often written to trigger a bug and to ensure the bug is detected if it
-resurfaces in the future, but also to validate behavior of components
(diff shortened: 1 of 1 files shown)


smoe:
Yes! Lovely! I'll update my patch accordingly - one second.

smoe:
Ok. Now, we have some final disagreement on how semantic the line breaks shall be, also when this does not affect weblate. If you do not like what I came up with then please just apply the patch you came up with and close this PR - this is basically a style guide discussion now - and I do not know if this needs to be in the style guide in the first place or should just be left to the discretion of the individual contributing authors.

smoe:
The build error is not related to this patch.

smoe:
Hm.

PR #2597, opened 2023-07-25
halcomp: insert shebang line rather than overwriting first line

phillc54:
This showed up on the [forum](https://forum.linuxcnc.org/38-general-linuxcnc-questions/49465-python-component-path?start=10#276196)

It possibly belongs in 2.8 as well, I currently don't have any easily available machines with python2.


andypugh:
This is rather clearly wrong. I think that it would belong in 2.8 too, even though it's unlikely that there will be another 2.8 release. 
I will merge it here and consider cherry-picking it into 2.8

PR #2610, opened 2023-08-10
Fix linking against libraries in Makefile.modinc

ighvh:
When linking a shared object/executable, libraries have to be added  after the object files and not before. Otherwise, symbols used in the object files are not resolved.

See https://forum.linuxcnc.org/9-installing-linuxcnc/41983-linuxcnc-ethercat-undefined-symbol-ecrt-slave-config-sdo?start=0

### DIFF
--- a/src/Makefile.modinc.in
+++ b/src/Makefile.modinc.in
@@ -122,7 +122,7 @@ $(foreach mod,$(patsubst %.o,%,$(obj-m)),\
 	$(Q)ld -d -r -o $*.tmp $^
 	$(Q)objcopy -j .rtapi_export -O binary $*.tmp $*.sym
 	$(Q)(echo '{ global : '; tr -s '\0' < $*.sym | xargs -r0 printf '%s;\n' | grep .; echo 'local : * ; };') > $*.ver
-	$(Q)$(CC) -shared -Bsymbolic $(LDFLAGS) -Wl,--version-script,$*.ver -o $@ $^ -lm
+	$(Q)$(CC) -shared -Bsymbolic $(LDFLAGS) -Wl,--version-script,$*.ver -o $@ $^ -lm $(EXTRA_LDFLAGS)
 	$(Q)chmod -x $@
 endif
 


andypugh:
I don't have any EtherCAT hardware, so I will just have to trust you on this....

ighvh:
To test this, no hardware is needed. The point is that the lcec module likes to link against other libraries, for instance libethercat.
It should lool like this:
```
# ldd /usr/lib/linuxcnc/modules/lcec.so 
        linux-vdso.so.1 (0x00007ffc8dfba000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ff15c747000)
        liblinuxcnchal.so.0 => /lib/liblinuxcnchal.so.0 (0x00007ff15c738000)
        libethercat.so.1 => /lib/x86_64-linux-gnu/libethercat.so.1 (0x00007ff15c72d000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff15c54c000)
        /lib64/ld-linux-x86-64.so.2 (0x00007ff15c862000)
```

But right now, without this patch and sittner/linuxcnc-ethercat#122, it looks like this:
```
# ldd src/lcec.so
        linux-vdso.so.1 (0x00007ffe8addf000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fb402a10000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fb40282f000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fb402b2b000)
```

When loading the module which is missing the link to libethercat, the symbol `ecrt_slave_config_sdo()` oviously cannot be found, because libethercat is not loaded.

The reason why it does not work with the current makefiles is the order of the linker arguments. Missing symbols are resolved left to right.
For example,
```
$ gcc -omyexe obj1.o obj2.o -lethercat
```
creates an executable `myexe` based on the compiled objects obj1 and obj2. Symbols which are not resolved yet (for example,  `ecrt_slave_config_sdo()`) are looked up in libethercat. This dependency is stored in the executable and can be examined by `ldd`, as seen above.

If the order of the arguments is swapped
```
$ gcc -omyexe -lethercat obj1.o obj2.o
```
missing symbols are **not** looked up in libethercat and stay unresolved. This is the current case.

PR #2611, opened 2023-08-15
docs: minor changes to suit sim configs

phillc54:
In doing these changes I see that the sim for craftsman seems to be python2/gtk2 and should be either converted to python3/gtk3 or removed. I don't have the time to attempt the conversion but I am willing to remove it if that is the way we should go.


hansu:
I think we should first see how big the work is to convert it to Python 3.

By the way @petterreinholdtsen - how is it with screenshots in different languages for the translated docs? Do the translator need to translate the image filename to match the correct file or can we do this somehow automatically?

petterreinholdtsen:
[Hans Unzner]
> By the way @petterreinholdtsen - how is it with screenshots in
> different languages for the translated docs? Do the translator need to
> translate the image filename to match the correct file or can we do
> this somehow automatically?

I believe the idea is to store translated images in parallell with the
english images, and adjust the path/filename in the translation PO
file.  I would recommend a language specific image directory if all
images are grouped together, and perhaps a different approach (language
specific file names) if they are not.  It depends on how many and what
kind of images.

-- 
Happy hacking
Petter Reinholdtsen

Sigma1912:
> I think we should first see how big the work is to convert it to Python 3.

Quite a bit of work , also it seems that the machine vendor that created this has not been in operation for a number of years.

petterreinholdtsen:
[Sigma1912]
> Quite a bit of work , also it seems that the machine vendor that
> created this has not been in operation for a number of years.

Linux distributions are dropping Python 2, so if the script should stay
relevant and usable, it need to be ported to Python 3.

-- 
Happy hacking
Petter Reinholdtsen

Sigma1912:
Yes I know, so  given that the machine vendor who created this for their machines seems to have disappeared years ago maybe it could be removed.
Also looking at the contents of '/configs/sim', this seems to be the only vendor-specific folder there is.

petterreinholdtsen:
[Sigma1912]
> Yes I know, so  given that the machine vendor who created this for
> their machines seems to have disappeared years ago maybe it could be
> removed.

I believe it is more relevant if there are users of this machine still
around.  I know several machines from now defunct vendors are still
being used all over the world.  I have no idea about this specifi
machine.

-- 
Happy hacking
Petter Reinholdtsen

Sigma1912:
On closer inspection this seems to be a custom gui that lives inside of a sim config.

petterreinholdtsen:
[Sigma1912]
> Quite a bit of work , also it seems that the machine vendor that
> created this has not been in operation for a number of years.

Btw, did you test with 2to3 to convert the script to Python 3?  Did not
make many changes when I ran int.  What cause you to say that it is
quite a bit of work to convert?

-- 
Happy hacking
Petter Reinholdtsen

Sigma1912:
Message ID: ***@***.***>
> What cause you to say that it is quite a bit of work to convert?

Well that is of course a subjective statement. When I see something with 
over 5k lines of python2/gtk2 code then I call that 'quite a bit of 
work' to make that run under python3/gtk3.

I find the python itself is usually not really the issue  but the gtk 
usually takes (me) a long time.


In my opinion we had better spend our time fixing things in the 
mainstream GUIs than dragging along broken things like this in the hope 
that maybe somebody will come along and fix it.


--------------CmuAkvp0q4nBmMuv7IRipz0B
Content-Type: text/html; charset=UTF-8
Content-Transfer-Encoding: 8bit

<html><head>
<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
  </head>
  <body>
    <span style="color: transparent; font-size: 0; display: none;
      visibility: hidden; overflow: hidden; opacity: 0; width: 0;
      height: 0; max-width: 0; max-height: 0; mso-hide: all">Message ID:
      <span>&lt;LinuxCNC/linuxcnc/pull/2611/c1729203950</span><span>@</span><span>github</span><span>.</span><span>com&gt;</span></span>
    <blockquote type="cite">What cause you to say that it is quite a bit
      of work to convert?</blockquote>
    <p><span style="color: transparent; font-size: 0; display: none;
        visibility: hidden; overflow: hidden; opacity: 0; width: 0;
        height: 0; max-width: 0; max-height: 0; mso-hide: all"><span></span></span>
      <script type="application/ld+json">[
{
***@***.***": "http://schema.org",
***@***.***": "EmailMessage",
"potentialAction": {
***@***.***": "ViewAction",
"target": "https://github.com/LinuxCNC/linuxcnc/pull/2611#issuecomment-1729203950",
"url": "https://github.com/LinuxCNC/linuxcnc/pull/2611#issuecomment-1729203950",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
***@***.***": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]</script></p>
    <p>Well that is of course a subjective statement. When I see
      something with over 5k lines of python2/gtk2 code then I call that
      'quite a bit of work' to make that run under python3/gtk3.</p>
    <p>I find the python itself is usually not really the issue&nbsp; but the
      gtk usually takes (me) a long time.</p>
    <p><br>
    </p>
    <p>In my opinion we had better spend our time fixing things in the
      mainstream GUIs than dragging along broken things like this in the
      hope that maybe somebody will come along and fix it.<br>
    </p>
    <p><br>
    </p>
  </body>
</html>

--------------CmuAkvp0q4nBmMuv7IRipz0B--

andypugh:
> On closer inspection this seems to be a custom gui that lives inside a sim config.

Yes, I don't think that it is in any way linked to a specific vendor or hardware, it's an alternative GUI that someone wrote. 

https://forum.linuxcnc.org/41-guis/27232-my-custom-designed-gui

Maybe the answer would be to ask them to update it?  
He appears to be still active on the forums, so I have sent an email.

Issue #2616, opened 2023-08-25
PncConf make wrong Tandem Axis for first "HOME_SEQUENCE = -0"


zz912:
RIP 2.9 Branche
Debian Bookworm
Oracle VM VirtualBox

Hello,

Maybe I'll make an interesting machine with "X Y1 Y2 Z1 Z2" kinematics.
That's why I'm preparing for this situation. I made the configuration "X Y1 Y2 Z1 Z2" by PncConf. I used it to simulate this configuration. Here are configuration sim files:
[Z0069_sim-003.zip](https://github.com/LinuxCNC/linuxcnc/files/12435799/Z0069_sim-003.zip)

The homing sequence must be:
1) Z
2) X
3) Y

This creates a situation where the INI file contains:
HOME_SEQUENCE = -0
Minus zero is not my invention. This was invented by PncConf.

The bug is that I can break the CNC machine in joint mode. I can move the Z axes asynchronously.
![Asynchronous_Z1_Z2](https://github.com/LinuxCNC/linuxcnc/assets/96618597/34918c84-9054-4a8d-878b-e1a90db4f396)

Automatic homing works fine:
![automatic_homing](https://github.com/LinuxCNC/linuxcnc/assets/96618597/edd5f20c-efe0-43d3-be9e-24afcde60c3c)

Manual homing is very, very dangerous, I can home:
1) Z1
2) Y1 and Y2
3) X
4) Z2
![manual_homing](https://github.com/LinuxCNC/linuxcnc/assets/96618597/a0023e47-fc17-48a6-a3f7-897ff627368a)

andypugh:
-0 is valid in IEEE floating point. But .home_sequence is an int, so I agree that the setup wizard should not be using -0. 
(Or, possibly, should refuse to use 0 as the sequence number for a tandem joint. I haven't run a test to see how this happens)

zz912:
I looked in the documentation:
http://linuxcnc.org/docs/2.9/html/config/ini-homing.html#sec:homing-section

> The initial HOME_SEQUENCE number may be 0, 1 (or -1).

This is NOT bug of motion, but this is a bug of PncConf.

Correct solution for my exmple is:
[JOINT_0]HOME_SEQUENCE = 2
[JOINT_1]HOME_SEQUENCE = -3
[JOINT_2]HOME_SEQUENCE = -3
[JOINT_3]HOME_SEQUENCE = -1
[JOINT_4]HOME_SEQUENCE = -1

[...]

PR #2621, opened 2023-08-28
PncConf - add Tip for home sequence

zz912:
I made another configuration for machine.
I started numbered home_sequence from 0 as usual.
Problem was, that first axis was tandem (Z1, Z2).
It caused a problems:
https://github.com/LinuxCNC/linuxcnc/issues/2616

After reviewing the documentation:
http://linuxcnc.org/docs/2.9/html/config/ini-homing.html#sec:homing-section
I found out that it is possible to start the number home_sequemce from the number 1 (-1). So that someone doesn't have the same problems as me, I edited PncConf.
I added: (If the first axis is a tandem, start from number 1)
image
Please take this as a suggested solution only. It would be ideal to check if the axis is not tandem and warn the user if 0 is filled in.


petterreinholdtsen:
[zz912]
> So that someone doesn&#39;t have the same problems as me, I edited
> PncConf.  I added: (If the first axis is a tandem, start from number
> 1)

I do not know the code, but suspect this change should be squashed into
one commit and rebased on the 2.9 branch, as it seem like a useful thing
to have in the next stable release.  @andypugh, what do you think?

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
Yes, whatever solution is settled on, it belongs in 2.9.

c-morley:
Thank you for the pull request.
I think  #2660 covers this problem yes?

zz912:
You're right.

PR #2624, opened 2023-08-30
python remap -expand tilde to user path

c-morley:
allows using ~ to represent the user path in PATH_APPEND, PATH_PREPEND and TOPLEVEL. This would bring these paths into alignment with other INI path entries options.

I am most interested in an audit of the C++ code.
I have tested and it seems to work fine with and without using the ~ substitute.

### DIFF
--- a/src/emc/pythonplugin/python_plugin.cc
+++ b/src/emc/pythonplugin/python_plugin.cc
@@ -353,8 +353,14 @@ int PythonPlugin::configure(const char *iniFilename,
     }
 
     char real_path[PATH_MAX];
+    char expandinistring[PATH_MAX];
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
[c-morley]
> allows using ~ to represent the user path in PATH_APPEND, PATH_PREPEND
> and TOPLEVEL. This would bring these paths into alignment with other
> INI path entries options.

This seem like a good idea.  Was the patch not ready?

-- 
Happy hacking
Petter Reinholdtsen

c-morley:
I pushed it to 2.9 with the related work it was holding up.
I do need to update the docs though.

PR #2626, opened 2023-09-04
Adding pin in hy_vfd to ease connection with Qtdragon GUI

looping40:
In hy_vfd, error-count and retval are parameters and are not easily connectable to QtDragon GUI
I added pins to also publish those values.
Qtdragon also requires spindle speed in rps. It can be connected using a range adaptor but with the new spindle-speed-fb-rps it's straightforward


c-morley:
Thank you for the pull request.
I have dug into this a little. It's a little bit of a wild west but...
It seems s32 pins for error count is common. I'm thinking maybe I should change Qtdragon to s32.
I still think adding the spindle RPS pin is a good idea.
I'll think on it some more.

andypugh:
Anything that you can do with a parameter (setp, getp) can be done with a pin. So it is probably better to just switch the parameters to pins.

looping40:
I didn't want to break something using params.
But yes I can juste replace params with pins.

OK, just let me know for s32 pins on error count.

andypugh:
There are quite a few params that have been changed to pins in various components over the years. 
Just don't forget that a pin needs a pointer to a pointer, not just a pointer.

c-morley:
qtdragon can now connect to U32 or S32.
So yes switching to pins and the spindle RPS pin would still be a good pull request.

looping40:
The new -u32 values are not take into account in the handlers.
May be can add the 2 values as only one should be connected

    def spindle_fault_changed(self, data):
        fault = hex(self.h['spindle-fault'] **+ self.h['spindle-fault-u32'])**
        self.w.lbl_spindle_fault.setText(fault)

    def mb_errors_changed(self, data):
        errors = self.h['spindle-modbus-errors'] **+ self.h['spindle-modbus-errors-u32']**
        self.w.lbl_mb_errors.setText(str(errors))

looping40:
@c-morley "can you please push [qtdragon/hd: -fix spindle fault pins to actually work](https://github.com/LinuxCNC/linuxcnc/commit/a179f6e6433118fab610f0f4325a1756f1bfe626) on master

c-morley:
pushed qtdragon work to master.

looping40:
Thanks
Since 1 or 2 days, I have the following error : 'AttributeError: "VCPWindow' object has no attribute "dro_button_stack_a'
I don't think that it is related because I have the same error with a fresh sim.qtdragon_hd.qtdragon_hd_xyz/qtdragon_hd_xyza.ini
In qtdragon_hd.ui changing name="dro_axis_4" to name="dro_axis_a" makes the error disappear

c-morley:
Ok fixed that error.

looping40:
@c-morley To me everything seems to be OK now, ready for a review ;)

c-morley:
ok .. I'm away working for a bit, if no one else does in the mean time I will when I'm back

PR #2648, opened 2023-09-20
Axis: add homing prompt to On/Off button.

DdotLdot:
For enable set HOMING_PROMPT = TRUE in DISPLAY section ini file.

homing_prompt

### DIFF
--- a/docs/src/config/ini-config.adoc
+++ b/docs/src/config/ini-config.adoc
@@ -283,6 +283,7 @@ Descriptions of the interfaces are in the Interfaces section of the User Manual.
   An under powered CPU may see improvement with a longer setting.
   Usually the default is fine.
 * `PREVIEW_TIMEOUT = 5` - Timeout (in seconds) for loading graphical preview of G-code. Currently AXIS only.
+* `HOMING_PROMPT = TRUE` - Show prompt message with homing request, when the Power On button is pressed in AXIS GUI. Pressing the "Ok" button in prompt message is equivalent to pressing the "Home All" button(or the Ctrl-HOME key).
 
 [NOTE]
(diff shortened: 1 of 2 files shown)


andypugh:
There is an existing (though non-obvious) feature that covers this use-case. 
http://linuxcnc.org/docs/stable/html/config/ini-config.html#_joint__lt_num_gt_section
See the option VOLATILE_HOME
If this is set then the machine automatically un-homes when turnd off. 

Does that cover your needs?

DdotLdot:
> There is an existing (though non-obvious) feature that covers this use-case. http://linuxcnc.org/docs/stable/html/config/ini-config.html#_joint__lt_num_gt_section See the option VOLATILE_HOME If this is set then the machine automatically un-homes when turned off.
> 
> Does that cover your needs?

Hi.
This has no effect when the Power Off button is pressed, the prompt message only appears when the Power On button is pressed. This is only a reminder to run homing when turning on the machine. Pressing the "Ok" button in message is equivalent to pressing the "Home all" button. By default this message is disabled.

andypugh:
My point is that is VOLATILE_HOME is set, if you turn off the machine then is un-homes and won't run G-code or MDI until re-homed. Whether it will jog in an unhomed state depends on the machine config. 

However, as you say, it is only displayed of configured on. 

But at the moment there is nothing in the documentation to inform the users of the new INI entry. 

If you add the parameter to the documentation then I will merge the change. 
https://github.com/LinuxCNC/linuxcnc/blob/master/docs/src/config/ini-config.adoc

andypugh:
Thanks, merged

DdotLdot:
Thank you

PR #2649, opened 2023-09-20
Trimmed set of translations built by default.

petterreinholdtsen:
Only translations touched the last 12 months or where >= 2% of the strings
are translated according to Weblate get build rules in docs/src/Submakefile, and
only languages with >50% get Debian packages with the translations.  The new
guidelines are documented in docs/README.adoc.

The set of translations built and published on https://linuxcnc.org/docs/devel/html/
are now Arabic, German, Spanish, French, Norwegian Bokmål, Russian, Swedish, Turkish
and Simplified Chinese.  Hungarian and Vietnamese are no longer built.  Reduced the
list in index.html to the built ones.

Debian packages for French, Spanish and Simplified Chinese are no longer created.


SebKuzminsky:
This is a good improvement and we should merge it.

smoe:
We need to do something about translations that have no active maintainers, indeed. And I agree that such a policy would be an improvement.

petterreinholdtsen:
[Steffen Möller]
> We need to do something about translations that have no active
> maintainers, indeed. And I agree that such a policy would be an
> improvement.

I wished more people saw the value of providing documentation in their
native language, and contributed to the translations even if they do not
want to take ownership like maintenance might be understood.

I guess the most important part of the change is the new and written
description on how translations will be handled and published.  I hope
we can raise the 2% limit over time as more and more translations climb
over the limit.  But it depend on volunters donating their time. :)

-- 
Happy hacking
Petter Reinholdtsen

hansu:
> This is a good improvement and we should merge it.

I agree. So any objections on the details?

smoe:
I frankly came to the conclusion that I prefer anyone talented enough to speak two languages and have a technical understanind of what is going on to please contribute to the English documentation. The translation we should address to see automated.

petterreinholdtsen:
[Steffen Möller]
> I frankly came to the conclusion that I prefer anyone talented enough to speak two languages and have a technical understand of what is going on to please contribute to the English documentation. The translation we should address to see automated.

While that might work for German, there are other languages where automated translation is not doing a good job.  I find the automated Norwegian Bokmål translations seen so far to be of too low quality.  So I believe we need to help translators also moving forward.

Issue #2646, opened 2023-09-20
False Limit violation warnings in AxisGUI 


Sigma1912:
Proposal:
 **Remove the function 'run_warn()' and the two function calls 'if run_warn(): return' from the 'axis.py' source code.**
 

AxisGUI produces false warnings about limit violations when the user executes gcode with active tool length compensation. Basically it will calculate limit violations using the currently active tool length offset regardless of toolchanges in the gcode.    

This came up recently on the forum but has been around for years:
https://forum.linuxcnc.org/20-g-code/50095-tool-offsets-issue-cannot-start-program-video-question#280947
and also here:
https://forum.linuxcnc.org/38-general-linuxcnc-questions/50135-error-program-exceeds-machine-maximum#281280

The check for limit violations 'run_warn()' which is called on 'run' and 'step' events: 

```
    def task_run(*event):
        if run_warn(): return

        global program_start_line, program_start_line_last
        program_start_line_last = program_start_line;
        ensure_mode(linuxcnc.MODE_AUTO)
        c.auto(linuxcnc.AUTO_RUN, program_start_line)
        program_start_line = 0
        t.tag_remove("ignored", "0.0", "end")
        o.set_highlight_line(None)

    def task_step(*event):
        if s.task_mode != linuxcnc.MODE_AUTO or s.interp_state != linuxcnc.INTERP_IDLE:
            o.set_highlight_line(None)
            if run_warn(): return
        ensure_mode(linuxcnc.MODE_AUTO)
        c.auto(linuxcnc.AUTO_STEP)
```

The run_warn() function looks like this:

```
def run_warn():
    warnings = []
    if o.canon:
        machine_limit_min, machine_limit_max = soft_limits()
        for i in range(3): # Does not enforce angle limits
            if not(s.axis_mask & (1<<i)): continue
            if o.canon.min_extents_notool[i] + to_internal_linear_unit(o.last_tool_offset[i]) < machine_limit_min[i]:
                warnings.append(_("Program exceeds machine minimum on axis %s")
                    % "XYZABCUVW"[i])
            if o.canon.max_extents_notool[i] + to_internal_linear_unit(o.last_tool_offset[i]) > machine_limit_max[i]:
                warnings.append(_("Program exceeds machine maximum on axis %s")
                    % "XYZABCUVW"[i])
    if warnings:
        text = "\n".join(warnings)
        return int(root_window.tk.call("nf_dialog", ".error",
            _("Program exceeds machine limits"),
[...]

PR #2652, opened 2023-09-21
remove buggy limit check from axis.py

Sigma1912:
This is with regard to this issue:
https://github.com/LinuxCNC/linuxcnc/issues/2646

### DIFF
--- a/src/emc/usr_intf/axis/scripts/axis.py
+++ b/src/emc/usr_intf/axis/scripts/axis.py
@@ -1872,27 +1872,6 @@ def get_max_jog_speed_map(a):
         a = "XYZABCUVW".index(axis_letter)
     return get_max_jog_speed(a)
 
-def run_warn():
-    warnings = []
-    if o.canon:
-        machine_limit_min, machine_limit_max = soft_limits()
-        for i in range(3): # Does not enforce angle limits
-            if not(s.axis_mask & (1<<i)): continue
-            if o.canon.min_extents_notool[i] + to_internal_linear_unit(o.last_tool_offset[i]) < machine_limit_min[i]:
-                warnings.append(_("Program exceeds machine minimum on axis %s")
-                    % "XYZABCUVW"[i])
-            if o.canon.max_extents_notool[i] + to_internal_linear_unit(o.last_tool_offset[i]) > machine_limit_max[i]:
-                warnings.append(_("Program exceeds machine maximum on axis %s")
-                    % "XYZABCUVW"[i])
-    if warnings:
-        text = "\n".join(warnings)
-        return int(root_window.tk.call("nf_dialog", ".error",
-            _("Program exceeds machine limits"),
-            text,
-            "warning",
-            1, _("Run Anyway"), _("Cancel")))
-    return 0
-
 def reload_file(refilter=True):
     if running(): return
     s.poll()
@@ -2341,8 +2320,6 @@ class TclCommands(nf.TclCommands):
             root_window.tk.call("exec", *e)
 
     def task_run(*event):
-        if run_warn(): return
-
         global program_start_line, program_start_line_last
         program_start_line_last = program_start_line;
         ensure_mode(linuxcnc.MODE_AUTO)
@@ -2354,7 +2331,6 @@ class TclCommands(nf.TclCommands):
     def task_step(*event):
         if s.task_mode != linuxcnc.MODE_AUTO or s.interp_state != linuxcnc.INTERP_IDLE:
             o.set_highlight_line(None)
-            if run_warn(): return
         ensure_mode(linuxcnc.MODE_AUTO)
         c.auto(linuxcnc.AUTO_STEP)
 


andypugh:
I would _much_ rather make the check not-buggy. I am sure that it used to work OK, possibly before JA changes? 

Does the tool length even belong in the calculation, if the system is simply logging the max and min joint positions as the code executes?

Sigma1912:
> I am sure that it used to work OK, possibly before JA changes?

I don't know that was before my time. 

> Does the tool length even belong in the calculation, if the system is simply logging the max and min joint positions as the code executes?

That is a good question, I'll investigate.

Sigma1912:
So I had a look at the values o.canon. calculates:
```
    for i in range(3): # Does not enforce angle limits
        print('# canon.max_extents_notool, machine_limit_max: ',i, ',', o.canon.max_extents_notool[i],',',machine_limit_max[i])
        print('# canon.min_extents_notool, machine_limit_min: ',i, ',', o.canon.min_extents_notool[i],',',machine_limit_min[i])

```
Running the preloaded Splash-gcode gives these values for X(0), Y(1) and Z(2) that seem sensible:

# canon.max_extents_notool, machine_limit_max:  0 , 5.373785039370079 , 10.0
# canon.min_extents_notool, machine_limit_min:  0 , 0.06920511811023623 , -10.0
# canon.max_extents_notool, machine_limit_max:  1 , 0.8222350393700788 , 10.0
# canon.min_extents_notool, machine_limit_min:  1 , 0.005690157480314961 , -10.0
# canon.max_extents_notool, machine_limit_max:  2 , -4.747889763779528 , 0.0
# canon.min_extents_notool, machine_limit_min:  2 , -4.944740157480315 , -12.598425196850394


However running this gcode: 

`G17 G21 G54 G80 G90  

G0 Z-10
G0 X10
G0 Y10
G0 X0Y0Z0 

M2
`
gives me this: 
# canon.max_extents_notool, machine_limit_max:  0 , -9e+99 , 10.0
# canon.min_extents_notool, machine_limit_min:  0 , 9e+99 , -10.0
# canon.max_extents_notool, machine_limit_max:  1 , -9e+99 , 10.0
# canon.min_extents_notool, machine_limit_min:  1 , 9e+99 , -10.0
# canon.max_extents_notool, machine_limit_max:  2 , -9e+99 , 0.0
# canon.min_extents_notool, machine_limit_min:  2 , 9e+99 , -12.598425196850394

So I get the impression that maybe canon extents calculation is somehow fragile.

Sigma1912:
Ok so it seems that if the gcode includes only rapid moves then the calculation does not work because if I change gcode to this:

`G17 G21 G54 G80 G90  

G0 Z-10
G1 X10 F100
G0 Y10
G0 X0Y0Z0 

M2`

Then I get better results:

 canon.max_extents_notool, machine_limit_max:  0 , 0.3937007874015748 , 10.0
 canon.min_extents_notool, machine_limit_min:  0 , 0.0 , -10.0
 canon.max_extents_notool, machine_limit_max:  1 , 0.3937007874015748 , 10.0
 canon.min_extents_notool, machine_limit_min:  1 , 0.0 , -10.0
 canon.max_extents_notool, machine_limit_max:  2 , -4.8660000000000005 , 0.0
 canon.min_extents_notool, machine_limit_min:  2 , -5.2597007874015755 , -12.598425196850394

andypugh:
Are these examples with a tool loaded? The numbers don't appear to match the G-code. 
Or perhaps these are home offsets?

Sigma1912:
No tool offset but with a G54 of -123.596mm (-4.86598 in)

Sigma1912:
So one fundamental problem seems to be that the method axis uses to calculate limit violations disregards tool changes in the gcode.

Example:
G54 of -40mm (-0.1969 in) and  tool loaded (TLO=20 -> 0.7874in):

gcode:

G17 G21 G54 G80 G90  
T10 M6 G43 (TLO = 10mm)
G0 Z25
G1 Z-10 F300
G0 Z0 
M2
 
Axis will calculate for the z axis:
 canon.max_extents_notool, machine_limit_max, last tool offset:  2 ,  -0.19685039370078744 ,  0.0 ,  0.7874015748031497
 canon.min_extents_notool, machine_limit_min, last tool offset:  2 ,  -1.5748031496062993 ,  -12.598425196850394 ,  0.7874015748031497
Program exceeds machine maximum on axis Z

Which is wrong. This gcode runs just fine because the tool is changed before any axis movement happens.

Sigma1912:
In my opinion this method cannot produce reliable limit checking because the calculation has no information about which tool offset is active for a particular move. If I understand the 'calc_extents' function right then it has no information about tool offsets at all.

line 209 glcanon.py:

    def calc_extents(self):
        self.min_extents, self.max_extents, self.min_extents_notool, self.max_extents_notool = gcode.calc_extents(self.arcfeed, self.feed, self.traverse)


If this ever worked correctly then I don't see how it did it for more than the tool loaded on start.

gmoccapy:
according to @andypugh this can be closed. https://github.com/LinuxCNC/linuxcnc/commit/5c9a79d186ea5e8b64772c71694debaddb0005bb

PR #2654, opened 2023-09-22
creating tool_db_utils

spacestate1:
I've added a c program to the sim/axis/db_demo to pull in tools form the sqlite3 databse. it's mostly based on the python code in the db_demo. It's basic But a good start to get Linuxnc working with sqlite3 and might open the door to track things like tool life and other things.


rene-dev:
I have a branch which integrates sqlite as a tooltable directly. It finally and once and for all solves all tooltable related issues properly. I will clean it up and create a PR.

spacestate1:
That would be awesome. Does it integrate the same way using the emc/tooldata code ?

andypugh:
Do either of these use the tool database schema that I worked out with Tormach? 

http://wiki.linuxcnc.org/cgi-bin/wiki.pl?ToolDatabase

spacestate1:
My code does not use that, but it can. I used a simple test db to make sure the connections were working and loading tools. I spindle hrs in the db, and one of my main motivations is to track individual tool usage time for tool life etc. I assume that can be added to your db without much trouble?

spacestate1:
@andypugh I've re-written the code to incorporate the database you mentioned here:
http://wiki.linuxcnc.org/cgi-bin/wiki.pl?ToolDatabase

It works the same it did before, but now it loads tools from that database into LinuxCNC. Now other features can be added to start tracking things like tool life and spindle usage.  

Here's the relevant changes 
https://github.com/spacestate1/linuxcnc/tree/master/configs/sim/axis/db_demo/tool_db_utils

rene-dev:
my code depends on this PR: https://github.com/LinuxCNC/linuxcnc/pull/2497
It greatly simplifies tool handling, and fixes all root causes of the problems with the tooltable.
It can adapt to any schema.

spacestate1:
I'll take a look. The code in this pull just interacts with the current system to load/unload tools and doesn't change anything internally.

gmoccapy:
stuttgart meeting agrees this will be obsolete when @rene-dev is done with the sqlite tooltable.

PR #2658, opened 2023-09-25
generate version in comment correctly

heeplr:


andypugh:
I don't think that this change has any effect as the line is a comment. 
And my reading of the comment is that it is saying that ,.linuxcnc_var LINUXCNCVERSION will return the value of the $EMCVERSION environment variable. 
So I don't think that this change is desirable. 
@dngarrett What do you think?

heeplr:
@andypugh I wasn't too sure myself if that's intentional or not. But at least there is a "C" missing in the varname.

Please feel free to edit as needed.

andypugh:
I will change the spelling, but leave the rest alone.

PR #2668, opened 2023-09-30
Added some subroutine information to the hallui mdi command section

xsnoopy:


andypugh:
Thanks, I am surprised to learn that the docs didn't already mention that.

PR #2671, opened 2023-10-02
change c++ standard to 20

rene-dev:
after the 2.9 release I would like to change the c++ version to 20, and the python version to 3.9.


petterreinholdtsen:
[Rene Hopf]
> after the 2.9 release I would like to change the c++ version to 20,
> and the python version to 3.9.

Why would you want to do that, which platforms will we drop support for
if we do this change, and which systems will be affected by such change?

I suspect we want to decide C++ version and python version
independently, and suggest to split this patch into two separate pull
requests.

-- 
Happy hacking
Petter Reinholdtsen

rene-dev:
debian 10 and ubuntu 20.04 need to be dropped for this change, as the gcc is too old. they happen to have python 3.9 and 3.10, so it makes sense to update the requirement at the same time. Keep in mind that this only affects master. Debian 10 will be EOL june 2024, and I dont think 2.10 will be released before that.

andypugh:
Currently LinuxCNC 2.8 supports OS versions back to Debian Wheezy (OOS 30 June 2020)  and Ubuntu precise(OOS  28 April 2017 )
There was a good argument for dropping these for 2.9 (mainly Py3 and GTK)  but I am not convinced that we should be in any hurry to drop older distros unless we absilutely have to, given the nature of our user base. 
Many of our users only know how to install from the ISO, and so will often lose their configurations doing so. 

So, _why_ do you want to change the C++ to 20 when we are currently dithering about moving on from C89 on the C side?

petterreinholdtsen:
Note, I am very positive to dropping any effort to try to keep LinuxCNC
running on operating systems without security support from the vendor,
because we lack the required manpower needed to do it well.  I still
believe any raised dependency requirement should be funded in functional
improvements and well articulated, hence my questions to @rene-dev for
his reasons to want to move from C++ 17 to 20 and from Python 3 to
Python 3.9.  I suspect these changes have different reasons and should
be decided independently.  If we still support C89 I suspect that too
deserve an update, a suggestion for a different issue, perhaps?

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
If we still support C89 I suspect that too deserve an update, a suggestion for a different issue, perhaps?

https://github.com/LinuxCNC/linuxcnc/issues/2672

gmoccapy:
@smoe will investigate if we can do this without dropping support using backports of gcc and python.

smoe:
Neither gcc nor python3 is in backports, and likely that is since there is no backports for buster. I checked on https://tracker.debian.org/pkg/python3 and found version 3.7 in oldoldstable (buster, see https://wiki.debian.org/LTS) and default gcc is version 8 (https://metadata.ftp-master.debian.org/changelogs//main/g/gcc-defaults/gcc-defaults_1.181_changelog, https://tracker.debian.org/pkg/gcc). Here is a summary of what gcc 8 can do: https://gcc.gnu.org/gcc-8/changes.html .
My suggestion is to drop support for Buster with 2.10.

rene-dev:
I agree, GCC 8 is really old. thanks for checking!

smoe:
> I agree, GCC 8 is really old. thanks for checking!

They came up with the early experimental routines "2a" for what I understood became "20", and the build failure suggests to change to it:
```
c++: error: unrecognized command line option '-std=gnu++20'; did you mean '-std=gnu++2a'?
```
This page https://gcc.gnu.org/projects/cxx-status.html has a summary of what feature was made available for what release, maybe it is not too bad?

rmu75:
C++20 really has a bunch of additional (vs. C++17) useful features like const eval, improved const expr support, improved STL like std::span that will make modernised code safer and easier to maintain.

andypugh:
I think that we should always consider whether it makes things better for the users rather than get carried away making things better for the programmers.

rmu75:
The question is how to quantify "better for users". Does any user suffer if support for Buster is dropped for 2.10? Making life easier for programmers will lead to "better" user experience, at least in the long run.

andypugh:
I anticipate much wailing and gnashing of teeth when they find that we have dropped support for _Wheezy_ and _Precise_ with 2.9.

c-morley:
The usually issue is linuxcnc/distribution is so difficult to get tweaked right, people prefer to keep the distribution for as long as possible. machine controllers don't _need_ to be continuously connected to the net so the usual security issue worry is not much of a real problem.  Our policy to this point is to put a lot of energy to keep old distributions supported.
Obviously there is a tipping point  but I think the policy is a good one.

NTULINUX:
As far as RTAI is concerned, kernel space doesn't support C++ anyway. You can use whatever C++ standard you want, none of it touches kernel space so don't worry about any of that. I have already created a `gnu11` branch in RTAI that builds against LinuxCNC with this: https://github.com/LinuxCNC/linuxcnc/pull/2685

rene-dev:
The glade version shipped with buster only works with python2 widgets. I don't understand why we support an os where glade panels can't be edited. This is very confusing to users. Gladevcp is unusable on buster.

andypugh:
> Gladevcp is unusable on buster.

Can you elaborate on this? I am sure lots of people are using gladevcp on Buster. 

Do you mean that it does not work with 2.9 and Master?  (I don't know, I haven't tried)

We do supply a special version of glade:
http://www.linuxcnc.org/dists/buster/base/binary-amd64/
(glade-gtk2)

rene-dev:
That's the old gtk2 version you are talking about. Im talking about gtk3.

https://packages.debian.org/buster/libglade2-dev

The glade version in buster is gtk3, with python2. This doesn't work with any version of linuxcnc.

Glade has an embedded python interpreter to work with custom widgets written in python, like gladevcp. The interpreter version cannot be changed at runtime.

rene-dev:
https://www.mail-archive.com/emc-developers@lists.sourceforge.net/msg21705.html

andypugh:
Can you raise a separate issue that explain the Glade  /  GladeVCP situation?

PR #2676, opened 2023-10-05
Test --enable-toolnml

heeplr:
"./configure ... --enable-toolnml" wasn't covered in CI


rene-dev:
What is this used for?

gmoccapy:
stuttgart meeting agrees this will be obsolete when @rene-dev is done with the sqlite tooltable.
anything tooltable related should not be a compiletime option.

heeplr:
I see. But does that really imply the tooltable being unusable from the NML layer or will sqlite support be implemented as an alternative to the current text format at similar API level like [the current implementation](https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/tooldata/tooldata_db.cc)? 

I think merging (or refactoring for that matter) toolnml would be a good step if I didn't miss anything, but removing it from NML might break some use cases. 

I'm currently working on getting a distributed network setup running with a more recent linuxcnc codebase and see lots of problems (including but not limited to the HAL layer).

Now if the general direction is towards a non-networkable desktop app with telnet interface, it'd be a kind of a showstopper for me.
It's also bad news for users of existing distributed setups who chose linuxcnc/emc2 for not being forced to run a complete windows environment directly on a controller in production.

heeplr:
This is an example resource listing some benefits of having a network capable tool db: https://www.wintool.com/en/tool-management-software/cnc-tool-library. Maybe something like ODBC would be better suited than sqlite?

petterreinholdtsen:
[heeplr]
> This is an example resource listing some benefits of having a network
> capable tool db:
> https://www.wintool.com/en/tool-management-software/cnc-tool-library.

Make sense.  Sad it is a non-free tool db.

> Maybe something like ODBC would be better suited than sqlite?

You mean a setup using a more high level API to allow any SQL database
to be used, also a networked one like PostgreSQL?  That would indeed be
nice. :)

I suspect it could be implemented already by using [DISPLAY]->DB_PROGRAM
instead of TOOL_EDITOR in the ini file to call a program fetching data
from an SQL database.

-- 
Happy hacking
Petter Reinholdtsen

heeplr:
> Make sense. Sad it is a non-free tool db.

Yeah, that functionality imposes quite some value and keeping a free & open implementation would be nice.

> I suspect it could be implemented already by using [DISPLAY]->DB_PROGRAM instead of TOOL_EDITOR in the ini file to call a program fetching data from an SQL database.

I think that resembles just partly what toolnml currently provides. So it's still curtailing functionality for non-database specific tasks.

andypugh:
> I see. But does that really imply the tooltable being unusable from the NML layer or will sqlite support be implemented as an alternative to the current text format at similar API level like [the current implementation](https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/tooldata/tooldata_db.cc)?

Where do you anticipate the tool data being stored? On the controller or on the remote PC? 

> Now if the general direction is towards a non-networkable desktop app with telnet interface, it'd be a kind of a showstopper for me. I

That's not a deliberate directional choice, but I am not sure that NML as a remote interface has ever actually completely worked.

You could maybe take a look at Machinekit and their Machinetalk interface, which I understand to work fairly well. However I don't know what the status of the rest of Machinekit is at the moment, I think it has unresolved issues with Python3.

heeplr:
> Where do you anticipate the tool data being stored? On the controller or on the remote PC?

Correct me if I'm wrong but as far as I understand, it can be stored on a any (multiple?) linuxcnc instances including while one being the master as configured in the .nml config file. In theory. (Atm there seem to be multiple ways to access the tool data including reading the file directly without any API)

> That's not a deliberate directional choice, but I am not sure that NML as a remote interface has ever actually completely worked.

You mean beyond HAL and tooltable? Is there anything else? I planned to look into NML support for those long ago but it's a rocky road to even get linuxcncrsh working. At least some of it must have worked at some point in the past.

andypugh:
I am not suggesting this as a _good_ solution, but while 2.9 is in maintenance mode, and master already has much of the NML tool interface already removed, could you use MDI G10 commands and read the data back with MDI (DEBUG, #5403) etc?

heeplr:
> while 2.9 is in maintenance mode, and master already has much of the NML tool interface already removed

yes, is there a benefit of removing it in master and not from a separate branch? or rather depreciate it while providing an alternative to adapt to?

> could you use MDI G10 commands and read the data back with MDI (DEBUG, #5403) etc?

I doubt any of the current network capabilities (telnet, halrmt, halstreamer?, etc.) can do what NML can currently does. They provide very useful niche functionality but are just frontends to the NML API and lack a lot of the genericness and extensibility.

It's also not so much about my specific setup as more about users that currently run ancient versions of linuxcnc in production with existing implementations. I'm just trying to speak for them since I know at least one.

Is it really that much work to complete the missing NML stuff so everything "just works", no matter if remote or not (including all GUI)? I'm still trying to understand the benefit for removing it beyond cutting the time needed to replace it with something modern (s.th. protobuffy probably), in contrast to the loss of existing functionality (even if it's rarely needed).

andypugh:
It could have been done in a feature branch, but experience with feature branches is that they become steadily harder to merge as time goes by. joints-axes was a feature branch when I first started with LinuxCNC in 2013. It finally got merged (ironically, by me as release manager) in late 2020.  
The tool table was removed from NML because it placed a hard limit of 50 on the number of tools that could be supported in the tool table, and at that point nothing was using the NML tooltable anyway. I am not expert on this subject, but I have been told that this is still the case, and that even if the tooltable is in NML nothing looks at it. 

Can I suggest that, until things settle down, you can run your own patched version of LinuxCNC while we work out what the long term solution is?

heeplr:
I looked a bit deeper into this and found that --enable-toolnml was [introduced in this commit](2dbb2f640fb8) along with
```
2) Define an applications interface: src/emc/tooldata/tooldata.hh
3) Support a new interface to a user-provided tool database
   management program specified by [EMCIO]DB_PROGRAM
```
and

```
4) Use a new mmapped tooldata interface between
   EMCIO and TASK while maintaining number of allowed
   pockets specified by CANON_POCKETS_MAX=1001
5) Deprecate tooldata in nml messages between EMCIO and TASK
   (legacy nml message method remains available using
     configure --enable-toolnml until removal)
```

While 2 & 3 are awesome, 4 and 5 are huge steps to move away from NML. But --enable-toolnml makes that optional, so it's cool probably.

I looked into libnml and there is a ton of stuff where I can't figure out why it was once added. There's queueing, locking, various checks and stuff. But I don't understand use-cases of every [config file option](https://linuxcnc.org/docs/html/code/code-notes.html#_configuration_file_format) for example or why [the default config](https://github.com/LinuxCNC/linuxcnc/blob/master/configs/common/linuxcnc_big.nml#L9) has different options for ```emcCommand```, ```emcStatus``` and ```emcError```?
I couldn't find any documentation for that.

It could be legacy but it could also cover some rare & really useful edge case someone implemented. Maybe the whole NML implementation enabled people to even run realtime components on different hosts. From what I understand, NML can run a distributed network of multiple linuxcnc instances/machines. (That could be an ancient form of "cloud computing" :) Probably employing some exotic Non-Transparent Bridge or similar hardware.)
The complexity/abstraction of CMS & NML and the initial question suggest that it's not trivial to understand every possible codepath - and thus it's easy to break use cases that once worked.

Unless there's a huge benefit to not make the code optional (like with the TOOL_NML macro), it might be better to not remove it from master without decent test coverage (e.g. tests/linuxcncrsh-tcp) and a LOT of releases with loud depreciation messages. (I don't know how often the average LinuxCNC user updates their setups. ;)

Also removing NML alltogether looks not so much like a mere feature change to me but more like a core paradigm shift so deep, it would probably benefit from a separate branch.

rene-dev:
I think the developers mostly agree on the following:

- anything tooltable related should not be a compile time option
- there should not be a limit on the number of tools
- in general, we want to move away from nml and replace it with something better and more flexible.

I still dont see how it is related in any way with a remote UI.

heeplr:
> anything tooltable related should not be a compile time option

That sounds arbitrary. What's the argument behind this? Making database dependencies optional sound like a good thing.

> there should not be a limit on the number of tools

I'm not sure about the limit. I suppose that's different from what "CANON_POCKETS_MAX" was? If not, it's probably some hardcoded value that would need some work. I can't see how CMS or NML could limit the number of tools.
Can you point me to the relation to (tool)nml or the tooldata.hh API?

> in general, we want to move away from nml and replace it with something better and more flexible.

I'd be the last to argue against replacing NML. I mean it works, but let's be honest... It smells a lot. 

But I certainly wouldn't just remove it from master without depreciation (--enable-toolnml never printed a depreaction warning at runtime). Maybe choosing an appropriate, modern 3rd party library that "can do everything NML can and more" first would enable additional test coverage to make a smoother transition.

> I still dont see how it is related in any way with a remote UI.

The UIs are meant to run as "xemc" process accessing CMS buffers that are shared between components (TASK in this case):  emcCommand, emcError, emcStatus (and toolCmd + toolSts before it got removed from the default configs).

IPC between UIs and EMCTASK (and, in an ideal world, other components aswell) happens through those channels. There's some stuff happening in the background.

Now UIs seem to employ a lot of different methods to query linuxcnc, including accessing linuxcncsvr owned files directly. I _think_ [this seems to be the main API where everything **should** go through by design](https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/nml_intf/emc_nml.hh) (please correct me if i'm wrong). With tool stuff going over NML and being a properly implemnted NML process, one could have a special GUI plugging directly into a tool database that serves multiple linuxcnc instances. With a custom DB backend, this could be a separate tooldb controller connecting to practically anything at almost any scale.

But it's not only about the UI but rather all other components that connect through NML. It's the IPC backbone that does queueing, locking, communication timestamping and all that other real-timey stuff that makes using, debugging & developing for linuxcnc tick.

andypugh:
> > anything tooltable related should not be a compile time option
> 
> That sounds arbitrary. What's the argument behind this?

If there is any configurabble option it should be something that ordinary users can change without recompiling. 

> I'm not sure about the limit. I suppose that's different from what "CANON_POCKETS_MAX" was? If not, it's probably some hardcoded value that would need some work. I can't see how CMS or NML could limit the number of tools.

It's based on the size of the NML buffee that is allocated:
https://github.com/LinuxCNC/linuxcnc/blob/2.7/configs/common/linuxcnc.nml#L15

(This is deliberately an old version, when the limit was 50 tools. It was later inreased to 1000 tools with a matching increase in the NML buffer size.)

But it doesn't really make sense to send all the tools all the time, the motion system only needs one at a time with the current setup where it only controls one tool at a time.

heeplr:
> If there is any configurabble option it should be something that ordinary users can change without recompiling.

This argument seems to aim at improving user friendliness. But I can't see how removing choice for a package maintainer without shifting it to the user (since you're not going to make toolnml a runtime choice) will improve usability. 
Instead, maintainers/users will [encounter problems](https://forum.linuxcnc.org/38-general-linuxcnc-questions/50257-configure-flag-enable-toolnml-gives-error-during-compilation) or [new bugs](https://github.com/LinuxCNC/linuxcnc/issues/2692) they can't work around and they're not even getting a benefit. If anything, this "benefits" only developers.

> It's based on the size of the NML buffee that is allocated: https://github.com/LinuxCNC/linuxcnc/blob/2.7/configs/common/linuxcnc.nml#L15
> 
> (This is deliberately an old version, when the limit was 50 tools. It was later inreased to 1000 tools with a matching increase in the NML buffer size.)

Ah, that makes sense. Thank you. 
 
> But it doesn't really make sense to send all the tools all the time, the motion system only needs one at a time with the current setup where it only controls one tool at a time.

I agree but it seems to be a bigger problem with buffersize just being a symptom:

A better solution might have been:
* renaming [```EMC_TOOL_STAT```](https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/nml_intf/emcops.cc#L155) to something more correct like ```EMC_TOOLS_STAT``` or ```EMC_TOOLTABLE_STAT``` 
* search/replace the linuxcnc codebase to use the new name + provide compile-time warning to notify 3rd-party code
* re-add ```EMC_TOOL_STAT``` that doesn't have the for-loop and only returns the currently used tool. 
* use that in all internal components that cause trouble
* then depreciate ```EMC_TOOLTABLE_STAT``` responsibly to provide a path for components to adapt by querying tools individually if they really need to know every tools' status. 
(There might be better solutions that also go without breaking design patterns but since this decision was already made long ago, I won't dig deeper and give just the example that first popped into my mind.)

Basically I think fixing the protocol by extending the API at one place, so things "just work" for the other components (like GUIs), generally is a much better solution in the long run than just pumping up limits (which surely are ancient in libnml), disabling code or plumbing in a quick fix with a shotgun surgery. Decisions like that keep a stable API but encourage inappropriate intimacy  where components do all sorts of weird stuff by simply abandoning or circumventing that API - including accessing the tooldb file directly *instead of using the very API, provided by linuxcnc* ... "it's just a textfile, right... what could possibly go wrong?"

heeplr:
Adding to this, I found this read explaining the culprits of distributed realtime capable messaging quite well: 

https://www.researchgate.net/publication/243775807_The_Neutral_Message_Language_A_Model_and_Method_for_Message_Passing_in_Heterogeneous_Environments

Naive implementations of non-distributed realtime IPC surely is easier to handle but still offers a trove of nasty pitfalls.

andypugh:
You will find the names of Shackleford and Proctor all over the LinuxCNC source code. 
At the LinuxCNC meetup in Wichita in 2013 we called Will Shackleford on the phone to discuss replacing NML. At that point he was very surprised that we were still using it.

heeplr:
> in 2013 we called Will Shackleford on the phone to discuss replacing NML.

So, here we are in 2023 with nothing changed, best intentions and a small PR. 
Are there any notable lessons learned or specific efforts to build upon from the last 10 years?

btw, one year after your phone call, Will [released the final version of rcslib](https://www.nist.gov/document/rcslib-20140429targz) into the public domain.

heeplr:
@gmoccapy 

> stuttgart meeting agrees this will be obsolete when @rene-dev is done with the sqlite tooltable. anything tooltable related should not be a compiletime option.

This didn't age well. Afais that code path is still untested.

I've seen that #2692 has been fixed in the meantime. That should make tests work now (if nothing else broke in the meantime, didn't test, yet).

PR #2679, opened 2023-10-06
more linuxcncrsh testing

heeplr:
* compare netcat output to expected output
* validate "set" calls with consecutive "get" calls
* check for error strings in result

This should catch a few things that were missed previously. If this gets approved, I'd apply the result to the linuxcncrsh-tcp test aswell in another PR.

EDIT: the spindle test without "-1" currently crashes linuxcncrsh, also it's currently sending "NAK" where it should send ACK (or nothing)


petterreinholdtsen:
Should this perhaps be targeting the 2.9 branch, or did the behaviour change between 2.9 and the master branch?

heeplr:
> Should this perhaps be targeting the 2.9 branch, or did the behaviour change between 2.9 and the master branch?

It doesn't pass CI so I thought it'd be best to first fix the bugs and then merge it.
I could change the expected output to have "the wrong output" accepted so tests pass but that seems smelly imho.

If failing tests are no problem, I can try to PR against the 2.9 branch.

petterreinholdtsen:
The failing tests seem to be in the documentation building system, so I suspect it is unrelated to your change.

heeplr:
The doc building is also failing currently but the failing rip-and-test job is caused by [this](https://github.com/LinuxCNC/linuxcnc/actions/runs/6435254986/job/17476118910?pr=2679#step:5:9259) ("get spindle" linuxcncrsh command without parameter)

PR #2682, opened 2023-10-07
Adjusted expected output for g72-facing and g72-missing-iteration tests.

petterreinholdtsen:
START_CHANGE() disappeared and CHANGE_TOOL(2) became CHANGE_TOOL() between version 2.9 and master. Not sure why, but as these lines are outside the features these tests were inteded to check, I just update the expected file to get these tests to succeed again in master.


andypugh:
Sorry Petter, but it looks like we were working in parallel, and this was fixed by dbd9d46a3f346296dc2040a104b83922b15dbf17

petterreinholdtsen:
[Andy pugh]
> Sorry Petter, but it looks like we were working in parallel, and this
> was fixed by dbd9d46a3f346296dc2040a104b83922b15dbf17

Aha. I try to use IRC for coordination.

-- 
Happy hacking
Petter Reinholdtsen

PR #2685, opened 2023-10-07
RTAI: Fix build against RTAI+GNU11

NTULINUX:
This may break RTAI builds that still use gnu89.

For some reason, if using GNU11, you don't need to define these anymore.


NTULINUX:
If all goes well such as trig functions in kernel space, notably:

atan
asin
acos
fmax
fmin

Then this should be backported to `2.9` otherwise the `2.9` branch won't work with modern RTAI.

NTULINUX:
One more thing, genserkins.ko has a stack frame size that's too big (kernel-space only) so I simply just removed the error from the Makefile.

smoe:
```
Runtest: 261 tests run, 260 successful, 1 failed + 0 expected, 2 skipped
Failed: 
	/home/runner/work/linuxcnc/linuxcnc/tests/linuxcncrsh
```

andypugh:
linuxcncrsh is failing on uspace builds too, and without this patch (I think)

NTULINUX:
What's the status of merging this? Can we re-run the tests? They seem to be passing now more regularly.

andypugh:
I have been bust trying to get 2.9-uspace packages ready. Then I will move on to trying to build RTAI packages. Then look at this.

PR #2689, opened 2023-10-09
fix EXTRA_DEBUG handling in Makefile

heeplr:
* EXTRA_DEBUG flags should override previous debug flags
* maintain same order in CXXFLAGS as in CFLAGS


andypugh:
Can you explain what issue this fixes?

heeplr:
sure. EXTRA_DEBUG can be passed from the commandline to alter CFLAGS dynamically while debugging (s. [this line in the Makefile](https://github.com/LinuxCNC/linuxcnc/blob/master/src/Makefile#L34)). 

e.g. Sometimes it's desireable during debugging to disable optimization and change debugging specific flags:  
```make V=1 EXTRA_DEBUG="-O0 -ggdb"```

The order fixes are just minor housekeeping that maybe might save future headaches as argument order matters for CFLAGS/CPPFLAGS.

andypugh:
Thanks, it obvious now you explain it. :-)

PR #2691, opened 2023-10-09
linuxcncrsh: fixes & partial cleanup

heeplr:
This fixes the current linuxcncrsh segfault bug and some other issues. I also cleaned/modernized parts of the code, added "get update" command and re-enabled the test (disabled in eca9685), which should pass now.
The individual commits provide special details.


andypugh:
I am inclined to just merge this, as I don't think anyone has paid any attention to linuxcncrsh for a long time.

PR #2702, opened 2023-10-19
Improve treatment of o-codes in documentation

VectorHasting:
I recently learned the NGC coding language by reading the page currently titled G-code overview. I found it unnecessarily confusing because all the remapping code I was trying to decipher uses o-codes and the documentation didn’t address them until after it had supposedly laid out all the acceptable ‘words.’ Also, in places the language describing them did not make much sense “1-30 subroutine local parameters of call arguments,” for example. Additionally, sometimes o-code is used, and sometimes o-word, but neither was clearly defined nor the interchangeability of the terms addressed.

The reason for this seems benign: this document was started 20 years ago at NIST before the LinuxCNC community extended it to include o-codes.  When that happened, an o-code page was created, and some references were added into the lower parts of this overview document. But it seems attention to integrating an introduction to this functionality didn’t happen.

I also clarified rules on Comments, to move mention of “active” comments like (MSG) to the first mention of commenting.

I propose also changing the page title from “G-code Overview” to “NGC-code Overview” because it’s kind of a misnomer since the language is NGC, and also the page g-code.adoc is titled “G Codes” which is confusingly similar. I checked for another reference to the chapter heading and did not find one, so I also changed the chapter heading from g-code-overview to ngc-code-overview.

On the O-code page, I added reference to the interchangeability of the terms o-code and o-word, and moved the “best practice” note to the introduction. I then changed all references to O-code or O-word to lower case o- to match that best practice.

I’m hopeful the community will accept my efforts to clarify for other newbs who find this page in the future.

### DIFF
--- a/docs/src/gcode/o-code.adoc
+++ b/docs/src/gcode/o-code.adoc
@@ -15,11 +15,16 @@
 O-codes provide for flow control in NC programs. Each block has an
 associated number, which is the number used after O. Care must be taken
 to properly match the O-numbers. O codes use the letter 'O' not the
-number zero as the first character in the number like O100 or o100.
+number zero as the first character in the number like O100 or o100. o-codes are also sometimes called o-words and these terms are interchangeable. 
+
+[NOTE]
+Using the lower case 'o' makes it easier to distinguish from a 0
+that might have been mistyped. For example 'o100' is easier to
+see than 'O100' that it is not a '0'.
 
 == Numbering
 
-Numbered O codes must have a unique number for each subroutine,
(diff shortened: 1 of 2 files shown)


VectorHasting:
My changes crashed. Invalid format [.indent], but this converts using asciidoctor on my system.

Asciidoctor 2.0.20 [https://asciidoctor.org]
Runtime Environment (ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x64-mingw-ucrt]) (lc:IBM437 fs:UTF-8 in:UTF-8 ex:UTF-8)

I'm going to change that and re-issue the pull request.

PR #2711, opened 2023-10-26
debian: review of build instructions

smoe:
Transitioned the debian/README.source file to markdown, so the code to be executed can be clearly separated from the main text describing the process. Nothing urgent, I would just like to see this discussed and also the new repository in which the official Debian packages are maintained should be referenced.

### DIFF
--- a/debian/README.source
+++ b/debian/README.source
@@ -13,10 +13,25 @@ Packaging by the LinuxCNC CI system is not covered in this document, see
 for a brief description on that.
 
 This document describes the official packaging for debian.org.
-Packaging of LinuxCNC for debian.org is split between linuxcnc.git
-(in <https://github.com/LinuxCNC/linuxcnc.git>) and linuxcnc-gbp.git
-(in <https://github.com/LinuxCNC/linuxcnc-gbp.git>).
+Packaging of LinuxCNC for debian.org is split between:
 
+linuxcnc.git (in <https://github.com/LinuxCNC/linuxcnc.git>)::
+  This is the regular LinuxCNC repository. This already provides instructions
+  to build a Debian package in the subdirectory "debian/". 
+  Branches for this repository reflect the ongoing developments for LinuxCNC.
+linuxcnc-gbp.git (in <https://github.com/LinuxCNC/linuxcnc-gbp.git>)::
+  This repository has the same content as the release version, just that
+  branches now describe the release of Debian that the package shall be
+  used for. In particular, the script debian/configure was already executed.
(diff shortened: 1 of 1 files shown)


smoe:
@petterreinholdtsen , may I ask you to have a look? And then eventually also @andypugh .

petterreinholdtsen:
[Steffen Möller]
> @petterreinholdtsen , may I ask you to have a look? And then
> eventually also @andypugh .

The change itself look good to me.  Is this related to #2688?  Seem to
be similar in scope, and I suspect it make sense to include it as part
of that change.  But it is standalone, so no real need.

-- 
Happy hacking
Petter Reinholdtsen

smoe:
Some questions remain - like who uploads to the https://github.com/LinuxCNC/linuxcnc-gbp repository and where are the details described for that process.

andypugh:
> Some questions remain - like who uploads to the https://github.com/LinuxCNC/linuxcnc-gbp repository and where are the details described for that process.

https://github.com/LinuxCNC/linuxcnc/blob/master/debian/README.source

smoe:
That README.source is the file that this PR intends to improve :) I am just a bit blank on linuxcnc-gbp how it is meant to be used by whom. I think this patch improves the wording of the generic parts of the upload and someone feeling more comfortable about it may want to fill the gaps.

smoe:
@petterreinholdtsen @andypugh @SebKuzminsky The problem with linuxcnc-gbp is that it does not containt the .orig.tar.xz files in a pristine-tar branch. I would need to recreate that from the upstream branch but then  do not know if you have removed the debian directory or not or ... 
Should we not transition to Debian's salsa.debian.org for all the Debian-critical parts? Linuxcnc-gbp is counterproductive in an attempt in our ambition to find more Debian-folks to help us out on a technical level.

andypugh:
I don't have the time or expertise to commit to this. (You saw how long it took me to fail to build a package in Stuttgart....)

Apart from anything else, aren't we hampered by bugs in dependiences that we can do nothing about?

smoe:
Those have all been addressed from what I understand on https://tracker.debian.org/pkg/linuxcnc. In the contrary - they are waiting for us to complete the testing of their transition to the new version of boost. We should upload a 2.9.2, I think, I just do not really feel that I have a mandate to do so and also feel like I risk ruining linuxcnc-gbp (not knowing if I have commit privileges in the first place). Would like to hear from Petter and Seb on who should be uploading and if we should prefer salsa over linxucnc-gbp.

PR #2717, opened 2023-10-30
improve RCS debug output control

heeplr:
Currently a lot of rcs_print*() calls are "commented out" by blocks of 
```
set_rcs_print_destination(RCS_PRINT_TO_NULL);
...
set_rcs_print_destination(RCS_PRINT_TO_STDOUT);
```
which supresses output completely. On the other hand, ```DEBUG |= EMC_DEBUG_RCS``` or ```DEBUG |= EMC_DEBUG_NML``` in the INI file will result in ```set_rcs_print_flag(PRINT_EVERYTHING);``` which naturally generates very noisy output.

This PR introduces the config options
* ```RCS_DEBUG``` to control [print mode flags](https://github.com/LinuxCNC/linuxcnc/blob/master/src/libnml/rcs/rcs_print.hh#L107) similar to the current ```DEBUG``` option
* ```RCS_DEBUG_DEST``` to control the destionation of rcs_print*() output
* ```RCS_MAX_ERR``` to limit the amount of errors printed by ```rcs_print_error()```

### DIFF
--- a/docs/src/config/ini-config.adoc
+++ b/docs/src/config/ini-config.adoc
@@ -218,7 +218,7 @@ Do _not_ use a file extension of '.ini' for included files.
   You can put whatever you want here as long as you make it a single line long.
 * `DEBUG = 0` - Debug level 0 means no messages will be printed when LinuxCNC is run from a <<faq:terminal,terminal>>.
   Debug flags are usually only useful to developers. See src/emc/nml_intf/debugflags.h for other settings.
-* `RCS_DEBUG = 0` RCS debug messages to show. Print nothing (0) by default if EMC_DEBUG_RCS and EMC_DEBUG_RCS bits in
+* `RCS_DEBUG = 1` RCS debug messages to show. Print only errors (1) by default if EMC_DEBUG_RCS and EMC_DEBUG_RCS bits in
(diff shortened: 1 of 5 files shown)


andypugh:
You seem to be the only one who has ever cared about RCS output, but I see no reason that it shoudn't be available,

heeplr:
I take that as a compliment :smiley: 
It helps a lot during debugging and it might even be useful to check output in some tests.

phillc54:
Is there a way to inhibit the below messages at startup?
```
task: machine: 'metric'  version 'unknown'
after iniLoad()
emcCommandChannel==NULL, attempt to create
emcStatusChannel==NULL, attempt to create

task: machine: 'metric'  version 'unknown'
task: machine: 'metric'  version 'unknown'
```

heeplr:
> ```task: machine: 'metric'  version 'unknown'```

not currently. I thought it should be shown in the output in case the user having more than one .ini on the same host. It could be supressed when DEBUG = 0, but why?

> ```emcCommandChannel==NULL, attempt to create```
> ```emcStatusChannel==NULL, attempt to create```

Those are [relics](https://github.com/heeplr/linuxcnc/blob/master/src/emc/task/emcsvr.cc#L124) that were previously redirected to NULL and should be changed to ```print_rcs_debug(PRINT_NML_CONSTRUCTORS, ...``` to only output it with debug flags set.
I'll plan to prepare a new PR for that in the next days.

Issue #2716, opened 2023-10-30
multiple parallel linuxcnc sessions


rmu75:
As @rene-dev mentioned to me it is currently not possible to run tests in parallel, mainly because there can only be one linuxcnc instance running at a time. After investigating for a bit and looking at old branches in linuxcnc and machinekit I think such a feature could be realized with little effort.

CMS/NML uses SYSV shared memory buffer and semphores ("ipcs"), those resources are identified by a key (32bit integer) and those keys are derived from specification in the .nml file, usually those numbers are 1001-1005 or so.

HAL uses its own shared memory segments identified by key 0x48484c34 ("HHL\"") and 0x48414c32 ("HAL "), halscope involves another one with hardcoded key, for testing purposes that should not be relevant.

So in order to run tests in parallel, either the sessions have to be isolated in some way such that each session gets it's own isolated SYSV ipc namespace (that would involve some linux namespace trickery and probably need root) or there has to be a mechanism to change the keys to non-overlapping unique values. AFAIK that should also be sufficient for parallel test runs in user space.

I propose using an environment variable like "LINUXCNC_INSTANCE" containing a number that is used as a kind of offset, e.g. first instance uses 0x48414c32 and 1001/1002/1003, second instance uses 0x49414c32 and 1011/1012/1013 and so on.

Please comment if I missed something. If this feature would be welcome and no major obstacles turn up I plan on implementing it.

andypugh:
Would this only be for testing, or do you see it being useful for running (for example) a second toolpath on a multi-axis lathe? 
Would the separate instances have any way to communicate with each other? 
(A simple soution to this might be a HAL component that could create pins in two share memory spaces, but there are probably cleverer ways to do it)

rmu75:
Yeah could be, but synchronizing two instances will be painful.

I already have something working but get some very strange effects, seems I still miss a few locations.

rmu75:
> Would this only be for testing, or do you see it being useful for running (for example) a second toolpath on a multi-axis lathe? Would the separate instances have any way to communicate with each other? (A simple soution to this might be a HAL component that could create pins in two share memory spaces, but there are probably cleverer ways to do it)

I think for that use-case it would be cleaner to use one HAL instance and make task/etc... multi instance capable.

andypugh:
> I think for that use-case it would be cleaner to use one HAL instance and make task/etc... multi instance capable.

[...]

PR #2722, opened 2023-10-31
allow multiple parallel linuxcnc instances

rmu75:
Set environment variable LINUXCNC_INSTANCE to some numeric value. This value times 16 is added to shared memory keys and to tcp port numbers. If unset it is treated as 0. issue #2716 
This is a draft and has probably unresolved issues.

### DIFF
--- a/src/rtapi/uspace_rtapi_app.cc
+++ b/src/rtapi/uspace_rtapi_app.cc
@@ -488,9 +488,13 @@ _get_fifo_path() {
            "rtapi_app: RTAPI_FIFO_PATH and HOME are unset.  rtapi fifo creation is unsafe.");
        return NULL;
     }
-    if(s.size() + 1 > sizeof(sockaddr_un::sun_path)) {
-       rtapi_print_msg(RTAPI_MSG_ERR,
-           "rtapi_app: rtapi fifo path is too long (arch limit %zd): %s",
+    if (getenv("LINUXCNC_INSTANCE")) {
+        s += "-";
+        s += getenv("LINUXCNC_INSTANCE");
+    }
+    if (s.size() + 1 > sizeof(sockaddr_un::sun_path)) {
+        rtapi_print_msg(RTAPI_MSG_ERR,
+                        "rtapi_app: rtapi fifo path is too long (arch limit %zd): %s",
                sizeof(sockaddr_un::sun_path), s.c_str());
        return NULL;
     }
--- a/scripts/linuxcnc.in
+++ b/scripts/linuxcnc.in
@@ -732,7 +732,7 @@ function Cleanup() {
 
 # Name of lock file to check for that signifies that LinuxCNC is up,
 # to prevent multiple copies of controller
-LOCKFILE=/tmp/linuxcnc.lock
+LOCKFILE=/tmp/linuxcnc-$LINUXCNC_INSTANCE.lock
 
 # Check for lock file
 if [ -f $LOCKFILE ]; then
--- a/src/hal/utils/halcmd_main.c
+++ b/src/hal/utils/halcmd_main.c
@@ -356,8 +356,17 @@ static int release_HAL_mutex(void)
         rtapi_print_msg(RTAPI_MSG_ERR, "ERROR: rtapi init failed\n");
         return -EINVAL;
     }
+
+    unsigned long key = HAL_KEY;
+    /* get instance env variable */
+    const char *instance = getenv("LINUXCNC_INSTANCE");
+    if (instance) {
+        long offset = strtol(instance, NULL, 10) * 16;
+        key += offset;
+    }
+    
     /* get HAL shared memory block from RTAPI */
-    mem_id = rtapi_shmem_new(HAL_KEY, comp_id, HAL_SIZE);
+    mem_id = rtapi_shmem_new(key, comp_id, HAL_SIZE);
     if (mem_id < 0) {
         rtapi_print_msg(RTAPI_MSG_ERR,
             "ERROR: could not open shared memory\n");
--- a/src/libnml/buffer/shmem.cc
+++ b/src/libnml/buffer/shmem.cc
@@ -73,6 +73,13 @@ SHMEM::SHMEM(const char *n, long s, int nt, key_t k, int m):CMS(s)
     master = m;
     key = k;
 
+    /* get instance env variable */
+    const char* instance = getenv("LINUXCNC_INSTANCE");
+    if (instance) {
+        long offset = strtol(instance, NULL, 10) * 16;
+        key += offset;
+    }
+
     /* open the shared mem buffer and create mutual exclusion semaphore */
     open();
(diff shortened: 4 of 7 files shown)


src/rtapi/uspace_rtapi_app.cc
```diff
@@ -488,9 +488,13 @@ _get_fifo_path() {
...
     }
-    if(s.size() + 1 > sizeof(sockaddr_un::sun_path)) {
-       rtapi_print_msg(RTAPI_MSG_ERR,
-           "rtapi_app: rtapi fifo path is too long (arch limit %zd): %s",
+    if (getenv("LINUXCNC_INSTANCE")) {
+        s += "-";
+        s += getenv("LINUXCNC_INSTANCE");
+    }
+    if (s.size() + 1 > sizeof(sockaddr_un::sun_path)) {
+        rtapi_print_msg(RTAPI_MSG_ERR,
```
smoe, line 496: I feel like the instance number should appear in all the error messages, too.

smoe:
Just some questions coming to mind:
 * Can multiple instances communicate with each other?
 * How shall the hardware be distributed between the instances? Would it be desirable to have hardware shared between two instances?

BsAtHome:
> Can multiple instances communicate with each other?

They are supposed to be completely isolated. When they are, you could then, for example, parallelize tests; reducing runtime from 17 minutes to approx. 2 minutes on my machine. This I would like very much.

> How shall the hardware be distributed between the instances? Would it be desirable to have hardware shared between two instances?

No, you cannot <i>share</i> hardware. But, you can control multiple machines if you partition hardware between instances. But with cheap small boards available (like rpi5), it may be too late for this option.


There are also several instances of shared memory IDs missed in the patch set. I think the change should modify the rtapi_shmem_new() function instead of local changes (and libnml should be updated to use it).

smoe:
> > Can multiple instances communicate with each other?
> 
> They are supposed to be completely isolated. When they are, you could then, for example, parallelize tests; reducing runtime from 17 minutes to approx. 2 minutes on my machine. This I would like very much.

I would like that too. But people will start multiple instances on a single machine with some real mill attached to it. When these all read the very same config then I just dare to assume that may trigger some sort of mischief.

> > How shall the hardware be distributed between the instances? Would it be desirable to have hardware shared between two instances?
> 
> No, you cannot share hardware. But, you can control multiple machines if you partition hardware between instances.
Oops - that is what I had meant. "Hardware" to me was the same interface card.
> But with cheap small boards available (like rpi5), it may be too late for this option.

Low-latency communication between those instances I have not yet given up upon. 

> There are also several instances of shared memory IDs missed in the patch set. I think the change should modify the rtapi_shmem_new() function instead of local changes (and libnml should be updated to use it).

I kind of sensed that not too much effort would be made to extend its functionality. IIRC then ZeroMQ was at some point considered as a possible substitute. The prospect to add a communication between instances, likely also across a network, I feel like may warrant a transition to something novel. I had just asked the friendly AI folks about possibly preferable alternatives and they pointed me to NNG (https://nng.nanomsg.org/) or DDS (https://www.dds-foundation.org/who-is-using-dds-2/), the latter being already used by ROS 2. My vision is to share a series of cameras the monitor a milling process and help coordinating a robot or detect a broken end mill.

rmu75:
If there is interest in this i can certainly develop it further and polish it to a useable state. Maybe even get rid of legacy SYS V shared memory segments on the way.

> I kind of sensed that not too much effort would be made to extend its functionality. IIRC then ZeroMQ was at some point considered as a possible substitute. The prospect to add a communication between instances, likely also across a network, I feel like may warrant a transition to something novel. I had just asked the friendly AI folks about possibly preferable alternatives and they pointed me to NNG (https://nng.nanomsg.org/) or DDS (https://www.dds-foundation.org/who-is-using-dds-2/), the latter being already used by ROS 2. My vision is to share a series of cameras the monitor a milling process and help coordinating a robot or detect a broken end mill.

If you want to take a look https://github.com/rmu75/linuxcnc/tree/rs/zmq-experiments it is certainly possible to replace all that NML stuff with zeromq. I used flatbuffers as serialisation format, others are certainly possible too.

It's been more than a year, so caveat emptor. IIRC DDS is a real behemoth, using CORBA IDL, openDDS even includes TAO and ACE -- I wouldn't want to touch that, it has a very strong "comittee", "military procurement" and "enterprise" smell.

Communication between the "relaxed" and the "harder" realtime part of linuxcnc doesn't involve NML (or zeromq in my experiment) and probably never will. That's just going through shared memory and an ad-hoc message queue.

Synchronisation between multiple instances will probably involve some HAL components, and those are free to communicate with each other in any way they like/need, e.g. shared memory, zeromq or something completely different. The question is why you need two instances in parallel and what kind of operation needs to be synchronised. Doing synchronised thread cutting with two instances will probably induce some headaches.

smoe:
ROS2 has a bridge to ZeroMQ, so this would be fine. Synchronisation for the very moment to me only means an exchange of material between some cobots and the mill. Harmless. My motivation for ZeroMQ would be my current understanding that I do not ultimately need to distinguish local and remote instances - they would just have different addresses, right?

I yet fail to see use cases for hobbyists, I must admit. But maybe those surface as soon as the technology is out?

andypugh:
I _think_ that Machinekit looked at replacing NML with 0MQ. They might even have done it. 
https://github.com/machinekit/machinekit/tree/master/src/machinetalk
Seems to mention zmq in some commit messages.

rmu75:
IIRC what they did is add a bridge to "machinetalk" that is based on protocol buffers and 0MQ. But the project seems to be dead / abandoned.

smoe:
I just had a look at the conflict which seems easy to address. Just one comment, in
```
SHMEM::SHMEM(const char * /*n*/, long s, int /*nt*/, key_t k, int m)
  : CMS(s),
    key(k),
    bsem_key(-1),
    second_read(0),
    shm(NULL),
    sem(NULL),
    master(m),
    sem_delay(0.00001),
    mao{},
    use_os_sem(1),
    use_os_sem_only(1),
    mutex_type(OS_SEM_MUTEX),
    shm_addr_offset(NULL),
    bsem(NULL),
    autokey_table_size(0)
{

    /* Set pointers to null so only properly opened pointers are closed. */
    shm = NULL;

    /* save constructor args */
    master = m;
    key = k;
```
the save of constructor args is redundant or needs some extra comment.

BsAtHome:
> ```
> SHMEM::SHMEM(const char * /*n*/, long s, int /*nt*/, key_t k, int m)
>   : CMS(s),
>     key(k),
>...
>     shm(NULL),
>...
>     master(m),
>...
> {
>     shm = NULL;
>     master = m;
>     key = k;
> ```


All three assignments are redundant because they are performed in the class contructor's init list.
This was already the case before #3331 that fixed the remaining initializations. You can add a comment in the constructor, but adding the assignments is a bad move.

PR #2728, opened 2023-11-04
Fix missing G59.(1,2,3) decimal in combi_dro.py

Sigma1912:
Fixes gladevcp 'combi_dro' from this:
DRO missing decimal
to this:

combi_dro


Sigma1912:
Ah jeez, only I can produce two commits for something like this.
Need to go back and try and fix this.

PR #2736, opened 2023-11-06
refactor interpl linked list to use std::unique_ptr

rmu75:
continuing the work from https://github.com/LinuxCNC/linuxcnc/pull/2486 this refactors NML_INTERP_LIST

* use std::unique_ptr<NMLmsg> instead of NMLmsg*
* adapt task to allocate message with std::make_unique

improvements:
* avoids multiple copying of message contents
* avoids some dynamic allocations in std::vector
* potentially saves heap space
* clear ownership / lifetime of message
* next step to more idiomatic use of standard library

### DIFF
--- a/src/emc/nml_intf/interpl.cc
+++ b/src/emc/nml_intf/interpl.cc
@@ -93,7 +93,7 @@ std::unique_ptr<NMLmsg> NML_INTERP_LIST::get()
 
 void NML_INTERP_LIST::clear()
 {
-    if (true /*emc_debug & EMC_DEBUG_INTERP_LIST*/) {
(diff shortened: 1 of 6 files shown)


royka1:
@rmu75 This commit causes that the M67/M68 codes already get applied during the motion of the previous line, can you see where it goes wrong?

https://github.com/LinuxCNC/linuxcnc/issues/2987

rmu75:
@royka1 I will look into it, no idea at the moment.

rmu75:
It seems I inadvertently removed some calls to flush_segments() in emccanon.cc -- I will prepare a patch.

PR #2738, opened 2023-11-07
silence debug message

heeplr:
As [promised](https://github.com/LinuxCNC/linuxcnc/pull/2717#issuecomment-1793357001):

* silence NML constructor debug output when not debugging
* improve iniLoad() debug output and move inside iniLoad()

### DIFF
--- a/src/emc/task/emcsvr.cc
+++ b/src/emc/task/emcsvr.cc
@@ -177,13 +177,13 @@ int main(int argc, char *argv[])
 	    || emcErrorChannel == NULL)
 	) {
 	if (NULL == emcCommandChannel) {
-	    rcs_print("emcCommandChannel==NULL, attempt to create\n");
+	    rcs_print_debug(PRINT_NML_CONSTRUCTORS, "emcCommandChannel==NULL, attempt to create\n");
 	    emcCommandChannel =
 		new RCS_CMD_CHANNEL(emcFormat, "emcCommand", "emcsvr",
 				    emc_nmlfile);
 	}
 	if (NULL == emcStatusChannel) {
-	    rcs_print("emcStatusChannel==NULL, attempt to create\n");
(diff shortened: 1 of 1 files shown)


phillc54:
Thank you, that looks good.

Is it possible to get the task: message to show only once, I get it three times which seems a bit excessive.

Screenshot 2023-11-08 at 2 24 06 pm

heeplr:
This is caused because the INI config is parsed by multiple processes from ```emcsvr.cc```, ```emctaskmain.cc```, ```halui.cc``` and ```shcom.cc``` since they use the same code. So depending on your configuration, you see the output multiple times.

Part of this is actually a copy/paste bug - it shouldn't say "task" three times. Thanks for pointing my nose at it ;)

So I wouldn't say it's excessive but I agree that it doesn't make a lot of sense like that since it's pretty redundant right now.
Could you bare with the output if it would include the processname and pid? That way, the message would indicate that the config was parsed in the specific process. It would look like this:

```
linuxcncsvr (4326) emcsvr: machine 'my_mill'   version 'unknown'
...
milltask (4329) emctask: machine 'my_mill'   version 'unknown'
...
linuxcncrsh (4329) shcom: machine 'my_mill'   version 'unknown'
```

That way, every process would indicate which machine config is parsed. (I used that when testing multiple configs on different hosts without using ```scripts/linuxcnc```.)

If we remove all messages but one, there might be two candidates: emcsvr.cc and emctaskmain.cc since they run always. But when using remote processes, emctask (milltask) is not run on every host. That leaves only emcsvr.cc which is always run on every host. (Although it doesn't need to run afais, when there are no remote processes configured so this might change in the future).

Also the ```iniLoad()``` code should actually be deduplicated at one point. When that happens, all processes will call the same function which would output the message for every process again.

I'd prefer keeping them with additional name/pid output but I don't really have a strong opinion on that.

PR #2739, opened 2023-11-07
linuxcncrsh: fixes & cleanup

heeplr:
This fixes the current linuxcncrsh segfault bug and a lot of other issues. 
I also cleaned/modernized parts of the code, added "get update" command and re-enabled the test (disabled in https://github.com/LinuxCNC/linuxcnc/commit/eca96850f9348ff1489e66050bb0d3e5c369a50e), which should pass now.
The individual commits provide verbose details.


andypugh:
Thanks for giving linuxcncrsh some long overdue attention. 
I have not gone through the code in detail, but I am a little concerned that yu seem to have changed the name of one of the commands (from set_wait to wait_mode). My worry is that this will possibly break existing scripts in user configs.

heeplr:
> Thanks for giving linuxcncrsh some long overdue attention.

I think it's a great way to automate linuxcnc and do testing. I also created a quick, uncleaned [proof of concept](https://github.com/heeplr/linuxcnc/blob/linuxcncrsh-libcli/src/emc/usr_intf/emcrsh.cc) for a rewrite using [libcli](https://dparrish.com/pages/libcli) that should provide a modern UX and cleaner code.

> you seem to have changed the name of one of the commands (from set_wait to wait_mode)

Yes. I couldn't find out what "set" refers to other than "set wait mode". I suspected it's an artefact from times when the ```set``` command didn't exist and someone forgot to remove the "set_" from the name. That resulted in the weird ```set set_mode ...``` construction.

> My worry is that this will possibly break existing scripts in user configs.

Indeed there is a possibility. But since linuxcncrsh was almost completly broken beyond sending mdi commands and set_wait, the chance should be pretty low. I'm quite sure that it should fail gracefully, tho unless users don't check for NAKs/errors. So mentioning it in the ChangeLog/Release notes should probably be fine.

I'd prefer to keep it if my assumption is right and the name is "just wrong". But I have no strong feelings about changing it back since it wouldn't exactly hurt to rename it later (e.g. on a new major release).

heeplr:
> As an interim measure, could the code respond to both the new and old commands identically? (possibly with a deprecation warning?)

That's a good solution. I'll add that.

> We have a script to auto-update configs

Autoupdate would be nice, but I'd argue that it's not a config value but an API change and updating user scripts/code/3rd party applications probably is out of scope of linuxcnc.

heeplr:
I've added the alias and encountered two race conditions when homing and probing in the tests.
The issue is likely unrelated to linuxcncrsh and I disabled the test for now.

andypugh:
@heeplr Can you look at why linuxcnc builds on the buildbot are failing? It's the linuxcncrsh test. But it doesn't always fail, and I have not been able to reproduce it. 
Example: https://github.com/LinuxCNC/linuxcnc/actions/runs/7380505624/job/20077951641?pr=2823

heeplr:
It's failing because of a race condition. The test does [**set teleop_enable on**](https://github.com/LinuxCNC/linuxcnc/commit/8a905c0da2f061746928226e403626c023f98a6b#diff-5bdd17bf27d29dbf085eb9f300d4ca3d6aad02327d494abdfb7a990df3bdebd9R108) but a subsequent **get teleop_enable** still returns [**off**](https://github.com/LinuxCNC/linuxcnc/actions/runs/7380505624/job/20077951641?pr=2823#step:8:2188).
The initial **set wait_mode done** should have prevented that if I understand correctly. But it seems that it's not working. I couldn't find a mechanism that would check when a command is actually done executing so the call could block until then.

I encountered multiple of those race conditions with probing and homing aswell. I ended up [disabling test commands to prevent the test from failing](https://github.com/LinuxCNC/linuxcnc/commit/8a905c0da2f061746928226e403626c023f98a6b#diff-5bdd17bf27d29dbf085eb9f300d4ca3d6aad02327d494abdfb7a990df3bdebd9L90).
This roots way deeper than linuxcncrsh but I can't tell exactly where the culprit is. It probably exists for a long time but hasn't been tested.

Maybe just comment out failing commands for now?

PS: Is there another integration test that does something like "run a complete real-world job on simulated hardware"? I couldn't find any (but maybe missed it).

petterreinholdtsen:
[heeplr]
> It's failing because of a race condition. The test does set
> teleop_enable on but a subsequent get teleop_enable still returns off.

What about adding code after the "set teleop_enable on" call to loop for
a few seconds to wait for 'get teleop_enable' to return on, to avoid the
race condition?  It could warn about the issue in the process, but make
the test more robust.
-- 
Happy hacking
Petter Reinholdtsen

heeplr:
> to loop for a few seconds

Adding a few "sleep" commands came to my mind but with all the other racing commands, that would be a lot of "few seconds" and would probably increase total runtime of the test significantly. From a testing POV, this would certainly be better than disabling the tests completely.

Fixing the race conditions or removing the **set_wait** (now **wait_mode**) command - i.e. not giving guarantees to the user - would be best.

Note that the test isn't complete. I'm not familiar enough to create a representative "real world job" test and couldn't find a template to copy. But in theory, it should test a lot more different modes/operations/combinations.

petterreinholdtsen:
[heeplr]
> Adding a few "sleep" commands came to my mind but with all the other
> racing commands, that would be a lot of "few seconds" and would
> probably increase total runtime of the test significantly. From a
> testing POV, this would certainly be better than disabling the tests
> completely.

Just adding sleep is quite unreliable, as the various machines running
the tests have widely different performance.  This is why I suggest a
loop with a upper bound on the number of seconds sleeping.

-- 
Happy hacking
Petter Reinholdtsen

heeplr:
@petterreinholdtsen yes, such a loop would help. Feel free to add it. I'd rather spend my time hunting down the actual issue tho, since there must be some mechanism for "block-until-cmd-done" already, which is broken.

petterreinholdtsen:
[heeplr]
> @petterreinholdtsen yes, such a loop would help. Feel free to add
> it. I'd rather spend my time hunting down the actual issue tho, since
> there must be some mechanism for "block-until-cmd-done" already, which
> is broken.

While you hunt, I propose to simply skip the test until it become stable
in <URL: https://github.com/LinuxCNC/linuxcnc/pull/2824 >, to avoid
blocking other patches from getting a successful test.

-- 
Happy hacking
Petter Reinholdtsen

heeplr:
> to avoid blocking other patches from getting a successful test.

@petterreinholdtsen are we 100% sure that this isn't caused by some recent patch? Since the test did run fine for three weeks until now. Might be worth confirming before merging #2824 to silence the issue.

PR #2742, opened 2023-11-10
Fix G64 tolerance getting reset on E-stop

mozmck:
Without this patch, G64 tolerance is reset on each E-stop.  This caused me problems in some cases, and this fix works well for me.  I don't think this fix will hurt anything for anyone else but I'm not completely sure - hence this pull request.

### DIFF
--- a/src/emc/tp/tp.c
+++ b/src/emc/tp/tp.c
@@ -464,8 +464,6 @@ int tpClear(TP_STRUCT * const tp)
     struct state_tag_t tag = {{0}};
     tp->execTag = tag;
     tp->motionType = 0;
-    tp->termCond = TC_TERM_COND_PARABOLIC;
-    tp->tolerance = 0.0;
     tp->done = 1;
     tp->depth = tp->activeDepth = 0;
     tp->aborting = 0;
@@ -515,6 +513,8 @@ int tpInit(TP_STRUCT * const tp)
     tp->spindle.waiting_for_atspeed = MOTION_INVALID_ID;
 
     tp->reverse_run = TC_DIR_FORWARD;
+    tp->termCond = TC_TERM_COND_PARABOLIC;
+    tp->tolerance = 0.0;
 
     ZERO_EMC_POSE(tp->currentPos);
 


c-morley:
Can you test that it doesn't break the default loaded setting of g64. The default setting of G64 is different for imperial then metric configs,

mozmck:
I'm not sure how to do that?  The only thing I do in this is move these two lines from tpClear() to tpInit()

tp->termCond = TC_TERM_COND_PARABOLIC;
tp->tolerance = 0.0;

c-morley:
Remove any setting of g64 from config or program and see what it's set at. then estop and see if it changes.
 9b5e2a is the commit for default setting and it could have been broken by estop before your change.
It would be good to know.

c-morley:
Interestingly when I added default g64 P setting I didn't add a default Q setting. That is probably a mistake.

mozmck:
I'm going to have to figure out how to view the G64 setting again - it's been a few years since I did this.

c-morley:
Ahh yes - open the linuxcnc status program from the GUI and look for the settings entry.
The third entry id the P code the fourth the Q

andypugh:
Is this ready to merge?

Mjankor:
Thanks for posting this fix. I just got stung by this bug. 
While on the topic, why is this the default in LinuxCNC - "(Blend Without Tolerance Mode) G64 is the default setting when you start EMC. G64 is just blending and the naive cam detector is not enabled. G64 and G64P0 tell the planner to sacrifice path following accuracy in order to keep the feed rate up. "

I would have thought that accuracy was far more important for most CNC work than maintaining feedrate.

andypugh:
I think the logic is that if the machine is properly set up, and the G-code matches its capabilities, then blend without tolerance will give the best results. 
It's mainly with machines that have very low acceleration relative to the commanded speed where issues are seen.

andypugh:
> Is this ready to merge?

@mozmck ?

Mjankor:
Makes sense. I wonder what percent of users are using blend without tolerance.

PR #2756, opened 2023-11-21
Bar widgets

tjtr33:
pyvcp_widgets has class bar
previously the bar was like a vumeter
it measured live data sifference form minimum
This adds

-    vertical bars ( acts like thermometer or flow meter vs previous vumeter)

-    new xml tags are add but defaulted so previous code should not break

-    option to set texts above below right left or centered  on bar 

-   add pins for datas min max actual origin when None

-   add minimal visual style (single line) with user defined graphic 'thumb'

-   allow thumb  graphics to be tkintter canvas drawing commands (.py file)
   or bitmaps suited to PhotoImage ( eg pmngs)

-  new 'origin' tag allows graph to measure from user reference point
  previously the bars measured for minimum only

 documentation will show these variants visually

tests folder will visualize the different tsyles and show error trap examples


andypugh:
Thanks. Though I am not sure how much "life" there is left in PyVCP. (we have a bit of an abundance of riches in the VCP area)

PR #2757, opened 2023-11-21
Update getting-linuxcnc.adoc

dpslwk:
update Debian download link

### DIFF
--- a/docs/src/getting-started/getting-linuxcnc.adoc
+++ b/docs/src/getting-started/getting-linuxcnc.adoc
@@ -287,7 +287,7 @@ there are very few surviving systems that can not run a 64-bit OS.
 
 . Install Debian Bookworm (Debian 12), amd64 version.
   You can download the installer here:
-  https://www.debian.org/releases/buster/
+  https://www.debian.org/distrib/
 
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
Why did you pick the /distrib/ link over https://www.debian.org/releases/bookworm/ ?

dpslwk:
`distrib` will always link to iso downloads for the latest release, so its no longer a link that needs to be update every two years or so

and given the context of the sentence "You can download the installer here:",  `/releases/xxxxxxxx/` does not actual point to a way to download an iso, its at least 2 or 3 more pages clicks to get to a page with download information

petterreinholdtsen:
[dps.lwk]
> `distrib` will always link to iso downloads for the latest release, so
> its no longer a link that needs to be update every two years or so
>
> and given the context of the sentence "You can download the installer
> here:",  `/releases/xxxxxxxx/` does not actual point to a way to
> download an iso, its at least 2 or 3 more pages clicks to get to a
> page with download information

Make sense to me.  I guess the release manager have the final say.
-- 
Happy hacking
Petter Reinholdtsen

PR #2761, opened 2023-11-26
S code fix

c-morley:
Setting a S code to anything before M3 or M4 releases the brake and (internally) sets the direction.
Setting S to 0 will not actually set the spindle speed to 0 - it will turn off the spindle.


c-morley:
Assigned @andypugh  as you are the release manager. I would like a second look at the code.

Issue #2490, opened 2023-05-12
axis raises exception when reloading file repeatedly


rene-dev:
1. open sim/axix on master (have not checked 2.9)
2. click reload a few times
3. the following exception occurs:

```
Exception in Tkinter callback
Traceback (most recent call last):
  File "/usr/lib/python3.10/tkinter/__init__.py", line 1921, in __call__
    return self.func(*args)
  File "/home/rene/dev/linuxcnc/bin/axis", line 2292, in reload_file
    reload_file()
  File "/home/rene/dev/linuxcnc/bin/axis", line 1907, in reload_file
    shutil.copyfile(loaded_file, tempfile)
  File "/usr/lib/python3.10/shutil.py", line 234, in copyfile
    raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
shutil.SameFileError: '/tmp/tmpps6fzowe/axis.ngc' and '/tmp/tmpps6fzowe/axis.ngc' are the same file

```

cakeslob:
Hey Rene, it might have something to do with this one, as it was the most recent thing changed related to files. 
https://github.com/LinuxCNC/linuxcnc/pull/2419/commits/3db35073cab2d054724ea2d4f26f7cc440a3ef59

I am using an older 2.9 with a gcode editor tab which reloads the gcode file many times, and I not seen this issue before

rmu75:
I see this also. Strange thing is that i can usually start it with one single step (key T) and then continue (key S).

rmu75:
this https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/usr_intf/axis/scripts/axis.py#L1906
```python
    if refilter or not get_filter(loaded_file):
        # we copy the file to a temporary file so that even if it subsequently
        # changes on disk, LinuxCNC is parsing the file contents from the time
        # the user opened the file
        tempfile = os.path.join(tempdir, os.path.basename(loaded_file))
        shutil.copyfile(loaded_file, tempfile)
        open_file_guts(tempfile, False, False)
    else:
        tempfile = os.path.join(tempdir, "filtered-" + os.path.basename(loaded_file))
        open_file_guts(tempfile, True, False)
    if line:
        o.set_highlight_line(line)
```

interacts badly with this https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/usr_intf/axis/scripts/axis.py#L1174 (notice the global variable)
```python
loaded_file = None
[...]

PR #2762, opened 2023-11-27
Fix reloading file with axis

sundtek:
This fixes following error:

Exception in Tkinter callback
File "/usr/lib/python3.7/tkinter/init.py", line 1705, in call return self.func(*args)
File "/home/sundtek/devel/linuxcnc/bin/axis", line 2796, in touch_off_system reload_file(False)
File "/home/sundtek/devel/linuxcnc/bin/axis", line 1911, in reload_file shutil.copyfile(loaded_file, tempfile)
File "/usr/lib/python3.7/shutil.py", line 104, in copyfile raise SameFileError("{!r} and {!r} are the same file".format(src, dst)) shutil.SameFileError: '/tmp/tmp3d2pd26j/sidewall.nc' and '/tmp/tmp3d2pd26j/sidewall.nc' are the same file

opening the tmp file will overwrite loaded_file, the patch backups the original loaded_file and sets loaded_file back to the original name after loading the temp file.

### DIFF
--- a/src/emc/usr_intf/axis/scripts/axis.py
+++ b/src/emc/usr_intf/axis/scripts/axis.py
@@ -1895,6 +1895,7 @@ def run_warn():
     return 0
 
 def reload_file(refilter=True):
+    global loaded_file
     if running(): return
     s.poll()
     if not loaded_file:
@@ -1902,7 +1903,7 @@ def reload_file(refilter=True):
         return
     line = vars.highlight_line.get()
     o.set_highlight_line(None)
-
+    orig_loaded_file = loaded_file
     if refilter or not get_filter(loaded_file):
         # we copy the file to a temporary file so that even if it subsequently
         # changes on disk, LinuxCNC is parsing the file contents from the time
@@ -1913,6 +1914,7 @@ def reload_file(refilter=True):
     else:
         tempfile = os.path.join(tempdir, "filtered-" + os.path.basename(loaded_file))
         open_file_guts(tempfile, True, False)
+    loaded_file = orig_loaded_file
     if line:
         o.set_highlight_line(line)
 


sundtek:
My patch will basically disable #2418 it's not a solution either, just a temporary workaround (don't pull).
I did not test the filters yet
Is the original author @jes still around (and would like to update it)?
By the way thanks for documenting your original pull request, even though it doesn't work as expected
it at least documents what you would like to achieve.

@rmu75 mentioned that linuxcnc is reloading the file when running a job.

So "user" open should create a copy of the original file.
The reload function should have an option - reloaded by user or reloaded by the system.
Reload by user should replace the the copy / cached file.

@andypugh , @rmu75 what do you think about that?

jes:
I am still around but evidently don't understand enough about what is going on to recognise the problem, it works fine on my system.

I also don't see how your patch disables #2418 - it looks like it keeps it intact to me?

jes:
I've attached the axis file from my system (as `.txt`, because github).

If I recall correctly, I made this by manually applying my patch to the version of axis that was already installed.

[axis.txt](https://github.com/LinuxCNC/linuxcnc/files/13486066/axis.txt)

rmu75:
I think I explained everything in https://github.com/LinuxCNC/linuxcnc/issues/2490. Maybe I'm missing something, but behaviour here is constistent with that explanation. 

First, a temporary filename is constructed from random but fixed temp dir location plus base filename of "loaded_file".
Second, in open_file_guts, "loaded_file" is overwritten with the temporary file name.

So when you reload, loaded_file already points to the temporary file and will try to copy it over itself (and miss any changes to the original file btw).

jes:
Oh, I hadn't realised there was a specific reload button. I always reload by reopening the same file, that'll be why I didn't notice anything wrong.

andypugh:
Is this still a live PR? It sounds like it might have been overtaken by events?

andypugh:
I am still confused about whether this bug still exists. (I do think it's a bug, you should need to reload the file to run the modified version, it should be safe to edit the file on disk.)
Though I guess this is a modern "infinite memory" point of view. Historically on systems with limited resources you would probably have expected the current file on disk to be executed from disk rather than from memory.

jes:
If you're asking for my view: I'm still just using the version with my fix from #2418 (which breaks the reload button, which I never use).

But sadly I'm planning to move away from LinuxCNC soon anyway so my view doesn't count for much.

PR #2763, opened 2023-11-28
Update updating-linuxcnc.adoc

xsnoopy:
I guess this is a copy past error. It seems to be 2.9-uspace instead of 2.9-rtpreempt. Apt threw me an error when I tried it with 2.9-rtpreempt.


andypugh:
2.9-uspace does seem corrct. 
https://www.linuxcnc.org/dists/buster/

PR #2768, opened 2023-11-30
Fix RTAI bodging for 4.19

NTULINUX:


andypugh:
The kernel serials currently in the code match the kernel packages that exist in the LinuxCNC server:
https://www.linuxcnc.org/dists/buster/base/binary-amd64/
Feel free to add more, but we should keep the ones that we supply / have supplied.

NTULINUX:
OK, will fix, thanks!

PR #2769, opened 2023-12-02
Fixed Inappropriate Logical Expression

asifhaider:
# Details

While triaging your project, our bug-fixing tool generated the following message(s)-

> In file: [gmoccapy.py, method: _button_pin_changed](https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/usr_intf/gmoccapy/gmoccapy.py#L5516), a logical equality check operation was performed with the same operand on both sides. The comparison operation always returns either true or false. Such logical short circuits in code lead to unintended behavior. iCR suggested that the logical operation should be reviewed for correctness.

In this specific case, the line 5516 of the file [gmoccapy.py](https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/usr_intf/gmoccapy/gmoccapy.py#L5516) is as follows:

```python
if number is not number:
```

In python, the `is` operator checks for identity. This means that the `is` operator checks whether the two operands refer to the same object or not. In this case, the same integer operand `number` is on both sides of the `is not` operator. Therefore, it will always evaluate to False, creating a logical short circuit. Consequently, the code block that follows the logical condition check will never actually be entered or executed. 

# Changes

- To ensure logical correctness and efficiency, the lines 5516-5518 are removed.

Suggestions related to these changes are welcomed. 

# CLA Requirements

This section is only relevant if your project requires contributors to sign a Contributor License Agreement (CLA) for external contributions.

All contributed commits are already automatically signed off.

The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit this work under the same license and agrees to a Developer Certificate of Origin (see [https://developercertificate.org/](https://developercertificate.org/) for more information).

[Git Commit SignOff documentation](https://developercertificate.org/)

# Sponsorship and Support

This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -5513,9 +5513,6 @@ class gmoccapy(object):
             return
 
         number = int(pin.name[-1])
-        if number is not number:
-            LOG.debug("Could not translate {0} to number".format(pin.name))
-            return
 
         button = self._get_child_button(location, number)
         if not button:


hansu:
Thanks for spotting this. Fixed in 457af9370dc4018dfc585d428b8a1c099e9f9c7e.

andypugh:
Is this necessarily wrong? "number is not number" might be intended as a NaN check. 
The N != N version is fairly common in C if it is desired to not use external libraries, and this also works (efficiently) in Python. 
The odd thing here, though, is the use of "is not" rather than "!=". 
And this only makes any sense if int() can return a NaN. It appears that it will actually raise an error rather than return an error value. 

I think that is this case a real bug has been located, I am just saying that in the general case N != N isn't always indicative of an error.

PR #2772, opened 2023-12-07
fix hardcoded description in Spanish language (Debian bug #1057312)

hansu:
Simple fix for bug reported on debian.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1057312


andypugh:
CI is passing now, so I will merge this.

PR #2775, opened 2023-12-10
docs: triggered while translating

smoe:
I did about 100 translations today and those triggered these few changes. Should be a straight-forward commit.

### DIFF
--- a/docs/src/config/ini-config.adoc
+++ b/docs/src/config/ini-config.adoc
@@ -1295,19 +1295,18 @@ image::images/encoder-scale.png[align="center"]
 
 The _<num>_ specifies the spindle number 0 ... (num_spindles-1) +
 The value of _num_spindles_ is set by `[TRAJ]SPINDLES=` . +
-By default maximum velocity of the spindle in forward and reverse is approximately
-  2147483000 RPM. +
+By default maximum velocity of the spindle in forward and reverse is approximately 2147483000 RPM. +
 By default minimum velocity of the spindle in forward and reverse is 0 RPM. +
 By default the increment is 100 RPM. +
 You change these default by setting the following INI variables:
 
 [NOTE]
-These settings are for the motion controller component. Control screens can limit
- these settings further. 
+These settings are for the motion controller component.
+Control screens can limit these settings further. 
 
 * `MAX_FORWARD_VELOCITY = 20000`
   The maximum spindle speed (in rpm) for the specified spindle. Optional.
-  This will also set MIN_FORWARD_VELOCITY to the negative value unless overridden.
+  This will also set MAX_REVERSE_VELOCITY to the negative value unless overridden.
 * `MIN_FORWARD_VELOCITY = 3000`
   The minimum spindle speed (in rpm) for the specified spindle. Optional.
   Many spindles have a minimum speed below which they should not be run.
--- a/docs/src/getting-started/updating-linuxcnc.adoc
+++ b/docs/src/getting-started/updating-linuxcnc.adoc
@@ -193,11 +193,11 @@ interpreter is, then this section does not affect you.
 A seldom-used feature of LinuxCNC is support for pluggable interpreters,
 controlled by the undocumented `[TASK]INTERPRETER` INI setting.
 
-Versions of LinuxCNC before 2.9.0 used to handle an incorrect
+Versions of LinuxCNC before 2.9.1 used to handle an incorrect
 `[TASK]INTERPRETER` setting by automatically falling back to using the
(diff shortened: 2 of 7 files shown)


smoe:
@hansu - please kindly have a look.

docs/src/config/ini-config.adoc
```diff
@@ -1295,19 +1295,19 @@ image::images/encoder-scale.png[align="center"]
...
-These settings are for the motion controller component. Control screens can limit
- these settings further. 
+These settings are for the motion controller component.
+Control screens can limit these settings further. 
 
 * `MAX_FORWARD_VELOCITY = 20000`
   The maximum spindle speed (in rpm) for the specified spindle. Optional.
-  This will also set MIN_FORWARD_VELOCITY to the negative value unless overridden.
+  This will also set MIN_FORWARD_VELOCITY to the negative value unless overridden
+  by MAX_REVERSE_VELOCITY.
```
hansu, line 1310: That doesn't sound logical to me. Not sure by which value this value could be overwritten.

hansu (changes requested):
One little thing...

docs/src/config/ini-config.adoc
smoe: Typo - by MIN_REVERSE_VELOCITY would be correct (when looking at the description of MIN_REVERSE_VELOCITY).  Thank you for spotting that.

hansu, line 1310: No, `MAX_REVERSE_VELOCITY` was correct. But it seems that it was wrong in the docs. So correct would be this i think:
```suggestion
  This will also set `MAX_REVERSE_VELOCITY` to the negative value unless overridden
  by MAX_REVERSE_VELOCITY.
```
But then the addendum "  by MAX_REVERSE_VELOCITY" doesn't make sense.

smoe: What I thought this meant is that you only set the forward direction (the one typically used for cutting as I interpret it) and the reverse direction is auto-set with it to the same value unless another value is given. The transition from MIN to MAX in the same paragraph is then to avoid.

smoe: Re-read it all again and agree that we have spotted a problem with the documentation. I think [e6384fe](https://github.com/LinuxCNC/linuxcnc/pull/2775/commits/e6384fe771061ce8773e01ed32b511ed6c3eef0f) now gets it right. Shall I squash the two?

smoe: squashed.

docs/src/getting-started/updating-linuxcnc.adoc
```diff
@@ -193,11 +193,11 @@ interpreter is, then this section does not affect you.
 A seldom-used feature of LinuxCNC is support for pluggable interpreters,
 controlled by the undocumented `[TASK]INTERPRETER` INI setting.
 
-Versions of LinuxCNC before 2.9.0 used to handle an incorrect
+Versions of LinuxCNC before 2.9.1 used to handle an incorrect
```
hansu, line 196: I think this was fixed in 2.9.0 and there was no change to 2.9.1, so I would not increase the version number here.

docs/src/getting-started/updating-linuxcnc.adoc
```diff
@@ -193,11 +193,11 @@ interpreter is, then this section does not affect you.
...
 `[TASK]INTERPRETER` setting by automatically falling back to using the
 default G-code interpreter.
 
-As of 2.9.0, an incorrect `[TASK]INTERPRETER` value will cause
+As of 2.9.1, an incorrect `[TASK]INTERPRETER` value will cause
```
hansu, line 200: Same here

hansu (changes requested):

docs/src/config/ini-config.adoc
hansu: Yeah nice, thanks!

docs/src/getting-started/updating-linuxcnc.adoc
smoe: Seems like I rushed something. Thank you.

docs/src/getting-started/updating-linuxcnc.adoc
smoe: Changed that to "Since ..".

hansu: Yeah that sounds better.

hansu (approved):

hansu:
Thanks

PR #2776, opened 2023-12-11
Fix module 'locale' has no attribute 'format'

dwrobel:
Fixes the following error on python 3.12:
  File "/usr/lib/python3.12/site-packages/qtvcp/widgets/origin_offsetview.py",
   line 383, in periodic_check
    self.reload_offsets()
  File "/usr/lib/python3.12/site-packages/qtvcp/widgets/origin_offsetview.py",
   line 270, in reload_offsets
    degree_tmpl = "%{}.2f".format(len(locale.format(tmpl, 0)))
                                      ^^^^^^^^^^^^^
 AttributeError: module 'locale' has no attribute 'format'. Did you mean: '_format'?

See also: https://github.com/python/cpython/issues/94226

### DIFF
--- a/lib/python/qtvcp/widgets/origin_offsetview.py
+++ b/lib/python/qtvcp/widgets/origin_offsetview.py
@@ -267,18 +267,18 @@ class OriginOffsetView(QTableView, _HalWidgetBase):
         else:
             tmpl = self.imperial_text_template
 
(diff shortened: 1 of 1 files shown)


hansu:
Looks good to me. Should better go into 2.9 I think.

c-morley:
I think locale is an artifact from converting from gladevcp anyways. Will back port/remove if required later. Thanks.

Issue #2784, opened 2023-12-11
/src/emc/usr_intf/emcrsh.cc function getToolOffset. Wrong param in snprintf.


johnchikov:
I think in emcrsh.cc (line 1677) use wrong argument type in snprintf instead of %f is %d

smoe:
This refers to https://github.com/LinuxCNC/linuxcnc/blob/7d9b5e504c1121f6d0be36120cd03a75ebf206da/src/emc/usr_intf/emcrsh.cc#L1677 and the question what the type emcStatus->task.toolOffset.tran.z may possibly be. That should be a double, right? And then, indeed, a "%lf" should possibly substitute the "%d".

If that is truly an error then this is also in version [2.9](https://github.com/LinuxCNC/linuxcnc/blob/fb4d6a175fcbfd279c069b2041173d90040a3f69/src/emc/usr_intf/emcrsh.cc#L1716), [2.8](https://github.com/LinuxCNC/linuxcnc/blob/fe43ab453bfd9728dbe32375403f75b53f7ce027/src/emc/usr_intf/emcrsh.cc#L1716), [2.7](https://github.com/LinuxCNC/linuxcnc/blob/d7ded6ef5e4c390267276dd72a59bfc4e319b5ce/src/emc/usr_intf/emcrsh.cc#L1602) and earlier, so one could assume.

smoe:
I just checked. It is a double since the ToolOffset is an [EmcPose](https://github.com/LinuxCNC/linuxcnc/blob/7d9b5e504c1121f6d0be36120cd03a75ebf206da/src/emc/nml_intf/emcpos.h#L19) , trans is a [PmCartesian](https://github.com/LinuxCNC/linuxcnc/blob/7d9b5e504c1121f6d0be36120cd03a75ebf206da/src/libnml/posemath/posemath.h#L121) of which the coordinates are doubles.

rmu75:
%f (or %g / %e) is the correct specifier for a double in printf, so as it is now it probably prints some garbled value. Even if repaired that method is somewhat incomplete as there may very well be additional offsets (besides Z).

smoe:
Since the value was garbled, the function cannot have been used much and, we have some freedom to correct it. What do you suggest? Add Z to the function name - and add variants for X and Y?

rmu75:
[...]

PR #2787, opened 2023-12-13
snprintf uses %d on a double (issue 2784)

smoe:
This PR reacts to issue https://github.com/LinuxCNC/linuxcnc/issues/2784 .


smoe:
Please note, the PR is against version 2.6, which may (or may not) be an unreasonably early version to see this patch migrating upwards.

PR #2788, opened 2023-12-13
Refactor count_names() for improved clarity

havardAasen:
Simplified the logic by changing to a while loop and removing a variable.

### DIFF
--- a/src/emc/motion/motion.c
+++ b/src/emc/motion/motion.c
@@ -213,16 +213,12 @@ static void emc_message_handler(msg_level_t level, const char *fmt, va_list ap)
     va_end(apc);
 }
 
-int count_names(char *names[]){
-  int namecount = 0;
-  int i;
-  for (i = 0; i < MAX_IO; i++) {
-    if (((names[i] == NULL) || (*names[i] == 0))){
-      break;
-    }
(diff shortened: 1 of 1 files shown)


andypugh:
I think that your code continues an error in the existing code, that the limit may not be MAX_IO for all cases where the function is called. This isn't currrently a problem. https://github.com/LinuxCNC/linuxcnc/blob/2.9/src/emc/motion/emcmotcfg.h#L33
But it feels like the function should be passed the relevant parameter rather than the local MAX_IO.

havardAasen:
> I think that your code continues an error in the existing code, that the limit may not be MAX_IO for all cases where the function is called. This isn't currrently a problem. https://github.com/LinuxCNC/linuxcnc/blob/2.9/src/emc/motion/emcmotcfg.h#L33 <https://github.com/LinuxCNC/linuxcnc/blob/2.9/src/emc/motion/emcmotcfg.h#L33>
> But it feels like the function should be passed the relevant parameter rather than the local MAX_IO.

Yes, I agree. Then we can remove `MAX_IO` and rather use `EMCMOT_MAX_*`
when we define the arrays, and send the macro in as a `max_len`
parameter.

PR #2789, opened 2023-12-13
Update python-interface.adoc

Sigma1912:
Remove reference to 'axes'  attribute as it has been removed from linuxcnc.stat


jethornton:
I feel it's wrong to remove this from the docs, the stat should be put back into src/emc/task/taskinf.cc as it's based on a required ini file entry and gives you the number of axes without having to be a math wizard.

I don't know why Jeff deleted it with commit
43deeedd065faf31254933ee56af01c314823827

Sigma1912:
Wouldn't this give you the number of axes:

n = s.axis_mask
axes = n.bit_count()

Sigma1912:
Just tried it: 

  number_of_axes =  (s.axis_mask).bit_count()

jethornton:
Well I would never been able to figure that out because I suspect like me most folks never heard of bit_count().

However if that snippet of code was in the manual then everyone that was trying to get the axis count by looking at the status documents would then know how to get the number of axes.

jethornton:
While bit_count() works it feels like a work around for a previously working stat item s.axes. Seems simpler to just put the s.axes back in...

jethornton:
Instead of just deleting the axes from the docs and letting people who are not python programmers and don't have a clue how to get that information fumble around for days trying to figure it out why not put that code in the  docs?

petterreinholdtsen:
[jethornton]
> Instead of just deleting the axes from the docs and letting people who
> are not python programmers and don't have a clue how to get that
> information fumble around for days trying to figure it out why not put
> that code in the docs?

I would also recommend not removing the axes entry completely from the
documentation, as it would confuse those that try to figure out why
their old setup no longer is working.  I believe it is better to move it
to a list of no longer present attributes, mention when it was removed
and explain how to get the same value in a different way using your
example code.

-- 
Happy hacking
Petter Reinholdtsen

jethornton:
I would add to Petter reply to leave the axes in the current spot so it's easy to find and add what Petter suggested.

Sigma1912:
I buggered it up and I don't have time to fix it.
Maybe jethornton can  create a pull request and I can then delete this one.

PR #2790, opened 2023-12-13
Fix warning for possible string truncation

havardAasen:
We check the length of the string before using `strncpy()`, we check with `>=` which means that the string must be one smaller and therefore have room for at least one NULL terminating character.

The previous solution also checked for NULL terminated strings but the compiler wasn't satisfied with it happening after the command to `strncpy()`.

For the error message, the unnecessary cast to unsigned long is removed, and the format string has been changed to `zu`.

### DIFF
--- a/src/emc/rs274ngc/interp_internal.hh
+++ b/src/emc/rs274ngc/interp_internal.hh
@@ -638,7 +638,7 @@ and is not represented here
 
 */
 #define STACK_LEN 50
-#define STACK_ENTRY_LEN 80
+#define STACK_ENTRY_LEN 256
 #define MAX_SUB_DIRS 10
 
 struct setup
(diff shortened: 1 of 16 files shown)


andypugh:
How many does this catch? I have notice that the compile log is sprinkled with hundreds of string truncation warnings.

havardAasen:
On 13.12.2023 17:13, andypugh wrote:
> How many does this catch? I have notice that the compile log is sprinkled with hundreds of string truncation warnings.
> 
Only three, it was all from that file. If you are satisfied with
how i solved it, I can see if I can't fix some more of them, but not
hundreds.

I forgot to check if they are present in 2.9, should i target any
specific branch?

andypugh:
There are many truncation warnings when compiling 2.9. So I would suggest targetting that. I probably _won't_ merge them until after 2.9.2 though.

havardAasen:
This fixes quite a lot of the warning for string truncation. The one responsible for most of them was fixed with 7f54de8.

andypugh:
I changed my mind, there was quite a backlog of 2.9 fixes which I am working through. This might as well go in too.

PR #2791, opened 2023-12-14
revert changes for strncat()

havardAasen:
Fixes the regression with tests `remap/fail/args.1` and `remap/fail/args.2`.


andypugh:
Nice to see the tests doing their job.  But I wonder why the CI didn't pick this up?

havardAasen:
> Nice to see the tests doing their job. But I wonder why the CI didn't pick this up?

Don't know, might have been something GitHub didn't like, when I force pushed and changed the targeting branch at the same time.

PR #2792, opened 2023-12-16
Update glcanon.py

Sigma1912:
Fix overflow error when clicking on the zoomed out preview or rapidly clicking on individual lines.

### DIFF
--- a/lib/python/rs274/glcanon.py
+++ b/lib/python/rs274/glcanon.py
@@ -592,9 +592,8 @@ class GlCanonDraw:
 
             try:
                 buffer = glRenderMode(GL_RENDER)
-            except OverflowError:
-                self.select_buffer_size *= 2
-                continue
+            except:
+                buffer = []
             break
 
         if buffer:


andypugh:
Are you sure that this is correct? 
The existing code attempts to double the size of the buffer on overflow (I think) whereas your new code appears to clear the buffer on any error. 
Should it perhaps be able to double the buffer size up to a certain limit?

Sigma1912:
I'm not certain that this is the best solution but it certainly seems to get rid of the errors while not breaking the functionality, at least in the testing I have done. As I understand it, we want the line number of the selected segment, which is the value of 'names[0]', to highlight the line in the Gcode window. The default buffer size is 100 and unless I'missing something, which is very much possible, I don't see why we would need to enlarge the buffer at all as we can only highlight the one line in the gcode.

hansu:
I would suggest this:

``` diff
            try:
                buffer = glRenderMode(GL_RENDER)
            except OverflowError:
                self.select_buffer_size *= 2
                continue
+           except OpenGL.error.GLError as e:
+             if e.err == GL_STACK_OVERFLOW:
+                   self.select_buffer_size *= 2
+               continue
            break
```

Sigma1912:
Still kind of begs the question as to why we would want to ingest hundreds or even thousands of segments to then only pick out a single one of them.
The only effect this has (as far as I can tell) is that you can click on a preview that is zoomed all the way out and you will get some line in the gcode to highlight. With the original pull request you will not be able to select anything until you zoom in enough to actually be able to make a sensible selection.

andypugh:
Feel free to submit a better solution, if you think of one.

Issue #2532, opened 2023-06-20
RFC: Re-work handling of coolant / lube / worklight pins


andypugh:
Prompted by #2471 and #2531 (the former trying to remove pins, the latter to add them) 

Both the references above show how much code is needed to control a single HAL pin. 
It seems that all IO pins could  be switched with a single message, and handled in a standardised way.  Still a lot of code, but at least it isn't redundant code. 

Furthermore, there will frequently be user requirements for additional pins (through-tool coolant, separate lube pump for each spindle, several worklights, possibly even include a stack-light in this) 

Initial proposal:

1)  Combine the existing pins in a single message, passing a 32 (or 64)-bit array of values. 

Sub-Proposals (for discussion) 

A)  Alias flood, mist (others?) to M64 / M65 using a P-word > 100 (or, use a different index word). There is currently a limit of 64 for the existing, conventional, digital IO pins, so the higher index number should be safe. A letter code might be better, though. I think that DHIJK are possibilities. D for "Digital" or I for "IO" would both fit (I haven't checked their current type) 
 
B) Allow override of the names flood / mist etc from the INI file, and also allow adding of extra signals the same way

Proposal (B) should include GUI updates to display GUI buttons on-demand in the same way as they currently do for mist/flood etc, but using the names from the INI. HALUI should use these names for HAL pins too. 

The pins should probably use NIST logic, so I think that we would also need to extend M66 to read back the status of these pins.

andypugh:
To elaborate on the motivation for this: The idea is to have a set of HAL pins that are equally easy to manipulate from G-code, HAL and UI code. You can achieve the same thing now using the motion.digital pins, and HAL and toggle-2-nist, mux and other logic. But even then the GUIs don't know about it.

hansu:
The idea sounds good to have all those in one message. Would that change the pin names?
Flood, mist and lube have their own messages currently, did I get that right?

grandixximo:
Why not using motion digital output pins? and set an alternative name for display from ini file.
I have found difficulty figuring out the toggle-2-nist logic when there are specific timing delays requirement at abort, and when I'd like to run it outside of a running program, like MDI for testing, maybe I just need look more into it.
But I was also thinking to help with user with similar questions
https://forum.linuxcnc.org/38-general-linuxcnc-questions/43038-motion-digital-output-value-on-stop
[...]

PR #1352, opened 2021-11-03
motion - Adding Names Arrays

satiowadahc:
Per #1349 Adding Parameters to motion to allow for labeling IO pins. 

WIP is pending a little more testing. And probably some documentation updates

Issue #1349, opened 2021-11-02
Add array of strings to label digital IO and analog IO to Motion


satiowadahc:
Writing thoughts down as this would be a nice to have, I might get to it later this week

Add three strings of arrays to motion:
- One for labeling digital pins 
- One for labeling analog pins
- One for labeling misc errors.

It should be an error if number of IO and the arrays are provided. 

EG:
```
din_names=endpoint_prox
dout_names=airvalve,vacuumservo,warningled
misc_errors=safety_door,motor_temp,chips_full
```
these would make hal pins:
motion.din-endpoint_prox
motion.dout-airvalve,
motion.dout-vacuumservo
motion.dout-warningled

motion.err-safety_door (Or keep the number but change the resulting error message)
motion.err-motor_temp
motion.err-chips_full

Since C variable names and halpins don't need to line up this should be pretty minimal changes

andypugh:
The more I think about this the more I like it.

SebKuzminsky:
Don't net names and pin aliases cover some of this use case?

satiowadahc:
Some. but sometimes I'm unlinking from halcmd so I can debug a mill and manually toggle a pin, and what pin was that again? let me just pull up the hal file... or did I change it there already.., ?

andypugh:
It feels like a natural extension of the "names=" option that many of the other HAL components have. It should only touch the code fairly lightly too, nothing changes after the point that the pin names are defined.

dngarrett:
It's hard to improve on the existing halcmd alias provisions which work on any hal pin or parameter

SebKuzminsky:
The `names` parameters that some/many modules have operates on instance names, not pin names.

I agree with Dewey we already have this requested feature, in the form of hal pin aliases: http://linuxcnc.org/docs/2.8/html/man/man1/halcmd.1.html#COMMANDS

c-morley:
This seems cleaner then alias. What is the complaint of having this option, other then there is another way to do it (which then using alias, shows two pins that do the same thing)?

satiowadahc:
Its another convenience function in my opinion. Defining parameters in the INI also makes it nicer to explicitly access in python.

 MR is up, I'm gonna keep using it, hope the community finds some use in it.

"If a programmer built a car, there would be 7 ways to turn the radio up. Including the using the steering wheel - Useful feature I must say"

c-morley:
[...]

PR #2794, opened 2023-12-16
Improves named IO pins

havardAasen:
There is still some limitations to the named arrays. If both `names_dout` and `names_din` is set, they have to be the exact same size. The same is true for `names_ain` and `names_aout`. Currently, we do not verify this.

Marking this as draft to see if the solution is accepted, I also thought to update the man-page for the named pins  `names_dout`, `names_din` and `names_ain`, `names_aout`.

There is also the issue with not sending in the max length to `count_names(), not sure if you want that in 2.9 as well @andypugh?

### DIFF
--- a/docs/man/man9/motion.9
+++ b/docs/man/man9/motion.9
@@ -1,10 +1,10 @@
-.TH MOTION "9" "" "LinuxCNC Documentation" "HAL Component"
+.TH MOTION "9" "2023-12-22" "LinuxCNC Documentation" "HAL Component"
 
 .SH NAME
 motion \- accepts NML motion commands, interacts with HAL in realtime
 
 .SH SYNOPSIS
-\fBloadrt motmod [base_period_nsec=\fIperiod\fB] [base_thread_fp=\fI0 or 1\fB] [servo_period_nsec=\fIperiod\fB] [traj_period_nsec=\fIperiod\fB] [num_joints=\fI[1-16]\fB] [num_dio=\fI[1-64]\fB] [num_aio=\fI[1-64]\fB] [num_misc_error=\fI[0-64]\fB] [num_spindles=\fI[1-8]\fB]\fR  \fB[unlock_joints_mask=\fR\fIjointmask\fR\fB]\fR \fB[num_extrajoints=\fI[0-16]\fB]\fR
+\fBloadrt motmod [base_period_nsec=\fIperiod\fB] [base_thread_fp=\fI0 or 1\fB] [servo_period_nsec=\fIperiod\fB] [traj_period_nsec=\fIperiod\fB] [num_joints=\fI[1-16]\fB] [num_dio=\fI[1-64]\fB | names_dout=\fIname[,...]\fB names_din=\fIname[,...]\fB] [num_aio=\fI[1-64]\fB | names_aout=\fIname[,...]\fB names_ain=\fIname[,...]\fB] [num_misc_error=\fI[0-64]\fB] [num_spindles=\fI[1-8]\fB]\fR  \fB[unlock_joints_mask=\fR\fIjointmask\fR\fB]\fR \fB[num_extrajoints=\fI[0-16]\fB]\fR
 
 The limits for the following items are compile-time settings:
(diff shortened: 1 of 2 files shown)


andypugh:
With this patch num_dio or num_aio is set to the minimum of count(inputs) and count(outputs). 
I feel it might be better to take the max(count(in), count(out) and then modify the check here: https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/motion/motion.c#L489C5-L489C5
to create default-named pins when it runs out of names. 
Currently users can define names for lots of inputs (say) and then not have them be created unless they have an equal number of outputs.

havardAasen:
Are you thinking we can split up the input and output? Like 15 inputs and 5 outputs, and of those inputs 10 is named and 5 is default-named, with something similar with the outputs, but only 5 outputs? Input and output does not have same amount of pins.

Or is the intention that if I want to create 10 named inputs and 5 named outputs, It would actually generate 10 named inputs, 5 named outputs + 5 default-named outputs? Input and output has same amount of pins.

This last one would give a limitation of 32 named pins, not sure if this will become a problem.

andypugh:
I was thinking of the latter, but may have missed a detail. Do the IOs of each type share an array with each being input or output (I would check, but it's late here, and probably later where you are). My (un-checked) impression was that they were separate but of equal size. 

I am _really_ keen to get 2.9.2 out (for reasons due to the Debian bugfix timescale) Do you think that the current partial fix is good enough for the moment?

havardAasen:
> I was thinking of the latter, but may have missed a detail. Do the IOs of each type share an array with each being input or output (I would check, but it's late here, and probably later where you are). My (un-checked) impression was that they were separate but of equal size.
> 
You might be correct , I based my assumption on these lines https://github.com/LinuxCNC/linuxcnc/blob/1db56381a8a1bb336b4cc9ab424d736eaecede64/src/emc/motion/emcmotcfg.h#L38-L40
and since it combines input and output pins into the same variable I assumed the same register. But as you say, it would make more sense if they were separate.

> I am _really_ keen to get 2.9.2 out (for reasons due to the Debian bugfix timescale) Do you think that the current partial fix is good enough for the moment?

Not sure if I want to have any opinion on that, for 2.9.1
```
motmod num_misc_error=10 # Creates 10 error pins
motmod names_misc_errors=one,two # Creates 0 error pins
```
In it's current state, this has been flipped
```
motmod num_misc_error=10 # Creates 0 error pins
motmod names_misc_errors=one,two # Creates 2 error pins
```
The `num_misc_error` was the only one documented in 2.9.1 as well.

andypugh:
As you spotted, I managed to commit an incomplete fix. I have now committed the fix that I intended (and tested). 
I think that what you have here is better, but I think it needs more careful assessment. 
The issue with the error pins was a clear bug, whereas the general improvement affects code that has (apparently) been working OK for some time. (Maybe not that long, I wasn't even aware that you could give names to the IO pins)

havardAasen:
> As you spotted, I managed to commit an incomplete fix. I have now committed the fix that I intended (and tested). I think that what you have here is better, but I think it needs more careful assessment.

Are you thinking on which solution to go for?

> The issue with the error pins was a clear bug, whereas the general improvement affects code that has (apparently) been working OK for some time. (Maybe not that long, I wasn't even aware that you could give names to the IO pins)

Not sure if you have looked it up, but linking in the discussion in #1349 and the initial pr #1352 for the named IO pins.

Now that @andypugh  fixed the bug in 2.9 this could target master instead, and we might be a bit more flexible with the solution and scope.

I don't think it is to hard to have equal amount of pins, where some is named and some isn't, splitting them up is a bit more work, though not by much.

havardAasen:
Rebased against master and added the necesarry documentation, the documentation for the error pins which is located in 2.9 is not included in this commit.

Don't like the global variables, but not sure how to solve it without moving everything into the same function.

andypugh:
Related to this (but not an immediate change) is https://github.com/LinuxCNC/linuxcnc/issues/2532

PR #2796, opened 2023-12-17
Minor modification to Inivar

havardAasen:
Some minor refactoring in inivar, most is related to Clang-Tidy and
Clang warnings.
- Renamed a couple of variables.
- Be more consistent regarding return value, -1 on error and 1 if
  variable is not found, 0 is still success.
- Change to headers adapted for C++.
- Simplified conditions for several of the if-statements.
- Removed two unused headers.
- Use strtol() instead of sscanf() to get number from command-line
  argument. This also sets a limit of accepted numbers between 0 and
  'INT_MAX'.
- Change name for '-num' argument to 'occurrence_number'.

The return value is the only change noticeable from the outside. 

The man page was extended and rewritten in asciidoc.


havardAasen:
Rebased against master to ensure successful CI integration.

PR #2806, opened 2023-12-27
Update Doxygen config file

havardAasen:
~~Update code comments so they can be read by Doxygen and IDEs.~~

Rename and update the old Doxygen config file.

For now I don't think we should do anything with Doxygen, but it has several possibilities that can be looked into

- The first is to create man pages from the source file, this means we could drop many of the `man3` files and have them generated instead.
- It is also possible to create API documentation of libraries, or just header files.


smoe:
Very nice! Many thanks!

petterreinholdtsen:
The comment changes look good to me.  I do not know Doxygen enough to know if there are risks involved in changing its setup like this.  Perhaps the change should be split in two, one modifying the code comments, and another changing the doxygen setup?

havardAasen:
I split up the PR, this PR only handles the Doxygen setup. The updated comments is located in #2827.

If someone used Doxygen to generate any documentation this change could surely affects it output, though seeing how old the config file was and the syntax of most of the source comments, I doubt anyone has used it for years.

Nothing internal in the LinuxCNC project uses it.

For the Doxygen configs, not that much deviates from the default.
- I added some predefined macro definition, so it can find and build documentation for the sections that is normally undefined.
- I enabled `dot` which generates diagrams, diagrams is changed from png to svg. This is mostly done to keep the size small and to increase the quality of the image. If `dot` isn't found it doesn't create the diagrams.
- The previous configuration included every file, I have limited the scope to only include header files.

Issue #2656, opened 2023-09-23
pthread error processing in uspace_rtapi_app not ok 


thgerner:
## Here are the steps I follow to reproduce the issue:

Start a LinuxCNC uspace application, e.g. ```latency-test 1000000```, on a system with a kernel denying setting the pthread priority. In my case kernel version 6.1.12-bone-rt-r15 (armhf) on a BeagleBone Black.
 
## This is what I expected to happen:

Getting an error EPERM "Operation not permitted"

## This is what happened instead:

The Application starts and does not work as the thread did not start.

## Information about my hardware and software:

Linux kernel 6.1.12-bone-rt-r15 armhf
BeagleBone Black

## Reason of the issue

According to the pthread documentation the affected functions return 0 on success or an error code. There is no definition that the error code is less the 0. Indeed EPERM is defined as being 1.

## Fix
```
diff --git a/src/rtapi/uspace_rtapi_app.cc b/src/rtapi/uspace_rtapi_app.cc
index 440727b17c..04fb995fc6 100644
--- a/src/rtapi/uspace_rtapi_app.cc
+++ b/src/rtapi/uspace_rtapi_app.cc
@@ -453,7 +453,7 @@ static pthread_t main_thread{};
 
 static int master(int fd, vector<string> args) {
     main_thread = pthread_self();
-    if(pthread_create(&queue_thread, nullptr, &queue_function, nullptr) < 0) {
+    if(pthread_create(&queue_thread, nullptr, &queue_function, nullptr) != 0) {
         perror("pthread_create (queue function)");
         return -1;
     }
@@ -1028,16 +1028,16 @@ int Posix::task_start(int task_id, unsigned long int period_nsec)
   int nprocs = sysconf( _SC_NPROCESSORS_ONLN );
 
   pthread_attr_t attr;
-  if(pthread_attr_init(&attr) < 0)
-      return -errno;
-  if(pthread_attr_setstacksize(&attr, task->stacksize) < 0)
-      return -errno;
-  if(pthread_attr_setschedpolicy(&attr, policy) < 0)
-      return -errno;
-  if(pthread_attr_setschedparam(&attr, &param) < 0)
-      return -errno;
-  if(pthread_attr_setinheritsched(&attr, PTHREAD_EXPLICIT_SCHED) < 0)
-      return -errno;
+  if(pthread_attr_init(&attr) != 0)
+    return -errno;
[...]

PR #2808, opened 2023-12-27
Check for correct return value after call to 'pthread_*'

havardAasen:
According to man pages, calls to 'pthread_*' can return 'a nonzero value'. This means it can be both negative and positive, we have to check for both.

This commit includes the diff supplied in #2656 and also fixes the same mistake in other files.
Fixes: #2656

### DIFF
--- a/src/rtapi/uspace_rtai.cc
+++ b/src/rtapi/uspace_rtai.cc
@@ -61,19 +61,20 @@ struct RtaiApp : RtapiApp {
         task->pll_correction_limit = 0;
         task->pll_correction = 0;
 
+        int ret;
         pthread_attr_t attr;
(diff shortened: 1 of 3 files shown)


cepelinas9000:
'pthread_*' functions do not set **errno** variable. 

The error handling is incorrect, reading  manuals (for example https://man7.org/linux/man-pages/man3/pthread_setschedparam.3.html) there where no mentioning settings errno. But there is macro which sets manually:
```c
// in https://man7.org/linux/man-pages/man3/pthread_setschedparam.3.html example
#define handle_error_en(en, msg) \
               do { errno = en; perror(msg); exit(EXIT_FAILURE); } while (0)
```

havardAasen:
Right, my bad. I'll update the patch, thanks.

PR #2809, opened 2023-12-27
documentation: Pointer to apt-get build-dep

smoe:
The retrieval of dependencies with dpkg-checkbuilddeps is still tedious that I found in our documentation, which disappointed me a bit, but to our rescue: https://serverfault.com/questions/127625/given-a-debian-source-package-how-do-i-install-the-build-deps/1150501 was equally missing that "apt-get build-deps ." also works for packages not already installed.

### DIFF
--- a/docs/src/code/building-linuxcnc.adoc
+++ b/docs/src/code/building-linuxcnc.adoc
@@ -186,14 +186,19 @@ $ make ../bin/froboz
 
 === Building Debian Packages
 
-When building Debian packages, the LinuxCNC programs are compiled from
-source and then stored in a Debian package, complete with dependency
-information.  This takes more time, and the programs can't be used until
-the Debian package is installed on a target machine.
+When building Debian packages, the LinuxCNC programs are compiled from source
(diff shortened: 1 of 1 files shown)


docs/src/code/building-linuxcnc.adoc
```diff
@@ -289,11 +294,38 @@ If you want just the traditional RTAI "kernel module" realtime, use
...
+
+This will prepare the file debian/control that contains lists of Debian packages
+to create with the runtime dependencies for those packages and for our cause also
+the build-dependencies for those to-be-created packages.
+
+The most straightforward way to get all build-dependencies installed is to just
+execute (from the same directory):
+
+----
+sudo apt-get build-dep .
```
andypugh, line 320: Is it worth stressing that the . is part of the command, not just punctuation?

smoe: The sentence continues below that verbatim block, but, donno, maybe yes. There could be a caption for that verbatim block with a respective partially redundant explanation of what this apt-get command is doing. I'll come up with something.

PR #2811, opened 2023-12-28
gmoccapy: add setting for G-code theme (GtkSourceView)

hansu:
Adds a setting to Gmoccapy for setting the GtkSourceView theme to avoid problems with dark themes.

Related to: https://forum.linuxcnc.org/gmoccapy/50636-can-i-change-the-colors-scheme-in-the-gcode-editor
and https://forum.linuxcnc.org/gmoccapy/51124-gmocappy-defects-in-release-2-9-2

I hope everybody is fine with the name "G-code Theme" because I think a user would not know what is meant with GtkSourceView :)
grafik

### DIFF
--- a/lib/python/gladevcp/hal_sourceview.py
+++ b/lib/python/gladevcp/hal_sourceview.py
@@ -126,6 +126,9 @@ class EMC_SourceView(GtkSource.View, _EMC_ActionBase):
             self.sm.set_search_path(path)
         self.buf.set_style_scheme(self.sm.get_scheme(style))
 
+    def get_style_schemes(self):
+        return self.sm.get_scheme_ids()
+
     def get_filename(self):
         return self.filename
 
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.glade
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.glade
@@ -719,6 +719,17 @@
       </row>
     </data>
   </object>
+  <object class="GtkListStore" id="lstst_sourceview_themes">
+    <columns>
+      <!-- column-name text -->
+      <column type="gchararray"/>
+    </columns>
+    <data>
+      <row>
+        <col id="0">default</col>
+      </row>
+    </data>
+  </object>
   <object class="GtkListStore" id="lstst_themes">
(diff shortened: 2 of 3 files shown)


Sigma1912:
Nice, thanks.
Note that you may also want to adjust the high-lighting for the currently executing line in the gcode window:
https://forum.linuxcnc.org/gmoccapy/50636-can-i-change-the-colors-scheme-in-the-gcode-editor#285585

hansu:
> Nice, thanks.
> 
> You may also want to adjust the high-lighting for the currently executing line in the gcode window: https://forum.linuxcnc.org/gmoccapy/50636-can-i-change-the-colors-scheme-in-the-gcode-editor#285585

That would be a bit too much of settings. I think the best would be to retrieve this color from the GtkSourceView theme, but I couldn't find a solution to get this color from which is defined in `/usr/share/gtksourceview-3.0/styles/<style>.xml `.
Instead I used `style.get_background_color(Gtk.StateFlags.SELECTED)` which is better than a fixed color I think.
Better would be that HAL_Sourceview handles the current line as selection and so automatically uses the same colors which are defined for a selected line in the theme.

Sigma1912:
I didn't mean that it should be user adjustable but that it should be changed to something more readable so what you did is just fine.
I guess you could use 'line number' if you wanted the current line to be different from the 'selected line'

Sigma1912:
> I guess you could use 'line number' if you wanted the current line to be different from the 'selected line'
Ah, for that we would need to access the color scheme in the xml which you have already tried.

hansu:
> Ah, for that we would need to access the color scheme in the xml which you have already tried.

exactly

hansu:
> I didn't mean that it should be user adjustable but that it should be changed to something more readable so what you did is just fine. 

Have you also tested this version?

Sigma1912:
I have just finished testing your last two commits:  
 gmoccapy: add setting for G-code theme (GtkSourceView) 
 gtksourceview: apply selection color from theme 

Works fine for me.
Thanks for your work!

hansu:
Okay then let's use this for now

PR #2814, opened 2023-12-30
Gmoccapy, don't hide recurring error messages

Sigma1912:
The original code will suppress showing an error message if it is identical to the one shown last. While this seems to make some sense at first glance it leads to confusing situations where the user thinks an error has been fixed while it actually still exists but is hidden by the gui.

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -2410,11 +2410,8 @@ class gmoccapy(object):
                 self._on_play_sound(None, "alert")
 
     def on_gremlin_gcode_error(self, widget, errortext):
-        if self.gcodeerror == errortext:
-            return
-        else:
-            self.gcodeerror = errortext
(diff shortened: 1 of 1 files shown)


gmoccapy:
I think I have done that that way, because I had an error with pop up every poll, so I had 100#erts of pop up windows.

Have you tested well the possible situations?

Norbert

Sigma1912:
> I think I have done that that way, because I had an error with pop up every poll, so I had 100#erts of pop up windows.

I suspected that may have been the reason. I have not had that happen yet, can you remember in what case you got that behavior?

hansu:
We could also change the type of warning dialog.

For example when I open a file with an error, I get this dialog:
grafik
But this doesn't prevent the file from loading, so a dialog like this could also be used (this info box appears when I try to start the program):
grafik
For this messages we already have the possibility to limit the number of displayed messages.

andypugh:
Assigning this to Norbert, I feel this should be his final decision.

PR #2816, opened 2023-12-30
Fix typos in src subdirectory

luzpaz:
Found via `codespell -q 3 -S *.po,*.pot,*.svg,*.ts,./.git/logs,./share,./docs/man/es,./configs/attic,*_fr.*,*_es.*,README_es,./src/hal/classicladder -L ans,ba,breal,bu,bulle,busses,childs,componentes,currenty,doubleclick,dout,dum,etxt,extint,faktor,fo,fpr,halp,ihs,inout,inport,mot,parm,parms,propertys,ro,seh,ser,smoe,te,tey,trough,ue,ure,wille,wont,wonte`


src/hal/drivers/hal_gm.c
```diff
@@ -1840,7 +1840,7 @@ RS485(void *arg, long period)
 	  {
 	    if((device-> RS485_mgr.ID[2*i]) != ((temp_u32 >> 8) & 0xff)) 
             {
-              //RS485 module falled off, error
+              //RS485 module failed off, error
```
hansu, line 1843: probably "fell off" is the correct version

src/hal/drivers/hal_gm.c
```diff
@@ -1854,7 +1854,7 @@ RS485(void *arg, long period)
 	  {
 	    if((device-> RS485_mgr.ID[2*i+1]) != ((temp_u32 & 0xff000000)>>24))
             {
-              //RS485 module falled off, error
+              //RS485 module failed off, error
```
hansu, line 1857: probably "fell off" is the correct version

luzpaz:
JFYI, there is another branch with misc. fixes at https://github.com/LinuxCNC/linuxcnc/compare/master...luzpaz:linuxcnc:typos-misc

Should I send a PR for that or further whittle it down?

src/hal/drivers/hal_gm.c
hansu: Or maybe just `//RS485 module failed, error` ?
"Fall off" does not make much sense in this context IMHO.

PR #2822, opened 2024-01-01
Documentation: Typo, punctuation, transition from 2.9.1 to 2.9.2

smoe:


docs/src/getting-started/updating-linuxcnc.adoc
```diff
@@ -20,7 +20,7 @@ updates. If you don't have an internet connection to your PC see
 == Upgrade to the new version
 
 This section describes how to upgrade LinuxCNC from version 2.8 to the
-new 2.9.1 version.  It assumes that you have an existing 2.8 install that you
+new 2.9.2 version.  It assumes that you have an existing 2.8 install that you
```
andypugh, line 23: Maybe this should be 2.9.x ?

smoe: Done, made it 2.9.y to avoid the impression that 2.8.x and 2.9.x would be the same x. I know ... just some gut feeling that we have a special audience.

andypugh:
I just noticed that it went into master, whereas is should probably have gone into 2.9?

smoe:
@andypugh , ouch, yes, sorry. What should I do? Prepare a second PR to 2.9?

andypugh:
I can just cherry-pick it

PR #2823, opened 2024-01-01
wj200: add actual speed feedback and RPM scaling

djdelorie:
Three new pins:

* frequency_scale (in, default 1): allows you to scale the commanded and actual frequency pins to read in actual RPM instead of HZ.  Example: a four pole motor would set this to 60 to change the "frequency" range of 0..400 to an "RPM" range of 0..24000.  The default is 1 for backwards compatibility.

* motor_frequency (out): Provides actual feedback of the motor's rotation as determined by back-EMF (i.e. not the "stored commanded" speed).

* is_stopped (out): True when the VFD has determined that the motor has come to a halt.

Note: due to latency issues, you cannot use "is_stopped" to tell the realtime system when the motor has reached a commanded speed of zero, because it won't "go false" fast enough.

Tested live on my Avid/S30C cnc router.

My config:

loadusr -W /usr/bin/wj200_vfd --device /dev/modbus0 --baud 38400 setp wj200-vfd.0.mbslaveaddr 1
setp wj200-vfd.0.frequency-scale 60

net spindle-reverse wj200-vfd.0.reverse <= spindle.0.reverse net spindle-run wj200-vfd.0.run <= spindle.0.on
net machine-is-enabled wj200-vfd.0.enable

loadrt near
addf near.0 servo-thread
setp near.0.scale 1.01

net vfd-commanded-speed near.0.in1 wj200-vfd.0.commanded-frequency <= spindle.0.speed-out-abs net vfd-motor-speed near.0.in2 <= wj200-vfd.0.motor-frequency net vfd-speed-near and2.1.in0 <= near.0.out
net vfd-at-speed and2.1.in1 <= wj200-vfd.0.is-at-speed net spindle-at-speed spindle.0.at-speed <= and2.1.out


petterreinholdtsen:
Are the pin names and functionallity also used by other HAL components?  In other words, are you trying to be consistent with other components or make up new names?

djdelorie:
is_stopped - other VFDs use is-stopped but all the pins in wj200_bfd use underscores

motor_frequency - other VFDs use frequency-out or OutF (but the frequency input pins are equally inconsistent).  The WJ200 has two outputs; "Output frequency monitor" and "Actual frequency monitor".  I'm reading the second, which is more like vfdb's motor-RPM pin, except it defaults to Hz scale (and motor-RPM is a vfd register, so the vfd is scaling it, not us).  In retrospect, I suspect this pin is signed, so it's more like motor velocity than speed.  I've never run my spindle in reverse prior to a tool change op.

So it's a mix ;-)

I suspect more work is needed in both the vfd code and the near module to handle fwd-rev changes (i.e. if you go from 3000 RPM fwd to 3000 RPM rev, for a moment, near say's it's OK because you're comparing speeds (unsigned), not velocities (signed).

djdelorie:
I should read my own code ;-)  (It's been a while since I wrote this).  "actual speed" is encoder-based, which I don't have, so I'm returning "output frequency monitor" which seems to track the motor's actual speed but is absolute speed not signed velocity.

And yes, the VFD's "at speed" pin blinks at 3000 RPM when you go from "m3 S5000" to "m4 S3000" so it's not considering direction either.

djdelorie:
New version that fixes the at-speed issue, and adds a motor_reverse pin.

andypugh:
> is_stopped - other VFDs use is-stopped but all the pins in wj200_bfd use underscores

There is a document somewhere that says that hyphens are preferred in HAL pin names. 
But that does mean that the loart lines (which use _ as it is a filename) end up inconsistent. 
loadrt my_component will make hal pins my-component.1, my-component.2.
https://linuxcnc.org/docs/stable/html/hal/canonical-devices.html

djdelorie:
I'm willing to break everyone's HAL config if you want me to change them all to hyphens ;-)

andypugh:
> I'm willing to break everyone's HAL config if you want me to change them all to hyphens ;-)

It's probably not worth doing, as those are rules mainly obvious in the breaking. But there is a solution to the "breaking configs" thing, if you also up-version the INI file:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/ini/update_ini.py#L425

PR #2827, opened 2024-01-03
Update code comments to be parsed by IDEs and Doxygen

havardAasen:
All functions was already well documented but it has been updated to be parsed by IDEs and Doxygen.

One line including a header file was moved up, together with the other inclusion.


andypugh:
Looking through this, I notice that much of the information is really outdated. Especially "Version 2.0 supports RTLinux and RTAI". Really it should be saying "preempt-rt, RTAI and Xenomai (under LXRT)"

PR #2829, opened 2024-01-03
Updates for the IniFile class

### LINKED ISSUE #2812: [RFC] Change return value from `const char *` to `std::optional<const char*>`
This is a request for comments, or perhaps ask for approval, to change the return value for public functions in the library `liblinuxcncini`, specifically file `src/libnml/inifile/inifile.hh`. This change will break the API of the library, I will of course update the code within the project, but I'm unsure if, or how many, users who also use this library.

The proposed change is to return `std::optional<const char*>` instead of `const char *`.

This has several benefits, it tells the developer that the function *can* return something else than a `const char *`. The `std::optional<>`can be used as a boolean, if it has a value or not, which can simplify error handling. It is also increases type safety in regards to NULL pointer dereferencing.

havardAasen:
The most significant change is the return value for some of the `Find()`. The return value was changed from `const char*` to `std::optional<const char*>`. This change breaks the API.

The main reason behind this change is to tell the users that the value can return something else, the compiler can also give us some help if we use the returned value in the wrong way, increasing type-safety.

Some of the code has also been refactored, changed and optimized, this is unrelated to the previously mentioned change.

I also added tests. It's does not have complete coverage, the C API is untested (the 'mb2hal' tests, uses the C API) as well as the 'tildeexpand' functionality.

Closes #2812


andypugh:
Can you fix the conflicts? It doesn't look hard, and can be done in the web editor here on github, but I am not particularly confident meddling with other folks' C++.

havardAasen:
Sure, no problem Andy.

Rebased it against master, perhaps not the best thing considering how large the diff became, still, the needed change was minor.

PR #2834, opened 2024-01-05
Docs typos

smoe:
As usual, something spotted while translating.

Please kindly confirm my interpretation "neither between" in docs/src/drivers/hal_gpio.adoc line 57.
Admittedly this PR was mostly motivated as a technical excercise to work with WSL.

### DIFF
--- a/docs/src/drivers/hal_gpio.adoc
+++ b/docs/src/drivers/hal_gpio.adoc
@@ -5,7 +5,7 @@
 
 = Generic driver for any GPIO supported by gpiod.
 
-This driver has been tested on the Raspberry Pi, and should also work on Bananapi, Beaglebone, Pine64 (et al) and other single board computers, and potentially on other platforms.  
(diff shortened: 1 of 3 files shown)


petterreinholdtsen:
I assume you are talking about this sentence: The pin names are case-sensitive and there must be no spaces in the strings, neither between the comma-separated pins lists of the "=" signs.

I would write "no spaces in the string, nor between the comma-separated pins" instead myself.  No native speaker, so not sure if it truly better.

hansu:
Both is fine I guess.
We can also write:

(3) 
`The pin names are case-sensitive and there must be no spaces in the strings, not even between the comma-separated pins lists of the "=" signs.`

or

(4)
`The pin names are case-sensitive and there must be neither spaces in the strings, nor between the comma-separated pins lists of the "=" signs.`

But I would prefer (1) (Steffens version) or (2) (Petters version).

smoe:
Heya, please then ("either" of you, pun intended) merge my request since this saves time, is apparently good enough, and closest to the original.

 




But I would prefer (1) (Steffens version) or (2) (Petters version).

hansu:
I agree :)

PR #2837, opened 2024-01-06
Gmoccapy: add configuration for 800x600

hansu:
I created a configuration using the a user command file to allow Gmoccapy to run on a 800x600 screen.

Related to https://forum.linuxcnc.org/gmoccapy/30793-screen-resolution

For this to work in fullscreen as well, there is a little modification on the gmoccapy base necessary.

Basically I
- reduced the button sizes
- re-organized the settings page
- moved the info box to the user tab

Now a window size of 770 x 580 is possible, BUT only in the English version. In other languages e.g. German, the size will be bigger because the translated text on the buttons might be longer.
This is a draft because I may reorganize the Gmoccapy setting pages and then I might need to adjust it here as well.

But please test this and let me know if it works for you.

gmoccapy_800x600-1
gmoccapy_800x600-info_box
gmoccapy_800x600_pref-1
gmoccapy_800x600_pref-4


zz912:
I dont need it, but I tested it.
800x600-monitor-CZ
As you wrote in a foreign language, it doesn't work:
800x600
So I switched it to EN:
800x600-EN
1. The Perspective button is displayed incorrectly. Check out your pictures above. You won't see the button at all.
2. When I start LCNC, Preview does not switch to perspective.
3. The right edges of the buttons are ugly.

zz912:
I dont need it, but I tested it.
800x600-monitor-CZ
As you wrote in a foreign language, it doesn't work: (I think it is not good to make Gmoccapy configuration only for EN languauge.)
800x600
So I switched it to EN:
800x600-EN
1. The Perspective button is displayed incorrectly. Check out your pictures above. You won't see the button at all.
2. When I start LCNC, Preview does not switch to perspective.
3. The right edges of the buttons are ugly.

hansu:
Thanks for testing!

> The Perspective button is displayed incorrectly. Check out your pictures above. You won't see the button at all.

Good point, I missed that.

> When I start LCNC, Preview does not switch to perspective.

Is this the case usually? I will check that.

>The right edges of the buttons are ugly.

That's GTK - we have to live with that if we still want to user a GtkButtonBox with reduced margin. Otherwise I have to rebuild all with single buttons in containers ... not for the few users who might need this resolution. And you don't know how it would look like in GTK4...

hansu:
> 
>     1. The Perspective button is displayed incorrectly. Check out your pictures above. You won't see the button at all.
> 
Fixed.

>     2. When I start LCNC, Preview does not switch to perspective.
> 
Gmoccapy remembers the last view in the .pref file, so everything is fine here.



I only have to figure out which part of the translation disallows the reduced window size.

zz912:
Results of my test:
Gmoccapy_czech
Gmoccapy_czech-home_page
Down part is missing.

hansu:
Ah there are multi-line strings in the czech translation. Ok I will reserve more space for the settings page.
Thanks for testing!

zz912:
Or could it be a different button layout?
buttons

hansu:
> Or could it be a different button layout? buttons

Sure. Suggestions? 
But that is not the problem that the lower part is cut if you had that in mind.

hansu:
Should be fine now

zz912:
I tried several languages. I renamed forign folder to my language and ran it.
/home/zdenek/linuxcnc/linuxcnc-gmoccapy-800x600/share/locale/

Size is OK:
french
But still Perspektive button is touched and Z view is viewed.

hansu:
> But still Perspektive button is touched and Z view is viewed.

But that has nothing to do with the 800x600 config. You discovered here a general bug in this version :-)

hansu:
> > But still Perspektive button is touched and Z view is viewed.
> 
> But that has nothing to do with the 800x600 config. You discovered here a general bug in this version :-)

Fixed now

zz912:
> > > But still Perspektive button is touched and Z view is viewed.
> > 
> > 
> > But that has nothing to do with the 800x600 config. You discovered here a general bug in this version :-)
> 
> Fixed now

I tested it, and it works. 




I tried select MDI, then custom and screen looks wierd. Is it OK?
mdi-custom

hansu:
Weird thing with that notebook widget, thanks for spotting this!
Fixed now.

PR #2848, opened 2024-01-16
docs: Fix bugs URL in man/man1/*.1 (smoe:Fix_Bugs_URL)

smoe:
Three prominent man pages had references to the sf.net bugs page, still.


smoe:
@petterreinholdtsen , @hansu - this should be a quick no-brainer.

PR #2863, opened 2024-01-27
limit_axis - Don't check range if not enabled

satiowadahc:
Enable check inside of an error check means if its disabled it'll only disable when there is an error

### DIFF
--- a/src/hal/components/limit_axis.comp
+++ b/src/hal/components/limit_axis.comp
@@ -65,11 +65,11 @@ FUNCTION(_){
 
   // Check all ranges
   for (i = 0; i < personality; i++){
-    if (max_range(i) < min_range(i)){
-      if(! enable(i)){
-        continue;
-      }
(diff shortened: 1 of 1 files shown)


satiowadahc:
This is the first I've used this since it was converted to a .comp file.

petterreinholdtsen:
Do you have a test setup for tests/ to demonstrate the bug and ensure it is fixed?

andypugh:
I think that this looks deliberate. Is the intention to mask the error before (for example) homing?

andypugh:
Maybe I am missing something, but this doesn't appear to change any behaviour. It might save a little CPU, I suppose.

satiowadahc:
If the range is disabled and there is not an error, the range currently will not disable.

Max<min is a sanity check that disables the range if it is mistyped. 

Enable allows ranges to be turned on/off

PR #2870, opened 2024-01-29
Hostmot2 encoder quad error clear when disabled fix.

JTrantow:
This change will reset the quad error when quadrature_error_enable is false. The previous commit fixed the problem for simple halrun scripts, but failed to fix the problem with complete configs running in linuxcnc.


JTrantow:
Made a mistake combining & operations.

pcw-mesa:
I cannot approve this change as it increases network load significantly (about 10% for a single encoder) ,
and I have been unable to duplicate the issue in current 2.9 or 2.10 on two different sets of hardware
(7I96S encoder input and 7I92+7I84S)

pcw-mesa:
I have finally  been able to duplicate this, it only happens with more than 1 encoder is enabled
 so I will look at the code when I get a chance.

JTrantow:
Thanks. I'd be happy to assist in testing as I have two computers (7i92,
7i96) and three configs that all exhibit this problem.

PR #2876, opened 2024-02-03
Reorder member variables in 'block_struct'

havardAasen:
The variables were arranged in alphabetical order, resulting in more than 10% of the total struct size being occupied by padding. However, by reorganizing the variables, the size was reduced from 1088 bytes to 960 bytes.

All variables is default initialized in the struct.

### DIFF
--- a/src/emc/rs274ngc/interp_internal.hh
+++ b/src/emc/rs274ngc/interp_internal.hh
@@ -426,83 +426,82 @@ typedef int_remap_map::iterator int_remap_iterator;
 
 struct block_struct
 {
-  block_struct ();
+  char comment[256]{};
+  double a_number{};
+  double b_number{};
+  double c_number{};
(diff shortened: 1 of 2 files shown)


andypugh:
Given that there may be many, many blocks, this might be worthwhile. Though at that point perhaps it is worth looking at the size of the comment field?

havardAasen:
> Given that there may be many, many blocks, this might be worthwhile. Though at that point perhaps it is worth looking at the size of the comment field?

I looked up RS274NGC version 3 and this specifies 256 characters, so if we decrease our field, we would deviate from the standard. Not sure how important it is to be consistent, I believe that 256 characters for a comment is on the larger side.

andypugh:
I don't recall if G-code supports multi-line comments. It might, using the braces version.
In that case 256 is possible.

Issue #2844, opened 2024-01-11
G70,G71,G72 looping not drawing


powerio25:
# The issue tracker is not a support forum

The LinuxCNC issue tracker is to report bugs in the software.
If you have a question about how to use the software, use one of the other methods detailed on our community support page: http://linuxcnc.org/community/

(delete this section before submitting your bug report)

## Here are the steps I follow to reproduce the issue:

 1.
[g71.txt](https://github.com/LinuxCNC/linuxcnc/files/13907002/g71.txt)

 2.
 3.

## This is what I expected to happen:
First the patch , this code in attachment run good.
## This is what happened instead:

in src/emc/rs274ngc/interp_g7x.cc
try {
	switch(cycle) {
	case 70: path.do_g70(&motion,x,z,d,e,p); break;
	-case 71: path.do_g71(&motion,subcycle,x,z,u,w,d,i,r); break;
	-case 72: path.do_g72(&motion,subcycle,x,z,u,w,d,i,r); break;
	case 71:
	    if(x!=imag(start)) {
		std::complex<double> end{real(start),x};
		path.emplace_back(std::make_unique<straight_segment>(
		    start, end
		));
	    }
	    path.do_g71(&motion,subcycle,x,z,u,w,d,i,r);
	    break;
	case 72:
	    if(z!=real(start)) {
		std::complex<double> end{z,imag(start)};
		path.emplace_back(std::make_unique<straight_segment>(
		    start, end
		));
	    }
	    path.do_g72(&motion,subcycle,x,z,u,w,d,i,r);
	    break;
	}
    } catch(std::string &s) {
	ERS("G7X error: %s", s.c_str());

## It worked properly before this:
(If the behavior changed after making a particular change in hardware or
software, describe the change you think is responsible.  E.g., "after upgrading
from LinuxCNC 2.7.3 to 2.7.4")

## Information about my hardware and software:

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`):
 * I am using this kernel version (shown by `uname -a`):
 * I am running ...
   * [ ] A binary version from linuxcnc.org (including buildbot.linuxcnc.org)
   * [ ] A binary I built myself
   * [ ] A binary version from some other source besides linuxcnc.org
 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`):
 * I am using this user interface (GUI) (e.g., AXIS, Touchy, gmoccapy, etc):
[...]

PR #2881, opened 2024-02-07
Add another G71 endless loop test and a fix for it

havardAasen:
A stripped down version of the problem in #2844, code structure is copied from 'g71-endless-loop'. The test is marked as skipped, since it does not always terminates correctly.


petterreinholdtsen:
@andypugh Should this change go into 2.9 or master?

andypugh:
I am happy for extra tests (which pass!) to go in 2.9. Especialliy if they relate to bug fixes which are also aimed at 2.9

petterreinholdtsen:
[andypugh]
> I am happy for extra tests (which pass!) to go in 2.9. Especialliy if
> they relate to bug fixes which are also aimed at 2.9

How does your view relate to this patch?  A skipped test, that is.
-- 
Happy hacking
Petter Reinholdtsen

havardAasen:
I updated the branch and enabled the test. I also included a fix for the endless loop problem. I believe this fix is robust enough to handle all cases where the algorithm ended up in an endless loop.

Still targeting master, but there is no problem targeting 2.9, except that it is untested.

fsabbatini89:
Test this and endless loop will happen: k-20 is ill formed. 
Already fixed in [#2907](https://github.com/LinuxCNC/linuxcnc/pull/2907)


```
G54 g95 g18 G21 G7 g61 g91.1

t02
g97 s350 m4
g0 x120 z4 m8


o110 sub 
g1 X79.995 Z0
g3 x83.9549 z-1.7176 k-20
g1 x95.25 z-47.4
g1 z-62
g1 x99 z-68
O110 endsub
;o15 call

G71 X104 Z5 Q110 D0 I1.5 R2 F0.3
g0 x200 z100
m30
```

Issue #2842, opened 2024-01-09
API for libgpiod has changed in v2


jasonKercher:
## Here are the steps I follow to reproduce the issue:

 1. Have libgpiod version >= 2
 2. Attempt to compile linuxcnc v 2.9.2

## This is what I expected to happen:

Build succeeds

## This is what happened instead:

```none
In file included from hal/drivers/hal_gpio.c:39:
hal/drivers/hal_gpio.c:188:50: error: invalid application of ‘sizeof’ to incomplete type ‘struct gpiod_line_bulk’
  188 |             (*ptr)[c].bulk = rtapi_kmalloc(sizeof(*(*ptr)[c].bulk), RTAPI_GFP_KERNEL);
```

...along with lots of warnings for implicit declarations of functions (due to API change).

## It worked properly before this:

libgpiod version <= 1.6.4

## Information about my hardware and software:

 * I am using Manjaro.  I know that is not officially supported, but we do keep a working [linuxcnc package](https://aur.archlinux.org/packages/linuxcnc) over there.
 Manjaro is part of the Arch Linux family, so we always have the bleeding edge latest software.  I imagine this will not be an issue on Debian for some time.
 * Linux Kernel Version 6.1.66
 * Linux CNC v2.9.2

andypugh:
This was something that I knew was coming, and that I don't really know how to solve. 

libgpiod is a mess, for example libgpiod2 is still v1.6, not 2. 

There are no packages in Debian  for libgpiod v2, with the new API, so I decided this was a problem that can be kicked down the road until at least after Trixie (which still has v1.6) https://packages.debian.org/search?suite=trixie&searchon=names&keywords=libgpiod

There is probably something horribly messy that can be done with conditional compilation, or a wrapper API. 

I am not even 100% sure that libgpiod has a 1:1 mapping between v1.x and v2.x functions. 

The libgpiod developers were keen that I should code for v2, but with no packages available from Debian, how would our users get it? We can host the packages on our own repository, but that isn't a fix for folk installing LinuxCNC from the main Debian mirrors.

FabioLolix:
Hello, on the AUR making a pkgbuild for libgpiod v1.6.4 installed in /opt/libgpiod-1.6.4 (or something similar) so it doesn't conflicts with the system's one isn't a problem

How I could pass that to autogen.sh or ./configure ?

FabioLolix:
[...]

PR #2883, opened 2024-02-09
Restrict libgpiod to < 2.0.0 when configuring

havardAasen:
This restricts libgpiod to versions < 2.0.0. mainly to prevent build failures if the system has any later version installed, which LinuxCNC does not support.

This prevents build failure noticed in #2842 but does of course not do anything to support the new API.

Using pkg-config to do the heavy lifting.

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -3,6 +3,7 @@
 /.tmp_versions
 /Makefile.inc
 /Makefile.modinc
+/aclocal.m4
 /config.guess
 /config.h
 /config.log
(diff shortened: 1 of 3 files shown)


andypugh:
Supporting the new API is on my list of things to do. And fairly high up.

havardAasen:
> Supporting the new API is on my list of things to do. And fairly high up.

Nice, you want to handle the build system as well?

This PR fails because it's missing `aclocal.m4`, could update `src/autogen.sh` to run `aclocal` before `autoconf`.

I haven't looked into this file before, it seems we handcraft a lot of things. Could we swap out everything in `src/autogen.sh` with `autoreconf --force --install`?

petterreinholdtsen:
[Håvard]
> I haven't looked into this file before, it seems we handcraft a lot of
> things. Could we swap out everything in `src/autogen.sh` with
> `autoreconf --force --install`?

Sound like a good idea to me.

-- 
Happy hacking
Petter Reinholdtsen

PR #2885, opened 2024-02-12
interp_g7x: Improve readability regarding `do_g70()`

havardAasen:
These patches simplifies and improves readability for function `do_g70()`, it also adds one of the example as a test.


andypugh:
I can't help feeling that it would be more useful to improve the readability (and detail) in the G7x _docs_ rather than re-working the source code.....

havardAasen:
I believe this is to separate issues, the docs, at least as it stands now, is more for our users. This pr is about code readability, and that is for developers. Could of course have documentation for both...

PR #2887, opened 2024-02-13
GCode Graphics - Show Bounding box of part when not showing preview.

satiowadahc:
Some files >1000000 lines or so, take a lot of resources to preview. Replacing the preview with a bounding box heavily reduces this issue.

While trying to find a good solution, IDE's are screaming at errors in this file, which I tried only addressing severe ones, such as reusing variables, while loop that was removed, unused imports, using builtin names, removed unused variables.

I tested with axis, qtdragon, our own gui. Foam and Mill versions.

### DIFF
--- a/lib/python/rs274/glcanon.py
+++ b/lib/python/rs274/glcanon.py
@@ -15,19 +15,16 @@
 #    along with this program; if not, write to the Free Software
 #    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
-from rs274 import Translated, ArcsToSegmentsMixin, OpenGLTk
+from rs274 import Translated, ArcsToSegmentsMixin
 from OpenGL.GL import *
 from OpenGL.GLU import *
-import itertools
 import math
-import glnav
 import hershey
 import linuxcnc
 import array
 import gcode
 import os
 import re
-import sys
 from functools import reduce
 
 def minmax(*args):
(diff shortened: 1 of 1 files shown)


lib/python/rs274/glcanon.py
```diff
@@ -272,7 +287,7 @@ def tool_offset(self, xo, yo, zo, ao, bo, co, uo, vo, wo):
         self.xo = xo
         self.yo = yo
         self.zo = zo
-        self.so = ao
+        self.ao = ao
```
satiowadahc, line 290: 99% Sure this was a typo that has always been an issue. But what is an "a" tool offset?

Sigma1912: Yes, that seems to be a typo. 
We can set tool offsets for all nine axis words (XYZABCUVW) in the tool table. I haven't used anything other then XYZ myself but maybe for a robot it's useful.

satiowadahc: On my gui, we go by either file size, or comment of `(DISABLE PREVIEW)`

```python
        if self.state.magic_comments["disable_preview"]:
            LOG.info("Preview Disabled by comments")
            self.ui.gcodegraphics.show_program = False
            return
        elif os.stat(file_name).st_size < self.max_graphics_file_size:
            self.ui.gcodegraphics.show_program = True
        else:
            self.ui.gcodegraphics.show_program = False
            LOG.warning(f"File To Large ({os.stat(file_name).st_size/1000000} mb), Preview disabled")

```

Sigma1912: Nice, looks quite useful.

satiowadahc: The alternative I would consider, but it would need adoption from axis, qtvcp and plasma is to make it a standalone toggle on/off.

c-morley:
Could you rebase this on master? I think Andy would agree that the changes are too big for 2.9

c-morley:
Great idea by the way.

satiowadahc:
Thanks! I could rebate... would you consider if I skip the code cleanup and just add the bounding box to 2.9?

c-morley:
I can't speak for @andypugh  - it's his decision, but I would think it would be acceptable as the changes are small and managable to debug.
I hope you would also make a pull request for the master for the rest of the work you did.

andypugh:
2.9 should only be getting bug-fixes at this point. I think that anything like this belongs in Master.

satiowadahc:
Closing in favour of #2897 #2896

PR #2893, opened 2024-02-18
Encapsulate static arrays, rename function and remove header files

havardAasen:
Move some of the `static const` arrays to the private section of the class, as the comments suggests. I also renamed the arrays, dropping the leading underscore.

Function `isreadonly()` is renamed to `is_parameter_readonly()`, made static and appended `nodiscard`.

For the removed header files, I would guess that the file was copied with the included headers untouched.


havardAasen:
Regarding those arrays, another approach is to change `interp_array.cc` to a header file, and introduce a namespace like `interpreter` or `interpreter::array`. Since these are all constants it's not that important if they are globally accessible in a namespace. It would also mean that we could change the arrays to `static constexpr` in a more elegant way.

PR #2896, opened 2024-02-19
Add Part Bounding box on Gcode Graphics

satiowadahc:
Replacing #2887, now rebased on master.


satiowadahc:
Set a default max file size, check with INI for updated value, then add getter/setters?

c-morley:
I'd think just a function to set it would be fine. A GUI then could set it if it's required, I suspect it won't happen often.
What do you think?

andypugh:
I have assigned this to @c-morley to decide whether to merge or not. I don't know enough to decide.

c-morley:
Thank you.

PR #2897, opened 2024-02-19
Correct Typo and Remove redundant loop in gcode graphics

satiowadahc:


andypugh:
The so - ao looks like a typo, did you work out what the affect was? 

The "While 1:" is a fairly common way to allow a block of code to be turned on or off during development. It probably shouldnt have made it to production code. (I did wonder if it was a way to limit the scope of a variable, but apparently that doesn't work in Python)

satiowadahc:
The while 1 was refactored a year ago. It used to be:
```python
while 1:
   try:
       makebuffer
   except
       makebufferbigger
       continue
   break
```

weird this caused stack overflows and was refactored to
```
while 1:
   try
       makebuffer
   except
       meh.
   break
```

satiowadahc:
https://github.com/linuxcnc/linuxcnc/commit/67a1e6c796b65a6455135baee310834511fa8a1b

PR #2899, opened 2024-02-20
Update python-interface.adoc

jethornton:
Note that the axes stat has been removed and show a way to get the same information from axis_mask


jethornton:
This should go in 2.9 but somehow it's pointing to master.

PR #2901, opened 2024-02-20
Doc Fix

jethornton:
Note that the axes stat has been removed and show a way to get the same information from axis_mask


petterreinholdtsen:
> Note that the axes stat has been removed and show a way to get the
> same information from axis_mask

Look good to me.  Perhaps a better phrase than "Removed from version
2.9" would be "Removed since version 2.9."?

-- 
Happy hacking
Petter Reinholdtsen

jethornton:
At what point did you remove it?

Issue #2414, opened 2023-03-30
Jog velocity with keyboard not continuous


dberndt391:
## Here are the steps I follow to reproduce the issue:

 1. Start linuxcnc 2.9pre-1 sim, power on.
 2. halscope joint.x.vel-cmd
 3. jog with arrow keys

## This is what I expected to happen:
Expect a constant jog velocity

## This is what happened instead:
Velocity drops sometimes. Velocity drop is roughly centered around the key repeat occurrence.

## It worked properly before this:
It worked on 2.8 on a entirely different computer.  I have no baseline with this hardware. It seems 2.8 can't be gone back to without major python2 related work.

## Information about my hardware and software:

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`): Linux Mint 21
 * I am using this kernel version (shown by `uname -a`): 6.1.0-5-rt-amd64
 * I am running ...
   * [x] A binary I built myself
 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`): 2.9~pre1
 * I am using this user interface (GUI) (e.g., AXIS, Touchy, gmoccapy, etc): Axis
 * I am using this interface hardware vendor and chipset (e.g., parallel port, ethernet port, FPGA card):  Simulation

## More details:

Issue does not happen on every key-repeat, but happens often. See attached image.
![lcnc-jog](https://user-images.githubusercontent.com/129317567/228937367-eb0bd5f4-873c-4888-8f7e-134ebc56ca17.png)

I know this looks similar to #434 However I'm fairly sure it's not a hardware/usb/keyboard issue. xev shows the key press up and down signals happening on the same ms every time. 

I instrumented the jog_on and jog_off functions in axis with some debug statements. Here is what pressing the jog keys with the machine off looks like:

*note times are deltas from the last jog debug event, and apparently they're *-1 as well as I wrote bad debug math apparently. 
```c
---f: jog_on start,                                 -4.98560
---f: jog_on stop-not manual_ok,                    -0.00004
^^^f: jog_off start,                                -0.52913
^^^f: jog_off stop-jog_after[a],                    -0.00003
---f: jog_on start,                                 -0.00115
---f: jog_on stop-not manual_ok,                    -0.00003
^^^f: jog_off start,                                -1.99852
^^^f: jog_off stop-jog_after[a],                    -0.00003
---f: jog_on start,                                 -0.00025
[...]

PR #2903, opened 2024-02-21
Update axis.py

grotius-cnc:
solves : 

https://github.com/LinuxCNC/linuxcnc/issues/2414


BsAtHome:
This patch does not always work to restore key repeat and fixes #2414 with severe side-effects. It fails if you exit <i>without</i> going through the confirm dialog. Even then, I've had it fail to restore key-repeat many times.

Besides being a nightmare in a development setting, it is fundamentally wrong, as an application, to change the global X-server state. Changing these operations are reserved for settings-applications in a desktop manager's control panel.

The proper way to do this is to use the equivalent of the X function <code>XkbSetDetectableAutoRepeat()</code>, which eliminates all key-release events for key-repeats. Then you keep track of the state of press/release in those functions that do not want any repeats.

PR #2904, opened 2024-02-22
Remove erroneous quote.

newellshk:
I noticed this because it messed up the syntax highlighting in my editor.
I have close to zero experience with tcl so I hope it's correct...

### DIFF
--- a/lib/hallib/check_config.tcl
+++ b/lib/hallib/check_config.tcl
@@ -165,7 +165,7 @@ proc warn_for_multiple_ini_values {} {
   foreach section $sections {
     set enforce 0
     foreach csection $sections_to_check {
-      if {[string first $csection $section"] >= 0} {
+      if {[string first $csection $section] >= 0} {
         set enforce 1
         break
       }


andypugh:
This does look wrong, but I don't know enough about Tcl to know if it really is wrong.

hansu:
Syntactically this is correct even though it seems not to make much sense. It just extends the string in which will be searched by a quotation mark.
But it doesn't hurt because the string the string `<csection>` will be found in `<section>` as well as in `<section>`+`"`.

andypugh:
I still don't know if rmu (on the IRC) or yourself are advocating for this change or against it. 

I understand that it will break syntax highlighting in many editors.

Perhaps the quote (if useful) could be escaped or quoted in some way?

hansu:
I think this is quite safe to merge. It searches in the INI section names and they don't contain a quote.
Furthermore I double-checked with the background knowledge of the corresponding change: https://github.com/LinuxCNC/linuxcnc/commit/6a3d0a434ebed19c41aecb3e6d09c712e1971b52

PR #2907, opened 2024-02-23
Update interp_g7x.cc fix endless loop issue (in axis and in milltask)

fsabbatini89:
Fix issue where G7x is ussed with g3 and g2 inside, and circle has errors inside it, endless loop would occur eating all ram memory and freezing the machine

uspace_xenomai.cc:
Fix dependendencies on POSIX xenomai skin

Also there is another fix to implement but i will not pull request it as it will modify behaviour but is as follows:
When doing the g72 movement reset the control and start over, as G7x code modifies incremental i/k coordinates to absolute coordinates and they will not be back to default incremental, all countour profile will be treated as i/k absolute.
The solution is put in the beginning of the file a g91.1

you can test the following code to see the fix at work:

```
G54 g95 g18 G21 G7 g61 g91.1

t02
g97 s350 m4
g0 x120 z4 m8


o110 sub 
g1 X79.995 Z0
g3 x83.9549 z-1.7176 k-2
g1 x95.25 z-47.4
g1 z-62
g1 x99 z-68
O110 endsub
;o15 call

G71 X104 Z5 Q110 D0 I1.5 R2 F0.3
g0 x200 z100
m30
```

Regards
Franco

### DIFF
--- a/src/emc/rs274ngc/interp_g7x.cc
+++ b/src/emc/rs274ngc/interp_g7x.cc
@@ -1112,6 +1112,31 @@ int Interp::convert_g7x(int mode,
 		    if(old.ijk_distance_mode()==DISTANCE_MODE::INCREMENTAL)
 			center+=start;
 		}
+
+        /* Verify if circular motion is valid */
+        double tpx;
+        double tpy;
+        double tpz;
+        double tAA_p;
+        double tBB_p;
+        double tCC_p;
+        double tu_p;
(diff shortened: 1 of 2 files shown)


andypugh:
I am trying to figure out how these PRs all fit together. 
Is this still needed?

fsabbatini89:
Hi Andy, yes, this is needed on top of the other fixes,

Franco

PR #2911, opened 2024-03-02
gmoccapy: re-order settings

hansu:
I reordered a bit the elements on Gmoccapys settings pages to find them better.
Let me know what you think or if you think some elements fit somewhere else better (especially @gmoccapy).

For comparison the current pages: https://linuxcnc.org/docs/html/gui/gmoccapy.html#gmoccapy:settings-page

Screenshot_2024-02-26_15-48-57

Screenshot_2024-02-26_15-49-05

Screenshot_2024-02-26_15-49-10


gmoccapy:
For me it is OK, even if I prefer to collect all Hardware related stuff in one page, with your new order it is mixed in two pages.

Norbert

zz912:
I don't see anything wrong with the new layout, but I don't understand why you're changing it. Do you have any plans to expand it in the future? Is it preparation for future stuff? I'm a conservative person and I don't like changes when the original solution is functional.

hansu:
[Norbert]
>>For me it is OK, even if I prefer to collect all Hardware related stuff in one page, with your new order it is mixed in two pages.
>

Hmm I can keep the name Hardware and the content but I don't associate those much with "Hardware":
- "Use Keyboard shortcuts"
- "Turtle Jog Button/factor"
- "Use unlock code/do not use / use hal pin" (ok except the last option)

[Zdeněk]
> I don't see anything wrong with the new layout, but I don't understand why you're changing it. Do you have any plans to expand it in the future? Is it preparation for future stuff? I'm a conservative person and I don't like changes when the original solution is functional.

1. Optimize the way how items can be found more quickly
2. Preparation for https://github.com/LinuxCNC/linuxcnc/pull/2837

1) Because this one I always search in "Appearance"
3_gmoccapy_settings_advanced

To move this to appearance I need more space there, so this is not really an appearance thing:
1_gmoccapy_settings_appearance_

And so on...

zz912:
>     2. Preparation for Gmoccapy: add configuration for 800x600 #2837

This argument convinced me that this PR is a good idea.

hansu:
I went a bit backward, so the Hardware Tab is almost untouched now.
Hope everyone is fine with that now =)

Screenshot_2024-03-16_20-01-05

Screenshot_2024-03-16_20-01-11

Screenshot_2024-03-16_20-01-17

PR #2914, opened 2024-03-04
Update interp_convert.cc

Sigma1912:
With ini setting [RS274NGC] 'DISABLE_G92_PERSISTENCE = 1' G52/G92 should be reset to zero: https://linuxcnc.org/docs/html/gcode/coordinates.html#sec:g52

The current code only resets the numbered parameters [5210 .. 5219]

### DIFF
--- a/src/emc/rs274ngc/interp_convert.cc
+++ b/src/emc/rs274ngc/interp_convert.cc
@@ -4681,10 +4681,43 @@ int Interp::convert_stop(block_pointer block,    //!< pointer to a block of RS27
     }
 
 /*10*/
-    if (settings->disable_g92_persistence)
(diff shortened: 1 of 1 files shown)


Sigma1912:
Maybe I'm reading the documentation wrong but it seems odd to me that an M2/M30 will leave G52/G92 active and only reset the parameter value to zero when DISABLE_G92_PERSISTENCE = 1 has been set in the ini file.
What is the general opinion on this?

Sigma1912:
This is from the NIST publication on the RS274NGC standard:

RS274NGC_m2

Which clearly states in Point 1 that G92/G52 is deactivated (G92.2) on M2/M30. 
So really this PR is actually not correct and I will change it and create an new one shortly

PR #2916, opened 2024-03-05
Update interp_convert.cc: fix broken G52/G92 reset on M2/M30

Sigma1912:
Current code does not reset G52/G92 offset to zero on M2/M30 this is contrary to RS274NGC standard (see Point1 below):

RS274NGC_m2

and can lead to unexpected behavior due to automatic application of G52/G92 offsets on startup:
https://linuxcnc.org/docs/html/gcode/coordinates.html#sec:g52
g92 note

Note: The inserted code is the same as is used for 'G92.2' in the same file:
 
g92_code


Sigma1912:
I'm sorry this should have been in master

PR #2920, opened 2024-03-05
2.9 update docs after fix for g52,g92 reset on m2,m30

Sigma1912:
This are documentation updates for 
https://github.com/LinuxCNC/linuxcnc/pull/2918


Sigma1912:
Descriptions are not quite right yet I noticed.

PR #2921, opened 2024-03-06
Update interp_convert.cc: fix broken G52/G92 reset on M2/M30 

Sigma1912:
Current code does not reset G52/G92 offset to zero on M2/M30 this is contrary to RS274NGC standard (see Point1 below):

RS274NGC_m2

This does not change the behavior of automatic reactivation of G92 offsets on config startup when G92 was active on shutdown if 'DISABLE_G92_PERSISTENCE = 0' . 
For 'DISABLE_G92_PERSISTENCE = 1' the G92 offset values will be zero on startup regardless of the state at shutdown. 
This will also not zero G92 offsets after an ABORT event. This can be  handled by the user with the ON_ABORT_COMMAND in the ini file.

Note: The inserted code is the same as is used for 'G92.2' in the same file:
 g92_code

### DIFF
--- a/src/emc/rs274ngc/interp_convert.cc
+++ b/src/emc/rs274ngc/interp_convert.cc
@@ -4681,10 +4681,35 @@ int Interp::convert_stop(block_pointer block,    //!< pointer to a block of RS27
     }
 
 /*10*/
+    // Clear active G92/G52 offset (same as G92.2 command)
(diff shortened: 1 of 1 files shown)


Sigma1912:
I'll update the documentation once this has been merged

Sigma1912:
included this in my other PR

Issue #2743, opened 2023-11-11
Active gcodes list does not properly reflect G92.x commands  


Sigma1912:
This is from the documentation:

![G92 Commands](https://github.com/LinuxCNC/linuxcnc/assets/46067220/d90eaa3e-cbaa-4500-9f4e-d3f552c8eaa6)

Looking at the 'Active gcodes' list in the axis gui when running G92.. commands: 

![active Gcodes](https://github.com/LinuxCNC/linuxcnc/assets/46067220/9f83af4a-e726-485a-a2d1-aa7ea3fad95e)

1. Command G92X100 will show 'G92' and 'G92.3'
2. Command G92.1 will show 'G92.2'
3. Command G92.2 will show 'G92.2'
4. Command G92.3 will show  'G92.3'   

The 3rd and the 4th are as expected-
The first seems odd.
The second is just plain wrong as it is telling me that the offset values were not set to zero (G92.2) when they really were set to zero (G92.1)
 
[edit]
Also 'G92' disappears from the active gcodes list after (M2,M30) despite the fact that the offset remains applied.  ( related issue for G52: #2744) 

This is using axis gui on a recent build of version 2.9

Sigma1912:
Looks like 'G92.x' are members of two modal groups (0 and 16) which have their own tags in stat.gcodes:

![g92_modal_groups](https://github.com/LinuxCNC/linuxcnc/assets/46067220/afdb0565-2501-4af2-8693-a55698c6df84)


In modal group 16  active G92 offset is indicated by 'G92.3' 
if inactive then it shows 'G92.2' 
![g92_23_active](https://github.com/LinuxCNC/linuxcnc/assets/46067220/41540820-f1c7-4d51-9a7b-0088a39c945c)


While in modal group 0 a 'G92 X...' seems to set the 'G92' tag but a 'G92.[1,2,3]' will remove it. 
Haven't found where that happens in the source code though.

Sigma1912:
and since 'GM_FLAG_G92_IS_APPLIED' directly reflects the state of parameter 5210 a 'G52' command will also set the gcode tag 'G92.3'.
![Flag_G92_is_applied](https://github.com/LinuxCNC/linuxcnc/assets/46067220/ff867280-b1fb-4cf7-9bd8-a32d1aefaa3a)

Sigma1912:
Seems the easier place to fix this would be in the gui's themselves. I'll see if I can come up with something for Axis and Gmoccapy.

[...]

PR #2922, opened 2024-03-06
Update axis.py: hide false display of '92.2'/'92.3' in active gcodes list

Sigma1912:
Fix false indications of 'G92.2' and 'G92.3' in the active Gcode display in Axis GUI

https://github.com/LinuxCNC/linuxcnc/issues/2743


Sigma1912:
Closed for now, as there may be a more elegant solution coming up.

PR #2923, opened 2024-03-06
Update gmoccapy.py: fix false display of 'G92.2'/'G92.3' in active Godes

Sigma1912:
see:
(https://github.com/LinuxCNC/linuxcnc/issues/2743)


hansu:
You have based your commit on master but this PR is targeted to 2.9.
As I think this should go into 2.9 I would please you to rebase your branch to 2.9.

c-morley:
I've been following along on this issue in the background a bit.
Do you think the right fix is to patch all the GUIs to checkbthis?
This seems the wrong approach.

hansu:
I haven't yet looked into the G92 issues in detail, so I cannot say much to the fix(es).

Sigma1912:
Thanks for the feed back.
Have a look at the comments in the issue #2743.

PR #2925, opened 2024-03-07
Made reset work with hal_gpio.

drummer315:
Made reset work with hal_gpio.  I have only been using linuxcnc for about a week, so I don't know if this change is reasonable.  But it works on my Raspberry Pi 5.

### DIFF
--- a/src/hal/drivers/hal_gpio.c
+++ b/src/hal/drivers/hal_gpio.c
@@ -95,12 +95,13 @@ typedef struct{
  *
  * hal_gpio flags
  * INVERT 		= BIT(5)
- * RESET		= BIT(6) - not currently supported
+ * RESET		= BIT(6)
  */
 
 typedef struct {
     int num_lines;
     int *vals;
+    int *old_vals;
     int *flags;
     hal_gpio_hal_t *hal;
     struct gpiod_chip *chip;
@@ -183,6 +184,7 @@ int build_chips_collection(char *name, hal_gpio_bulk_t **ptr, int *count){
 	    (*ptr)[c].chip = NULL;
 	    (*ptr)[c].flags = NULL;
 	    (*ptr)[c].vals = NULL;
+	    (*ptr)[c].old_vals = NULL;
 	    (*ptr)[c].num_lines = 0;
 	    (*ptr)[c].chip = gpiod_line_get_chip(temp_line);
 	    (*ptr)[c].bulk = rtapi_kmalloc(sizeof(*(*ptr)[c].bulk), RTAPI_GFP_KERNEL);
@@ -198,6 +200,7 @@ int build_chips_collection(char *name, hal_gpio_bulk_t **ptr, int *count){
     gpiod_line_set_flags(temp_line, (*ptr)[c].flags[(*ptr)[c].num_lines - 1]);
 #endif
     (*ptr)[c].vals = rtapi_krealloc((*ptr)[c].vals, (*ptr)[c].num_lines * sizeof(int), RTAPI_GFP_KERNEL);
+    (*ptr)[c].old_vals = rtapi_krealloc((*ptr)[c].old_vals, (*ptr)[c].num_lines * sizeof(int), RTAPI_GFP_KERNEL);
     gpiod_line_bulk_add((*ptr)[c].bulk, temp_line);
 
     return 0;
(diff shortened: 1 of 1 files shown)


andypugh:
Do you definitely need the old_vals? 
The use-case for reset is working with a stepgen, and in that case the step-time is set to zero. 
I think that this means that the input HAL pin to the hal_gpio driver pin will always return to zero on the next thread invocation, so it is safe to blindly write the HAL value to the hardware pin without edge checking. 

If you look at the equivalent code in the hal_parport driver, I don't see any edge-detection there: https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/drivers/hal_parport.c

You seem to have put the test for whether the deadline has expired before the calculation of what the output pin states will be. 
I had it deliberately after to save cycle time (as you are then assembling the code _during_ the reset time, rather than after it.

drummer315:
If I understand what you are saying, I should set stepgen.steplen = 0 when using reset.  Maybe that's why I could not get reset to work without my patch.  There was comment saying reset wasn't implemented, so I thought it was incomplete.  I don't know how stepgen worked and stepconf set steplen=1.  Then logic told me that I wanted to generate a pulse starting at each transition and ending 5ms later.  So that's what I did--but from what you are saying, I think it is the wrong way to do it.  I will try without the patch and setting steplen=0.

andypugh:
steplen = 1 should also be fine, that's 1 nanosecond.

drummer315:
I haven't been able to get reset to work without the patch.  But that could be because I don't know how to use linuxcnc.  I think you are saying that reset works without this patch.  But you haven't stated that clearly.  Does reset work with hal_gpio as is in Master?  When I try to use without patch, I get no errors, but it steps at same rate as without reset.  With this patch, it steps at double the rate.  If reset works in Master, then I will delete this pull request.

drummer315:
I must be stupid.  I can't get reset to work without detecting both edges.  I don't understand how hal_gpio or hal_parport communicate to stepgen that it reset the line.  As far as I can tell, they do not.  And if they do not, I don't see how it can work without detecting both edges.  But then, I am stupid, so maybe that's why I don't understand.

> It is also debatable if the reset function is particularly useful with real mechanical systems, too. The maximum step rate is one step per cycle. The next slower rate is one thread per two cycles. Can the stepper motor track that doubling in speed?

Well, this seems better than maximum step rate is 1/2 per cycle and next slower rate is 1/3 per cycle.  I don't have any experience with CNC or steppers or any of this stuff, so I could be wrong.  I just got a CNC machine to make some stupidly expensive and hard to find drum parts.

drummer315:
I realize you don't need to teach a beginner like me how this stuff works.  And I am thankful for your help so far.  But I just can't get this thing to work without detecting both edges.  I don't understand how hal_parport can work.  This is what I think is happening based on my reading of stepgen.c, hal_parport.c and hal_gpiod.c.  Let's assume stepping at maximum rate and reset active and no inverting.

1. stepgen.make-pulses sets stepgen.step high.
2. hal_gpio.write tells gpiod to set the physical pin high.
3. hal_gpio.reset busy waits 5000ns.
4. hal_gpio.reset tells gpiod to set the physical pin low.
5. yield for the balance of BASE_PERIOD
6. stepgen.make-pulses sets stepgen.step low.
7. hal_gpio.write tells gpiod to set the physical pin low--nothing happens because it is already low.
8. yield for the balance of BASE_PERIOD
9. start over at step 1.

Result:  one step every two BASE_PERIOD.

With the patch, this is what I think is happening:

1. stepgen.make-pulses sets stepgen.step high.
2. hal_gpio.write detects edge and tells gpiod to set the physical pin high.
3. hal_gpio.reset busy waits 5000ns.
4. hal_gpio.reset tells gpiod to set the physical pin low.
5. yield for the balance of BASE_PERIOD
6. stepgen.make-pulses sets stepgen.step low.
7. hal_gpio.write detects edge and tells gpiod to set the physical pin high.
8. hal_gpio.reset busy waits 5000ns.
9. hal_gpio.reset tells gpiod to set the physical pin low.
10. yield for the balance of BASE_PERIOD
11. start over at step 1.

Result:  one step every BASE_PERIOD.

I must be misunderstanding something.

drummer315:
I see.  I misunderstood how linuxcnc works.  I thought, incorrectly, with reset enabled, versus reset disabled, and no other changed to the hal or ini files, that it should step twice as fast.  But based on what you are saying, it should not.  It should step at the same rate.  However, it should work at twice the step rate.  

With my new understanding, I believe hal_gpio.reset works perfectly as is in Master.  The only thing I would change is get rid of that comment saying that reset isn't supported yet.  Well, except that it doesn't work if you have no input pins.  Work around is to pick some random pin to use as a dummy input pin.

Thank you for your help.  I am sorry for wasting your time with my ignorance.

andypugh:
> I see. I misunderstood how linuxcnc works. I thought, incorrectly, with reset enabled, versus reset disabled, and no other changed to the hal or ini files, that it should step twice as fast. But based on what you are saying, it should not. It should step at the same rate. However, it should work at twice the step rate.

Yes that's right. Perhaps the documentation needs to be clearer on this point. 

> The only thing I would change is get rid of that comment saying that reset isn't supported yet. 

Yes, that does seem like it should be changed. 

> Well, except that it doesn't work if you have no input pins. Work around is to pick some random pin to use as a dummy input pin.

That sounds like an interesting bug that I was unaware of. Can you raise an issue on the tracker about this? (and assign it to me if the system lets you do that)

andypugh:
I have removed the misleading comments: https://github.com/LinuxCNC/linuxcnc/commit/fadbc8d0ce914df90f7e8a213cca884b36f8b478

Issue #2744, opened 2023-11-11
Active Gcodes list does not properly reflect applied G52 offsets


Sigma1912:
1. Command G52 X100 -> Offset is applied in DRO and active GCodes will show 'G52'
2. Command M2  (or M30)  -> Offset remains applied in DRO but active GCodes no longer shows 'G52'

This is from the documentation:

![G52 caution](https://github.com/LinuxCNC/linuxcnc/assets/46067220/0d9a391c-eeea-492e-99f1-ab9fd83d4f65)

Since the G52/G92 offset remains applied after  M2/M30 then 'G52'/'G92' should not be removed from the active gcodes list.

( related issue for G92: #2743) 

This is using a recent build of version 2.9

Sigma1912:
Looking at this again it seems that the actual issue here is that G52 persists after M2/M30 when it should really be cleared.

Sigma1912:
Seems that G52/G92 handling in LinuxCNC does not follow NISTs RS274NGC standard.
While in the standard the offset  is deactivated on 'm2'/'m30' and the offset values are made 'persistent' by storing them, LinuxCNC will not deactivate G92/G52 offsets on 'm2'/'m30' and both values AND activation state are 'persistent'.
https://tsapps.nist.gov/publication/get_pdf.cfm?pub_id=823374

![RS274NGC_m2](https://github.com/LinuxCNC/linuxcnc/assets/46067220/d9a3de60-4e77-4b2d-b859-048ccc994dbb)  
Looking at the number of tests failing when changing this makes me think that this must have been done intentionally.

andypugh:
Feel free to re-open this if #2935 wasn't actually the fix

PR #2931, opened 2024-03-10
2.9 g92 fix

Sigma1912:
See https://github.com/LinuxCNC/linuxcnc/issues/2743

Summary:
This reverts https://github.com/LinuxCNC/linuxcnc/commit/26aeab962d82a8249275e98089bff37a9e342845
due to issue https://github.com/LinuxCNC/linuxcnc/issues/2743

Fixes the original issue of missing indication of active G92 offsets after an abort (the issue originally to be fixed by the reverted commit above)

Disables automatic activation of persistent G92 offset values on startup

Fixes missing G92/G52 reset on M2,M30 after M2,M30 as described in the RX274NGC standard:

RS274NGC_m2 

See https://github.com/LinuxCNC/linuxcnc/issues/2744


Sigma1912:
Not sure why the checks keep failing on most of my PRs

andypugh:
It's probably not you. Let me investigate.

andypugh:
Tests are passing on the buildbots and my test machine, I will try re-running the tests on this PR.

Sigma1912:
It is indeed me. I was not aware that this non-compliance with the standard seems to be built into the tests.
I'm retracting this as this would probably require a more engaged discussion.

PR #2953, opened 2024-03-27
hal_stream_create: fix bogus error check

jlama:
hal_stream_create() did not check the return value of halpr_parse_types() properly, so the stream could be created with any kind of invalid cfg string.


andypugh:
This feels like a bugfix, and might be appropriate for inclusion in 2.9. Do you want to rebase it against 2.9?

jlama:
Sure, but I'm not sure how to proceed. Feel free to do it.

hansu:
Rebase would cause a lot of merge conflict so I would recommend creating a new branch and cherry pick that commit and create a new PR (or force push that branch to [jlama:hal_stream](https://github.com/jlama/linuxcnc/tree/hal_stream)).

andypugh:
I will do it via cherry-pick.

andypugh:
Actually, I did it by saving the PR out of Github as a patch, and then applying that. 
The trick there is to click to the commit page, eg:
https://github.com/LinuxCNC/linuxcnc/pull/2953/commits/253c3b9056841abec95586f46063c120c7906f72
Then append ".patch" to the end
https://github.com/LinuxCNC/linuxcnc/pull/2953/commits/253c3b9056841abec95586f46063c120c7906f72.patch

And that gives a plain-text patch that can be applied with "git am" and that keeps attributions. 

It's in as a83d4e0e8

andypugh:
Merged indirectly into 2.9

Issue #2902, opened 2024-02-20
pncconf default firmware choice for 7I96S confuses newbees.


pcw-mesa:
pncconfs default firmware for the 7I96S is 7I96s_7i74 which creates configurations
that will not run on the standard  pre-installed firmware (7i96s_d) 

7I96s_d should be the default

zz912:
7i96 has the same problem. 7i96d is not first.
![pncconf-default_firmware](https://github.com/LinuxCNC/linuxcnc/assets/96618597/180ecb5b-d17d-45e6-b408-41d8c0565302)

pcw-mesa:
Thank you for fixing this!

phillc54:
Fixed by [PR #2955](https://github.com/LinuxCNC/linuxcnc/pull/2955)

PR #2955, opened 2024-04-01
pncconf: don't sort firmware names

phillc54:
Use the order the firmwares appear in private_data.
See issue #2902


zz912:
I tested it. It works well.

c-morley:
If this helps people - push away :)

c-morley (approved):
All good

PR #2959, opened 2024-04-07
update translation files (.pot)

hansu:


petterreinholdtsen:
Look like a good start.  I normally do the POT file updates directly
into master without a pull request, as doing such update include locking
weblate and updating the PO files as well, which can not reasonably be
handled as a asynchronous pull request.

Lets sit down this evening to see how to best do this.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
Just preparation for this evening to be able to do this without the VM =)

PR #2977, opened 2024-05-15
qtaxis: Move major window sections into splitters so you can resize

hachi:
This diff looks a little overwhelming if viewed normally, but viewed with whitespace changes ignored it really is just:

Wrap leftTab and rightTab in a splitter
Wrap that splitter and gcode_editor in a splitter
Wrap rightTab and gcode_editor in a frame/layout combo to make them render properly
Tweak a few default height/width values
Remove height limit on gcode_editor box
Make leftTab adjustable from 300 to 400px wide
Change handler code so edit mode hides the upper portion (as intended)


c-morley:
I have pushed your changes to master branch manually. Thank you for your work!

PR #2981, opened 2024-05-18
qtaxis: change all print calls in handler to LOG.something

hachi:
This changes any print() calls in the qtaxis_handler.py to LOG.something() calls which are silenced unless you start with -v


hachi:
Closing, @c-morley fixed this in https://github.com/LinuxCNC/linuxcnc/commit/378ad14590984b11f3f14251a43f6c560cb10c6b

PR #2984, opened 2024-05-21
fix float arg passed to QSize, caused Qt designer to crash when loading

tangentaudio:
When loading the qttouchy ui file into Qt Designer, it would crash:

```
  File "/usr/lib/python3/dist-packages/qtvcp/widgets/tab_widget.py", line 43, in tabSizeHint
    return QSize(size.width(), w)
           ^^^^^^^^^^^^^^^^^^^^^^
TypeError: arguments did not match any overloaded call:
  QSize(): too many arguments
  QSize(w: int, h: int): argument 2 has unexpected type 'float'
  QSize(a0: QSize): argument 1 has unexpected type 'int'
```

Fixed the 'w' argument by converting it to an int.


c-morley:
Thank you.

PR #2990, opened 2024-05-31
2.9

fsabbatini89:
Fix error joint limits zeroing variables before using


andypugh:
Can you describe what this fixes in more detail? Perhaps there is a related bug report or a forum thread?

PR #2991, opened 2024-05-31
Update interp_convert.cc: Fix bug in G43.2

Sigma1912:
Fix error when using eg 'G43.2 z1  '  (ie H word is omitted)
No point trying to find the tool using the h_number when the user specified axes parameters instead.


Sigma1912:
This is in line with the original commit: 
https://github.com/LinuxCNC/linuxcnc/commit/2df2871ba06988f51a0b314d1962113ac9c3ed8b

I'm not sure what happened in the commits after that as the code seems to revert back and then the current form appears here
https://github.com/LinuxCNC/linuxcnc/blob/3765decc8626c191c0c2797f91ea4ffff6ab7513/src/emc/rs274ngc/interp_convert.cc

Issue #2986, opened 2024-05-28
Posemath #define toLine() is same defined as in Qt library. 


krcwrhvgyxzqelljes:
Hi,

~/linuxcnc/src/libnml/posemath/posemath.h
Line 978

The problem is that same name is used in qt libraries.

#define toLine(src,dst) {toPose((src).start, &((dst)->start)); toPose((src).end, &((dst)->end)); toCart((src).uVec, &((dst)->uVec));}

Can you rename this to :

#define toLine_(src,dst) {toPose((src).start, &((dst)->start)); toPose((src).end, &((dst)->end)); toCart((src).uVec, &((dst)->uVec));}

Then it compiles along with qt libraries and problem is solved.

~/qt-creator/5.15.1/gcc_64/include/QtCore/qvariant.h
line 324 :    QLine toLine() const;

Compile ouput:
error: expected ';' at end of declaration list
    QLine toLine() const;

rmu75:
These toLine etc... macros shouldn't be macros in the first place and they should be moved inside the #ifdef __cplusplus block.

rmu75:
These toLine etc... macros shouldn't be macros in the first place and they should be moved inside the #ifdef __cplusplus block.

rene-dev:
closed by https://github.com/LinuxCNC/linuxcnc/pull/2999

Issue #2995, opened 2024-06-11
Remove StateTag tag. As it is never used.


krcwrhvgyxzqelljes:
/home/user/linuxcnc/src/emc/rs274ngc/interp_convert.cc

Line 4357 

StateTag tag;

This can be deleted as it is never used.

andypugh:
In which version? 
In master it is: https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/rs274ngc/interp_convert.cc#L4357
In 2.9 it is https://github.com/LinuxCNC/linuxcnc/blob/2.9/src/emc/rs274ngc/interp_convert.cc#L4357

rmu75:
https://github.com/LinuxCNC/linuxcnc/blob/2.9/src/emc/rs274ngc/interp_convert.cc#L3811
https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/rs274ngc/interp_convert.cc#L4352

andypugh:
OK, yes, that seems to never be used in that scope. 
The compiler probably ignores it, but it does seem untidy.

rene-dev:
closed by https://github.com/LinuxCNC/linuxcnc/pull/2999

PR #2999, opened 2024-06-26
refactor posemath macros

rmu75:
posemath.h: replace some macros with function templates.

posemath.h contains some macros for conversion between C structs and C++
classes. Those macros are named like 'toLine' and don't follow usual
convention that #define should use CAPITALIZED_NAMES. To avoid
name collisions, replace macros with function templates and make sure
those are only enabled if compiled as C++.


rmu75:
fixes #2986 and #2995

PR #3004, opened 2024-07-04
Unset FORTIFY_SOURCE before setting to value 2

NTULINUX:
On some configurations (whether it's custom {C,CXX}FLAGS on ./configure or Gentoo where users specify -D_FORTIFY_SOURCE in make.conf) you will see many warnings about FORTIFY_SOURCE.

This unsets whatever FORTIFY_SOURCE was set to, then sets it to 2.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -252,7 +252,7 @@ INCLUDE += $(LIBTIRPC_CFLAGS)
 # Compilation options.	Perhaps some of these should come from Makefile.inc? (CXXFLAGS now does)
 INTEGER_OVERFLOW_FLAGS := -fwrapv
 OPT := -Os $(INTEGER_OVERFLOW_FLAGS)
-DEBUG := $(DEBUG) -g -Wall -D_FORTIFY_SOURCE=2
+DEBUG := $(DEBUG) -g -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2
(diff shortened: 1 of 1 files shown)


petterreinholdtsen:
[NTU]
> On some configurations (whether it&#39;s custom {C,CXX}FLAGS on
> ./configure or Gentoo where users specify -D_FORTIFY_SOURCE in
> make.conf, you will see many warnings about FORTIFY_SOURCE.
>
> This unsets whatever FORTIFY_SOURCE was set to, then sets it to 2.

Why is the correct fix to change FORTIFY_SOURCE, and not to change the
code where the warning show up?

-- 
Happy hacking
Petter Reinholdtsen

NTULINUX:
The warning is built into the compiler not LinuxCNC. GCC used to complain not sure anymore, Clang still does and the problem is that you're setting a variable without getting rid of the previous variable so it's being redefined. I forget the exact error because I'm not hitting it with GCC at the moment but most projects that use -D_FORTIFY_SOURCE, you'll see they have -U_FORTIFY_SOURCE to clear it. Maybe it's with older versions of GCC and they pulled the warning out, not sure.

NTULINUX:
Ah ha! I got the warnings back. For some reason, passing CFLAGS to CXXFLAGS to ./configure doesn't work but running `export CFLAGS="-O2 -D_FORTIFY_SOURCE=3" && export CXXFLAGS="-O2 -D_FORTIFY_SOURCE=3` makes it trigger both in Clang and GCC.

NTULINUX:
OK, this is odd. The reason why it only works with `export CFLAGS/CXXFLAGS` is because LinuxCNC doesn't respect CFLAGS/CXXFLAGS on `./configure` but rather `make`.

i.e. `CXXFLAGS="-O2 -D_FORTIFY_SOURCE=3" make`

PR #3015, opened 2024-07-07
Allow uniq_id to be used to select hal_input devices.

mark-v-d:
When using several input devices this is usefull


andypugh:
Seems like a good idea to me, but you missed the 2.9.3 release by a matter of minutes

PR #3021, opened 2024-07-09
hal_gpio - fix output length check using invalid inputs variable

otaviojr:
Using the input length will break if the number of inputs is less than the number of outputs.

Besides, using input length is probably not the intended thing to do.


andypugh:
Oops! I am sure you are right.

Issue #2966, opened 2024-04-20
Glade Tooltable widget destroys tooltable in a particular case


hansu:
When using locale settings for e.g. Germany, so having a "," as decimal point, the tool table shows the correct decimal point after starting.
But after loading an (empty) G-code file and reloading the tool table, the decimal point changes to a dot:

https://github.com/LinuxCNC/linuxcnc/assets/67957916/421d6f9a-7a8d-4f6c-90cb-40ac777b398f

This can lead to loss of the complete tool table in the following (quite common) case:

1. Load a G-code file
2. Add the missing tool
3. Apply the changes ... boom! all tools are gone!

https://github.com/LinuxCNC/linuxcnc/assets/67957916/a59e732b-f6b2-4aa1-a6b7-9ae81574fa16

Version: latest from 2.9 branch.

```
Traceback (most recent call last):
  File "/home/cnc/linuxcnc/linuxcnc-2.9/bin/gmoccapy", line 4947, in on_btn_apply_tool_changes_clicked
    self.widgets.tooledit1.save(None)
  File "/home/cnc/linuxcnc/linuxcnc-2.9/lib/python/gladevcp/tooledit_widget.py", line 348, in save
    line = line + "%s%s "%(KEYWORDS[num], locale.atof(test))
                                          ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/locale.py", line 338, in atof
    return func(delocalize(string))
           ^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: could not convert string to float: '0,0000'
 (gmoccapy:62)
Tool file was modified since it was last read
/usr/lib/python3/dist-packages/gi/overrides/GObject.py:493: Warning: ../../../gobject/gsignal.c:2772: instance '0x460ff90' has no handler with id '113342181'
  return func(*args, **kwargs)
```

zz912:
I can confirm this behavior. Any G-code file can be used for recall, not just an empty G-code file.

In place for tool information is used as a decimal point and a dot. It is also in the Czech localization.
![tool](https://github.com/LinuxCNC/linuxcnc/assets/96618597/ba2ed833-6340-4f82-91e7-d7667499f2a4)
It looks wierd.

hansu:
One possible workarounds are:

hansu:
The problem does not appear on Debian Bullseye / Python 3.9.2

hansu:
[...]

PR #3026, opened 2024-07-11
Tool edit widget: backup tool table and add some checks

hansu:
Resolves https://github.com/LinuxCNC/linuxcnc/issues/2966

Edit: It creates a backup of the tooltable when attempting to save it and some checks if the locale is valid before converting the numbers before saving.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,5 @@ docs/help/hu/*
 docs/help/nb/*
 docs/help/vi/*
 docs/help/zh_CN/*
+configs/**/*.tbl.bak
+configs/**/halshow.preferences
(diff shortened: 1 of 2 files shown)


zz912:
I tested it. It works.

hansu:
Originally thought as a fix for https://github.com/LinuxCNC/linuxcnc/issues/2966 which is now resolved by https://github.com/LinuxCNC/linuxcnc/pull/3041

But what do you think about backing up the tool table before saving and the additional checks?

zz912:
> But what do you think about backing up the tool table before saving and the additional checks?

It should definitely be merged.

PR #3035, opened 2024-07-21
litehm2 support

sensille:
litehm2 support was added to master with cb81d5ba73db5a34683718e7f853f0820ea3dcad. It also made it partially into 2.9 with d72b9d444710b28814100766bfb88ca2d7e55aab. This pull request is to merge the remaining bits into the next release, so litehm2 users don't need to build lcnc from master.


andypugh:
I am a bit nervous about merging this without checking that it doesn't cause problems with normal Mesa cards. That will have to wait a couple of days.

pcw-mesa:
I have been using master since the change was added ~1.5 years
ago and have not seen any issues with our cards.

PR #3036, opened 2024-07-21
glcanon.py - reset localization

zz912:
Added locale.setlocale(locale.LC_ALL, '') for reset localization after using 'gcode' library. The 'gcode' library is a 'C' language and caused delocalization.

### DIFF
--- a/lib/python/rs274/glcanon.py
+++ b/lib/python/rs274/glcanon.py
@@ -25,7 +25,15 @@ import array
 import gcode
 import os
 import re
+import sys
 from functools import reduce
+
+# localization
+import locale
+BASE = os.path.abspath(os.path.join(os.path.dirname(sys.argv[0]), ".."))
+LOCALEDIR = os.path.join(BASE, "share", "locale")
+locale.setlocale(locale.LC_ALL, '')
+
 
 def minmax(*args):
     return min(*args), max(*args)
@@ -1941,6 +1949,10 @@ class GlCanonDraw:
             self.stale_dlist('program_norapids')
             self.stale_dlist('select_rapids')
             self.stale_dlist('select_norapids')
+
+        # library gcode is written in 'C' language which causes delocalization.
+        # Therefore, we have to activate the default localization again
+        locale.setlocale(locale.LC_ALL, '')
 
         return result, seq
 


zz912:
More information about this PR can be found here: https://github.com/LinuxCNC/linuxcnc/issues/2966

HansU made PR https://github.com/LinuxCNC/linuxcnc/pull/3026. 

I consider this PR to be a complement to the HansU PR. It is not a replacement for HansU PR.

I would like to ask for advice: Why
some checks were not successful? Is it my fault?

hansu:
If I understood you correctly this doesn't resolve it? (https://github.com/LinuxCNC/linuxcnc/issues/2966#issuecomment-2241035733)

zz912:
This PR resolve https://github.com/LinuxCNC/linuxcnc/issues/2966#issuecomment-2241035733

BUT I don't know if the problem can be solved even more low. I don't know if it is possible to do the correct localization in linuxcnc-2.9/lib/python/gcode.so . I don't have enough skills for that. If it were possible to set the correct localization in the gcode.so source, then this PR would be unnecessary.

Please be patient with my bad English.

rmu75:
IMO the root cause of the problem needs to be traced, randomly sprinkling setlocale around will not really fix the problem.

zz912:
> randomly sprinkling setlocale around will not really fix the problem.

I tried not to sprinkle randomly. I tried to find the delocalization as deep as possible.
Can you solve it better? I would like to ask for help or advice on how to do it better.

rmu75:
I think the real fix would be to use a proper numeric input widget in tool table edit, but I don't know how to do that in Gtk or if  that is even possible with stock widgets. (GtkSpinButton?) I looked at the C code and am surprised that something is changing locale.

Same problem exists on work offset page and other gmoccapy inputs of floats.

rmu75:
Maybe something fishy is going on here https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/rs274ngc/interp_internal.hh#L996

This should restore locale but I'm not sure -- *checking*.

rmu75:
yes there is something fishy going on there -- the restore doesn't work as expected.

rmu75:
yes, there is the problem. I have a fix.

rmu75:
for the record if somebody wants to fix numeric entry in gmoccapy and co, here is a link to the solution in the tryton client  https://foss.heptapod.net/tryton/tryton/-/blob/branch/default/tryton/tryton/common/number_entry.py

zz912:
PR https://github.com/LinuxCNC/linuxcnc/pull/3040 makes this PR unnecessary.

PR #3040, opened 2024-07-24
fix restoring of locale setting in interp

rmu75:
locale string returned by setlocale needs to be copied, subsequent calls to setlocale overwrite location pointed to by previous call rendering it illegal.


rmu75:
This should fix locale-related problems occuring aftercalls to the interpreter, e.g. #2966 and should make PRs like #3036  resetting locale in python unnecessary.

zz912:
Wow, it is working. You are genius. I dont know, how this patch work, but it works. Thank you.

rmu75:
setlocale returns a pointer to a string describing (in an opaque way, so may not be directly readable) the current locale. but that string is not valid any more if you set a new locale. so the string needs to be copied. std::string is an "owning" container and copies the string it is initialized with.

hansu:
Really great, thank you!
(But one PR to 2.9 would have been sufficient ;-) )

PR #3048, opened 2024-07-29
gmoccapy: add sim config for gcmc support

hansu:
Add a sim config for Gmoccapy with support for gcmc (G-code Meta Compiler, https://www.vagrearg.org/content/gcmc) analog to the gcmc Axis config.


configs/sim/gmoccapy/gmoccapy_pyngcgui_gcmc.txt
```diff
@@ -0,0 +1,24 @@
...
+
+The .ngc and .gcmc files can be intermixed with
+ini file lines like:
+
+NGCGUI_SUBFILE    = drill.gcmc
+NGCGUI_SUBFILE    = wheels.gcmc
+NGCGUI_SUBFILE    = simp.ngc
+
+Installation instructions:
+1. Download from https://gitlab.com/gcmc/gcmc
```
smoe, line 17: If this is something we want for LinuxCNC, should we then not go and package it?

hansu: This would be a good idea!

andypugh:
Gcmc is something that I have been aware of for a long time, but I have never had any interest in, and so haven't looked at in in any detail. 

But, would it perhaps make more sense for the LinuxCNC config to be part of the gcmc download than the LinuxCNC one?

hansu:
>But, would it perhaps make more sense for the LinuxCNC config to be part of the gcmc download than the LinuxCNC one?

But there is already some gcmc stuff inside LinuxCNC (configs/nc_files/gcmc_lib), gcmc support in ngcgui and a sim config for axis so this config will just round that off.

andypugh:
OK, then I am happy to merge it.

Issue #3065, opened 2024-08-12
TCL error with master build: bad menu entry index "G-code properties..."


Sigma1912:
This error pops up on a current (12. August) build of master:
Maybe related to https://github.com/LinuxCNC/linuxcnc/commit/4cb5170b4368d79254b0220a79480d2f5b9d7c78

```
LINUXCNC - 2.10.0~pre0
Machine configuration directory is '/home/user/git/linuxcnc-master-august-12/configs/sim/axis'
Machine configuration file is 'axis.ini'
Starting LinuxCNC...
linuxcncsvr (374409) emcsvr: machine 'LinuxCNC-HAL-SIM-AXIS'  version '1.1'
linuxcnc TPMOD=tpmod HOMEMOD=homemod EMCMOT=motmod
Note: Using POSIX non-realtime
milltask (374423) task: machine 'LinuxCNC-HAL-SIM-AXIS'  version '1.1'
halui (374425) halui: machine 'LinuxCNC-HAL-SIM-AXIS'  version '1.1'
Found file(lib): /home/user/git/linuxcnc-master-august-12/lib/hallib/core_sim.hal
Found file(lib): /home/user/git/linuxcnc-master-august-12/lib/hallib/sim_spindle_encoder.hal
Found file(lib): /home/user/git/linuxcnc-master-august-12/lib/hallib/axis_manualtoolchange.hal
Found file(lib): /home/user/git/linuxcnc-master-august-12/lib/hallib/simulated_home.hal
Found file(lib): /home/user/git/linuxcnc-master-august-12/lib/hallib/check_xyz_constraints.hal
Found file(REL): ./cooling.hal
USRMOT: ERROR: command 30 timeout
emcMotionInit: emcTrajInit failed
link (updating variable file): No such file or directory
note: MAXV     max: 5.000 units/sec 300.000 units/min
note: LJOG     max: 5.000 units/sec 300.000 units/min
note: LJOG default: 0.250 units/sec 15.000 units/min
note: jog_order='XYZ'
note: jog_invert=set()
TCL error in asynchronous code:
bad menu entry index "G-code properties..."
    while executing
"$m entrycget $idx -state"
    (procedure "state" line 9)
    invoked from within
"state  {$taskfile != ""} {.menu.file "G-code _properties..."}"
    (procedure "update_state" line 13)
    invoked from within
"update_state"
    ("after" script)

```

Sigma1912:
[...]

PR #3070, opened 2024-08-16
Fix typo in 'share/axis/axis.tcl' creating error 'bad menu entry index "G-code properties..."'

Sigma1912:
Fixes error:
bad menu entry index "G-code properties..."

https://github.com/LinuxCNC/linuxcnc/issues/3065


hansu:
Thanks for spotting that. But the correct fix would be to remove the underscore in axis.tcl:80.

Sigma1912:
Ah yes. I was too rushed to get the error away :) 
I'm closing this, so maybe you can push the correct fix.

hansu:
Yes, fixed with d8a09efdaecb4a23596501ff27a48963714cd0bd

PR #3071, opened 2024-08-17
build: d/configure gives constructive comments, severity prefix for messages

smoe:
These "I:|W:|E:" prefixes for messages to the console are rather useful and found in many other projects. I suggest to adopt them where we can.

Was about to revisit https://github.com/LinuxCNC/linuxcnc/pull/2647https://github.com/LinuxCNC/linuxcnc/pull/2647 when I thought that too many folks likely do not reread the manual and that little piece of magic should likely look good in debian/configure.


smoe:
Improved wording and eliminated some typos, rebased to master.

smoe:
I like it, want to test a bit more, yet.

petterreinholdtsen:
I agree that such prefixes are useful.  In my own code, I tend to use error: warning: info: etc instead of the short form, to make searching for them easier, as it is unlikely to get any false matches.

petterreinholdtsen:
Still look good to me.

petterreinholdtsen (approved):

PR #3077, opened 2024-08-18
Fix glade panel and remove depricated 'Features' entries in ini

Sigma1912:
This sim config has been broken for a while now


Sigma1912:
Closing as this should have gone into 2.9.

PR #3080, opened 2024-08-19
New simulation config for machine with one table rotary and two spindle rotary joints

Sigma1912:
As presented here
https://forum.linuxcnc.org/show-your-stuff/49103-kinematic-model-for-a-5axis-mill-with-universal-nutating-head?start=0


Sigma1912:
Seems adding  the manpages for the vismach guis doesn't work like this so I'll revert that commit and try again.

PR #3083, opened 2024-08-24
Migrate to GtkSourceview 4

hansu:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1079419


petterreinholdtsen:
My idea is that such change is a rather big one, and thus belong in
master, not 2.9.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
> My idea is that such change is a rather big one, and thus belong in master, not 2.9.
> […](#)
> -- Happy hacking Petter Reinholdtsen

This was my first thought too then I realized that the bug was posted on the Debian Bug Tracker for 2.9 and might be expected to be fixed in 2.9...

andypugh:
Indeed. Fixing it in master won't address the Debian bug.

hansu:
> Can we close the debian bug without fixing it? I don't see it as a genuine bug, if it actually currently works.

It does not really work actually in my opinion. Not sure if this resolves it, but on my 2.9 system which was installed from the Linuxcnc ISO, the sourceview does not have any syntax highlighting. My dev system which runs on bookworm from a debian ISO does have highlighting. A bit weird. I didn't have the time yet to investigate this further unfortunately.

rene-dev:
libgtksourceview-4-dev is available in all debian versions, so I see no issue in changing master and 2.9

andypugh:
OK, I will merge it. But that only goes part-way to addressing the Debian "Bug". Someone still needs to go through the (not inconsiderable) pain of releasing a new version of LinuxCNC 2.9. And then someone else also has the (not inconsiderable) pain of submitting that new release into the Debian repository.

hansu:
When building for RIP, it is a bit inconvenient that  `sudo apt-get build-dep .` doesn't install `libgtksourceview-4-dev` because it is defined under "Depends".
Is there a possibility to get also the "Depends" additionally to the "Build-Depends"?

PR #3087, opened 2024-08-26
fix configs/apps/gladevcp: update to gtk3

Sigma1912:
'sourceview' and 'animated backdrop' still not 100% functional but at least they load ok.


Sigma1912:
Closing this as it should have gone into 2.9.

PR #3088, opened 2024-08-26
fix configs/apps/gladevcp: update to gtk3

Sigma1912:
'sourceview' and 'animated backdrop' still not 100% functional but at least they load ok.

### DIFF
--- a/configs/apps/gladevcp/animated-backdrop/cairodraw.ui
+++ b/configs/apps/gladevcp/animated-backdrop/cairodraw.ui
@@ -1,18 +1,21 @@
-<?xml version="1.0"?>
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.38.2 -->
 <interface>
-  <!-- interface-requires gladevcp 0.0 -->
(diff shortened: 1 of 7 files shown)


hansu:
> 'sourceview' and 'animated backdrop' still not 100% functional but at least they load ok.

Doesn't look like a huge thing to get also the `_set_line function` compatible with GtkSourview 3. Or am I wrong?
Would be nice if we could have this complete.

Sigma1912:
Alright :) 
I'll be back with another one.

andypugh:
Is this superceded by moving to GtkSourceview4 ?

Sigma1912:
I'm afraid I don't know what moving 2.9 to GtkSourceview4 means for GUIs that use GTK3. I would expect @hansu to be more aware of the consequences.

Note, the current PR is over here:
https://github.com/LinuxCNC/linuxcnc/pull/3090

hansu:
> Is this superceded by moving to GtkSourceview4 ?

No,  "Both GtkSourceView 3 and GtkSourceView 4 depend on GTK 3."

PR #3089, opened 2024-08-26
Add note about required 'XTerm' to README 'configs/apps/xhc-hb04' and remove spanish README

Sigma1912:
The bash script errors out in terminal apps other than 'XTerm'


petterreinholdtsen:
Why do you remove README_es in a commit claiming to only add a documentation note?

Sigma1912:
Correct thanks. Do you want me to put it back in? I don't speak spanish so I cannot modify the content.
Although I would want to question the practicality of having README files in different languages, doesn't make code maintenance any easier.

petterreinholdtsen:
[Sigma1912]
> Correct thanks. Do you want me to put it back in? I don't speak
> spanish so I cannot modify the content.
> Although I would want to question the practicality of having README
> files in different languages, doesn't make code maintenance any
> easier.

I do not care much about what happen to the README file, my issue is
with a correct commit description accurately expression what is intended
in the change, to make it obvious that the file was not removed by
mistake, and explain why it was removed.

On the other hand, I believe translated README files should be handled
by the po4a framework we use for documentation, to avoid placing a
burden on the code maintainers.

-- 
Happy hacking
Petter Reinholdtsen

Sigma1912:
I put the README.es back in as the issue about the README.es files may need more discussion.
I'm closing this PR to be replaced by #3092

PR #3090, opened 2024-08-26
2.9 fix configs apps gladevcp

Sigma1912:
 Note 'animated backdrop' window is transparent (which is kind of cool in itself) but I'm afraid updating the loading of the png as backdrop is beyond my capabilities.

### DIFF
--- a/configs/apps/gladevcp/animated-backdrop/cairodraw.py
+++ b/configs/apps/gladevcp/animated-backdrop/cairodraw.py
@@ -1,5 +1,10 @@
-
-import gtk, sys, cairo
+#!/usr/bin/env python3
+import gi
+gi.require_version("Gtk","3.0")
+gi.require_version("Gdk","3.0")
+gi.require_version("GtkSource","3.0")
+from gi.repository import Gtk
+import sys, cairo
 from math import pi
 
 
@@ -49,8 +54,8 @@ class HandlerClass:
         # This sets the windows colormap, so it supports transparency.
         # This will only work if the wm support alpha channel
         screen = win.get_screen()
-        rgba = screen.get_rgba_colormap()
-        win.set_colormap(rgba)
+        rgba = screen.get_rgba_visual()
+        win.set_visual(rgba)
 
         # scaling the bitmap is possible by turning this on
         # however the fixed widget does NOT do proportional scaling
--- a/configs/apps/gladevcp/animated-backdrop/cairodraw.py
+++ b/configs/apps/gladevcp/animated-backdrop/cairodraw.py
@@ -1,10 +1,4 @@
-#!/usr/bin/env python3
-import gi
-gi.require_version("Gtk","3.0")
-gi.require_version("Gdk","3.0")
-gi.require_version("GtkSource","4")
-from gi.repository import Gtk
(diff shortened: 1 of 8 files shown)


Sigma1912:
Just realized that the move  to GtkSourceview4 has been merged.
I'll update my local repo and test my pull requests. Will report on results.

Sigma1912:
Oh, thanks.

Sigma1912:
Confirmed this PR works now with GtkSourceview4
Thanks for the help.

hansu:
>Note 'animated backdrop' window is transparent 

Maybe this is related to that error message:

```
  File "/home/cnc/linuxcnc/linuxcnc-2.9/configs/apps/gladevcp/animated-backdrop/cairodraw.py", line 18, in on_expose
    cr = widget.window.cairo_create()
         ^^^^^^^^^^^^^
AttributeError: 'Window' object has no attribute 'window'
```



However, the other looks fine, thanks for that!

Sigma1912:
Yes but that is just the start. I followed the trail and ended here (the handler runs without errors but the image does not show):

'''
#!/usr/bin/env python3
import gi
gi.require_version("Gtk","3.0")
gi.require_version("Gdk","3.0")
gi.require_version("GtkSource","4")
from gi.repository import Gtk
import sys, cairo
from math import pi


pngfile = 'vortex.me.png'

class HandlerClass:

    def on_window1_realize(self,widget,data=None):
        print("on_window1_realize")
        window = widget.get_window()
        cr = window.cairo_create()

        # Sets the operator to clear which deletes everything below
        # where an object is drawn
        cr.set_operator(cairo.OPERATOR_CLEAR)
        # Makes the mask fill the entire window
        cr.rectangle(0.0, 0.0, *widget.get_size())
        # Deletes everything in the window (since the compositing
        # operator is clear and mask fills the entire window
        cr.fill()
        # Set the compositing operator back to the default
        cr.set_operator(cairo.OPERATOR_OVER)

        if self.scale:
            w = widget.get_allocation().width
            h = widget.get_allocation().height
            cr.scale(1.0 *w / self.width, 1.0*h/self.height)

        cr.set_source_surface(self.img, 0, 0)
        cr.paint()



    def __init__(self, halcomp,builder,useropts):
        self.builder = builder
        win = self.builder.get_object("window1")
)
        #win.set_decorated(False)

        self.img = cairo.ImageSurface.create_from_png(pngfile)
        self.width = self.img.get_width()
        self.height = self.img.get_height()

        # Makes the window paintable, so we can draw directly on it
        win.set_app_paintable(True)
        win.set_size_request(self.width, self.height)

        # This sets the windows colormap, so it supports transparency.
        # This will only work if the wm support alpha channel
        screen = win.get_screen()
        rgba = screen.get_rgba_visual()
        win.set_visual(rgba)

        # scaling the bitmap is possible by turning this on
        # however the fixed widget does NOT do proportional scaling
        self.scale = 1

def get_handlers(halcomp,builder,useropts):

    return [HandlerClass(halcomp,builder,useropts)]

'''

Sigma1912:
Nice, looks easy now. Thanks for helping me with this.

PR #3092, opened 2024-08-29
Add note about required 'XTerm' to README 'configs/apps/xhc-hb04'

Sigma1912:
The bash script errors out in terminal apps other than 'XTerm'


petterreinholdtsen:
I suspect this change is ok for 2.9, and that the pull request should be rebased before being merged into 2.9.

PR #3093, opened 2024-08-29
Remove all README_es files from configs folder

Sigma1912:
It seems that the translation framework used for the documentation makes these translated README files obsolete. This would remove some of the burden when maintaining these example configs.


petterreinholdtsen:
[Sigma1912]
> It seems that the translation framework used for the documentation
> makes these translated README files obsolete. This would remove some
> of the burden when maintaining these example configs.

Note, most/all of these files are not listed in docs/po4a.cfg, and thus
not handled by the translation framework.  I never understood the
advantage of having these README files translated, so I never tried to
get them integrated.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
Furthermore I assume that they will not be visible to users running a spanish system as the config selection tool only picks the file with the name that matches the config file name.

Maybe those files should be sourced into the tool and handled as a string which ends up in a po file.

andypugh:
These spanish README translations came in as a large set of translations done before the weblate switch. 
In some cases they provided a translation that over-wrote the english version. 

It would be a nice touch if the config picker could read the locale information and show the appropriate file. 

I am not sure that I want to remove them, as they do represent some work done, that might need to be repeated.

Sigma1912:
Thanks for your prompt feedback.

The sim configs are quite handy to point new users to regarding questions that arise on the forum and also for experimentation. The README files in the example configs are very useful as they give a description of what the particular config is actually about and as a general place for notes about the topic covered in the simulation.

I'm not saying that these files need to be translated. I'm saying that IF it was felt that they need to be translated the new weblate system would seem to be a MUCH better approach that would not just cater to a single language (and I certainly don't think we should add more translated README files for other languages)

Generally I simply no longer see the point of having random translated files in the source code.  

Thanks for your attention

andypugh:
I agree with your points. And, on reflection, the file contents will remain available in Git anyway.

Issue #3094, opened 2024-08-29
2.9.3 Several broken sim configs in sim/gscreen


Sigma1912:
These configs in 'configs/sim/gscreen/' in Version 2.9.3 seem to be broken:

- all configs in the subfolder 'gscreen_custom/'
- 'industrial_lathe_wear'
- 'silverdragon'
- 'spartan'

If nobody is interested in fixing these then maybe they should be removed?

hansu:
There are a few broken sim configs. See https://github.com/LinuxCNC/linuxcnc/issues/1762

Sigma1912:
Thanks.

So these configs haven't worked for at least 2 years. I guess the question is what do we do with these broken things. Do we just simply drag them along for another few years or should the fact that nobody has bothered fixing them be taken as an indication that they should be removed?

Personally I find it a somewhat sad testament to the project as a whole if we present things on the config picker  that fail on load and have done so for several releases.

c-morley:
Looks like they are bit rotted from GTK3/GLADE changes over the years.
I looked at one and it's more work then I wish to expend. I'm not savvy with GTK anymore. 
Removing seems fine.

c-morley:
Oh I wonder if it matters how old a system one is using. I see someone mentions using silverdragon with 2.9.2 here:
https://github.com/LinuxCNC/linuxcnc/issues/2951

Sigma1912:
Thank you for your feedback.

I do all my testing on a recent Debian Bookworm installation with Linuxcnc 2.9.3 installed through the package manager and also on recent RIP installations. All the configs listed in this PR fail on load with gtk related errors.

Note that this quote from the linked page 'I runing simulate example sim.gscreen or sim.gscreen.silverdragon.' could simply mean that the user was not quite sure what he was running  as  the screenshot clearly shows 'gscreen.ini' which still works and is not among the configs I'm suggesting to remove.

Sigma1912:
The mentioned PR is here:
https://github.com/LinuxCNC/linuxcnc/pull/3096

Sigma1912:
Closing this as most have been fixed.

PR #3096, opened 2024-08-30
Remove some broken configs inside configs/sim/gscreen

Sigma1912:
This removes some of the simulation configs in the 'gscreen' folder. 
These have been broken and unmaintained for at least two years. Nobody, including the original authors, have shown any interest in maintaining these since.

The simulation configs included in LinuxCNC should showcase the capabilities of the application, which is now part of the debian repository, it seems a rather sad state of affairs to have unmaintained configs that have been failing to load over several releases.   
 
See also https://github.com/LinuxCNC/linuxcnc/issues/3094

Thank you.


andypugh:
Should probably remove the mention of Silverdragon here, too.
https://linuxcnc.org/docs/stable/html/gui/gscreen.html
(Though it looks like the screenshot shown is actually of QTdragon anyway)

Sigma1912:
Thanks for pointing that out. 

Looking at this again it seems that I might have got myself a somewhat bigger job then expected. 

1. It seems that really only the most basic configs in the main folder are still working:
    'gscreen.ini'
    'gscreen_lathe.ini'
    'gscreen_mm.ini 

2. Given that all other configs (with the possible exception of the three mentioned above) should really be removed there seems to be a question as to what should happen to all the files in 'share/gscreen/' and indeed also all the references in 'src/Makefile':

```
./src/Makefile:182:    emc/usr_intf/gremlin emc/usr_intf/gscreen emc/usr_intf/pyui emc/usr_intf/qtvcp \
./src/Makefile:681:		$(DESTDIR)$(datadir)/gscreen/images \
./src/Makefile:682:		$(DESTDIR)$(datadir)/gscreen/skins \
./src/Makefile:752:	$(FILE) ../share/linuxcnc/gscreen.glade $(DESTDIR)$(prefix)/share/linuxcnc
./src/Makefile:774:	$(DIR) $(DESTDIR)$(SITEPY)/gscreen
./src/Makefile:797:	$(FILE) ../lib/python/gscreen/*.py $(DESTDIR)$(SITEPY)/gscreen
./src/Makefile:829:	$(EXE) ../bin/gscreen $(DESTDIR)$(bindir)
./src/Makefile:840:	$(FILE) ../share/gscreen/images/*.gif $(DESTDIR)$(datadir)/gscreen/images
./src/Makefile:841:	$(TREE) ../share/gscreen/skins/* $(DESTDIR)$(datadir)/gscreen/skins
```

Given that really very little of gscreen is still in working order it seems to me that if a more thorough cleanup is desired then it may make a lot more sense to just remove all of gscreen now rather then leave bits and pieces that will probably remain in the source code for years to come.


So should we just remove the broken content in 'configs/sim/gscreen' and ignore the obsolete files in the background or should this be a more thorough cleanup?

Thanks

andypugh:
I have had a play around, and now have Silverdragon loading. 

This mainly involved loading the .glade file into the new version of Glade and re-saving targetting 3.16. 
Then working through the Gtk3 changes. 

Screenshot 2024-09-05 at 23 00 46

andypugh:
See https://github.com/LinuxCNC/linuxcnc/commit/6fe73e76a51be3d4a05aa006a354024042683435

andypugh:
What do you think about #3106 as an alternative?

Sigma1912:
Nice. If it's fixed then it doesn't need removing.
Does this fix all the configs in the 'configs/sim/gscreen' folder or does it need more work?
Thanks.

andypugh:
I have worked on all the fies in share/gscreen/skins (I think) and launched at least one config that uses each then homed it.
I haven't gone through them all, and I haven't done much testing of any of them. Party because I was working through ssh to my dev machine, rather than directly at the machine. A consequence of this is no open-gl so the graphical pane doesn't work. 
I still need to sit in front of the actual machine (buried under other projects) to check the varuous functions where I have had to make changes in the handler file. 

I can see why Chris shifted loyalties to QT. Gtk is awful. For example to set the DRO text colour some handlers create a [Pango attribute](https://github.com/LinuxCNC/linuxcnc/blob/master/share/gscreen/skins/silverdragon/silverdragon_handler.py#L1303) and insert that into the widget: 
But others use the [modify_fg](https://github.com/LinuxCNC/linuxcnc/blob/2.9/share/gscreen/skins/silverdragon/silverdragon_handler.py#L292) method: 
but modify_fg was deprecated and replaced by [override_color](https://github.com/LinuxCNC/linuxcnc/blob/andypugh/gscreen/share/gscreen/skins/industrial/industrial_handler.py#L231) in 3.0:  which looks the same, but takes a Gtk.RGBA as an argument, not a Gtk.Color. And then it turns out the override_color is itself deprecated in 3.16, with them now wanting you to use CSS for everything. Which is a really painful way to just change some text colour. It's hard to know what to use if you want the code to work after the next seemingly random update.

Sigma1912:
I'm sure it won't be very long before this has to be fixed again, that is why I lean towards removing things rather than investing a lot of time trying to fix it again and again. 

I'll try to checkout your branch and see how things look.

Sigma1912:
So I had a look at your branch. The result is rather mixed:

1. 'gscreen', 'gscreen_lathe' and 'gscreen_mm' fail on start 
2. the configs in 'gscreen_custom' starts with errors, the GUI is responsive and homing works.
4. 'industrial_lathe_wear' starts with errors, the GUI is responsive and homing works.
3. 'ngcgui' loads without errors but the 'gladevcp' window is empty
4. 'silverdragon' starts with errors, the GUI is responsive and homing works but there are runtime errors in the background
5. 'spartan' loads without errors (except 'Encoder Fault' for X,Y,Z) but homing doesn't work 

Not sure how much time you are willing to invest but given the likelihood of all this breaking again in the foreseeable future I'm not inclined to invest any time in fixing these bugs.

andypugh:
They all work for me, though with errors. (fewer now)
The ngcgui one works, with the ngcgui tab displaying as expected. 

I have cleaned up all but one of the silverdragon errors. And that's proving tricky:
````
(gscreen:938046): Gtk-WARNING **: 00:45:48.122: Don't know color '#27FFD8CFFF30D8FF27'
````
Obviously gscreen doesn't have that many lines, and the Hex number appears nowhere in the code. 
(I had some other errors due to old data in ~/.gscreen_preferences, but those were cleared by deleting the file.)

Sigma1912:
Closing this since these seem to be mostly working again.

Thanks for all the work.

Issue #3084, opened 2024-08-24
Cannot use a long list of keys with mqtt-publisher


ondras12345:
## Here are the steps I follow to reproduce the issue:

 1. Generate basic configuration with `stepconf`
 2. Add to `custom.hal`:
```
loadusr -W mqtt-publisher --mqtt-broker=[MQTT]BROKER --mqtt-prefix=[MQTT]PREFIX --mqtt-user=[MQTT]USERNAME --mqtt-password=[MQTT]PASSWORD keys=halui.estop.is-activated,halui.machine.is-on,halui.mode.is-auto,halui.mode.is-manual,halui.mode.is-mdi,halui.mode.is-teleop,halui.mode.is-mdi,halui.mode.is-manual,halui.program.is-idle,halui.program.is-paused,halui.program.is-running,halui.mist.is-on,halui.flood.is-on,halui.spindle.is-on
```
 3. Add to ini file:
```
# ...
[HAL]
# ...
HALUI = halui

# ...

[MQTT]
BROKER = somehost
PREFIX = linuxcnc
USERNAME = linuxcnc
PASSWORD = pass
```
 4. Run `linuxcnc path-to.ini` from terminal

## This is what I expected to happen:
All signals listed under `keys=` are published.

## This is what happened instead:
Published MQTT messages do not contain all signals:
```
"halui.estop.is-activated": false, "halui.machine.is-on": false, "halui.mode.is-auto": false, "halui.mode.is-manual": true, "halui.mode.is-mdi": false, "halui.mode.is-teleop": false, "halui.program.is-idle": true, "halui.program.is-paused": false, "halui.program.is-running": false, "mqtt-publisher.period": 10}
```
Warning is emitted on startup:
```
Preparing LinuxCNC2MQTT
warning: Missing pin halui.mist.is-o not sent to MQTT
info: Publishing MQTT message (linuxcnc): {"halui.estop.is-activated": false, "halui.machine.is-on": false, "halui.mode.is-auto": false, "halui.mode.is-manual": false, "halui.mode.is-mdi": false, "halui.mode.is-teleop": false, "halui.program.is-idle": false, "halui.program.is-paused": false, "halui.program.is-running": false, "mqtt-publisher.period": 10}
info: MQTT connected: Connection Accepted
```

It seems like only the first 255 characters of `keys=` are passed to `mqtt-publisher`:
```
$ ps aux | grep mqtt-publisher
[...]

PR #3098, opened 2024-08-31
Add workaround for long keys= in mqtt-publisher

ondras12345:
This fixes https://github.com/LinuxCNC/linuxcnc/issues/3084 (but I would prefer to see a better solution, this feels like an ugly workaround).


petterreinholdtsen:
[Ondřej Sluka]
> This fixes https://github.com/LinuxCNC/linuxcnc/issues/3084 (but I
> would prefer to see a better solution, this feels like an ugly
> workaround).

I agree, but it seem like an elegant solution to a ugly problem.  Note
there is a behaviour change in the code, if the same key is specified
several times.  I believe using key={}, duplicates would be removed,
with key=[] they will show up several times instead.  Perhaps duplicate
removal should be added?

Also, is there a warning somewhere if the .ini file contain lines longer
than 255 characters?  I guess there should be a popup letting the user
know that it is going to fail...

-- 
Happy hacking
Petter Reinholdtsen

ondras12345:
Previously, the initialization of `keys` with an empty dict would be unused as `keys` was reassigned with a new list:
https://github.com/LinuxCNC/linuxcnc/blob/69feb3b601b3e2d6cf07586cd42a0dfab03948b2/src/hal/user_comps/mqtt-publisher.py#L150
There should be no change in behavior regarding duplicates with my patch.

That being said, it might be a good idea to use a set instead of a list to drop duplicates (but that introduces a change in behavior).

The only warning I encountered was this:
```
warning: Missing pin halui.mist.is-o not sent to MQTT
```
It comes from the `mqtt-publisher` process and was only printed to my terminal.

andypugh:
I rebased this on 2.9 and committed as 317595efa0b800b537d4919a74cc8a5253590480
And, then in master I committed what I think is a more elegant fix: 9eb94c3bfb4e15724ecceac425cdb8409eb7f90a

I left in the ability to specify multiple keys= sections.

PR #3107, opened 2024-09-09
2.9 fix sim config manual toolchange with tool length switch

Sigma1912:
This sim config has been broken for a while now


Sigma1912:
I thought testing in Sid had been fixed:

 Failed: 
	./tests/mdi-queue/oword-queue-buster
	./tests/mdi-queue/simple-queue-buster

PR #3109, opened 2024-09-12
2.9 fix sim config stop lookahead

Sigma1912:
This sim config has been broken since 'G70' has been claimed. The included remap has been moved to 'G70.9'. Other minor updates to remove broken links and avoid 'abort' messages on startup.


andypugh:
Not a big thing, but G70.9 doesn't feel like the natural choice to me. 
If it's replication a G103, maybe G13 would work? It's in the middle of open space. 
http://linuxcnc.org/docs/stable/html/remap/remap.html#remap:remappable-codes
G13 is used for full circle moves on some controls, but LinuxCNC achieves that by just not specifying an end point. 
As I say, not a big thing, and we are pretty unlikely to have so many variant lathe finishing cycles that we need G70.9

Sigma1912:
Done, see https://github.com/LinuxCNC/linuxcnc/pull/3111
Closing this PR.
Thanks

Issue #2613, opened 2023-08-23
Gmoccapy - offset Z - wrong after tool 0 


zz912:
RIP 2.9 Branche
Debian Bookworm
Oracle VM VirtualBox
Test sim Gmoccapy.

If any tool other than 0 is selected at the start, the z offset is displayed correctly.
If I choose tool 0, the z offset is displayed incorrectly.
![Bug-tool-offset](https://github.com/LinuxCNC/linuxcnc/assets/96618597/2f65d8e1-322a-4896-a88c-4fdfdeaa293b)

rmu75:
Before selecting tool 0, G43 is active, after you select tool 0, G49 is active, i.e. tool compensation is off.

zz912:
So the question is why is G43 automatically disabled when tool 0 is selected?

zz912:
I did some research on this issue and found out that it is again a problem with the G43 auto-starting after a tool change.

https://github.com/LinuxCNC/linuxcnc/blob/c2b55957e91242ccddfb70c68bf0b59b783c19cd/src/emc/usr_intf/gmoccapy/gmoccapy.py#L3540-L3544

Norbert knows about this problem and plans to solve it. However, I will leave this Issue open so that we don't forget to check if his solution removes this consequence as well.

My solution is:
1) Remove lines auto-starting g43
2) Add the lines here:
https://github.com/LinuxCNC/linuxcnc/blob/c2b55957e91242ccddfb70c68bf0b59b783c19cd/src/emc/usr_intf/gmoccapy/gmoccapy.py#L4980-L4982
```
            self.command.wait_complete()
            self.command.mdi("G43")
            self.command.wait_complete()
```

rmu75:
I don't see how "tool offset == 0" can influence the code you cited.

zz912:
The cited code makes random problems. It makes cace conditions. I have been dealing with this for a long time.

lookt at:
https://github.com/LinuxCNC/linuxcnc/pull/2552

lookt at:
https://forum.linuxcnc.org/38-general-linuxcnc-questions/49445-python-command-wait-complete-does-not-wait-for-c-halui-commands

lookt at:
https://github.com/LinuxCNC/linuxcnc/issues/2453#issuecomment-1565086089

I would like ask you. Could you try reproduce this bug?  I wouldn't be surprised if I was the only one with the problem.

rmu75:
This issue may be related but I don't think this is caused by a race condition.

[...]

PR #3113, opened 2024-09-14
Gmoccapy - remove AUTOMATIC_G43

zz912:
Gmoccapy had automatic activation of tool offsets using the G43 command after a tool change. The automatic G43 caused "race condition problems" in some configurations. That's why it was removed. Automatic G43 should be implemented by LCNC remap functionality not GUI.

This PR fix many bugs:
https://github.com/LinuxCNC/linuxcnc/issues/2489
https://github.com/LinuxCNC/linuxcnc/issues/2613
......

Here rmu75 explain why automatic_G43 is not good in GUI:
https://github.com/LinuxCNC/linuxcnc/pull/2841#issuecomment-2275257199

Here Norbert allow me remove automatic_G43 from Gmoccapy:
https://github.com/LinuxCNC/linuxcnc/pull/2841#issuecomment-2325656940

Sorry that this PR contains a lot of commits. I can only use Github in a web browser.

Unfortunately, this patch produced, or rather pointed out, more bugs:
1) bug:
using M61 disable MDI window [SOLVED - this bug is not related with this PR]
2) bug:
remap broke these configurations: lathe_macros.ini, xyzac-trt.ini [SOLVED]

I would like ask for help.


Sigma1912:
Since the bug with M61 disabling MDI mode is not related to this PR I would suggest to remove this particular issue from the description of this pull request and file it in the 'Issues' list.
Thanks for working on this.

Sigma1912:
as for the broken configs, try these modifications:

in 'configs/sim/gmoccapy/lathe_configs/lathe_macros.ini'

replace this line:
SUBROUTINE_PATH = macros:./
with this:
SUBROUTINE_PATH = ../macros:./
 
in configs/sim/gmoccapy/non_trivial_kinematics/table-rotary-tilting/xyzac-trt.ini
remove these two lines:
SUBROUTINE_PATH = ./examples
SUBROUTINE_PATH = ../../macros
and replace with this:
SUBROUTINE_PATH = ./examples:../../macros

Sigma1912:
Could you edit the initial comment with the bug description and remove everything regarding the M61 bug and move that over to the issue you opened about this https://github.com/LinuxCNC/linuxcnc/issues/3120

This way it would be much clearer that the two are not related.

zz912:
After solving the bugs I will mark them [SOLVED]. I'll test your solution suggestions tonight. I hope it's clearer now.

Sigma1912:
My apologies, I was under the impression that the m61 bug was not related to your changes to the ini file.
Now I made a mess, I'm sorry.

Sigma1912:
Testing  'configs/sim/gmoccapy/lathe_configs/lathe_macros.ini' with my suggested ``` SUBROUTINE_PATH = ../macros:./ ```

I actually get errors:

```
[Gmoccapy][ERROR]  g-file-error-quark: Failed to open file “/home/user/git/linuxcnc-zz/configs/sim/gmoccapy/lathe_configs/macros/images/i_am_lost.png”: No such file or directory (4) (gmoccapy:803)
[Gmoccapy][ERROR]  could not resolve the image path '/home/user/git/linuxcnc-zz/configs/sim/gmoccapy/lathe_configs/macros/images/i_am_lost.png' given for button 'macro_0' (gmoccapy:805)
[Gmoccapy][ERROR]  g-file-error-quark: Failed to open file “/home/user/git/linuxcnc-zz/configs/sim/gmoccapy/lathe_configs/macros/images/goto_x_y_z.png”: No such file or directory (4) (gmoccapy:803)
[Gmoccapy][ERROR]  could not resolve the image path '/home/user/git/linuxcnc-zz/configs/sim/gmoccapy/lathe_configs/macros/images/goto_x_y_z.png' given for button 'macro_4' (gmoccapy:805)
```

the issue here is that the original   
``` SUBROUTINE_PATH = macros:./ ```
does actually not point to an existing 'macros' folder as that is actually located in the parent directory.

So this is really something that was already broken in all the configs inside a sub folder of 'configs/sim/gmoccapy'. In none of those configs can we use for example 'o<i_am_lost>call'.


Fixing this seems to be somewhat tricky because of this path that is used inside the macro ngc files:
```
(IMAGE, ./macros/images/i_am_lost.png)
```



So I would suggest to move your two new ngc files into a new 'subroutines' folder and then to add that to the 'SUBROUTINE_PATH' as either ':subroutines' or ':../subroutines' depending on whether the .ini file is located in the 'configs/sim/gmoccapy' folder or a subfolder.

new folder

**Comment by zz912** [comment]
Thank you for explain lathe_macros.ini bug.
BUT i prefer this solution:

Use SUBROUTINE_PATH = ../macros:./
and remove all [MACROS] in INI file.
**The behavior will be the same as before.**

I want these macros together:
macros

**Comment by zz912** [comment]
I have a question for @Hansa. I wanted to add a note about deleting AUTOMATIC_G43 here:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/usr_intf/gmoccapy/release_notes.txt

Currently the version of Gmoccapa is 3.4.8. Unfortunately, this designation is the same in both stable (2.9) and master (2.10). Will there be a change for Gmoccapa versioning? In the past, most PRs was merged into STABLE, so versioning was not a problem. But now there is pressure to have only BUGfix in STABLE. (I agree with this, but grey zone exist) So maybe it would also require a change in versioning, or its definition.

This PR is again an exception. This is a BUGfix, but it changes the behavior a lot, so it must belong to the MASTER branch.

**Comment by hansu** [comment]
We can increase the version number of gmoccapy for LinuxCNC 2.10 to 3.5.0.
But we definitely should avoid making changes in gmoccapy.glade in 2.9 AND master. Because the glade file would be hard to merge.

**Comment by zz912** [comment]
Can I ask you explain why? I am interesting.

**Comment by hansu** [comment]
Because the glade file is generated by the Glade Editor and sometimes it happens that the order changes, maybe by a version change or something different. and if you then change something within those lines ... good lock :)

File: src/emc/usr_intf/gmoccapy/release_notes.txt [hunk 1]
```diff
@@ -1,3 +1,9 @@
+ver 3.5.0
+ - version for 2.10 (master) branche
```
hansu, line 2:
```suggestion
 - version for 2.10 (master) branch
```

configs/sim/gmoccapy/gmoccapy.ini
```diff
@@ -54,7 +54,15 @@ py = python3
 [RS274NGC]
 RS274NGC_STARTUP_CODE = G17 G21 G40 G43H0 G54 G64P0.005 G80 G90 G94 G97 M5 M9
 PARAMETER_FILE = sim.var
-SUBROUTINE_PATH = macros
+SUBROUTINE_PATH = ./macros
```
hansu, line 57: Is this really needed?

zz912: In some INI files is ../macros or ../../macros needed. I wanted to unify it.

src/emc/usr_intf/gmoccapy/release_notes.txt
zz912: Thank you.

zz912:
> Because the glade file is generated by the Glade Editor and sometimes it happens that the order changes, maybe by a version change or something different. and if you then change something within those lines ... good luck :)

I think, that exist 2 ways:
1) same Gmoccapy.glady for 2.9 and 2.10 branche
2) make bugfix for 2.9 manualy, then make bugfix for 2.10 manualy. Dont merge from 2.9 to 2.10

Did you think about second way?

hansu:
> I think, that exist 2 ways:
> 
>1. same Gmoccapy.glade for 2.9 and 2.10 branche

That would prevent you from adding new graphical features to Gmoccapy, so no real option.

> 2. make bugfix for 2.9 manualy, then make bugfix for 2.10 manualy. Dont merge from 2.9 to 2.10

That's only the "emergency option" when merging is not really feasible.


We only have to keep an eye of the changes in the glade file before merging and try to fix the 2.9 glade file manually or make sure Glade does not change the order of elements.

zz912:
I am confused now.

HansU:
> But we definitely should avoid making changes in gmoccapy.glade in 2.9 AND master. Because the glade file would be hard to merge.

Then:

zz912:
> same Gmoccapy.glade for 2.9 and 2.10 branche

HansU:
>That would prevent you from adding new graphical features to Gmoccapy, so no real option.

What is the procedure for bugfix to 2.9 gmoccapy.glade?
What is the procedure for adding new graphical features to 2.9 gmoccapy.glade?

hansu:
Just keep the changes in the glade file in 2.9 small.

andypugh:
Try making _just_ the glade changes in (local) 2.9, then see if you can merge them into a local version of master. 
Then carefully check that the new glade file isn't messed up in some subtle way (that might be hard to be sure about)

gmoccapy:
Looks good for me.

Norbert

zz912:
I would like ask @andypugh for squeze.

zz912:
Thank you.

Issue #3120, opened 2024-09-16
2.10 Gmoccapy - remap M61 disable show MDI window


zz912:
I added to LCNC remap for M61 and I found out, that after using, Gmoccapy MDI window is disabled.

For more information and test configuration look here: https://github.com/LinuxCNC/linuxcnc/pull/3113

zz912:
![M61](https://github.com/user-attachments/assets/78e706f1-7c53-4bbf-a6be-df32cf291369)

Modification INI file:
```
[RS274NGC]
........
SUBROUTINE_PATH = ./macros
REMAP=M6  modalgroup=6 prolog=change_prolog ngc=change_g43 epilog=change_epilog
REMAP=M61  modalgroup=6 prolog=settool_prolog ngc=settool_g43 epilog=settool_epilog

# the Python plugins serves interpreter and task
[PYTHON]
PATH_PREPEND = ./python
TOPLEVEL = ./python/toplevel.py
LOG_LEVEL = 0
```
 needed file:
[macros_2024-09-16.zip](https://github.com/user-attachments/files/17013067/macros_2024-09-16.zip)
[python_2024-09-03.zip](https://github.com/user-attachments/files/17013069/python_2024-09-03.zip)

Sigma1912:
I can reproduce this. 
Toggling machine off then on seems to get things working again.

zz912:
Or you must change tool by M6 and it works again.

Sigma1912:
The odd thing is that even if I remove these lines from the ini I still get the bug:

'''
SUBROUTINE_PATH = ./macros
REMAP=M6  modalgroup=6 prolog=change_prolog ngc=change_g43 epilog=change_epilog
REMAP=M61  modalgroup=6 prolog=settool_prolog ngc=settool_g43 epilog=settool_epilog

# the Python plugins serves interpreter and task
[PYTHON]
PATH_PREPEND = ./python
TOPLEVEL = ./python/toplevel.py
LOG_LEVEL = 0
'''

Sigma1912:
Just did some investigating:

1. I just built current master on my test machine and I get this bug without any modification at all. (Gmoccapy version 3.4.8)
2. I modified an older installation of 2.9 with Gmoccapy version 3.4.7.1 and I don't get the bug. (but I notice that sometimes it does not switch to mdi the first time I click on the 'MDI' button)

So I would conclude that this is not related to your remap.

zz912:
Hi,

[...]

PR #3123, opened 2024-09-17
Gmoccapy - added more LOG.debug

zz912:
I am makeing it still again, again, again.
So I made PR.

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -2524,6 +2524,7 @@ class gmoccapy(object):
                 return
             self.reload_tool()
             self.command.mode(linuxcnc.MODE_MANUAL)
+            LOG.debug("MODE MANUAL was activated by all_homed")
 
     def on_hal_status_not_all_homed(self, widget, joints):
         LOG.debug("Hal Status not all homed {0}".format(joints))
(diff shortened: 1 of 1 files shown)


hansu:
I think this should better go into master

zz912:
I would like to ask you to consider merge it to branche 2.9.

Reasons:
- Gmoccapy in branch 2.9 still shows strange random behavior in the tool change area even with automatic G43 deleted. These notes might help other users in finding the source of the bug.
I have done a retrofit from version 2.9 and I don't want to go to version 2.10 if there is only need to find bugs (maybe my bugs, maybe LCNC bugs) related to ATC.
- branche 2.9 is still a STABLE version and is used for comparison when looking for bugs in other versions. Therefore, I want the same messages in 2.9 and in 2.10
For example here:
https://github.com/LinuxCNC/linuxcnc/issues/3120
- I don't think this PR is so dangerous that it can't be merged until 2.9

gmoccapy:
I agree with hansu,
This changes are to be merged to master.

- Changing of a behavior of a GUI and therefor the machine in a stable release is not good to do.

Norbert

zz912:
I added only LOG.debug for better description of bugs in 2.9. How did I changed behavior?

andypugh:
I am undecided on where this should go. I don't think that it actually changes the GUI behaviour, but at the same time it isn't _quite_ a bugfix. 

Does it have any effect on the "runtests" stage? Tests tend to compare terminal output. But given that the checks have passed (which include running the tests) I assume that it hasn't changed anything there. 

So, I have no real opinion, but don't _object_ to it going into 2.9.

gmoccapy:
After the "misbehavior" has been fixed all the debug output is not needed any more. 
I vote to not include the addinonal outputs at all.

But it is just my opinion and I have not checked the latest changes done in gmoccapy

zz912:
Software without bugs is just poorly tested software. :-D

PR #3125, opened 2024-09-19
Gmoccapy/gscreen: remove deprecation warning

hansu:
@andypugh  Into 2.9 or master?


andypugh:
Good for 2.9 I think.

PR #3130, opened 2024-09-23
gmoccapy: fix deprecation warning "Gtk.StyleContext.get_background_color is deprecated"

hansu:
A bit tricky one that removes the warning
`Gtk.StyleContext.get_background_color is deprecated`
and uses the correct color from the theme: "current-line"

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -4582,10 +4582,12 @@ class gmoccapy(object):
 
     def _set_sourceview_theme(self, name):
         self.widgets["gcode_view"].set_style_scheme(name)
-        style  = self.widgets["gcode_view"].get_style_context()
-        color = style.get_background_color(Gtk.StateFlags.SELECTED)
-        color.alpha = 0.5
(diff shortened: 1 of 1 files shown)


andypugh:
Gtk is just a nightmare isn't it? It's a moving traget that isn't even shaped like a target.

hansu:
>Gtk is just a nightmare isn't it? It's a moving traget that isn't even shaped like a target.

Exactly! You got to the heart of it.

c-morley:
As I understand it, glade will not suport gtk4. So more to come in the future :(

hansu:
Yeah, time to think about a QTmoccapy version.

Do you think we should better start from scratch or use QTDragon and reshape it?

andypugh:
Interesting read here: https://news.ycombinator.com/item?id=27783545
(Saying that Gtk3 has a stable API in 2021. Did it? I am sure we have been forced to change newer code than that)
There is a "new glade" which seems to be in active development, but doesn't seem to be in Debian. 
https://flathub.org/apps/ar.xjuan.Cambalache

hansu:
Interesting, thanks for the info.

PR #3131, opened 2024-09-24
Fix python path errors on startup in newly added trsrn sim configs

Sigma1912:
Looks like the build process messes with the python vismach simulation files in the config folders. I'm not sure why that happens but moving the files relevant to the vismach simulation models into a new subfolder inside the config folder seems to fix it.
Also adds a small descriptive comment inside the files about their purpose


Sigma1912:
I'd like to fix this sim config of mine. Could somebody merge this please.

Sigma1912:
Thanks, seems to work now.

Issue #3135, opened 2024-09-28
QTDragon HD: properties textTemplate and alt_textTemplate missing for lbl_tool_offset in qtdragon_hd.ui


Eckeneckepen:
## Here are the steps I follow to reproduce the issue:

 1. Start and home linuxcnc/qtdragon_hd
 2. In Tool List add a tool with diameter 123.123456 and tool offset 123.123456
 3. Change tool with button "M61 Qn"

## This is what I expected to happen:

Tool Status displays:
- DIAMETER: 123.123
- Z OFFSET: 123.123

## This is what happened instead:

Tool Status displays:
- DIAMETER: 123.123
- Z OFFSET: 123.12346

## It worked properly before this:
N/A

## Information about my hardware and software:

Simulation. LinuxCNC 2.9.3.

PR #3136, opened 2024-09-28
fix: https://github.com/LinuxCNC/linuxcnc/issues/3135

Eckeneckepen:
QTDragon HD: properties textTemplate and alt_textTemplate missing for lbl_tool_offset in qtdragon_hd.ui #3135


c-morley:
Thank you!

PR #3142, opened 2024-10-03
Change the gcode editor background when the focus changes. 

sfroome:
This merge request adds event handling for the focusIn and focusOut events of the gcode editor. If the editor is in focus, it sets the editor's background white. Otherwise, the background is set to lightgrey.

### DIFF
--- a/lib/python/qtvcp/widgets/gcode_editor.py
+++ b/lib/python/qtvcp/widgets/gcode_editor.py
@@ -634,6 +634,14 @@ class GcodeDisplay(EditorBase, _HalWidgetBase):
             STATUS.connect('interp_idle', lambda w: self.set_line_number(0))
         self.markerDeleteHandle(self.currentHandle)
 
+    def focusOutEvent(self, event):
+        self.setColorBackground(QColor(211, 211, 211))
+        super().focusOutEvent(event)
+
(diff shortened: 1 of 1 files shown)


andypugh:
Can you suggest a GUI to use that will show the old and new behaviour? I assume this is a qtvcp-only change?

c-morley:
It will show in any Qtvcp screen. The colors would need to be selectable. I'll think on this request.

hansu:
In Gmoccapy we can chose a theme for the G-code editor (Gtk SourceView) which controls the background color e.g.   
I wonder if there is something similar for Qt?

c-morley:
Yes, Qt has stylesheets, which can also manipulate arbitrary properties.
We would have to add two color properties to the widget.
But I think focus event changes are better set in the screen handler code, rather then in the widget it's self.

c-morley:
Thank you for the merge request.
In this case, I think this is best to do it in the handler file of the screen. Al the versions of Qtdragon use a similar technique to color borders when selecting MPG scrolling in master. This is where you could add code to color the background if wanted.
Please do not hesitate to comment if you think of problems with using the handler file or would like help in some way to code from the handler file.
I look forward to more suggestions/code.

Issue #3151, opened 2024-10-13
qtvcp uses obsolete(?) env variable EMC2_HOME instead of LINUXCNC_HOME


Eckeneckepen:
## Here are the steps I follow to reproduce the issue:

1.     git clone https://github.com/LinuxCNC/linuxcnc
2.     cd linuxcnc/src
3.     ./autogen.sh
4.     ./configure --with-realtime=uspace --prefix=/opt --enable-non-distributable=yes
5.     make -j 2
6.     make install
7.     /opt/bin/linuxcnc 
8.     Select sim.qtdragon_hd.qtdragon_hd_xyz

## This is what I expected to happen:
LinuxCNC starts normally.

## This is what happened instead:
qtvcp cannot find ui-files and qtdragon_hd_handler.py

## It worked properly before this:
n/a

## Information about my hardware and software:

KVM-virtual-machine, Debian 12, current master

Changing EMC2_HOME to LINUXCNC_HOME solves the problem for me.

andypugh:
Use of --prefix is somewhat deprecated. But while it exists it ought to work.

PR #3152, opened 2024-10-13
fix:  qtvcp uses obsolete(?) env variable EMC2_HOME instead of LINUXCNC_HOME #3151

Eckeneckepen:
Solves problem with relocation of LinuxCNC to /opt.


c-morley:
Thank you

PR #3156, opened 2024-10-19
docs: update some gitignores for new languages

phillc54:


.gitignore
```diff
@@ -26,12 +26,15 @@ src/modules.order
 /configs/*/emc.nml
 !/configs/common/emc.nml
 /docs/src/source-highlight/local
+docs/src/*/gui/gmoccapy_release_notes.txt
```
smoe, line 29: Why?

.gitignore
```diff
@@ -26,12 +26,15 @@ src/modules.order
...
 lib/python/gremlin.py
 lib/python/qt5_graphics.py
 lib/python/touchy
 configs/sim/gmoccapy/*.pref
 configs/sim/gmoccapy/gmoccapy_plasma/*.pref
 configs/sim/gmoccapy/gmoccapy-JA/*.pref
+configs/**/*.tbl.bak
```
smoe, line 36: Maybe even *.bak would make sense as a global rule?

.gitignore
phillc54: They are pre existing, I just moved them near more aligned rules as they looked quite silly following the commented line `# Ignore VSCode settings`

.gitignore
petterreinholdtsen: Perhaps, but that can wait for another patch.

PR #3155, opened 2024-10-16
fix: Commit fa2bfe0cd4eecd1d06d9ca548e4cb852aad67053 kills my build on openSUSE #3154

Eckeneckepen:
1. Interrupt configure if no path is found.
2. Call `sysconfig.get_path()` depending if scheme `deb_system` exists.

Issue #3154, opened 2024-10-16
Commit fa2bfe0cd4eecd1d06d9ca548e4cb852aad67053 kills my build on openSUSE


Eckeneckepen:
1. git clone https://github.com/LinuxCNC/linuxcnc
 2. cd linuxcnc/src
 3. ./autogen.sh
 4. ./configure --with-realtime=uspace --prefix=/opt --enable-non-distributable=yes

## This is what I expected to happen:
Configure should run with no error.

## This is what happened instead:
checking for site-package location... Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib64/python3.11/sysconfig.py", line 673, in get_path
    return get_paths(scheme, vars, expand)[name]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/sysconfig.py", line 663, in get_paths
    return _expand_vars(scheme, vars)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/sysconfig.py", line 319, in _expand_vars
    for key, value in _INSTALL_SCHEMES[scheme].items():
                      ~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 'deb_system'


## It worked properly before this:
Before commit fa2bfe0cd4eecd1d06d9ca548e4cb852aad67053.

## Information about my hardware and software:

 openSUSE Tumbleweed

PR #3158, opened 2024-10-19
Dropped use of depricated python3-setuptools / distutils.

petterreinholdtsen:
Make sure sysconfig.get_path("platlib") return correct path on Debian systems, where /usr/lib/ should be used over /usr/local/lib.

Based on e2c10a8ac20c37c231367d3731c79846082b3a53 by Andy Pugh,

Fixes Debian issue #1080668.

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1684,7 +1684,7 @@ fi
 AC_MSG_RESULT([$PYTHON_TK_VERSION])
 
 AC_MSG_CHECKING(for site-package location)
-SITEPY=`$PYTHON -c 'import distutils.sysconfig; print(distutils.sysconfig.get_python_lib())'`
+SITEPY="$($PYTHON -c 'import sysconfig; s = sysconfig.get_scheme_names(); m=list(set(("deb_system", "rpm_prefix")) & set(s)); print(sysconfig.get_path("platlib", m.__getitem__(0))) if m else print("/usr/lib/python3/dist-packages");')"
 AC_MSG_RESULT($SITEPY)
 
 AC_MSG_CHECKING(for working GLU quadrics)
--- a/src/m4/ax_python.m4
+++ b/src/m4/ax_python.m4
@@ -64,7 +64,7 @@ if test x$ax_python_bin != x; then
      AC_CHECK_LIB(${ax_python_bin}m, main, ax_python_lib=${ax_python_bin}m, ax_python_lib=no)
    fi
    if test x$ax_python_lib != xno; then
-     ax_python_header=`$ax_python_bin -c "from distutils.sysconfig import *; print(get_config_var('CONFINCLUDEPY'))"`
+     ax_python_header=`$ax_python_bin -c "import sysconfig; print(sysconfig.get_path('include'))"`
      if test x$ax_python_header != x; then
        break;
      fi


petterreinholdtsen:
See also #3154 and https://bugs.debian.org/1080668 .

smoe (approved):

petterreinholdtsen:
I do not understand why configure still report "checking for site-package location... /usr/lib/python3.9/dist-packages" on bullseye, when it should be reporting /usr/lib/python3/dist-packages after my regex rewrite.  When running the code in question manually on bullseye, the correct path show up.

andypugh:
Bullseye is not supported by the master branch any more, anyway. Though arguably this fix needs to go into 2.9 to ge tinto Trixie if Trixie is to lose distutils.

petterreinholdtsen:
[andypugh]
> Bullseye is not supported by the master branch any more,
> anyway. Though arguably this fix needs to go into 2.9 to get tinto
> Trixie if Trixie is to lose distutils.

Good point.  I'll try to remember to rebase on 2.9 when I am happy with
the change.  I do not like that I have no idea why the python oneliner
still fail on bullseye, and am thus reluctant to merge it anywhere.

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
> [andypugh]
> Bullseye is not supported by the master branch
> Good point. 

It wasn't a good point, I deleted it, but not fast enough. 

_Buster_ is not supported, Bullseye is still supported (and, in fact, current?) 

src/configure.ac
```diff
@@ -61,6 +61,17 @@ else
...
 AC_SUBST(AUTODIRS)
 
+# Switch sysconfig 'platlib' path from
+# /usr/local/lib/python3.11/dist-packages to
+# /usr/lib/python3/dist-packages, see
+# <URL: https://bugs.debian.org/1080668 >.  This affect SITEPY and
+# might also affect AX_PYTHON and AX_PYTHON_DEVEL.
+case "$(lsb_release -is)" in
+    Debian|Ubuntu)
+	export DEB_PYTHON_INSTALL_LAYOUT=deb
```
stefanor, line 71: Speaking with a Debian hat, without knowing anything about linuxcnc's build system:

This is appropriate for a `/usr` install, but not a `/usr/local` install.

I suggested that in the context of a Debian packaging patch, not something upstreamable.

src/m4/ax_python_devel.m4
```diff
@@ -186,34 +186,23 @@ variable to configure. See ``configure --help'' for reference.
...
 
-	if test $ax_python_devel_found = yes; then
-	   #
-	   # Check if you have distutils, else fail
-	   #
-	   AC_MSG_CHECKING([for the sysconfig Python package])
-	   ac_sysconfig_result=`$PYTHON -c "import sysconfig" 2>&1`
-	   if test $? -eq 0; then
+	#
+	# Check if you have sysconfig, else fail
```
stefanor, line 190: Note that upstream `ax_python_devel` has support for the Debian layout (see my patches from 2022):

https://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=history;f=m4/ax_python_devel.m4

src/configure.ac
```diff
@@ -1653,8 +1664,16 @@ fi
 AC_MSG_RESULT([$PYTHON_TK_VERSION])
 
 AC_MSG_CHECKING(for site-package location)
-SITEPY=`$PYTHON -c 'import distutils.sysconfig; print(distutils.sysconfig.get_python_lib())'`
+# The sed replacing python3.7 and similiar with python3 is to work
+# with buster, which return version specific path.
+# The sed replacing site-package to work with bullseye and older,
+# which return ../site-packages, not .../dist-packages.
+SITEPY="$($PYTHON -c 'import sysconfig; import re; p1=sysconfig.get_path("platlib"); print(p1); p2=p1.replace("/site-packages", "/dist-packages"); print(p2); print(re.sub("/python3[^/]+/","/python3/", p2))')"
```
stefanor, line 1671: Again, not knowing anything about linuxcnc, this seems to be completely unaware of PREFIX.

stefanor (commented):
Sorry, I don't have a whole lot of context for the package, abut I can comment on supporting installing on Debian's python.

Eckeneckepen:
Configure has the following result:
`2024-10-20T14:44:57.8858950Z checking for Python site-packages path... /usr/lib/python3.9/site-packages`
instead of the required `/usr/lib/python3.9/site-packages`

If you call your python statement, you get 3 rows. The last one is correct.  Configure seems to consider the first row only:
`python3 -c 'import sysconfig; import re; p1=sysconfig.get_path("platlib"); print(p1); p2=p1.replace("/site-packages", "/dist-packages"); print(p2); print(re.sub("/python3[^/]+/","/python3/", p2))'`
=>
```
/usr/lib/python3.9/site-packages
/usr/lib/python3.9/dist-packages
/usr/lib/python3/dist-packages
```

If you remove the unnecessary print()s it should work on Bullseye:
`python3 -c 'import sysconfig; import re; p1=sysconfig.get_path("platlib"); p2=p1.replace("/site-packages", "/dist-packages"); print(re.sub("/python3[^/]+/","/python3/", p2))'`
=>
`/usr/lib/python3/dist-packages`

But it won't work on Bookworm due to an additional "local":
`python3 -c 'import sysconfig; import re; p1=sysconfig.get_path("platlib"); p2=p1.replace("/site-packages", "/dist-packages"); print(re.sub("/python3[^/]+/","/python3/", p2))'`
=>
`/usr/local/lib/python3/dist-packages`

Possible solution, but then you can also hardcode it directly:
`python3 -c 'import sysconfig; import re; p1=sysconfig.get_path("platlib"); p2=p1.replace("/site-packages", "/dist-packages"); p3=re.sub("/python3[^/]+/","/python3/", p2); print(p3.replace("local/", ""))'`

dist-packages is only used by Debian and its derivates and not by other distributions.
This solution is incompatible to non-Debian distributions.

Why not using my proposal?
https://github.com/LinuxCNC/linuxcnc/pull/3155

It hard codes the old Bullseye only. This could be replaced by 
`python3 -c 'import sysconfig; import re; p1=sysconfig.get_path("platlib"); p2=p1.replace("/site-packages", "/dist-packages"); print(re.sub("/python3[^/]+/","/python3/", p2))'`

src/m4/ax_python_devel.m4
petterreinholdtsen: It seem to me that both upstream ax_python.m4 and ax_python_devel.m4 uses the distutils python module we are trying to avoid.  I am thus reluctant to just upgrade to the latest upstream edition.

src/configure.ac
petterreinholdtsen: I changed the approach and moved the environment variable to debian/rules(.in) instead.

src/configure.ac
petterreinholdtsen: Indeed.  I changed the approach to avoid these string replacements.

petterreinholdtsen:
See #3159 for new upstream editions of the ax_python*.m4 files.

src/m4/ax_python_devel.m4
petterreinholdtsen: Looking closer, I see that the distutils part is only a fallback, and the new version can be used without modifications.

PR #3166, opened 2024-10-26
Fix and update remap documentation regarding remappable Mcodes

Sigma1912:
The current documentation on remapping has not been reflecting the situation in the source code for some time.
Note: This includes the changes made in PR #3165


zz912:
I think it would also like to expand this chapter:
https://linuxcnc.org/docs/devel/html/remap/remap.html#_remapping_other_existing_codes_s_m0_m1_m60

Sigma1912:
Thanks for the comment, I cleaned that section up and added a small paragraph regarding M7/M8/M9

PR #3167, opened 2024-10-26
Remove references to removed 'iocontrolv2'

Sigma1912:
'iocontrol-v2' has been removed with this commit
https://github.com/LinuxCNC/linuxcnc/pull/2497/commits/9d79c2aa23e6e6762dfc32b4ad7c4b7d219e9ea4


Sigma1912:
There are some other references to iocontrol-v2. I'm somewhat unclear about some if them (eg parameters 5600, 5601 or the comment in the various 'stdglue.py' files):

```
./docs/src/remap/remap.adoc:1583:* The following three steps are applicable only if the `iocontrol-v2` component is used:
./docs/src/remap/remap.adoc:1600:* If parameter 5600 (fault indicator) is greater than zero, this indicates a Toolchanger fault, which is handled as follows (`iocontrol-v2`-only):
./docs/src/remap/remap.adoc:1774:Note that the use of M61 currently requires the use of `iocontrol-v2`.
./docs/src/remap/remap.adoc:2228:. M61 (remapped or not) is broken in `iocontrol` and requires `iocontrol-v2` to actually work.
./docs/src/gcode/overview.adoc:342:* '5600' - Toolchanger fault indicator. Used with the iocontrol-v2 component.
./docs/src/gcode/overview.adoc:344:* '5601' - Toolchanger fault code. Used with the iocontrol-v2 component.
./src/emc/rs274ngc/interp_parameter_def.hh:201:// 5600 - Toolchanger fault indicator. Used with the iocontrol-v2 component. 1: toolchanger faulted, 0: normal. Volatile.
./src/emc/rs274ngc/interp_parameter_def.hh:203:// 5601 - Toolchanger fault code. Used with the iocontrol-v2 component. Reflects the value of the toolchanger-reason HAL pin if a fault occurred. Volatile.
./configs/sim/gmoccapy/python/stdglue.py:59:        # this is relevant only when using iocontrol-v2.
./configs/sim/woodpecker/python/stdglue.py:170:        # this is relevant only when using iocontrol-v2.
./configs/sim/woodpecker/python/stdglue.py:199:        # this is relevant only when using iocontrol-v2.
./configs/sim/axis/remap/extend-builtins/README:16:Note that using M61 requires iocontrol-v2 as set tool number support in iocontrol is broken.
./nc_files/remap_lib/python-stdglue/stdglue.py:170:        # this is relevant only when using iocontrol-v2.
./nc_files/remap_lib/python-stdglue/stdglue.py:199:        # this is relevant only when using iocontrol-v2.
```

hansu:
>'iocontrol-v2' has been removed with this commit
https://github.com/LinuxCNC/linuxcnc/commit/9d79c2aa23e6e6762dfc32b4ad7c4b7d219e9ea4

And with this PR: https://github.com/LinuxCNC/linuxcnc/pull/2497

Not only iocontrol v2 also iocontrol.

Then I wonder if the EMCIO parameter can be removed from all the sample configs?
```
[EMCIO]
EMCIO = 		io
```

Furthermore there are also the man pages of io and iov2 remaining. Can be removed as well?

rene-dev:
yes, can all be removed.

rene-dev:
ah, sorry no. the man pages should remain, as the pins are still there.

Sigma1912:
```
ah, sorry no. the man pages should remain, as the pins are still there.
```

I presume you mean the manpage for 'io' should remain and the one for 'iov2' can be removed?

Sigma1912:
So it looks like I'm done (I assume that the files in the '/docs/po' folder are updated automatically)

rene-dev:
> I presume you mean the manpage for 'io' should remain and the one for 'iov2' can be removed?

yes

hansu:
It seems that something went wrong with your force push.
And also note the conflicts.

andypugh:
Even though we don't have iov2 any more, I think we still support alternative io modules (if someone wants to write their own)
So, do we really want to remove that line from all the sample .ini files?

Sigma1912:
Not sure we do still support alternative io modules as the 'EMCIO' entry seems to be ignored no matter what its value is.

PR #3172, opened 2024-11-03
Fix Python syntax warning

havardAasen:
- Change deprecated and removed `locale.format()` to `locale.format_string()`
 - Change to raw string to fix Python SyntaxWarning, this also addresses Debian bug 1085669
 - Update a regex in gcode_ripper


andypugh:
Thanks for this, #1085669 has been stuck in my inbox awaiting attention for far too long,

Issue #3174, opened 2024-11-10
interp error in 2.9.3, started somewhere between 2.9.3-83-g1d5d02d836 and  2.9.3-2099-g9f5a1c8f50


Lcvette:
## Here are the steps I follow to reproduce the issue:

 1. start axis sim from terminal 
 2. look for USRMOT: ERROR  command 30 timeout and emcMotionInit:  emcTrajInit failed
 3. load the tool table with tools for a program and then load the program, on screen error message occurs: "parse_file interp_error"
 4. Terminal error shows as:
 
 py3dev@py3dev:~/linuxcnc/configs/sim.axis$ linuxcnc axis.ini

LINUXCNC - 2.9.3-2099-g9f5a1c8f50

Machine configuration directory is '/home/py3dev/linuxcnc/configs/sim.axis'

Machine configuration file is 'axis.ini'

Starting LinuxCNC...

linuxcncsvr (8811) emcsvr: machine 'LinuxCNC-HAL-SIM-AXIS'  version '1.1'

linuxcnc TPMOD=tpmod HOMEMOD=homemod EMCMOT=motmod

Note: Using POSIX realtime

milltask (8824) task: machine 'LinuxCNC-HAL-SIM-AXIS'  version '1.1'

halui (8826) halui: machine 'LinuxCNC-HAL-SIM-AXIS'  version '1.1'

Found file(lib): /usr/share/linuxcnc/hallib/core_sim.hal

Found file(lib): /usr/share/linuxcnc/hallib/sim_spindle_encoder.hal

Found file(lib): /usr/share/linuxcnc/hallib/axis_manualtoolchange.hal

Found file(lib): /usr/share/linuxcnc/hallib/simulated_home.hal

Found file(lib): /usr/share/linuxcnc/hallib/check_xyz_constraints.hal

Found file(REL): ./cooling.hal

USRMOT: ERROR: command 30 timeout

emcMotionInit: emcTrajInit failed

note: MAXV     max: 5.000 units/sec 300.000 units/min

note: LJOG     max: 5.000 units/sec 300.000 units/min

note: LJOG default: 0.250 units/sec 15.000 units/min

note: jog_order='XYZ'

note: jog_invert=set()

!!!emc/rs274ngc/gcodemodule.cc: parse_file() f=/home/py3dev/linuxcnc/nc_files/aa_error_test.ngc

!!!interp_error=1 result=2 last_sequence_number=16

[...]

PR #3176, opened 2024-11-11
fix tool parameter mismatch in StatMixin

rmu75:
StatMixin is called with tool number instead of tool table index since refactoring in https://github.com/LinuxCNC/linuxcnc/commit/0ab84a6ab7aef8f14a396fdb2296631cb2fa888d, so we need to search for tool in tool table


rmu75:
fixes #3174

Sigma1912:
Works for me. Thanks for the fix.

PR #3181, opened 2024-11-15
Gmoccapy duplicated settings

hansu:
In some cases, especially when switching between 2.9 and 2.10, the gmoccapy preference file gets duplicate entries.
Here a fix.

For more details: https://www.forum.linuxcnc.org/gmoccapy/50371-duplicate-settings-in-pref-file?start=20#314578


hansu:
Closed in favor of #3269

PR #3182, opened 2024-11-16
emcrsh: fix bug with extra reply line for 'get spindle_override'

Sigma1912:
currently 'get spindle_override' command in linuxcncrsh will return an extra line containing the reply from the 'get' command before last:

```
get teleop_enable
TELEOP_ENABLE OFF
get spindle_override
SPINDLE_OVERRIDE 0 49.000000
TELEOP_ENABLE OFF

```

I'm not sure why that happens but using 'OUT' instead of 'dprintf' seems to fix this issue.


Sigma1912:
Ready, as far as I'm concerned

PR #3189, opened 2024-11-22
Gardenweazel-branch

6XoCtujg2C0gne:
Was:  self.buildToolTip(self.input_max_travel, 'Maximum Probe Search Distance', 'rappid')


c-morley:
Thank you.

PR #3194, opened 2024-11-27
qt_istat: remove "Error" from logger strings

hansu:
A warning log message should not include the string "Error". And for critical log messages it is redundant.

### DIFF
--- a/lib/python/qtvcp/qt_istat.py
+++ b/lib/python/qtvcp/qt_istat.py
@@ -402,7 +402,7 @@ class _IStat(object):
         self.MAX_FEED_OVERRIDE = float(self.get_error_safe_setting("DISPLAY", "MAX_FEED_OVERRIDE", 1.5)) * 100
         if self.INI.find("TRAJ", "MAX_LINEAR_VELOCITY") is None:
             if self.LINUXCNC_IS_RUNNING:
-                log.critical('INI Parsing Error, No MAX_LINEAR_VELOCITY Entry in TRAJ')
+                log.critical('INI Parsing: No MAX_LINEAR_VELOCITY Entry in TRAJ')
         self.MAX_TRAJ_VELOCITY = float(self.get_error_safe_setting("TRAJ", "MAX_LINEAR_VELOCITY",
                                             self.get_error_safe_setting("AXIS_X", "MAX_VELOCITY", 5))) * 60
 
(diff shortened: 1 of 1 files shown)


c-morley:
What are you trying to fix?
The word 'error' was the correct use.
for instance:
'warning: Valid Extension Parsing Error' is the opposite of  'warning: Valid Extension Parsing'

hansu:
>'warning: Valid Extension Parsing Error' is the opposite of 'warning: Valid Extension Parsing'

Oh yes in this case you are totally right as well as in some further. That's why I assigned you...

My starting point was this log message:
```
[Gmoccapy.QTVCP.QT_ISTAT][WARNING]  INI Parsing Error, No MAX_SPINDLE_0_SPEED Entry in DISPLAY, Using: 2500 (qt_istat.py:532)
```
And for this, the "Error" does not really fit into a warning.

So IMHO either
```
[Gmoccapy.QTVCP.QT_ISTAT][WARNING]  INI Parsing: No MAX_SPINDLE_0_SPEED Entry in DISPLAY, Using: 2500 (qt_istat.py:532)
```

or
```
[Gmoccapy.QTVCP.QT_ISTAT][ERROR]  INI Parsing (Error), No MAX_SPINDLE_0_SPEED Entry in DISPLAY, Using: 2500 (qt_istat.py:532)
```

But for most of the others, the pattern is different and simply removing the "Error" does not work.
So feel free to revert this commit...

c-morley:
I'm just interested in what the best message for users. It didn't occur to me that a warning with the word error would be confusing or does it just 'feel' wrong?

This is how I intended it (I think):
'Warning (not super important): Syntax error in the parsing of INI file, missing something'

I would say I have been pretty loose in the definition of message levels.
The levels available are: VERBOSE, DEBUG, INFO, WARNING, ERROR, CRITICAL
Maybe we could roughly define them to help with consistency?

hansu:
>This is how I intended it (I think):
'Warning (not super important): Syntax error in the parsing of INI file, missing something'

I agree if it is a warning about a syntax error, but if it's just a message about a missing entry, then I would tend to 
` log.warning('INI Parsing: No {} Entry in {}, Using: {}'.format(detail, heading, default))`

So I would suggest to revert everything of this PR except that line - are you fine with that?


> Maybe we could roughly define them to help with consistency?

Yes, we can add it to your document https://github.com/LinuxCNC/linuxcnc/blob/gui-reference-docs/docs/src/gui/gui-dev-reference.adoc

andypugh:
Is there a way to revert from the Github interface?

hansu:
There is a revert button a bit up, but not sure if this is better than a manual revert...
grafik

Issue #3185, opened 2024-11-18
LinuxCNC and musl C library (possible?)


NTULINUX:
## Here are the steps I follow to reproduce the issue:

 1. Use a Clang 19 system with musl
 2. Download and attempt to build LinuxCNC
 3. Wait for errors

## This is what I expected to happen:

A proper compilation.

## This is what happened instead:

A few errors, maybe an easy fix?

## It worked properly before this:

Clang 16 and glibc built fine when I last tested this.

## Information about my hardware and software:

 * I am using Gentoo

 * I am using this kernel version (shown by `uname -a`): Linux ntu-gentoo 6.12.0-rc7-BUILD #2 SMP Mon Nov 11 10:20:31 CST 2024 x86_64 AMD Ryzen 7 2700X Eight-Core Processor AuthenticAMD GNU/Linux

 * I am running ...
   * [ ] A binary version from linuxcnc.org (including buildbot.linuxcnc.org)
   * [X] A binary I built myself
   * [ ] A binary version from some other source besides linuxcnc.org

 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`): latest 2.9 git checkout

NTULINUX:
These are the errors:

`Done reading realtime dependencies
Compiling libnml/buffer/tcpmem.cc
libnml/buffer/tcpmem.cc:108:46: error: use of undeclared identifier 'u_short'; did you mean 'short'?
  108 |     server_socket_address.sin_port = htons(((u_short) tcp_port_number));
      |                                              ^~~~~~~
      |                                              short
1 error generated.`

`rtapi/uspace_rtapi_app.cc:699:18: error: use of undeclared identifier 'M_TRIM_THRESHOLD'
  699 |     if (!mallopt(M_TRIM_THRESHOLD, -1)) {
      |                  ^
rtapi/uspace_rtapi_app.cc:704:18: error: use of undeclared identifier 'M_MMAP_MAX'
  704 |     if (!mallopt(M_MMAP_MAX, 0)) {
      |                  ^
rtapi/uspace_rtapi_app.cc:1053:21: error: use of undeclared identifier 'pthread_attr_setaffinity_np'; did you mean 'pthread_setaffinity_np'?
 1053 |           if((ret = pthread_attr_setaffinity_np(&attr, sizeof(cpuset), &cpuset)) != 0)
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #3206, opened 2024-12-02
Fix build on musl

NTULINUX:
While probably not perfect; this patch series allows LinuxCNC to be successfully compiled on a musl system.

### DIFF
--- a/src/emc/rs274ngc/interp_o_word.cc
+++ b/src/emc/rs274ngc/interp_o_word.cc
@@ -526,6 +526,15 @@ void Interp::loop_to_beginning(setup_pointer settings)
     settings->sequence_number = 0;
 }
 
+// Prevent any possibility of the Posix version of basename(3) being used
+// (possibly altering the argument). Therefore, local reimplementation
+// of the *GNU version* of basename(3).
+static const char *thebasename(const char *filename)
+{
+     const char *p = strrchr(filename, '/');
+     return p ? p + 1 : filename;
+}
+
 //
 // TESTME!!! MORE THOROUGHLY !!!KL
 //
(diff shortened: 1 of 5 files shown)


NTULINUX:
Please see #3185 for discussion on the changes.

BsAtHome:
The call to pthread_setaffinity_np() is wrong. It is performed before the thread is created by pthread_create(). The thread's ID is unknown at that time. In my earlier description, I wrote that such change should go after creating the thread (see https://github.com/LinuxCNC/linuxcnc/issues/3185#issuecomment-2500671710).

Also, using pthread_setaffinity_np() must be conditional and only be used if pthread_attr_setaffinity_np() is not available. Therefore, you need to have a autoconf check.

NTULINUX:
I never wrote an autoconf test before for a C function but I'll try :) I have some experience with autotools and I think there's a few places I know to look for example code.

andypugh:
Any updates to this? I think I saw some activity on a related thread?

BsAtHome:
Apparently not yet.

However, I think that splitting this PR into parts is the proper way to proceed.

There are two includes added, a fixed wrong freopen() call and a local basename() version. These are harmless changes and can be added. These may be fixed in a (new) PR.

The last part, the thread affinity, that is a different problem. The suggested change is wrong as described in https://github.com/LinuxCNC/linuxcnc/pull/3206#issuecomment-2512265551. It will require more work to get that right.

There has been some overlap in the discussion because it identified a bug due to optimization with clang, which got fixed in PR #3261.

NTULINUX:
I'm working on a QEMU image for Gentoo with musl+clang and a working desktop environment if anyone wants to contribute to this.

BsAtHome:
There are still problems with the musl library that cannot be solved easily. There are missing interfaces/functions for the RT interface to work properly. The thread issue could be implemented with creating our own trampoline (with quite a bit of effort), but the malloc/page-lock issue cannot. See also #3185.

NTULINUX:
OK, got it! Will scratch VM.

PR #3208, opened 2024-12-05
Update ci.yml - introducing caching of Ubuntu packages

smoe:
I admit not to fully know what I am doing here. What this is meant to do is to avoid all those frequent downloads of the very same Ubuntu packages. This seems to have failed many times, causing the CI to indicate a false positive failure. I have no idea about how close this attempt to have the files in /var/cache/apt indeed cached will work.


smoe:
No, the parallel access I think troubles it.

petterreinholdtsen:
If it is any comfort, I do not know what you are doing either. :)

PR #3209, opened 2024-12-07
New unified hostmot2 SPI driver hm2_spix

BsAtHome:
This is the new unified SPI driver for the Mesa cards supporting hostmot2 over SPI. This driver currently supports Raspberry Pi 3, 4 and 5 natively, including compute modules 4 and 5, and will automatically fall back to the kernel's spidev interface if an unsupported board is found. The driver is written for easy addition of new native hardware drivers like future RPi boards or existing other "fruity" pi boards.

The new driver also includes fixes for SPI1 on the RPi3 and RPi4 boards, which was non-functional with the previous driver.

The new hm2_spix driver is easily back-ported if required.
The code body includes some separated source files for future de-duplicating of some routines and constants, which other HM2 drivers should make use of later on. A second PR will be created after integration of this driver to streamline and implement code de-duplication for the other drivers(s).


andypugh:
Thanks. 
If it was just a new driver I would be looking to port it to 2.9, but it touches lots of other files in hm2 and that makes me a little nervous.

BsAtHome:
> If it was just a new driver I would be looking to port it to 2.9, but it touches lots of other files in hm2 and that makes me a little nervous.

Actually, it only touches two files from the previous driver: hm2_rpspi.c and spi_common_rpspi.h.
The changes in hm2_rpspi.c are moving the register read/write macros (always inlined function) into a header hwregaccess.h. The changes in spi_common_rpspi.h are moving the __I, __O and __IO macros into the hwregaccess.h header.

A backport to 2.9 would change exactly the same thing as this PR did in HEAD.

Now, HEAD should evolve a bit further, where a cleanup is in order, which I mentioned in the text accompanying this PR. None of the evolution needs to be backported for the hm2_spix driver to function in 2.9.

ozzyrob:
Wow Thank You Sir.
Very much appreciated.

PR #3218, opened 2024-12-10
fix race condition in single stepping

### LINKED ISSUE #3213: Test single-step fails randomly
#### Here are the steps I follow to reproduce the issue:

 1. Get master branch and compile
 2. run from the shell while standing in tests directory:
 ```sh
 while ../scripts/rip-environment runtests -p -s single-step; do true; done
```
 3. Wait until it fails (maybe do some other things on the machine while running)

#### This is what I expected to happen:

No failures, at all.

#### This is what happened instead:

Random fails. Sometimes it runs 1, 2, 3, 4, 5, 6,... times before failure. However, every instance so far failed eventually. The point of failure in the ngc-file  seems to vary too (see attached logs).

#### It worked properly before this:

Unknown.

A CI failure after commit https://github.com/LinuxCNC/linuxcnc/commit/697fe4d326eae2989087a86d6857305b229198ed was suspicious. It turned out to be a failure in the single-step test, which I'd seen several times on my local machine.

Looking into the failure: it is always a timeout in tests/single-step/test-ui.py in function wait_complete_step(). Adding some debug messages seems to indicate that it always times out while the state is in <code>linuxcnc.EXEC_WAITING_FOR_MOTION_AND_IO</code> (integer 7).

The fact that it fails randomly seems to indicate a race-condition, somewhere.

[single-step-failure.zip](https://github.com/user-attachments/files/18061319/single-step-failure.zip)

rmu75:
fixes #3213

### DIFF
--- a/src/emc/task/emctask.cc
+++ b/src/emc/task/emctask.cc
@@ -718,7 +718,7 @@ int emcTaskUpdate(EMC_TASK_STAT * stat)
 					emcStatus->motion.traj.tag);
 	if (res_state != INTERP_ERROR) {
             rtapi_strxcpy(stat->file, emcStatus->motion.traj.tag.filename);
-            stat->currentLine = emcStatus->motion.traj.tag.fields[0];
+        //    stat->currentLine = emcStatus->motion.traj.tag.fields[0];
 	}
     } 
     // If we get an error from trying to unpack from the motion state, always


andypugh:
I would like to see at least an explanatory comment about why that line exists in a commented-out state.

PR #3221, opened 2024-12-12
Fix classicladder warnings

BsAtHome:
There are several warning in the classicladder component. The ones involving string copy operations are quite serious because the used buffer sizes are wrong.

Other warning deal with missing checks on return values of fgets() and write(). The fgets() problem is solved quite easily. Fixing the return value check of write() on a serial port requires some extra code because there may be partial and restartable writes.

### DIFF
--- a/src/hal/classicladder/drawing.c
+++ b/src/hal/classicladder/drawing.c
@@ -52,14 +52,14 @@ int PrintRightMarginNumExpr = 0;
 int PrintRightMarginPosiX;
 int PrintRightMarginWidth;
 
-void CreateVarNameForElement( char * pBuffToWrite, StrElement * pElem, char SymbolsVarsNamesIfAvail )
+void CreateVarNameForElement( char * pBuffToWrite, size_t BuffSize, StrElement * pElem, char SymbolsVarsNamesIfAvail )
 {
 	char VarIndexBuffer[20];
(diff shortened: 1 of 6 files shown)


smoe (approved):

smoe (approved):

src/hal/classicladder/drawing.c
```diff
@@ -215,11 +215,12 @@ char * DrawExprForCompareOperate( cairo_t * cr, int BaseX, int BaseY, int Width,
 		if ( DrawingOption==DRAW_FOR_PRINT )
 		{
 			// print the expression entirely in the right margin
-			char * pReportNumAndText = malloc( strlen(Text)+10 );
+			size_t TextSize;
+			char * pReportNumAndText = malloc( TextSize = strlen(Text)+10 );
```
smoe, line 219: Where is this freed again?

BsAtHome:
> The original code feels a bit weird. Error management anyone? But that is not ours, or is it?

That is an understatement.

If you diff the LCNC version with the current upstream, then you get a 21000+ line diff. If you try to <i>compile</i> the upstream version, then you get about 6000+ lines of warnings. And that includes a lot of deprecation warnings.

> I liked what I saw, but it is a memory leak, right?

No memory leak identified. The malloc in question is freed a little further on.

PR #3226, opened 2024-12-14
Backport fix hm2_spi driver

BsAtHome:
A flaw in hm2_spi was discovered in master and fixed in PR #3225. This simply backports that fix to 2.9.


andypugh:
Looks reasonable to me. (though I was initially a bit surprised by the single mention of delay_usecs, then found it in the spidev.h file)

Issue #3202, opened 2024-12-01
Superfluous NULL check


BsAtHome:
There is a warning-generating check on a pointer CMS::BufferName that is not a real pointer, but a character array declared in the CMS class:
<code>libnml/nml/nmldiag.cc:134 col 26: warning: the address of ‘CMS::BufferName’ will never be NULL [-Waddress]</code>

The check will always be true and is superfluous. The use of the buffer in a print function should be fine because it is initialized to zeros by the constructor.

```diff
diff --git a/src/libnml/nml/nmldiag.cc b/src/libnml/nml/nmldiag.cc
index 1eec85bf0a..fd5bf203a6 100644
--- a/src/libnml/nml/nmldiag.cc
+++ b/src/libnml/nml/nmldiag.cc
@@ -131,9 +131,7 @@ int nml_print_diag_list()
                }
                rcs_print
                    ("\n*********************************************\n");
-               if (NULL != nml->cms->BufferName) {
-                   rcs_print("* Buffer Name: %s\n", nml->cms->BufferName);
-               }
+               rcs_print("* Buffer Name: %s\n", nml->cms->BufferName);
                NML_DIAGNOSTICS_INFO *ndi = nml->get_diagnostics_info();
                if (NULL != ndi) {
                    ndi->print();
```

smoe:
I agree. The BufferName (declared at https://github.com/LinuxCNC/linuxcnc/blob/1ed894e830e871ad6f3f84c3213a1cd7dd1e92cc/src/libnml/cms/cms.hh#L354) is a char[] of the the CMS class.

@BsAtHome, the way to proceed, so I presume, is a PR. Would you be up for it?

BsAtHome:
Sure, a PR would be possible. I reported several issues, #3200, #3201, #3202 and #3204, which are all rather trivial changes. For a PR I'd would want to put all of them in one PR.

Issue #3200, opened 2024-12-01
Printf strings with single %


BsAtHome:
A translation text used in a printf-style format uses a single '%' and is missing a proper format specifier. It looks, from the context, that a literal percent sign is meant to be printed.

A fix is to change the format to '%%', to print the percent sign.

That said, this is a change in a translatable string, which has consequences for all translations. I have not looked at the i18n and l10n procedures, so I have no idea how to fix it for all languages. Someone with knowledge of the translation procedure can use the diff below and then fix all translations.

```diff
diff --git a/src/emc/rs274ngc/rs274ngc_return.hh b/src/emc/rs274ngc/rs274ngc_return.hh
index 29e320625e..24dcf27ca5 100644
--- a/src/emc/rs274ngc/rs274ngc_return.hh
+++ b/src/emc/rs274ngc/rs274ngc_return.hh
@@ -77,8 +77,8 @@
 #define NCE_EQUAL_SIGN_MISSING_IN_PARAMETER_SETTING _("Equal sign missing in parameter setting")
 #define NCE_F_WORD_MISSING_WITH_INVERSE_TIME_ARC_MOVE _("F word missing with inverse time arc move")
 #define NCE_F_WORD_MISSING_WITH_INVERSE_TIME_G1_MOVE _("F word missing with inverse time g1 move")
-#define NCE_FILE_ENDED_WITH_NO_PERCENT_SIGN _("File ended with no percent sign (%)")
-#define NCE_FILE_ENDED_WITH_NO_PERCENT_SIGN_OR_PROGRAM_END _("File ended with no percent sign (%) or program end (M2)")
+#define NCE_FILE_ENDED_WITH_NO_PERCENT_SIGN _("File ended with no percent sign (%%)")
+#define NCE_FILE_ENDED_WITH_NO_PERCENT_SIGN_OR_PROGRAM_END _("File ended with no percent sign (%%) or program end (M2)")
 #define NCE_FILE_NAME_TOO_LONG _("File name too long")
 #define NCE_G_CODE_OUT_OF_RANGE _("G-code out of range")
 #define NCE_I_WORD_GIVEN_FOR_ARC_IN_YZ_PLANE _("I word given for arc in yz plane")
```

smoe:
[...]

PR #3227, opened 2024-12-14
Backport issues 3200 and 3202

BsAtHome:
This is a backport of issues #3200 and #3202, which were fixed in master PR #3216 and PR #3215.


andypugh:
OK, looks reasonable

PR #3233, opened 2024-12-19
Fix invalid escape sequence warnings

BsAtHome:
Several hal components have invalid escape sequences by using single \ followed by an unknown character to escape. This results in a SyntaxWarning in python 3.12 and will become a SyntaxError in the future (see: https://docs.python.org/3/library/re.html second paragraph).

This patch fixes all instances of bad escapes. The resulting code from the components and halcompile were checked between pre- and post-error-fix. Both versions of the generated code and documentation are 100% the same.

### DIFF
--- a/src/hal/components/bldc.comp
+++ b/src/hal/components/bldc.comp
@@ -313,7 +313,7 @@ This can be over-ridden by the \\fBT\\fR tag. Sinusoidal commutation is signific
 To use an encoder for commutation a reference 0-degrees point must be found.
 The component uses the convention that motor zero is the point that an unloaded
 motor aligns to with a positive voltage on the A (or U) terminal and the B & C
-(or V and W) terminals connected together and to \\-ve voltage. There will be
+(or V and W) terminals connected together and to -ve voltage. There will be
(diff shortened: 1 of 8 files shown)


andypugh:
The code will be the same as most (or all) of these escapes are in the documentation strings. 
I think that in most cases the correct fix would be to remove the \ rather than to add a second one. 
I have a feeling that some of this goes back to a problem we were having where the docs creation was replacing hyphens with em-dashes. The result of this was that code copied out of the online docs wouldn't work. (leading to user confusion)

I think that this fix needs to be checked to see what happens if \- is replaced by - rather than by \\-, specifically regarding the legibility of the docs.

BsAtHome:
You are right. All <code>\\\\-</code> escapes are translated into a single hyphen (0x2d) when man(1) renders the page. A change to remove that particular escape was committed to the branch.

And, yes, most changes are in the documentation except for one. The halcompile.g source has a few missing escapes which python does not agree with. These changes do not change the operation of the program (verified). The changes just make the program comply with proper escaping.

smoe (commented):
Very nice!

PR #3234, opened 2024-12-20
Function calls must adhere to their prototypes

BsAtHome:
Several function calls added arguments to the function call but the functions in question do not take any arguments.

### DIFF
--- a/src/hal/classicladder/classicladder_gtk.c
+++ b/src/hal/classicladder/classicladder_gtk.c
@@ -176,7 +176,7 @@ void GetTheSizesForRung( void )
 
 	// size of the page or block changed ?
 	if ( InfosGene->PageHeight!=PageHeightBak || InfosGene->BlockHeight!=BlockHeightBak )
-		UpdateVScrollBar( TRUE/*AutoSelectCurrentRung*/ );
+		UpdateVScrollBar( /*AutoSelectCurrentRung*/ );
 	PageHeightBak = InfosGene->PageHeight;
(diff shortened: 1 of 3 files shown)


andypugh:
It does take an argument in the upstream version, curiously. 
https://github.com/MaVaTi56/classicladder/blob/main/src/classicladder_gtk.h#L46

This feels like a merge error during the update to Gtk3: 
https://github.com/LinuxCNC/linuxcnc/commits/master/src/hal/classicladder/classicladder_gtk.c

BsAtHome:
I don't know about any merges, but the functions UpdateVScrollBar() (hal/classicladder/classicladder_gtk.c:211) and SerialOpen() (hal/classicladder/serial_linux.c:50) are defined without having any arguments.

It is my understanding that LinuxCNC's version has diverged considerably from the upstream version. There are two options: 1) fix all warnings in upstream and get in sync with upstream, or 2) fix the local version.

I think that option 2 is the only viable option at this time.

andypugh:
I think I would prefer to look at the changes necessary to allow "auto select current rung" to work, as that sounds good. 
But I don't see me doing it, so will merge this.

BsAtHome:
I just took a look at the original and compiled it. The versions LCNC and source are essentially more than 10 years apart.

However, that is not the worst of. It would be possible to port again, but the current source version has over 350 compiler warnings (many are GTK deprecation warnings requiring partial rewrites). And I haven't even used -Wextra. Fixing all this in upstream and then porting to LCNC is a lot of work. I'm not sure it is worth the effort.

PR #3235, opened 2024-12-20
Use consistent silencing of make messages

BsAtHome:
Making man pages is way more verbose than all other jobs. This PR streamlines the a2x call for making man pages to use the same noise limiting procedure. It removes a lot of clutter in the terminal when doing a build.

Additionally, reducing the clutter exposed some formatting warnings which are also fixed. There may be warnings left about "too many manpage names", but this is an internal asciidoc issue.

### DIFF
--- a/docs/src/Submakefile
+++ b/docs/src/Submakefile
@@ -997,10 +997,11 @@ TARGETS += manpages
 # For more information, see https://github.com/asciidoc/asciidoc/issues/44
 GENERATED_MANPAGES += $(patsubst $(DOC_DIR)/src/man/%.adoc, $(DOC_DIR)/man/%, $(wildcard $(DOC_DIR)/src/man/man?/*.adoc))
 $(DOC_DIR)/man/%: $(DOC_DIR)/src/man/%.adoc
-	@mkdir -p `dirname $@`
-	a2x -v --doctype manpage \
+	$(ECHO) Making manpage $(notdir $@)
+	@mkdir -p $(dir $@)
+	$(Q)a2x --doctype manpage \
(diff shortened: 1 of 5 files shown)


docs/src/Submakefile
```diff
@@ -997,10 +997,11 @@ TARGETS += manpages
...
 GENERATED_MANPAGES += $(patsubst $(DOC_DIR)/src/man/%.adoc, $(DOC_DIR)/man/%, $(wildcard $(DOC_DIR)/src/man/man?/*.adoc))
 $(DOC_DIR)/man/%: $(DOC_DIR)/src/man/%.adoc
-	@mkdir -p `dirname $@`
-	a2x -v --doctype manpage \
+	$(ECHO) Making manpage $(notdir $@)
+	@mkdir -p $(dir $@)
+	$(Q)a2x --doctype manpage \
 		--format manpage \
-		--destination-dir `dirname $@` \
+		--destination-dir $(dir $@) \
```
hansu, line 1004: What about

`--destination-dir $(@D) \`

?

BsAtHome:
Actually, the reason is that <code>$(notdir $@)</code>is already used many places (or <code>$*</code>, <code>$&lt;</code>, ... instead of <code>$@</code>). Using <code>$(dir $@)</code> is IMO more clear and consistent.

docs/src/Submakefile
BsAtHome: Hm, should have replied in this box... Oh well, see below :-)

snowgoer540:
So much better ... thanks for this!

PR #3241, opened 2024-12-28
Fix type-punned dereference warning

### LINKED ISSUE #3204: Type punned dereference in libnml
A cast to int& is performed in libnml/nml/stat_msg.cc:38:
```c++
    cms->update((int&)(((RCS_STAT_MSG *) buf)->status));
```
This int& cast evokes a warning:
<code>libnml/nml/stat_msg.cc:38:48: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]</code>

It is a serious warning because the type of the class member 'status' (class RCS_STAT_MSG in libnml/nml/stat_msg.hh:26) is of an enum type. There are no guarantees that an enum has the size of an int. It usually has, but there are no guarantees. The C++ standard only states that the underlying type <i>is no larger than</i> int. Since C++11 you can force an underlying type to be used, but that does not help in this case. The declaration of the enum in libnml/rcs/rcs.hh:31 also means that it is <i>not</i> a scoped enumeration. Otherwise it would have been int by default, but the use-case here cannot make such guarantees. Therefore, there are possible size and alignment issues in the cast to int&.

One solution would be to declare RCS_STAT_MSG::status as int and the type-punning removed. However, there are many instances that then would require int casts in comparisons. Therefore, leaving it as is would be preferable. Removing the type-pun can be accomplished by using a temporary as shown in the diff below. It simply copies the current status member value into a local int, calls the CMS::update() and copies the result back in the status member.
```diff
diff --git a/src/libnml/nml/stat_msg.cc b/src/libnml/nml/stat_msg.cc
index 7e1301c287..830a930d06 100644
--- a/src/libnml/nml/stat_msg.cc
+++ b/src/libnml/nml/stat_msg.cc
@@ -35,7 +35,9 @@ int RCS_STAT_MSG_format(NMLTYPE t, void *buf, CMS * cms)
 {
     cms->update(((RCS_STAT_MSG *) buf)->command_type);
     cms->update(((RCS_STAT_MSG *) buf)->echo_serial_number);
-    cms->update((int&)(((RCS_STAT_MSG *) buf)->status));
+    int i = (int)((RCS_STAT_MSG *) buf)->status;
+    cms->update(i);
+    ((RCS_STAT_MSG *) buf)->status = (RCS_STATUS)i;
     cms->update(((RCS_STAT_MSG *) buf)->state);
 
     switch (t) {
```

BsAtHome:
This fixes #3204.

No formats or sizes change. Just the way the value of the enum is addressed. The enum was already defined as an <code>enum class</code> which has int as the default underlying type. This has not changed but the underlying type is made explicit because the union absolutely needs both fields as ints to match size (and alignment). See also added code comments.

### DIFF
--- a/src/libnml/nml/stat_msg.cc
+++ b/src/libnml/nml/stat_msg.cc
@@ -35,7 +35,7 @@ int RCS_STAT_MSG_format(NMLTYPE t, void *buf, CMS * cms)
 {
     cms->update(((RCS_STAT_MSG *) buf)->command_type);
     cms->update(((RCS_STAT_MSG *) buf)->echo_serial_number);
-    cms->update((int&)(((RCS_STAT_MSG *) buf)->status));
+    cms->update(((RCS_STAT_MSG *) buf)->status_int);
     cms->update(((RCS_STAT_MSG *) buf)->state);
 
     switch (t) {
--- a/src/libnml/nml/stat_msg.hh
+++ b/src/libnml/nml/stat_msg.hh
@@ -23,7 +23,22 @@ class RCS_STAT_MSG:public NMLmsg {
     RCS_STAT_MSG(NMLTYPE t, size_t sz);
     NMLTYPE command_type;
     int echo_serial_number;
-    RCS_STATUS status;
+    // The anonymous union is provided to allow for RCS_STAT_MSG_format() in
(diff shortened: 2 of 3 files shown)


andypugh:
Thanks for your work on this. (and everything else)

PR #3248, opened 2025-01-05
drowidget: Fix missing property 'background_color'

hansu:
The missing property 'background_color' lead to a lot of error messages:  
` Got unknown property in <<set_style>>`  
when using the drowidget VCP, for example with `gmoccapy_with_user_tabs.ini`.

This fix provided by @phillc54 fixes this. Thanks Phill!

### DIFF
--- a/lib/python/gladevcp/drowidget.py
+++ b/lib/python/gladevcp/drowidget.py
@@ -16,7 +16,6 @@
 
 import sys, os
 import math
-import linuxcnc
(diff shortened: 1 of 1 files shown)


andypugh:
Is the deletion of "import linuxcnc" deliberate? Is it necessary?

hansu:
It works also with the import, I guess it's just not needed.

PR #3251, opened 2025-01-07
Docs: add clearer info about lowercase letters in INI entries

Sigma1912:
See
https://forum.linuxcnc.org/38-general-linuxcnc-questions/54922-2-9-3-ini-vars-issue


Sigma1912:
Thanks for the clarification. I updated the PR.

PR #3260, opened 2025-01-09
Fix make empty variable warnings

BsAtHome:
Fixed in this PR:
* Some hal modules do not exist in uspace that should be in a conditional for the build. The object build rules are in a conditional, but the target build rules were not.
* A change in 2021 (https://github.com/LinuxCNC/linuxcnc/commit/37758168e509400e29bc957f8e52f3c6bdc6150f) removed t_on/t_off/t_p when they got component'ized and renamed. This removes the lingering deps.
* Reduce the terminal clutter, comment out adding --warn-undefined-variables to MAKEFLAGS. Older debian did not show the clutter, but debian:sid does (as wel as other distros).

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -27,7 +27,7 @@ install: install-software install-docs
 install-software: install-kernel-dep install-kernel-indep install-menu
 
 ifeq ($(origin KERNELRELEASE), undefined)
-MAKEFLAGS += --warn-undefined-variables
+#MAKEFLAGS += --warn-undefined-variables
 endif
 
 EXTRAFLAGS :=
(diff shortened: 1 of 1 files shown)


andypugh:
I wonder _why_ those (mainly) drivers do not exist in uspace? They seem to be a mix of ISA and PCI boards. 
Other PCI boards (Mesa, for example) work fine in uspace. 
I wonder if this is just because the uspace developer didn't have hardware to test on?

andypugh:
It looks like Motenc-Lite is stilla current project, though it looks like an expensive choice. 
https://www.vitalsystem.com/motion/motenc/motenc_lite.php

andypugh:
If the code is 15 years old then I think that it is not related to uspace (which is younger), but rather to the old "sim" build target which explicitly excluded hardware drivers. 
At the moment there are RTAI packages that users can use, but its not clear to me how long that will continue. At the present i have to build those packages by hand for each release using a modified version of the LinuxCNC build system (the current mainline build system fails to build RTAI debs)

BsAtHome:
Well, for example the hal_motenc.c driver (for the one you mentioned) was added between 21 and 19 years ago. The last fix was fixing a PCI API deprecation 13 years ago. With uspace, the commit 1d6c0d0646217d461d91ead139b6a99f1cf5c96f just moved the 'sim' exclusions into 'uspace' exclusions. I guess nobody cared about these drivers being ported to uspace functionality.

That said, this PR does not _change_ current behavior. I merely fixes the rules such that no warnings are generated. If someone wants these drivers to become active in uspace, they are welcome to do the port, but that is a different problem.

I don't have/run RTAI, so I can't comment on the specific problems on that build. But, maybe it should receive some caring attention, reading the required out-of-tree efforts to make it work. But this is also beyond the scope of this PR.

PR #3264, opened 2025-01-10
Fix snprintf() buffer size warning

BsAtHome:
The string expansion buffer must be larger that the string argument.


BsAtHome:
The failing CI rip-and-test-clang is caused by the package installer, not actually the code in question.

The failing CI job must simply be restarted.

BsAtHome:
@andypugh the CI build failed on installing the firefox-snap package. Not the first time this happens. Anyway, just restart the job to make it succeed. The code change is trivially small, but fixes annoying warnings.

Issue #3203, opened 2024-12-01
Python 3.12 deprecation warning


BsAtHome:
The follwing warning is produced when compiling with python 3.12:
<code>emc/pythonplugin/python_plugin.cc:329 col 3: warning: ‘Py_UnbufferedStdioFlag’ is deprecated [-Wdeprecated-declarations]</code>

The <a href="https://docs.python.org/3/c-api/init.html">python documentation</a> states:
<code>int Py_UnbufferedStdioFlag
    This API is kept for backward compatibility: setting [PyConfig.buffered_stdio](https://docs.python.org/3/c-api/init_config.html#c.PyConfig.buffered_stdio) should be used instead, see [Python Initialization Configuration](https://docs.python.org/3/c-api/init_config.html#init-config).
    Force the stdout and stderr streams to be unbuffered.
    Set by the [-u](https://docs.python.org/3/using/cmdline.html#cmdoption-u) option and the [PYTHONUNBUFFERED](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONUNBUFFERED) environment variable.
    <b>Deprecated since version 3.12, will be removed in version 3.14.</b></code>

I am no python expert and especially not versed in how it is integrated in LinuxCNC. Can someone who knows about this suggest a fix?

rene-dev:
closed by https://github.com/LinuxCNC/linuxcnc/pull/3265

PR #3265, opened 2025-01-10
Fix Py_UnbufferedStdioFlag deprecation warning

BsAtHome:
The use of Py_UnbufferedStdioFlag has been deprecated in python 3.12 (current debian:sid and other distros) and will be removed in 3.14. This PR fixes the warning (see also: #3203) and also fixes additional problems with the PyConfig code.

@rene-dev: The python documentation states that you need to use Py_InitializeFromConfig() when you change the PyConfig and you need to release the configuration with PyConfig_Clear() (see: https://docs.python.org/3.12/c-api/init_config.html#initialization-with-pyconfig).

Note: the merging #2487 to fix a previous deprecation warning introducing use of PyConfig caused all python versions lower than 3.8 to fail or anything older than Bullseye (see https://wiki.debian.org/Python). Therefore, no CPP conditional are placed around the code to support older distributions.

### DIFF
--- a/src/emc/pythonplugin/python_plugin.cc
+++ b/src/emc/pythonplugin/python_plugin.cc
@@ -326,8 +326,9 @@ PythonPlugin::PythonPlugin(struct _inittab *inittab) :
         }
       }
   }
-  Py_UnbufferedStdioFlag = 1;
-  Py_Initialize();
+  config.buffered_stdio = 0;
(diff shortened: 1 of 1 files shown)


BsAtHome:
@rene-dev: What is the status of this PR?

BsAtHome:
Thanks!

Yedfree (approved):

PR #3266, opened 2025-01-10
Fix ignored qualifiers

BsAtHome:
Running the compile with <code>-Wignored-qualifiers</code> results in more warnings that indicate wrong types being used. This PR fixes two cases:
* ignored volatile qualifier on cast
* ignored const or volatile return qualifier

Returning const or volatile _values_ is inappropriate, just like casting a _value_ to volatile is inappropriate. The hal_X_t types are replaced with the underlying type and the const on return values is removed.

BTW, yes, I'm trying to eliminate <i>all</i> warnings from the compile.

### DIFF
--- a/src/emc/motion/axis.c
+++ b/src/emc/motion/axis.c
@@ -457,7 +457,7 @@ void axis_apply_ext_offsets_to_carte_pos(int extfactor, double *pcmd_p[])
     }
 }
 
-hal_bit_t axis_plan_external_offsets(double servo_period, bool motion_enable_flag, bool all_homed)
+bool axis_plan_external_offsets(double servo_period, bool motion_enable_flag, bool all_homed)
 {
     static int first_pass = 1;
     int n;
--- a/src/emc/motion/axis.h
+++ b/src/emc/motion/axis.h
@@ -14,7 +14,7 @@ void axis_initialize_external_offsets(void);
 int axis_init_hal_io(int mot_comp_id);
 
 void axis_handle_jogwheels(bool motion_teleop_flag, bool motion_enable_flag, bool homing_is_active);
-hal_bit_t axis_plan_external_offsets(double servo_period, bool motion_enable_flag, bool all_homed);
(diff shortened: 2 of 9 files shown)


andypugh:
> BTW, yes, I'm trying to eliminate all warnings from the compile.

I think I achieved that once (but at a lower level of compiler pedantry). I appreciate the effort.

andypugh:
You have been in there a lot recently. What do you think would be the result of replacing all hal/rtapi_u32/u64 with hal/rtapi_unsigned and the same for signed, where hal_unsigned == u64.
Some HAL files using the dev version might have to change, but that could be handled by update_ini. Some components such as stepgen will need their internal accumulators looking at (in many cases these are already 64 bits, but they all should be to eliminate the risk of wrapping)

BsAtHome:
> > BTW, yes, I'm trying to eliminate all warnings from the compile.
>
> I think I achieved that once (but at a lower level of compiler pedantry). I appreciate the effort.

Well, pedantry is necessary for optimal performance ;-)

More seriously, the codebase should compile warning free with <code>-Wextra</code> and then we also should add <code>-Werror</code> by default. There is no excuse for warnings. Warnings are (potential) bugs and generally bad coding practice. There is no need for any warning to be present, at all.

andypugh:
My feeling is that on 64-bit CPUs there is probably no difference in speed between 32-bit and 64-bit operations. 

The background to this is that initially LinuxCNC used 32-bit floats and 32-bit ints. At some point all the floats were switched to double (and this was before 64-bit CPUs were ubiquitous). Then much later the 64-bit ints were introduced in addition to the 32-bit ones. The 2.9 branch does not have 64-bit HAL pins.
My feeling is that we should not have introduced 64 bits alongside 32, we should have made a clean switch to 64-bit ints in master. And it's not (yet) too late to do that. 
This should be pretty much invisible to the users. Some kludges may be needed, LUT5 would become LUT6 for example, but that can be handled in the update_ini script.

andypugh:
For further discussion see https://github.com/LinuxCNC/linuxcnc/issues/3286

PR #3269, opened 2025-01-11
Gmoccapy 3.4.9 - several bugfixes for 2.9

hansu:
This fixes some annoying bugs in the current 2.9 release:
- Fix problems with duplicated entries in the preferences file (#3181)
- Remove Gtk.Paned for upper main area because this prevents one to lick the DRO axes letters on touch screens (https://www.forum.linuxcnc.org/gmoccapy/51526-unwanted-actions-with-a-touch-screen)
- Fix wrong states of spindle button images after program abort

### DIFF
--- a/lib/python/gladevcp/offsetpage.glade
+++ b/lib/python/gladevcp/offsetpage.glade
@@ -37,7 +37,7 @@
     </columns>
     <data>
       <row>
-        <col id="0" translatable="yes">Tool</col>
+        <col id="0">Tool</col>
         <col id="1">0</col>
         <col id="2">0</col>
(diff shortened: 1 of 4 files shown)


hansu:
@andypugh can we do a 2.9.4 release after this has been merged? Apart from this one there are also some bugfixes for the gmoccapy tooltable that has been added after the 2.9.3 release. Would feel better to have them out :)

andypugh:
Making the release is easy. Getting all the debs made and in the right places on the server is quite a bit of work, especially as Pi / RTAI are hand-built even if the buildbot is working at the moment the release is made.

PR #3272, opened 2025-01-11
Add format length limit to prevent format-truncation warning

BsAtHome:
A spurious format truncation warning appears when extra warnings are enabled (using g++ 14.2.1). This might be a false positive, but is easily fixed by limiting the expansion size. All the names have a maximum length of 20 characters so limiting the expansion to that length should be fine.

### DIFF
--- a/src/emc/usr_intf/emclcd.cc
+++ b/src/emc/usr_intf/emclcd.cc
@@ -491,7 +491,7 @@ static int createWidgets()
   int i;
 
   for (i=0; i<WIDGET_COUNT; i++) {
-    snprintf(sockStr, sizeof(sockStr), "widget_add %s %s %s\n", widgets[i].screenName,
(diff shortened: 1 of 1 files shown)


andypugh:
I suspect that no-one has ever used emclcd, but whilst it exists, it might as well compile cleany.

BsAtHome:
> I suspect that no-one has ever used emclcd, but whilst it exists, it might as well compile cleany.

It may need a caring hand. Well, a firm hand might be required. The code references "/dev/hda1". That is a long time ago I've seen that device.

andypugh:
It is retained mainly as an example of a user interface using a different paradigm to Python + GUI framework. 
It's actually a userinterface running on a particular 4x20 LCD display with integrated buttons intended to mount in the front panel of your PC tower case (which could have been quite a neat installation for a special-purpose CNC machine) 
This might be one:  https://www.ebay.co.uk/itm/115938873276

PR #3277, opened 2025-01-11
Add missing backslash

BsAtHome:
Another Python SyntaxWarning about invalid escapes was encountered while doing a <code>make pycheck</code>. This PR fixes these instances.


andypugh:
Would n r-string be better here? ie:
```Python
m = re.match(r'(.*)\.ngc', ngc)
```

andypugh:
I think I would have preferred raw strings everywhere, as they are easier to read. But I suspect that some of the code is old enough to pre-date them. (though a bit of a Google suggests that this is unlikely)

PR #3278, opened 2025-01-12
Fix assigned but unused variables

BsAtHome:
This PR removes variables that are set but not used (clang warnings). Some are only commented out if it makes sense in the context of those variables.
One case in src/hal/user_comps/xhc-hb04.cc now actually uses the value because it is a collection of possible errors and is passed on to report any error.


andypugh:
I am almost surprised there are so few.

BsAtHome:
> I am almost surprised there are so few.

There are many more, but they are exposed with a different warning option. The bulk of unused warnings are from unused function arguments. That list was too long to take on while all the other warnings were blinding me.

PR #3281, opened 2025-01-12
Fix clang vla problems

BsAtHome:
Variable length arrays (vla) are a feature of C and are not in the C++ standard. Both gcc and clang have fixes for this, but they remain non-standard and result in a warning in clang.

This PR removes the use of vla and replaces it with new[]/delete[]. This is all non-RT code, so that is no problem. In the process it uncovered some subtle and less subtle problems in the code:
* strtok_r() reused the state pointer as string argument. This is very inappropriate because the state pointer is an opaque libc internal state. You should never depend that. The code is modified to use documented strtok_r() procedures.
* Several instances did not check the vla size for being zero or negative. Both cases are not handled well in vla's. The code is modified to test for negative or zero lengths and acts accordingly.

### DIFF
--- a/src/hal/utils/halcmd_commands.cc
+++ b/src/hal/utils/halcmd_commands.cc
@@ -56,7 +56,7 @@
 #include <errno.h>
 #include <time.h>
 #include <fnmatch.h>
-
+#include <vector>
 
 static int unloadrt_comp(char *mod_name);
 static void print_comp_info(char **patterns);
@@ -2605,8 +2605,8 @@ static void save_comps(FILE *dst)
         return;
 	}
 
-    hal_comp_t **comps = new hal_comp_t*[ncomps];
-    hal_comp_t **compptr = comps;
+    std::vector<hal_comp_t *> comps(ncomps, nullptr);
+    hal_comp_t **compptr = comps.data();
     next = hal_data->comp_list_ptr;
     while(next != 0)  {
 	comp = SHMPTR(next);
@@ -2628,9 +2628,6 @@ static void save_comps(FILE *dst)
                 (char *)SHMPTR(comp->insmod_args));
         }
     }
-
-    delete[] comps;
-
 #if 0  /* newinst deferred to version 2.2 */
     next = hal_data->comp_list_ptr;
     while (next != 0) {
--- a/src/hal/utils/halcmd_commands.cc
+++ b/src/hal/utils/halcmd_commands.cc
@@ -2599,7 +2599,14 @@ static void save_comps(FILE *dst)
 	next = comp->next_ptr;
     }
 
-    hal_comp_t *comps[ncomps], **compptr = comps;
+    if(!ncomps) {
+        // No components found, bail
+        rtapi_mutex_give(&(hal_data->mutex));
+        return;
+	}
+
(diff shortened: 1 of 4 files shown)


BsAtHome:
You are probably right. However std::vector is unlikely to work considering the way the rest of the code works. I will look into the alternatives.

rmu75:
std::vector always allocates a contiguous array, it is kind of auto-resizing dynamic array. when constructing with size it should do only one allocation. So it should be possible to get the pointer to first element and leave the code more or less unchanged. I will take a look.

Better solution would be something like dynamic std::array, but that doesn't seem to exist in standard C++. Maybe std::valarray? Sometime in the future.

We should put comment there why it's done like that, sometime in the future either val arrays will come to C++ or prober stack-allocated std::array with at-construction size (vs. at-compile size as it is now).

src/rtapi/uspace_rtapi_app.cc
```diff
@@ -346,14 +346,25 @@ static int read_number(int fd) {
...
+        return string();
+    char *buf = new char[len];
+    if(read(fd, buf, len) != len) {
+        delete[] buf;
+        throw ReadError();
+    }
+    string str(buf, len);
+    delete[] buf;
+    return str;
 }
```
rmu75, line 361: I think we could read into std::string directly, something like

`
string str{len, 0};
if (read(fd, &str[0], len) != len) throw ReadError();
return str;
`

src/hal/utils/halcmd_commands.cc
```diff
@@ -2599,7 +2599,14 @@ static void save_comps(FILE *dst)
...
 
-    hal_comp_t *comps[ncomps], **compptr = comps;
+    if(!ncomps) {
+        // No components found, bail
+        rtapi_mutex_give(&(hal_data->mutex));
+        return;
+	}
+
+    hal_comp_t **comps = new hal_comp_t*[ncomps];
+    hal_comp_t **compptr = comps;
```
rmu75, line 2609:
``` 
auto comps = std::vector<hal_comp_t*>(ncomps, null_ptr);
hal_comp_t **compptr = &comps[0];
```

src/rtapi/uspace_rtapi_app.cc
BsAtHome: Fixed in https://github.com/LinuxCNC/linuxcnc/pull/3281/commits/b8434f81cd20d84e4d6eff6757b6633a7478ced1

src/hal/utils/halcmd_commands.cc
BsAtHome: Fixed in https://github.com/LinuxCNC/linuxcnc/pull/3281/commits/b8434f81cd20d84e4d6eff6757b6633a7478ced1

BsAtHome:
> std::vector always allocates a contiguous array,...

Yes, but I was thinking about string handling, not generic arrays. There you need the terminating NUL for C-string compatibility. Luckily, the string class keeps the array NUL terminated since C++11.

Anyway, it should now be resolved with the commit. Please review.

rmu75:
I like that much better, LGTM.

Issue #3129, opened 2024-09-22
2.9 Gmoccapy - wrong tool disable MDI mode


zz912:
If I set wrong tool (tool is not in tooltable), then I cannot activate MDI windows. MDI mode is activated, but users cannot see it.

https://github.com/user-attachments/assets/f3256603-3337-47f3-93fa-c32a9908f774

I entered G49 to rule out bugs related to AUTOMATIC_G43.

For better bug diagnosis, I made this PR:
https://github.com/LinuxCNC/linuxcnc/pull/3123

Sigma1912:
Confirmed. 
Here is the debug from my simulation machine running master.

First changing to a  tool nr that is present in the tool table (Note the 'RUN' and 'IDLE' messages that point to GStat correctly sending messages about the interpreter mode changing first to run and then back to idle):
```
[Gmoccapy][DEBUG]  ntb_button_switch_page (gmoccapy:5256)
3 2
[Gmoccapy][DEBUG]  MDI Mode, tool_change = True (gmoccapy:2745)
[Gmoccapy][DEBUG]  ntb_button_switch_page (gmoccapy:5256)
[Gmoccapy][DEBUG]  RUN (gmoccapy:2619)
[Gmoccapy][DEBUG]  hal status motion mode changed (gmoccapy:2821)
[Gmoccapy][DEBUG]  IDLE (gmoccapy:2567)
[Gmoccapy][DEBUG]  hal signal tool changed (gmoccapy:2638)
[Gmoccapy][DEBUG]  Tool is now 1 (gmoccapy:3497)
[Gmoccapy][DEBUG]  G43 is active (gmoccapy:3499)

```


Then here when calling a non-existent tool nr (note the absence of 'RUN' and 'IDLE' ):

```
task: main loop took 0.022615 seconds
emc/task/emctask.cc 68: interp_error: Requested tool 45 not found in the tool table
Requested tool 45 not found in the tool table
task: main loop took 0.019085 seconds
[Gmoccapy][DEBUG]  MDI Mode, tool_change = True (gmoccapy:2745)
[Gmoccapy][DEBUG]  ntb_button_switch_page (gmoccapy:5256)
[Gmoccapy][DEBUG]  hal status motion mode changed (gmoccapy:2821)
[Gmoccapy][DEBUG]  _on_play_sound <__main__.gmoccapy object at 0x7fe3cc270d00> None error (gmoccapy:5501)

```
[...]

PR #3282, opened 2025-01-13
gmoccapy: fix bugs caused by GStat missing change in interpreter mode

Sigma1912:
this fixes issues #3120 and #3129


Sigma1912:
Should have gone into 2.9, closing this.

PR #3284, opened 2025-01-13
Fix clang minmax coercion

BsAtHome:
This PR fixes the clang warnings about values not fitting the type and implicit conversion.

The previous solution to generating the type conversion components used the C preprocessor with very large literal int and uint values, passed to the script, to determine whether some code should be enabled or not. This is prone to error as the preprocessor may not handle types the same way as the actual language. This results in errors and warnings.

All conversions are a matrix of possible combinations that have predetermined outcome how they should be handled. The minimum and maximum values for the target types are defined in the C language and are the authoritative values:  <code>0</code>, <code>1</code>, <code>INT{32,64}_{MIN,MAX}</code> and <code>UINT{32,64}_MAX</code>.

The PR changes the build to use a programmed approach to determine which parts of the component template are used and how they are activated. The preprocessor is no longer used with large values but only simple true/false conditions. Template generation invocation no longer requires the min/max values to be passed as they are determined by the script using C defined symbolic values. Finally, the conversion components use the underlying type for caching the value, which may be properly optimized by the compiler and values are cast to their targets.

### DIFF
--- a/src/hal/components/Submakefile
+++ b/src/hal/components/Submakefile
@@ -109,119 +109,116 @@ TARGETS += ../bin/panelui
 
 hal/components/conv_float_s32.comp: hal/components/conv.comp.in hal/components/mkconv.sh hal/components/Submakefile
 	$(ECHO) converting conv for $(notdir $@)
-	$(Q)sh hal/components/mkconv.sh float s32 "" -2147483647-1 2147483647 < $< > $@
+	$(Q)sh hal/components/mkconv.sh float s32 < $< > $@
 
 hal/components/conv_float_u32.comp: hal/components/conv.comp.in hal/components/mkconv.sh hal/components/Submakefile
 	$(ECHO) converting conv for $(notdir $@)
-	$(Q)sh hal/components/mkconv.sh float u32 "" 0 2147483647 < $< > $@
+	$(Q)sh hal/components/mkconv.sh float u32 < $< > $@
 
 hal/components/conv_float_s64.comp: hal/components/conv.comp.in hal/components/mkconv.sh hal/components/Submakefile
 	$(ECHO) converting conv for $(notdir $@)
-	$(Q)sh hal/components/mkconv.sh float s64 "" -9223372036854775807u-1 9223372036854775807u < $< > $@
+	$(Q)sh hal/components/mkconv.sh float s64 < $< > $@
 
-# Clang error out when using 18446744073709551616u, reduce range by
-# one as a workaround:
-#  error: integer literal is too large to be represented in any integer type
 hal/components/conv_float_u64.comp: hal/components/conv.comp.in hal/components/mkconv.sh hal/components/Submakefile
 	$(ECHO) converting conv for $(notdir $@)
-	$(Q)sh hal/components/mkconv.sh float u64 "" 0 18446744073709551615u < $< > $@
+	$(Q)sh hal/components/mkconv.sh float u64 < $< > $@
 
 hal/components/conv_bit_s32.comp: hal/components/conv.comp.in hal/components/mkconv.sh hal/components/Submakefile
 	$(ECHO) converting conv for $(notdir $@)
-	$(Q)sh hal/components/mkconv.sh bit s32 // < $< > $@
+	$(Q)sh hal/components/mkconv.sh bit s32 < $< > $@
 
 hal/components/conv_bit_u32.comp: hal/components/conv.comp.in hal/components/mkconv.sh hal/components/Submakefile
 	$(ECHO) converting conv for $(notdir $@)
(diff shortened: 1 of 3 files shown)


andypugh:
I have just noticed that this breaks the build with an RTAI kernel. 

Specifically this line:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/components/conv.comp.in#L11

Appears to conflict (indirectly) with:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/rtapi/rtapi.h#L834

```
CC [M]  /home/andypugh/linuxcnc-dev/src/objects/hal/components/conv_bit_float.o
In file included from /usr/lib/gcc/x86_64-linux-gnu/12/include/stdint.h:11,
                 from hal/components/conv_bit_float.comp:11:
/usr/lib/gcc/x86_64-linux-gnu/12/include/stdint-gcc.h:43:24: error: conflicting types for ‘int64_t’; have ‘long int’
   43 | typedef __INT64_TYPE__ int64_t;
      |                        ^~~~~~~
In file included from ./arch/x86/include/asm/alternative.h:7,
                 from ./arch/x86/include/asm/barrier.h:5,
                 from ./include/linux/compiler.h:325,
                 from ./include/linux/ioport.h:13,
                 from /home/andypugh/linuxcnc-dev/src/../include/rtapi.h:834,
                 from /home/andypugh/linuxcnc-dev/src/objects/hal/components/conv_bit_float.c:2:
./include/linux/types.h:114:33: note: previous declaration of ‘int64_t’ with type ‘int64_t’ {aka ‘long long int’}
  114 | typedef s64                     int64_t;
      |                                 ^~~~~~~
/usr/lib/gcc/x86_64-linux-gnu/12/include/stdint-gcc.h:55:25: error: conflicting types for ‘uint64_t’; have ‘long unsigned int’
   55 | typedef __UINT64_TYPE__ uint64_t;
      |                         ^~~~~~~~
./include/linux/types.h:112:33: note: previous declaration of ‘uint64_t’ with type ‘uint64_t’ {aka ‘long long unsigned int’}
  112 | typedef u64                     uint64_t;
      |                                 ^~~~~~~~
make[2]: *** [scripts/Makefile.build:262: /home/andypugh/linuxcnc-dev/src/objects/hal/components/conv_bit_float.o] Error 1
make[1]: Leaving directory '/usr/src/linux-headers-5.4.258-rtai-amd64'
make[1]: *** [Makefile:1754: /home/andypugh/linuxcnc-dev/src] Error 2
make: *** [Makefile:547: modules] Error 2
```

There are probably not many of us building LinuxCNC on RTAI at the moment, but I would prefer it to remain possible.

BsAtHome:
Sorry about that. Fixed in PR #3295.

andypugh:
I am not sure that this is working correctly, see for example these cases

```
     4  float IN                  -3.14  conv-float-u64.0.in
     4  u64   OUT    0xFFFFFFFFFFFFFFFD  conv-float-u64.0.out

     7  u64   IN     0xFFFFFFFFFFFFFFFF  conv-u64-s64.0.in
     7  s64   OUT                    -1  conv-u64-s64.0.out
```

I am not sure that the bottom clamp in float-to-unsigned and the top clamp in unsigned-to-signed are active. 

Also, do you have any link to something that documents -o and -a as "or" and "and"? Nothing that I have found mentions these.

andypugh:
@BsAtHome

BsAtHome:
> I am not sure that this is working correctly, see for example these cases

>      4  float IN                  -3.14  conv-float-u64.0.in
>      4  u64   OUT    0xFFFFFFFFFFFFFFFD  conv-float-u64.0.out

It seems that the clamp test is performed on the target, which should be the source.


>      7  u64   IN     0xFFFFFFFFFFFFFFFF  conv-u64-s64.0.in
>      7  s64   OUT                    -1  conv-u64-s64.0.out

Might be same problem.

> I am not sure that the bottom clamp in float-to-unsigned and the top clamp in unsigned-to-signed are active.

I'll look into this.


> Also, do you have any link to something that documents -o and -a as "or" and "and"? Nothing that I have found mentions these.
```bash
$ man test
...
  EXPRESSION1 -a EXPRESSION2
    both EXPRESSION1 and EXPRESSION2 are true
  EXPRESSION1 -o EXPRESSION2
    either EXPRESSION1 or EXPRESSION2 is true
...
```
Should be there right after <code>! EXPRESSION</code>

BsAtHome:
> I might have expected "clamp" to be set by default. It is presumably too late to change it now, with systems "in the wild" with the current behaviour.

That may be true. Though, you can do stuff in development...

That said, the clamp param is described in the man-pages where clamp is implemented. But that does not seem to be adequate.

The real question is what the <i>expected</i> behaviour is amongst users. A poll, maybe?

PR #3285, opened 2025-01-14
ELIF typo

6XoCtujg2C0gne:
Changed line 676 from IF to ELIF


Sigma1912:
Looks like you need to exclude all the ...rc.py files from your commit.

Sigma1912:
@c-morley  will probably have something to say about this.

BsAtHome:
Wouldn't it be better to commit the sources and use make + Qt resource compiler to create the compiled resource .py files?
Embedding compiled resources like this is IMO bad coding style and each change generates huge opaque diffs.

c-morley:
Why do you think the if should be elif? I don't think it should.

I don't know what all the '_rc' code is about, but it shouldn't be included.

Please let us know what problem you are trying to fix.

6XoCtujg2C0gne:
I compared it against the same Def from the other handler files.

Jan 15, 2025 7:42:23 AM c-morley ***@***.***>:

c-morley:
What problem are you trying to fix ?

Sigma1912:
I guess it might be the question why is it different from this, there are other examples:
qtaxis_handler

6XoCtujg2C0gne:
There was no “problem” other than consistency across “like” files and class definitions.

 

I’ve added “*_rc.py”  to my ignore file.

 

If you are not comfortable with the change I can either pull it back or you may cancel/delete it.

 

Andy

6XoCtujg2C0gne:
The resource files were committed inadvertently.

 

*_rc.py has been added to my ignore file.

c-morley:
Ok ill look into the other screen code to check the logic. Thanks

andypugh:
@c-morley Any decision?

c-morley:
The difference is:
using IF no keyboard jogging when gcode text display is in focus. cursor movement instead
using ELIF keyboard jogging, bo cursor movement

One might expect the gcode display to scroll text rather then jog.
maybe we should change the other screens.

andypugh:
> One might expect the gcode display to scroll text rather then jog. maybe we should change the other screens.

Given that page-up / page-down are Z jogging, but would logically be expected to scroll the program listing, I tend to agree.

6XoCtujg2C0gne:
Yes, agreed.

 

Thanks.

andypugh:
@c-morley Should we close this one, or leave it open until the alternative solution mentioned above is carried out?

c-morley:
Ok I'll close this then.

PR #3287, opened 2025-01-16
Fix unused parameter warnings

BsAtHome:
Fix all instances of unused parameters passed to functions. This PR is large, but only fixes the warnings generated by <code>-Wunused-parameter</code>. There are no actual code path changes.
Unused function parameters in C functions have added <code>(void)parm;</code> to the start of the body. Unused function parameters in C++ function have the parameter name commented out.


andypugh:
There is a merge conflict. Not difficult to sort out but...

The commented-out parameters look really rather odd to me. Especially as the type declarations are retained. Are you sure that is the best way to do it? 

Are these function parameters always unused? If they are, then why keep them at all?

BsAtHome:
The commented out parameters are there to remind one of the (once) intended use, but I admit that some naming is currently not very to the point. Anyway, the type declarations are the "master-view" of how a function should be. Therefore you'd normally have them all in there. Just removing the name of the parameter in the definition may lead to confusion later on.

Also, you cannot just remove parameters from the declaration and definition. Many functions are API related and have several uses. Not all options/parameters may be needed when implemented. An example of this is emc/sai/saicanon.cc, which is basically a stub. Another example is in classicladder's many files interfacing with gtk.

Essentially, what is left is a choice. We remove the unused parameter names or comment them out. My preference is to have them commented out, but others may have another preference. You decide.

PR #3288, opened 2025-01-16
Fix cppcheck to check near all sources

BsAtHome:
This PR rewrites the cppcheck script to do a (near) full source-tree check of C and C++ files.
The new version exposes many, many more problems than previously seen. Fixing these will be a next project.

### DIFF
--- a/scripts/cppcheck.sh
+++ b/scripts/cppcheck.sh
@@ -72,3 +72,5 @@ do
     echo "I (4/4): checking $d"
     docheck "$d"
 done
+
+exit 0
--- a/scripts/cppcheck.sh
+++ b/scripts/cppcheck.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash
 
 if ! command -v cppcheck; then
   echo "E: Please install the program 'cppcheck' prior to executing this script."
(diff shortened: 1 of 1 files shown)


andypugh:
Are these "many, many" more problems real problems? 

When is this script run? If it is part of the buildbot tests, and if failing the checks will lead to packages not being built, I am not keen on merging without also merging the fixes.

BsAtHome:
Nobody has every looked at the cppcheck results I imagine... A list of issues was there already before I made the change. My change made the list longer.

Cppcheck is about quality control of the code. The static analysis is a good way to see if you overlooked some potential problem. There may be false positives, but they can be isolated if they occur. Currently, the cppcheck.sh script does not make the CI fail the, ever. The strategy should be to clean up the indicated issues and then make the cppcheck mandatory clean to pass. BTW, this also goes for shellcheck, but have not looked at it yet.

Cppcheck's static analysis exposes issues that should be addressed. There are three classes: warnings, compatibility and performance.
* All warnings are actually programming errors. They may not have bitten you yet, but will eventually (see f.ex. #3290). Therefore, a compile should be completely warning free with highest warning setting (-Wextra and cppcheck's static analysis).
* For the portability warnings. We (are supposed to?) support both gcc and clang and different distros. Each has some small difference to cope with, which are typical portability issues. Also versions of compilers may pose challenges.
* The performance issues are often bad coding styles, like passing classes as values. They are cheap to fix and add improvements to the code base.

That is why I have been hunting warnings and making so many PRs.

PR #3289, opened 2025-01-17
Fix most of the incompatible function cast warnings

BsAtHome:
This PR takes care of nearly all warnings from <code>-Wcast-function-type</code>. A few were left to be examined closer.

Part of these warnings were caused by using functions missing an unused parameter, which could be added without changing functionality. The other set were because of initializing function pointers with different prototypes in one and the same struct field. Here the field's type is changed to a union to encompass all the function prototypes used. Dereferencing this union also reduced the amount of casting needed to call the functions.

Additionally, a non-portable call to malloc(0), allocating zero bytes, was replaced to allocate one element instead. The code flow reallocs this later.

Finally, the PR fixes a few signed to unsigned conversion with subsequent compare to less than zero. Unsigned comparison to less than zero makes no sense, but the values come from function return values that can be negative on error, which was tested for.

### DIFF
--- a/src/emc/rs274ngc/gcodemodule.cc
+++ b/src/emc/rs274ngc/gcodemodule.cc
@@ -81,10 +81,10 @@ typedef struct {
     int mcodes[ACTIVE_M_CODES];
 } LineCode;
 
-static PyObject *LineCode_gcodes(LineCode *l) {
+static PyObject *LineCode_gcodes(LineCode *l, void *) {
     return int_array(l->gcodes, ACTIVE_G_CODES);
 }
-static PyObject *LineCode_mcodes(LineCode *l) {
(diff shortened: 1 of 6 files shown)


andypugh:
```
l = PyLong_AsLongLong(tmp);
    if(PyErr_Occurred()) goto fail;
    if(l < 0 || l >= (long long)UINT64_MAX)
```
Can l (a long long) ever be > UINT64_MAX?

The original test looks suspect too (though may function as a NaN trap as well as, just possibly, an overflow detector)

BsAtHome:
You are right, long long is (for both gcc and clang) 64 bit and has an original maximum of INT64_MAX, which is about half that of UINT64_MAX. The vatiable <code>l</code> can only be negative (an error) or successfully be within unsigned long long range.

PR #3292, opened 2025-01-18
Remove the kludge and fix the race between task and motion

BsAtHome:
This PR addresses the problem that shows itself at startup with these two messages:
```
 USRMOT: ERROR: command 30 timeout
 emcMotionInit: emcTrajInit failed
```

Commit https://github.com/LinuxCNC/linuxcnc/commit/9a45ed94f75f2cef5cef29fd92d859d7ab0662a9 added a kludge to re-copy the command in case of timeout in a specific command. However, it is a kludge and it turns out that it was always the _first_ command that timed out. That clearly indicated a race/overwrite situation.

The problem exists between task and motion, usually milltask and motmod, but also motion-logger is affected. The race is an overwrite scenario with following sequence:
1. the writer, milltask, copies a command into shared memory command structure
2. the reader, motmod or motion-logger, attaches to shared memory and clears the shared memory overwriting the command structure
3. the writer never sees its command completing.

The timeout happens consistently because the milltask process starts before the motmod thread (though it is not guaranteed which process attaches first and writes the memory). The solution is obvious. The reader must not write to the command structure during initialization because there may already be a command in there.

This PR removes the kludge and removes the initialization of the command structure in the readers. The initial value of shared memory (when created) is always zero, so that works out fine if the reader attaches before the writer. The reader may set status and other structures to its liking, but the command structure is left alone.

Additionally, a micro-sleep is added in the split read functions (<code>usrmotReadEmcmotStatus()</code>, <code>usrmotReadEmcmotConfig()</code> and <code>usrmotReadEmcmotInternal()</code>). These were running a too tight loop with three retries. A tight loop is over in a matter of nanoseconds and will most likely timeout. Adding a minute sleep forces a reschedule and possible yield, giving the other side time to actually catch up. The micro sleep is no problem because the routines are only called from non-realtime.

### DIFF
--- a/src/emc/motion-logger/motion-logger.c
+++ b/src/emc/motion-logger/motion-logger.c
@@ -96,9 +96,6 @@ static int init_comm_buffers(void) {
 	return -1;
     }
 
-    /* zero shared memory before doing anything else. */
-    memset(emcmotStruct, 0, sizeof(emcmot_struct_t));
-
     /* we'll reference emcmotStruct directly */
     c = &emcmotStruct->command;
     emcmotStatus = &emcmotStruct->status;
--- a/src/emc/motion/motion.c
+++ b/src/emc/motion/motion.c
@@ -826,9 +826,6 @@ static int init_comm_buffers(void)
 	return -1;
     }
 
-    /* zero shared memory before doing anything else. */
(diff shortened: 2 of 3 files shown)


snowgoer540:
Thank you so much for looking into this and fixing the root of the problem.  I really appreciate it!

PR #3297, opened 2025-01-23
Fix rtai warnings

BsAtHome:
Many warnings not caught in uspace compilation but appear with rtai have been fixed in this PR.

The implicit fall through warnings are removed by making it explicit using the standard <code>/* Fallthrough */</code> marker, which is understood by the compiler. A variable length array warning was not actually using variable length arrays.

There are some warnings left that will be fixed in a later PR. These are about the stackframe size (larger than 2048). Maybe it could/should be enlarged?

BTW, it also fixes properly inlining the pmSq() function.

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -200,7 +200,7 @@ AC_ARG_ENABLE(toolnml,
 
 #at this point if RTS is empty, we need to find RT ourselves
 AC_MSG_CHECKING([for rtai-config])
-if test -z "$RTDIR"; then
+if test -z "$RTS"; then
   DIRS="/usr/realtime-`uname -r` /usr/realtime /usr/realtime* /usr /usr/src/rtai*"
 else
   DIRS="$RTS"
--- a/src/emc/motion/command.c
+++ b/src/emc/motion/command.c
@@ -369,6 +369,7 @@ STATIC int is_feed_type(int motion_type)
         return 1;
     default:
         rtapi_print_msg(RTAPI_MSG_ERR, "Internal error: unhandled motion type %d\n", motion_type);
+        /* Fallthrough */
     case EMC_MOTION_TYPE_TOOLCHANGE:
     case EMC_MOTION_TYPE_TRAVERSE:
     case EMC_MOTION_TYPE_INDEXROTARY:
--- a/src/emc/motion/homing.c
+++ b/src/emc/motion/homing.c
@@ -350,7 +350,7 @@ static void do_homing_sequence(void)
             }
         }
         sequence_is_set = 1;
-        //drop through----drop through----drop through----drop through
+        /* Fallthrough */
 
     case HOME_SEQUENCE_DO_ONE_SEQUENCE:
         // Expect multiple joints with home_state==HOME_START
@@ -378,8 +378,7 @@ static void do_homing_sequence(void)
             }
         }
         sequence_state = HOME_SEQUENCE_START;
-
-        //drop through----drop through----drop through----drop through
+        /* Fallthrough */
 
     case HOME_SEQUENCE_START:
         // Request to home all joints or a single sequence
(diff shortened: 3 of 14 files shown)


andypugh:
Thanks.

andypugh:
Compiled almost cleanly (RTAI) which is a great improvement. I am running the testsuite now. 
For reference, the warnings are

```
/home/andypugh/linuxcnc-dev/src/emc/tp/tp.c: In function ‘tpHandleBlendArc.isra’:
/home/andypugh/linuxcnc-dev/src/emc/tp/tp.c:2037:1: warning: the frame size of 2512 bytes is larger than 2048 bytes [-Wframe-larger-than=]
 2037 | }
      | ^

In file included from /home/andypugh/linuxcnc-dev/src/emc/motion/stashf.c:123:
/home/andypugh/linuxcnc-dev/src/emc/motion/stashf_wrap.h: In function ‘snprintdbuf’:
/home/andypugh/linuxcnc-dev/src/emc/motion/stashf_wrap.h:34:9: warning: ISO C90 forbids variable length array ‘block’ [-Wvla]
   34 |         char block[fmt_len + 1];
      |         ^~~~

/home/andypugh/linuxcnc-dev/src/emc/kinematics/genserfuncs.c: In function ‘genser_kin_jac_inv’:
/home/andypugh/linuxcnc-dev/src/emc/kinematics/genserfuncs.c:271:1: warning: the frame size of 2112 bytes is larger than 2048 bytes [-Wframe-larger-than=]
  271 | }
      | ^

/home/andypugh/linuxcnc-dev/src/emc/kinematics/genserfuncs.c: In function ‘genserKinematicsInverse’:
/home/andypugh/linuxcnc-dev/src/emc/kinematics/genserfuncs.c:564:1: warning: the frame size of 2576 bytes is larger than 2048 bytes [-Wframe-larger-than=]
  564 | }
      | ^

/home/andypugh/linuxcnc-dev/src/hal/drivers/mesa-hostmot2/sserial.c: In function ‘hm2_sserial_get_param_value’:
/home/andypugh/linuxcnc-dev/src/hal/drivers/mesa-hostmot2/sserial.c:389:17: warning: ISO C90 forbids variable length array ‘buf’ [-Wvla]
  389 |                 char buf[g->DataLength/8];
      |                 ^~~~

hal/components/mesa_pktgyro_test.comp: In function ‘extra_setup.constprop’:
hal/components/mesa_pktgyro_test.comp:288:1: warning: the frame size of 2400 bytes is larger than 2048 bytes [-Wframe-larger-than=]

```

6 warnings is fewer than I think i have ever seen.

BsAtHome:
Yes, these are the warnings left.

The variable length warnings need to be looked into. These are probably a bit harder to fix, but I need to investigate. Most other warnings were trivially solved.

The frame size warnings are also problematic. The local variables of the functions in question need to be analyzed to see if they can be reduced. If not, then the stack frame size limit must be increased. But that should be last resort.

The one in hal/components/mesa_pktgyro_test.comp was even larger until I declared  a few (constant data) arrays as static. This also relates to a bad declaration in a uart send routine, which should take a const char pointer, but does not. There are more places where this is the case and changing them optimized the code and data.

andypugh:
If necessary pktgyro can probably be dropped. I don't think it's especially useful in CNC, but was written for testing the Mesa packet UART module (and a gyro was a cheap test device)

BsAtHome:
Actually, I've fixed all warnings. Changes are committed but must run CI before I make a PR.

The frame sizes cannot be reduced. They are mostly a necessity of the calculations (kinematics has matrices and vectors in there). These are bound by an added pragma.

The pktgyro exposed something else. There are many function that have non-const parameters which should be declared const. This gave the opportunity to fix one of them ;-)

PR #3301, opened 2025-01-28
tests-hardening: addressing shellcheck errors in our tests

smoe:
These changes shall prepare LinuxCNC's test routines to accept blanks in file names.

### DIFF
--- a/scripts/cppcheck.sh
+++ b/scripts/cppcheck.sh
@@ -25,19 +25,19 @@ CPPCHKCC="$CPPCHKOPT $CCSTD"
 CPPCHKCX="$CPPCHKOPT $CXSTD"
 
 # Do this from the source directory
-cd "$(dirname "$0")/../src"
+cd "$(dirname "$0")/../src" || (echo "E: Could not return to source directory" ; exit 1 )
 
 docheck() {
     files=$(find "$1" -maxdepth 1 \( -name "*.c" -o -name "*.h" \) )
-    [ ! -z "$files" ] && cppcheck -I"$1" $CPPCHKCC $files
+    [ -n "$files" ] && cppcheck -I"$1" "$CPPCHKCC" "$files"
 
     files=$(find "$1" -maxdepth 1 \( -name "*.cc" -o -name "*.hh" \) )
-    [ ! -z "$files" ] && cppcheck -I"$1" $CPPCHKCX $files
+    [ -n "$files" ] && cppcheck -I"$1" "$CPPCHKCX" "$files"
 }
 
 # *** HAL files ***
 echo "I (1/4): checking HAL folders with both C and C++ code"
-for d in $(find hal/ -type d -not -name "*__pycache__")
+find hal/ -type d -not -name "*__pycache__" | while read -r d
 do
     # Don't care about examples
     case "$d" in hal/user_comps/mb2hal/examples*) continue;; esac
@@ -48,7 +48,7 @@ done
 
 # *** EMC files ***
 echo "I (2/4): checking EMC folders with both C and C++ code"
-for d in $(find emc/ -type d -not -name "*__pycache__")
+find emc/ -type d -not -name "*__pycache__" | while read -r d 
 do
     # Will give Tcl problems
     case "$d" in emc/usr_intf/axis/extensions*) continue;; esac
@@ -59,7 +59,7 @@ done
 
 # *** NML files ***
 echo "I (3/4): checking LIBNML folders with both C and C++ code"
-for d in $(find libnml/ -type d -not -name "*__pycache__")
+find libnml/ -type d -not -name "*__pycache__" | while read -r d
 do
     echo "I (3/4): checking $d"
     docheck "$d"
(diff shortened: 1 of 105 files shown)


BsAtHome:
Darn! I have many of the same changes in my tree.

However, there are many more and a few fixes. See https://github.com/LinuxCNC/linuxcnc/compare/master...BsAtHome:linuxcnc:fix_shellcheck for a diff. I'm still missing a few.

Notably scripts/runtests.in has added left-over shared memory check and tests/halcompile/serial-out-of-tree/checkresult fixes testing the correct target.

BsAtHome:
And, BTW, these changes are not yet enough to run with spaces in the path.

There are a could of hard ones that, for example, expand compiler flags that can contain paths which are not escaped. Expanding it cannot use quotes because there are individual arguments and that will split an embedded path too. The right way is to use arrays, but that needs to be done consistently throughout the code. That will be very hard.

BsAtHome:
in, for example, cppcheck.sh:
```bash
# Do this from the source directory
cd "$(dirname "$0")/../src" || (echo "E: Could not return to source directory" ; exit 1 )
```

You cannot use <code>( )</code> because it runs in a sub-shell. You need to use <code>{ }</code> to exit the <i>current</i> shell, like in:
```bash
# Do this from the source directory
cd "$(dirname "$0")/../src" || { echo "E: Could not return to source directory"; exit 1; }
```

BsAtHome:
in scripts/cppcheck.sh:
```bash
find emc/ -type d -not -name "*__pycache__" | while read -r d
...
```
This will fail because read does word-splitting. If there are any spaces in the path then it fails. See https://github.com/BsAtHome/linuxcnc/blob/fix_shellcheck/scripts/cppcheck.sh how it is fixed there.

BsAtHome:
In scritps/swich.sh
```bash
SRCDIR=$(cd "$MYDIR/../src" && pwd || exit 1)
```
This is just plain ugly and the exit is run in the sub-shell instead of the current shell. May I suggest:
```bash
SRCDIR=$(realpath "$MYDIR/../src")
[ -d "$SRCDIR" ] || exit 1
```

BsAtHome:
Disabling quote warning:
# shellcheck disable=SC2086

This is a seriously bad idea in $CPPFLAGS (f.ex. in tests/build/header-sanity/test.sh) to disable the warning. The CPPFLAGS variable may actually include paths that contain spaces. It needs to be fixed at a project-wide scale and requires arrays.

BsAtHome:
Generally: For each added shellcheck disable I suggest you add a comment <i>why</i> it is done.

In tests/overrun/test.sh:
```bash
trap "rm -rf '$TMPDIR'" 0 1 2 3 15
# not trapping 9 since it cannot be trapped
```
Using single quotes inside will not expand TMPDIR when the trap executes. The outside must be single quoted then the inside will evaluate when executed. Pedantic note: the comment is not really useful.


In tests/tooledit/test.sh
```bash
if [ "$(wc -l tclerror.log)" -ne 0 ]; then
```
It should read <code>$(wc -l < tclerror.log)</code> because wc on a file outputs multiple fields. You only want it to return a number for which you need read from stdin, hence the redirect.

smoe:
Well done. Thank you very much for your exceptional review. Shall I proceed by fixing my changes accordingly and then rebase everything on your branch followed by a PR against it?

smoe:
> Disabling quote warning: # shellcheck disable=SC2086
> 
> This is a seriously bad idea in $CPPFLAGS (f.ex. in tests/build/header-sanity/test.sh) to disable the warning. The CPPFLAGS variable may actually include paths that contain spaces. It needs to be fixed at a project-wide scale and requires arrays.

Do you have a good pointer to how it is done properly? I have undone the disabling of SC2086 and am with you the need for arrays. But with "make" that is a bit difficult, right?

BsAtHome:
> > Disabling quote warning: # shellcheck disable=SC2086
> > This is a seriously bad idea in $CPPFLAGS...

> Do you have a good pointer to how it is done properly? I have undone the disabling of SC2086 and am with you the need for arrays. But with "make" that is a bit difficult, right?

Yes and no. In principle it is easy where you can use mapfile or read (I used it several places in my patches).
```bash
# using IFS word-splitting:
read -r -a ARRAYVAR < <(echo word word word)

# using full input lines (-t removes trailing newline):
mapfile -t ARRAYVAR < <(find "$SOMEPATH"  -name "*.c")
```

The real challenge is in the input values from third party sources. For example, path building in the configure step may already include spaces in paths. If these end up in a statement like <code>CFLAGS=$CFLAGS "-IPath With Space"</code>, then you are in for a real pain later on when expanding CFLAGS. This goes for any constructed option variable.

I do not "see" the right thing to do, so I kept the shellcheck warnings to remind us that it needs to be addressed. We even may need to revert the "$HEADERS" expansion because it states HEADER<i><b>S</b></i> (plural) and that may require it to be unquoted. But I have not yet checked the definition(s) of HEADERS to see how it is constructed.

smoe:
https://github.com/BsAtHome/linuxcnc/pull/1

andypugh:
What's the status of this? It is showing as still WIP.

BsAtHome:
This has been fixed in the other PRs.

PR #3302, opened 2025-01-30
Fix shellcheck trivial double quote

BsAtHome:
This fixes the first part of shellcheck problems with help from @smoe.
The current set is almost exclusively small changes of double quote issues.

### DIFF
--- a/bin/profile_axis
+++ b/bin/profile_axis
@@ -1,4 +1,4 @@
 #!/bin/bash
-SCRIPT_LOCATION=$(dirname $(readlink -f $0))
-python3 -mcProfile -o axis.profile $SCRIPT_LOCATION/axis "$@"
+SCRIPT_LOCATION=$(dirname "$(readlink -f "$0")")
+python3 -mcProfile -o axis.profile "$SCRIPT_LOCATION"/axis "$@"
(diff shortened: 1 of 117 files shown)


smoe:
Maybe just rebase and squash the two commits into one?

BsAtHome:
It turns out that there are a lot of subtle problems (see https://github.com/LinuxCNC/linuxcnc/pull/3302/commits/150b369929b98559238d95e6c017e30fc3aa4e53). That is why I opted to divide the patches into multiple groups.

Your changes have been merged into my tree (thanks for your work!), but the lot gets too difficult to wade through and verify. There are 80 more files that are changed in my tree and they are of increasing patch size.

smoe:
My previous comment may not have been clear. From what I see you have just two commits. And the second commit is undoing something from the first commit. If you squash the two into one commit then it all becomes a bit simpler. Not much simpler. But simpler :-)  The patches are just all so nice and important ( I think ) that I felt like I drop that suggestion so the PR better reflects the inner beauty.

BsAtHome:
> ...so the PR better reflects the inner beauty.

Well, yes, there is that...

But that said... there is quite an amount of (local) commits and undo/redo you don't see anymore. Just one slip-up before in the final pass. I think I'll let that pass ;-)

PR #3304, opened 2025-01-30
Fix more shellcheck problems

BsAtHome:
This PR is mostly about  <code>cd || exit</code> because cd may fail. Also includes missing double quotes and few other simple changes. The file-set is primarily config/demo related. The patch-set was created in cooperation with @smoe.

### DIFF
--- a/configs/apps/xhc-hb04/xhc-hb04.sh
+++ b/configs/apps/xhc-hb04/xhc-hb04.sh
@@ -2,7 +2,7 @@
 
 thisfile=$(readlink -f "$0")
 thisdir=$(dirname "$thisfile")
-cd "$thisdir" || exit
+cd "$thisdir" || { echo "E: Could not change directory to '$thisdir'"; exit 1; }
 
 LAYOUT=${1:-""}
 
--- a/configs/apps/xhc-hb04/xhc-hb04.sh
+++ b/configs/apps/xhc-hb04/xhc-hb04.sh
@@ -2,14 +2,14 @@
 
 thisfile=$(readlink -f "$0")
 thisdir=$(dirname "$thisfile")
-cd "$thisdir"
+cd "$thisdir" || exit
 
 LAYOUT=${1:-""}
 
-[ -x "$(which $COLORTERM)" ] && TERM=$COLORTERM
-[ -z $TERM ] && TERM=xterm
+[ -x "$(which "$COLORTERM")" ] && TERM="$COLORTERM"
+[ -z "$TERM" ] && TERM=xterm
 # when invoked from Application menu and RIP, may have TERM=dumb
-[ x"$TERM" = xdumb ] && TERM=xterm
+[ "$TERM" = "dumb" ] && TERM=xterm
 
 XHC_HB04=$(which xhc-hb04)
 HALLIB_DIR=$(linuxcnc_var HALLIB_DIR)
@@ -20,7 +20,7 @@ case $LAYOUT in
 esac
 dashI="-I $CFG"
 
-if [ ! -z "$debug" ] ; then
+if [ -n "$debug" ] ; then
   echo "COLORTERM=$COLORTERM"
   echo "TERM=$TERM"
   echo "XHC_HB04=$XHC_HB04"
--- a/scripts/make-docs-pdf-index
+++ b/scripts/make-docs-pdf-index
@@ -9,19 +9,23 @@ set -e
 # --show-toplevel` because when building from dsc we don't have a
 # git repo.
 TOPLEVEL=${PWD}/..
-cd ${TOPLEVEL}/docs
+cd "${TOPLEVEL}"/docs
 
-TITLE="LinuxCNC PDF docs ($(cat ${TOPLEVEL}/VERSION))"
+TITLE="LinuxCNC PDF docs ($(cat "${TOPLEVEL}"/VERSION))"
 
 rm -f index.html
 
-echo "<html>" >> index.html
-echo "<head><title>${TITLE}</title></head>" >> index.html
-echo "<body><h1>${TITLE}</h1><p>" >> index.html
+{
+    echo "<html>"
+    echo "<head><title>${TITLE}</title></head>"
+    echo "<body><h1>${TITLE}</h1><p>"
+} >> index.html
 
-for F in $(ls -1 *.pdf | sort); do
+for F in $(ls -1 ./*.pdf | sort); do
     echo "    <a href=\"${F}\">$F</a><br>" >> index.html
 done
 
-echo "</body>" >> index.html
-echo "</html>" >> index.html
+{
+    echo "</body>"
+    echo "</html>"
+} >> index.html
--- a/scripts/pyvcp_demo
+++ b/scripts/pyvcp_demo
@@ -108,7 +108,6 @@ done
 cd "$(dirname "$HALFILE")" || { echo "E: Could not change directory to '$(dirname "$HALFILE")'"; exit 1; }
 
 halmsg=/tmp/pyvcp_demo_halcmd.err
-true >| "$halmsg"
 halcmd -f "$HALFILE" > "$halmsg" 2>&1
 halstatus=$?
 if [ "$halstatus" != 0 ] ; then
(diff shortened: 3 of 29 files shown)


configs/apps/xhc-hb04/xhc-hb04.sh
```diff
@@ -2,14 +2,14 @@
 
 thisfile=$(readlink -f "$0")
 thisdir=$(dirname "$thisfile")
-cd "$thisdir"
+cd "$thisdir" || exit
```
smoe, line 5: The text is missing here.

scripts/make-docs-pdf-index
```diff
@@ -9,19 +9,23 @@ set -e
...
 rm -f index.html
 
-echo "<html>" >> index.html
-echo "<head><title>${TITLE}</title></head>" >> index.html
-echo "<body><h1>${TITLE}</h1><p>" >> index.html
+{
+    echo "<html>"
+    echo "<head><title>${TITLE}</title></head>"
+    echo "<body><h1>${TITLE}</h1><p>"
+} >> index.html
```
smoe, line 22: Nice!

scripts/pyvcp_demo
```diff
@@ -74,49 +74,52 @@ case $# in
...
 done
 
-cd $(dirname "$HALFILE") ;# so HALFILE can source relative files
+# so HALFILE can source relative files
+cd "$(dirname "$HALFILE")" || { echo "E: Could not change directory to '$(dirname "$HALFILE")'"; exit 1; }
 
 halmsg=/tmp/pyvcp_demo_halcmd.err
->|$halmsg
-halcmd -f $HALFILE >$halmsg 2>&1
+true >| "$halmsg"
```
smoe, line 111: Maybe leave a comment here explaining that this command just truncates the $halmsg file?

BsAtHome: Actually, the whole <code>>|$halmsg</code> is useless. The next line uses '>' as redirection char and that truncates $halmsg anyway.

scripts/make-docs-pdf-index
BsAtHome: I stole (inspiration) from experts :-)

scripts/pyvcp_demo
BsAtHome: BTW, I assume that noclobber is off, which is normally the case.
And, I cannot see any <code>set -C</code> or <code>set -o noclobber</code> in the script.

configs/apps/xhc-hb04/xhc-hb04.sh
BsAtHome: Fixed in https://github.com/LinuxCNC/linuxcnc/pull/3304/commits/eec772f1d40c77402accea0686e6d3dc7dfc66af

scripts/pyvcp_demo
BsAtHome: Fixed in https://github.com/LinuxCNC/linuxcnc/pull/3304/commits/7439f102f5135623a051c03ed8c0dcd04d2d5716

smoe: This is not uncommon to be set in the user's .bashrc.

scripts/make-docs-pdf-index
smoe: shellcheck proposes it, but I was like "so what". Now I really like it, though.

scripts/pyvcp_demo
BsAtHome: Doesn't matter. The redirection just below the truncate will fail anyway.

Below example result in an error:
```bash
$ set -o noclobber
$ >| junkfile
$ echo hello > junkfile
bash: junkfile: cannot overwrite existing file
```

smoe: That was unexpected. But this also is a nice reminder that ">| $halmsg" mail fail, which again suggests that we test if that was successful and somehow react properly. I lack the overview about where we are in the process of operating a CNC machine, so maybe an exit is not an option, but something like

```
true >| "$halmsg" || { echo "W: Could not write to '$(realpath "$halmsg")' ; exit 1; }
```
sounds like a natural consequence since later attempts to add output will then also fail.

The shell scripts of ours typically do not set the "-e" flag (have not checked this one), so they will continue in case of an error, so exiting here may change the behaviour of the script in case of an error.

smoe: Er - actually, if `true >| "$halmsg"` fails because of the noclobber then adding extra lines to the output will not fail.

BsAtHome: The >| doesn't fail. The normal redirect fails, which is the next line under the truncate. That is also how the scripts/pyvcp_demo script has it. I just reduced the example to the essentials.

This scripts/pyvcp_demo is for demo. There would have been complaints if anybody has noclobber enabled and ran this script because it doesn't work as expected.

The point is that the next line in the script: <code>halcmd -f "$HALFILE" > "$halmsg" 2>&1</code> fails if noclobber is active, regardless of what happened before, and that means nothing works as expected. The whole truncate is superfluous.

smoe: Sorry, this got me wrong when I first read this. 

```[bash]
$ set -o noclobber
$ echo > a && echo ok || echo failed
ok
$ echo > a && echo ok || echo failed
$ echo > a && echo ok || echo failed
-bash: a: cannot overwrite existing file
failed
```
So you can create a new file, but not create it again since noclobber is set.

```
$ echo >> a && echo ok || echo failed
ok
```
But you can always append content to it - which is what the script is doing.
```[bash]
mc00023497:~ u005069$ echo >| a && echo ok || echo failed
ok
mc00023497:~ u005069$ echo >> a && echo ok || echo failed
ok
```
This combination ensures that you append content to an empty file, also when noclobber is set.

So I think this is just fine.

BsAtHome: > ... But you can always append content to it - which is what the script is doing. ...

No, it is not appending in the script. That is the whole point of my objection.

The relevant part in the original file:
```bash
...
halmsg=/tmp/pyvcp_demo_halcmd.err
>|$halmsg
halcmd -f $HALFILE >$halmsg 2>&1
halstatus=$?
if [ $halstatus != 0 ] ; then
...
```
What it does:
* set halmsg var to a filename
* truncate the file
* (try to) run halcmd and redirect the output to the file including anything from stderr
* save the status
* the following if [] on basis of the status is <i>always</i> true if noclobber is set because the redirection failed with noclobber set and halcmd was <i>never</i> executed.

What is left is the following:
a) fix noclobber, in which case we can remove the superfluous truncate
b) just remove the superfluous truncate and don't care about noclobber

The truncate is in both cases superfluous. Also, I opt for scenario b). If anyone ran this code previously, then they would already have had problems. Nobody complained.

smoe: I suggest to leave the >| in and then indeed append to the file:
```[diff]
- halcmd -f $HALFILE >$halmsg 2>&1
+ halcmd -f "$HALFILE" >> "$halmsg" 2>&1
```

BsAtHome: Maybe @andypugh can resolve the choice?

BsAtHome: Maybe, as a compromise, we can both have our wish.

What about:
```diff
- >|$halmsg
- halcmd -f $HALFILE >$halmsg 2>&1
+ halcmd -f "$HALFILE" >| "$halmsg" 2>&1
```

I still think noclobber is an abomination, first saying <i>"you are not allowed to create files"</i> and then, with a new operator say <i>"well,... if you really insist, I'll let you create files anyway"</i>.

smoe: > Maybe, as a compromise, we can both have our wish.

:)

> What about:
> 
> ```diff
> - >|$halmsg
> - halcmd -f $HALFILE >$halmsg 2>&1
> + halcmd -f "$HALFILE" >| "$halmsg" 2>&1
> ```
> 
> I still think noclobber is an abomination, first saying "you are not allowed to create files" and then, with a new operator say "well,... if you really insist, I'll let you create files anyway".

Yip. Preferable would be something like 
```
if [ -f "$halmsg" ]; then
    mv "$halmsg" "$halmsg".$( ls -l "$halmsg" -D "%Y%m%d" | awk '{print $6}' )
fi
```

I kind of like the >| since the semantics "I write my log to this file no matter what", though. Just because someone set the noclobber flag should not stop you from running the CNC machine. But noclobber saves you from unintentionally overwriting your G-code.

BsAtHome: [snip]
> Yip. Preferable would be something like...

[snip]

This is a <i>demo</i> script and is not for production. Lets not go overboard.

PR #3307, opened 2025-01-31
Fix shellcheck various

BsAtHome:
This PR contains various updated scripts, mainly double quotes and $() instead of back-ticks. The patch-set was created in cooperation with @smoe.
The cppcheck.sh script has a more complex while/read loop to allow paths to be handled properly. Also options and file lists are handled in arrays to preserve the path in its entirety.
The docs/src/asciideps routine adds newline-to-space translation because quoted expansions are no longer word-split using IFS. Each individual file path in the collection of files in the variable is considered whitespace-free because the paths are relative to LCNC's topdir and it is assumed that no path or file name in the LCNC hierarchy contains white-space.

### DIFF
--- a/docs/src/checkref
+++ b/docs/src/checkref
@@ -10,24 +10,21 @@ fi
 LANGUAGE="$1"
 shift
 
-FILES="$@"
-
-if [ -z $(which linuxcnc-checklink) ]; then
+if [ -z "$(which linuxcnc-checklink)" ]; then
     echo "ERROR: checklink not found, install w3c-linkchecker for HTML link validation" 1>&2
     exit 1
 fi
 
 
 BAD_LINKS=0
-for F in $FILES; do
-    OUT=.checklink.$LANGUAGE.$(basename $F).tmp
-    rm -f $OUT
-    linuxcnc-checklink --quiet --exclude "(http|https|irc)://" $F  2>&1 | tee $OUT
-    grep -E -q 'List of (broken links and other issues|duplicate and empty anchors)' $OUT
-    if [ $? -ne 1 ]; then
+for F in "$@"; do
+    OUT=.checklink.$LANGUAGE.$(basename "$F").tmp
+    rm -f "$OUT"
+    linuxcnc-checklink --quiet --exclude "(http|https|irc)://" "$F"  2>&1 | tee "$OUT"
+    if grep -E -q 'List of (broken links and other issues|duplicate and empty anchors)' "$OUT"; then
         BAD_LINKS=1
     fi
-    rm -f $OUT
+    rm -f "$OUT"
 done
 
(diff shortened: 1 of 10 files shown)


docs/src/checkref
```diff
@@ -10,24 +10,21 @@ fi
...
 
 
 BAD_LINKS=0
-for F in $FILES; do
-    OUT=.checklink.$LANGUAGE.$(basename $F).tmp
-    rm -f $OUT
-    linuxcnc-checklink --quiet --exclude "(http|https|irc)://" $F  2>&1 | tee $OUT
-    grep -E -q 'List of (broken links and other issues|duplicate and empty anchors)' $OUT
-    if [ $? -ne 1 ]; then
+for F in "$@"; do
```
smoe, line 20: I was shying away from this - it works, though:

In the script
```[bash]
#!/bin/bash                                                                                                                                                               
for i in "$@"
do
    echo "Hello '$i'"
done
```
it performs as it should, keeping blanks in check:
```
$ ./s.sh "a" "b" "a b"
Hello 'a'
Hello 'b'
Hello 'a b'
```
without the quotes, it would be
```
$ ./s.sh "a" "b" "a b"
Hello 'a'
Hello 'b'
Hello 'a'
Hello 'b'
```
which is a bummer and likely also a security concern.

scripts/cppcheck.sh
```diff
@@ -13,64 +13,64 @@ fi
 # See if cppcheck accepts --check-level
 EXHAUSTIVE=$(cppcheck --check-level=exhaustive --version > /dev/null 2>&1 && echo "--check-level=exhaustive")
 
-CPPCHKOPT="-j $nproc --force $EXHAUSTIVE"
-CPPCHKOPT="$CPPCHKOPT --enable=warning,performance,portability"
-CPPCHKOPT="$CPPCHKOPT -I$(realpath "$(dirname "$0")/../include")"
+CPPCHKOPT=( -j "$nproc" --force "$EXHAUSTIVE" )
```
smoe, line 16: Nice!

docs/src/checkref
BsAtHome: The "$@" actually bit me in PR #3302 scripts/halrun (where we had a discussion). The problem there is that the script manipulates the arguments in the script using <code>set -- ...</code> and relies on word splitting later on. This is wrong, but requires much more work to solve.

The case you mention here was clearly the right call to use "$@" because the word splitting must not be performed on the argument itself. The "$@" expands to individual quoted arguments, which is what you want (just like your example here).

smoe (approved):
Well done!

scripts/cppcheck.sh
BsAtHome: Just like normal development. You do 10 stupid things and then do the right thing (by accident).
;-)

PR #3315, opened 2025-02-03
Fix last -Wextra warnings

BsAtHome:
This PR solves the last few warnings when compiling with -Wextra:
* one signed/unsigned comparison,
* one unused parameter,
* three instances where intentional switch case fall-through is made explicit,
* four obviously missing breaks in switch cases.

It runs clean with both gcc 14.2.1 and clang 18.1.8. It still needs to be verified on older Debian instances, after which we should be able to add -Wextra in configure in a future PR.

### DIFF
--- a/src/emc/motion/stashf_wrap.h
+++ b/src/emc/motion/stashf_wrap.h
@@ -42,7 +42,7 @@ const char *fmt, *efmt;
 	// after the format.
         char block[63 + 1];
         int fmt_len = efmt - fmt;
-	if(fmt_len >= sizeof(block))
+	if(fmt_len >= (int)sizeof(block))
 		fmt_len = sizeof(block) - 1;
         memcpy(block, fmt, fmt_len);
         block[fmt_len] = 0;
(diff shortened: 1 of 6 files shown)


src/emc/motion/stashf_wrap.h
```diff
@@ -42,7 +42,7 @@ const char *fmt, *efmt;
 	// after the format.
         char block[63 + 1];
         int fmt_len = efmt - fmt;
-	if(fmt_len >= sizeof(block))
+	if(fmt_len >= (int)sizeof(block))
```
andypugh, line 45: Would it be better to change the type of fmt_len?

BsAtHome: Formally, the difference between pointers is an integer. Therefore, the statement in line 43 results in a signed integer value.

The code makes sure that <code>fmt_len</code> will be larger or equal zero because the value of <code>efmt</code> is a search from <code>fmt</code> and beyond (and cannot be NULL or it would have dropped out of the while()).

In that sense, either cast, signed at RHS or implicit by type-change unsigned at LHS, would work.

andypugh: OK, so what you have there is probably better.

PR #3316, opened 2025-02-03
Remove class instance passed by value and make then pass by const reference

BsAtHome:
There were considerable amount of places where complex structures/classes were passed as values to functions. These create unnecessary copy operations and impede performance. The correct way is to use const pass-by-reference. The issue was detected by cppcheck (passedByValue).

This PR changes all the instances from pass-by-value to const pass-by-reference. There were only few place where a parameter was altered in a function:
* An inch/mm conversion was demultiplexed instead of altering the passed parameter.
* There were instances where a parameter passed vector was cleared before exiting a function. This has no effect and the clear was removed.

### DIFF
--- a/src/emc/nml_intf/canon.hh
+++ b/src/emc/nml_intf/canon.hh
@@ -480,84 +480,84 @@ extern void STRAIGHT_FEED(int lineno,
 
 extern std::vector<unsigned int> nurbs_G5_knot_vector_creator(unsigned int n, unsigned int k);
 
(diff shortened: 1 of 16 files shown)


andypugh:
This seems to be largely in the NURBS code. This was written as a student project, which might explain why the programmer was unaware that what they were doing was suboptimal. 

I suspect that the NURBS code is rarely used (certainly no CAM  system is aware of it) so I have very few qualms about merging it with only a fairly limited inspection.

PR #3318, opened 2025-02-04
Fix undefined signed 32-bit integer signed shift left by 31

BsAtHome:
Integers are 32 bit by default. Performing left shift of 31 on signed integers is undefined by the standard because it is an overflow condition. This happens in statements like <code>1 << 31</code>. The fix is to state them as unsigned: <code>1u << 31</code>.


andypugh:
Good catch.

BsAtHome:
Found by cppcheck...

Issue #3296, opened 2025-01-23
C++ standard to use


BsAtHome:
The configure script tests for c++17 (<code>AX_CXX_COMPILE_STDCXX(17, , mandatory)</code> in <code>src/configure.ac:42</code>. However, the <code>src/Makefile:258</code> sets the standard to gnu++20.

If the standard is supposed to be c++20 (with gnu extensions), then it must be tested in configure. For the current build, I suggest to use <code>--std=c++17</code> to be compatible with clang and change the Makefile accordingly.

Other suggestions?

rene-dev:
Its supposed to be 20, please change the configure script accordingly.

rmu75:
so can we finally use std::span and std::format?

BsAtHome:
> Its supposed to be 20, please change the configure script accordingly.

Debian Bullseye, which is a supported platform afaik, uses g++ 10.2, which does not fully support c++20. It is however c++17 feature complete. That would suggest that setting the standard to c++20 is inappropriate.

rmu75:
We should stay with c++20 even if it is not completely supported on current build platforms, things like concepts, imporved constexpr and consteval are useful language features supported by even gcc-10. https://gcc.gnu.org/projects/cxx-status.html

IIRC concepts are already used, so that ship has sailed anyways...

andypugh:
> Debian Bullseye, which is a supported platform afaik, uses g++ 10.2, which does not fully support c++20. It is however c++17 feature complete. That would suggest that setting the standard to c++20 is inappropriate.

http://wiki.linuxcnc.org/cgi-bin/wiki.pl?MinimumSoftwareVersions agrees with you. 

But then the question is whether we want to continue to support Bulleye. Master was supported in Buster until Rene changed it during the Norway meetup last year (or the year before...)

Trixie will almost certainly be out before 2.10 is released.

rmu75:
Is there a scenario where bullseye works on a machine and upgrade to bookworm would not be possible?

smoe:
Once the shellcheck-triggered changes are in, I guess more work on cppcheck will be addressed. Maybe the one or other C++-17/20 feature will sneak in while reviewing all that code.

BsAtHome:
> Once the shellcheck-triggered changes are in, I guess more work on cppcheck will be addressed. Maybe the one or other C++-17/20 feature will sneak in while reviewing all that code.

[...]

PR #3326, opened 2025-02-06
Use C++20 as compile standard

BsAtHome:
This PR changes the required C++ standard to C++20, as discussed in #3296. The consequence is that Bullseye/Debian 11 is dropped as supported distribution and removed from CI.

Also fixed in this PR:
* Drop useless compiler-flag test in configure. It is never added to CFLAGS/CXXFLAGS anyway due to a spelling error (<code>flag</code> vs. <code>commonflag</code>). Regardless, the tested compiler-flag is unnecessary because the code has already been fixed (and the test always failed on clang). If compiler-flags are required you should add CFLAGS= or CXXFLAGS= to the configure command-line.
* Remove a contraction in an explanatory text (Don't -> Do not) because vim has trouble syntax highlighting beyond it.
* Always add compiler-flag <code>-Wextra</code>. We are currently completely warning free. Lets keep it that way.
* Add a configure option <code>--enable-werror</code> to set the <code>-Werror</code> compiler-flag. In the near future this should be added to CI's rip-and-test and rip-and-test-clang tests to stay warning free.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
     runs-on: ubuntu-24.04
     strategy:
       matrix:
-        image: ["debian:bullseye", "debian:bookworm", "debian:sid"]
+        image: ["debian:bookworm", "debian:sid"]
     container:
       image: ${{ matrix.image }}
       # IPC_OWNER is needed for shmget IPC_CREAT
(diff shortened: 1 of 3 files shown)


SebKuzminsky:
I think this broke builds on buildbot2, eg http://buildbot2.highlab.com/buildbot/#/builders/7/builds/3289

Bullseye errors out because it doesn't have c++20, so the builds fail and no debs are produced. 

Probably someone needs to update `scripts/platform-is-supported`.

BsAtHome:
And, while you are at it,... I don't think i386 is supported anymore and arm is probably either arm64 or aarch64. It also detects xenomai but will report it as unsupported.

There may be more to fix in the script.

PR #3327, opened 2025-02-06
Remove duplicated non-virtual base-class method

BsAtHome:
A non-virtual method cannot be "properly" overloaded. It is also a pointless construct because the base-class method claims it should never be called. Therefore, it should not exists at all, resulting in it never being called (and the compiler will error if you try).

This particular method cannot be made virtual because of the vtable, which cannot be located in shared memory (pointers are not valid across shmem maps). Anyway, there is no point in having it at all in the base-class. Derived classes must declare and define it where necessary.

Cppcheck found this issue [duplInheritedMember] and this PR is the first part of fixing the this problem class. This PR roughly halves the reported instances of the warning.

### DIFF
--- a/src/libnml/nml/nmlmsg.cc
+++ b/src/libnml/nml/nmlmsg.cc
@@ -95,12 +95,3 @@ void NMLmsg::clear()
 	size = sizeof(NMLmsg);
     }
 }
-
-/* Error message stub for base class. */
- /* update should only be called with derived classes. */
-void NMLmsg::update(CMS * cms)
-{
-    rcs_print_error("update called for NMLmsg base class.");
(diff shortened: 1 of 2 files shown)


andypugh:
It looks like it exists to trap improper usage? All it does is say "you shouldn't have called me".

BsAtHome:
And that is exactly why it is pointless.
If it was a <i>virtual</i> function, then it would have made some sense (making it abstract would have been better). However, it is a normal method and it cannot be virtual due to its place in shared memory.

Therefore, if you don't want this base-class method to get called, ever, then you simply don't include it in the base (!) class.

PR #3330, opened 2025-02-07
gmoccapy: fix increasing window size on edit (2.9)

hansu:
For some translations the text for the search/replace buttons is too wide to fit the smallest window size. The result is that the window width increases in edit mode:

gmoccapy_edit_de_en

gmoccapy-resize-edit

Changing the text to icons for the undo and redo buttons resolves this:

gmoccapy_edit_new_de

@gmoccapy Are you okay with the button solution or do you have a better proposal?
For the master branch I thought to optimize this further, maybe similar to how it's done in [kate](https://github.com/user-attachments/assets/daba5fbe-34c3-4dc5-a700-9699bf2e7a66)


gmoccapy:
Hallo Hans,

i like the solution with the icon button.
They look better than the text solution i made.

So go ahead!

Norbert

PR #3331, opened 2025-02-07
Fix class initializations and suppress duplicate inherited update() message

BsAtHome:
This PR mainly covers all missing class member initialization in the class constructor. It uses init-lists to do so and several places replaces old-style in-constructor assignment with init-list initialization.

Secondly, the PR fixes (suppresses) the duplicate inherited method message on update() as detected by cppcheck. These messages are caused by the sub-class containing the same method as the base-class. However, the sub-class implementation of update() calls the base-class update() correctly. The reason for the construct is because a virtual method is not possible for the set of classes. The scripts/cppcheck.sh script is modified to enable inline suppression of cppcheck messages.

A few instances where the copy-constructor is disabled have been modernized to use <code>= delete</code> syntax on the copy-constructor and copy-assignment.


smoe (approved):

smoe:
Impressive. Many thanks!

BsAtHome:
> Impressive. Many thanks!

Janitorial work... Tedious but has to be done once in a while ;-)

PR #3334, opened 2025-02-09
Fix various problems and bugs detected by cppcheck

BsAtHome:
This PR has an assortment of fixes for cppcheck detected problems and some are real bugs. Fixes include:
* The hal/drivers/mesa-hostmot2/sserial.c had a 5 year old warning "fix" that demoted 64-bit values to int. That fix ignored the actual use-case and resulted in a 32-bit value being shifted by 48 and failing the intended test.
* Several instances had array index tests that were performed <i>after</i> the array index dereference. These tests needed to be reversed.
* Pushing NULL as a (string list) terminating argument in an vararg function may fail because NULL is not guaranteed to be a pointer. It must be force-cast to pointer to become a proper terminator.
* Multiple places where variables were not properly initialized.
* realloc() can fail and it will leak memory if the argument pointer is reassigned before testing.
* Several cppcheck false positives or intentional constructs were marked to have the message suppressed.
* Limit sscanf %s format to the argument buffer size minus one (like in <code>%255s</code>).
* A buffer free intended to clear the pointer variable but was missing a dereference.

### DIFF
--- a/src/emc/ini/initraj.cc
+++ b/src/emc/ini/initraj.cc
@@ -280,7 +280,7 @@ static int loadTraj(EmcIniFile *trajInifile)
                     continue;    // position t at index of next non-zero mark
                 }
                 // there is a mark, so read the string for a value
-                if (1 == sscanf(&homes[len], "%s", home) &&
+                if (1 == sscanf(&homes[len], "%255s", home) &&
                     1 == sscanf(home, "%lf", &d)) {
                     // got an entry, index into coordinateMark[] is 't'
                     if (t == 0)
--- a/src/emc/ini/initraj.cc
+++ b/src/emc/ini/initraj.cc
@@ -295,12 +295,15 @@ static int loadTraj(EmcIniFile *trajInifile)
                         homePose.b = d;
                     else if (t == 5)
                         homePose.c = d;
+/*
+ * The following have no effect. The loop only counts [0..5].
                     else if (t == 6)
                         homePose.u = d;
                     else if (t == 7)
                         homePose.v = d;
                     else
                         homePose.w = d;
+*/
 
                     // position string ptr past this value
                     len += strlen(home);
--- a/src/emc/ini/initraj.cc
+++ b/src/emc/ini/initraj.cc
@@ -280,7 +280,7 @@ static int loadTraj(EmcIniFile *trajInifile)
                     continue;    // position t at index of next non-zero mark
                 }
                 // there is a mark, so read the string for a value
-                if (1 == sscanf(&homes[len], "%255s", home) &&
+                if (1 == sscanf(&homes[len], "%254s", home) &&
                     1 == sscanf(home, "%lf", &d)) {
                     // got an entry, index into coordinateMark[] is 't'
                     if (t == 0)
--- a/src/hal/user_comps/xhc-whb04b-6/hal.cc
+++ b/src/hal/user_comps/xhc-whb04b-6/hal.cc
@@ -58,7 +58,7 @@ void Hal::freeSimulatedPin(void** pin)
     if (*pin != nullptr)
     {
         free(*pin);
-        pin = nullptr;
+        *pin = nullptr;
     }
 }
 // ----------------------------------------------------------------------
@@ -573,11 +573,6 @@ void Hal::init(const MetaButtonCodes* metaButtons, const KeyCodes& keyCodes)
     newHalBit(HAL_OUT, &(memory->out.axisCSelect), mHalCompId, "%s.halui.axis.c.select", mComponentPrefix);
 
     mIsInitialized = true;
-}
-// ----------------------------------------------------------------------
-bool Hal::isInitialized()
-{
-    return mIsInitialized;
 }
 // ----------------------------------------------------------------------
 real_t Hal::getAxisXPosition(bool absolute) const
--- a/src/hal/utils/scope_disp.c
+++ b/src/hal/utils/scope_disp.c
@@ -515,7 +515,7 @@ static int handle_click(GtkWidget *widget, GdkEventButton *event, gpointer data)
     } else {
         int z = select_trace(event->x, event->y);
         int new_channel = z & 0xff;
-        int channel_part = z >> 8;
+        int channel_part = (unsigned)z >> 8;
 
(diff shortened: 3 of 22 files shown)


src/hal/user_comps/xhc-whb04b-6/hal.cc
```diff
@@ -58,7 +58,7 @@ void Hal::freeSimulatedPin(void** pin)
     if (*pin != nullptr)
     {
         free(*pin);
-        pin = nullptr;
+        *pin = nullptr;
```
smoe, line 61: Good one.

src/emc/ini/initraj.cc
```diff
@@ -295,12 +295,15 @@ static int loadTraj(EmcIniFile *trajInifile)
                         homePose.b = d;
                     else if (t == 5)
                         homePose.c = d;
+/*
+ * The following have no effect. The loop only counts [0..5].
```
smoe, line 299: So uvw are left uninitialized - are those bits used elsewhere?

src/hal/utils/scope_disp.c
```diff
@@ -515,7 +515,7 @@ static int handle_click(GtkWidget *widget, GdkEventButton *event, gpointer data)
     } else {
         int z = select_trace(event->x, event->y);
         int new_channel = z & 0xff;
-        int channel_part = z >> 8;
+        int channel_part = (unsigned)z >> 8;
```
smoe, line 518: I looked up the function "select_trace" in the same file (https://github.com/LinuxCNC/linuxcnc/blob/1b61c580ee35b6ad7b6ef7cabad859e8dcc71555/src/hal/utils/scope_disp.c#L449) and if I am not erroneous then it always returns the value -1.

smoe:
This is very nice work of yours, thank you tons for this.

smoe (commented):
My attempts to educate myself on this one have failed. My initial hunch was "alignment", but that should have happened right at the initialisiation of the char*? Maybe by introducing https://en.cppreference.com/w/cpp/language/alignas ?

The transition from the char* pByte to the double* could be more formally done by reinterpret_cast, but since the compiler already accepts this, I think this this just fine?

src/hal/user_comps/xhc-whb04b-6/hal.cc
BsAtHome: Struck me too ;-)

src/emc/ini/initraj.cc
BsAtHome: No, they are initialized in the declaration of homePose (line 267). The homePose variable is used later in line 322 in a call to emcTrajSetHome().

Alternative was to extend the loop to count [0..8]. However, that would fail on the coordinateMark[] variable (line 262), which only had 6 entries. Its initialization {1,1,1,0,0,0} means that only XYZ are used anyway.

I have no clue why this code was written as it was, but I am sure that the if-then-else clauses (lines 286..303) never can reach 6, 7 or 8.

src/hal/utils/scope_disp.c
BsAtHome: No, 'target' is a global variable and changed other places.

smoe: You are right, changed by the function line which is called as part of at least the draw_triggerline function. But that only if the other variable DRAWING is set. Just in case -1 is returned, is that a special case that should be caught?

BsAtHome:
> The transition from the char* pByte to the double* could be more formally done by reinterpret_cast, but since the compiler already accepts this, I think this this just fine?

This is C-code, no reinterpret_cast available...

The problem with a char* to double* cast is alignment. It is a type-punned pointer and can fail dereference at runtime if not careful. However, the case here guarantees that the pointer is properly aligned.

The default trick is to cast twice, once over void*, to force the compiler to accept the type-pun.

src/hal/utils/scope_disp.c
BsAtHome: This code is shared between drawing and selecting a line. Function select_trace() clears DRAWING and refresh_display() sets it.

smoe:
> > The transition from the char* pByte to the double* could be more formally done by reinterpret_cast, but since the compiler already accepts this, I think this this just fine?
> 
> This is C-code, no reinterpret_cast available...
:-) And no alignas.
 
> The problem with a char* to double* cast is alignment. It is a type-punned pointer and can fail dereference at runtime if not careful. However, the case here guarantees that the pointer is properly aligned.
> 
> The default trick is to cast twice, once over void*, to force the compiler to accept the type-pun.

You then tricked the compiler, but the cpu could then still not accept the double at that position? To avoid that, should then at
https://github.com/LinuxCNC/linuxcnc/blob/8b3a6d0810d09ea922b61fc91c871c39192bb6b6/src/hal/classicladder/arrays.c#L176
the pByte be initialized as a double* ?

src/hal/utils/scope_disp.c
smoe: Sounds very much like you know what you are doing. 👍

BsAtHome:
> > The default trick is to cast twice, once over void*, to force the compiler to accept the type-pun.

> You then tricked the compiler, but the cpu could then still not accept the double at that position? To avoid that, should then at

[snip]
This code has never failed, so the alignment is probably fine. It is (very) ugly code when you need this extra casting at all, but fixing that is not an option.

On amd64 it does not fail because the pointer is 2<sup>2</sup> aligned. For arm64 it might be a problem when only on a 2<sup>2</sup> boundary, but that has not happened either yet. Therefore, I think we are safe here to silence the compiler warning.

smoe:
> > > The default trick is to cast twice, once over void*, to force the compiler to accept the type-pun.
> 
> > You then tricked the compiler, but the cpu could then still not accept the double at that position? To avoid that, should then at
> 
> [snip] This code has never failed, so the alignment is probably fine. It is (very) ugly code when you need this extra casting at all, but fixing that is not an option.
> 
> On amd64 it does not fail because the pointer is 22 aligned. For arm64 it might be a problem when only on a 22 boundary, but that has not happened either yet. Therefore, I think we are safe here to silence the compiler warning.

I agree. Maybe just leave a comment?

smoe (approved):

PR #3337, opened 2025-02-14
Fix 25 year old typo

kirberich:
I assume the code actually does the right thing, as I think it would be quite obvious if not, but I just noticed this typo in the original spec (https://tsapps.nist.gov/publication/get_pdf.cfm?pub_id=823374) and saw it's been copied to the LinuxCNC docs as well.

G94 is actually executed first and is duplicated here, instead of the much more sensible G92.3, which is missing.

### DIFF
--- a/docs/src/gcode/overview.adoc
+++ b/docs/src/gcode/overview.adoc
@@ -1182,7 +1182,7 @@ position of each item on the line, but by the following list:
 * Set distance mode (G90, G91).
 * Set retract mode (G98, G99).
 * Go to reference location (G28, G30) or change coordinate system
-  data (G10) or set axis offsets (G52, G92, G92.1, G92.2, G94).
+  data (G10) or set axis offsets (G52, G92, G92.1, G92.2, G92.3).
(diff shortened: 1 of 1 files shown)


kirberich:
Judging from how every bit of documentation I can find has this same error, I think the spec is really the only place this is ever mentioned - I've attached a screenshot of the bit of the spec that I'm basing this on. I even checked the original version of the spec, but that doesn't have the order of execution in as much detail

Screenshot 2025-02-15 at 18 21 55

petterreinholdtsen:
I found the same duplication replicated on https://www.cnccookbook.com/g-code-basics-program-format-structure-blocks/ and https://docs.edingcnc.com/order-of-execution, so I guess it is widespread.  Perhaps the issue should be reported as a errata for the specification too?

kirberich:
I did wonder about that, but thought it might be a lost cause for a small thing like this 😅 as it doesn't look like it's been updated since 2000 (as far as I could tell, maybe I'm wrong about that!)

andypugh (approved):

PR #3339, opened 2025-02-16
.gitignore fixes

NTULINUX:
Ignore mod files generated by build and text editor "~" files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ build-stamp
 *.mod.c
 *.cmd
 .tmp*
+src/*.mod
 # Ignore generated html files, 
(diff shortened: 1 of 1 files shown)


smoe (approved):
I am blanking about the .mod files and if anything other than .mod.c (which are removed in the Makefile upon clean) can create them.
The missing "*~" would be much appreciated to be added, indeed.

rmu75:
while at it you might also add `.ccls-cache`

smoe:
And also `*_rc.py` as it happened in https://github.com/LinuxCNC/linuxcnc/pull/3285.

PR #3340, opened 2025-02-16
.gitignore fixes and 5.4.290 debian support

NTULINUX:


NTULINUX:
Disabling docs weren't supposed to be part of this PR, sorry.

PR #3346, opened 2025-02-21
QtVcp: new calc features, new homing option on ActionButton, new OperatorValueLine widget, indicatorMixin bugfix.

tangentaudio:
1. Adds two new calculator dialog features, configured via the [DISPLAY] section of the INI file.  [documented in the .adoc]

* `CALCULATOR_CONST_VALUES` - a comma-delimited list of common values you might enter, that will appear on a dedicated row of buttons at the bottom of the calculator.  e.g. setting to `0.100,-0.100` would provide two buttons for +0.100 and -0.100 which are commonly used when edge-finding on inch mills.  Up to six (6) values may be entered, beyond that the list will be truncated.  Values must be valid floating point or integer.
* `CALCULATOR_ON_SHOW` - optionally set to `CLEAR_ALL` to issue a "Clear All" each time the calculator is shown.  This will clear any previously entered values from the last time the calculator was used and open with the display value set to `0`

2. Bug fix in indicatorMixin for flashing status, it previously was not honoring the flashRate property setting and would only flash at 200ms intervals (the default value).

3. Adds new `home_no_unhome_option` property to ActionButton.  When set, this will make a home action button *only* do homing, it will disable the logic which makes it act as a home/unhome toggle.  In my GUI, I did not want it to be so easy to un-home an axis.

4. Adds new 'Operator Value Line' QtVcp widget that allows manual entry of values, optionally using the popup keyboard.  It can format the value and issue an MDI command automatically or in a manual fashion.  It also tracks pending state so it can be styled when a value is pending.  This can be used to create spindle speed and tool change fields much like Tormach PathPilot has.  [documented in the .adoc]


c-morley:
Excellent work. Thank You!

PR #3347, opened 2025-02-23
Add 5.4.290 support, update docs

NTULINUX:
5.4.290 fixes a lot of issues and should be the new replacement all around.

### DIFF
--- a/debian/configure
+++ b/debian/configure
@@ -238,7 +238,7 @@ case $TARGET in
 	EXTRA_RECOMMENDS="$EXTRA_RECOMMENDS, linux-image-rt-amd64 [linux-amd64], linux-image-rt-686-pae [linux-i386]"
 	CONFIGURE_REALTIME_ARG=--with-realtime=uspace
     ;;
-    4.14.174-rtai-amd64|4.19.195-rtai-amd64|5.4.258-rtai-amd64)
+    4.14.174-rtai-amd64|4.19.195-rtai-amd64|5.4.258-rtai-amd64|5.4.290-rtai-amd64)
         CONFIGURE_REALTIME_ARG=--with-realtime=/usr/realtime-$KERNEL_VERSION
     ;;
     *)


BsAtHome:
Two things, 1) aren't the translations supposed to be handled through weblate and 2) what is it that makes the suggested change fail CI?

andypugh:
I think it would be best to only fix the english docs, though I am not clear how this sort of thing would be expected to propagate through weblate. 

Also, do we actually have a 5.4.290-rtai kernel image? Note that we supply the kernel and linuxcnc packages as a set, so whilst I see some value in _allowing_ 5.4.290 it is quite a lot more work to distribute it.

andypugh:
@BsAtHome CI doesn't run an RTAI kernel. So it shouldn't even _notice_ the changes to RTAI specific things.

BsAtHome:
> @BsAtHome CI doesn't run an RTAI kernel. So it shouldn't even _notice_ the changes to RTAI specific things.

That is why I'm puzzled as to the <i>complete</i> CI failure.

andypugh:
The issue is that my public 2.9-rtai branch, against which this patch is to be applied, doesn't actually build debs without parameters to the debian/configure command. And the CI.yml doesn't have those. 
https://github.com/LinuxCNC/linuxcnc/blob/master/.github/workflows/ci.yml#L155
(Should be debian/configure -r )

andypugh:
I have added the new kernel, but not changed the docs yet as a separate commit. 
https://github.com/LinuxCNC/linuxcnc/commit/488645dc9c2955d47d82803c74a31982a9df5447

hansu:
> Two things, 1) aren't the translations supposed to be handled through weblate

Yes, but only for master.

I suggest to change only the adoc file (which is not included here)
```
#: src/getting-started/getting-linuxcnc.adoc:369
msgid ""
"Reboot the machine, ensuring that the system boots from the new 5.4.258-rtai "
"Reboot the machine, ensuring that the system boots from the new 5.4.290-rtai "
```
and do not touch the po files (this will probably cause merge conflicts and you don't want to solve them all one by one =) ). The pot file change can be done here, but this file is generated (manually), so there is no need to do this change manually.

NTULINUX:
I am actually a bit worried now about the 5.4.290 kernel because when you load a config, not a sim, LinuxCNC crashes and it'll log you out and you'll get a black screen, and it seg faults a lot. I'm going to try the 5.4.290 kernel in Gentoo and see if it's any better. If 5.4.290 gives me trouble on Gentoo, I'll have to downgrade the kernel. 5.4.279 definitely works, so something may have happened between 5.4.279 and 5.4.290.

smoe:
Hi @NTULINUX, do you have an update on how to adjust the documentation? As already pointed out by others, it will be on Weblate contributors to update their respective language, so we can avoid conflicts in the version management.

NTULINUX:
I would put this on hold for now because 5.4.290 is having issues with LinuxCNC+parallel port cards. The simulator and everything else works, but after using stepconfig and actually using the parallel port card, the system crashes. I need to downgrade the kernel but I have not yet had access to a machine with a parallel port card in awhile, and it will be a long time until I do. For the mean time, please use 5.4.279.

andypugh:
I have put 5.4.279 on the servers, and re-issued 2.9.8-rt based on that.

PR #3351, opened 2025-02-25
QtVcp: adds tool chooser widget and dialog, add new logging severity and MachineLog view.

tangentaudio:
Intended to be used along with the Operator Value Widget, this new dialog will allow the operator to select a tool from the tool table.  It can be used to make a friendly manual tool change control where the operator can select the tool from the table rather than having to remember the correct number.

There are also a few minor fixes to Operator Value Widget in this PR as they were developed together.

image

This PR now also includes changes to extend the 'update-machine-log' signal so that the 'option' parameter can optionally contain a severity field.  Previously this field would be e.g. 'TIME' but now is extended to allow a severity to be added by specifying a comma-delimited list, e.g. 'TIME,ERROR'.

I also extended the MachineLog widget to have a new optional way of displaying the machine log with severity, using a QTableWidget.  The widget has configurable colors for the various severities that can be set in Designer.

image


c-morley:
You have merged master in a couple times. I think it would be better to rebase on master so the additions are all in a row. Can you do that? Otherwise I'll try to get to it this weekend.

PR #3356, opened 2025-02-28
QtVcp: new ToolChooser widget and dialog, new logging severity levels and MachineLog widget additions

tangentaudio:
Adds a new Tool Chooser dialog, which will allow the operator to select a tool from the tool table. It can be used to make a friendly manual tool change control where the operator can select the tool from the table rather than having to remember the correct number.

Extends the 'update-machine-log' signal so that the 'option' parameter can now optionally contain a severity field. Previously this field would be simply 'INITIAL', 'DATE', 'TIME', or None.   Now it is extended to allow a severity to be added by specifying a severity level after a comma, e.g. 'TIME,ERROR'.

Extends the MachineLog widget to have a new optional way of displaying the machine log with severit highlighting, using a QTableWidget. The widget has configurable colors for the various severities that can be set in Designer.  The old plaintext log style is still available as well.

@c-morley hopefully this is a more concise PR than the last one with the extra merges in it.


c-morley:
I see you have coded using some of Qtvcp's unique paradigms. Nice job.
I was curious how difficult was it for you to code for Qtvcp and was the docs useful enough?

tangentaudio:
@c-morley the learning curve has not been too bad so far.  I have the advantage of a lot of years of both Qt and LinuxCNC experience.  I've used the source code more than documentation in order to understand how things work, but I certainly have also been looking at the docs.  For the new widgets I created, it was easy enough to find something similar and use it as a template and then rewrite as necessary.  As for trying to understand Qtvcp paradigms, I think it's also been a case of studying existing code more than consulting docs.

I did start out trying to accomplish my goals using a handler file with bits of monkey patching, but it quickly became obvious that wasn't going to get me where I wanted.  At that point I figured I might as well learn the internals and contribute my work since others may be interested.

-Steve

PR #3358, opened 2025-03-01
QtVcp calculator and operator value line improvements

tangentaudio:
Per a request from @snowgoer540 :
https://github.com/LinuxCNC/linuxcnc/commit/f16d729d4778d4b37c703a707f856152166513f6#commitcomment-153113741

The new calculator options that I added in a previous PR have been moved from the INI file into the Preferences.

This PR also adds a couple new options like the ability to force focus and accept-on-return-key to the calculator.  This makes it work much better with a physical keypad.

Also a couple fixes to my OperatorValueLine widget along the same lines to improve behavior with a physical return key.

Changes documented in the .adoc


snowgoer540:
Thank you for making these changes!

PR #3368, opened 2025-03-06
QtVcp more calculator tweaks

tangentaudio:
These tweaks help improve the behavior of the popup calculator, especially when using a physical numpad.  It's not perfect but it might be "good enough" for now until a full rethink and rewrite can be done.


snowgoer540:
No worries at all!  Sorry I jumped the gun, didn’t realize you had some changes in queue.

PR #3370, opened 2025-03-06
Account for differences between cppcheck versions

BsAtHome:
Different versions of cppcheck result in some differences in warnings and errors. Specifically, Debian 12 is at 2.10 and Ubuntu 24.4 is at 2.13 (used in CI), while current version is 2.17. Interestingly, older cppcheck versions found some problems not seen by newer versions.

This PR fixes the messages. Some are actual problems, like printf/scanf formats or arguments. Also missing initialization could occur in some corner-cases. Other messages are more like informational noise or design choices that cannot easily be changed (like virtual call in constructor).

There were initially messages left in Debian 12 because cppcheck 2.10 lacks some suppress and depth features, but these problems are handled for now by reformulating some suppression and ignoring some.

Additionally, the scripts/cppcheck.sh script now ignores header files because some .h files are not necessarily C headers. Headers are checked anyway because cppcheck does includes. The script's return/exit value is non-zero if any cppcheck fails.

### DIFF
--- a/scripts/cppcheck.sh
+++ b/scripts/cppcheck.sh
@@ -34,30 +34,41 @@ cd "$(dirname "$0")/../src" || { echo "Could not change directory to '$(dirname 
 
 # Only process individual files if passed on the command line.
 if [ $# -gt 0 ]; then
+    retval=0
     for f in "$@"; do
         if [ -r "$f" ]; then
             case "$f" in
             *.hh|*.cc)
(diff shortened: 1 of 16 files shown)


andypugh:
Thanks for your continuing efforts on this.

BsAtHome:
The good part... as of now, cppcheck is completely error and warning free. And it has squashed quite a few bugs in the process 8-)

PR #3376, opened 2025-03-14
Halscope: set button padding to allow smaller window width

hansu:
Sets the left and right padding of buttons to 0 to lower the window width to fit smaller screens.

Before:
halscope-1030

After:
halscope-860px

Not sure if this should better go into master instead, but it's a very small change.


hansu:
:+1:

PR #3377, opened 2025-03-16
debian: Changes associated with upload of 2.9.4

smoe:
Hello,

These are a couple of changes I have done to the linuxcnc/debian folder for the upload to the distribution. The most important bit may be that I merged the d/changelog of what is in Debian  with the one we have as upstream.  I am not ultimately confident that this is the right thing to do, but I liked it.

I added "debian/source/format" and set that to quilt. This possibly interferes with the  autobuilders in case that these are performing a native build, i.e. one without an orig.tar.gz. The orig.tar.gz was the official release of 2.9.4.

Added Andy to list of format Uploaders - did not ask :-) . Also:
 - Merged d/changelog with what was found in d/changelog of 2.9.3 in Debian.
 - Lintian requested d/source/format - set to quilt
 - Removal of files no longer problematic/removed in d/source/lintian-overrides - also requested by lintian.


andypugh:
I think that this change to quilt _might_ be causing a problem with the buildbot? 
http://buildbot2.highlab.com/buildbot/#/builders/8/builds/2590

Buildbot2 hasn't been even trying to build master debs since February (I hadn't noticed) and this seems to be because scripts/platform-is-supported was saying that Bullseye was supported, that build as failing, and so deb creation was skipped. 

I have modified platform-is-supported and now it is attempting to build Bookworm debs, but failing, as reported in the link above.

PR #3378, opened 2025-03-17
stop-lookahead sim config: remove section with broken web links from the README

Sigma1912:


andypugh:
I had a good look for that forum thread, it should still exist, but I couldn't find it.

PR #3384, opened 2025-03-24
debian/patches: remove the explicit use of FORTIFY_SOURCE

pushkarnk:
Starting Ubuntu 24.04, dpkg-buildflags passes _FORTIFY_SOURCE=3 by default.
On Debian, dpkg-buildflags passes _FORTIFY_SOURCE=2. The explicit use of _FORTIFY_SOURCE=2
in the Makefile causes redefinition errors on Ubuntu, failing the build. This patch enforces
depending on `dpkg-buildflags` for the _FORTIFY_SOURCE value.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,8 +1,14 @@
 linuxcnc (1:2.10.0~pre0) UNRELEASED; urgency=medium
 
+  [ Andy Pugh ]
   * Master branch open for new features.
 
- -- Andy Pugh <andy@bodgesoc.org>  Mon, 09 Oct 2023 10:39:03 +0100
+  [ Pushkar Kulkarni ]
+  * debian/patches: Remove the use of FORTIFY_SOURCE and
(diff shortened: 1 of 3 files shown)


BsAtHome:
So, this is set automatically on package builds, but it differs between Debian and Ubuntu.

 My understanding is that package builder call configure with CFLAGS and CXXFLAGS set to force the _FORTIFY_SOURCE in the build. So, then, why create a patch-patch during package build and not remove it from out source-tree? It may be an idea to add some --enable-xxx thing to configure so that local builds can enable it at will?

Also, the debian/changelog file is auto-generated iirc.

andypugh:
debian-changelog is semi-manually created at the tie that a new release is made. It is based on the commit messages, and shouldn't normally be included in patches. 
However, I see no harm in there being a reminder to me what this one was when I edit the next changelog.

PR #3388, opened 2025-03-26
debian: usrmerge - Debian bugs #1073692,#1101346

smoe:
This PR complements a thread on our mailing list about the efforts of Debian and other distros to merge /lib with /usr/lib and the recent bug report 1101346: My upload to Debian reintroduced a bug that Petter had already closed for 2.9.3, see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1073692 and https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1101346 .

No idea if I had seen that if I had followed the linuxcnc-gbp approach for uploads to Debian.

Apparently the https://wiki.debian.org/UsrMerge is already a thing since Buster, so LinuxCNC can just broadly adopt it? Or should that be an option to configure?

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,28 @@
+linuxcnc (1:2.9.4-2) unstable; urgency=medium
+
+  * Team upload.
+
+  * Adding build dependency on pyqt5.qtweb{engine,kit} to avoid build
+    failures on, as kindly addressed by Adrian <Bunk>. (Closes: #1100882)
+  * Re-applied Petter's change to address usrmerge, kindly pointed out
+    by Michael <Biebl>. (Closes: #1101346, #1073692)
+  * Bumped policy to 4.7.2.
+  * Now truly added Andy to Uploaders.
+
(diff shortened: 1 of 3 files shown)


andypugh:
Does this belong in master or 2.9?

smoe:
Good point. You are right. This should go to the 2.9. branch.

andypugh:
Well, changing the base here in github goes disasterously wrong, so I won't do that. 
I can export the PR as a patch and manually apply it to 2.9 ( by appending .patch, ie https://github.com/LinuxCNC/linuxcnc/pull/3388.patch ) but that's probably more work than you rebasing your PR. (and it's definitely more work for _me_ :-)

smoe:
I'll perform the rebase. :-)

smoe:
I have uploaded a version -2 to the Debian archive a few minutes ago. That may now be too late for an auto-transition to the new Debian release but if the problems disappear that are currently shown on https://tracker.debian.org/pkg/linuxcnc then I am confident that we will ship with the trixie release.

PR #3390, opened 2025-03-27
Add bitmerge as inverse of bitslice

BsAtHome:
The inverse function of bit slicing is bit merging. Add the component bitmerge to do so.
Also, prevent bitslice to read the source pin multiple times because volatiles are not cached and re-loaded on every dereference.
This is a slightly better implementation than originally submitted in #3386. The current version is not using a floating bit-mask. However, the compiler optimizer stages are so good nowadays that, either way, code will be fast when multiple volatile loads are prevented.

### DIFF
--- a/docs/man/.gitignore
+++ b/docs/man/.gitignore
@@ -261,6 +261,7 @@ man9/axis.9
 man9/axistest.9
 man9/bin2gray.9
 man9/biquad.9
+man9/bitmerge.9
 man9/bitslice.9
 man9/bitwise.9
 man9/bldc.9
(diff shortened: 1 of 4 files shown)


BsAtHome:
> I wonder if you can already use 64 bit data types here?

That depends on what you connect the pin to. If you do, then you can only connect it to u64 pins or need to do a conversion. If you don't, you can only connect to u32 pins or need a conversion.

However, the status quo is u32 and moving it to u64 means an incompatible change wrt. all existing use if you change it.

hansu:
Yes you are right, we should change all at the same time.

andypugh:
> Yes you are right, we should change all at the same time.
This is a work-in-progress: https://github.com/LinuxCNC/linuxcnc/compare/master...andypugh/long_hal_ints

By default the "weighted sum" component does exactly the same as bitmerge. Are we sure we need both?

BsAtHome:
Although mathematically both (can) perform the same job, the weighted sum component is a very inefficient way to perform bit-merging.

Also, the weighted sum component uses signed output, whereas the bitslice component uses unsigned input. Therefore, the results from slicing a value and putting it back together with weighted sum are not identical.

PR #3381, opened 2025-03-22
debian: Reducing lintian warnings.

smoe:
* file permissions
* man page conversion fro adoc to troff
* hardening linker flags for panelui

Lintian errors of the latest upload of LinuxCNC are on
https://udd.debian.org/lintian/?packages=linuxcnc

PR #3398, opened 2025-04-14
debian: lintian indicates hardening-no-relro: panelui, module_helper

smoe:
I wish there was a comment for panelui if the LDFLAGS were intentionally omitted.

We have no policy about the use of "@" (don't show the command) in Makefiles. I removed the one for the compilation of linuxcnc_module_helper since arguments to the compiler ar eimportant to know.  Was tempted to introduce it for all the $(ECHO) instructions, though, since we have little incentive to read the same lines twice.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -44,14 +44,17 @@ linuxcnc (1:2.9.4) UNRELEASED; urgency=medium
   * Update m-code.adoc - typo exection -> execution
   * Merge pull request #3247 from Sigma1912/patch-1
   * gmoccapy.adoc: correct INI section name [MACROS]
-  * Disable "override limits" at the end of the jog. This fixes the issue that the override was not cleared in teleop (homed) mode Issue #2482
+  * Disable "override limits" at the end of the jog.
+    This fixes the issue that the override was not cleared
+    in teleop (homed) mode Issue #2482
   * Doc: User Defined Command, add note on exit codes != 0
   * gmoccapy: update release notes
   * docs: fix list in building-linuxcnc.adoc
   * Typos in carousel.comp
   * Merge pull request #3227 from BsAtHome/backport_2.9_3200-3202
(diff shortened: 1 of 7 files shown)


smoe:
The warnings were observed on
https://udd.debian.org/lintian/?packages=linuxcnc

smoe:
Hm. Seems like I was addressing panelui before at https://github.com/LinuxCNC/linuxcnc/pull/3381,
albeit slightly differently.

I cannot confirm myself if panelui and linuxcnc_module_helper are indeed functional with the the -z relro added. If some good soul that is using those binaries could please confirm that for me and then this should go into the 2.9 branch, not against master. I would then also go and rebase what I did in https://github.com/LinuxCNC/linuxcnc/pull/3381 accordingly.

andypugh:
I think that panelui is something @c-morley wrote? 
There seems to be a sample config to test it with:
sim/axis/panelui-demo

smoe:
Just started the panelui demo and experienced no crashes. Seems like fine.
Lintian complained about lines too long in d/changelog and typos that I addressed (all typos, not all long lines).
Edited lintian overrides to eliminate some more warnings..

smoe:
The build was cancelled. Did not break or anything. I think this can go in.

BsAtHome:
> The build was cancelled. Did not break or anything. I think this can go in.

See #3408.

smoe:
Have seen that #3388 was merged, will rebase to the latest 2.9 later tonight, which should then also fix the build.

andypugh:
Is this still live?

smoe:
Thank you for your reminder.

smoe:
@andypugh , I just skimmed through the changes again and think this is fine. Should also be good to find its way to 2.10.

smoe:
That said, should I squash those patches?

andypugh:
> That said, should I squash those patches?

I can squash on merge.

PR #3399, opened 2025-04-18
Fix HAL_TYPE_MAX to be the largest numerical value

BsAtHome:
The numerical constant of the largest legal value in the hal type enum <code>HAL_TYPE_MAX</code> was, although properly placed last, not the correct value because the other enumerations were not in order while using fixed value assignments. Unassigned enums always take the "last value plus one" approach and that resulted in the wrong enum value for <code>HAL_TYPE_MAX</code>.

The above fix exposed enum switch-cases that were not handled properly. These are fixed to use a default clause now.

### DIFF
--- a/src/hal/hal.h
+++ b/src/hal/hal.h
@@ -279,9 +279,9 @@ typedef enum {
     HAL_FLOAT = 2,
     HAL_S32 = 3,
     HAL_U32 = 4,
+    HAL_PORT = 5,
     HAL_S64 = 6,
     HAL_U64 = 7,
-    HAL_PORT = 5,
     HAL_TYPE_MAX,
 } hal_type_t;
 
--- a/src/hal/halmodule.cc
+++ b/src/hal/halmodule.cc
@@ -376,8 +376,8 @@ static PyObject *pyhal_read_common(halitem *item) {
             case HAL_S64: return to_python(*(item->u->pin.s64));
             case HAL_FLOAT: return to_python(*(item->u->pin.f));
             case HAL_PORT: // HAL_PORT is currently not supported
-            case HAL_TYPE_UNSPECIFIED: /* fallthrough */ ;
-            case HAL_TYPE_UNINITIALIZED: /* fallthrough */ ;
+            default:
+                break;
         }
     } else {
         switch(item->type) {
@@ -388,8 +388,8 @@ static PyObject *pyhal_read_common(halitem *item) {
             case HAL_S64: return to_python(item->u->param.s64);
             case HAL_FLOAT: return to_python(item->u->param.f);
             case HAL_PORT: // HAL_PORT is currently not supported
-            case HAL_TYPE_UNSPECIFIED: /* fallthrough */ ;
-            case HAL_TYPE_UNINITIALIZED: /* fallthrough */ ;
+            default:
+                break;
         }
     }
     PyErr_Format(pyhal_error_type, "Invalid item type %d", item->type);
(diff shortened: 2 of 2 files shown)


andypugh:
Is this mainly an aesthetic problem, or does it cause issues that should be addressed in 2.9 too?

smoe:
I admittedly did not want to believe this at first sight, the compiler should raise a warning I thought, but, yes, indeed. And cppcheck does not consider this noteworthy either.

This should go to 2.9., I rest assured.

@andypugh , maybe you just do the change in your local 2.9 checkout? 

What I did not like too much about the patch was that the technically admittely redundant lines

```
            case HAL_TYPE_UNSPECIFIED: /* fallthrough */ ;
            case HAL_TYPE_UNINITIALIZED: /* fallthrough */ ;
```
were removed. I wish there would have been a comment left, but that is a matter of taste.

andypugh:
I am (largely) away from my LinuxCNC machines for Easter. A PR for 2.9 would be a fair bit easier for me. 
(But you both have push rights? "Make it so" :-)

BsAtHome:
> What I did not like too much about the patch was that the technically admittely redundant lines
> ```
>             case HAL_TYPE_UNSPECIFIED: /* fallthrough */ ;
>             case HAL_TYPE_UNINITIALIZED: /* fallthrough */ ;
> ```
> were removed. I wish there would have been a comment left, but that is a matter of taste.

Well, yes, they are redundant and more so when using a default case. However, they are also just noise because the meaning of these specific enums cannot be implemented in the context of the code. As such, it will become a case of strategy and taste how to handle enums in switches, whether to list only the used ones and add a default case or list them all at all times.

If there are others who dislike my choice, please let your voice be heard or accept fate ;-)

@andypugh I'll make a 2.9 PR too.

smoe:
You just convinced me (and likely others). Your patch is fine. Thank you.

andypugh:
I tried exporting this as a patch, and hit a bit of a puzzle. 
The patch applies apart from two sections of halmodule.cc. The patch deletes the 
-            case HAL_TYPE_UNSPECIFIED: /* fallthrough */ ;
-            case HAL_TYPE_UNINITIALIZED: /* fallthrough */ ;
And replaces with a default; break. Which seems fine. 

But in 2.9 there is already a default. 

Compare:
master:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/halmodule.cc#L1246
2.9:
https://github.com/LinuxCNC/linuxcnc/blob/2.9/src/hal/halmodule.cc#L1238

I asked GitHub copilot who deleted the default case and it said that you (Bertho) did, but the commit it gave me didn't do that, so it was hallucinating in true AI style.

BsAtHome:
> I tried exporting this as a patch, and hit a bit of a puzzle. The patch applies apart from two sections of halmodule.cc.
>[snip]
> And replaces with a default; break. Which seems fine.
> But in 2.9 there is already a default.

I'll have a look. A bit busy finishing documenting a new modbus driver...


> I asked GitHub copilot who deleted the default case and it said that you (Bertho) did, but the commit it gave me didn't do that, so it was hallucinating in true AI style.

Well, I'm known to exist, not, and be a ghost, do don't exist not surely maybe and am completely halfway existential. It is also well documented that I have an advanced degree in producing bullshit of any kind. Are you sure you want to ask me?

"AI" is the (Dutch) abbreviation for "Afnemende intelligentie", which translates to "Decreasing Intelligence". That will be approximately appropriate for anyone who tries to take a very very large Markov Chain machine generated reply seriously.
;-)

BsAtHome:
> I tried exporting this as a patch, and hit a bit of a puzzle.
> [snip]

The point is, this particular patch does not need to be backported.

The 2.9 branch does not include the same enums as the master branch. Looking at 2.9 branch's hal/hal.h:275, then there is no HAL_TYP_MAX and all enums are in numerical order.

The master branch got updated by https://github.com/LinuxCNC/linuxcnc/commit/7a4c0908567e27f034c7cb2fe2f6c5b1341c933c a long time ago and the problem was introduced there.

andypugh:
In that case, it seems we are all good.

PR #3401, opened 2025-04-23
Fix most remaining warnings on ARM

BsAtHome:
This PR fixes another set of warnings that is exclusive to the ARM compile, mostly due to unavailable IO-mapped IO instructions. However, it also deals with an invalid pointer/integer comparison and a type range warning.

### DIFF
--- a/src/hal/classicladder/protocol_modbus_master.c
+++ b/src/hal/classicladder/protocol_modbus_master.c
@@ -256,9 +256,9 @@ int PrepPureModbusAskForCurrentReq( unsigned char * AskFrame )
 
 /* Response given here start directly with function code 
   (no IP header or Slave number) */
-char TreatPureModbusResponse( unsigned char * RespFrame, int SizeFrame )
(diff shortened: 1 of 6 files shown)


BsAtHome:
Then I'd suggest that we do away with the #define statements and replace them with inline functions. That would probably be a better and nicer way to deal with it. Expanding to literal versions of some <code>(void)x</code> may be fragile.

andypugh:
I think that my objection to the (void)variable approach is that it seems to suggest "we know that these variables are never used", whereas they actually are used, in the case of x86. It just feels potentially misleading (right up to the point that someone reads the next line, though)

BsAtHome:
I see your point. That is why I suggest to change the underlying problem from <i>"blindly" expanding a preprocessor macro</i> into <i>using an inline function to cover differing expansions</i>.

It would centralize the problem to one place, the inline functions, and you are done with it. Say the word...

andypugh:
> Say the word...

"Make it so, Mr Stulthiens"

PR #3402, opened 2025-04-23
Halshow: Add bit toggle button in watch for HAL_BIT pins and signals

BsAtHome:
You can "Set" and "Clr" bit pins and signals in the Watch tab. However, it was determined, after using halshow for a while, that a toggle option would be a helpful addition.
This PR adds the extra button "Tgl" and does what it says.

### DIFF
--- a/tcl/bin/halshow.tcl
+++ b/tcl/bin/halshow.tcl
@@ -905,17 +905,23 @@ proc watchHAL {which} {
             -fill lightgray -tag oval$i
         if {$writable == 1} {
             if {$vartype == "sig"} {
+                canvasbutton::canvasbutton $::cisp [expr $canvaswidth - 76] \
+                    [expr {$i * 20 + 4}] 24 17 "Tgl" [list hal_tgls $label] 1
                 canvasbutton::canvasbutton $::cisp [expr $canvaswidth - 48] \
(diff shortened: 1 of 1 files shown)


hansu:
I haven't implemented this because it needs extra space and it can also be done with the set and clear buttons.
However, please also update the docs.

BsAtHome:
> I haven't implemented this because it needs extra space and it can also be done with the set and clear buttons.

Actually, the toggle function cannot be done with the set/clr buttons. You have to move your mouse to perform a <i>similar</i> function. From a user interaction standpoint, you want to reduce complexity. Having to coordinate both click <i>and</i> move is substantially more complex (hence more difficult) than just clicking because you need to look, whereas a repeated toggle function can be performed blindly.

With respect to the extra space; you want halshow as a debug and inspection tool. The space requirements are subordinate because the window and its content scales well.

> However, please also update the docs.

Sure can do. But exactly which document needs updating? I thought the button was self-explanatory.

andypugh:
Yes, I see your point, sometimes you want to toggle a signal while looking elsewhere, possibly without even being able to see the screen (like, in the back of the cabinet with the mouse in your hand...)

hansu:
> > However, please also update the docs.
> 
> Sure can do. But exactly which document needs updating? I thought the button was self-explanatory.

Just the screenshots in: https://linuxcnc.org/docs/html/hal/halshow.html

PR #3404, opened 2025-04-24
Fix warnings on RPi/ARM

BsAtHome:
This PR fixes another set of warnings that is exclusive to the ARM compile, mostly due to unavailable IO-mapped IO instructions. However, it also deals with an invalid pointer/integer comparison and a type range warning.

This PR replaces #3401.

### DIFF
--- a/src/hal/classicladder/protocol_modbus_master.c
+++ b/src/hal/classicladder/protocol_modbus_master.c
@@ -256,9 +256,9 @@ int PrepPureModbusAskForCurrentReq( unsigned char * AskFrame )
 
 /* Response given here start directly with function code 
   (no IP header or Slave number) */
-char TreatPureModbusResponse( unsigned char * RespFrame, int SizeFrame )
+int TreatPureModbusResponse( unsigned char * RespFrame, int SizeFrame )
 {
-	char cError = -1;
+	int cError = -1;
 
 	if ( RespFrame[ 0 ]&MODBUS_FC_EXCEPTION_BIT )
 	{
--- a/src/hal/drivers/hal_gpio.c
+++ b/src/hal/drivers/hal_gpio.c
@@ -165,7 +165,7 @@ int build_chips_collection(char *name, hal_gpio_bulk_t **ptr, int *count){
     struct gpiod_line *temp_line;
     
     temp_line = gpiod_line_find(name);
-    if (temp_line <= 0) {
+    if (!temp_line) {
 	    rtapi_print_msg(RTAPI_MSG_ERR, "The GPIO line %s can not be found\n", name);
 	    return -EINVAL;
     }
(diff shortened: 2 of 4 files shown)


src/hal/classicladder/protocol_modbus_master.c
```diff
@@ -256,9 +256,9 @@ int PrepPureModbusAskForCurrentReq( unsigned char * AskFrame )
 
 /* Response given here start directly with function code 
   (no IP header or Slave number) */
-char TreatPureModbusResponse( unsigned char * RespFrame, int SizeFrame )
+int TreatPureModbusResponse( unsigned char * RespFrame, int SizeFrame )
```
smoe, line 259: Would it be preferable to  just
- char
+ signed char
so something with the same bit length is better compatible with the current API?

BsAtHome: This function is used only locally in that same source file (it should've been declared static).

The return value can only ever be zero (0) or minus one (-1). Why they used 'char' as a return type at all boggles the mind.

smoe: Ah, great! And if you declare it static now? Well, let's hear if Andy is any opinionated on that.

Well,  the "char" is kind of classic, I think. But yes, 
https://www.reddit.com/r/C_Programming/comments/1df85hl/why_return_char_instead_of_int_for_functions_that/
it should preferably be int, i.e. the size of the processor's registers.

src/hal/drivers/hal_gpio.c
```diff
@@ -165,7 +165,7 @@ int build_chips_collection(char *name, hal_gpio_bulk_t **ptr, int *count){
     struct gpiod_line *temp_line;
     
     temp_line = gpiod_line_find(name);
-    if (temp_line <= 0) {
+    if (!temp_line) {
```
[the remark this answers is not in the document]
smoe: It is also in 2.9 and should be fixed (https://github.com/LinuxCNC/linuxcnc/blob/df7a294781a6eaf74ab01dc46824fb4d80dc7407/src/hal/drivers/hal_gpio.c#L168). 

Just below that one, there is a retrieval of a chip
https://github.com/LinuxCNC/linuxcnc/blob/df7a294781a6eaf74ab01dc46824fb4d80dc7407/src/hal/drivers/hal_gpio.c#L172
and that return value is used without further checking on the retrieval being successful. 
And then another line is retrieve but not checked again
https://github.com/LinuxCNC/linuxcnc/blob/df7a294781a6eaf74ab01dc46824fb4d80dc7407/src/hal/drivers/hal_gpio.c#L193 .

But maybe that is for 2.10 :-)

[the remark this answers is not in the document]
BsAtHome: There are many missing checks in the code all over the place. It is a hell to look at...
It will need a complete code review and often rework/refactor to fix most or all of them.
At least I backported the compare (#3410).

PR #3414, opened 2025-04-27
Fix modcompile issues

BsAtHome:
The modcompile program is updated to take command-line arguments (see also #3380) and now properly removes temporary files, unless instructed to keep them.
The template is updated to fix a missing comma in the exception list.


andypugh:
Thanks. 
Sorry, but I am not really getting much free time at the moment.

PR #3422, opened 2025-04-29
Backport: Fix invalid escape sequence warnings

BsAtHome:
This is a backport of #3233.

It fixes the SyntaxError messages by fixing escape-sequences.

### DIFF
--- a/src/hal/components/bldc.comp
+++ b/src/hal/components/bldc.comp
@@ -131,7 +131,7 @@ Note that a number of incorrect commutations will have non-zero net torque which
 
 If your motor lacks documentation it might be worth trying every pattern.
 
-.ie '\*[.T]'html' \\{\\
+.ie '\\*[.T]'html' \\{\\
 .HTML \\
 <STYLE> \\
 #pattern TD { text-align: center; padding-left: .5ex; padding-right: .5ex } \\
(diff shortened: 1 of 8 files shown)


andypugh:
I don't see many significant examples of this in the changed files, but there was a reason for the escaped minus signs in the .comp files. 
The issue was that HAL code examples in the docs were getting em-dash characters instead of minus signs as hyphens in HAL pin names, which meant that you couldn't copy and paste from the docs into your HAL file without getting (possibly invisible) errors. 

See https://github.com/LinuxCNC/linuxcnc/commit/4a897203bcdfc1fb57438e98ea56a7417c9c88ad

BsAtHome:
The problem is that I get a lot of SyntaxError invalid escape sequence when building.

The file is preprocessed with python and python barks at the wrong escaped (like \\-). The comment https://github.com/LinuxCNC/linuxcnc/pull/3233#issuecomment-2561892297 suggested that these particular hyphens (minus signs) were alright and were actually changed from '\\\\-' to '-'.

But for these specific man pages, what should it matter if it was an en-dash? There is only one instance where it marginally references hal pins (sim_parport.comp). All other instances are numeric values that are indicative.

andypugh:
As I mentioned, none of these cases seem to be ones where the precise format of the dash matters, I was just trying to add some context as I recall this was a complicated issue to get to the bottom of, and to find any sort of answer to.

andypugh:
Giving the user a warning as to _why_ the HAL pin apparently doesn't exist might be useful. Though maybe it would even make sense to just accept all dash variants as valid when searching the pin names (the search is typically only done once, at load time)

BsAtHome:
I do not really understand the discussion. This PR fixes wrong escape sequences that generate python SyntaxWarning messages, which soon become SyntaxError, just like done in the master branch a long time ago. There is no change in the sources other than that the complaining stops when *building* LCNC. The PR doesn't involve HAL searches nor does it involve changed code paths.

If, for some reason, you want the em-dash/en-dash/hyphen issues solved, then may I suggest you open an issue with examples of where and when things go wrong? I have not been able to reproduce any of these problems and this PR does not address that issue.

The question is: are there objections to merging this PR?

PR #3425, opened 2025-04-29
Backport: fix deprecated warnings

BsAtHome:
This is a backport of #2487 and #3265.


BsAtHome:
This will probably need a test for python version to work properly on debian 10.

BsAtHome:
There now is a python version check in place so that all supported Debian versions 10, 11, 12 and sid are covered and that the deprecation warning no longer occurs for newer python.

smoe:
Looks good to me!

PR #3427, opened 2025-04-30
Add new Hostmot2 Modbus driver hm2_modbus

BsAtHome:
This PR adds a new driver for Modbus communication via Mesa cards' PktUART interface(s).

The new driver can handle up to eight RS-485 busses and has no limit on how many Modbus slave devices share the same RS-485 bus using distinct device slave IDs and includes broadcast support. The driver can map any Modbus coil, input or register to the proper HAL types for reading and writing. The register mappings support 16-bit, 32-bit and 64-bit types in signed and unsigned quantities as well as floating point values of float16, float and double. Data is clamped to the type sizes by default (can be disabled). Pins support dynamic offset/scaling HAL pins on input and output when configured to make them available.

A Modbus Command Control file configures the pins, mappings and communication using one hm2_modbus driver. An XML mbccs source is compiled using mesambccc into a binary mappings file (mbccb file). The mbccb is read by the driver when the module is loaded with loadrt and configures itself. Very complex configurations can be created. Device initialization commands are supported to be sent once at startup to configure Modbus slave devices on the bus.

A couple of examples can be found in configs/by_interface/mesa/hm2-modbus/. Manual pages are supplied for hm2_modbus(9) and mesambccc(1).

Note: you probably need to upgrade your Mesa card and flash it with a PktUART bit-file with V3 support. Older V2 is accepted with a warning. However, it may not support your configuration due to missing features or subtle bugs.


andypugh:
It might have been a good idea to discuss this before doing all the work? It has rather taken me by surprise. 
The reason that mesa_modbus is written the way that it is is because I was not sure how to pass complex configuration data into a realtime component. My understanding is that RTAI realtime components run in kernel space, and so have no access to the filesystem (with the exception of the lib/firmware tree). Does this mean that your new driver is oly usable with LinuxCNC-uspace? (Or am I misunderstanding the limitations of kernel modules?)

BsAtHome:
Currently, it is only compiled in uspace. However, the file-read part has been written to use the kernel infrastructure when it is compiled as a kernel module. Kernel modules _can_ read arbitrary files in the filesystem. The read_mbccb() function is the only thing that touches the file system. And even if that didn't work, you'd probably be able to make an interface via procfs (or sysfs).

See: https://github.com/BsAtHome/linuxcnc/blob/8753f28a6ed5526bdc0c55aa563764e166b7b945/src/hal/drivers/mesa-hostmot2/hm2_modbus.c#L2080 and the read_mbccb() function.

But, I have no RTAI installation at hand and cannot test it. Therefore, only uspace has been enabled for now. And, FWIW, the hm2_eth driver also requires uspace and I only have been able to use a 7i96s to test so far (and the RPi spi interface also requires uspace, but I have errors on my spi card; and, no, not because of this driver). For a kernel module test, I'd need an appropriate PCIe card and a PC to test and I have neither.

andypugh:
It doesn't build as-is under RTAI. 

```/home/andypugh/linuxcnc-dev/src/hal/drivers/mesa-hostmot2/sserial.c: In function ‘hm2_sserial_update_params’:
/home/andypugh/linuxcnc-dev/src/hal/drivers/mesa-hostmot2/sserial.c:1553:33: error: implicit declaration of function ‘labs’; did you mean ‘fabs’? [-Werror=implicit-function-declaration]
 1553 |                             if (labs((p->s64_param - p->s64_written) >> shift) > 2) break;
      |                                 ^~~~
      |                                 fabs
cc1: some warnings being treated as errors
make[2]: *** [scripts/Makefile.build:262: /home/andypugh/linuxcnc-dev/src/hal/drivers/mesa-hostmot2/sserial.o] Error 1
make[1]: Leaving directory '/usr/src/linux-headers-5.4.258-rtai-amd64'
make[1]: *** [Makefile:1754: /home/andypugh/linuxcnc-dev/src] Error 2
make: *** [Makefile:547: modules] Error 2
make: Leaving directory '/home/andypugh/linuxcnc-dev/src'
andypugh@rm-one:~/linuxcnc-dev/src$ 
```
Looking in https://github.com/LinuxCNC/linuxcnc/blob/master/src/rtapi/rtapi_math.h the kernel mode rtapi_math doesn't have any integer abs, which rather surprises me. I don't know why that would be.

andypugh:
There is an abs macro in linux/kernel.h that we could almost certainly use, but it's a bit late here for me to start figuring it out:
https://github.com/torvalds/linux/blob/master/include/linux/math.h#L150

BsAtHome:
> It doesn't build as-is under RTAI.
> <code>
> .../<b>sserial.c:1553</b>:33: error: implicit declaration of function ‘labs’; did you mean ‘fabs’? [-Werror=implicit-function-declaration]
> </code>

This is not a problem with *this* PR. It was caused by #3334 three months ago where the line in question was changed. This PR doesn't touch the sserial.c file.

You can try the following patch which redirects the labs() call to become abs() for kernel compiles and include linux/math.h so it may find the type-adapting abs macro.

```diff
diff --git a/src/hal/drivers/mesa-hostmot2/sserial.c b/src/hal/drivers/mesa-hostmot2/sserial.c
index 7cb724316b..d9c1e0763d 100644
--- a/src/hal/drivers/mesa-hostmot2/sserial.c
+++ b/src/hal/drivers/mesa-hostmot2/sserial.c
@@ -29,6 +29,13 @@
 #include "hostmot2.h"
 #include "bitfile.h"
 
+#if defined(__KERNEL__)
+#include <linux/math.h>
+// The kernel has a special type-adapting macro for abs(). We use labs() in
+// here, which isn't available in the kernel.
+#define labs abs
+#endif
+
 
 int getbits(hm2_sserial_remote_t *chan, rtapi_u64 *val, int start, int len){
     //load the bits from the registers in to bit 0+ of *val
```

BsAtHome:
Maybe this patch is better because it does not require any kernel headers or declarations:
```diff
diff --git a/src/hal/drivers/mesa-hostmot2/sserial.c b/src/hal/drivers/mesa-hostmot2/sserial.c
index 7cb724316b..01e3ab8137 100644
--- a/src/hal/drivers/mesa-hostmot2/sserial.c
+++ b/src/hal/drivers/mesa-hostmot2/sserial.c
@@ -29,6 +29,12 @@
 #include "hostmot2.h"
 #include "bitfile.h"
 
+#if defined(__KERNEL__)
+// The kernel has no labs(). There is a type-adapting macro in the kernel
+// headers, but this is just a one-off use.
+static inline rtapi_s64 labs(rtapi_s64 v) { return v < 0 ? -v : v; }
+#endif
+
 
 int getbits(hm2_sserial_remote_t *chan, rtapi_u64 *val, int start, int len){
     //load the bits from the registers in to bit 0+ of *val
```

andypugh:
> Maybe this patch is better because it does not require any kernel headers or declarations:

I am wondering if it would be better to define labs in rtapi_math.h ? Actually, maybe rtapi_math,h should #include linux/math,h?
In any case, this feels like it should be raised as a separate [issue](https://github.com/LinuxCNC/linuxcnc/issues/3435), and I will bodge things to work for the purposes of testing hm2_modbus.

andypugh:
Sorry to be testing this so slowly. 

There seems to be a disparity in the naming convention for the config files. 
the docs say 
"*mbccbs* [default: <empty>]::
   A comma separated list of "Modbus Command Control Binary" (MBCCB) file paths
   to use for each PktUART port as specified in the *ports* parameter. The path
   should be an absolute path to prevent nasty surprises.
   Example: `mbccbs="/path/to/rly-and-spindle.mbccb","/path/to/lightsparks.mbccb"`"
   
   But the sample files are brushless-dc-motor-driver.mbccs  rs485io.mbccs (etc) 
   
   And then the error I get on trying to load is again using ".mbccb"
   
```
[696332.921962] hm2_5i25.0: initialized AnyIO board at 0000:03:00.0
[696487.527391] hm2_modbus.0: error: Invalid signature in mbccb file: '
[696487.527392] <
[696487.527394] ?
[696487.527394] x
[696487.527395] m
[696487.527395] l
[696487.527396]  
[696487.527396] v
[696487.527397] e
[696487.527397] ', expected 'MesaMB01'
[696496.957947] hm2_5i25.0: dropping AnyIO board at 0000:03:00.0```

Oh, and, there is this error.....

**Comment by BsAtHome** [comment]
> There seems to be a disparity in the naming convention for the config files.

Well, no. The _source_ file is an mbccs file (<b>M</b>od<b>B</b>us <b>C</b>ommand <b>C</b>ontrol _<b>S</b>ource_). The accepted configuration file for the driver is a binary _mbccb_ file (<b>M</b>od<b>B</b>us <b>C</b>ommand <b>C</b>ontrol _<b>B</b>inary_) .

The source is converted to a binary with the _mesambccc_ program:
```bash
$ mesambccc -o outputname.mbccb inputfile.mbccs

# Use the -v flag for verbose output and pin mappings:
$ mesambccc -v -o outputname.mbccb inputfile.mbccs
```

The compiler step is implemented because you cannot do this kind of complicated conversion in-kernel. You need a complex text parser (XML parser in this case) and there is a ton of required error and consistency checking. The driver also does verification and validation of the binary file, but that is much easier and more compact.
The compiler is 1700 lines of code of which nearly all is validation as well as consistency- and error checking. The actual conversion to binary is just 250 lines or so.

The module arguments ***ports*** and ***mbccbs*** are both plural forms of _port_ and _mbccb_ because you can enter up to 8 in a comma separated list.

Suggestions to improve the docs to make it more clear that a configuration compile step is necessary are always welcome.

> But the sample files are brushless-dc-motor-driver.mbccs rs485io.mbccs (etc)

These are the _source_ files. I'd rather not include binary files in the source-tree. However, your confusion does make a point that these probably should be in the Makefile to convert them to binary mbccb files in the build process.

> And then the error I get on trying to load is again using ".mbccb"
> [696487.527391] hm2_modbus.0: error: Invalid signature in mbccb file: '

Because you try to use the source file. The signature clearly shows the XML header. However, the file's magic should be <code>MesaMB01</code>.

andypugh:
Part of the problem here is that I haven't built the docs, so have been scanning the asciidoc source on github rather than reading formatted docs. 
Maybe the error message could add "Have you compiled the mbccs file into mbccb?"

BsAtHome:
> Maybe the error message could add "Have you compiled the mbccs file into mbccb?"

It is now changed so that the message would read:
```
hm2_modbus.0: error: Invalid signature in mbccb file: '<?xml ve' (expected 'MesaMB01')
hm2_modbus.0: error: Have you compiled the mbccs source into a binary mbccb file using mesambccc?
```

andypugh:
We should probably add a deprecation notice to the mesa_modbus page too.

Issue #3395, opened 2025-04-09
file tool table.tbl error formatting


soundlike:
System Debian 12
lcnc 2.10.0
test with axis/sim/random_tc
In spindle T2
if in tool table , you have a tool:

T1 P14 D+0.750000 Z-5.319396 ;3/4 roughing 

and execution t1 m6 or m61 q1 , the formatting text is correct.

if have no comment the tool  in tool table file :

T1 P14 D+0.750000 Z-5.319396 ;
and execution t1 m6 or m61 q1 , the formatting text not is correct, the string next is concatenated al before string

This not is correct, i can't have the comment on tool

soundlike:
Yes there are the problem with zero lenght the comment.

If in tooltable.tbl have one or more tool without comment, the same at : T1 P14 D+0.750000 Z-5.319396 ;
and command a tool change t1 m6 or m61 q1 , the file tooltable.tbl have error in saving.
Not can reopen the tool table  with tooledit.tcl , because file forrmatting have error.

You can test your self , but in next day i can attachment the file

soundlike:
![Image](https://github.com/user-attachments/assets/8b59318f-2436-41c8-ab9e-43c14f2129c8)
![Image](https://github.com/user-attachments/assets/bc2a97f4-221b-4850-827b-0fd655e09058)

THis is screenshot the proble.
When start machine have in spindle t2
After command t35m6 (t35 not have comment after ;)
The table is refreshed, but not correct format , please see picture t35_in_spindle

soundlike:
![Image](https://github.com/user-attachments/assets/b67cf2d6-a4db-4c10-87a5-e37a09fff4f8)
![Image](https://github.com/user-attachments/assets/ff91b6de-8103-4f88-9e20-27201bd2976a)

This is the picture for see the problem
The lcnc start with t2 in spindle, pics t2_in_spindle
Open tooledit and uncommnet t35 and save file tool table
start command t35 m6 and file tool table is refreshed , but with error format , see pics t35_in_spindle

if no tool has comment, the tool table file becomes so unreadable that even tooledit cannot reopen it

andypugh:
Is this with a RANDOM_TOOLCHANGER?

soundlike:
Yes is random tool change

andypugh:
I can reproduce the problem. It isn't in 2.9 (which is a relief). 
[...]

PR #3428, opened 2025-05-01
Issue #3395: Fix tool table formatting error with random tool changer

andypugh:
Issue introduced in https://github.com/LinuxCNC/linuxcnc/commit/088be31097b4cadf5e6e5e65273e43ad7313f367

### DIFF
--- a/src/emc/tooldata/tooldata_common.cc
+++ b/src/emc/tooldata/tooldata_common.cc
@@ -270,9 +270,11 @@ void tooldata_format_toolline (int idx,
 #undef F_ITEM
 #undef I_ITEM
     if (tdata.comment[0]) {
-        snprintf(tmp,sizeof(tmp)," ;%s\n",tdata.comment); \
-        strncat(formatted_line,tmp,CANON_TOOL_ENTRY_LEN-1); \
-    } 
+        snprintf(tmp,sizeof(tmp)," ;%s\n",tdata.comment);
+    } else {
+        snprintf(tmp,sizeof(tmp),"\n");
+    }
+    strncat(formatted_line,tmp,CANON_TOOL_ENTRY_LEN-1);
     return;
 } // tooldata_format_toolline()
 


src/emc/tooldata/tooldata_common.cc
```diff
@@ -270,9 +270,11 @@ void tooldata_format_toolline (int idx,
...
 #undef I_ITEM
     if (tdata.comment[0]) {
-        snprintf(tmp,sizeof(tmp)," ;%s\n",tdata.comment); \
-        strncat(formatted_line,tmp,CANON_TOOL_ENTRY_LEN-1); \
-    } 
+        snprintf(tmp,sizeof(tmp)," ;%s\n",tdata.comment);
+    } else {
+        snprintf(tmp,sizeof(tmp),"\n");
+    }
+    strncat(formatted_line,tmp,CANON_TOOL_ENTRY_LEN-1);
```
smoe, line 277: strncat "appends  at most ssize non-null bytes from the array pointed to by src". It does not check how much is already written to the string or how large the buffer truly is. So, the length of _formatted_line_ in my understanding needs to be substracted from CANON_TOOL_ENTRY_LEN-1 to be safe.

This is all a bit cumbersome. I suggest to not execute the strncat in line 241 and instead keep the value "tmp" intact. Then directly perform the snprintf of line 273 into formatted_line and check the number of characters written. Like this:

new line 273:

```
int charsThatWouldBeWritten snprintf(formatted_line,CANON_TOOL_ENTRY_LEN,"%s ;%s\n", tmp, tdata.comment);
if (charsThatWouldBeWritten<0) {
    fprintf(stderr,"Could not be printed. No good. Maybe this machine should not control any machine.\n");
if (! charsThatWouldBeWritten<CANON_TOOL_ENTRY_LEN) {
    fprintf(stderr,"Ouch! May still be ok if the bits up the comment were copied. Not checked.\n");
}
```
and analogous for the part just attaching the "\n".

This would somewhat guarantee that all non-comment bits are truly read  - This may avoid some weird situation to have some tool's length shortened by a couple of decimals positions and nobody notices that. Admittedly, after a quick check on http://wiki.linuxcnc.org/cgi-bin/wiki.pl?ToolTable it seems unlikely to create such a long line. 

The code does not check if strlen(tdata.comment) < CANON_TOOL_COMMENT_SIZE . This would allow the routine to create formatted lines that do not follow the constraints set by the format.

andypugh: This seems wrong? 
```
char tmp[CANON_TOOL_ENTRY_LEN-1] = {0};
```
Shouldn't tmp be the full size, then we should use sizeof-1 to leave space for the null? 

As for the rest of it, I wonder if this works? 

```
char tmp[CANON_TOOL_ENTRY_LEN-1] = {0};
space = sizeof(tmp);
    space -= snprintf(tmp,sizeof(tmp),"T%-3d P%-3d"
            ,tdata.toolno
            ,is_random_toolchanger ? idx : tdata.pocketno);
    strncat(formatted_line,tmp,space);
// format zero float values as %.0f for brevity
#define F_ITEM(item,letter) if (!ignore_zero_values || tdata.item) { \
                                if (tdata.item) { \
                                    space -= snprintf(tmp,sizeof(tmp)," " letter "%+f", tdata.item); \
                                } else { \
                                    space -= snprintf(tmp,sizeof(tmp)," " letter "%.0f",tdata.item); \
                                } \
                                strncat(formatted_line,tmp,space; \
                            }
#define I_ITEM(item,letter) if (!ignore_zero_values || tdata.item) { \
                                space -= snprintf(tmp,sizeof(tmp)," " letter "%d",tdata.item); \
                                strncat(formatted_line,tmp,space); \
                            }

    F_ITEM(diameter,       "D");
    F_ITEM(offset.tran.x,  "X");
    F_ITEM(offset.tran.y,  "Y");
    F_ITEM(offset.tran.z,  "Z");
    F_ITEM(offset.a,       "A");
    F_ITEM(offset.b,       "B");
    F_ITEM(offset.c,       "C");
    F_ITEM(offset.u,       "U");
    F_ITEM(offset.v,       "V");
    F_ITEM(offset.w,       "W");
    F_ITEM(frontangle,     "I");
    F_ITEM(backangle,      "J");
    I_ITEM(orientation,    "Q");
#undef F_ITEM
#undef I_ITEM
    if (tdata.comment[0]) {
        space -= snprintf(tmp,space," ;%s\n",tdata.comment);
    } else {
        space -= snprintf(tmp,space,"\n");
    }
    strncat(formatted_line,tmp,space);
 
```

PR #3434, opened 2025-05-04
PNCconf - automatic G43 for Gmoccapy

zz912:
Gmoccapy had automatic activation of tool offsets using the G43 command after a tool change. The automatic G43 caused "race condition problems" in some configurations.
This was solved by:
https://github.com/LinuxCNC/linuxcnc/pull/3113

Automatic G43 should be implemented by LCNC remap functionality not GUI.
This Pull Request modifies PNCconf so that the configuration files for Gmoccapy include a remap with AUTOMATIC_G43.

More information:
https://forum.linuxcnc.org/gmoccapy/55011-lcnc-2-10-ask-for-automatic-g43

### DIFF
--- a/src/emc/usr_intf/pncconf/build_INI.py
+++ b/src/emc/usr_intf/pncconf/build_INI.py
@@ -174,6 +174,17 @@ class INI:
                 p =.001
             print ("RS274NGC_STARTUP_CODE = G{} G40 G90 G94 G97 G64 P{}".format(unit,p), file=file)
 
+        if self.d.frontend == _PD._GMOCCAPY:
+            print("SUBROUTINE_PATH = ./macros", file=file)
+            print("REMAP=M6  modalgroup=6 prolog=change_prolog ngc=change_g43 epilog=change_epilog", file=file)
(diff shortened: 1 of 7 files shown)


andypugh:
How (if at all) does this interact with the already-existing RETAIN_G43 setting? 
http://linuxcnc.org/docs/stable/html/config/ini-config.html#sub:ini:sec:rs274ngc

zz912:
Hello Andy,

AUTOMATIC_G43 is not RETAIN_G43

Differences:
AUTOMATIC_G43
- created using remap
- works in all modes (PROGRAM, MDI, MANUAL)
- is always active
- adding G43 after executing M6/M61 is often used in other control systems. Users often add it themselves.

RETAIN_G43
- activated in INI file
- works ONLY in PROGRAM mode
- is active only when the *.ngc program is running and only after the first G43 command in the program has been read. After the *.ngc program has been read, this function is deactivated.
- I think this feature was important in the old days when *.ngc programs were written manually to protect the machine when G43 was forgotten after M6/M61.

PR #3441, opened 2025-05-17
Gmoccapy-integration halui MDI commands

zz912:
When the "halui MDI command" is run, the modes are switched in the background:
1) Original (Manual or MDI or Auto)
2) MDI
3) Original (Manual or MDI or Auto)

This Pull Request ensures that the mode screens in the Gmoccapy GUI are not switched.

### DIFF
--- a/configs/sim/gmoccapy/gmoccapy.ini
+++ b/configs/sim/gmoccapy/gmoccapy.ini
@@ -96,7 +96,11 @@ HALUI = halui
 
 # Trajectory planner section --------------------------------------------------
 [HALUI]
-#No Content
+MDI_COMMAND = M6 T1
+MDI_COMMAND = M6 T2
+MDI_COMMAND = M61 Q1
+MDI_COMMAND = M61 Q2
+MDI_COMMAND = M6 T3 G4 P10
 
 [TRAJ]
 COORDINATES =      X Y Z 
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -213,6 +213,17 @@ class gmoccapy(object):
         self.jog_increments = []  # This holds the increment values
         self.unlock = False       # this value will be set using the hal pin unlock settings
 
+        # if halui MDI commands are not defined in the INI file, the pin halui.halui-mdi-is-running will not be created.
+        try:
+            hal.get_value("halui.halui-mdi-is-running") 
+            # this pin prevents mode switching during execution of halui MDI commands
+            self.halui_cmd_exist = True 
+            # this value will allow reading the halui.halui-mdi-is-running pin only if it exists
+            LOG.debug("halui MDI commands are used")
+        except Exception:
+            self.halui_cmd_exist = False
+            LOG.debug("halui MDI commands are NOT used")
+
         # needed to display the labels
         self.system_list = ("0", "G54", "G55", "G56", "G57", "G58", "G59", "G59.1", "G59.2", "G59.3")
         self.dro_size = 28           # The size of the DRO, user may want them bigger on bigger screen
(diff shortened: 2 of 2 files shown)


Sigma1912:
Maybe @hansu could review this.

gmoccapy:
This patch does the MDI Running Check in the button press functions, IMHO the check should be done in Hal Status change function instead.

On the other Hand I do not like that the code is introduced on several places, so we have the same code on several places instead of having it in its own function.

Norbert

zz912:
> This patch does the MDI Running Check in the button press functions, IMHO the check should be done in Hal Status change function instead.

This patch has three parts:
1) in init it detects if LCNC configuration contains HALUI MDI commands
line 216
2) blocks mode buttons during run HALUI MDI commands
line 2534 - 2558
It is debatable whether this part is important. Theoretically this situation should not occur, because during run commands these buttons are no sensitive. However I have problems with real implementation of LCNC, when during use of remap and HALUI commands I have random horrible things happening and therefore here are logs that could help in finding problems.
3) react to Hal Status change and block screen change
line 2804 - 2888
This is the main part of this patch.

> On the other Hand I do not like that the code is introduced on several places, so we have the same code on several places instead of having it in its own function.

I also thought about enclosing part of the code in a function.

But does it make sense to make a function for 3 lines?
```
    if self.halui_cmd_exist:
        if hal.get_value("halui.halui-mdi-is-running"):
            MESSAGE
            return
```

The problem is that the MESSAGE is different every time.
Moreover, in part 2 I use LOG.error
in part 3 I use LOG.debug

I don't know how to do it so that it doesn't make it even worse for clarity.

Or do you want to move the code in the init part into the function?
To make the init function cleaner?

zz912:
Is there anything I can do to get this Pull Request merged? Or is it just that no one has gotten to it and I should wait?
I don't want us to wait for each other.

andypugh:
> Is there anything I can do to get this Pull Request merged? Or is it just that no one has gotten to it and I should wait? I don't want us to wait for each other.

I saw that @hansu had self-assigned it, so I have not been paying much attention to it.

zz912:
Hello Andy,

Thanks for your response. I don't want to put any pressure on Hans.
I just wanted to make sure that there was no need to do anything on my part.

I didn't want a situation where I would be waiting for Hans and Hans would be waiting for me.

Zdeněk

zz912:
This PR solves the problem with switching modes using a HAL signal. I now know that HAL signals in GUIs are not a very clean solution. On the other hand, I still think that this PR should be merged as a temporary solution to the problem, which will make life more pleasant for users for at least a year, until the switching pages issue is properly resolved.

hansu:
I have it on my list =)

hansu:
I wonder why it is a problem when the screen switches to MDI mode while a MDI command is being executed.
This has the advantage that you get an abort button while the MDI command is running...

zz912:
Short commands are very often used. For example:
MDI_COMMAND = M6 T1
MDI_COMMAND = M61 Q2
MDI_COMMAND = G43
MDI_COMMAND = G49
Mainly in physical panel with physical button.

So you only see a screen flash.
Then there is the problem that it starts very quickly :
**def on_hal_status_mode_mdi(self, widget):**
and immediately after:
**def on_hal_status_mode_manual(self, widget):**
Which causes a race condition on some PCs and some parts of the previous functions are not implemented completely. Then some buttons in gmoccapy are grayed out and they shouldn't be. Mainly it behaves unstable.

I admit that I have never needed long HALUI mdi command.

hansu:
>Then some buttons in gmoccapy are grayed out and they shouldn't be. 

You mean they stay grayed out then due to the race condition?


But for longer lasting MDI commands we should do something on the screen At least disabling the mode-switch buttons and maybe also display a message like "MDI command is currently executed". Otherwise the user just wonders why the GUI doesn't respond.
Do you thin a pop-up dialog is annoying, which pops up while a halui MDI command is running and then disappears automatically?

grafik

zz912:
> You mean they stay grayed out then due to the race condition?

I think Yes. 90%

> Do you thin a pop-up dialog is annoying, which pops up while a halui MDI command is running and then disappears automatically?

The message "External MDI command is currently executed" is not suitable for the operator. The CNC machine operator (not integrator) does not know what an "External MDI command" is. Mayby "Machine is currently working."

 I would make a 1s delay so that this message does not appear for short commands. Pop-up dialog should be annoying, but with automatically dissapearing it is not a big problem for me.



Before you start working on this PR, please read this https://github.com/LinuxCNC/linuxcnc/pull/3580 .
Because then this PR might not make sense.
This is especially important:
https://github.com/LinuxCNC/linuxcnc/pull/3580#issuecomment-3413938283

PR #3446, opened 2025-05-20
Backport: hm2_modbus driver

BsAtHome:
This is the same hm2_modbus driver as added to master (#3427). The only difference is that there is no support for 64-bit HAL types because they do not exists in 2.9 branch.

The code was slightly adapted in two lines to be accepted by the gcc compiler in Debian 10 and 11 (Buster and Bullseye). The old compilers would not accept a label at a combined declaration/assignment and required a label to be attached to a statement. A second adaption was in the mesambccc utility where python 3.7 in Debian 10 needed an alternative syntax for dictionary merges.


hansu:
@BsAtHome Please do the 2.9-to-master-merge, because there are many conflicts you know best to solve.

BsAtHome:
> Please do the 2.9-to-master-merge, because there are many conflicts you know best to solve.

Many changes were _backported_ from master to 2.9 (many by me, in fact). That is how you normally do it, you backport and not forwardport. Master is getting more and more divergent and it becomes harder and harder to cope with 2.9 branch modifications moving forward.

And, BTW, all that backported code I did _cannot_ move back forward. It was specifically modified to be in the 2.9 branch.

But then, I have no clue how to forward port. I have no idea from when to when and need to get more info on how you imagine I do a forward port. Command line? Via github web interface? What tests need to be performed? You will need to give me some guide(lines) here which parts must/should move forward and which parts should not. And, FWIW, the user interface stuff fixed in 2.9 may not be forward portable from what I read. Some are bugs present in 2.9 but not in master or vice versa or both. I am not familiar with that code and have not really paid attention to any modifications.

What I can do is clean a diff/patch, based on a 2.9-to-master target, where all non forward portable things, that I know of, are removed. You still need to tell me exactly from where-to-where to generate the initial diff. Or maybe you have a better idea?

hansu:
> That is how you normally do it, you backport and not forwardport. 

In this project we use the merge upwards workflow...

>Or maybe you have a better idea?

I can do the merge to my fork or to another branch which derives from the current master with basically "git merge -s ours" for the conflicts and you can tell me if everything is fine.

hansu:
Here you can see the conflicts: https://github.com/LinuxCNC/linuxcnc/commit/7604246cd2e9034179ec01cc368b713c36279ce1

And this branch would be the result if I chose the version from master when resolving the conflicts:
https://github.com/LinuxCNC/linuxcnc/commits/master-2.9-merge

Please verify that. So far I can see it, all conflicts are caused by your backports except that one in the gitignore.

andypugh:
This was an unusual case, where a new driver was added to master, and tested there. And then it was suggested that as a new standalone driver it could be safely included in 2.9 too.

BsAtHome:
These files must not be touched in master because there is a good reason why these are different (missing 64-bit hal support in 2.9 and the makefile duplicates the already present rule):
* configs/by_interface/mesa/hm2-modbus/bldc_and_wavebox.mbccs
* configs/by_interface/mesa/hm2-modbus/brushless-dc-motor-driver.mbccs
* docs/src/man/man1/mesambccc.1.adoc
* src/Makefile
* src/hal/drivers/mesa-hostmot2/hm2_modbus.c
* src/hal/drivers/mesa-hostmot2/mesambccc.py

File docs/man/.gitignore is interesting. See added comments.

hansu:
Ok, done. Thanks for your comments.

Issue #3447, opened 2025-05-23
Gmoccapy - Odd behaviour in G96 (CSS) mode


andypugh:
LinuxCNC 2.9.4

Open Gmoccapy, home and go to MDI mode

G96 S100
M3

The HAL pin spindle.0.speed-out and the Gmoccapy display briefly show the correct speed, then go to 100. 

M3 (again)

Now the HAL pin works properly, but the Gmoccapy speed display has many too many digits. 

![Image](https://github.com/user-attachments/assets/bc2ecf12-f51e-4c66-b34b-ee07baf18145)

andypugh:
(Tested with Axis GUI, and the HAL pin part of the problem is not seen)

Sigma1912:
Not sure if this is related but I additionally notice a very large value for 'max acceleration' in the python status.

![Image](https://github.com/user-attachments/assets/4442372d-1a3c-43aa-9eec-717d4a10fe33)

Sigma1912:
Same for the 'acceleration' value:

![Image](https://github.com/user-attachments/assets/123c2164-d9f6-4903-878f-80ed9c56556c)

Note that this is right after starting the config (master branch).

Sigma1912:
The 'infinite' acceleration values are due to missing INI entries:

[TRAJ]
DEFAULT_LINEAR_ACCELERATION 
MAX_LINEAR_ACCELERATION  

However, adding these to the ini file does not fix the described behavior of the spindle rpm value.

Sigma1912:
There seems to be something generally wrong with the python status channel when 'G97' is active (Note the 'speed' entry):

![Image](https://github.com/user-attachments/assets/9b8e7791-072b-49fc-902f-039e55eb41fa)

This after 'M4' command:

![Image](https://github.com/user-attachments/assets/a571b436-2408-468d-a566-ceb34e9327a4)

andypugh:
The HAL pin works fine if you issue a second M3 command. It smells like some sort of special-handling in Gmoccapy, possibly aimed at working around the Python interface problem.

Sigma1912:
Yes, it also works if I add      self.linuxcnc.mdi('g4 p0.1')   after   self.linuxcnc.mdi(cmd):

![Image](https://github.com/user-attachments/assets/4099e779-64c7-46cd-af6e-31f40d8ca82b)

[...]

PR #3450, opened 2025-05-24
Gmoccapy: Fix odd spindle behavior when using G96

Sigma1912:
Fixes [issue #3447](https://github.com/LinuxCNC/linuxcnc/issues/3447):
-  gui spindle speed display showing a very large number when G96 (CSS) mode is used
-  spindle.0.speed-out pin being reset to a wrong value when the gui spindle buttons are updated after calling G96 through MDI

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.glade
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.glade
@@ -3226,6 +3226,7 @@
                                                             <property name="draw-indicator">False</property>
                                                             <property name="group">rbt_stop</property>
                                                             <signal name="clicked" handler="on_rbt_reverse_clicked" swapped="no"/>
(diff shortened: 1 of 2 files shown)


andypugh:
Can you rebase this on 2.9? It's a clear bug.

Sigma1912:
I notice that the relevant functions have been somewhat changed in master and I assume that the added lines will also need to be put into my two new functions. 
This is 2.9:


And this is what it should look like in master (the three last lines in each function are new additions made by @hansu  ):

master

I'm not sure how this should be handled. Please advice.

Sigma1912:
I see, so it seems that my repo is not up to date, thanks.

Sigma1912:
See rebased and improved PR https://github.com/LinuxCNC/linuxcnc/pull/3454

Issue #3064, opened 2024-08-11
Tool table wiodget: expand for column "rotation"


hansu:
![grafik](https://github.com/user-attachments/assets/68ee4fe6-a717-444d-85d1-67871a4f873e)

Sigma1912:
I agree. This makes much more sense as each WCS has it's own 'R' value.

andypugh:
And having a spurious rotation is an occasional cause of trouble. (I saw a recent forum post where someone said that their X and Y axes had swapped. That was the cause.

c-morley:
Does each system have it's own rotation? I thought the rotation was was separate and added after the user system.

Sigma1912:
Yes, each system has it's own rotation. AxisGUI displays R values correctly for each system.

hansu:
Yes, except G92. See https://linuxcnc.org/docs/devel/html/gcode/g-code.html#gcode:g10-l2

c-morley:
What python attribute gives the rotation setting for systems not in use?
rotation_xy only gives the current rotation.

Sigma1912:
As far as I'm aware there are only python attributes for for the currently active WCS system.  I think one would have to query parameters #5230 + i*20, i=[0..8]

c-morley:
Ahh yes ,it's coming back to me now. We read the var file directly for the other user systems. I assume there is a var in that file for the other rotations too.

hansu:
@c-morley Are you currently working on this? If yes - I thought it would be nice if we could use the calculator widget to enter the offset values. What do you think?

hansu:
Fixed by https://github.com/LinuxCNC/linuxcnc/pull/3457

PR #3457, opened 2025-05-28
Gmoccapy: improvements to the offsetpage widget

Sigma1912:
- Remove rows 'G5x' and 'Rot'
- Add rows for G28 and G30 offsets
- Add column 'R' for rotation R value in offsets 'G54' .. 'G59.3'
- Rename column 'name' to 'Comment'

### DIFF
--- a/lib/python/gladevcp/offsetpage.glade
+++ b/lib/python/gladevcp/offsetpage.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.38.2 -->
+<!-- Generated with glade 3.40.0 -->
 <interface>
   <requires lib="gtk+" version="3.24"/>
   <object class="GtkListStore" id="liststore2">
@@ -24,6 +24,8 @@
       <column type="gchararray"/>
       <!-- column-name W -->
       <column type="gchararray"/>
+      <!-- column-name Rot -->
+      <column type="gchararray"/>
       <!-- column-name display_row -->
       <column type="gboolean"/>
       <!-- column-name editable -->
@@ -47,14 +49,15 @@
         <col id="7">0</col>
         <col id="8">0</col>
         <col id="9">0</col>
-        <col id="10">True</col>
-        <col id="11">False</col>
-        <col id="12">0</col>
-        <col id="13">black</col>
-        <col id="14" translatable="yes">Tool</col>
+        <col id="10">0</col>
+        <col id="11">True</col>
+        <col id="12">False</col>
+        <col id="13">0</col>
+        <col id="14">black</col>
+        <col id="15" translatable="yes">0</col>
       </row>
       <row>
-        <col id="0">G5x</col>
+        <col id="0">G28</col>
         <col id="1">0</col>
         <col id="2">0</col>
         <col id="3">0</col>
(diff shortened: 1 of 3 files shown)


Sigma1912:
It seems that this widget might be used by GUIs other than Gmoccapy. I have not been able to find any though and have thus only tested this in Gmoccapy.

Looks like this:

offset_page

hansu:
Really great, that you!
So this fixes https://github.com/LinuxCNC/linuxcnc/issues/3064.

But I would suggest to use 'Rot' for rotation instead of 'R' to make sure that it's not a linear offset.

In which case was 'G5x' displayed? In the above mentioned Issue there is no 'G5x'.

hansu:
Looks good, only one question:

>Rows 1-3 (Tool, G28, G30) are not editable.

Why? Intentional or not possible?

Sigma1912:
Thanks for testing, I have expanded editing to all rows:
- Editing a 'Tool' value  will call 'G43.1 [x,y,z,a,b,c,u,v,w] [value]'
- Editing a 'G28' value will write to the appropriate parameter[5161..5169]
- Editing a 'G30' value will write to the appropriate parameter[5181..5189]

hansu:
>Rows 1-3 (Tool, G28, G30) are not editable.

Sorry I was a bit unclear, I thought about their comment column, but very cool to have them editable!

Only for the tool I am not sure if this could cause some trouble. Of course it is very handy, but imagine if you change the value there and don't change it in the tool table and then change to another tool and then back to this tool...

The default comments for G28 and G30 are now displayed as '0'. Can we have them just empty or name them G28 and G30? (I know a bit duplicated, but consistent to the rest)

And when I look at the tool row - do you think it is possible to write there the comment from the tool table?

PS: It is fine and in most cases also better to have multiple commits in a pull request. It allows me for example to trace your additions and makes is easier to fetch your additions. Fetch updates from a force-pushed HEAD is not so handy.

But anyway, already a good improvement!

Sigma1912:
> Only for the tool I am not sure if this could cause some trouble. Of course it is very handy, but imagine if you change the value there and don't change it in the tool table and then change to another tool and then back to this tool...

The offset page should give an overview of all applied offsets and also a way to quickly change them. 
The offsets in the 'Tool' row do not necessarily only reflect the tool offset values of the loaded tool. These might also contain additional offsets set by G43.2. 
While it would be possible to save these to the tool table I think saved offsets for tools in the tool table should be handled in the tool table page. 

> And when I look at the tool row - do you think it is possible to write there the comment from the tool table?

Following the reasons above I would rather put a note like 'NOT saved to tooltable!' in the comment of the 'Tool' row. The comment from the tool table is already displayed in the 'Tool Information' frame below the offset page.
comment

toolinfo

hansu:
>The offsets in the 'Tool' row do not necessarily only reflect the tool offset values of the loaded tool. These might also contain additional offsets set by G43.2.

Yes of course

>While it would be possible to save these to the tool table I think saved offsets for tools in the tool table should be handled in the tool table page.

No I don't want to change the tool table from here as well.

Yeah maybe it's better to not have the comment from the tool table there to not give the impression that this corresponds always to the tool table.
Maybe a comment like "Working offset" ?

Sigma1912:
I've pushed the change you can easily change the default comment for the 'Tool' row.

hansu:
:+1: 
It wasn't my intention to dictate you what should be in the comment column, it was more to throw out a suggestion.
But if you are fine with that, we can keep it.

Issue #2753, opened 2023-11-18
gmoccapy/qtdragon: preview not respecting WCS other than G54 on file load/reload 


Sigma1912:
This is running 2.9.1 and master.

1. start configs/sim/gmoccapy/gmoccpy
2. load file without WCS definition, for example 'arcspiral.ngc'
3. Note how the spiral is centered on the origin in the preview:
![spiral_centered](https://github.com/LinuxCNC/linuxcnc/assets/46067220/16a70f1a-e03d-435d-8999-4f09ca553e06)

4: run mdi-command 'g55'
5: run mdi-command 'g10 l20 p0 x100'
6: reload the ngc file.
What I expect it that the preview shows the center of the spiral in the new origin of g55 (ie moved 100mm to the right) however the preview only shows the new origin but does not move the preview:
 
![spiral not centered](https://github.com/LinuxCNC/linuxcnc/assets/46067220/abfdb429-9b71-41dc-855a-91abd30f10e4)

[edit] same in qtdragon

This used to work in previous versions (2.9pre) and still works in 'axis'.

Sigma1912:
Not sure what is going on here.
I copied and built the rip install folder from the machine where this works to one of the ones where it does not work and I see the same bug.

Maybe this is dependent on the installed opengl version or something?

hansu:
But if you run the program, the spiral is machined around the new origin?

Sigma1912:
Yes, the machine moves correctly. It's just the preview that is off.

Sigma1912:
I've done a bit of digging in '/lib/python/rs274/glcanon.py' and 'interpret.py'
This, I think, is here is where the program preview is constructed in 'interpret.py' (with my print statements):
![interpret_py](https://github.com/LinuxCNC/linuxcnc/assets/46067220/b0d4a228-3329-4792-b214-b3ed4afaa2d3)

I then ran AXIS gui changed the offset from G54 to G55 and clicked on reload:

![reload arcspiral axis](https://github.com/LinuxCNC/linuxcnc/assets/46067220/a325d91a-d083-4285-82e0-8405728cf4f4)

Then I did the same in gmoccpy:
![reload arcspiral gmoccapy](https://github.com/LinuxCNC/linuxcnc/assets/46067220/0d183293-e888-44cc-a04b-c19cfa6ea897)

[...]

PR #3460, opened 2025-06-06
Don't use  RS274NGC_STARTUP_CODE when updating the preview

Sigma1912:
This fixes issue #2753


andypugh:
Thanks for the fix, and for the clear explanation.

PR #3465, opened 2025-06-07
fix: rs274 nullptr will collapse the milltask when hal para is wrong

fghnntp:
rs274 nullptr will collapse the milltask when hal para is wrong.

### DIFF
--- a/src/emc/rs274ngc/gcodemodule.cc
+++ b/src/emc/rs274ngc/gcodemodule.cc
@@ -193,28 +193,8 @@ static InterpBase *pinterp;
 
 #define callmethod(o, m, f, ...) PyObject_CallMethod((o), (char*)(m), (char*)(f), ## __VA_ARGS__)
 
-static void maybe_new_line();
 static void maybe_new_line(int sequence_number);
-
-static void maybe_new_line() {
-    if(!pinterp) return;
-    if(interp_error) return;
-    int sequence_number = pinterp->sequence_number();
-    if(sequence_number == last_sequence_number)
-        return;
-    LineCode *new_line_code =
-        (LineCode*)(PyObject_New(LineCode, &LineCodeType));
-    pinterp->active_settings(new_line_code->settings);
-    pinterp->active_g_codes(new_line_code->gcodes);
-    pinterp->active_m_codes(new_line_code->mcodes);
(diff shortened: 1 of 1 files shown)


andypugh:
This looks reasonable on inspection. Can you describe how to demonstrate the problem that this fixes?

fghnntp:
When using the rs274 model directly, you might write code like this:
InterBase *b = makeInterp();
b->init();
In the init function, SETG54_OFFSET() and maybe_new_line() is called. If the static pinterp is NULL at this point, it will suddenly cause the process to collapse without any error message or warning.
This makes the rs274 module not robust, especially when used independently.

BsAtHome:
Apparently, my review comments were not seen, again, ...

```diff
- static void maybe_new_line(int sequence_number=pinterp->sequence_number());
+ static void maybe_new_line(int sequence_number=0);
  static void maybe_new_line(int sequence_number) {
      if(!pinterp) return;
      if(interp_error) return;
+     sequence_number=pinterp->sequence_number()
```

Your fix would alter the code's operation. The change will assign `sequence_number` unconditionally. There are also calls where the default argument is overridden, which you now unconditionally replaced (see lines 335, 355, 372, 617 and 629).

fghnntp:
Thank you for your feedback and clarification.
I have updated the code according to your suggestion:

- I removed the default argument that depended on a runtime variable.
- Now, there are two overloaded functions: one with no argument (which uses the current sequence_number from pinterp), and one that accepts a specified sequence_number.
- This avoids unconditionally overwriting the parameter and keeps the logic clear and robust.

Sorry for the earlier mistake, and thank you for pointing it out.

BsAtHome:
Why a full duplication? Can't you just thunk the arg-less version like in:

```c++
static void maybe_new_line(int sequence_number);
static void maybe_new_line();

static void maybe_new_line(int sequence_number)
{
...
}

static void maybe_new_line()
{
    if(!pinterp) return;
    maybe_new_line(pinterp->sequence_number());
}

```

fghnntp:
Thanks for the suggestion! I've updated the code to delegate the argument-less version to the one with sequence_number, removing the duplication.

BsAtHome:
Now it should be fine.

PR #3471, opened 2025-06-14
Gmoccapy: Use calculator widget to edit offset values in offsetpage

Sigma1912:
might be useful.

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -123,7 +123,7 @@ TCLPATH = os.environ['LINUXCNC_TCL_DIR']
 ALERT_ICON = "dialog_warning"
 INFO_ICON = "dialog_information"
 
-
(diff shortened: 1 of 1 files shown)


hansu:
Haha really cool that you take this up - this has been in my backlog for almost 3 years now: https://github.com/hansu/linuxcnc/issues/5

Some notes:

1. I would keep the behavior that you click "edit offsets" and then you can edit multiple offsets and return by clicking this button again. Now you need to click "edit offsets" for every value you change.
2. I would keep the text in the calculator dialog short like "G59 Z-offset:" instead of "Set G59 Z-offset to:"
3. It would be good if you could preset the current value in the calculator. 
For example if you click of the G54 Z column, the Z-value should be already in the entry like this.
This would come close to the request in the forum: https://forum.linuxcnc.org/gmoccapy/46488-and-value-feature#249332
Image

Sigma1912:
Done, thanks for testing.

hansu:
Yeah cool, thanks!

PR #3472, opened 2025-06-15
Gmoccapy: Use calculator widget to edit numeric values in tooltable

Sigma1912:


hansu:
Nice that you also pick this up from my to do list :+1: 
But thinking a bit more about this let me come to the conclusion, that this could be annoying sometimes.
Imagine if you create a new tool table or add several tools, then you have this pop up every time.
So I would suggest to either add a setting or a toggle button "use calculator for editing".
Not sure if it's intuitive enough to have a button there or if someone expects something different there:
grafik

Sigma1912:
Maybe we could offer a TAB function on new tools so that pressing 'Ok' would automatically open the next cell in the calculator?

hansu:
Then we need to add some buttons to the calculator widget...
I tend to have the calculator widget optionally. What do the other think?

Sigma1912:
I have no problem with making the calculator optional.

hansu:
>I have no problem with making the calculator optional.

I would appreciate this ;-)

hansu:
We already had a different behavior for editing the cells in the offsetpage and in the tool table.
For the tool table the row need to be selected first and then the cell can be edited:
select_tooltable

For the offsetpage that is not necessary:
select_offsetpage

I wonder if we should unify the behavior for these two?

Further I wonder if we could highlight the current cell somehow (for both widgets). I mean we have the designator in the calculator widget but a visual highlight would make it quicker to verify if we have clicked the right cell:

grafik

Do you think this is possible?

Sigma1912:
> We already had a different behavior for editing the cells in the offsetpage and in the tool table.

This _might_ be because editing is handled somewhat differently but I have not been able to find the exact reason.
offset_page
tooledit

> Further I wonder if we could highlight the current cell somehow (for both widgets)

I've tried a few things but it seems that the individual cells are not individual widgets (like a button)  and I have not been able to find an example how this would be possible. That's not to say that it is definitely impossible though.

Sigma1912:
Regarding making the calculator popup optional:

I don't think it makes much sense to to this in the settings page as an operator may not have access to that if a pin lock has been set. So it's likely the easiest to implement the screen button you suggested.
Do you have still have the required modification to the glade file with the icon?

Sigma1912:
I guess we would be using the material icon 'calculate'
calculate
As a side question: Should we consider retiring the 'classic' icon theme?

Sigma1912:
We might also want to consider adding the calculator button to the mdi and gcode editor pages:
mdi

program edit

Sigma1912:
> Do you have still have the required modification to the glade file with the icon?

Disregard, I figured it out.

Sigma1912:
Now that I got the buttons for MDI and EDIT modes working I realize that in lathe mode the tool table page has no space to add the calculator button:
lathe

gmoccapy:
First of all:

Thanks for going on, improving gmoccapy day by day! I hope to be back on development from end of the year.

Regarding the calculator widget, I do like the idea a lot. IMHO, it should be implemented in a way, that every entry on every page should be done that way. My intention was to make a GUI for touch screen use, but I have not been able to make that possible in all cases. Now I see a possibility other do my job ;-)

If you follow my suggestion, we should only have a settings entry, or we make a global button placed on the right button bar as additional widget. I would prefer the Settings solution, as normally it will be set only once (use the feature or not).

Norbert

hansu:
> > Do you have still have the required modification to the glade file with the icon?
> 
> Disregard, I figured it out.

I only used the stock icon for this example...


> As a side question: Should we consider retiring the 'classic' icon theme?

Hmm I like parts of the classic theme. But most of them are to colorful IMHO.
I am currently in the process creating an own theme as an mix of the material theme, classic and some CNC standard symbols...

grafik



>We might also want to consider adding the calculator button to the mdi and gcode editor pages:

I think this does not make much sense there because you also have to enter non-numerical values there. So the onboard keyboard would be the better choice for these pages.

Sigma1912:
Just as a demonstration of how the calculator could be used  in mdi mode:
Recording

and in edit mode
Recording

hansu:
>Just as a demonstration of how the calculator could be used in mdi mode:

Ok you mean use it as a calculator not only as numeric input, I got it.

hansu:
>If you follow my suggestion, we should only have a settings entry, or we make a global button placed on the right button bar as additional widget. 

But that would break many hardware designs using aligned hardware buttons on the right side of the screen...

>I would prefer the Settings solution, as normally it will be set only once (use the feature or not).

That would be applicable for the offsetpage and tool table. But for the idea to use it as calculator in MDI mode e.g. that would not work.


For the latter I would tend to a button in the bottom button box. But this box is already full in most cases.

Sigma1912:
> Ok you mean use it as a calculator not only as numeric input, I got it.

Yes. The nicest solution would be if the calculator could be called from the virtual keyboard but I don't know it that is possible. Btw, the current behavior of using the numpad in the virtual keyboard is a bit painful:
Recording

hansu:
>Yes. The nicest solution would be if the calculator could be called from the virtual keyboard but I don't know it that is possible. 

I wouldn't say so in every case. I have a panel with a physical keyboard without numpad. So I prefer to enter numerical values on the touch screen, but for text and more editing I choose the keyboard.

Sigma1912:
> For the latter I would tend to a button in the bottom button box. But this box is already full in most cases.

Maybe we could move the two 'touchoff tool .' buttons to another button page (similar to what we have on the mdi page)?

hansu:
We could introduce a second row...

grafik

Sigma1912:
I certainly like the larger tool table! 
Not sure about the buttons stacked up like that though. I think it would need to blend into the tooltable a bit more.

Sigma1912:
We already have something like it in the preview frame:
preview buttons

Sigma1912:
If we consider adding a row of buttons to the tool table frame then I would suggest to move the 'Delete', 'Add', 'Reload' and 'Apply' buttons into the table frame as well.

andypugh:
The numeric input widget in my lathe macros has some buttons that I find very useful. 
in->mm, mm->in, pitch->tpi and tpi->pitch. 

The second two are probably less useful in general, but very handy for the threading macro, for example cutting an 9tpi thread on a metric lathe. (tpi->pitch then ini>mm)

And the Touchy MDI interface is well worth a look. You only need to type numbers, it knows what letters you will need.

Sigma1912:
> If we consider adding a row of buttons to the tool table frame then I would suggest to move the 'Delete', 'Add', 'Reload' and 'Apply' buttons into the table frame as well.

Maybe like this:
tooltable_proposal

Sigma1912:
> And the Touchy MDI interface is well worth a look. You only need to type numbers, it knows what letters you will need.

touchy-mdi


It's an interesting concept for sure. I like how it interactively guides the operator. 

The calculator is also nice (I like the idea of being able to use axis positions in the calculation):
mdi calculator

Sigma1912:
With the added space we could also add a 'deselect' button:

deselect_proposal

hansu:
>Maybe like this:

Yeah looks good :+1:. I only would let the buttons have the same size like the lower ones. Or at least a bit wider. In this direction we have plenty of space.

>With the added space we could also add a 'deselect' button:

You mean "deselect all"?

Sigma1912:
> You mean "deselect all"?

Yes.

> I only would let the buttons have the same size like the lower ones. Or at least a bit wider. In this direction we have plenty of space.

Sure a _bit_ wider or a bit farther apart is fine. The idea was that it picks up the design of the preview window. 
I would advise against making them too similar to the ones below. This will make it confusing on a panel with aligned buttons along the bottom as it suggests that there is somehow a connection to the buttons in the upper row as well.

Sigma1912:
Unless of course, you are thinking of having the upper row of buttons connect to the physical buttons by some sort of 'shift' hal pin (ie a button combination).

hansu:
>I would advise against making them too similar to the ones below. This will make it confusing on a panel with aligned buttons along the bottom as it suggests that there is somehow a connection to the buttons in the upper row as well.


Yeah you are right, this would make it hard to determine the mapping of the hardware buttons.

>Unless of course, you are thinking of having the upper row of buttons connect to the physical buttons by some sort of 'shift' hal pin (ie a button combination).

I would say that is too much complexity here

hansu:
>Maybe we could move the two 'touchoff tool .' buttons to another button page (similar to what we have on the mdi page)?

But then you loose one button in favor of the arrow for the default view.

Can you add a class variable to make it easier to implement a toggle button or setting later? So that this can be merged?

Sigma1912:
> But then you loose one button in favor of the arrow for the default view.

I believe the dedicated button row at the bottom of the tool table frame (as discussed above) is really the best way forward.

>Can you add a class variable to make it easier to implement a toggle button or setting later? 

Sure, I'll have a look at it tomorrow.

hansu:
Thanks!

PR #3473, opened 2025-06-15
Some improvements to the Gladevcp/calculator widget

Sigma1912:
- inputs and deletions in the entry box are now respecting cursor position instead of being always at the end
- a bit of clean up

### DIFF
--- a/lib/python/gladevcp/calculatorwidget.py
+++ b/lib/python/gladevcp/calculatorwidget.py
@@ -217,9 +217,14 @@ class Calculator( Gtk.Box ):
         if i not in "+-*/" and self.eval_string != "":
             if self.eval_string[-1] == " ":
                 self.eval_string = ""
-
-        self.eval_string = self.eval_string + i
+        cursor_pos = self.wTree.get_object( "displayText" ).get_position()
+        self.eval_string = self.eval_string[:cursor_pos] + i + self.eval_string[cursor_pos:]
         self.wTree.get_object( "displayText" ).set_text( str( self.eval_string ) )
+        if i == 'Pi':
+            cursor_move = 2
+        else:
+            cursor_move = 1
+        self.wTree.get_object( "displayText" ).set_position(cursor_pos + cursor_move)
 
     def displayText_changed( self, widget ):
         self.eval_string = widget.get_text()
@@ -236,10 +241,20 @@ class Calculator( Gtk.Box ):
         if(text == "Error"):
             self.delete()
         else:
-            if text[-2:] == "Pi":
-                self.wTree.get_object( "displayText" ).set_text(text[:-2])
+            cursor_pos = self.wTree.get_object( "displayText" ).get_position()
+            text_left = self.eval_string[:cursor_pos]
+            text_right = self.eval_string[cursor_pos:]
+            if text_left[-2:] == "Pi":
+                self.wTree.get_object( "displayText" ).set_text(text_left[:-2] + text_right)
+                cursor_move = -2
+            elif text_left[-1:] == "P":
+                self.wTree.get_object( "displayText" ).set_text(text_left[:-1] + text_right[1:])
+                cursor_move = -1
             else:
-                self.wTree.get_object( "displayText" ).set_text(text[:-1])
+                self.wTree.get_object( "displayText" ).set_text(text_left[:-1] + text_right)
+                cursor_move = -1
+            if cursor_pos > 0:
+                self.wTree.get_object( "displayText" ).set_position(cursor_pos + cursor_move)
 
     def displayLeftBracket( self, widget ):
         self.displayOperand( "(" )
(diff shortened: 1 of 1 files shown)


Sigma1912:
Is there any reason for not using the unicode '\u03C0' for the small greek letter "pi" instead of the string "Pi"? 
<pre>
&#x03C0
</pre>
That would make almost all the special handling of the two character string 'Pi' in the code obsolete.

Maybe we could also replace the 'DEL' with the more correct '\u232B':
<pre>
&#x232B
</pre>

Like this:
calc

hansu:
> Is there any reason for not using the unicode '\u03C0' for the small greek letter "pi" instead of the string "Pi"?
> 
> π

I guess it's because your keyboard doesn't have a π-key, but you can enter 'Pi' with your keyboard....



> Maybe we could also replace the 'DEL' with the more correct '\u232B':
> 
> ⌫

Yes, we can change that.

Sigma1912:
Looks like this now:

calc_widget

hansu:
I prefer the small font size (like in the rest of the GUI). 
Squeezing the text with the large font size on the buttons just makes them look ugly using the standard theme.
And also I have duplicated icons here which you seem not to have :thinking: 

Before:
grafik
After:
grafik

andypugh:
Are there any objections to this? I don't use Gmoccapy so feel unqualified to do the merge.

hansu:
I don't agree to the graphical changes except that one with the backspace-symbol.

> And also I have duplicated icons here which you seem not to have 🤔

I was talking about duplicated icons - text and icons are fine. Maybe you misunderstood me...

And I don't really get the point with the cursor. Can you give an example of the behavior before and after?

hansu:
>And I don't really get the point with the cursor. Can you give an example of the behavior before and after?

@Sigma1912 ?

Sigma1912:
I don't have access to a linuxcnc installation to record the behavior but the issue is easily reproduced. Note that the issue does not appear if we use the keyboard to type, so make sure you use the mouse or touchscreen to press the buttons on the calculator widget: 

Example 1

1. open the calculator
2. type '12345'
3. try to use the mouse/touch to delete the '3' (ie but the cursor between the '3' and the '4' and press 'DEL')
4. result is that the '5' is deleted.

Example 2

1. open the calculator
2.  type '12345'
3. try to use the mouse/touch to make the number negative (ie but the cursor to the left of  the '1' and press the '-')
4. result is that the '-' is added to the end '12345-'.

hansu:
Ah good catch! I never noticed (used) that.

hansu:
Your commit [Allow using calculation result in new operations](https://github.com/LinuxCNC/linuxcnc/pull/3473/commits/ad56459e6ad568bf2d146f0dd9fefc39b0809aed) is a bit contrary to my change 4 years ago: https://github.com/LinuxCNC/linuxcnc/commit/450dabe1919ed634192592e26a4421eb070b23f5

I don't remember why I did it that way. @ others: opinions?

Sigma1912:
> I don't agree to the graphical changes except that one with the backspace-symbol.

Generally I find the mix of icons and text labels somewhat unfortunate as it breaks the styling of the buttons. I can kind of understand the use of icons for buttons like 'OK' and 'Cancel' but I don't really see a point for any of the others.
In the end it is your decision.   
Note that when using material icon themes apparently only the 'Pi' icon is shown.

hansu:
It's just based on the old-style Gtk dialog buttons.

grafik

Yeah maybe the labels "Ok" and "Cancel" are superfluous as the icons should be clear enough.

hansu:
Compromise?

grafik

Sigma1912:
sure, note though that the 'Pi' icon does not seem to handle the dark theme all that well:
pi


 Unfortunately using the unicode  '\u03C0'  for the small greek letter 'pi' seems to be translated to the capital letter Pi.

Sigma1912:
Seems that using a different font 'font-family: Arial, sans-serif;' solves the pi unicode issue:
arial

This way we could do without the pi icon, what do you think?

Sigma1912:
```
        self.wTree.get_object( "Pi" ).set_label('\u03c0')
```

```
        css = b"#calc_widget {font-family: Arial, sans-serif;}"
```

hansu:
>This way we could do without the pi icon, what do you think?

Yeah that's fine.

I just increased the font size for that symbol a bit (because it is a small letter)

`#pi-symbol {font-size: 18px; font-family: sans-serif;}`

https://github.com/hansu/linuxcnc/commits/calculatorwidget-respect-cursor-position

grafik


> Note that when using material icon themes apparently only the 'Pi' icon is shown.

Hmm I get also the OK and Cancel button. But yes they should be added to the icon theme to be able to adapt to the material theme.
grafik

hansu:
I have done the icon changes and added a prefix in the commit messages in this branch https://github.com/linuxcnc/linuxcnc/commits/calculatorwidget-respect-cursor-position
If you are fine with the changes, I suggest to force-push that branch into your branch so that we can merge it via GitHub.

hansu:
Weird that I am still the Committer of the commits that I rebased and you force pushed. However...
Thanks!

Issue #3469, opened 2025-06-14
Docs G43.1: wrong description of behavior in example gcode


Sigma1912:
Looking at this explanation of 'G43.1' on the G-Codes page makes me think that the comment in the example does not correctly describe the behavior:

![Image](https://github.com/user-attachments/assets/e1d66804-6806-475e-88e7-da4397fc3641)

https://linuxcnc.org/docs/html/gcode/g-code.html#gcode:g43.1

The above is not what I see when testing (master). In my opinion the 3rd line should read:

```
G43.1 Z0.250 (replace current tool offset with 0.250, DRO now shows Z0.250)
```

andypugh:
Fixed by c07dfed71d81b139c07c259ef886ae455b9278ea

But I don't want to close this yet, as in the process I spotted:

```
////
FIXME Use of G43.1 with I and K words ?
To use dynamic tool length compensation in a program, use 'G43.1 I- K-',
where 'I-' gives the compensation of tool length in X (for lathes) and 'K-'
gives the compensation length in Z (for lathes and milling machines).
////
```

Which I think is just wrong? Maybe very old? I am sure you would just use X and Z.

Also, there is no mention that G43.2 can also take axis words to apply incremental offsets. (the docs only describe using H)

Sigma1912:
Thanks.

> Fixed by https://github.com/LinuxCNC/linuxcnc/commit/c07dfed71d81b139c07c259ef886ae455b9278ea

If I may, I'd suggest to replace 'offset current tool offset' with '_replace_ current tool offset'



> Which I think is just wrong? Maybe very old? I am sure you would just use X and Z.

I agree, possibly a very old feature idea. Current implementation of G43.1 does not support any other than axis words:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/rs274ngc/interp_convert.cc#L6235

I've filed a PR for this

Sigma1912:
completed

PR #3479, opened 2025-06-19
Docs, clarify G43.1 and G43.2

Sigma1912:
for discussion see https://github.com/LinuxCNC/linuxcnc/issues/3469


andypugh:
Just one tiny tweak, and I will push it.

PR #3481, opened 2025-06-20
Abort homing on all joints when encoutering an error while homing any single joint

### LINKED ISSUE #2983: Broken homing error handling! If one joint fails to home the other joints will continue anyways.
# The issue tracker is not a support forum

The LinuxCNC issue tracker is to report bugs in the software.
If you have a question about how to use the software, use one of the other methods detailed on our community support page: http://linuxcnc.org/community/

(delete this section before submitting your bug report)

My machines, z homes first. Then x any y together. Noticed this issue when my z stepper wire broke. 

 1. Unplug the stepper wire going to the  joint which homes first. In my case z axis
 2. Home the machine.
 3. Linuxcnc will throw an error saying it failed to home z but.....
 4. X and y axis will start homing after the error shows up causing the head to crash. 

Z joint tries to home, fails, pops up a warning saying z joint failed to home. Machine then stops. 

Their should be an option in ini file to disable this function for those without homing switches.

Machine crashes head cause the z axis was not moved up before x and y due to broken stepper motor wire

Don't think this worked

Seems to be an issue across linuxcnc. Tried it on a few machines and was able to replicate. Other users are reporting the same issue.
https://forum.linuxcnc.org/plasmac/48236-qtplasmac-2-10-arc-ok-thcad-thc-and-thcad-ohmic-troubleshooting-help-needed?start=40

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`):
 * I am using this kernel version (shown by `uname -a`):
 * I am running ...
   * [ ] A binary I built myself
 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`):
 * I am using this user interface (GUI) (e.g., AXIS, Touchy, gmoccapy, etc):
 * I am using this interface hardware vendor and chipset (e.g., parallel port, ethernet port, FPGA card):

itaib:
fixes #2983

There are 2 issues fixed with this PR:
1) The HOME_ABORT state in the homing state machine only aborts the homing of a single joint while other joints continue to travel, leading to potential crashes.
2) In all but one case (in the HOME_FINAL_MOVE_WAIT state), despite an error being reported, the state moved to  HOME_IDLE instead of HOME_ABORT.

### DIFF
--- a/src/emc/motion/homing.c
+++ b/src/emc/motion/homing.c
@@ -757,7 +757,8 @@ static int base_1joint_state_machine(int joint_num)
             if (H[joint_num].home_flags & HOME_IS_SHARED && home_sw_active) {
                 rtapi_print_msg(RTAPI_MSG_ERR, _("Cannot home while shared home switch is closed j=%d"),
                                 joint_num);
-                H[joint_num].home_state = HOME_IDLE;
+                H[joint_num].home_state = HOME_ABORT;
+                immediate_state = 1;
                 break;
             }
             /* set flags that communicate with the rest of EMC */
@@ -815,7 +816,8 @@ static int base_1joint_state_machine(int joint_num)
                 } else {
                     rtapi_print_msg(RTAPI_MSG_ERR,
                          _("invalid homing config: non-zero LATCH_VEL needs either SEARCH_VEL or USE_INDEX"));
-                    H[joint_num].home_state = HOME_IDLE;
+                    H[joint_num].home_state = HOME_ABORT;
+                    immediate_state = 1;
                 }
             } else {
                 if (H[joint_num].home_latch_vel != 0.0) {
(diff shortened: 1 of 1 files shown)


c-morley:
If I home an axis successfully, then home another unsuccessfully, it unhomes the first axis.
Is that intended?

itaib:
I won't say it was intended, but it seems to me like reasonable desired behavior. 
re-homing a failed joint might rely on a previous joint being at the home position, which might not be the case if it's been moved after homing and not re-homed. 
for example, the z axis being at home position before homing x,y, so that the head doesn't crash into fixtures.
if were z homed successfully, but x then failed,  when attempting to re-home x it wouldn't be guaranteed that z is at the correct position if it remained homed.

c-morley:
I see what you are saying but I would account that to the operator.
If you are auto homing all the joints, then yes it should home all the joints again.
If I ask it to home one joint then I expect to make sure it is in a safe position to do so.

I was hoping someone else would chime in. If I don't hear something soon, I will merge and then see if anyone complains :)

PR #3482, opened 2025-06-24
HAL pin "conventional_directions" for configuration of rotational axis direction

DerAndere1:
Following discussion in PR https://github.com/LinuxCNC/linuxcnc/pull/2443:

Introduce HAL pin "conventional_directions" for configuration of rotational axis direction with selected kinematics.
For maxkins, xyzac-trt and xyzbc-trt kinematics, the value of this pin is set to 0 (false) by default so that behaviour of these kinematics is not changed by default for backwards compatibility.
The pin can be set to 1 (true) if the user wants conventional axis direction as defined in the NIST RS274NGC interpreter, version 3 and at https://linuxcnc.org/docs/html/gcode/machining-center.html#_rotational_axes.

I have no idea how to do the same for "5axiskins", so issue https://github.com/LinuxCNC/linuxcnc/issues/2435 should stay open.
This should probably be unit-tested and/or tested using a simulator, but I am not familiar enough with LinuxCNC.

### DIFF
--- a/configs/sim/axis/vismach/5axis/table-rotary-tilting/xyzac-trt.ini
+++ b/configs/sim/axis/vismach/5axis/table-rotary-tilting/xyzac-trt.ini
@@ -69,6 +69,7 @@ HALCMD = sets :z-offset     10
 HALCMD = setp xyzac-trt-kins.x-rot-point 0
 HALCMD = setp xyzac-trt-kins.y-rot-point 0
 HALCMD = setp xyzac-trt-kins.z-rot-point 0
+HALCMD = setp xyzac-trt-kins.conventional-directions 0
 
 [HALUI]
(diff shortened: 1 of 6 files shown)


BsAtHome:
Have you tested it locally? It fails to compile.

The `emc/kinematics/trtfuncs.c` file line 64 is missing a reference in the definition on the pin.

Sigma1912:
I'm opposed to have another copyright added:
 
copyright

DerAndere1:
Thanks, I updated this PR accordingly... A similar test to what can be found under tests/matrixkins would be interesting. The configs under config/sim/axis/vismach/table-rotary-tilting are a good starting point but I still fail to set up a test. I will have to leave that task for someone else.

andypugh:
Sorry for the delay, I have been rather busy with work this week.

PR #3483, opened 2025-06-24
Pncconf add Mesa 7i95t support

kworm83:
This pull requests adds support for the Mesa 7i95t board to pncconf.   The first encoder port on the 7i95t has the pin number 0 so I had to correct the true checking on self.d._arcvpin to check for != None or the THCAD screen wouldn't display.


c-morley:
Thank you.

PR #3489, opened 2025-06-30
Gmoccapy: Glade, use ComboBoxText instead of ComboBox to avoid gtk warning on startup

Sigma1912:
I'm not exactly sure what is wrong with the current 'GtkComboBox' implementation but using 'GtkComboBoxText' instead gets rid of this gtk related warning on startup:
gtk_warning on startup

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.glade
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.glade
@@ -687,40 +687,6 @@
     <property name="visible">True</property>
     <property name="can-focus">False</property>
   </object>
-  <object class="GtkListStore" id="lst_button_mode">
-    <columns>
-      <!-- column-name button_mode -->
-      <column type="gint"/>
-      <!-- column-name button_mode_description -->
-      <column type="gchararray"/>
-    </columns>
-    <data>
-      <row>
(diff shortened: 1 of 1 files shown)


hansu:
But GtkComboBoxText needs more space. With English locales it just fits the window width but in other translation it makes the window wider.
@Sigma1912 can you look for a fix?
grafik

hansu:
Another question - how did you trace the error message so that you end up in the mouse button mode combo box?

Sigma1912:
> Another question - how did you trace the error message so that you end up in the mouse button mode combo box?

There _may_ be a more sophisticated way to do this but this is what I do:
I open 'gmoccapy.glade' in a text editor, scroll about halfway down the code and introduce a syntax error (eg by replacing 'object'  with 'objec') then I start the config and check the terminal output to find out whether the 'Gtk-CRITICAL ...' error is printed before the syntax error aborts the starting process or not. This indicates whether the cause is in the first or second half of the glade code.  Knowing this I can then move the syntax error to the middle of the remaining half and check the terminal output again.  I repeat this until it points me to the object causing the error, so kind of what 'git bisect' does.

hansu:
Interesting to use the glade file for that, thanks :+1:

PR #3492, opened 2025-07-01
Gmoccapy - Documents - update Embedded Tab locations

zz912:
I wanted to add a new location for Embedded Tab to Gmoccapy.
I need to put some *.glade on the setup page (9).
After studying gmoccapy.py I found that Gmoccapy already has this functionality. Just use the localization "ntb_setup".
So I just updated the Gmoccapy documentation.


hansu:
Good catch!
But can you combine `gmoccapy_embedded_tabs-c.png `and `gmoccapy_embedded_tabs.png` into one image?
And the position of 3a and 3b is wrong! Good to point that out because with the introduced G-code property tab it can be mixed up. Correct would be the upper part of the tab "Tool info and G-codes" for 3a and the lower part for 3b.
And I don't think we need an extra image to show the position for "ntb_setup".

zz912:
I fulfilled your requirements.

hansu:
Thanks!

Issue #982, opened 2020-11-19
Spindle using other speed than in GCode, 2.8


mydani:
Here's a short clip about the behaviour:
https://www.youtube.com/watch?v=7x6jb6AMuRA

## Here are the steps I follow to reproduce the issue:

 1. Load a GCode with M3 and M4 commands using different spindle speed and place M1 after each command.
 2. Run it once.
 3. Re-Run it - check the speed of the spindle at 1st m3 - it's not what it is set to in GCode.

## This is what I expected to happen:
Spindle always uses the speed given within the GCode.

## This is what happened instead:
Spindle used the RPM last used in the previous program.

## It worked properly before this:
No clue about the version, but it works properly if some axis movements happen before the M3/M4 command.

## Information about my hardware and software:

 * 2.8 vanilla + gmoccapy
 * 3.x RTAI something...
* mesa card

andypugh:
I think that this might be a Gmoccapy display problem. 

I ran a test with Axis and got the result I expected (when watching the spindle.0.speed-out pin, as Axis does not have an on-screen speed command display) 

Can you repeat your test with a halmeter on the spindle.0.speed-out pin and report if the problem is seen there too?

andypugh:
Any update on this? Can I close the issue?

mydani:
Hello Andy,
Sorry for not getting back until now - I am currently rebuilding parts of my work shop. I will test until end of next week!

andypugh:
Did you complete rebuilding your workshop? Is this issue still open in 2.9.4 ?

Sigma1912:
I just tested on 2.9.4 with the following gcode and cannot reproduce this issue.
```
g0 x0
S100 M3
g0 x5
M1
S200 M4
g0 x10
M1
g0 x5
m2
```

Sigma1912:
HOWEVER, testing with this gcode makes strange things happen to the spindle speed:
```
g0 x0
S100 M3
M1
g0 x10
S200 M4
M1
g0 x5
m2
``` 

![Image](https://github.com/user-attachments/assets/cdcb9429-9e01-4173-9242-87d84058b0eb)

This does not happen in AXIS GUI.

This might be fixed with this PR https://github.com/LinuxCNC/linuxcnc/pull/3454.
But I'll have to test this tomorrow.

andypugh:
Thanks @Sigma1912

PR #3494, opened 2025-07-03
Gmoccapy 3.5.1

hansu:
Main changes:

 - Add button "refresh dir" in open file page
 
- Main window / jog:
  - Add icon for continuous jog, change alignment of jog increment button box
  - Add setting for jog button size + right pane width
  - Style velocity/spindle info box (remove S-prefix, add units)
  
  jog

 - G-code editor:
    - Add button to show preview beside editor (split view). **The split-state is currently remembered during the session. Not on exit.**
    - Add comment/uncomment button for G-code + keybinding
    - Re-arrange search and replace bar

   gcode-edit

- MDI-entry: bigger enter button (touch-friendly) 
  
  MDI-button

And the changes from other PRs which are already in master:
 - Add button to call calculator in mdi and gcode edit modes (#3496, thanks Sigma1912)
 - Fix several warnings. Thanks Sigma1912
 - Improvements to the calculator widget (#3473, thanks Sigma1912)
 - Use calculator widget to edit numeric values in tooltable (#3472, thanks Sigma1912)
 - Use calculator widget to edit offset values in offsetpage (#3457, thanks Sigma1912)
 - Improvements to the offsetpage widget. Especially add column for rotation in offsets G54..G59.3 (#3457, thanks Sigma1912)
 - Fix issues #3447 and #982 (#3454, thanks Sigma1912)
    - Fix the display of very large numbers for spindle speed and 'Vc' in the gui when using G96
    - Fix 'spindle.0.speed-out' pins being reset to wrong values when using G96
    - Fix wrong spindle speed set when using M1 with optional stop enabled
    - The 'Spindle [rpm]' value will now track hal pin 'spindle.0.speed-out'
    - The 'S' value in the 'G-Code' frame now tracks 'self.stat.settings[2]'
 - Fix external_load function -wrong function name - makes external loading of files possible again (Thanks CMorley)

@PeterMue Can you please add resp. recreate the new icons to your https://github.com/PeterMue/gmoccapy-icon-themes? For those I get weird colors, but I think you will add them as SVG and then generate the PNGs so this problem will disappear I guess. Only for the Comment-button I haven't created a symbolic one. Thanks!

Sorry @Sigma1912 @BsAtHome @zz912 that I create this in favor of reviewing your PRs, but these changes are mostly from march and I will get this done to avoid conflicts especially when there will be some activity in the LinuxCNC meeting in Oslo this weekend.


PeterMue:
I'll take a look

zz912:
Hi,

I tested 3.5.1 and here are my findings:
1) is possible make tooltips for continous buttons? Not everyone will understand the symbol, so a tooltip might help.
2) The same "back change" symbol and the same "back to main menu" symbol are close together. Would it be possible to differentiate them at least a little?
Back
3) Calculator in tooltable is good, but it put numbers with decimal comma. After Apply it crashes.
Convert_string

hansu:
> Hi,
> 
> I tested 3.5.1 and here are my findings:
> 
>     1. is possible make tooltips for continous buttons? Not everyone will understand the symbol, so a tooltip might help.

Sure.

>     2. The same "back change" symbol and the same "back to main menu" symbol are close together. Would it be possible to differentiate them at least a little?

I tried but it was rejected by the community. I didn't remember the issue or thread, but I kept the images in the latest release: https://github.com/LinuxCNC/linuxcnc/pull/3335/commits/f3a81369bc5e133c9e5cca2f8b5501a226d6b4ca

>     3. Calculator in tooltable is good, but it put numbers with decimal comma. After Apply it crashes.

Oh, I didn't test that. The everlasting problems with the localization in the tool table...
@Sigma1912 We decided to drop localization in the tool table, but the calculator still does. For the offsetpage it fits but here not.

hansu:
@Petermue Please also for the button that came via https://github.com/LinuxCNC/linuxcnc/pull/3496. Thanks!

Sigma1912:
> We decided to drop localization in the tool table, but the calculator still does. For the offsetpage it fits but here not.

I cannot test this right now but I think replacing this line:
```
                store[row][col] = locale.format_string("%11.4f", value)
```
https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/usr_intf/gmoccapy/gmoccapy.py#L1992

with this
```
                store[row][col] = ("%11.4f" % value)
```
should do it.

hansu:
>                store[row][col] = ("%11.4f" % value)

That would bee to easy :smile:. But yeah, almost that :wink:  (4bcfd3e1ba3e6838ed284d2df454cb9d5bfd0ec5). Thanks.

zz912:
> We decided to drop localization in the tool table, but the calculator still does. For the offsetpage it fits but here not.

I would rather disable localization in the calculator.
Decimal comma makes only problems in technical software.

hansu:
>I would rather disable localization in the calculator.
Decimal comma makes only problems in technical software.

Yes, but it should be possibly to use the comma key on the keyboard as decimal separator.
I suggest to show only the dot in the calculator and treat comma and dot as decimal separators.

zz912:
> > I would rather disable localization in the calculator.
> > Decimal comma makes only problems in technical software.
> 
> Yes, but it should be possibly to use the comma key on the keyboard as decimal separator. I suggest to show only the dot in the calculator and treat comma and dot as decimal separators.

It was Norberts conditions.

Maybe this will help:
https://github.com/LinuxCNC/linuxcnc/blob/4bcfd3e1ba3e6838ed284d2df454cb9d5bfd0ec5/lib/python/gladevcp/tooledit_widget.py#L479

hansu:
Hmm but weird that it crashes though.

zz912:
It crashed here:
https://github.com/LinuxCNC/linuxcnc/blob/4bcfd3e1ba3e6838ed284d2df454cb9d5bfd0ec5/lib/python/gladevcp/tooledit_widget.py#L354
In **save** function:
https://github.com/LinuxCNC/linuxcnc/blob/4bcfd3e1ba3e6838ed284d2df454cb9d5bfd0ec5/lib/python/gladevcp/tooledit_widget.py#L320


Not in function :
https://github.com/LinuxCNC/linuxcnc/blob/4bcfd3e1ba3e6838ed284d2df454cb9d5bfd0ec5/lib/python/gladevcp/tooledit_widget.py#L463

The col_editted function was probably skipped by the calculator.

Sigma1912:
> Yes, but it should be possibly to use the comma key on the keyboard as decimal separator.

We could have the calculator widget accept ',' as well as '.'  by converting all commas to dots:

Changing this
```
        temp = temp.replace(",", ".")
```
https://github.com/Sigma1912/linuxcnc/blob/master/lib/python/gladevcp/calculatorwidget.py#L174

to this: 
```
        temp = temp.replace(",", ".")
```

Sigma1912:
An other option would be to do it in the signal handler function similar to what is done in 'col_editted':

Changing this
```
            else:
                store[row][col] = f"{value:11.4f}"
```
https://github.com/Sigma1912/linuxcnc/blob/master/src/emc/usr_intf/gmoccapy/gmoccapy.py#L1992 

to this
```
            else:
                value = float(str(value).replace(",", "."))
                store[row][col] = f"{value:11.4f}"
```

Sigma1912:
I think I would prefer the second option thus not allowing mixing comma AND dots in the calculator but either commas OR dots depdending on the locale set.
Also it wouldn't disable the calculator widget from handing back float values according to locale setting. 


> The col_editted function was probably skipped by the calculator.

Yes, the calculator uses the 'editing-started' signal which is emitted as soon as a cell is in editing mode while the 'edited' signal is emitted _after_ the cell has been edited using the keyboard.

zz912:
[Sigma1912]
> I think I would prefer the second option thus not allowing mixing comma AND dots in the calculator but either commas OR dots depdending on the locale set.
Also it wouldn't disable the calculator widget from handing back float values according to locale setting.

I would say that history is repeating itself here.
Please see:
https://github.com/LinuxCNC/linuxcnc/issues/2966#issuecomment-2254653717

I also preferred to use only comma, or only dot.
But then we found out that somewhere decimal dot is necessary.
For example, using decimal comma in g-code is nonsense.
Using a decimal comma in DRO is ugly.
I concluded from the above discussion:
1) Use decimal point in all localizations.
2) Make inputs resistant decimal comma.

If we could agree on this, we could make it a rule and list it here:

Sigma1912:
@zz912 
So are you saying that you would prefer the calculator to hand back only values with decimal point (ie option 1)?  
That's also fine with me as I live in decimal dot land.

zz912:
> So are you saying that you would prefer the calculator to hand back only values with decimal point (ie option 1)?

Yes
> That's also fine with me as I live in decimal dot land.

You are lucky man.

Sigma1912:
Could you test this please?
In 'lib/python/gladevcp/calculator_widget.py'  around line 210 change this:

```
        except:
            b = "Error"
        self.entry.set_text( b )
        self.entry.set_position(len(self.eval_string))
```
to this:
```
        except:
            b = "Error"
        b = b.replace(",", ".")
        self.entry.set_text( b )
        self.entry.set_position(len(self.eval_string))
```

This should have the advantage of only accepting the decimal symbol set in the locale (ie dot OR comma but not both) and then hand back values with decimal dot.

zz912:
I'll look at it tonight.

zz912:
It did not help.

I tested it with print command:
```
        except:
            b = "Error"
        b = b.replace(",", ".")
        self.entry.set_text( b )
        print(b)
        self.entry.set_position(len(self.eval_string))
```

b was with decimal point, but in table cell was comma.

It is weird.

If you want to test other localizations, here is the manual:
https://github.com/LinuxCNC/linuxcnc/issues/2966#issuecomment-2211733881

hansu:
>That's also fine with me as I live in decimal dot land.

Where do you live? Because your name sounds German... Or just a German living abroad? :thinking:

Sigma1912:
I'm in Switzerland

hansu:
Ah, didn't know that they use dots.

Sigma1912:
> Ah, didn't know that they use dots.

I think we are about the only ones in europe beside the irish and the brits


I notice that the offset page uses decimal comma format, is that intended? Seems somewhat odd to use decimal dot in the dro and the tool table but decimal comma in the offset table:
offsetpage

Wouldn`t it be preferable to have the decimal dot format everywhere in Gmoccapy?

hansu:
> I notice that the offset page uses decimal comma format, is that intended? 

I think this was originally intended by Norbert, like the tool table also used localization.
But now as we agreed to use dots in the tool table and tool information, we should use dots everywhere in Gmoccapy.

zz912:
> Wouldn`t it be preferable to have the decimal dot format everywhere in Gmoccapy?

Yes, yes, yes. My opinion. But don't forget to accept decimal comma in inputs. The best is plain:
``.replace(",",` ".")``

Sigma1912:
See https://github.com/LinuxCNC/linuxcnc/pull/3500

PeterMue:
@hansu i'm almost done with the material icons. I've slightly changed the design to match the one from the original material library (like undo/redo are well known icons).

Accordingly, I would complete the material icons with the following:
image
(chevron_down, comment, jog-continuous, redo, split_view, undo)

I would also suggest to change the "continuous jog" icon with an "infinity" icon instead of the "fast-forward".

How do you want to integrate the changes? I could open a PR for the main branch, since it doesn't matter whether the icons are already in use or not - that's probably the easiest.

hansu:
>@hansu i'm almost done with the material icons.

Cool, thanks!

The infinity symbol makes sense, but for me it took a second look. I am open to try it - let's see what the community says.
For this symbol I got inspired by a BZT control:
grafik

>How do you want to integrate the changes? I could open a PR for the main branch, since it doesn't matter whether the icons are already in use or not - that's probably the easiest.

I first thought about a PR against this branch, but you are right - against master would be easier :+1:

PeterMue:
I'm not fully confident with the icon either, but at least in my head it makes some sense.
I also considered whether a tilde (~) might be suitable, i think have seen a similar icon on a sinumerik.

Let's give it a try, if it fails, feel free to assign me a issue to change it.

PeterMue:
By the way, IMHO are the undo/redo buttons in the wrong order:
image

Usually left is "undo" and right is "redo".

I'm not sure if there is any "background story" i don't know, but fixing the order is as simple as:

```diff
diff --git a/src/emc/usr_intf/gmoccapy/gmoccapy.glade b/src/emc/usr_intf/gmoccapy/gmoccapy.glade
index 3c04ad8ab6..d27c7cde00 100644
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.glade
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.glade
@@ -1940,7 +1940,7 @@ uncomment selection</property>
                                 <signal name="clicked" handler="on_btn_undo_clicked" swapped="no"/>
                               </object>
                               <packing>
-                                <property name="left-attach">8</property>
+                                <property name="left-attach">7</property>
                                 <property name="top-attach">0</property>
                                 <property name="height">2</property>
                               </packing>
@@ -1962,7 +1962,7 @@ uncomment selection</property>
                                 <signal name="clicked" handler="on_btn_redo_clicked" swapped="no"/>
                               </object>
                               <packing>
-                                <property name="left-attach">7</property>
+                                <property name="left-attach">8</property>
                                 <property name="top-attach">0</property>
                                 <property name="height">2</property>
                               </packing>
```

hansu:
>By the way, IMHO are the undo/redo buttons in the wrong order:

Oh yes, thanks! :+1:

Sigma1912:
Could somebody explain how to the icons in the 'material-light' theme are created?

I have tried matching the color of the other icons in the material-light folder for the new 'calculator_open.symbolic.png' but the color in the gui is off:
light theme

The icon is located here in current master:
https://github.com/LinuxCNC/linuxcnc/blob/master/share/gmoccapy/icons/material-light/32x32/actions/calculator_open.symbolic.png

I need to add a couple more, so I would like to know how to create these myself.
Thanks

PeterMue:
@Sigma1912 Check this out: [https://github.com/PeterMue/gmoccapy-icon-themes (Section "Symbolic Icon](https://github.com/PeterMue/gmoccapy-icon-themes?tab=readme-ov-file#symbolic-icons)

This is the source of the material icons, if you like, please send me a PR or let me know, then I will add the new icons.

hansu:
Also a bit about that in the docs: https://linuxcnc.org/docs/html/gui/gmoccapy.html#gmoccapy:icon-theme-section

Sigma1912:
Two more icons ('Add' and 'Delete') in this recent PR
https://github.com/LinuxCNC/linuxcnc/pull/3509

Sigma1912:
@hansu 
We had a small discussion about the tool table and using 'switch to highlighted' instead of 'switch to checked' tool somewhere. 
I can't find it, but I was wondering if the 'Select'/'Auswählen' column could be 'Delete'. With my PR the 'Delete' button is right underneath that column so it might be fairly intuitive.

See
https://github.com/LinuxCNC/linuxcnc/pull/3509#issuecomment-3044091355 

The  'switch to highlighted' instead of 'switch to checked' tool change would then be in sync with the 'switch to selected offset' in the offset page.

Sigma1912:
I have also prepared adding a buttonbox to the offsetpage:

 
offsetpage

Maybe we could move some of the  'edit offsets', 'Zero G92' , 'Set selected offset' buttons to the new button box.
What do you think?

PeterMue:
> Two more icons ('Add' and 'Delete') in this recent PR #3509

I'll integrate them in https://github.com/PeterMue/gmoccapy-icon-themes and update the https://github.com/hansu/linuxcnc/pull/9 accordingly.

Sigma1912:
Proposed tooltable 'change to highlighted':
Recording

Todos:

- remove automatically checking checkbox of currently loaded tool
- maybe show currently loaded too in the new buttonbox as tool info frame is hidden when using the onboard keyboard.

Comments welcome.

zz912:
So what are check buttons for now?

Sigma1912:
Checkbuttons would only be used to delete the selected lines

Sigma1912:
Deleting is really the only use case where selecting multiple lines makes any sense.

zz912:
I think it would be confusing to have both there. Both "select line" and "checkbuttons".

Select line can multiple lines too.

Sigma1912:
> Select line can multiple lines too.

Yes, but it requires holding down the shift key which is a pain when using a touch screen (without a physical keyboard)

gmoccapy:
Hallo,

IMHO we should delete the column with the checkboxes.
The checkbox column is an historical relict. Long time ago, when I coded that widget, I have not been able to find a solution to edit cells without the checkbox. And also multiple line selection was not possible at that state. Sure some of the problems are caused by my poor python knowledge at that time.

Is it possible to select several lines holding <STRG> down and clicking that line, as we all are used to do in Excel? That would allow deleting several tools at once (I can not remember having done that in all my time working on CNC machines).

The single selected line will be edited, if several lines have been selected, which one will be used for the next step? That is why you get an error if you change tool and have not selected any row or several.

Norvbert

hansu:
Almost all have completely misunderstood this PR :'(
It should have been a discussion about the features I presented, not about any new ones.
It's now hard to follow the topics itself.

So please continue for the several topics that arose from this in the dedicated topics:

Offset page: #3513
Tool table: #3509
Calculator - localization: #3500

hansu:
I would suggest to merge this like it is now and do the localization stuff for the calculator and so on in the next gmoccapy version. What do you think?

zz912:
No problem for me.

gmoccapy:
I do agree to push this one now and discuss the comma dot stuff later.
I vote for the replace solution, so leaving the cell can change any number to dot separation and we allow at the same time the use of commy on the keyboard.
IMHO normally Noone will use the coma as thousand separator

zz912:
Issue for "Calculator widget - localization - decimal separator" is made:
https://github.com/LinuxCNC/linuxcnc/issues/3686

PR #3495, opened 2025-07-03
Gmoccapy: Fix unwanted expansion of ComboBoxText for mouse-button-mode

Sigma1912:
This fixes issue with horizontal expansion of the ComboBoxText reported in #3489

mouse-button-mode


hansu:
That was quick, thanks.
But it was weird that you can delete the text in the ComboBox.  Are you okay with my addition?

Sigma1912:
Yes, I missed that,  that would be weird indeed. 
I cannot test it at the moment but I presume you have so sure go ahead.

PR #3496, opened 2025-07-03
Gmoccapy: Add button to call calculator in mdi and gcode edit modes

Sigma1912:
As presented here https://github.com/LinuxCNC/linuxcnc/pull/3472#issuecomment-2987254101

edit
mdi

Note: There is a problem with color of the calculator icon when using 'material-light' theme. Not sure exactly what the icon color needs to be to work like the other icons. 
light


zz912:
Did you tested it with Localization with decimal comma? CS or DE for example.

hansu:
@Sigma1912  That is cool, thanks!


>Did you tested it with Localization with decimal comma? CS or DE for example.

I did.

PR #3501, opened 2025-07-05
Documentation: Indentation warnings by po4a, spellcheck (smoe:Translations:20250705)

smoe:


BsAtHome:
`100 HZ` (and other places): True that there must be a space. But it should be a non-breaking space. Not sure that can be done portable between html/(t)roff/pdf. How do you make a _portable_ non-breaking space in asciidoc?

The list of frequencies in hm2_rpspi.9.adoc should probably be a table. Maybe just change it into:
```adoc
[cols=">,>,>"]
|===
|2|62500|62.500 MHz
|3|41667|41.667 MHz
|4|31250|31.250 MHz
|5|25000|25.000 MHz
|6|20834|20.833 MHz
|7|17858|17.857 MHz
|8|15625|15.625 MHz
|9|13889|13.889 MHz
|10|12500|12.500 MHz
|11|11364|11.364 MHz
|12|10417|10.417 MHz
|13|9616|9.615 MHz
3+|etc.
|===
```

hansu:
Here we had a change to non-breaking space, but not in asciidoc: fd42d30834

BsAtHome:
> Here we had a change to non-breaking space,...

That is still the case. I just checked toggle.comp (one of the files I recently updated to use adoc). The UTF-8 sequence is 0xc2 0xa0 for code-point u00A0. These are still in the file, so the non-breaking spaces have survived.

So, I guess we need to change all the instances where values are followed by (SI) units to have a non-breaking space there. That'll be a real task.

smoe:
Hello, adjusted the PR along BsAtHome's comments, introduced a few "&#8239;"s.
This patch was about remove warnings by po4a, not so much about the space, admittedly.

BsAtHome:
Have you verified that "32&#9239;MHz" works for all versions of the docs?

I have had problems while using entities being taken literally into the destination format instead of being resolved into their Unicode code point. It may work properly for HTML, but map-pages seem to be affected.

smoe:
We have "&#8239;" at many places throughout the documentation already - it is easy to substitute with something better as soon as we have it. Maybe this is something to be addressed in another issue?

BsAtHome:
Why was this merged with such urgency?

There is that ETB substitute (which apparently means _End of Transmission Block_). That is not the same as a non-breaking space. Using man on the toggle(9) man-page gets the non-breaking space converted to normal space in the terminal, but the manual source page actually still has it embedded. Also, the HTML version of toggle(9) has the non-breaking space still in place (as UTF-8 sequence) and is therefore used.

The table is all messed up:
hm2_rpspi9-bad-man

Additionally, the changes are inconsistent:
hm2_rpspi9-inconsistent

petterreinholdtsen:
It was merged while I had time and opportunity to look at it, to avoid it having to linger for a long time until the next opportunity present itself.  If you have other improvements independent of indentation warnings and these spell check discovered issues, I recommend you create a separate pull request for those improvements.

PR #3507, opened 2025-07-06
Software Weblate - Update of src/po/linuxcnc.pot (smoe:WeblatePo4aNonDocumentation)

smoe:
Weblate is locked.

This PR updates the src/po/linuxcnc.pot file and adjusts the .po files accordingly.

### DIFF
--- a/src/po/ar.po
+++ b/src/po/ar.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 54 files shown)


smoe:
Analogously for src/po/gmoccapy/gmoccapy.pot.

hansu:
I think only the pot-files should be updated. The po-files are updated by Weblate. Otherwise we get a lot aof conflicts with Weblate if I remember correctly.

smoe:
Er - you have a point. Thank you - I only locked the documentation-weblate repository because of mass-changes for Tamil.

PR #3508, opened 2025-07-06
Documentation: Fix missformatted table and typo of non-breakable space

smoe:
This PR is meant to continue where https://github.com/LinuxCNC/linuxcnc/pull/3501 has stopped. The table looks reasonable now, fixed the typo for the non-breakable white space character.


BsAtHome:
The table is both wrong and was committed without testing the change first... see the failing CI.

The table now states two _identical_ values, which is **wrong** and not what the table is supposed to show.

Please revert it to my original 3-column design as stated in https://github.com/LinuxCNC/linuxcnc/pull/3501#issuecomment-3039263023. The context of the table must be considered too.

If you really want a header, then you must use the proper indicators and descriptions. The first value is the divider, the second the configuration value and the third the actual frequency.

PR #3474, opened 2025-06-15
Gmoccapy - update toolinfo after Apply

zz912:
Toolinfo is updated after click on Apply button in tooltable page.
Update_toolinfo_after_Apply

Issue #2818, opened 2023-12-31
Gmoccapy - confirm value in Tool table


zz912:
RIP 2.9 Branche
Debian Bookworm

If I want save changed value in Tool table I must confirm changed value by picking mouse in tooltable or leave cell by Enter or by arrow.

This is a parody of what it looked like when I was learning to use the Tool table.
![Confirm_inserted_value](https://github.com/LinuxCNC/linuxcnc/assets/96618597/227e3dcc-b9bd-461c-bd04-bc015451ffc2)

Could be possible confirm changed value by clicking on some horizontal button?

zz912:
If a person knows about this error, he will be careful about it.

@Sigma1912 is working with tooltable now. I wanted to ask if he could create a signal that would be triggered by the Apply button and would accept the value written in the cell. I tried that, but I couldn't.

hansu:
The question is if we want this. If you accidentally remove a tool or so, you have no abort option.

zz912:
I guess we're misunderstanding each other. When you delete a tool, you don't have an unfinished cell.

This table is similar to:
- Microsoft Excel
- LibreOffice Calc
- Google Sheets
You dont have to confirm anything when you write something in Cell.

hansu:
But you have a save and an undo function there.
The apply button is equal to the save function.

What about this - move the apply button next to the exit button and perform two actions:
1. apply
2. return to main page
?

<img width="992" height="785" alt="Image" src="https://github.com/user-attachments/assets/768e8ea4-350e-40c7-9ade-7d43d5a58dc2" />

Sigma1912:
> You dont have to confirm anything when you write something in Cell.

That is true but in those applications you also have an undo button.

Sigma1912:
> What about this - move the apply button next to the exit button and perform two actions:

This does not change the behavior of unconfirmed edits being rejected.

I've been planning to implement moving the table editing buttons into the table frame as presented here:
https://github.com/LinuxCNC/linuxcnc/pull/3472#issuecomment-2988010896

In my opinion having unfinished edits rejected on leaving the page is not totally unexpected behavior.

hansu:
> > What about this - move the apply button next to the exit button and perform two actions:
> 
> This does not change the behavior of unconfirmed edits being rejected.

Yeah but you can just click on another cell for that input to be accepted. That is not uncommon. And only applies when you disable using the calculator for this.

[...]

PR #3509, opened 2025-07-07
Gmoccapy: Move tool table buttons to table frame, add toggle button for calculator use

Sigma1912:
tooledit classic
tool edit material


Sigma1912:
Auto scroll to the bottom of the tool list when adding new tool:
Recording

hansu:
[Norbert]
> Hallo,
> 
> IMHO we should delete the column with the checkboxes. The checkbox column is an historical relict. Long time ago, when I coded that widget, I have not been able to find a solution to edit cells without the checkbox. And also multiple line selection was not possible at that state. Sure some of the problems are caused by my poor python knowledge at that time.
> 
> Is it possible to select several lines holding down and clicking that line, as we all are used to do in Excel? That would allow deleting several tools at once (I can not remember having done that in all my time working on CNC machines).
> 
> The single selected line will be edited, if several lines have been selected, which one will be used for the next step? That is why you get an error if you change tool and have not selected any row or several.
> 
> Norvbert

hansu:
[Sigma1912]
> @hansu We had a small discussion about the tool table and using 'switch to highlighted' instead of 'switch to checked' tool somewhere. I can't find it, but I was wondering if the 'Select'/'Auswählen' column could be 'Delete'. 

Sounds good.

But if only using that column for deleting - then I don't see in the tool table which tool is actually in the spindle after I changed the selection.

If keeping the checkbox, it makes clear that I am changing from tool 6 to 8:
grafik

Of course I can see the current tool in the Tool information area, but only if that is visible.
Or can we format the line of the current tool in bold?

And I would keep the four buttons left aligned - it looks a bit odd so.

hansu:
And For the save button I would suggest the check button like mentioned in https://github.com/LinuxCNC/linuxcnc/issues/2818#issuecomment-3041057883.
Yes it does a save action but also tells LinuxCNC about the changed tool table, so a check button would be mor intuitive I think.
If we are really fancy, we create a combined button.

Sigma1912:
> But if only using that column for deleting - then I don't see in the tool table which tool is actually in the spindle after I changed the selection.

Note that if the loaded tool and the tool to be changed to are far enough apart in you don't actually see the checkbox anymore.

Current tool loaded would be displayed in the button box:
current tool

Sigma1912:
> And I would keep the four buttons left aligned - it looks a bit odd so.

I want to keep buttons that change the table to the left and buttons that dont to the right. Pulling the ones on left apart gives more room for error when using a touch screen.

hansu:
>Note that if the loaded tool and the tool to be changed to are far enough apart in you don't actually see the checkbox anymore.

That is true

>Current tool loaded would be displayed in the button box:

:+1: 


That's a good solution. But I would move the tool display to the left to be closer to the tool-nr column:

grafik

Sigma1912:
> And For the save button I would suggest the check button like mentioned in https://github.com/LinuxCNC/linuxcnc/issues/2818#issuecomment-3041057883.
Yes it does a save action but also tells LinuxCNC about the changed tool table, so a check button would be mor intuitive I think.

Maybe we should just replace the yellow arrow on the 'back' button with the green check mark? I'm not overly keen on having another button to leave with accepting unfinished entries.
One of the things that have always bugged me was the label 'Apply' which is not the same as 'Save', we might want to indicate if the table has not been saved on leaving the page.

Sigma1912:
Also we need to decide whether we want to keep the checkbox column for deleting multiple tools or remove it as norbert has suggested.

hansu:
> Maybe we should just replace the yellow arrow on the 'back' button with the green check mark? 

Hmm you might want to have an abort option, if you accidentally removed some tools.

>One of the things that have always bugged me was the label 'Apply' which is not the same as 'Save', we might want to indicate if the table has not been saved on leaving the page.

We can add a warning dialog like when exiting from the G-code editor when having changes?

Sigma1912:
> Hmm you might want to have an abort option, if you accidentally removed some tools.

As long as the table has not been saved you can just click on the reload button. Even if you clicked on not saving in the proposed popup on leaving the page,
I think we should add an ESC key handler to abort editing of a cell, seems intuitive to expect that to work.

gmoccapy:
One question after a very very short look at the code changes:
Within tooled it widget the "new button line" is created. So the widget could be used also stand allone. But in the gmoccapy.py file the buttons from the widget are hidden and created new in gmoccapy. IMHO the buttons from the widget should send a signal and gmoccapy should just react to that signal.

Sorry, if I haven't check the code enaugh and beeping wrong here.

Norbert

hansu:
I only wonder why you remove the buttons inside the widget's buttonbox and create new ones. Does it not work to change only the labels to icons?

Sigma1912:
I'll see what I can do to reuse more of the existing code.

Sigma1912:
How about adding a settings option to show/hide 'Select' column given that selecting multiple rows with a touchscreen only is inconvenient. If a keyboard is used the CTRL/SHIFT keys could be used to select multiple rows and the 'Select' column is not required.

hansu:
Sure we can add a setting to hide the select column. But I don't really get the point why - Is it mostly to avoid confusion? Or space reasons?

I agree that this column is now only used in a rare use case - deleting multiple tools by mouse or touch only. 
We can also use a toggle button for deleting. So every selection with an active delete button will remove the line/tool. But might be also a bit unexpected.

--- 
What do you think of highlighting the active tool by a blue-ish color like it is done in the offset page widget?

hansu:
I mentioned resp. asked recently somewhere (didn't find now where it was) if it's the best strategy to change the buttons of the widgets like this one in the gmoccapy code and not in the widget itself.
I think it would be better for maintainability if this is done in the widget's code. Sure, this will affect other GUIs like Gscreen which use these widgets, but maybe they benefit also from these changes. @c-morley what do you think?

c-morley:
If buttons makes sense for general use, (I didn't read back enough to see what buttons you are referring to) then I would think adding them to the widget would be better.
Similarly you could add a function that would be used to add a button to the existing button box, so a third party could add whatever buttons they want.

If you must change all the buttons for some reason, then adding a function or property to hide/delete the button box and (possibly) add a new button box with new specified buttons would seem reasonable.

hansu:
>If buttons makes sense for general use, (I didn't read back enough to see what buttons you are referring to) then I would think adding them to the widget would be better.

It's more about changing the existing buttons to have an icon instead of just a label and change the size and maybe the order.

Sigma1912:
> Sure we can add a setting to hide the select column. But I don't really get the point why - Is it mostly to avoid confusion? Or space reasons?

It's to avoid confusion. The only use case for the Select column is to select multiple tools with a touch screen w/o a keyboard.

c-morley:
> > If buttons makes sense for general use, (I didn't read back enough to see what buttons you are referring to) then I would think adding them to the widget would be better.
> 
> It's more about changing the existing buttons to have an icon instead of just a label and change the size and maybe the order.

Can that not be controlled with css styles?

hansu:
I don't think we can add an icon with CSS...

Sigma1912:
> What do you think of highlighting the active tool by a blue-ish color like it is done in the offset page widget?

This would require a hidden column added to the tool data with the 'foreground' color. At least that is how it is done in the offset table and I don't know of an other way to change the text color. This column would need to be inserted when reading the tool table file and would need to be ignored when writing to the file.

hansu:
@Sigma1912 I am happy to merge your latest version. Can you rebase it to the current master and resolve the conflict?
Furthermore the label of the loaded tool seems to got lost during you latest change. Can you re-add this? (and I would prefer "Tool" instead of "Tool#"). Thanks!

grafik

Sigma1912:
Modified and rebased as requested

tooltable

hansu:
Nice, that was quick. Thanks!

zz912:
Hello @Sigma1912 

I wanted to ask if you are using the buttons from the tooledit widget?
Tooledit

If I understand correctly, there are unused functions left in gmoccapy.py:
    def on_btn_delete_tool_clicked(self, widget, data=None):
    def on_btn_add_tool_clicked(self, widget, data=None):
    def on_btn_reload_tooltable_clicked(self, widget, data=None):
    def on_btn_save_tool_changes_clicked(self, widget, data=None):

I wanted to resolve the conflict in https://github.com/LinuxCNC/linuxcnc/pull/3474 . I managed to resolve the conflict. I changed apply to save. Unfortunately https://github.com/LinuxCNC/linuxcnc/pull/3474 stopped working for me. I found that def on_btn_save_tool_changes_clicked(self, widget, data=None): is not called by anything.

Can you please check if gmoccapy.py contains the unused functions mentioned above?

Would it be possible to disable label "Tool loaded: " when the keyboard is off? It looks strange:
Tooledit-002

Would it be possible to add tooltips for buttons delete, add, reload, save?

hansu:
>I wanted to ask if you are using the buttons from the tooledit widget?

Yes they are used and re-arranged. But the tooltips seem to got lost. Maybe even in an earlier commit.

>If I understand correctly, there are unused functions left in gmoccapy.py:
def on_btn_delete_tool_clicked(self, widget, data=None):
def on_btn_add_tool_clicked(self, widget, data=None):
def on_btn_reload_tooltable_clicked(self, widget, data=None):
def on_btn_save_tool_changes_clicked(self, widget, data=None):


Yes it seems they are unused. E.g. this one is not called. It's currently possible to delete the current tool which this function disallows:


```
    def on_btn_delete_tool_clicked(self, widget, data=None):
        act_tool = self.stat.tool_in_spindle
        if act_tool == self.widgets.tooledit1.get_selected_tool():
            message = _("You are trying to delete the tool mounted in the spindle\n")
            message += _("This is not allowed, please change tool prior to delete it")
            self.dialogs.warning_dialog(self, _("Warning Tool can not be deleted!"), message)
            return

```

Sigma1912:
See https://github.com/LinuxCNC/linuxcnc/pull/3706

Thanks for testing.

hansu:
>Would it be possible to disable label "Tool loaded: " when the keyboard is off? It looks strange:

Imagine this situation (picture).
I think it's better to have it redundant than to not have it in some situations. It doesn't harm IMHO as there is space enough.

grafik

zz912:
I vote for solution without redundand information. What is sigma opinion?

hansu:
The check could be improved to really avoid redundance like "If tool info is shown --> hide tool information in table".
But it's a bit to much effort for just avoiding that redundance.

zz912:
> The check could be improved to really avoid redundance like "If tool info is shown --> hide tool information in table". But it's a bit to much effort for just avoidung that redundance.

I love details. This should be nice detail. Can I ask Sigma for it? Please.

Issue #3510, opened 2025-07-07
Gmoccapy - open_file in gmoccapy.pref cause crash


zz912:
RIP instalation
branch master

If I add this to gmoccapy.pref:
`open_file = /home/zdenek/linuxcnc/linuxcnc-zz912/configs/nc_files/3D_Chips.ngc`
it will crash Gmoccapy

```
[Gmoccapy][ERROR]  Found an error!
The following information may be useful in troubleshooting:

Traceback (most recent call last):
  File "/home/zdenek/linuxcnc/linuxcnc-zz912/bin/gmoccapy", line 6136, in <module>
    inifile = sys.argv[2]
              ~~~~~~~~^^^
IndexError: list index out of range
 (gmoccapy:62)
```

Sigma1912:
Same here.

Debugging like this:

![Image](https://github.com/user-attachments/assets/585988db-a53f-4d1d-98e4-8f8143a3f10b)



With 'open_file = /home/user/git/linuxcnc-master/configs/nc_files/arcspiral.ngc'
```
2-sys.argv:  ['']
```

![Image](https://github.com/user-attachments/assets/2f660867-9fd2-40b6-8f27-a81c02ef8cb3)

Sigma1912:
Storing 'sys.arcv' seems to fix the startup issue and loads the file:

```
    store_sys_argv= sys.argv
    # instantiate gmoccapy
    app = gmoccapy(sys.argv)

    # get the INI path
    inifile = store_sys_argv[2]

    # find the POST_GUI HAL file path
    LOG.debug("inifile = {0} :".format(store_sys_argv[2]))
    postgui_halfile = app.get_ini_info.get_postgui_halfile()
```

However there is another error when trying to open the 'Calibration' tool on the settings page:

![Image](https://github.com/user-attachments/assets/2fbf49d1-acbd-4b05-a29e-55411bbc2217)

Sigma1912:
the second error _could_ also be fixed by using a local copy of argv:

![Image](https://github.com/user-attachments/assets/204feca5-f018-4555-b48f-97079c304920)

![Image](https://github.com/user-attachments/assets/4a9fb043-3cff-4b61-bd8c-a2156ccb38b2)


I haven't really found out why the 'sys.argv' list is suddenly empty when an 'open_file' is set.
So the above 'fixes' seem rather amateurish.

Sigma1912:
Maybe @hansu has a more educated fix?

[...]

PR #3511, opened 2025-07-08
Gmoccapy: Fix args list getting lost

Sigma1912:
Fixes https://github.com/LinuxCNC/linuxcnc/issues/3510

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -49,6 +49,9 @@ from collections import OrderedDict # needed for proper jog button arrangement
 from time import strftime  # needed for the clock in the GUI
 #from Gtk._Gtk import main_quit
 
+# store the command line arguments passed
+args_passed = sys.argv
+
 # Throws up a dialog with debug info when an error is encountered
 def excepthook(exc_type, exc_obj, exc_tb):
     try:
@@ -126,7 +129,7 @@ INFO_ICON = "dialog_information"
 AXISLIST = ['offset', 'X', 'Y', 'Z', 'A', 'B', 'C', 'U', 'V', 'W', 'name']
 
 class gmoccapy(object):
-    def __init__(self, argv):
+    def __init__(self):
 
         # prepare for translation / internationalisation
         locale.setlocale(locale.LC_ALL, '')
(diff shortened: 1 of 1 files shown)


c-morley:
Do you know why this is necessary? I don't see the problem on my system.

I do see this error now that I know how to invoke it. (preload a file)

Sigma1912:
I don't. 
Maybe @BsAtHome knows how his commit   https://github.com/LinuxCNC/linuxcnc/commit/d71c67c37639d20b021dbd271c4ac6e779ded754 is related to https://github.com/LinuxCNC/linuxcnc/issues/3510

BsAtHome:
The question I have is: Is the complete python `sys.argv` array empty? Or are just _some_ values missing? What happens in between start of the program, the first use of `sys.argv` and the error where it seems to have lost index 2?


If _**all**_ sys.argv values are missing, _from the start_, then the following may be an explanation:
I'm not entirely sure why there should be any interference caused by the change. However, there might be a call missing (see https://docs.python.org/3/c-api/init_config.html):
```c++
PyStatus status = PyConfig_SetBytesArgv(&config, argc, argv);
if(PyStatus_Exception(status)) {
 ...handle exception...
}
```
The call should be done before `Py_InitializeFromConfig()`. The above snippet should take the arguments from the command line and propagate them to the config object used. That _might_ solve the problem.

However, I'm not sure how you can get to argc/argv in the PythonPlugin class that is a singleton instantiated from _somewhere_. Instantiation seems performed in `emc/rs274ngc/rs274ngc_pre.cc`, which in turn is in the constructor for the Interp class, sub-classed from InterpBase and sub-classed from boost::noncopyable. The instantiation of Interp seems to be in `emc/task/emctask.cc`, but there might be others that I missed.

Al in all, a mess of opaque layers and access to real argc/argv is lost in the layers. I sincerely hope that the answer to the question(s) I posed at the start of this message already solved the mystery ;-)

Sigma1912:
@BsAtHome please see https://github.com/LinuxCNC/linuxcnc/issues/3510 for a description 
- sys.argv list is empty (ie ['']) after some manipulation in the glade gui.

BsAtHome:
See in depth analysis in https://github.com/LinuxCNC/linuxcnc/issues/3510#issuecomment-3050274242

The solution is not this PR, but more likely requires a fix in the Interp class.

c-morley:
Hacky work around for gremlin:

`

           # FIXME Hack to remember argv settings (canon strips them somehow)
            temp = sys.argv

            result, seq = self.load_preview(filename, canon, unitcode, initcode)

            # FIXME reset argv
            sys.argv = temp

`

BsAtHome:
See https://github.com/LinuxCNC/linuxcnc/issues/3510#issuecomment-3051754954 for suggested fix.

Sigma1912:
I'll close this PR once your fix has been merged (maybe somebody needs a quick fix until then)

PR #3512, opened 2025-07-08
Gmoccapy: Fix arrow key up/down bug in tooleditor

Sigma1912:
The key_press handler inside 'tooledit_widget.py' interferes when using the arrow up/down keys (one keypress results in moving two rows ):

Recording_current

New (fixed) behavior:
Recording

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -1924,7 +1924,9 @@ class gmoccapy(object):
         self.widgets.tooledit1.set_visible("abcxyzuvwijq", False)
         for axis in self.axis_list:
             self.widgets.tooledit1.set_visible("{0}".format(axis), True)
-
+        # disconnect the key_press handler in the widget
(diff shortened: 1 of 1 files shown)


hansu:
But with this fix, this handy function wouldn't work any longer:

tool-table

Sigma1912:
Not sure I like it much.
Notice that this 'handy function' makes it impossible to move the cursor _inside_ the cell left or right as it will automatically jump to cell beside it.

Sigma1912:
Also when you hit enter to finish the edit it automatically goes into editing the cell below.

Sigma1912:
What we could possibly do is retain the 'Tab' key function so we can go on editing the cell to right.

Sigma1912:
Tab key enabled:

Recording

gmoccapy:
I like Sigmas solution with the Tab key ;-)

Sigma1912:
Demo with Tab and Esc keys:
Recording

Sigma1912:
I'll put the Tab and Esc key feature in a separate PR against master as it is not a bug fix.

PR #3514, opened 2025-07-09
Do not initialize python if already initialized

### LINKED ISSUE #3510: Gmoccapy - open_file in gmoccapy.pref cause crash

BsAtHome:
This PR fixes #3510 and replaces #3511.

See also discussion in #3510 for more details. Completely removing the initialization would cause tests to fail because the python plugin code is also used in standalone code where it is required. Now it is only called when python is not yet initialized.
It does make one code-path redundant and a comment is added. In the future, that redundant code-path may be removed. We'll see how this change works out.

### DIFF
--- a/src/emc/pythonplugin/python_plugin.cc
+++ b/src/emc/pythonplugin/python_plugin.cc
@@ -294,13 +294,21 @@ PythonPlugin::PythonPlugin(struct _inittab *inittab) :
     abs_path(0),
     log_level(0)
 {
-  PyConfig config;
-  PyConfig_InitPythonConfig(&config);
-  if (abs_path) {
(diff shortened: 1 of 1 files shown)


BsAtHome:
Superseded by #3516

Issue #3485, opened 2025-06-26
xhc-whb04b-6 not regonising feed rotary switch "lead" position


zajc3w:
This is the output when turning the left, or feed rotary switch from 100% to "lead" position

`pndnt wheel total counts {counts=0 activeCounter=0 isLeadActive=0}
in    0x04 10 00 00 1b 11 00 10 delta 27 => | 04 | 10 |                 |                 |    10(100%) |     X(    ) |   0 | 10
pndnt feed   inactive event feed={code=0x1a text='5' altText='60%'} feedButton={{{key={code=0x1b text='10' altText='100%'}}} isPermitted=TRUE stepSize=100 stepMode=0x00}
pndnt feed   active   event feed={code=0x1b text='10' altText='100%'} feedButton={{{key={code=0x1b text='10' altText='100%'}}} isPermitted=TRUE stepSize=100 stepMode=0x00}
pndnt wheel total counts {counts=0 activeCounter=0 isLeadActive=0}
in    0x04 10 00 00 9b 11 00 10 delta 155 => | 04 | 10 |                 |                 |      (    ) |     X(    ) |   0 | 10
pndnt failed to interpret axis code feed axisCode={�}
pndnt feed   inactive event feed={code=0x1b text='10' altText='100%'} feedButton={{{key={code=0x00 text='' altText=''}}} isPermitted=FALSE stepSize=100 stepMode=0x00}
pndnt feed   active   event feed={code=0x00 text='' altText=''} feedButton={{{key={code=0x00 text='' altText=''}}} isPermitted=FALSE stepSize=100 stepMode=0x00}
pndnt wheel total counts {counts=0 activeCounter=0 isLeadActive=0}
`
it appears that the "lead" key is now 0x9B, not 0x1c
this is with 2025 purchased pendant

label at the back says:
model: WHB04B-6
serial:1703010004
Version:TX:V08 V20  RX:V10 V18 

this was supposedly fixed in 2020, but it's still not working with latest distributed binary, this is on a fresh 2.9.4 bookworm iso
```
cnc@cnc:~$ xhc-whb04b-6 -h
xhc-whb04b-6 version 2.9.4 Jan 25 2025 12:20:01

SYNOPSIS
    xhc-whb04b-6 [-h | --help] | [-H] [OPTIONS] 
.
.
.
```

andypugh:
It seems that it ought to be working with 0x9B in 2.9. 

[...]

PR #3518, opened 2025-07-15
fix for https://github.com/LinuxCNC/linuxcnc/issues/3485

zajc3w:
Old patch #987 mentioned in commits wasn't complete, it missed one check of the key code.
tested with only version of xhc-whb04b-4 and xhc-whb04b-6 I had access to.
I ordered second pendant from another supplier in hope to get differend version, hence the delay.
Unfortunatey both use the 0x9B code, not 0x1C.

tested on debian preempt_rt install of latest stable ISO.

edit: spelling


andypugh:
Thanks

PR #3521, opened 2025-07-19
Fix runtests races

BsAtHome:
Fix the runtests race condition running `checkresult` before the `result` file is closed. This can happen when the test-process runs asynchronously and has not yet exited, keeping the from runtests redirected stdio descriptors in an open state.

The mb2hal process, which exhibited the race, has another race fixed, where the signal handlers were installed too late in the process. Additionally, the mb2hal internal `quit_flag` must be declared volatile because it is written in the signal handler and read in separate threads.

Finally, a longstanding bug in runtests prevented it to take multiple directories from the command-line when, for example, done using a wildcard. The expansion to find tests mistakenly used `$*` instead of `$@`.

### DIFF
--- a/scripts/runtests.in
+++ b/scripts/runtests.in
@@ -56,6 +56,41 @@ clean () {
 	-print0 | xargs -0 rm -f
 }
 
+wait_for_result_close() {
+    # Test for the 'result' and 'stderr' files in the current testdir to be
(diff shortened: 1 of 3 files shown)


andypugh:
Thanks for doing this. As it's a bug-fix I will apply it to 2.9 then merge up.

PR #3525, opened 2025-07-24
Allow for changing default G64 Value

satiowadahc:
Yes, I'm aware you should put it in your safety block. Yes, I'm aware [RS274NGC]Start-Up Code exists. On the other hand consider a shop with multiple machines and someone... not saying names... copies a program from Haas or Fanuc to Linuxcnc. Well I'd like to be able to use the same code cross platform, and not cut corners so to speak.

image

This defaults to the classic behaviour if you don't use it. If you do use it, calling G64 all by itself will give a reasonable answer.


andypugh:
Seems like a good idea to me.

PR #3527, opened 2025-07-25
PNCconf - integration of default values for G64 in *.ini file

zz912:
@satiowadahc made new functionality, so I integrete it to mesa wizard PNCconf 
https://github.com/LinuxCNC/linuxcnc/pull/3525/

g64-default

### DIFF
--- a/src/emc/usr_intf/pncconf/build_INI.py
+++ b/src/emc/usr_intf/pncconf/build_INI.py
@@ -172,8 +172,13 @@ class INI:
             else:
                 unit = 20
                 p =.001
-            print ("RS274NGC_STARTUP_CODE = G{} G40 G90 G94 G97 G64 P{}".format(unit,p), file=file)
-
+            print("RS274NGC_STARTUP_CODE = G{} G40 G90 G94 G97 G64".format(unit), file=file)
+            print("", file=file)
+            print("# Default P value for G64 if P is not called out", file=file)
+            print("G64_DEFAULT_TOLERANCE = {}".format(p), file=file)
+            print("# Default Q value for G64 if Q is not called out", file=file)
+            print("G64_DEFAULT_NAIVETOLERANCE = 0", file=file)
+            print("", file=file)
         if self.d.frontend == _PD._GMOCCAPY:
             print("SUBROUTINE_PATH = ./macros", file=file)
             print("REMAP=M6  modalgroup=6 prolog=change_prolog ngc=change_g43 epilog=change_epilog", file=file)


satiowadahc:
Does G64 in startup become redundant now as there's a default value?

I seen an issue earlier that multiple p words in startup code cause issues. But if g64 now has defaults, i would think startup taking default behavior would be preferred

zz912:
G64 is in startup so everyone knows that G61 is not active.

c-morley:
Should the Q value not be set too?

zz912:
https://linuxcnc.org/docs/devel/html/gcode/g-code.html#gcode:g64

> If you set Q to a non-zero value it turns on the Naive CAM Detector: when there are a series of linear XYZ feed moves at the same [feed rate](https://linuxcnc.org/docs/devel/html/gcode/other-code.html#sec:set-feed-rate) that are less than Q- away from being collinear, they are collapsed into a single linear move.

I'm afraid this function would make flat areas for circles. These areas could be almost unmeasurable, but they would be observable by eye. I would leave this functionality for more experienced LCNC users, not for PNCconf users.

c-morley:
Well As I understand it only if your circle was less then the setting (say .001 inch) - thats a pretty small circle :)
I also understand it can be very helpful for more constant machine speed.

Though I have not tested to qualify either.

zz912:
My theories above were wrong. Tests using millimeter.

Tests with "char of infinity".
[G64_Q.ngc.zip](https://github.com/user-attachments/files/22432714/G64_Q.ngc.zip)
Circle_G64Q

G64 Q14
Circle_G64Q14

G64 Q16
Circle_G64_Q16


Tests with heart.
[Srdicko.ngc.zip](https://github.com/user-attachments/files/22432542/Srdicko.ngc.zip)
G64 Q0.5
G64Q0 5

G64 Q2
G64Q2

G64 Q5
G64Q5


My opinion is that for a beginner it would be best to use G61. A beginner with problems will install Debian and count steps per revolution.
I know a lot of people, including me, who have struggled with G64 without P0.01. G64 Q is not very well known and when they have a bad products they will look for fault everywhere.

andypugh:
I am not clear now whether you think this commit is correct or not.
I have asked Chris to review too.

zz912:
Chris and I discussed whether G64_DEFAULT_NAIVETOLERANCE should be zero or have some value.

I don't prefer that, but it's not a problem to modify this PR so that G64_DEFAULT_NAIVETOLERANCE contains some value for mm and some value for inches. Please define them.

c-morley:
I would suggest using the same value as G64 P

zz912:
I've edited the PR to have G64_DEFAULT_NAIVETOLERANCE have the same value as the P.

To Chris:
Feel free to make another request if needed.

c-morley:
Thank you!

PR #3529, opened 2025-07-28
tests: Add a test for spindle INI speed limits function

andypugh:
Prompted by a discussion on the mailing list.


smoe:
Nothing damatic, just to show some respect to the effort:

This is a rather advanced test, I wish the README had a few more insights in how the test works. For instance there is a spindle.ini (invoked by the test script) and a test.ini. This may be confusing.

Another thing is that the .ini files are not minimal. For instance the unit "inch" should not be required for testing the spindle's RPM. Or maybe they are minimal because of some other constraints that I am not aware of, which would then possibly be discussed in another issue :-)

andypugh:
There should only be one INI file, I later deleted the spurious one.

I copied the INI from another test. I don't know what the minimum set is for an INI, but I would expect you need at least one axis fully defined in a test like this where the full linuxcnc app is being loaded. 

In writing this test I found: https://linuxcnc.org/docs/stable/html/code/writing-tests.html rather less helpful than I thought it should be.  Having now done the test I think that everything I needed to know was there, but I didn't recognise it at the time. For example I had no "checkresult" and spent a long time wondering why, if the python and shell scripts both exited with zero the test still failed.

smoe:
Once we have multiple instances of LinuxCNC cooperating, I anticipate that devices without axes may be a thing.  A LinuxCNC-controlled washing machine  or the centrifuge in our lab comes to mind that can spin, open/close doors, set temeratures, let water flow, sense various bits and pieces. Maybe I just found myself a challenge :-)

andypugh:
I have done quite a few hal-only configs without axes (including a sheet metal bender).
But this test is testing the "motion" module, and I think that expects axes. 
If you are not moving axes, then you can omit the motion module.

PR #3533, opened 2025-08-03
docs: optics while translating (smoe:docs_fixes_20250803)

smoe:
Hello,

I had a first look at the many changes that the update of po4a brought us. It was a huge improvement. The one or other bit however did still not look right, because of parsing (sometimes) or because of some smallish bits like a missed period or line breaks when there should be a continuous text. So I had a look into it and these patches I consider an improvement of what we had before.

At least one paragraph I thought was erroneous, but that I will comment on below.

### DIFF
--- a/docs/src/man/man1/hal_input.1.adoc
+++ b/docs/src/man/man1/hal_input.1.adoc
@@ -39,7 +39,7 @@ A number _N_::
 
 When several devices are identified by the same string, add "**:**_N_"
 where _N_ is the index of the desired device. For example, if *Mouse*
-matches *input3* and *input10*, then *Mouse* and *mouse:0* select *input3*.
+matches *input3* and *input10*, then *Mouse* and *Mouse:0* select *input3*.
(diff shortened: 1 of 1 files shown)


docs/src/man/man1/hal_input.1.adoc
```diff
@@ -39,7 +39,7 @@ A number _N_::
 
 When several devices are identified by the same string, add "**:**_N_"
 where _N_ is the index of the desired device. For example, if *Mouse*
-matches *input3* and *input10*, then *Mouse* and *Mouse:0* select *input3*.
```
smoe, line 42: Should this possibly be all lower-case, i.e. always mouse not Mouse ?

andypugh (approved):
This all looks OK to me. (I still am not 100% sure on the Mouse/mouse thing in HAL_input, but I don't think this makes anything worse)

smoe:
I have undone that [mM]ouse thingy. Should be fine.

PR #3536, opened 2025-08-06
probe_error_decel_no_inhibit - rebased for 2.9

DauntlessAq:
This is the bugfix for the probe deceleration error, rebased for the 2.9 branch.

The current state of this pull request is based on discussion from my previous pull request into master:
https://github.com/LinuxCNC/linuxcnc/pull/3534

The previous pull request kept the original behaviour (with the deceleration error) by default, but also included the ability to inhibit the error.

However, in this pull request the error is effectively permanently inhibited (so, fully patched) with no option to enable it, due to the determination that this error is definitely a bug which should be fully patched).


DauntlessAq:
(accidentally opened request against master - closed and resubmitting against 2.9)

PR #3537, opened 2025-08-06
probe_error_decel_no_inhibit rebased for 2.9

DauntlessAq:
This is the bugfix for the probe deceleration error, rebased for the 2.9 branch.

The current state of this pull request is based on discussion from my previous pull request into master:
https://github.com/LinuxCNC/linuxcnc/pull/3534

The previous pull request kept the original behaviour (with the deceleration error) by default, but also included the ability to inhibit the error.

However, in this pull request the error is effectively permanently inhibited (so, fully patched) with no option to enable it, due to the determination that this error is definitely a bug which should be fully patched).

### DIFF
--- a/src/emc/motion/control.c
+++ b/src/emc/motion/control.c
@@ -716,59 +716,63 @@ static void process_probe_inputs(void)
     } else if (!old_probeVal && emcmotStatus->probeVal) {
         // not probing, but we have a rising edge on the probe.
         // this could be expensive if we don't stop.
-        int i;
-        int aborted = 0;
 
         if(!GET_MOTION_INPOS_FLAG() && tpQueueDepth(&emcmotInternal->coord_tp)) {
             // running an command
-            tpAbort(&emcmotInternal->coord_tp);
-            reportError(_("Probe tripped during non-probe move."));
-	    SET_MOTION_ERROR_FLAG(1);
-        }
-
-        for(i=0; i<NO_OF_KINS_JOINTS; i++) {
-            emcmot_joint_t *joint = &joints[i];
-
-            if (!GET_JOINT_ACTIVE_FLAG(joint)) {
-                /* if joint is not active, skip it */
-                continue;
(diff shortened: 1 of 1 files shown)


andypugh:
I want to run this on my mill (where I see the problem) before commiting. It might take me a few days to find the time.

DauntlessAq:
That's great Andy, thanks so much!

andypugh:
I found the time to find that my mill is broken. I spent this evening on it and it's close to working now. 
I won't be releasing 2.9.5 until this is in, so it's not going to miss any deadlines.

andypugh:
I finally got my mill back up and running, and tested this. 
The existing 2.9.4 installation errored out on my testing about half the time, whereas with the new logic it did 10 consecutive tests without any issue.

andypugh:
Thanks for getting deep into this. It had been causing me trouble for some time, and I was thinking it was something wrong in my setup that I didn't have the time to fix. Finding that it was an inherent problem in the logic is an important improvement.

DauntlessAq:
Excellent, thanks so much! I'm really glad this got into 2.9.5!
Hoping this will save a lot of frustration for people setting up probing.

It's still possible to get bounce on the pulloff move if it's not a probing move, but it's less likely and the error message when this happens is accurate (the probe has tripped during a non-probe move), so most people will find that the bounce issue is no longer a problem that requires hal logic to fix.

But, if you make the pulloff move a probing move as well (G38.4 or G38.5), it will also be bounce-tolerant during deceleration.
Since I'm assuming the majority of users probe using prewritten macros, if these macros are modified to take advantage of this change, it should be possible to effectively completely eliminate the bounce issue for these users, making it a *completely* solved issue for the majority.

Also, https://github.com/LinuxCNC/linuxcnc/pull/3534 (the original pull request against master which is dated compared the logic eventually merged into 2.9) is still open, shall I close that now?

PR #3544, opened 2025-08-13
Add Joint to Axis Translator

satiowadahc:
Open to comments. 

Motivation is some guys can't memorize Joint 0 - X Joint 1 is Y etc. This is a brute force way off adding a popup for which Axis has faulted, while being flexible enough for a variety of kinematics.

The other options I seen for doing this was in one of the 3 modules:
- Motion only has the fault data
- Kinematics only has the axis-to-joint mapping
- RS274 only has the INI data
So without adding spaghetti of plumbing those three modules I opted for a halcomp

Also naming things is hard... is this too generic of a name.?

### DIFF
--- a/src/hal/components/joint_axis_mapper.comp
+++ b/src/hal/components/joint_axis_mapper.comp
@@ -1,4 +1,4 @@
-component translate """Translate from Joint to Axis""";
+component joint_axis_mapper """Translate from Joint to Axis""";
 
 license "GPL";
 
@@ -28,7 +28,7 @@ char fault_letter[16];
 
 int rtapi_app_main(void) {
     int r = 0;
-    comp_id = hal_init("translate");
+    comp_id = hal_init("joint_axis_mapper");
     char buf[HAL_NAME_LEN + 1];
 
     if(comp_id < 0) return comp_id;
@@ -36,7 +36,7 @@ int rtapi_app_main(void) {
     number_joints = strlen(coord);
     faults_in = hal_malloc(sizeof(faults_in_t));
 
-    rtapi_print_msg(RTAPI_MSG_INFO, "Translate - Initializing with coords %s\n", coord);
+    rtapi_print_msg(RTAPI_MSG_INFO, "Joint Axis Mapper - Initializing with coords %s\n", coord);
 
     int this_count = 0;
     for(int i =0; i < strlen(coord); i++){
@@ -65,19 +65,19 @@ int rtapi_app_main(void) {
         if(coord[i] == 'W' || coord[i] == 'w'){ axis_counts[8]++; this_count = axis_counts[8];  }
 
         if(this_count == 1){
-            hal_pin_bit_newf(HAL_IN,  &(faults_in->fault_in[i]),  comp_id, "translate.%c-fault", coord[i]);
-            hal_pin_bit_newf(HAL_OUT, &(faults_in->fault_out[i]), comp_id, "translate.%c-fault-latched", coord[i]);
+            hal_pin_bit_newf(HAL_IN,  &(faults_in->fault_in[i]),  comp_id, "jam.%c-fault", coord[i]);
+            hal_pin_bit_newf(HAL_OUT, &(faults_in->fault_out[i]), comp_id, "jam.%c-fault-latched", coord[i]);
         }
         else if(this_count > 1){
-            hal_pin_bit_newf(HAL_IN, &(faults_in->fault_in[i]), comp_id, "translate.%c%d-fault", coord[i], this_count);
-            hal_pin_bit_newf(HAL_OUT, &(faults_in->fault_out[i]), comp_id, "translate.%c%d-fault-latched", coord[i], this_count);
+            hal_pin_bit_newf(HAL_IN,  &(faults_in->fault_in[i]),  comp_id, "jam.%c%d-fault", coord[i], this_count);
+            hal_pin_bit_newf(HAL_OUT, &(faults_in->fault_out[i]), comp_id, "jam.%c%d-fault-latched", coord[i], this_count);
         }
         else {
-            rtapi_print_msg(RTAPI_MSG_ERR, "Translate - %c Axis not found in coord string\n", coord[i]);
+            rtapi_print_msg(RTAPI_MSG_ERR, "Joint Axis Mapper - %c axis not found in coord string\n", coord[i]);
         }
     }
 
-    hal_export_funct("translate", (void(*))_, 0, 1, 0, comp_id);
+    hal_export_funct("joint_axis_mapper", (void(*))_, 0, 1, 0, comp_id);
     hal_ready(comp_id);
     return 0;
 }
@@ -90,7 +90,7 @@ void rtapi_app_exit(void) {
 FUNCTION(_) {
         for(int i = 0; i < number_joints; i++) {
             if(*faults_in->fault_in[i] && ! *faults_in->fault_out[i]){
-                rtapi_print_msg(RTAPI_MSG_ERR, "Translate - %c Axis fault detected\n", fault_letter[i]);
+                rtapi_print_msg(RTAPI_MSG_ERR, "%c Axis fault detected\n", fault_letter[i]);
             }
             *(faults_in->fault_out[i]) = *(faults_in->fault_in[i]);
         }


andypugh:
I guess that this would be for situations where the system builder is not the system user? 
I think I would have called it joint-axis-mapping or joint_to_axis (but hyphens/underscores in comp names is confusing...) 
I have changed "translater" to "translator" in the description because Debian flags spelling errors in the changelog,

satiowadahc:
Exactly. "Joint 0 fault" never seems to make sense to our operators, this just adds a message that maps it to Joint X Fault. And assuming someone wants to use this with non trivial kinematics it allows for weird mapping. I can change it to either - or _, is there a preference?

andypugh:
> I can change it to either - or _, is there a preference?

File names have to use underscores but the (not well followed) policy is that HAL pin names should use hyphens. 

This leads to the unfortunate position where "loadrt name_with_underscores" will create HAL pins with names such as "name-with-underscores.00.number-of-hyphens" 

https://linuxcnc.org/docs/html/hal/general-ref.html#sec:hal-gr:naming-conventions

satiowadahc:
```bash
halcmd: loadrt joint_axis_mapper coord=XXYZ
Note: Using POSIX realtime
halcmd: show pin
Component Pins:
     4  bit   IN          FALSE  jam.X-fault
     4  bit   OUT         FALSE  jam.X-fault-latched
     4  bit   IN          FALSE  jam.X2-fault
     4  bit   OUT         FALSE  jam.X2-fault-latched
     4  bit   IN          FALSE  jam.Y-fault
     4  bit   OUT         FALSE  jam.Y-fault-latched
     4  bit   IN          FALSE  jam.Z-fault
     4  bit   OUT         FALSE  jam.Z-fault-latched
     4  s32   OUT             0  joint_axis_mapper.time
```

zz912:
Documentation is missing. Automatic compiling in same "make file" is missing too. I'm writing this here so it won't be forgotten. The component looks interesting, but I haven't figured out what it's for. Is it a replacement for kinematics, or a supplement to kinematics?

satiowadahc:
See the first comment. It's just to add a fault message when Joint 0 faults - which axis is that? Most of our machinist don't understand Joints but every single machinist knows what the XYZABC axis are. 

Arent .comp files auto compiled? If not why not? Sounds like a lot of redundant make files?

zz912:
> See the first comment.

I read it, but I didn't understand it. Is this component also usable for trivkins or just for some special kinematics? Do you have any example of its use?

Nobody will look for documentation for the HAL component in a Pull Request. Everyone will look for it here:
http://linuxcnc.org/docs/devel/html/
man_pages


> Arent .comp files auto compiled? If not why not? Sounds like a lot of redundant make files?

I will test it tonight.

zz912:
.comp files are auto compiled. I apologize for confusing. I tested it.

I don't know if it makes sense to deal with this, but there are a limited number of axes.
```
zdenek@cnc:~/linuxcnc/linuxcnc-zz912$ halcmd -f
halcmd: loadrt joint_axis_mapper coord=XXYZZAABBCCUUVVWW
Note: Using POSIX realtime
HAL: ERROR: pin_new(jam.W2-fault) called with already-initialized memory
```

zz912:
I originally thought it is a component for creating special kinematics. Today I loaded this component into LinuxCNC.
I found out that it is a better joint.N.amp-fault-in and joint.N.faulted

Your component writes messages and creates pins according to coord.
I don't think it's a good idea. I have drivers connected to a mesa card and there are stepgens 0 to N. Physical drivers should be named by joints not by axis.

I think this component should not be part of LinuxCNC because it is a duplication of joint.N.amp-fault-in and joint.N.faulted.

It would be great if you improved joint.N.amp-fault-in and joint.N.faulted. You would add an error message to them:
joint N axis M fault detected
For example:
joint 7 axis Z3 fault detected

To Andy:
I think that HAL components without man pages should not be merged. Otherwise I would have dealt with it earlier.

jam

andypugh:
HAL components defined by a .comp create their own manpages. Unfortunately buildbot builds are failing at the moment so they new manpage hasn't made it to the docs server. 

This component exists as an option for a system integrator to give extra info to the machine operator in cases where these are two different people.  It's entirely optional, as are most of the HAL components.

zz912:
Where is source for man page of this component?

andypugh:
> Where is source for man page of this component?

Well, it won't have much of a manpage, just the two strings. 

It would be better if the original commit included a "description" field.

satiowadahc:
Happy? #3550

Sigma1912:
Seems that this produces a bunch of compiler warnings during the build process:

```
Preprocessing joint_axis_mapper.comp
Compiling realtime objects/hal/components/joint_axis_mapper.c
hal/components/joint_axis_mapper.comp: In function ‘rtapi_app_main’:
hal/components/joint_axis_mapper.comp:57:21: warning: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
   57 |     for(int i =0; i < strlen(coord); i++){
      |                     ^
hal/components/joint_axis_mapper.comp:69:22: warning: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
   69 |     for(int i = 0; i < strlen(coord); i++) {
      |                      ^
hal/components/joint_axis_mapper.comp:47:10: warning: unused variable ‘buf’ [-Wunused-variable]
   47 |     char buf[HAL_NAME_LEN + 1];
      |          ^~~
hal/components/joint_axis_mapper.comp:45:9: warning: unused variable ‘r’ [-Wunused-variable]
   45 |     int r = 0;
      |         ^
hal/components/joint_axis_mapper.comp: In function ‘_’:
objects/hal/components/joint_axis_mapper.c:60:62: warning: unused parameter ‘__comp_inst’ [-Wunused-parameter]
   60 | #define FUNCTION(name) static void name(struct __comp_state *__comp_inst, long period)
      |                                         ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
hal/components/joint_axis_mapper.comp:105:1: note: in expansion of macro ‘FUNCTION’
  105 | FUNCTION(_) {
      | ^~~~~~~~
objects/hal/components/joint_axis_mapper.c:60:80: warning: unused parameter ‘period’ [-Wunused-parameter]
   60 | #define FUNCTION(name) static void name(struct __comp_state *__comp_inst, long period)
      |                                                                           ~~~~~^~~~~~
hal/components/joint_axis_mapper.comp:105:1: note: in expansion of macro ‘FUNCTION’
  105 | FUNCTION(_) {
      | ^~~~~~~~
objects/hal/components/joint_axis_mapper.c: At top level:
objects/hal/components/joint_axis_mapper.c:40:12: warning: ‘export’ defined but not used [-Wunused-function]
   40 | static int export(char *prefix, long extra_arg) {
      |            ^~~~~~
Linking ../rtlib/joint_axis_mapper.so
```

Sigma1912:
Also note that the Usage description does not match the pin name and it might be more intuitive to use 'joint.n.faulted':

```
net x-fault joint.0.amp-fault-in => translate.X-fault
```
should be 
```
net x-fault joint.0.faulted => jam.X-fault
```

Just noticed that the axis letter in the created pins change from upper to lower case for 'coord=xyz' which is a bit unusual.

Sigma1912:
I'm sorry to be going on about this but I cannot seem to get it to work.

I'm testing on master using 'axis.ini' with this modification.

```
[HAL]
HALCMD = loadrt joint_axis_mapper coord=[TRAJ]COORDINATES
```
 Note here that the component will error on load if there are spaces between the axis letters like in many of the simulation configs:
```
[TRAJ]
COORDINATES = X Y Z
```


After homing I toggle 'jam.x-fault' to TRUE.
Not sure but I gather that there should be an error popup with "X Axis fault detected"?
I see nothing.

zz912:
Popup worked me in gmoccapy.

Sigma1912:
Disregard. Of course I forgot the 'addf joint_axis_mapper servo-thread'

Sigma1912:
I got rid of most of the compiler warnings:

```
diff --git a/src/hal/components/joint_axis_mapper.comp b/src/hal/components/joint_axis_mapper.comp
index 3637e75b10..6013631ea3 100644
--- a/src/hal/components/joint_axis_mapper.comp
+++ b/src/hal/components/joint_axis_mapper.comp
@@ -5,6 +5,7 @@ license "GPL";
 pin in bit temp "Fault in";
 
 function _;
+option period no;
 option singleton yes;
 option rtapi_app no;
 
@@ -27,9 +28,7 @@ char fault_letter[16];
 
 
 int rtapi_app_main(void) {
-    int r = 0;
     comp_id = hal_init("joint_axis_mapper");
-    char buf[HAL_NAME_LEN + 1];
 
     if(comp_id < 0) return comp_id;
 
@@ -39,7 +38,7 @@ int rtapi_app_main(void) {
     rtapi_print_msg(RTAPI_MSG_INFO, "Joint Axis Mapper - Initializing with coords %s\n", coord);
 
     int this_count = 0;
-    for(int i =0; i < strlen(coord); i++){
+    for(unsigned i =0; i < strlen(coord); i++){
         if(coord[i] != ' '  && coord[i] != ';'  && coord[i] != '{'  && coord[i] != '}'  && coord[i] != ',' &&
            coord[i] != '\n' && coord[i] != '\r' && coord[i] != '\t' && coord[i] != '\"' && coord[i] != '\'') {
             coord[this_count++] = coord[i];
@@ -51,7 +50,7 @@ int rtapi_app_main(void) {
     this_count = 0;
     number_joints = strlen(coord);
 
-    for(int i = 0; i < strlen(coord); i++) {
+    for(unsigned i = 0; i < strlen(coord); i++) {
         fault_letter[i] = coord[i];
         this_count = 0;
         if(coord[i] == 'X' || coord[i] == 'x'){ axis_counts[0]++; this_count = axis_counts[0];  }
@@ -77,7 +76,7 @@ int rtapi_app_main(void) {
         }
     }
 
-    hal_export_funct("joint_axis_mapper", (void(*))_, 0, 1, 0, comp_id);
+    export("joint_axis_mapper", 0);
     hal_ready(comp_id);
     return 0;
 }
@@ -94,4 +93,5 @@ FUNCTION(_) {
             }
             *(faults_in->fault_out[i]) = *(faults_in->fault_in[i]);
         }

```

only one left is this one:
warning


Any ideas?

andypugh:
__comp_inst is the (slightly hidden) macro that points to the component instance data. 

It might be related to option_singleton being set to "yes". It would be interesting to see if the same warning is seen compiling other "singleton" comps. If so then we might need to add a compiler macro to halcompile.g

Sigma1912:
I get no warnings when compiling 'thc.comp' or 'plasmac.comp' which both have 'option singleton yes;':
```
user@user-iMac:~/git/mylinuxcnc-repo-master$ halcompile --install src/hal/components/thc.comp 
Compiling realtime thc.c
Linking thc.so
cp thc.so /home/user/git/mylinuxcnc-repo-master/rtlib/
user@user-iMac:~/git/mylinuxcnc-repo-master$ halcompile --install src/hal/components/plasmac.comp 
Compiling realtime plasmac.c
Linking plasmac.so
cp plasmac.so /home/user/git/mylinuxcnc-repo-master/rtlib/

```

Sigma1912:
Note though that the warning was resolved by adding this to the FUNCTION:

```
    dummy=0; // if we don't use at least one halpin in FUNCTION we get a compiler warning
```
see related PR #3570

PR #3547, opened 2025-08-22
DRAFT: Converting G-code graphics to use shaders

satiowadahc:
Working on a minimum viable product. The old GL pipeline has been deprecated for some time and now sometimes when gcodegraphics.redraw() gets called I get a black triangle across the display. 

The "Modern" pipeline (Circa 2004) using shaders seems to be a valid option to upgrade to. Needs a little bit of polishing yet, but working to get the feature set that was existing first. 

This also allows for moving into constructive solid geometry which could render stock and then subtract tool paths, rendered on the GPU. 

Going to keep working on this but was wanting feedback before going to far.

Some Philosophy here:
- The Canon class only calculates positions now.
- Trying to move all gl related calls into CanonShaders Class 
- Mixin is now integrated in canon and no need for gtk and qt to both make their own
- I think an options object is needed.
  - Scroll directions
  - Mouse Button options
  - Color options


c-morley:
This is looong overdue, Thank you for taking this on.
I tried this on my system and got a display.

print out in terminal:

python3 canonshaders.py /home/chris/emc/configs/sim/qtdragon/qtdragon_xyz
No canon set, cannot render.
No canon set, cannot draw bounding box.
No canon set, cannot draw machine limits.
Loading /home/chris/linuxcnc/nc_files/examples/3D_Chips.ngc with parameters from /home/chris/emc/configs/sim/qtdragon/qtdragon_xyz/qtdragon.var

shader

a few immediate observations:
blue bounding box seemed miss drawn on one side
CPU usage was at  25 % on the window
rotation with the mouse was a bit strange

andypugh:
Thanks for doing this. I admit that I have never looked at the preview code (except in passing to fix an arc bug) and hadn't realised that it was using deprecated calls. 

There may be some scope for confusion in the naming as "canon" as that also refers to the "canonical" motion commands in emccanon. (But this comment is in the context of not looking at the code as I am camping in a field at the moment)

satiowadahc:
On second thought have the update loop internal to the graphics code would mean you can't customize performance per machine. That should reduce the cpu usage. 

I personally don't like the current controls in gcodes graphics in particular right click zoom. 

I'm open to name suggestions, CanonShade is a mixin from Canonmodule.cc in rs274. CanonShaders... could make it like something to do with shaders/opengl/and drawing.

I need to update my dev machine as the C++ upgrade made this challenging, did most of this on an actual mill. So I'll probably do that today and keep working on improving this

c-morley:
I should have been specific, the rotation with the mouse only rotated in one direction rather then back and forth with the mouse pad.

I way more generic way to select controls would be a nice touch. Maybe we could get gestures one day too.

satiowadahc:
- Fixed Initial Startup Errors
- Removed event loop in favor of update as needed
- Added Option class for each UI's own customization (Defaulted to my preferred controls lol )
- Renamed classes to be a little more descriptive. Yes it could be confusing with canonical motion commands, but its pulling data from there hence CanonInfo.

andypugh:
I assume this is still a draft?

Sigma1912:
Maybe have a look at QTpyvcp's vtk based backplot:
https://github.com/search?q=repo%3Akcjengr%2Fqtpyvcp%20import%20vtk&type=code

satiowadahc:
Welp. I no longer have a job that allows me work time to develop LCNC anymore. I can take this down for the meantime. I might be able to get back to it when the snow falls. 

VTK looks interesting... I see lots of references to it being high memory intensive which large cam files take a ton of memory already. 

If someone is going to take on this torch.. I think gcode.parse_file needs to be split into a parse n lines of code so visually you can render a bit at a time and not lock the display up rendering millions of lines of code. 

https://github.com/LinuxCNC/linuxcnc/blob/f7405d5bd7153aa820f2aa444a43de88dc3257eb/src/emc/rs274ngc/gcodemodule.cc#L855

I know I didn't contribute much to the project, but its still been a fun 6 years.

BsAtHome:
What is the status of this PR?

Sigma1912:
This can be closed as solved by #4293

PR #3550, opened 2025-08-26
Adding documentation for Hal Component

satiowadahc:
Hopefully this helps someone sleep at night. Didn't think it was a complicated concept.


zz912:
I have a young wife who always puts me to sleep well, so I don't have any problems sleeping.

I'm not an LCNC developer, so you can ignore my opinions. I really appreciate the developers here accepting my opinions. I don't mind if they disagree with me, but they can express their disagreement politely.

Component and2 is non-complicated concept, but it has good manual page.
https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/components/and2.comp
https://linuxcnc.org/docs/devel/html/man/man9/and2.9.html
You can inspire here.


`net x-fault joint.0.amp-fault-in => translate.X-fault`

I don't think this will work. The net command needs a signal name. You can't connect two input pins.

It is customary to denote the order of joints with the letter N.
It is customary to denote the order of axis with the letter L.
https://linuxcnc.org/docs/devel/html/man/man9/axis.9.html

andypugh:
> `net x-fault joint.0.amp-fault-in => translate.X-fault`
> 
> I don't think this will work. The net command needs a signal name. You can't connect two input pins.

I think that in this case it is assumed that the signal "x-fault" is being driven by logic elsewhere in the HAL file. 
The statement above will drive both input pins from the signal, and is valid (but will only work if the signal has a "writer" elsewhere)

PR #3555, opened 2025-09-07
deb: lintian warnings and errors (smoe:lintian_warning)

smoe:
nodocs -> nodoc to avoid lintian error.
Also eliminated terminal white space in d/changelog and updated copyright to 2025.


smoe:
@andypugh , I ported this identification of dependencies that are only contributing to the documentation back to the stable branch, in anticipation that we will soon also update the Debian package and I felt like it belongs there. The netpbm library I presume to be dragged in by another dependency that is now documentation only.
 
The master branch has this still written as "nodocs", which is wrong (me in https://github.com/LinuxCNC/linuxcnc/commit/e6b72db3e1b9614a692b672f02d2efb349a08ee3 and you pointed out that lintian was complaining agout it in pedantic mode on the LinuxCNC build servers) . This PR originally meant to correct just that master branch, but "nodoc" (no s) is corrrect (https://www.debian.org/doc/debian-policy/ch-source.html). This will then induce a merge conflict but I am convinced that the stable release would benefit from it.

smoe:
Hm. Why does this not auto-build?

andypugh:
I think that there might be rather more to this. There are several places in the docs where "nodocs" is described as the switch. 

Also, I think that the buildbots invoke "nodocs" and that is possibly what Lintian is complaining about?  

I _think_ that this is why buildbot builds are currently failing,  which means that document updates to wlo from buildbot2 are not happening.

smoe:
https://wiki.debian.org/BuildProfileSpec tells what it should be. To my understanding there is no direct mapping between that environment variable DEB_BUILD_OPTIONS that one then checks in debian/rules and those switches.

I'll skim through the docs and also adjust debian/rules for nodocs/nodoc and nochecks/nocheck.

andypugh:
~Well, I don't know exactly what is going on, but I can't see anything else preventing docs updates from the buildbot to wlo at the moment.~
~(Can we move this to Github? Maybe, I don't know)~

~http://buildbot2.highlab.com/buildbot/#/builders/8/builds/2825~

Actually, probably nothing to do with this. I am now thinking that it is some other issue, looking at the filtered list of error lines. 

http://buildbot2.highlab.com/buildbot/#/builders/8/builds/2825/steps/16/logs/warnings__20_

smoe:
I looked at
```
E: linuxcnc source: invalid-profile-name-in-source-relation nodocs [Build-Depends: docbook-xsl <!nodocs>]
N:
E: linuxcnc source: invalid-profile-name-in-source-relation nodocs [Build-Depends: ghostscript <!nodocs>]
N:
E: linuxcnc source: invalid-profile-name-in-source-relation nodocs [Build-Depends: groff-base <!nodocs>]
N:
E: linuxcnc source: invalid-profile-name-in-source-relation nodocs [Build-Depends: imagemagick <!nodocs>]
N:
E: linuxcnc source: invalid-profile-name-in-source-relation nodocs [Build-Depends: po4a <!nodocs>]
```
which triggered my motivation to get this into the 2.9 branch. With the invocation "--pedantic" this is then the reason for lintian to fail.

I very much hope and expect that this build failure will go away once the buildd updates its source tree.

smoe:
> I very much hope and expect that this build failure will go away once the buildd updates its source tree.

I had this against the stable branch, which explains why the build bot does not see it.

andypugh:
It looks like I accidentally made my docs change in Master (it was meant to be in Stable).

smoe:
I am a bit lost :-) Do you have a pointer for a build-bot run on the now corrected profile? The pointer of yours was to a version that still reads "nodocs", which should be "nodoc" once it sees the PR.

andypugh:
I am a bit lost too. But http://buildbot2.highlab.com/buildbot/#/builders/8/builds/2826 just completed on Stable, and the changes I made to the docs have magically appeared in https://linuxcnc.org/docs/stable/html/getting-started/getting-linuxcnc.html#cha:Installing-RTAI

PR #3561, opened 2025-09-11
docs: fix typos in man pages

luzpaz:
Found via codespell


hansu:
it seems that the HTML man page is a bit erroneous converted so that the changed line among others isn't displayed at all.
It starts with the code section in "MBCCS FILE FORMAT":

https://linuxcnc.org/docs/devel/html/man/man1/mesambccc.1.html

smoe:
@hansu - The patch however is perfectly fine, right? I would want to accept it.

smoe:
Referenced the problem @hansu identified in separate issue #3562.

PR #3567, opened 2025-09-18
correct bad fix for 2753

Sigma1912:
This PR

- reverts https://github.com/LinuxCNC/linuxcnc/commit/736a3a904aaa70532a7ee83ef1c196853ab7ba42
- applies a better fix for #2753


andypugh:
I assume you have tested this against both bugs?

andypugh:
I have checked both bugs in a sim, and it looks good. 

Thanks.

Issue #3563, opened 2025-09-15
2.9.5 axis - touch off with "end" key pressed: not possible to input any numbers then 0 - O.K. is greyed out


dd3t:
2.9.5 axis - touch off with "end" key pressed: not possible to input any numbers then 0 - O.K. is greyed out

<img width="1022" height="647" alt="Image" src="https://github.com/user-attachments/assets/941b3cb1-8236-4e83-98d3-2b06fd98355c" />

https://forum.linuxcnc.org/21-axis/57208-touch-off-ok-button-greyed-out

andypugh:
Also reported here: https://forum.linuxcnc.org/21-axis/56723-touch-off-don-t-work-file-not-open
(But with 2.10)

andypugh:
Correction.
Git bisect says that 736a3a904aaa70532a7ee83ef1c196853ab7ba42 is the bad commit.

andypugh:
I have done some digging, but remain confused. 

The touch-off entry box has a callback to check_valid that is called whenever the entry changes:
https://github.com/LinuxCNC/linuxcnc/blob/2.9/src/emc/usr_intf/axis/scripts/axis.py#L1688
This, in turn calls parse_gcode_expression:
https://github.com/LinuxCNC/linuxcnc/blob/2.9/src/emc/usr_intf/axis/scripts/axis.py#L1580
Which attempts to load a parameter file. 
But I am not sure why, as check_valid checks for the # character and rejects it saying that variables can not be used. 
I suspect that this "File not open" error is set here:
https://github.com/LinuxCNC/linuxcnc/blob/2.9/src/emc/rs274ngc/rs274ngc_pre.cc#L1602
As described:
https://github.com/LinuxCNC/linuxcnc/blob/2.9/src/emc/rs274ngc/rs274ngc_pre.cc#L1485

I think that we may just have been getting away with this, as the behaviour removed in #2753 was setting one of the pointers referenced non-null. 

But in the case of checking a G-code expression, that's perfectly fine, and we can probably just add in special handling for error 4, 
INTERP_FILE_NOT_OPEN that treats it as not-an-error

andypugh:
Changing line 1590 in axis.py to
```py
    # The "4" should be INTERP_FILE_NOT_OPEN 
    if result > 4: return False, gcode.strerror(result)
```
Seems to bypass the problem. 
But I am not sure if this is a proper fix for a not-really-a-problem or a kludge.

[...]

PR #3568, opened 2025-09-18
AxisGui: Allow parameters in 'Touch OFF' dialog

Sigma1912:
This is a spin off from https://github.com/LinuxCNC/linuxcnc/issues/3563

### DIFF
--- a/src/emc/usr_intf/axis/scripts/axis.py
+++ b/src/emc/usr_intf/axis/scripts/axis.py
@@ -1690,10 +1690,6 @@ class _prompt_float:
         st = 0
         ok = 1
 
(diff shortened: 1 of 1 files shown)


andypugh:
To add some explanation. In trying to explain why just typing a number in a box should give a "file not found" error we determined that it was due to the Axis touch of Entry widget using the G-code parser to check input (and calculate numeric results)
It was assumed that this was at leas in-part to allow variables to be used, but variable use was specifically tested for and denied. 
Removing the code that prohibited it seems to show that it works perfectly OK (but may not have done 18 years ago)

PR #3570, opened 2025-09-19
Update joint_axis_mapper.comp

Sigma1912:
- fixes all compiler warnings
- fixes usage example
- expands component description

see https://github.com/LinuxCNC/linuxcnc/pull/3544

### DIFF
--- a/src/hal/components/joint_axis_mapper.comp
+++ b/src/hal/components/joint_axis_mapper.comp
@@ -1,25 +1,27 @@
 component joint_axis_mapper "Translate faults from Joint to Axis";
 
 description """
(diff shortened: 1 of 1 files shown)


satiowadahc:
Thanks For fixing those.

Curious which os/compiler are you using? I don't see the same errors on Ubuntu 24 or build-bot?

Sigma1912:
> I don't see the same errors on Ubuntu 24 or build-bot?

Interesting, my test PC runs LinuxMint Zara
'$ gcc -v' returns
```
gcc version 13.3.0 (Ubuntu 13.3.0-6ubuntu2~24.04) 
```

PR #3586, opened 2025-10-20
gmoccapy: allow use of TAB key in the tooledit widget

Sigma1912:
as promised here:
https://github.com/LinuxCNC/linuxcnc/pull/3512


hansu:
Can you check if it doesn't conflict with https://github.com/LinuxCNC/linuxcnc/pull/3494 ?

Sigma1912:
Should not interfere now, note though that this PR requires the fix in 2.9 mentioned above:
https://github.com/LinuxCNC/linuxcnc/pull/3512/files

PR #3587, opened 2025-10-23
[ci] publish releases with artifacts

f355:
* before this, CI built packages but didn’t create a release with downloadable .deb's, now it does, for both amd64 and arm64 architectures and three Debian versions: bookworm, trixie and sid.
* in order to make the .deb versioning better, version detection has been changed - on tagged commits, packages use the tag verbatim; otherwise the prevous logic is preserved.
* tag matching on master is derived from VERSION (major.minor) instead of hard‑coded. This yields versions that align with release tags and clearer pre‑release identifiers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,10 +92,11 @@ jobs:
         # Note that the package build covers html docs
 
   package-arch:
-    runs-on: ubuntu-24.04
+    runs-on: ${{ matrix.runner }}
     strategy:
       matrix:
-        image: ["debian:bookworm", "debian:sid"]
+        runner: ["ubuntu-24.04", "ubuntu-24.04-arm"]
+        image: ["debian:bookworm", "debian:trixie", "debian:sid"]
     container:
       image: ${{ matrix.image }}
       # IPC_OWNER is needed for shmget IPC_CREAT
@@ -129,7 +130,7 @@ jobs:
         DEBIAN_FRONTEND: noninteractive
       run: |
         case "${{matrix.image}}" in
-            debian:sid|debian:bookworm)
+            debian:sid|debian:bookworm|debian:trixie)
                 exit 0
                 ;;
             *)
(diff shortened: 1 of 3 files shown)


f355:
an example of a release tag build can be seen here: https://github.com/f355/linuxcnc/releases/tag/v2.10.0-pre1

andypugh:
Thanks for this, it has almost (but not quite..) perfect timing as I released v2.9.7 on Wednesday. 
At the moment buildbot2 isn't working, so I have been building debs manually. 

I am assuming that the debs are built on the standard (non-realtime) kernel? This probably doesn't matter with uspace builds as the tests still run, and I think that the packages will install regardless. A solution for the (rather neglected) RTAI version would be a nice-to-have but I can't see that happening in Github. 

Is there a way to separate actual releases (with a new signed tag like v2.9.7) from incremental builds? There is maybe some room for confusion if we have (potentially broken) master builds in a space called "releases".

f355:
yes, the debs are built on the bog-standard, non-realtime kernel. I don't think it should cause any major problems, since the kernel headers it is building against are the same, but the tests are running in non-realtime, so it's not totally perfect. not much we can do about this.

we can try building RTAI debs too, but I'm not familiar with what that entails.

regarding possible confusion between incremental and release builds - it's not an issue as there are no incremental builds :) the github releases are only created on tag pushes, regular commits to master still run the pipeline with tests and everything, but don't create releases. the artifacts can still be downloaded from the action.

andypugh:
The next question, then, is whether this PR should be into 2.9 (or even 2.8, but I don't anticipate any more 2.8 releases)

f355:
I don't know, you tell me :) I can rebase it on any branch if needed.

andypugh:
Please rebase on 2.9, then I will merge-up to Master.

andypugh:
Actually, forget the re-base. I will merge into master and then cherry-pick to 2.9.

f355:
@andypugh I have it cherry-picked already, it's building now - I'll send a PR for 2.9 in a few minutes.

f355:
#3588

smoe:
This is darn cool! Thank you, @f355 !

andypugh:
I pushed my version as I was heading out for the weekend and wanted to test it with 2.9.7
I will look at your (probably less hurried) version when I get back. 

Again, many thanks for this.

PR #3591, opened 2025-10-25
docs/src/Submakefile: Extra patterns for Makefile simplification (smoe:docs_submakefile_taming)

smoe:
Hello,
There is some more work to do, but as a start, this shall help to reduce the effort to add additional languages to our build system. Also, the Submakefile can now (or again) be executed from the docs subdirectory.
The $(ECHO) variable was used in the code before, have only now also defined it if it was previously undefined.
This is also a preparation for the Ukrainian documentation, which appears to be complete and building nicely. Well done!

### DIFF
--- a/docs/src/Submakefile
+++ b/docs/src/Submakefile
@@ -1,6 +1,8 @@
 .PHONY: docs docclean checkref checkref_en checkref_ar checkref_de checkref_es checkref_fr checkref_nb checkref_ru checkref_sv checkref_ta checkref_tr checkref_ua checkref_zh_CN
 .PHONY: pdfdocs htmldocs install-doc install-doc-pdf install-doc-html
 
+SHELL=/bin/bash
+
 # To make linuxcnc-checklink widely available
 export PATH:=$(BASEPWD)/../scripts:$(PATH)
 
@@ -196,7 +198,7 @@ DOC_SRCS_EN := \
 	Master_Developer.adoc
 
 # Map two-letter codes to human-readable language names
-# (this can later be extended to handle zh_CN, zh_TW, etc.)
+# (this can extended on demand to handle zh_TW, etc.)
 define lang_name
 $(if $(filter $1,ar),arabic,\
 $(if $(filter $1,de),german,\
@@ -212,8 +214,6 @@ $(if $(filter $1,ua),ukranian,\
 $(if $(filter $1,zh_CN),chinese,\
 unknown))))))))))))
 endef
-
-
 
 GENERATED_TRANSLATED = $(foreach l, $(LANGUAGES), \
 	$(DOC_SRCDIR)/$(l) \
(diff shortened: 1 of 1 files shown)


docs/src/Submakefile
```diff
@@ -168,8 +195,24 @@ DOC_SRCS_EN := \
 	Master_Integrator.adoc \
 	Master_Developer.adoc
 
-LANGUAGES := $(shell sed -e's/#.*//' < $(DOC_DIR)/po4a.cfg | grep '^\[po4a_langs\]' | cut -d" " -f2-)
-LANGUAGES_MATCH := $(shell echo $(LANGUAGES) | tr " " "|")
+# Map two-letter codes to human-readable language names
+# (this can later be extended to handle zh_CN, zh_TW, etc.)
```
andypugh: You say that it can be expanded for zh_CN then include it? (and also state "two-letter")

Also the Chinese docs use all lower case zh_cn.

(I don't know enough to know if either matter)

docs/src/Submakefile
```diff
@@ -335,11 +343,22 @@ PDF_TARGETS_TR = $(addprefix $(DOC_DIR)/, $(subst tr/,, \
...
+	$(subst Master_,LinuxCNC_, $(filter ua/Master_%,$(DOC_SRCS_UA))))))
+
 PDF_TARGETS_ZH_CN = $(addprefix $(DOC_DIR)/, $(subst zh_CN/,, \
 	$(patsubst %.adoc,%_zh_CN.pdf, \
 	$(subst Master_,LinuxCNC_, $(filter zh_CN/Master_%,$(DOC_SRCS_ZH_CN))))))
 
-PDF_TARGETS = $(PDF_TARGETS_EN) $(PDF_TARGETS_AR) $(PDF_TARGETS_DE) $(PDF_TARGETS_ES) $(PDF_TARGETS_FR) $(PDF_TARGETS_NB) $(PDF_TARGETS_RU) $(PDF_TARGETS_SV) $(PDF_TARGETS_TA) $(PDF_TARGETS_TR)
+PDF_TARGETS = $(PDF_TARGETS_EN) $(PDF_TARGETS_AR) $(PDF_TARGETS_DE) $(PDF_TARGETS_ES) $(PDF_TARGETS_FR) $(PDF_TARGETS_NB) $(PDF_TARGETS_RU) $(PDF_TARGETS_SV) $(PDF_TARGETS_TA) $(PDF_TARGETS_TR) $(PDF_TARGETS_UA)
+# Do not add $(PDF_TARGETS_ZH_CN) to the line above, it is added below - only for XETEX
+# FIXME: Since we are no longer supporting stretch, this may need to be revisited
```
andypugh: Time to re-visit, I think

andypugh:
I will re-run this to see if the CI fail is as random as it looks.

PR #3592, opened 2025-10-26
docs: vismach, elaborate on the functionality of 'Capture()'

Sigma1912:


Sigma1912:
Looks like this:
vismach

c-morley:
I think this would apply to qt vismach too. Could you add it to your pull request?

Sigma1912:
Added to qtvcp-vismach:
qtvcp-vismach

c-morley:
Ur the best!

PR #3598, opened 2025-11-03
Preparing everything for the Ukranian language (smoe:docs_adding_lang_ua)

smoe:
Hello,

This is a first-timer for me. We'll see how this goes. This PR now adds all that is needed to compile the Ukranian documentation. You may recall that I was so very excited about the build working as intended, when I now have seen that docs/po/ua.po does not feature the translations. So something does not work out, yet :)

@petterreinholdtsen, please kindly lend me your extra eyeballs.

My plan was to see some more fixes to the documentation with a then updated .pot file, have that once processed via weblate, and then add the ua.po file while weblate is locked.

### DIFF
--- a/docs/help/.gitignore
+++ b/docs/help/.gitignore
@@ -20,6 +20,7 @@ sai/*
 sk/*
 sv/*
 tr/*
+ua/*
 uk/*
 vi/*
 zh_CN/*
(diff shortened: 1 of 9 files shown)


smoe:
Argh. I have now regenerated the .pot file without the UA language. Will update that one, obviously.

PR #3600, opened 2025-11-03
Fixing some typos, cleaning up some language

johnobrien:


smoe:
This PR cannot be accepted as it is since it contains extra patches that are not yours with all the bits that have since happened in the linuxcnc/master branch. Those extra bits need to be removed. Only the first commit shall remain.

Here how I think this would work:
 * Rename your local master branch to something else, like "typos": `git branch -m master typos`
 * Check out the real master branch again: `git checkout -b master origin/master`
 * Back to the typos branch: `git checkout typos`
 * Make all commits disappear that are not your first commit with the typos: `git rebase -i master`, there put a "d" for "drop" instead of the "pack" as the very first word for all commits that is not the one with the typo
 * Push that typo branch to your linuxcnc repository on github
 * Open https://github.com/LinuxCNC/linuxcnc/pull/3600 and find the "Edit" button to the upper right, there change the branch to the "typos" branch
 
 
The result should then look just like my attempt to save your work on http://github.com/LinuxCNC/linuxcnc/pull/3602.

PR #3597, opened 2025-11-03
Fixing typos, cleaning up some stuff, take 2

johnobrien:
Trying again, I think I have the correct changes this time.

PR #3602, opened 2025-11-03
Fixed some typos, removed references to wiki as obsolete and outdated.

smoe:
This is a rescue of many smallish changes that I liked in https://github.com/LinuxCNC/linuxcnc/pull/3597 .

The Wiki seems like "gone". It thus seems fair to also remove it from our documentation.


andypugh:
The Wiki is still there, you just can't get to it with https.

PR #3604, opened 2025-11-04
Typos PR, take 4?

johnobrien:
Hi smoe, trying this again.

### DIFF
--- a/docs/src/getting-started/about-linuxcnc.adoc
+++ b/docs/src/getting-started/about-linuxcnc.adoc
@@ -64,14 +64,14 @@ Some IRC etiquette::
 * If you're really new to all this, think a bit about your question
   before typing it.  Make sure you give enough information so
   someone can answer your question or solve your problem.
-* Have some patience when waiting for an answer, sometimes it takes a
-  while to formulate an answer or everyone might be busy working or
+* Have some patience when waiting for an answer. Sometimes it takes a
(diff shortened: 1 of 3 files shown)


andypugh:
You seem to have each commit twice, but I don't think that will matter.

andypugh:
However it can't be merged while marked as "Draft".

smoe:
Looks much better already :-) Two bits:
 * Since this is a bit of a training-PR, I think we should get this right and have the changes only once. I am not sure how you got those in, my hunch is that it is a leftover from your previous work on the master branch of yours. The master branch on the LinuxCNC repository has since changed in the mean time, so you need to update that again. And then, well, it is a rebase of the typos branch to the then updated master branch.
 * As @andypugh stated on https://github.com/LinuxCNC/linuxcnc/pull/3602, the wiki still exists, just https does not work. It should not be skipped from the documentation, yet. So, please correct for that.

I a bit swamped this week, but if your trouble is persistent, I would be up for a live support with screen sharing.

johnobrien:
@smoe I think I did what you asked, please let me know if I didn't and I'll try again.

smoe:
We have chatted about your changes in the Video meeting. The Wiki URL is due to soon be changed to a completely new URL in a very new future. Just please be aware for this change to come. Thank you for your fixes already!

PR #3617, opened 2025-11-11
Removing po/ua.po - redundant to po/uk.po

smoe:
Both uk and ua represent the Ukranian language, weblate was irritated.


petterreinholdtsen:
[Steffen Möller]
> Both uk and ua represent the Ukranian language, weblate was irritated.

No they do not.  'uk' is the language code according to ISO 639, while
'ua' is the country code according to ISO 3166.  No wonder Weblate
reacted, it process languages, not countries.  I guess it might be wise
to track down how was ua.po introduced and explain something to the
person doing this.

-- 
Happy hacking
Petter Reinholdtsen

PR #3643, opened 2025-11-23
GTK widget little probe

zz912:
# Modification of Probe Screen application by Serguei Glavatski (c) 2015
This modification moved the icons from one window to containers with switchable vertical tabs. The goal of this modification is to minimize the window.
The modification was made by zz912.
little_probe-001

little_probe-002

little_probe-003

little_probe-004

little_probe-005

little_probe-006

Integration probe screen to Gmoccapy is first click:
pncconf-use_probescreen

Integration of debounce probe signal is second click:
pncconf-debounce_signal


andypugh:
As of 2.9.7 we shouldn't need the probe debounce (or, at least, not as much)

zz912:
I tested it on my CNC tester toy with Aliexpress probe in LCNC 2.10 and I needed deboubence. If you deactivate check box with probe debounce. Debounce component is not used.

hansu:
@zz912 is this probe screen only available if you generate a config with PNCconf? Or is there also a sample config?

zz912:
If you dont use pncconf, you should read readme.md.
https://github.com/LinuxCNC/linuxcnc/blob/master/lib/python/gladevcp/builtin-panels/gtk_little_probe/README.md

hansu:
Ah okay thanks :+1:

Issue #2954, opened 2024-03-31
 /tests/abort/g64/test.ini is not working well.


zz912:
Branche 2.9 RIP Installation

I experimented with warning Python WaitCommandComplete timed out 
https://forum.linuxcnc.org/38-general-linuxcnc-questions/52141-warning-python-waitcommandcomplete-timed-out

I found out that /tests/abort/g64/test.ini is not working well.

If you run it normally, the result is OK.
```
zdenek@cnc:~/linuxcnc/linuxcnc-2.9-WAIT_COMPLETE/src$ ../scripts/linuxcnc -r ../tests/abort/g64/test.ini
Machine configuration file is 'test.ini'
Starting LinuxCNC...
emc/iotask/ioControl.cc 786: can't load tool table.
linuxcnc TPMOD=tpmod HOMEMOD=homemod EMCMOT=motmod
Note: Using POSIX non-realtimezdenek@cnc:~/linuxcnc/linuxcnc-2.9-WAIT_COMPLETE/src$ ../scripts/linuxcnc -r ../tests/abort/g64/test.ini
Machine configuration file is 'test.ini'
Starting LinuxCNC...
emc/iotask/ioControl.cc 786: can't load tool table.
linuxcnc TPMOD=tpmod HOMEMOD=homemod EMCMOT=motmod
Note: Using POSIX non-realtime
Found file(LIB): /home/zdenek/linuxcnc/linuxcnc-2.9-WAIT_COMPLETE/lib/hallib/core_sim.hal
Found file(REL): ./motion-test.hal
task: main loop took 0.120195 seconds
task: main loop took 0.061635 seconds
task: main loop took 0.019434 seconds
Running command 'G61'
G61 post-command:  Control mode = 610; tolerances = P0.001 Q0.000
Running program 'test.ngc'
task: main loop took 0.020032 seconds
Error - Python WaitCommandComplete timed out 
G61 pre-abort:  Control mode = 610; tolerances = P0.000 Q0.000
task: main loop took 0.035752 seconds
G61 post-abort:  Control mode = 610; tolerances = P0.000 Q2.000
z=0.0; max_x = 5.000000; max_y = 1.000000; samples = 6245

task: main loop took 0.019740 seconds
G61 redo post-command:  Control mode = 610; tolerances = P0.000 Q2.000
Running program 'test.ngc'
task: main loop took 0.021903 seconds
Error - Python WaitCommandComplete timed out 
[...]

PR #3656, opened 2025-12-08
Gtk verser & little probe screen -fix timeout wait_complete

zz912:
The default wait time for command.wait_complete() is 5 seconds.
If the probe movements lasted more than 5 seconds, they were interrupted, or the measurement behavior was unexpected.
I set the wait time for command.wait_complete() to 600 seconds, which is 10 minutes. If the time is exceeded during an extremely long measurement, the operator will be informed.

### DIFF
--- a/lib/python/gladevcp/builtin-panels/gtk_little_probe/gtk_little_probe.py
+++ b/lib/python/gladevcp/builtin-panels/gtk_little_probe/gtk_little_probe.py
@@ -98,6 +98,8 @@ class ProbeScreenClass:
         self.textarea = builder.get_object("textview1")
         self.e = linuxcnc.error_channel()
 
+        self.timeout = 600 # waiting for the MDI command to complete in seconds
+
         self.buffer = self.textarea.get_property('buffer')
         self.chk_set_zero = self.builder.get_object("chk_set_zero")
         self.chk_set_zero.set_active( self.prefs.getpref( "chk_set_zero", False, bool ) )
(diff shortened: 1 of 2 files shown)


andypugh:
Thanks. This seems sensible. 

Perhaps the sample code segments on https://linuxcnc.org/docs/stable/html/config/python-interface.html should use an explicit timeout, until a day ago I thought that it waited indefinitely, and 5s seems very short for some possible commands (such as an MDI move)

zz912:
> until a day ago I thought that it waited indefinitely

I am not happy that the command.wait_complete() function waits 5 seconds. On the other hand, I understand the author's concerns that an infinite wait could often cause LCNC to freeze.

I have already learned to think about this and accept it. What I consider a problem is that this function does not inform in any way that an overtime has occurred.
Errors caused by this are very difficult to find. I wanted to change this https://forum.linuxcnc.org/38-general-linuxcnc-questions/52141-warning-python-waitcommandcomplete-timed-out , but this post did not generate much response. I would need help choosing a suitable way to inform the user that a timeout has occurred. I have already found cases where exceeding 5 seconds is used intentionally for the command.mdi() function, so this should be a non-violent form of informing the user. I find the use of fprintf() barbaric, I would like some debug message.

andypugh:
Could wait_complete() hook into the linuxcnc.comand.display_msg(string) code? I _think_ that is probably the same as the G-code (DEBUG, ....) which will print an informational message at the bottom of the GUI screen.

zz912:
Are you using Gtk verser or little probe screen?
I would be happy for feedback.

Which GUI are you using?
I use Gmoccapy and it has no informational message at the bottom of the GUI screen.

> Could wait_complete() hook into the linuxcnc.comand.display_msg(string) code?

What do you mean? Do you want to add a message for wait_complete() in the probe screen, or for all uses of wait_complete()?


Do you want add linuxcnc.comand.display_msg(string)? Or do you want replace self.add_history() by linuxcnc.comand.display_msg(string)?

andypugh:
I was wondering if a timeout in wait_complete() should _always_ pop-up a message, and be built-in to the wait_complete() code itself. 
Eventually folk might get annoyed enough to fix it.

zz912:
> I was wondering if a timeout in wait_complete() should _always_ pop-up a message, and be built-in to the wait_complete() code itself.
> Eventually folk might get annoyed enough to fix it.

I agree with message  built-in to the wait_complete() code itself. 
https://forum.linuxcnc.org/38-general-linuxcnc-questions/52141-warning-python-waitcommandcomplete-timed-out
Nobody react for creating Pull Request.

I dont agree with  pop-up message.  There may be situations where timeout is used intentionally and it would be unnecessarily annoying. wait_complete() is very old function and it is used in many users plugin surely.
I think, this is example https://github.com/LinuxCNC/linuxcnc/issues/2954 , where timeout is necessary.

andypugh:
I wasn't meaning a modal must-be-acknowledged dialog, rather a message in the LinuxCNC "rtapi_print" stream (or equivalent) that would be shown in the message area of the GUI.

zz912:
Hello,

I vote for silent message, but we can try message in GUI. What should I do now? Or how can I help you.

Issue #3607, opened 2025-11-06
Support for additional device ID (0xeb91) for XHC-WHB04B-6 clone (EDING CNC rebrand)


Gogonfa:
Hi there,

it would be great if you could add support for an additional productID for the XHC-WHB04B-6 handwheel.

I own an EDING CNC rebranded clone, and they have changed the device ID from the standard 0xeb93 to 0xeb91. It is specified in the USB.h file of this user component. I made a rebuild and it worked.

Everything else seems to be identical to the regular XHC-WHB04B-6 device.

Adding support for this ID would be a huge benefit for EDING CNC users (like me 😊) who want to use their existing hardware with LinuxCNC instead of having to buy a second pendant.

Thanks a lot for your great work and for considering this small but helpful addition!

Gogonfa

zz912:
If an LCNC developer doesn't take on this task, I think you should make a Pull Request with this modification.
It would be ideal if it could test itself which ID is correct and use that.
The second option is to make an OPTIONS for the component, which will change the ID for EDING CNC.

Sigma1912:
see https://github.com/LinuxCNC/linuxcnc/pull/3659

andypugh:
This change has been added to the master (2.10 / development) branch. 
If you want to try it you should be able to get a .deb file to install that version from the buildbot in a couple of hours from now. 

http://buildbot2.highlab.com/debian/dists/bookworm/master-uspace/binary-amd64/

PR #3659, opened 2025-12-09
XHC-WHB04B-6: Add command line option -P for devices with ProductId other than 0xeb93

Sigma1912:
ProductId defaults to 0xeb93 if the -P option is omitted.

Example use:
	xhc-whb04b-6 -ue -P 0xeb91

Also see:
https://github.com/LinuxCNC/linuxcnc/issues/3607

### DIFF
--- a/src/hal/user_comps/xhc-whb04b-6/main.cc
+++ b/src/hal/user_comps/xhc-whb04b-6/main.cc
@@ -114,9 +114,15 @@ static int printUsage(const char* programName, const char* deviceName, bool isEr
         << "    Force being silent and not printing any output except of errors. This will also inhibit messages "
             "prefixed with \"init\"." << endl
         << endl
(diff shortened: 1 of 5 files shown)


andypugh:
Have you been able to test this with hardware? (I imagine that checking with the normal XHC and then setting it to the wrong ID and seeing it fail to load would be a good test, if you don't have a device with an unusual ID)

Sigma1912:
Yes, tested with real hardware (productId 0xeb93)

1. Using 'xhc-whb04b-6 -ue' (ie using default 0xeb93) shows successful connection.
2. Setting the default to 0xeb91 and using 'xhc-whb04b-6 -ue', connection fails. 
3. Using 'xhc-whb04b-6 -ue -P 0xeb93' shows successful connection again.

PR #3660, opened 2025-12-10
Widget - GTK Mesa tests screen

zz912:
# GTK Mesa tests screen
Embedded panel for PC and Mesa tests by zz912

This screen allows the user to verify whether they have a suitable and well-tuned PC for their Mesa card.

Based on the design of Mesa Configuration Tool II
https://github.com/jethornton/mesact
Copyright (c) 2022 jethornton
gtk_mesa_tests-001
gtk_mesa_tests-002
gtk_mesa_tests-003
gtk_mesa_tests-004

### DIFF
--- a/lib/python/gladevcp/builtin-panels/gtk_mesa_tests/README.md
+++ b/lib/python/gladevcp/builtin-panels/gtk_mesa_tests/README.md
@@ -1,0 +1,26 @@
+# GTK Mesa tests screen
+Embedded panel for PC and Mesa tests by zz912
+
+This screen allows the user to verify whether they have a suitable and well-tuned PC for their Mesa card.
+
+Based on the design of Mesa Configuration Tool II
+https://github.com/jethornton/mesact
+Copyright (c) 2022 jethornton
+
+---
+
+### Load with gladevcp command:
+```bash
+gladevcp gtk_mesa_tests
+```
+
+---
+
+## Modification INI file for embedded panel in Gmoccapy:
+```ini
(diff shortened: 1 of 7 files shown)


zz912:
The functionality is done, but I'm not sure about the licensing terms.

https://forum.linuxcnc.org/gmoccapy/58037-gtk-mesa-tests-screen

zz912:
I would like to ask @jethornton to look at the license and see if he agrees with it.

https://github.com/LinuxCNC/linuxcnc/pull/3660/changes#diff-c03e806812352483661b08418577e330bf5a8cf24bdb631c3915a79f51cc850dR2-R19

jethornton:
I had a temporary change in my license due a jerk. I've returned it back to the original MIT license which means feel free to use my code.

zz912:
I would like to ask for a merge. Maybe @pcw-mesa would like this screen. I think it would help many Mesa card users to find out if their hardware is suitable for Mesa cards.

hansu:
It looks for me like a configuration tool which is not needed when Linuxcnc is running. Is that right?
In this case I wonder if it makes sense to integrate it into the GUI (even though when optional).

zz912:
> It looks for me like a configuration tool which is not needed when Linuxcnc is running. Is that right?

No, it's exactly the other way around. LCNC must be running. The tmax values ​​are based on a running LCNC.

To Hans:
Do you use Mesa cards?

hansu:
> Do you use Mesa cards?

Not yet, but I have one for the bigger mill.

Issue #3314, opened 2025-02-02
Homing breaks if a second axis home command is requested before the first completes homing.


Lcvette:
linuxcnc 2.9.4 deb installation fully up to date
linux Debian 12 Bookworm, fully up to date
axis sim

I was asked if there was a bug in Probe Basic when requesting a second axis to be homed prior to the first axis completing its homing routine.  I tested this in axis sim config to find that the bug is present there as well.  the axis will actually move in the wrong direction indefinitely until user intervenes (no soft limits are active because homing fails to complete)

how to test:
I had to slow down the home_search and home_latch velocities to 0.1 in the ini for all axes in order to be able to trigger each axis before it completed and verify the bug.

select the x axis and press home key, then select y axis and press home key, the screen would freeze and then unfreeze and the homing process was then broken with the x axis moving away from the home switch.

not sure of a solution besides perhaps disabling any subsequent user homing requests while an axis is actively homing.

rmu75:
Can confirm with current master and sim axis mm config with slowed down velocities.

After issuing second home command, screen freezes until first home completes, second home is hosed.

```
Component Pins:
Owner   Type  Dir                 Value  Name
    31  float OUT                     0  joint.0.acc-cmd
    31  bit   OUT                  TRUE  joint.0.active
    31  bit   OUT                  TRUE  joint.0.amp-enable-out
    31  bit   IN                  FALSE  joint.0.amp-fault-in
    31  float OUT                     0  joint.0.backlash-corr
    31  float OUT                     0  joint.0.backlash-filt
    31  float OUT                     0  joint.0.backlash-vel
    31  float OUT               32.1068  joint.0.coarse-pos-cmd
    31  bit   OUT                 FALSE  joint.0.error
    31  float OUT                     0  joint.0.f-error
    31  float OUT                 0.254  joint.0.f-error-lim
    31  bit   OUT                 FALSE  joint.0.f-errored
    31  bit   OUT                 FALSE  joint.0.faulted
    31  float OUT               1000.15  joint.0.free-pos-cmd
    31  bit   OUT                  TRUE  joint.0.free-tp-enable
    31  float OUT                   0.1  joint.0.free-vel-lim
    31  s32   OUT                     7  joint.0.home-state
    31  bit   IN                   TRUE  joint.0.home-sw-in <== XZhomesw
    31  bit   OUT                 FALSE  joint.0.homed
    31  bit   OUT                  TRUE  joint.0.homing
[...]

Issue #2630, opened 2023-09-06
Gmoccapy - Interupting homing


zz912:
RIP 2.9 Branche
Debian Bookworm
Oracle VM VirtualBox
Configuration: [Z0069_sim-001.zip](https://github.com/LinuxCNC/linuxcnc/files/12535504/Z0069_sim-001.zip)

If I move the GUI while homing, the homing of one axis is interrupted and another axis starts to be referenced.
![Interupt_homing](https://github.com/LinuxCNC/linuxcnc/assets/96618597/38911213-acfb-4fe3-840e-5ce32481a8e2)
Failed to simulate this in Axis.

hansu:
Does this also happen on a real machine with hardware limit switches?

Sigma1912:
Yes, this also happens on my real hardware.
Also happens when I switch to another desktop while homing.

zz912:
> Does this also happen on a real machine with hardware limit switches?

Yes

Sigma1912:
Some observations:

1. [edit] This bug shows up intermittently on all my installations[/edit]
2. Activating debugging for gmoccapy shows this when homing breaks:

> [Gmoccapy][DEBUG]  on button home clicked = ref_all (gmoccapy:4062)
[Gmoccapy][DEBUG]  Hal Status not all homed 01 (gmoccapy:2703)
[Gmoccapy][DEBUG]  Stopped jogging on focus-out-event (gmoccapy:3184)
[Gmoccapy][DEBUG]  Hal Status not all homed 012 (gmoccapy:2703)


Which points to this:
https://github.com/LinuxCNC/linuxcnc/blame/master/src/emc/usr_intf/gmoccapy/gmoccapy.py#L3177
 
If I disable the function and add another debug print:

<img width="841" height="227" alt="Image" src="https://github.com/user-attachments/assets/eb7b063b-faa2-468f-b017-a10b2ef042e3" />

I get this output and the bug disappears:

> [Gmoccapy][DEBUG]  on button home clicked = ref_all (gmoccapy:4064)
[Gmoccapy][DEBUG]  Hal Status not all homed 01 (gmoccapy:2703)
[Gmoccapy][DEBUG]  WOULD HAVE Stopped jogging on focus-out-event (gmoccapy:3178)
[Gmoccapy][DEBUG]  Hal Status not all homed 0 (gmoccapy:2703)
[Gmoccapy][DEBUG]  hal status motion mode changed (gmoccapy:3028)
[...]

PR #3664, opened 2025-12-12
Fix homing bug when trying to home while a homing sequence is already in progress

Sigma1912:
Fixes https://github.com/LinuxCNC/linuxcnc/issues/3314 
The modification to 'command.c' prevents the homing process breaking where the joint would continue moving indefinitely.
The modifications to axis.py and gmoccapy.py prevent the gui freezing that occurs due to wait_complete() timing out to the default 5 seconds.

### DIFF
--- a/src/emc/motion/command.c
+++ b/src/emc/motion/command.c
@@ -1363,7 +1363,10 @@ void emcmotCommandHandler_locked(void *arg, long servo_period)
 	                   joint_num);
                 return;
 	    }
-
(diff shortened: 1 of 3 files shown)


zz912:
Is this fix also related to this issue?
https://github.com/LinuxCNC/linuxcnc/issues/2630

Sigma1912:
https://github.com/LinuxCNC/linuxcnc/issues/2630 is not fixed by this.

PR #3667, opened 2025-12-15
gmoccapy: fix homing abort on focus out event

Sigma1912:
Don't cancel joint jogging on focus out if the joint is currently homing.
This fixes https://github.com/LinuxCNC/linuxcnc/issues/2630


zz912:
Thank you.

Issue #3658, opened 2025-12-09
Python Interface motion_mode fails to change with VOLATILE_HOME = 1


jethornton:
If a joint has VOLATILE_HOME = 1 and all joints are homed the motion_mode is TRAJ_MODE_TELEOP as expected. Turn the power off and the joint(s) with VOLATILE_HOME = 1 are unhomed. The motion_mode does not change to TRAJ_MODE_FREE as expected when the joints are unhomed. If you only have one joint with VOLATILE_HOME = 1 the motion_mode does not change when you home that joint. If you have more than one joint with VOLATILE_HOME = 1 when you home the first joint then the motion_mode changes to TRAJ_MODE_FREE.

rmu75:
I assume you observed that on current master branch. Is this a new problem?

Sigma1912:
testing on current master shows that it is not just the python status but also the 'motion.teleop-mode' halpin that is not updating properly:

<img width="1142" height="646" alt="Image" src="https://github.com/user-attachments/assets/d8e25032-f232-4926-8d8c-8d32c3831e20" />

Sigma1912:
Unhoming of joints set to volatile is handled here:
https://github.com/LinuxCNC/linuxcnc/blob/69d26ee514d67f6cf9accdc75218ae593b3b1409/src/emc/motion/homing.c#L283

It seems that outside of homing.c we have no way of knowing if there are any joints set to volatile, we don't even know if 'set_all_unhomed(-2, motstate) did unhome any joints at all unless we explicitly check if all joints are homed after calling it.

Since this call may or may not leave joints unhomed we cannot switch to mode_free before calling it and there seem to be no provisions to do so after.

andypugh:
VOLATILE_HOME is intended to unhome joints if encoder tracking is lost. I am not sure that amp-fault necessarily implies tracking lost. 
The docs say that it is activated if machine power is off or e-stop is on. 
The motion manpage doesn't say what amp-fault does.

Sigma1912:
> VOLATILE_HOME is intended to unhome joints if encoder tracking is lost. I am not sure that amp-fault necessarily implies tracking lost.

Neither does switching the machine off or even an Estop condition. 

Problem is that the current behavior is VERY inconsistent:
- Estop unhomes volatile joints
-  Switching the machine state to off using the gui button or halui.machine.off unhomes volatile joints
[...]

PR #3669, opened 2025-12-16
Set motion mode to FREE when volatile joints have been unhomed

Sigma1912:
Joints set to volatile are automatically unhomed when task state switches to OFF or ESTOP 
If a joint is thus unhomed motion state needs to be set to 'free' 
Fixes https://github.com/LinuxCNC/linuxcnc/issues/3658

### DIFF
--- a/src/emc/motion/control.c
+++ b/src/emc/motion/control.c
@@ -866,6 +866,15 @@ static void set_operating_mode(void)
     emcmot_joint_t *joint;
     double positions[EMCMOT_MAX_JOINTS];
 
+    /* joints set to volatile are unhomed when task state switches to OFF or ESTOP
+       in this case we need to switch to 'free' mode */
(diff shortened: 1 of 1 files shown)


Sigma1912:
Closing this for further investigation into failing tests.

Sigma1912:
The reason some test are failing is that some of these tests actually run in unhomed mode.
This PR assumes that motion state should not be 'teleop' or 'coord' while not all-homed which is not correct. 
Linuxcnc can also be run in unhomed state (NO_FORCE_HOMING = 1).
Hence we would need to know if this option has been set before checking  'if (!get_allhomed() ...' 
However this information seems to only be available in task:
https://github.com/LinuxCNC/linuxcnc/blob/f25a2ede8b1cbfdd4ada1a317e62d155d1838be8/src/emc/task/emctaskmain.cc#L3228

PR #3671, opened 2025-12-17
docs: bump linuxcnc version, details on dd for image handling, (((index))) not in header (smoe:docs image handling)

smoe:
po4a has difficulties separating the header from the index when there is a ")" in the index and I see warnings in the asciidoc parsing that I accredit to the positioning of these index entries in the header.
Substituted some 2.9.4 occurrences with 2.9.7.

### DIFF
--- a/docs/src/common/emc-history.adoc
+++ b/docs/src/common/emc-history.adoc
@@ -2,8 +2,9 @@
 :toc:
 
 [[cha:linuxcnc-history]]
-= Origin(((History)))
+= Origin
 
+(((History)))
 EMC (the Enhanced Machine Controller) was created by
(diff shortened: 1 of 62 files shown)


andypugh:
Should this be in 2.9 or master?

smoe:
2.9 would likely also benefit, and since this only some improved formatting with no functional effect, so I redid it for the release branch. When eventually forwarding to 2.10 please just ignore any conflicts (of which there are quite a few expected) and ping me.

The proper use of indices is explained here https://docs.asciidoctor.org/asciidoc/latest/sections/user-index/ . The multi-level indices have found only partial entry in our documentation and are in parts confused with having multiple entries for the same. I fixed the obvious but at some point we need to discuss where we want to go with this.

Issue #1528, opened 2022-01-24
Linuxcnc interprets Freecad postprocessor Gcode  lines with big Parameters I and j wrong, 


Strahlensauger:
gcode created with FreeCAD consisting of G2/G3  lines with big I an J parameters,

## This is what I expected to happen:

loading the gcode with linuxcnc should show the shape of the drawing (a dwarf door)

## This is what happened instead:

Linuxcnc shows huge circles

## It worked properly before this:
(If the behavior changed after making a particular change in hardware or
software, describe the change you think is responsible.  E.g., "after upgrading
from LinuxCNC 2.7.3 to 2.7.4")

## Information about my hardware and software:

 * I am  using this Linux distribution and version (often, shown by `lsb_release -a`):
 * No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 10 (buster)
Release:	10
Codename:	buster

 * 
 * I am using this kernel version (shown by `uname -a`):
 * Linux cnc 4.19.0-18-rt-amd64 #1 SMP PREEMPT RT Debian 4.19.208-1 (2021-09-29) x86_64 GNU/Linux

 * I am running ...
   * [ x] A binary version from linuxcnc.org (including buildbot.linuxcnc.org)
   * [ ] A binary I built myself
   * [ ] A binary version from some other source besides linuxcnc.org
 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`):
 * I am using this user interface (GUI) (e.g., AXIS, Touchy, gmoccapy, etc):
 * AXIS
 * I am using this interface hardware vendor and chipset (e.g., parallel port, ethernet port, FPGA card): 
 * MESA 7i76e

[Wichteltuer.txt](https://github.com/LinuxCNC/linuxcnc/files/7924432/Wichteltuer.txt)

andypugh:
This is rather pathological G-code, but it does seem that LinuxCNC is interpreting it incorrectly. 
(for example an arc length of about 5mm with a radius of 7km...)
It should be a nearly-straight and short line, but is instead being interpreted as slightly more than a full revolution. 

My guess is that the ends are so close together relative to the arc radius that the interpreter is deciding to make a full circle (as it normal ix the start and ends are identical) 

[...]

PR #3683, opened 2026-01-02
Add optional S-curve trajectory planner with jerk limiting

grandixximo:
Just to let you guys know what I am working on, you can read more in the forum

https://forum.linuxcnc.org/38-general-linuxcnc-questions/33459-linuxcnc-s-curve-accelerations?start=520#340865

and I am also available on the discord, just `@grandixximo` I think I should get a notification, or post in this PR.

I could have the PR cleaner without the SCURVE_QUICKREF.txt that's just for general reference of how this PR works.

The s-curve is for now optional, and has to be enabled for testing, so you could keep it in and do testing in the background, or I could replace the TP now, but I don't think it is time to do that yet, could keep the PR open till you guys feel is time to merge it, I'll try to keep the fork in line with master, so when is time it could be merged cleanly.


grandixximo:
cannot build because of motion-logger expected output, should now include JERK

how to proceed?

c-morley:
fix this file on your system:
https://github.com/LinuxCNC/linuxcnc/blob/master/tests/interp/m98m99/12-M99-endless-main-program/expected.motion-logger
This file is what the expected test output should be. add the required lines. 

Probably other tests fail too? You can run the tests on your RIP system to test.

grandixximo:
my apologies, I'm a bit dumbfounded by the tests

 Runtest: 273 tests run, 272 successful, 1 failed + 0 expected, 3 skipped, 0 shmem errors
Failed: 
./tests/build/header-sanity

this means the header-sanity failed?

on my system I get 

Runtest: 276 tests run, 226 successful, 50 failed + 0 expected, 0 skipped, 5 shmem errors

grandixximo:
ok, that worked, it's still not working on my system, but that must be something with my system in particular, not sure...

Some nice pictures for effect.

image

Run this configuration

image

Any issue, report here or in the forum, thanks!

rene-dev:
That looks interesting, can't wait to test it on my machine. It is probably a good idea to check which tests are actually testing the tp, and duplicating them with this enabled.

andypugh:
Is it possible to package this as a separate, pluggable TP?
Does this TP handle finite jerk in more than 3 axes? Does it have N-segment lookahead in more than 3 axes? 
The reason I ask is because there is work going on in parallel to merge the Tormach 9-axis TP, which has full lookahead for combined linear and angular moves. (Whereas the current TP drops down to single-segment lookahead in that case)
LinuxCNC does have the option to choose the TP at load-time, and I can see advantages in being able to choose between these three options depending on the nature of their work. 
This is controlled by the "TPMOD" parameter in the INI file. 
(There really does need to be better documentation of the use of TPMOD, HOMEMOD and INTERPRETER INI settings. (The last one isn't even mentioned in the docs, but is used in sim/axis/canterp.ini))

grandixximo:
maybe I'll try and do that in the coming days, is my approach agreeable so far? Any chance this will be merged? Or is a different approach required? You guys let me know how I should approach this properly, I tried my best to only make a PR for a reasonable change to linuxcnc master, let me know if this will ruff someone feathers and I should not be doing that, or if it's all good and game, I'm really not sure...

grandixximo:
sorry Andy just saw your reply, it still uses the old lookahead, it would be cool to compare it with the Tormach 9-axis TP, at the time this was developed there was no TPMOD, that's what the guys in the forum also pointed me to, but unfortunately I already made it on a different path, I would have to look at TPMOD and possibly refactor the entire code, but if Tormach is already doing the 9 axis one, then maybe this is not needed?
With my TP
All 9 axes get jerk limiting
Full lookahead works
Velocity limiting treats the path as a single entity, not separately limiting linear vs rotary components

andypugh:
> maybe I'll try and do that in the coming days, is my approach agreeable so far? Any chance this will be merged? Or is a different approach required? 

I would love to have a jerk-limited planner, but I have always expected it to be an option controlled by choosing a different TPMOD. I see that you have set things up such that the behaviour is run-time configurable. Would it simplify your code if this was not required?
I would prefer to have the option of running the existing 3-axis blended lookahead TP unmodified, especially considering concerns such as threading/spindle synch motion mentioned at the beginning of the forum thread, just because it is well tested and a known quantity. I might be persuaded to change this view, but consider this in the context also of the 9-axis blend work. 

Grotius on the forum was also looking at this, did he get anywhere? I am not able to keep up with all forum threads. (or even most) at the moment. 

I see jerk-control being relevant to robots and very fast routers. I doubt that anyone needs it on a lathe, but lathes do need spindle-synch motion to work at least as well as it does currently.

andypugh:
We seem to be commenting with a phase lag :-)

> but if Tormach is already doing the 9 axis one, then maybe this is not needed?

As I understand it the difference between the Tormach TP and the current LinuxCNC one (both written by the same chap) is that the current LinuxCNC planner reverts to single-segment lookahead if axes other than XYZ are involved in the motion, whereas the Tormach TP retains N-segment lookahead regardless of which axes are involved. (Note that this is just my understanding of the Tormach TP, I haven't analysed the code in detail to confirm that this is what it does) 

It was my understanding that PP does not have jerk limitation, but the Google AI insists that it does, as does the LinuxCNC forum.  However I would have thought that it would be something that the Tormach web site would mention if it did?

grandixximo:
Yes totally lagging between our comments, I got your email, thanks I will be looking at it for sure, yes one of the main limitations I experienced is the standard look ahead is a bit limited, but I made the best with what we had, yes with trajectory planner set at 0 the TP is basically unchanged.

Sigma1912:
> Grotius on the forum was also looking at this, did he get anywhere?

Certainly not anywhere nearly as functional as this. Also he seems to have since ceased to work on the linuxcnc planner:
https://forum.linuxcnc.org/38-general-linuxcnc-questions/33459-linuxcnc-s-curve-accelerations?start=510#340686

snowgoer540:
Tormach did write a new kinematics module called "square3kins".  I don't recall what version it showed up in, but it's in v2.12.1 for sure.  "full 6dof squareness compensation for an XYZ machine" is in the header of the .c file.  Their website was updated to include "high speed machining", but from what I can see it is certainly not an S-Curve planner.

It's also worth noting that Grotius did claim at one point that they approached him to help write them a S-Curve planner, and he told them to go pound sand because he doesn't believe they give back to LinuxCNC anymore, and his license actually limited them from using his work ... I say that because if it's true, they definitely don't have an S-Curve planner.  

Found it:  https://forum.linuxcnc.org/pathpilot/45286-pathpilot-v2-source-code?start=90#333361
And this:  https://codeberg.org/skynet/libclothoid3d/src/branch/master/LICENCE.md

grandixximo:
Well, I'm much more flexible on licensing, as in, I don't really care as long as it works, and the community is happy to get some new toy, does it really matter? It's not up to me to change Tormach Licensing, but I'd be glad to work together, IMO we shouldn't try to outcompete each other, rather cooperate, anyway now the code is here for all to see, it ain't perfect, but it is a good start for sure. And I'd be glad to talk with the guys at Tormach and find a viable solution for all, the important thing IMO is that both parties feel respected and everyone involved worries be taken care of to a reasonable degree. I'm not here to make a fuss about anything, I had been working and testing this curve for a while, years, I was never in a comfortable situation to share it till now, I am glad I did it though, and I think some good things may come out from this, I don't think there is really much to loose, no regrets.

snowgoer540:
I really intended all of that just to say that I am sure they don't have an S-Curve planner.  They do have 9 axis look ahead and all the stuff Andy mentioned.  

I agree that it benefits all, and we should all be working together.

I am glad you shared this.  I am hopeful it will be that we end up with a 9 axis version of both options in the end :)

grandixximo:
@andypugh I appreciate you being open to persuasion regarding TPMOD. The thing is, I need motmod modifications regardless, jerk parameters have to be read from INI per-axis and per-joint, and that code lives in the motion controller, not the TP layer. TPMOD doesn't eliminate touching core code.

With the integrated approach, users get runtime switching via HAL pin (test both planners on the same job instantly), unified INI configuration (jerk limits alongside velocity/acceleration where they expect them), and we maintain shared lookahead/blending infrastructure. TPMOD would give us architectural separation but we'd still patch motmod anyway, and the performance overhead is negligible, a single branch versus all the kinematics already happening.

I can see real value in runtime switching for hybrid machines, users dialing in settings, and development without restarts. The plugin model is elegant when modules are truly independent, but jerk control creates tight coupling with the motion layer anyway.

That said, I'm genuinely interested if there's a TPMOD path I'm not seeing that preserves runtime switching and clean INI integration. What am I missing?

andypugh:
Motmot is also a loadable module... :-)
(as is TASK, we _always_ use milltask but the idea was to have lathetask, robottask etc.)

My motivation for preferring the new features in a loadable module is that I would prefer to have a guaranteed fallback for our tens of thousands of users if we find a problem with the jerk-limited control.  It would also create a cleaner separation if we end up with a separate 9-axis-blend TP, or any other special purpose planner. 

Saving CPU cycles is not really part of the motivation. 

I imagine that the difficulty is that you need extra elements in data structures and the existing code would need to be taught to ignore those?  I see no difficulty in motmod looking for jerk data in the INI and using a default value if it isn't found, that happens all over the place. 

Does a plugin model specifically exclude live switching? Is that because TPMOD doesn't look at HAL pins?

grandixximo:
are you saying

INI file for users who want S-curve

```
[TASK]
TASK = scurve_task

[EMCMOT]
EMCMOT = scurve_motmod

[TRAJ]
TPMOD = scurve_tpmod
PLANNER_TYPE = 1
MAX_LINEAR_JERK = 1000.0
DEFAULT_LINEAR_JERK = 500.0
```

Am I understanding this right?

To answer your questions and worries, to the best of my knowledge,

Current implementation defaults to PLANNER_TYPE = 0 which is default trapez, scurve is opt in completly, you can enable from ini, or Hal while running, you can turn it off from ini or Hal while running.

I'm not sure about the difficulty, the jerk ini settings already fall trough without problem, if I am allowed to touch those which I already did in this PR we should be good, I then need to get them in the module via tpMotData()

The ini.jerk pins will always be present, but not working with trapez fallback.

Switching via HAL pins is possible, the TPCOMP module can create its own pins, but I think my scurve_tpmod would need to include BOTH trapezoidal and S-curve implementations for runtime switching to work, we would have duplicate code to maintain, not that it changes often...

what are your thoughts about this?

andypugh:
> are you saying
> 
> INI file for users who want S-curve
...

It's one possibility, but I think it is far too much unnecessary duplication. 

In fact unnecessary duplication is a strong argument against the TPMOD approach, as it might mean that bugs need to be fixed in two places. We have had an example like this (https://github.com/LinuxCNC/linuxcnc/issues/1528#issuecomment-1027360082) with very large arcs that needed to be fixed in both the planner and the G-code preview code, but in different ways. 

I wonder if @dngarrett is still around and has an opinion? (He introduced the pluggable TP option)

Sigma1912:
Not sure if the TPMOD option is the right way to integrate this particularly since the current behavior remains the default and the Scurve option will likely be used by an increasing number of users. I have always thought TPMOD (and HOMEMOD) to be for user customization or testing rather than feature integration.
Also, I'm very much in favor of retaining runtime switching between (trapezoidal/scurve) planners.

snowgoer540:
> Also, I'm very much in favor of retaining runtime switching between (trapezoidal/scurve) planners.

x2

andypugh:
We discussed this in our regular Sunday meeting and have concluded that it is best to merge is as soon as we can so that it can get as much testing as possible before the release of 2.10 (which is some months away)

grandixximo:
That's cool, do you need anything from my end?

andypugh:
> That's cool, do you need anything from my end?

I plan on moving the relevant info from the QUiCKREF.txt file into the correct places in the docs before I merge the commit, but then once people start using the planner I expect that there will be a number of questions and issues that arise.

grandixximo:
By the way there is also a PLANNER_TYPE = 2 which I have not documented at all, I intended to remove it, but have not come around to do the cleaning yet, that could be cleaned, or the community could be doing some further testing.

The main practical difference is in how they handle cornering velocities during blends, type 1 is more cautious, while type 2 balances smoothness with speed, in my testing type 1 wins out in most scenarios.

grandixximo:
I did initial doc fix, see if that's acceptable.

andypugh:
Thanks very much for this.

BsAtHome:
Cppcheck warnings introduced with this PR:
```
emc/tp/tp.c:1709:9: warning: %u in format string (no. 1) requires 'unsigned int' but the argument type is 'signed int'. [invalidPrintfArgType_uint]
        tp_debug_print("Wrong term cond = %u\n", prev_tc->term_cond);
        ^
emc/tp/tp.c:1724:5: warning: %u in format string (no. 1) requires 'unsigned int' but the argument type is 'signed int'. [invalidPrintfArgType_uint]
    tp_debug_print("Motion types: prev_tc = %u, tc = %u\n",
    ^
emc/tp/tp.c:1724:5: warning: %u in format string (no. 2) requires 'unsigned int' but the argument type is 'signed int'. [invalidPrintfArgType_uint]
    tp_debug_print("Motion types: prev_tc = %u, tc = %u\n",
    ^
```

grandixximo:
Will check

grandixximo:
Apologies Bertho, it is something you had fixed, and I inadvertently rebased back in. #3985 fixes it back...

BsAtHome:
As they say, shit happens. It is one inevitability of life ;-)

Issue #3391, opened 2025-03-28
RPi install missing dependencies


BsAtHome:
Just did a clean install of the official Raspbian/bookworm on a RPi5 and compiled master from source.

All dependencies were installed using:
```bash
$ scripts/travis-install-build-deps.sh
```
The build went fine, but RIP running default axis example config missed several dependencies:
- python3-opengl
  - libgl1-mesa-dev
  - libglut-dev
  - libglut3.12
- python3-gi-cairo
  - python3-gi
  - python3-cairo
- python3-numpy
  - ibblas3
  - libgfortran5
  - liblapack3
- x11-xserver-utils

(the second indent are the dependencies pulled in)

I can see nearly all dependencies being listed in the <code>debian/control</code> file <i>except (!)</i> the <code>x11-server-utils</code>, which provides in the <code>xset</code> program.

So, how are you supposed to depend-install packages for running and not just building? Secondly, I think package x11-server-utils needs to be added to the dependency list.

smoe:
I think I would add x11-server-utils to the "Recommends"-line of the package, not as a full dependency. This would get it installed by default but removing all the X environments would keep LinuxCNC installed.

The list of packages to install for a complete functionality should go somewhere into the documentation, I suggest, not as a script, maybe to https://github.com/LinuxCNC/linuxcnc/blob/master/docs/INSTALL.adoc ?

What came to mind - we could add some tests and come up with a target that installs the runtime dependencies. We would just need to find a way to have this mostly distribution-agnostic.

andypugh:
Does it make sense to build a system that won't work? 
Whilst some things are not _exactly_ build dependencies,  but the built system won't work without them, isn't it maybe legitimate to have them as build-deps?

smoe:
The dependencies for testing are indeed regular build dependencies. And the -dev libraries usually specified have the package shipping the .so files as a dependency, so these are also installed. A test that is not executable is a failure, so it seems like there are build-time tests missing.

Dependencies for tests that are to be performed after the package was installed go into debian/tests/control.

I have another concern in that the dependencies of linuxcnc-uspace are already fairly impressive. For instance I had to install python3-opencv and python3-espeak as runtime dependencies - but  how many of our installations do indeed use a camera or speak?

BsAtHome:
Maybe there should be a further split in packages, splitting in separate front-ends and a back-end? Then the necessary dependencies can be added to the front-end packages.

andypugh:
[...]

PR #3687, opened 2026-01-06
Add runtime dependency x11-xserver-utils providing xset (smoe:add-runtime-dependency-for-xset)

smoe:
I yet fail to understand where exactly xset is invoked but have no difficulty to accept that one wants the tools shipping with  x11-xserver-utils:

```
/usr/bin/iceauth
/usr/bin/sessreg
/usr/bin/showrgb
/usr/bin/xcmsdb
/usr/bin/xgamma
/usr/bin/xhost
/usr/bin/xkeystone
/usr/bin/xmodmap
/usr/bin/xrandr
/usr/bin/xrdb
/usr/bin/xrefresh
/usr/bin/xset
/usr/bin/xsetmode
/usr/bin/xsetpointer
/usr/bin/xsetroot
/usr/bin/xstdcmap
/usr/bin/xvidtune
```
Addresses https://github.com/LinuxCNC/linuxcnc/issues/3391.

Took the freedom to also harmonize the number of blanks at the start of the lines in d/control files.

### DIFF
--- a/debian/control.main-pkg.in
+++ b/debian/control.main-pkg.in
@@ -34,10 +34,12 @@ Depends:
 Recommends:
     linuxcnc-doc-en | linuxcnc-doc,
     librsvg2-dev,
+    x11-xserver-utils,
     @EXTRA_RECOMMENDS@,
     @PYTHON_IMAGING@,
     @PYTHON_IMAGING_TK@
-Suggests: mesaflash,
+Suggests:
+    mesaflash,
     onboard
 Description: motion controller for CNC machines and robots
  LinuxCNC is a fully-realised CNC machine controller that can interpret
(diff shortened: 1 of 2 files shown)


BsAtHome:
> I yet fail to understand where exactly xset is invoked...

The `xset` program is invoked in axis (`src/emc/usr_intf/axis/scripts/axis.py`) and qtvcp (`lib/python/qtvcp/qt_action.py`).
It is used to disable key repeat while LCNC is running. Whether this is a good idea is a matter of opinion and I am not a friend of programs changing my xserver setting. Especially because it works unreliably when running over a remote connection.

> ...but have no difficulty to accept that one wants the tools shipping with x11-xserver-utils:

Then the better option is then to remove the invocation of `xset` completely and be done with it.

PR #3690, opened 2026-01-07
Gmoccapy: Move offsetpage buttons to table frame, add toggle button for calculator use

### LINKED ISSUE #3513: Gmoccapy offset page - buttons
Just pull this out of #3494.

[@Sigma1912]
> I have also prepared adding a buttonbox to the offsetpage:
> 
> 

Image

> 
> Maybe we could move some of the 'edit offsets', 'Zero G92' , 'Set selected offset' buttons to the new button box. What do you think?

Sigma1912:
- calculator_use state (on/off) can be changed in tooltable and offsetpage but applies to both those tables
- calculator_use state is retained in the preferences file (initial state is OFF)

Resolves #3513

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.glade
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.glade
@@ -762,6 +762,10 @@ uncomment selection</property>
     <property name="visible">True</property>
     <property name="can-focus">False</property>
   </object>
+  <object class="GtkImage" id="img_offset_calculator">
+    <property name="visible">True</property>
+    <property name="can-focus">False</property>
+  </object>
   <object class="GtkListStore" id="lstst_icon_themes">
     <columns>
       <!-- column-name path -->
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -649,17 +649,16 @@ class gmoccapy(object):
         # Activate the recently open tab page
         self.widgets.ntb_tool_and_code_info.set_current_page(self.prefs.getpref("info_tab_page", 0, int))
 
+
         self.jog_btn_size = self.prefs.getpref("jog_btn_size", 48, int)
         self.widgets.adj_jog_btn_size.set_value(self.jog_btn_size)
-            
+
         jog_box_width = self.prefs.getpref("jog_box_width", 360, int)
         self.widgets.adj_jog_box_width.set_value(jog_box_width)
         self.widgets.vbx_jog.set_size_request(jog_box_width, -1)
-        
-        # enable/disable calculator widget for tooltable/offsetpage
-        self.toolpage_use_calc = self.prefs.getpref("toolpage_use_calc", True, bool)
-        self.offsetpage_use_calc = self.prefs.getpref("offsetpage_use_calc", True, bool)
-        self.widgets.chk_offsetpage_use_calc.set_active(self.offsetpage_use_calc)
+
+        # calculator use in offset pages
+        self.offsetpages_use_calc = self.prefs.getpref("offsetpages_use_calc", False, bool)
 
 ###############################################################################
 ##                     create widgets dynamically                            ##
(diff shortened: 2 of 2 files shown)


Sigma1912:
4 Axis with 'Block Height'
4axis_block
9 Axis with 'Block Height'
9axis_block
9 Axis without  'Block Height'
9axis_noblock

hansu:
I am fine with this. Very handy :)
But we should style the scroll bar so that it is always shown. Currently it looks as if there is no G59.3 in the list.
And I wonder if we should keep the position of touch off buttons, so with one empty button space to the left.
But probably not. @gmoccapy ?

Before:
grafik


After:
grafik

hansu:
And you have been a bit too quick. Now with Gmoccapy 3.5.1 in master, you have some conflicts to solve :/.

hansu:
```

  File "/home/cnc/linuxcnc-master/bin/gmoccapy", line 6394, in <module>
    app = gmoccapy(sys.argv)
          ^^^^^^^^^^^^^^^^^^
  File "/home/cnc/linuxcnc-master/bin/gmoccapy", line 193, in __init__
    self.builder.add_from_file(XMLNAME)
gi.repository.GLib.GError: gtk-builder-error-quark: /home/cnc/linuxcnc-master/share/gmoccapy/gmoccapy.glade:766:1 Duplicate object ID 'img_tool_delete' (previously on line 631) (8)
 ```

**Comment by Sigma1912** [comment]
Now I'm confused. With which branch are you getting that error?

**Comment by Sigma1912** [comment]
Seems I have rebased it on gmoccapy-3-5-2 which wasn't my intention. I'll have to sync my repo and fix this tomorrow.

**Comment by hansu** [comment]
>Now I'm confused. With which branch are you getting that error?

Your branch, [Sigma1912:Gmoccapy_Add_buttons_to_offsetpage_frame](https://github.com/Sigma1912/linuxcnc/tree/Gmoccapy_Add_buttons_to_offsetpage_frame)

>Seems I have rebased it on gmoccapy-3-5-2 which wasn't my intention. I'll have to sync my repo and fix this tomorrow.

That's correct, I have targeted your branch to that. But it's the same as master right now.


The glade file should only contain this change:

```
  <object class="GtkImage" id="img_offset_calculator">
    <property name="visible">True</property>
    <property name="can-focus">False</property>
  </object>
```

Sigma1912:
Yes, that is the only change to the glade file. Looks like I made a mistake there while resolving the conflict.

Sigma1912:
Should be fixed now.

hansu:
Got Norbert's Ok :)

gmoccapy:
I am sure, this has to be reopened!
We / You moved the buttons, resulting in new Hal-pin order for the hardware buttons.
First hardware button in bottom bar was edit offsets, now with new placement it must activate "set X to value"

That change need to be changed also in the documentation!
And I did not see the code changes for the hardware button Hal pins to react to other buttons now!

Am I right, or are the changes in an other push?

Norbert

Sigma1912:
Thanks.
I indeed forgot the halpins.
I'll create a new PR for this.

hansu:
> We / You moved the buttons, resulting in new Hal-pin order for the hardware buttons.

I also thought about leaving the first button empty to keep the old positions.
But actually it doesn't matter which hardware button you press - this has the advantage that there is space for one more touch-off button without need to "scroll".

The screenshots in the documentation I usually update when I release the version. 

>Am I right, or are the changes in an other push?

Currently it's in a feature branch linuxcnc/gmoccapy-3-5-2

Sigma1912:
The bug I introduced is that the 'back' button  (ie the last one) in the offsetpage is now connected to 'gmoccapy.h-button.button-8' instead of ..button-9. 
I'm working on this.

> The screenshots in the documentation I usually update when I release the version.

So, should I only update the table entries for the halpin functions here?
https://linuxcnc.org/docs/html/gui/gmoccapy.html#_right_and_bottom_button_lists

hansu:
>So, should I only update the table entries for the halpin functions here?

Yes, please

Sigma1912:
see https://github.com/LinuxCNC/linuxcnc/pull/3728

Issue #3686, opened 2026-01-06
Calculator widget - localization - decimal separator


zz912:
Hello,

I would like to summarize my view on the localization of "Calculator Widget" https://github.com/LinuxCNC/linuxcnc/pull/3500

At this moment, there is a rule for how the localization of decimal separators should look like:

> 2. Localization of float numbers in GUIs
> 
> Different locales use different decimal separators and thousands separators. Locale-specific string-to-float functions should be avoided as they may give unexpected results. (For example the text string "1.58" in de_DE will be converted to 158 by atof()). The following guidelines (based on avoiding ambiguity rather than on "correctness" in any specific locale) are suggested if parsing float to string and vice-versa:
> 
>In the case of input allow either comma (,) or point(.) as a decimal separator, but reject any input that has more than one of either. Space should be accepted but not required as a thousands separator.
> 
> In the case of display either use point (.) consistently or use the current localisation format consistently. The emphasis here being on "consistently".

Source:
https://linuxcnc.org/docs/devel/html/gui/gui-dev-reference.html#_localization_of_float_numbers_in_guis

I think we shouldn't discuss how to localize decimal separators at this moment.
We should only discuss whether some localization complies with the rules.

I know that every rule can be changed. But please don't suggest changes to the rules, because we will never agree. It took a lot of work to make the rules quoted above.

Sigma did a localization of decimal separator for Calculator widget:
![Image](https://github.com/user-attachments/assets/0eba1988-c193-46ea-9d5c-11ffd431d7b3)

I was happy with the functionality, but two problems arose:
1) I disagreed that this behavior is not default, but optional by value self.allow_dot_and_comma = True/False
2) At the time Sigma was doing this localization, there were no rules for localization of decimal separator. So there was a long debate about how it should work. This debate had no conclusion for Sigma. That's why this Issue was created.

My conclusion for Sigma:
1) Leave the Calculator widget - localization - decimal separator functionality as Sigma did it. Ideally, double-check that the functionality matches the rules above (mainly in DE localization).
2) Remove the value self.allow_dot_and_comma
3) The rules allow creating a GUI with decimal commas, but under the condition of format consistently. Therefore, displaying decimal comma in Gmoccapy is not allowed for consistency. Maybe someday some GUI will be created where the author will want to display the current localization format. Then it will be necessary to make the widgets variable. However, this situation has not yet occurred.

[...]

PR #3691, opened 2026-01-07
Gmoccapy: Make value conversion independent of locale setting

Sigma1912:
Commit for offsetpage_widget:
Conversion of calculator entry to float is now independent of locale:
10,123 --> 10.123
10.123 --> 10.123
10.12,3 --> error
10,1 + 10.2 --> error

The old behavior using locale setting can be activated by
setting 'self.use_localization = True'

Commit for Calculator_widget:
Conversion of calculator entry to float is now independent of locale:
10,123 --> 10.123
10.123 --> 10.123
10.12,3 --> error
10,1 + 10.2 --> 20.3

The old behavior using atof() had issues but can be activated by
setting 'self.use_localization = True'

For discussion see https://github.com/LinuxCNC/linuxcnc/issues/3686

### DIFF
--- a/lib/python/gladevcp/offsetpage_widget.py
+++ b/lib/python/gladevcp/offsetpage_widget.py
@@ -88,6 +88,7 @@ class OffsetPage(Gtk.Box):
         self.status = linuxcnc.stat()
         self.cmd = linuxcnc.command()
         self.hash_check = None
+        self.use_localization = False # Set to True for float value conversion using locale settings (not recommended)
         self.display_units_mm = 0 # imperial
         self.machine_units_mm = 0 # imperial
         self.program_units = 0 # imperial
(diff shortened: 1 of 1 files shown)


hansu:
Looks good for me :+1: 

The tool table already accepts dot and comma, but the offset page currently doesn't. Can you change that?

Sigma1912:
Added another commit for the value conversion in the offsetpage widget.

hansu:
It seems that it doesn't work in current master with locale de_DE. Weird, I am sure it had worked when it was merged.
With en_US it's no problem.

calculatorwidget-error

@Sigma1912 can you have a look?

Sigma1912:
My mistake, see https://github.com/LinuxCNC/linuxcnc/pull/3789

Issue #3698, opened 2026-01-15
Add "nonstring" attribute to hostmot2.h: rtapi_u8 board_name[8];  // ascii string, but not NULL terminated!


smoe:
```
hal/drivers/mesa-hostmot2/llio_info.c:173:31: warning: initializer-string for array of ‘char’ truncates NUL terminator but destination lacks ‘nonstring’ attribute (9 chars into 8 available) [-Wunterminated-string-initialization]
  173 |                 .board_name = "MESA7C80",
      |                               ^~~~~~~~~~
hal/drivers/mesa-hostmot2/llio_info.c:184:31: warning: initializer-string for array of ‘char’ truncates NUL terminator but destination lacks ‘nonstring’ attribute (9 chars into 8 available) [-Wunterminated-string-initialization]
  184 |                 .board_name = "MESA7C81",
      |                               ^~~~~~~~~~
hal/drivers/mesa-hostmot2/llio_info.c:195:31: warning: initializer-string for array of ‘char’ truncates NUL terminator but destination lacks ‘nonstring’ attribute (9 chars into 8 available) [-Wunterminated-string-initialization]
  195 |                 .board_name = "MESA7I90",
      |                               ^~~~~~~~~~
hal/drivers/mesa-hostmot2/llio_info.c:206:31: warning: initializer-string for array of ‘char’ truncates NUL terminator but destination lacks ‘nonstring’ attribute (9 chars into 8 available) [-Wunterminated-string-initialization]
  206 |                 .board_name = "MESA7I43"
```
and indeed
```
hal/drivers/mesa-hostmot2/hostmot2.h:    rtapi_u8 board_name[8];  // ascii string, but not NULL terminated!
```
I suggest to follow the advice and add the https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#nonstring .

PR #3699, opened 2026-01-15
hostmot2 - board_name: Explicitly indicate __atribute__(nonstring)

smoe:
Addressing https://github.com/LinuxCNC/linuxcnc/issues/3698.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/hostmot2.h
+++ b/src/hal/drivers/mesa-hostmot2/hostmot2.h
@@ -145,7 +145,7 @@ typedef struct {
     rtapi_u32 idrom_type;
     rtapi_u32 offset_to_modules;
     rtapi_u32 offset_to_pin_desc;
-    rtapi_u8 board_name[8];  // ascii string, but not NULL terminated!
+    rtapi_u8 board_name[8] __attribute__ ((nonstring));  // ASCII string, but not NULL terminated!
     rtapi_u32 fpga_size;
     rtapi_u32 fpga_pins;
     rtapi_u32 io_ports;
--- a/src/hal/drivers/mesa-hostmot2/llio_info.c
+++ b/src/hal/drivers/mesa-hostmot2/llio_info.c
@@ -26,7 +26,7 @@
 #include "llio_info.h"
 
 typedef struct __info_entry_t {
-	char board_name[8];
+	char board_name[8] __attribute__ ((nonstring));
 	const char *base_name;
 	int num_ioport_connectors;
 	int pins_per_connector;


BsAtHome:
Fixing this in `hostmot2.h` may not be the right place. The warning comes from a local struct in the `llio_info.c` file. I think you need to fix this in `llio_info.c` line 29.

Now you mentioned the 'board' field and comparison, there are many references to (a different) 'board' field in other hostmot2 driver files and the comparisons are, well, a bit fragile. There are a combination of string and non-string comparisons spread out in the code (strncmp and memcmp) and I never fixed them all because it possibly broke things. The code to detect the different cards based on identifiers should be revamped not to use any string functions.

BsAtHome:
Why are you still changing `hostmot2.h`? Is that also giving warnings, #3698 did not indicate that? If not, then you should leave it alone until all of the board ID code gets fixed.

smoe:
> Fixing this in `hostmot2.h` may not be the right place. The warning comes from a local struct in the `llio_info.c` file. I think you need to fix this in `llio_info.c` line 29.

Thank you so much, yes, right, done.

> Now you mentioned the 'board' field and comparison, there are many references to (a different) 'board' field in other hostmot2 driver files and the comparisons are, well, a bit fragile. There are a combination of string and non-string comparisons spread out in the code (strncmp and memcmp) and I never fixed them all because it possibly broke things. The code to detect the different cards based on identifiers should be revamped not to use any string functions.

I skimmed through it. Would not know how to start or how to test. Somehow I happily leave this to a future generation of some AI coding tool :-)

BsAtHome:
> I skimmed through it. Would not know how to start or how to test. Somehow I happily leave this to a future generation of some AI coding tool :-)

Aie, aie, aie, yes, that will be a guarantee that no board will ever be found again...

I think @pcw-mesa needs to get involved here (if he wants to) and provide a /full/ list of all supported board identifiers (all 8 bytes for each board) so that they can be properly compared with a full 8-byte memcmp() call and mapped accordingly. Then the code can be fixed and streamlined for all hostmot2 board drivers. If there are any boards that use fewer than 8 ID bytes, with the remaining being random, then we need to fix the code of the comparisons/mappings a bit further.

The current comparisons are (can be) a bit fragile and I did not dare to touch previously it because I do not know the complete list of valid identifiers with their corresponding mapping(s).

pcw-mesa:
There are two distinct board name sources: One is the one embedded in the HM2ID ROM
These are used in PCI/PCIE/EPP/SPI cards. These are all 8 bytes long. The other name 
source is on Ethernet cards where the name is stored on an on card EEPROM. 
This is a normal null terminated string of up to 16 characters. These names are
often shortened when creating the hal name to make differing implementations
of the same card functionality use the same hal name (7I97/7I97T for example)

PR #3701, opened 2026-01-15
libnml: Fixed memleak in configuration parsing

smoe:
Also suppressing false positive report on use of undefined variable.

### DIFF
--- a/src/emc/usr_intf/axis/extensions/emcmodule.cc
+++ b/src/emc/usr_intf/axis/extensions/emcmodule.cc
@@ -2054,6 +2054,10 @@ static PyObject *pydraw_lines(PyObject * /*s*/, PyObject *o) {
             if(!first) glEnd();
             return NULL;
         }
+
+        // Suppress cppcheck false positive:
+        // 'first' == 1 when 'pl' is undefined and therefore not a problem.
+        // cppcheck-suppress uninitvar
         if(first || memcmp(p1, pl, sizeof(p1))
                 || (for_selection && n != nl)) {
             if(!first) glEnd();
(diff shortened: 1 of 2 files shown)


BsAtHome:
Instead of suppressing the cppcheck uninitialized warning, it may be appropriate just to initialize the `pl` variable like:
```C++
double p1[9], p2[9], pl[9] = {};
```

No comment on the mem-leak :-)

smoe:
> Instead of suppressing the cppcheck uninitialized warning, it may be appropriate just to initialize the `pl` variable like:
> 
> ```c++
> double p1[9], p2[9], pl[9] = {};
> ```

I would like to keep the suppression as a reminder that the code is a bit tricky to read. Also, erroneously assigning a default value 0 or undefined is equally unfortunate, so the suppression is something like a quick statement that the devs are aware of the situation.
 
> No comment on the mem-leak :-)

:)

BsAtHome:
> I would like to keep the suppression as a reminder that the code is a bit tricky to read.

Tricky,... that is an understatement.

> Also, erroneously assigning a default value 0 or undefined is equally unfortunate

That is not necessarily a problem. The real problem is that cppcheck cannot perform the correct analysis. The code and compound condition makes `first` and the value of `pl` mutually exclusive.

The cppcheck message is a false positive. If you suppress the message, then you should at least precede it with a comment why the message needs to be suppressed. That would help to understand the "tricky" code a bit better ;-)

smoe:
@BsAtHome, many thanks, it is indeed much nicer with the comment preceding, done. Broke the comment into two lines since it became a bit long (108 characters). @andypugh - ready?

Issue #966, opened 2020-10-21
Halcope - Can save data files but not open them


andypugh:
This is more an obvious enhancement than a bug. 

Halscope is a really useful tool, and it has the capability of saving data files. (albeit in a rather odd and inefficient format)

It would be very useful (for example when analysing other peoples plots from the mailing list and forum) to be able to load their log files back in to Halscope to zoom in and otherwise analyse. 

I think that there is probably no reason to maintain the current format. Just changing it to CSV (or maybe mf4?) would be an improvement in usability.

rodw-au:
I would love to see halscope and halstreamer both use csv files. Maybe there is a command line switch. I think CSV is the more broadly used format. In the past I have analysed data collected this way in Google sheets.

havardAasen:
If you are interested, I can submit a PR with a patch that saves the log files in csv format. I also made a python script to create plots, I found it easier then changing halscope.

I found the plot to be quite nice to work with, (interactive html). The only downside is that to run the python script it requires Bokeh library, and you can't 'apt-get' it. For me, the easiest was to install it with pip.

I attached a csv file created with halscope. The html file contains a interactive plot created with the python script. The plot is based on information from the documentation, HAL Tutorial - stepgen example.

GitHub didn't like html files... The dark theme is optional.
[halscope.html.csv](https://github.com/LinuxCNC/linuxcnc/files/6626192/halscope.html.csv)

[halscope_slow.csv](https://github.com/LinuxCNC/linuxcnc/files/6626190/halscope_slow.csv)

andypugh:
Saving .CSV would definitely be a start, but what I would really like to see Halscope do would be to import the files that it exports.
Then users on the forum could post their halscope data and we could look at it, and see what they see. 
(CSV would at least allow import in to Excel, but Excel is not great for interactive data plotting)

grandixximo:
This could be closed if PR  #3722 is merged

PR #3722, opened 2026-01-20
Halscope always 16 channels, and add user configurable samples

grandixximo:
The default memory has been increased, and user can set custom memory and restart to get less or more.

During the work on the jerk control have been using the scope a lot, this makes it a better tool for debugging.

### DIFF
--- a/src/hal/utils/scope.c
+++ b/src/hal/utils/scope.c
@@ -340,6 +340,33 @@ void start_capture(void)
 	/* already running! */
 	return;
     }
+
+    /* Check if data is from log file */
+    if (ctrl_usr->data_from_log_file) {
+        GtkWidget *dialog = gtk_message_dialog_new(
+            GTK_WINDOW(ctrl_usr->main_win),
+            GTK_DIALOG_MODAL,
+            GTK_MESSAGE_WARNING,
+            GTK_BUTTONS_OK_CANCEL,
+            _("Overwrite loaded log file data?"));
+
+        gtk_message_dialog_format_secondary_text(
+            GTK_MESSAGE_DIALOG(dialog),
+            _("Starting acquisition will clear the data loaded from the CSV file. Continue?"));
+
+        int response = gtk_dialog_run(GTK_DIALOG(dialog));
+        gtk_widget_destroy(dialog);
+
+        if (response != GTK_RESPONSE_OK) {
+            /* User cancelled, switch back to STOP mode */
+            set_run_mode(STOP);
+            return;
+        }
+
+        /* Clear the flag */
+        ctrl_usr->data_from_log_file = 0;
+    }
+
     for (n = 0; n < 16; n++) {
 	/* point to user space channel data */
 	chan = &(ctrl_usr->chan[n]);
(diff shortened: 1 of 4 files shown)


grandixximo:
I think I pulled at the wrong time, I will try to fix it

grandixximo:
Is retro compatibility important?
Like if switch back to 2.9, you want to open the 2.10 new halscope files, should that be supported?
I would need to add my new settings in a comment as the .halscope file, is that acceptable? or screw retro-compatibility?

2.10 can open 2.9 settings, but 2.9 will not understand 2.10 settings

grandixximo:
I implemented full bidirectional compatibility, so whomever switches branches won't be bothered by the changes

hansu:
What happens if 2.9 halscope reads a 2.10 halscope file? If it still reads all "old" parameters except the new ones it would be no problem. But I haven't looked into it if it may fail to read the file at all?

grandixximo:
> What happens if 2.9 halscope reads a 2.10 halscope file? If it still reads all "old" parameters except the new ones it would be no problem. But I haven't looked into it if it may fail to read the file at all?

Previous commit 2.9 would just not open the new files, probably cause I had the new config in the header, but I think it makes more sense to have it in the header as a comment which 2.10 can parse, and 2.9 ignores, already pushed the commit to fix this...

andypugh:
I am not sure that Halscope usefully opens files?
You can save-out the data files but AFAIK you can't load one in to analyse it. 
https://github.com/LinuxCNC/linuxcnc/issues/966
While you are at it.... :-)

hansu:
>I am not sure that Halscope usefully opens files?

It can open configuration files...

grandixximo:
On it

grandixximo:
Simple job for my stochastic parrot 🤭
I tested, everything seems in order, you can open a log and view it, even if you don't have the pins of the original logger, phantom log pins will be created for you, and forced to be enabled, the CVS also contains the configuration, so everything will look like what it did in the original log scaling and position, but it can be changed zoomed etc, rolling will clear the loaded log.

andypugh:
Thanks for this.

snowgoer540:
This seems in a few ways a step in the right direction for modern hardware, but it seems there are also a few issues...

1. On my system, no matter what I do, I cant get it to increase the number of samples.  It is locked at "Current: 32000 samples (2000 per channel)".  I've tried everything I can think of, it appends the autosave.halscope file correctly, but upon the next launch of halscope, the only thing that has changed is the "Samples (16 channels):" field.  It shows the value I saved, but it remains at "Current: 32000 samples (2000 per channel)".
2. I am a bit unsure why you did away with the chooser for the number of channels (and subsequently the samples per channel).  In the new version the number of Samples I choose is divided among all 16 channels, always, whether I use them or not.  So if I want 32000 samples on 1 channel. I have to make my Samples 512,000 (32000 * 16).  Is that the intent?  
3. When I launch halscope. I get these errors in the terminal window:

halscope: command not implemented: '# SAMPLES 32000'
halscope: config file 'autosave.halscope' caused 1 warnings

I _think_ (pretty sure) that is because in the function "read_config_file" in scope.c, there is no way to handle comments.  So it tries incorrectly to execute the command # SAMPLES 32000, when that line should be ignored.

grandixximo:
> This seems in a few ways a step in the right direction for modern hardware, but it seems there are also a few issues...
> 
>     1. On my system, no matter what I do, I cant get it to increase the number of samples.  It is locked at "Current: 32000 samples (2000 per channel)".  I've tried everything I can think of, it appends the autosave.halscope file correctly, but upon the next launch of halscope, the only thing that has changed is the "Samples (16 channels):" field.  It shows the value I saved, but it remains at "Current: 32000 samples (2000 per channel)".
> 
>     2. I am a bit unsure why you did away with the chooser for the number of channels (and subsequently the samples per channel).  In the new version the number of Samples I choose is divided among all 16 channels, always, whether I use them or not.  So if I want 32000 samples on 1 channel. I have to make my Samples 512,000 (32000 * 16).  Is that the intent?
> 
>     3. When I launch halscope. I get these errors in the terminal window:
> 
> 
> halscope: command not implemented: '# SAMPLES 32000' halscope: config file 'autosave.halscope' caused 1 warnings
> 
> I _think_ (pretty sure) that is because in the function "read_config_file" in scope.c, there is no way to handle comments. So it tries incorrectly to execute the command # SAMPLES 32000, when that line should be ignored.

Need restart linuxcnc to get bigger buffer.
I can ask user to input per channel memory, and then multiply internally by 16, I did away with choose how many channels, because the messages when you needed more channels were annoying.
For retro compatibility the new sample count is in a comment, but the counter effect is that comments are not really ignored anymore, can you share the halscope file? did you write comments manually?

snowgoer540:
> Need restart linuxcnc to get bigger buffer.

Wow, I feel a bit silly, but that was not apparent to me.  I took the message at face value: "The new sample count (512000) will take effect the next time halscope is started."  It seems it should be modified to say that linuxcnc needs to be restarted entirely.  

I do wonder why we need to restart though.  It seems like it's memory allocation, but is there a problem with just setting the necessary memory to a max?  In modern PCs this can only be a few Mb, right?

> I can ask user to input per channel memory, and then multiply internally by 16, I did away with choose how many channels, because the messages when you needed more channels were annoying. 
It was just a bit counter intuitive.  I think things might be better if restarting linuxcnc was not necessary.

>For retro compatibility the new sample count is in a comment, but the counter effect is that comments are not really ignored anymore, can you share the halscope file? did you write comments manually?

the contents of the file are just:

```
# SAMPLES 512000
MAXCHAN 16
THREAD servo-thread
HMULT 1
HZOOM 1
HPOS 5.000000e-01
TMODE 0
RMODE 0
```

I did make a quick modification to read_config_file to ignore lines with comments, and it made the aforementioned errors go away.  It's trying to pass the commented line to parse_command() and it does not like it.

snowgoer540:
I created this pull request that addresses the above:

https://github.com/LinuxCNC/linuxcnc/pull/3788

PR #3728, opened 2026-01-22
Gmoccapy offsetpage fix halpins and docs

Sigma1912:
Fix mismatched hal pin functionality and update the docs re bottom buttons/halpin function in 'Touch off mode'.

For discussion see 
https://github.com/LinuxCNC/linuxcnc/pull/3690#issuecomment-3782722012


hansu:
Thanks

PR #3730, opened 2026-01-22
Do not use xset when AXIS_NO_XSET is set

BsAtHome:
Having axis using `xset` to kill the key repeat wreaks havoc on the workflow when axis crashes or otherwise does not finish through the key repeat enable path.

This PR adds a simple environment variable `AXIS_NO_XSET` to disable messing with the key repeat. The default behaviour is not changed.

### DIFF
--- a/docs/src/gui/axis.adoc
+++ b/docs/src/gui/axis.adoc
@@ -611,6 +611,12 @@ full list of keyboard shortcuts can be found in the AXIS Quick
 Reference, which can be displayed by choosing Help > Quick Reference.
 Many of the shortcuts are unavailable when in MDI mode.
 
+The keyboard key repeat function is normally forcefully disabled in
+AXIS when LinuxCNC is started and forcefully re-enabled when you exit
+LinuxCNC. You can change this behavior by setting the environment
+variable 'AXIS_NO_XSET' to any value, causing AXIS to refrain from
+touching the key repeat settings.
+
 === Feed Override Keys
 
 [NOTE]
--- a/src/emc/usr_intf/axis/scripts/axis.py
+++ b/src/emc/usr_intf/axis/scripts/axis.py
@@ -120,7 +120,8 @@ inifile = linuxcnc.ini(sys.argv[2])
 ap = AxisPreferences()
 
 os.system("xhost -SI:localuser:gdm -SI:localuser:root > /dev/null 2>&1")
-os.system("xset r off")
+if None == os.getenv('AXIS_NO_XSET'):
+    os.system("xset r off")
 root_window = Tkinter.Tk(className="Axis")
 dpi_value = root_window.winfo_fpixels('1i')
 root_window.tk.call('tk', 'scaling', '-displayof', '.', dpi_value / 72.0)
@@ -154,7 +155,8 @@ except TclError:
 def General_Halt():
     text = _("Do you really want to close LinuxCNC?")
     if not root_window.tk.call("nf_dialog", ".error", _("Confirm Close"), text, "warning", 1, _("Yes"), _("No")):
-        os.system("xset r on")
+        if None == os.getenv('AXIS_NO_XSET'):
+            os.system("xset r on")
         root_window.destroy()
 
 root_window.protocol("WM_DELETE_WINDOW", General_Halt)


andypugh:
How noisy is xset? Would it be plausible to disable repeat at the start of a jog and re-enable on key-release? 

I think that the chosen INI entry name might not mean mush to most users. 
AXIS_NO_DISABLE_KEYREPEAT might be more meaningful.

BsAtHome:
`xset` isn't noisy. The problem is that re-enabling key repeat is <i>very</i> unstable. Calling `xset` repeatedly will only make things worse.
<sub>[I have about a crash about every 5..10 minutes debugging tcl9 and it freaks me out every time that axis interfered with the xserver settings]</sub>

It isn't an INI name. It is an environment variable. It is called AXIS_NO_XSET because I'm nearly sure someone will, at some stage, (ab)use `xset` for other global settings that will interfere with the xserver settings. It is a bad habit to change global settings in a local program.

An alternative not to use `xset` in axis at all. If you don't want key repeat, configure it using the standard window/desktop manager configuration utilities. The current situation affects the xserver config for all programs when axis is running. You cannot change windows and expect other behaviours anyway (like alt-tab'ing to an editor and expecting key repeat to work in there).

If you really want to disable key repeat on a <i>per application</i> basis, then you need to use the XKB interface and set XkbSetDetectableAutorepeat() (somewhere in the script with the equivalent call). If set, it will suppress key-release events until the physical key is released and you can track that for the appropriate keys that should not repeat for the application.

BsAtHome:
There is another way to detect key-repeat. Key press and release events have a timestamp associated with them. When a key repeat occurs, then the KeyRelease and KeyPress events have the <i>same</i> timestamp and can be detected that way. The following program is a simple demonstration, using the arrow-Up and the 'a' key as examples.

The only thing now is for someone to specify exactly which keys must not repeat and it can then be implemented quite easily without resorting to `xset` or any other xserver/xsession/global modifications.

```python
#!/usr/bin/env python3

import tkinter as tk

class Frame(tk.Frame):
    def __init__(self, master=None):
        super().__init__(master)

frame = Frame()
frame.master.title("Key-repeat tester")
frame.master.minsize(640, 480)

# We save the last time a key was released. If the following key is pressed at
# the same time, then X has generated a synthetic release/press combo and the
# event is a repeated key.
keySymState = {}

def f_up(event):
    global keySymState
    if event.keysym in keySymState and keySymState[event.keysym] == event.time:
        print("repeat:", event)
    else:
        print("press :", event)

def f_down(event):
    global keySymState
    keySymState[event.keysym] = event.time

frame.master.bind('<KeyPress-a>', f_up)
frame.master.bind('<KeyRelease-a>', f_down)
frame.master.bind('<KeyPress-Up>', f_up)
frame.master.bind('<KeyRelease-Up>', f_down)
frame.mainloop()
```

andypugh:
Key-repeat causes issues when keyboard-jogging in Axis. So I would have thought that the check could go in the jogging code rather than be mapped to specific keys.

BsAtHome:
I've just seen that some jog keys use start at press and stop at release:
```python
root_window.bind("<KeyPress-minus>", nomodifier(commands.jog_minus))
root_window.bind("<KeyPress-equal>", nomodifier(commands.jog_plus))
root_window.bind("<KeyRelease-minus>", commands.jog_stop)
root_window.bind("<KeyRelease-equal>", commands.jog_stop)
```
These cannot be handled without setting XkbSetDetectableAutorepeat(). However, Tcl has an example how to do that at https://wiki.tcl-lang.org/page/Disable+autorepeat+under+X11 and it looks quite easy.

BsAtHome:
I have a new version in my tree eliminating using `xset` in axis completely by adding `XkbSetDetectableAutoRepeat()` in a python module.
However,  I need #3714 and #3739 merged or it will be difficult to replace this PR with a new one due to potential merge conflicts and problems testing the new PR independently.

snowgoer540:
For what it's worth, we have the issue with autorepeat affecting Jogs in QtPlasmaC as well.  Despite most attempts Phill and myself made over the years to eliminate it, xset ended up being the solution that worked for everyone, always.  It seems the root of the issue is the interaction between X11 and Qt.  event.isAutoRepeat() returns False, even when it's True.  No amount of trying to force it seems to fix it.  I'm pulling from 5 years ago Greg's mind, but I seem to recall that it was acknowledged as an issue in qt, but has to my knowledge never been addressed.

I wonder if something similar goes on with tkinter as well, since some axis folks claim to have the issue, while others do not (that's how xset stuff ended up in there).  It'll be interesting to see if your solution has any merit with Qt, or is limited to tkinter.  

Admittedly using xset is annoying on my PC where I do development as anytime the qtplasmac window is open, autorepeat is off system wide.  I sort of deal with it, open the editor (which turns autorepeat back on), or close the window.  Otherwise I don't notice it because the computers that run the CNC stuff is used exclusively for CNC.  

Much like your original commit, we give the option for a user to disable it if they'd like.

BsAtHome:
Replaced by #3750.

Issue #3700, opened 2026-01-15
src/emc: cppcheck warns about null pointer derefences upon memory allocation failures


smoe:
I do not know how to react in such situations. Does the machine stop when the controller crashes? This may be worthwhile to state in the documentation if this is the case. In former times some aberrant java script in a browser was sufficient to drain all memory, so I do not find the scenario completely absurd to expect.

```
Checking src/emc/usr_intf/schedrmt.cc ...
src/emc/usr_intf/schedrmt.cc:1157:3: warning: If memory allocation fails, then there is a possible null pointer dereference: context [nullPointerOutOfMemory]
  context->cliSock = client_sockfd;
  ^
src/emc/usr_intf/schedrmt.cc:1156:41: note: Assuming allocation function fails
  context = (connectionRecType *) malloc(sizeof(connectionRecType));
                                        ^
src/emc/usr_intf/schedrmt.cc:1156:13: note: Assignment 'context=(struct connectionRecType*)std::malloc(sizeof(struct connectionRecType))', assigned value is 0
  context = (connectionRecType *) malloc(sizeof(connectionRecType));
            ^
src/emc/usr_intf/schedrmt.cc:1157:3: note: Null pointer dereference
  context->cliSock = client_sockfd;
  ^
src/emc/usr_intf/schedrmt.cc:1158:3: warning: If memory allocation fails, then there is a possible null pointer dereference: context [nullPointerOutOfMemory]
  context->linked = false;
  ^
src/emc/usr_intf/schedrmt.cc:1156:41: note: Assuming allocation function fails
  context = (connectionRecType *) malloc(sizeof(connectionRecType));
                                        ^
src/emc/usr_intf/schedrmt.cc:1156:13: note: Assignment 'context=(struct connectionRecType*)std::malloc(sizeof(struct connectionRecType))', assigned value is 0
  context = (connectionRecType *) malloc(sizeof(connectionRecType));
            ^
src/emc/usr_intf/schedrmt.cc:1158:3: note: Null pointer dereference
  context->linked = false;
  ^
src/emc/usr_intf/schedrmt.cc:1159:3: warning: If memory allocation fails, then there is a possible null pointer dereference: context [nullPointerOutOfMemory]
  context->echo = true;
  ^
src/emc/usr_intf/schedrmt.cc:1156:41: note: Assuming allocation function fails
  context = (connectionRecType *) malloc(sizeof(connectionRecType));
                                        ^
[...]

PR #3734, opened 2026-01-23
Emc race and cppcheck warnings

smoe:
Took the patch proposed by @BsAtHome for https://github.com/LinuxCNC/linuxcnc/issues/3700 and prepared this PR for it. It looks fine to me but I also did not test it. Any takers?

The most important bits are
 * elimination of dynamic memory allocation
 * the declaration of the sessions variable as atomic. It is decreased when the client can no longer be read (within a thread), and increased in sockMain when a new thread is started (outside a thread) and also when the return value of the thread is != 0 (outside a thread).

There are some meta talking points:
 * There is no pthread_join or pthread_detach anywhere.
 * Is the exit(0) upon an acceptance failure (the immediate stop of the controller with no explanation whatsoever) what we truly want?

### DIFF
--- a/src/emc/usr_intf/emcrsh.cc
+++ b/src/emc/usr_intf/emcrsh.cc
@@ -528,7 +528,7 @@ static int initSocket()
   server_address.sin_addr.s_addr = htonl(INADDR_ANY);
   server_address.sin_port = htons(port);
   server_len = sizeof(server_address);
-  err = bind(server_sockfd, (struct sockaddr *)&server_address, server_len);
+  err = bind(server_sockfd, reinterpret_cast<struct sockaddr*>(&server_address), server_len);
   if (err) {
       rcs_print_error("error initializing sockets: %s\n", strerror(errno));
       return err;
--- a/src/emc/usr_intf/schedrmt.cc
+++ b/src/emc/usr_intf/schedrmt.cc
@@ -328,7 +328,7 @@ static int initSockets()
   server_address.sin_addr.s_addr = htonl(INADDR_ANY);
   server_address.sin_port = htons(port);
   server_len = sizeof(server_address);
-  bind(server_sockfd, (struct sockaddr *)&server_address, server_len);
+  bind(server_sockfd, reinterpret_cast<struct sockaddr*>(&server_address), server_len);
   listen(server_sockfd, 5);
   signal(SIGCHLD, SIG_IGN);
   return 0;
@@ -1210,7 +1210,7 @@ void sockMain()
         int res = -1;
         struct sockaddr_in client_address;
         socklen_t client_len = sizeof(client_address);
-        int client_sockfd = accept(server_sockfd, (struct sockaddr *)&client_address, &client_len);
+        int client_sockfd = accept(server_sockfd, reinterpret_cast<struct sockaddr*>(&client_address), &client_len);
         if (client_sockfd < 0) {
             perror("sockMain: accept failed\n");
             exit(EXIT_FAILURE);


BsAtHome:
Good to see that you fixed the off-by-one in read() :-)

Seeing `pthread_join()` normally indicates that the application is well behaved. It clearly is not when `exit()` is called upon failure. The program is mostly written as a C program masked in a C++ source file. There are many things that should be streamlined, but I'm not sure whether this has any real value at this time. There are many interactions to consider, for example with NLM.
That said, the missing `pthread_join()` does make it necessary to add `pthread_detach()` after creation because the resources should be released automatically when a thread ends. It can be added as an else-clause to the `if(res != 0)` just after the thread is created.

There are also no allowances for signals causing interrupted syscalls. That would simply cause the application to bail and exit.

The `exit(0)` in failed `accept()` is interesting because it ignores all network errors that should be treated as `EAGAIN` (see accept(2)). Your machine will stop accepting (remote) connections when the network wobbles. Is this good/bad/whatever?

That brings us to a general LCNC issue: error handling. There are many more instances in LCNC programs and utilities that are not exactly well-behaved when we talk about error resilience and recovery. But that would require a complete review and well defined strategy how to cope.

rmu75:
Is this stuff even thread safe? It looks like it uses emcsched.cc, and addProgram there just uses push_back on a std::list... did I muss a lock?. Seems incorrect.

smoe:
I like to where this cppcheck warning brought us. The pthread_detach should always declare the pthread_t to be reallocatable after it returned, but correct and direct me.

At one of the very last lines, just prior to the invocation of sockMain, another pthread_create is performed. That should be detached, too, but I wait for the clarification on the position of this pthread_detach.

@rmu75, I also just skimmed through emcsched.cc and emcsched.hh - the queueStatus is an enum that may indeed immediately benefit from some thread safety, as in testing for a condition and changing its status should not be interrupted. And the addProgram changes the queue and then triggers a queue sort, which may come as a disturbing inconvenience to anyone iterating through the queue at the same time. Should we have this as a separate issue?

BsAtHome:
> Is this stuff even thread safe? ... Seems incorrect.

You are right! It is incorrect and not thread safe. There are probably many more problems with thread safety. These would not have been seen as the "normal" use is to attach only one remote client to send stuff.

Seen in this light, the choice is to limit to one client only (or use a select/poll construct to serialize access) or to review and fix the <i>entire</i> code path to be thread safe.

BsAtHome:
> I like to where this cppcheck warning brought us. The pthread_detach should always declare the pthread_t to be reallocatable after it returned, but correct and direct me.

A detached thread will have its resources purged when it exits. That means that the return value cannot be retrieved, for which you would have to use a join. In this case, it doesn't matter because no return value is used. Therefore, a simple detach should suffice.

> At one of the very last lines, just prior to the invocation of sockMain, another pthread_create is performed. That should be detached, too, but I wait for the clarification on the position of this pthread_detach.

That doesn't matter. That specific thread is the "queue runner" and there is only one during the lifetime of the program. Any resources associated with it will automatically die when the program dies.

smoe:
We have two separate threads already.  One is started in https://github.com/LinuxCNC/linuxcnc/blob/46c3b99c788adf3f8fb1610737305836ba1ce188/src/emc/usr_intf/schedrmt.cc#L1289 and the other from within the socksMain we are editing.  And both threads are changing the queue, one via parseCommand -> commandSet -> ... -> addProgram, the second via https://github.com/LinuxCNC/linuxcnc/blob/46c3b99c788adf3f8fb1610737305836ba1ce188/src/emc/usr_intf/emcsched.cc#L313.

smoe:
> That doesn't matter. That specific thread is the "queue runner" and there is only one during the lifetime of the program. Any resources associated with it will automatically die when the program dies.

Done it anyway, may eventually avoid false positive memory leaks in valgrind.

rmu75:
I think the likelihood that concurrency-problems are triggered in real application of schedrmt are very low and in this case it would be acceptable to just document the shortcomings. Introducing a global mutex that is held while processing data from read() should be possible and relatively easy. As no significant amounts of data are moved that should have no performance implications.

Proper way to fix this amounts to a rewrite with boost::asio or equivalent (which would also get rid of the threads btw) (or port to python).

BsAtHome:
With a global mutex, I think only invocation of `parseCommand()` in schedrmt.cc:1177 and `updateQueue()` in schedrmt.cc:1141 need to be protected.

rmu75:
with something like this

```c++

class MutexGuard {
    static pthread_mutex_t mutex_;
public:
    MutexGuard() { while (pthread_mutex_lock(&mutex_)) /* sleep? exit?*/; }
    ~MutexGuard() { pthread_mutex_unlock(&mutex_); }
};

pthread_mutex_t MutexGuard::mutex_ = PTHREAD_MUTEX_INITIALIZER;
```

it should be only a matter of creating a local MutexGuard object at the start of the functions to be mutex'ed. Don't pthread_exit though.

BsAtHome:
There are no thread-die-exit-or-else afaics. So just mutex the two functions should handle the problem.

Then, thisQuit() is called from a signal handler and uses exit(0). However, you should use _exit(0) when called from a signal handler.

BsAtHome:
@smoe, did you see my review comments?

smoe:
> @smoe, did you see my review comments?

Well, yes and no, had not mentally parsed the exit -> _exit one. This is all a bit beyond my daily routines - what is the signal meant to stop? I was wrong to remove the exit from the implementation of shutdown, from how interpret this all now, but what flavour of exit is appropriate when the shutdown code also frees everything?

smoe:
> With a global mutex, I think only invocation of `parseCommand()` in schedrmt.cc:1177 and `updateQueue()` in schedrmt.cc:1141 need to be protected.

Don't we need to expand any such global mutex also to emcsched.cc, etc?

rmu75:
I would keep the mutexes in the main loop of the threads. Blocking read returns data -> acquire mutex -> do stuff -> release mutex -> go back to blocking read of socket.

smoe:
@rmu75 , @BsAtHome , please kindly help me with a patch.

smoe:
> > With a global mutex, I think only invocation of `parseCommand()` in schedrmt.cc:1177 and `updateQueue()` in schedrmt.cc:1141 need to be protected.
> 
> Don't we need to expand any such global mutex also to emcsched.cc, etc?

I can answer that question for me now - no, we don't, since schedrmt is an application on its own, so there is no other route to invoke the same lower-level routines. Sorry, this seems rather obvious but somehow I needed that.

rmu75:
not really a patch and untested, but that should do it

line 1186
```C++
     else if (j > 0)
      {
        context->inBuf[j] = 0;
        MutexGuard mutex;
        if (parseCommand(context) == -1) goto finished;
        j = 0;
      }
```

line 1137
```C++
void *checkQueue(void * /*arg*/)
{
  while (1) {
    {
      MutexGuard mutex;
      updateQueue();
    }
    sleep((unsigned)pollDelay);
    }
  return 0;
}
```

rmu75:
:+1:  for using C++ std-facilities, I didn't think of that.

smoe:
I'll rebase to master, build and try this locally.

BsAtHome:
> I'll rebase to master, build and try this locally.

Have you checked out my <i>other</i> code review comments? There are two about constant strings and two about pthread_detach() being wrongly placed.

smoe:
> Have you checked out my other code review comments? There are two about constant strings and two about pthread_detach() being wrongly placed.

No, have not seen them. And I do no see them. Can you please provide them again?

BsAtHome:
They are created as "reviews". They should both be visible in this thread (somewhere higher up) and in the "Files changed" tab they appear in the diff.

BsAtHome:
FWIW...

Line 410:
There is no point of having am extra pointer to the array as a variable, You can simply use the pointer to the const char array. Therefore, you should instead change that to:
```C
  static const char runStr[] = "RUN";
  static const char stopStr[] = "STOP";
  static const char pauseStr[] = "PAUSE";
  static const char resumeStr[] = "RESUME";
```

Line 1106:
Same as other comment, no point of having yet another pointer indirection. Should change to:
```C
  static const char helloNakStr[] = "HELLO NAK\r\n";
  static const char shutdownNakStr[] = "SHUTDOWN NAK\r\n";
  static const char helloAckStr[] = "HELLO ACK %s 1.1\r\n";
  static const char setNakStr[] = "SET NAK\r\n";
```

Line 1229:
The pthread_detach must be as an else-clause of the test 'res != 0' because you cannot detach a thread that did not get created.

Line 1306:
You can't detach updateThread if thread creation failed. There is a test on pthread_create's result one line down and the detach must be conditional on pthread_create to succeed.

BsAtHome:
Wouldn't it be more appropriate to put the `std::lock_guard<std::mutex> lck(queue_mtx);` from line 1192 to be at the start of function `parseCommand()` just after the open brace at line 1102?
The current position has a goto to exit the block, which, although it works, is not nice to look at. The lock solely protects `parseCommand()` and therefore starting that function with the lock may be "cleaner".

smoe:
> Wouldn't it be more appropriate to put the `std::lock_guard<std::mutex> lck(queue_mtx);` from line 1192 to be at the start of function `parseCommand()` just after the open brace at line 1102? The current position has a goto to exit the block, which, although it works, is not nice to look at. The lock solely protects `parseCommand()` and therefore starting that function with the lock may be "cleaner".

Yes. I also prefer locking parseCommand, indeed. It took me a while to understand you wanted, so I eliminated that goto first :-)  Not sure if I prefer the goto, though.

BsAtHome:
> Yes. I also prefer locking parseCommand, indeed. It took me a while to understand you wanted, so I eliminated that goto first :-) Not sure if I prefer the goto, though.

I don't have problems with goto. It is, after all, how we get to where we want to be, eventually ;-)
But having it break the loop is fine too and is effectively identical.

However, I'd place the guard mutex <i>before</i> the strtok and snprintf call (actually, I'd rather place it after the opening brace of the function before the local variables, as I said previously, but that may be a bit of nitpicking).

smoe:
> However, I'd place the guard mutex before the strtok and snprintf call (actually, I'd rather place it after the opening brace of the function before the local variables, as I said previously, but that may be a bit of nitpicking).

My  "concern" was that the lock should preferably only lock what needs to be locked. Those initially string processing bit I happily granted it access.

BsAtHome:
> > However, I'd place the guard mutex before the strtok and snprintf call (actually, I'd rather place it after the opening brace of the function before the local variables, as I said previously, but that may be a bit of nitpicking).
> 
> My "concern" was that the lock should preferably only lock what needs to be locked. Those initially string processing bit I happily granted it access.

But why then not <i>inside</i> the `if(pch != NULL) { ...lock.this.block... }` block? You'd forego locking entirely if the lock is not necessary.

smoe:
> > > However, I'd place the guard mutex before the strtok and snprintf call (actually, I'd rather place it after the opening brace of the function before the local variables, as I said previously, but that may be a bit of nitpicking).
> > 
> > 
> > My "concern" was that the lock should preferably only lock what needs to be locked. Those initially string processing bit I happily granted it access.
> 
> But why then not inside the `if(pch != NULL) { ...lock.this.block... }` block? You'd forego locking entirely if the lock is not necessary.

Done.

rmu75:
my intention to keep the locks in the main thread function was that it is somewhat more obvious who is holding a lock. if the checkQueue function were to call parseCommand there would be a problem and would need `std::recursive_mutex`.

rmu75:
> Yes. I also prefer locking parseCommand, indeed. It took me a while to understand you wanted, so I eliminated that goto first :-) Not sure if I prefer the goto, though.

nitpicking: I don't think the goto there is unclean, but if you eliminate it, then you should also remove the other goto (and the label).

smoe:
> > Yes. I also prefer locking parseCommand, indeed. It took me a while to understand you wanted, so I eliminated that goto first :-) Not sure if I prefer the goto, though.
> 
> nitpicking: I don't think the goto there is unclean, but if you eliminate it, then you should also remove the other goto (and the label).

Yes, had it under my fingertips, but this would introduce a non-natural use of ret or extend the scope of j and so somehow I liked it better with that goto remaining.

smoe:
Worse: schedrmt needs -h and --help as arguments.

BsAtHome:
You removed the `char` from `const char`
You added a size parameter to a `strcmp` call but there are only two arguments to strcmp. Did you want to strncmp? If so, why? You are adding a false positive compare when the source is longer than the compare value. And, the source is from the network.

BsAtHome:
Just to add why using strncmp is inappropriate.

If you compare _some buffer_ with a constant string, as is the case here, then you will **always** terminate the compare at most at the sizeof the constant string because it has a guaranteed terminating NUL character. When you use the _some buffer_ size, then you may end up with a false positive in the comparison is that size is less than the size of the constant string.

Consider this example:
```C
#include <stdio.h>
#include <string.h>
char a[] = "AAA";
char b[] = "AAB";
int main(void)
{
	printf("% d\n", strcmp(a, b));
	printf("% d\n", strncmp(a, b, sizeof(b)));
	printf("% d\n", strncmp(a, b, 2));
	return 0;
}
```

It will print:
```
-1
-1
 0
```
That means: the two strings are considered equal when you limit the length of the comparison. However, the strings are **not** the same.

If `a` is a generic buffer and `b` is a constant string, then you can still create a buffer overrun by using `strncmp(a, b, sizeof(b))` when the size of `a` is less than the size of `b`.
If you use `strncmp(a, b, sizeof(a))` and the size of `a` is less than the size of `b`, then you prevent any buffer overrun but may generate a false positive in the compare.

The use case in schedrmt.cc is that the buffer is always prepared from a strtok() call. This ensures that the buffer is always NUL terminated and there is no need to limit the string compare in length. Therefore, simple strcmp() is adequate and appropriate and the only correct thing to do.

BsAtHome:
And that reminds me... strtok() is not thread safe. It may need to be replaced with strtok_r().

rmu75:
> And that reminds me... strtok() is not thread safe. It may need to be replaced with strtok_r().

why not leave the lock outside, nobody cares about performance or a maximum of concurrence in this app. This should be a correctness fix of an app that happens to be there and (probably) nobody uses.

smoe:
> And that reminds me... strtok() is not thread safe. It may need to be replaced with strtok_r().

For this comment alone, me messing up with that strncmp was worth it. Thank you.

Somewhat overly excited while commuting I went into some immediate robot work to transition the strtok to strtok_r, but reality caught up quickly.

rmu75:
> > And that reminds me... strtok() is not thread safe. It may need to be replaced with strtok_r().
> Somewhat overly excited while commuting I went into some immediate robot work to transition the strtok to strtok_r, but reality caught up quickly.

just put the mutex in the top level functions of the thread so no strtok can happen in parallel.

BsAtHome:
> > And that reminds me... strtok() is not thread safe. It may need to be replaced with strtok_r().

> why not leave the lock outside, nobody cares about performance or a maximum of concurrence in this app. This should be a correctness fix of an app that happens to be there and (probably) nobody uses.

Fully agree! It is a pain to convert to strtok_r because you suddenly need to pass the save pointer around.

It is much, much easier just to move the lock guard to above the strtok in parseCommand().

smoe:
> > And that reminds me... strtok() is not thread safe. It may need to be replaced with strtok_r().
> 
> why not leave the lock outside, nobody cares about performance or a maximum of concurrence in this app. This should be a correctness fix of an app that happens to be there and (probably) nobody uses.

I should move the lock to the beginning of parseCommand just like @BsAtHome suggested upfront since this eliminates the need for strtok_r (and passing an extra char **pointer to all the command handlers).

Concerning the interest in these remote execution interfaces, I guess that this will soon rise, anticipating some remote management of multiple LinuxCNC machines and/or collaborating machines.

Could you all please also have a quick look at emcrsh.cc? Do you see any chance / motivation to share some code between emcrsh and schedrmt?  Cppcheck thinks so:
```
$ cppcheck *.cc
Checking emclcd.cc ...
1/7 files checked 11% done
Checking emcrsh.cc ...
2/7 files checked 33% done
Checking emcsched.cc ...
3/7 files checked 36% done
Checking emcsh.cc ...
4/7 files checked 59% done
Checking halui.cc ...
Checking halui.cc: TOOL_NML...
Checking halui.cc: __GNUC__...
5/7 files checked 81% done
Checking schedrmt.cc ...
6/7 files checked 92% done
Checking shcom.cc ...
7/7 files checked 100% done
emcrsh.cc:426:9: error: The one definition rule is violated, different classes/structs have the same name 'connectionRecType' [ctuOneDefinitionRuleViolation]
typedef struct {
        ^
schedrmt.cc:241:9: note: The one definition rule is violated, different classes/structs have the same name 'connectionRecType'
typedef struct {
        ^
emcrsh.cc:426:9: note: The one definition rule is violated, different classes/structs have the same name 'connectionRecType'
typedef struct {
```

BsAtHome:
Please run it with the LCNC cppcheck wrapper `scripts/cppcheck.sh`. There are several setup thingies in there.
```sh
$ cd src
$ ../scripts/cppcheck.sh emc/usr_intf/*.cc
```
Edit: you need to run it from the src directory

BsAtHome:
BTW, do you actually try to compile this locally <i>before</i> you commit changes? It doesn't seem so considering the basic errors you get in CI. Or is this just AI slop we're commenting on?

BsAtHome:
Why exactly are you const'ifying the arguments? Is the context pointer prone to be misinterpreted and changed in inappropriate ways? Can you please state your case for dong this?

On a side note: If there ever was any case for using references, this would be it. That would be a much better solution than to make complex const constructs.

But, these type of changes are rather irrelevant, do not really improve the code and do not fix real bugs. If you really want to fix the program to be top in class, then you should rewrite the <i>entire</i> code (also mainly because it uses C thinking in a C++ disguise).

smoe:
> Why exactly are you const'ifying the arguments? Is the context pointer prone to be misinterpreted and changed in inappropriate ways? Can you please state your case for dong this?

I liked it. It reflects in parts how I think.

> On a side note: If there ever was any case for using references, this would be it. That would be a much better solution than to make complex const constructs.

I like references better, indeed.

> But, these type of changes are rather irrelevant, do not really improve the code and do not fix real bugs. If you really want to fix the program to be top in class, then you should rewrite the entire code (also mainly because it uses C thinking in a C++ disguise).

Maybe that is something I would indeed want to do.

smoe:
> Please run it with the LCNC cppcheck wrapper `scripts/cppcheck.sh`. There are several setup thingies in there.
> 
> ```shell
> $ cd src
> $ ../scripts/cppcheck.sh emc/usr_intf/*.cc
> ```

This points to 

```
emc/usr_intf/emcrsh.cc:530:29: warning: Potentially invalid type conversion in old-style C cast, clarify/fix with C++ cast [dangerousTypeCast]
  err = bind(server_sockfd, (struct sockaddr *)&server_address, server_len);
                            ^
emc/usr_intf/schedrmt.cc:331:23: warning: Potentially invalid type conversion in old-style C cast, clarify/fix with C++ cast [dangerousTypeCast]
  bind(server_sockfd, (struct sockaddr *)&server_address, server_len);
                      ^
emc/usr_intf/schedrmt.cc:1213:51: warning: Potentially invalid type conversion in old-style C cast, clarify/fix with C++ cast [dangerousTypeCast]
        int client_sockfd = accept(server_sockfd, (struct sockaddr *)&client_address, &client_len);
```

which I would then address with `reinterpret_cast<sockaddr*>` unless suggested otherwise. 

If the introduced "const"s for the context are seriously frowned upon then I would remove them. Transitioning to references would come with quite another set of changes.

BsAtHome:
> emc/usr_intf/emcrsh.cc:530:29: warning: Potentially invalid type conversion in old-style C cast, clarify/fix with C++ cast
> ... which I would then address with `reinterpret_cast<sockaddr*>` unless suggested otherwise.

Indeed, for C++ you do need that change because cppcheck is picky.

> If the introduced "const"s for the context are seriously frowned upon then I would remove them. Transitioning to references would come with quite another set of changes.

I frown upon them (and specifically in the context of fixing these programs). They make no real improvement and just make things look complex. The correct change would be to convert to references.

rmu75:
I don't understand the mutex decl in emcrsh. If you want to backport the scheduling-stuff, I suggest to split off changes to emcrsh, the remote shell got a bunch of updates since 2.9

smoe:
> I don't understand the mutex decl in emcrsh. If you want to backport the scheduling-stuff, I suggest to split off changes to emcrsh, the remote shell got a bunch of updates since 2.9

The parseCommand also uses strtok, so we should not allow any reentry by a separate thread. Here the diff of emcrsh to master:

```diff
--- a/src/emc/usr_intf/emcrsh.cc
+++ b/src/emc/usr_intf/emcrsh.cc
@@ -22,6 +22,8 @@
 #include <sys/socket.h>
 #include <netinet/in.h>
 #include <getopt.h>
+#include <atomic>              // for sessions counter
+#include <mutex>               // parseCommand is not reentrant

 #include "emcglb.h"            // EMC_NMLFILE, TRAJ_MAX_VELOCITY, etc.
 #include "inifile.hh"          // INIFILE
@@ -449,7 +451,7 @@ int enabledConn = -1;
 char pwd[16] = "EMC\0";
 char enablePWD[16] = "EMCTOO\0";
 char serverName[24] = "EMCNETSVR\0";
-int sessions = 0;
+std::atomic_int sessions = 0;
 int maxSessions = -1;

 const char *cmdTokens[] = {
@@ -479,6 +481,8 @@ struct option longopts[] = {
   {"path", 1, NULL, 'd'},
   {0,0,0,0}};

+std::mutex queue_mtx; // controls access to queue
+
 /* format string to outputbuffer (will be presented to user as result of command) */
 #define OUT(...) snprintf(context->outBuf, sizeof(context->outBuf), __VA_ARGS__)

@@ -524,7 +528,7 @@ static int initSocket()
   server_address.sin_addr.s_addr = htonl(INADDR_ANY);
   server_address.sin_port = htons(port);
   server_len = sizeof(server_address);
-  err = bind(server_sockfd, (struct sockaddr *)&server_address, server_len);
+  err = bind(server_sockfd, reinterpret_cast<struct sockaddr*>(&server_address), server_len);
   if (err) {
       rcs_print_error("error initializing sockets: %s\n", strerror(errno));
       return err;
@@ -2917,6 +2921,8 @@ cmdType lookupCommand(char *s)
 // handle the linuxcncrsh command in context->inBuf
 int parseCommand(connectionRecType *context)
 {
+  std::lock_guard<std::mutex> lck(queue_mtx); // Only one thread enters this function at a time.
+                                              // Caveat: Change strtok to strtok_r if this ever changes.
   int ret = 0;
   char *cmdStr;

@@ -3041,15 +3047,9 @@ int sockMain()
         sessions++;
         // enforce limited amount of clients that can connect simultaneously
         if ((maxSessions == -1) || (sessions <= maxSessions)) {
-            pthread_t *thrd;
+            pthread_t thrd;
             connectionRecType *context;

-            thrd = (pthread_t *)calloc(1, sizeof(pthread_t));
-            if (!thrd) {
-                fprintf(stderr, "linuxcncrsh: out of memory\n");
-                exit(1);
-            }
-
             context = (connectionRecType *)malloc(sizeof(connectionRecType));
             if (!context) {
                 fprintf(stderr, "linuxcncrsh: out of memory\n");
@@ -3067,7 +3067,10 @@ int sockMain()
             context->commProt = 0;
             context->inBuf[0] = 0;

-            res = pthread_create(thrd, NULL, readClient, (void *)context);
+            res = pthread_create(&thrd, NULL, readClient, (void *)context);
+            if (!res) {
+                pthread_detach(thrd);
+            }
         } else {
             fprintf(stderr, "linuxcncrsh: maximum amount of sessions exceeded: %d\n", maxSessions);
             res = -1;
```

smoe:
No more to add from my side other than testing.

andypugh:
I have something of a mental block with C++ so any review of the code by me would be pointless. 
I know that emcrsh is used by several people, and presumably that's just a subset of all the users.

rmu75:
I was a bit confused about the scope of this PR when emcrsh entered the picture.
I checked manually, the changes to emcrsh should also apply to 2.9 (despite there having been many improvements to emcrsh since)

smoe:
> I was a bit confused about the scope of this PR when emcrsh entered the picture.

That is because of thread safety conflating with the initial PR in https://github.com/LinuxCNC/linuxcnc/pull/3734#issuecomment-3790280470 .
We discussed this rather intensively and so I also applied that mutex to the neighbouring code that is very similar in its structure.

And missing pthread_{join,detach} was introduced to this PR in https://github.com/LinuxCNC/linuxcnc/pull/3734#issuecomment-3790029863

A pthread_detach seems to be also missing for
https://github.com/LinuxCNC/linuxcnc/blob/5814ae203aef0600513ca7d110efacd620e49e1b/src/hal/classicladder/socket_server.c#L129
https://github.com/LinuxCNC/linuxcnc/blob/5814ae203aef0600513ca7d110efacd620e49e1b/src/hal/classicladder/socket_modbus_master.c#L107
https://github.com/LinuxCNC/linuxcnc/blob/5814ae203aef0600513ca7d110efacd620e49e1b/src/libnml/cms/tcp_srv.cc#L1045 (but needs some thinking)
https://github.com/LinuxCNC/linuxcnc/blob/5814ae203aef0600513ca7d110efacd620e49e1b/src/hal/utils/halrmt.c#L3472

which may indeed be preferable to address in separate, individual PRs.

I never used strtok myself before, which was introduced to this PR in https://github.com/LinuxCNC/linuxcnc/pull/3734#issuecomment-3808238183
which has an internal state remembering where it was last active so it is passed a NULL in subsequent invocations on the same string. So if my newly acquired insights serve me right, only one single strtok invocation can be executed at the very same time for all threads, and the variant strtok_r now demands that variable to be contributed externally so that problem is gone.

Suspicious invocations of strtok I found in
https://github.com/LinuxCNC/linuxcnc/blob/5814ae203aef0600513ca7d110efacd620e49e1b/src/emc/tooldata/tooldata_db.cc#L244
https://github.com/LinuxCNC/linuxcnc/blob/5814ae203aef0600513ca7d110efacd620e49e1b/src/emc/rs274ngc/rs274ngc_pre.cc#L1003
https://github.com/LinuxCNC/linuxcnc/blob/5814ae203aef0600513ca7d110efacd620e49e1b/src/emc/rs274ngc/interp_remap.cc#L424
https://github.com/LinuxCNC/linuxcnc/blob/5814ae203aef0600513ca7d110efacd620e49e1b/src/hal/user_comps/mb2hal/mb2hal_init.c#L141
https://github.com/LinuxCNC/linuxcnc/blob/501f44ad7b9939cade275788d3c8ba7d3bd502df/src/emc/tooldata/tooldata_common.cc#L96
... others ...
which do not look overly complicated to transition to strtok_r. I would make this a separate strtok_r PR, though.

So, yes, this PR got utterly sidetracked but I loved it and thank @BsAtHome and @rmu75 for training me on it.

> I checked manually, the changes to emcrsh should also apply to 2.9 (despite there having been many improvements to emcrsh since)

I would not want to introduce mutexes into a stable version without some serious testing. Maybe it would even be preferable to leave this in 2.9 as it is and have that extra (then tested) safety as a feature for 2.10.

smoe:
Rebased to master.

rmu75:
> Suspicious invocations of strtok I found in
> 
> https://github.com/LinuxCNC/linuxcnc/blob/5814ae203aef0600513ca7d110efacd620e49e1b/src/emc/tooldata/tooldata_db.cc#L244
> 
> https://github.com/LinuxCNC/linuxcnc/blob/5814ae203aef0600513ca7d110efacd620e49e1b/src/emc/rs274ngc/rs274ngc_pre.cc#L1003
> 
> https://github.com/LinuxCNC/linuxcnc/blob/5814ae203aef0600513ca7d110efacd620e49e1b/src/emc/rs274ngc/interp_remap.cc#L424
> 
> https://github.com/LinuxCNC/linuxcnc/blob/5814ae203aef0600513ca7d110efacd620e49e1b/src/hal/user_comps/mb2hal/mb2hal_init.c#L141
> 
> https://github.com/LinuxCNC/linuxcnc/blob/501f44ad7b9939cade275788d3c8ba7d3bd502df/src/emc/tooldata/tooldata_common.cc#L96
> 
> ... others ...
> which do not look overly complicated to transition to strtok_r. I would make this a separate strtok_r PR, though.

I suggest to leave that stuff alone. interp isn't thread safe anyways and can't be made without extensive rearchitecture of the canon interface.

> > I checked manually, the changes to emcrsh should also apply to 2.9 (despite there having been many improvements to emcrsh since)
> 
> I would not want to introduce mutexes into a stable version without some serious testing. Maybe it would even be preferable to leave this in 2.9 as it is and have that extra (then tested) safety as a feature for 2.10.

then we should remove the label "2.9-must-fix".

smoe:
Hm. I admit not to know about how many threads there are executed in parallel that possibly use strtok, but the problem is real, as in this little test program
```c
#include <stdio.h>  // printf
#include <string.h> // strtok
#include <pthread.h>

const char delim[]=" ,?";
const int num=500;

void * f(void*) {
    for(int i=0; i< num; i++) {
        char a[]="hello world, how are you?";
        char *s1 = strtok(a,delim);
        printf("1: %s\n", s1);
        char *s2 = strtok(NULL,delim);
        printf("2: %s\n", s2);
        char *s3 = strtok(NULL,delim);
        printf("3: %s\n", s3);
    }
    return(nullptr);
}

void * g(void*) {
    char *safeptr = nullptr;
    for(int i=0; i< num; i++) {
        char a[]="hello world, how are you?";
        safeptr = NULL;
        char *s1 = strtok_r(a,   delim,&safeptr);
        printf("1: %s\n", s1);
        char *s2 = strtok_r(NULL,delim,&safeptr);
        printf("2: %s\n", s2);
        char* s3 = strtok_r(NULL,delim,&safeptr);
        printf("3: %s\n", s3);
    }
    return(nullptr);
}

int main() {
    for(int i=0; i<100; i++) {
        pthread_t t;
        int res = pthread_create(&t,NULL,g,NULL);
        if (res) {
           fprintf(stderr,"E: %d -> %d\n", i, res);
        } else {
           pthread_detach(t);
        }
    }
}
```
produces with strtok (function f)
```
$ ./t | sort | uniq -c
   2464 1: hello
    203 2: are
     48 2: are you?
    379 2: how
     31 2: how are you?
    395 2: (null)
   1197 2: world
     29 2: world, how are you?
    120 2: you
     29 2: you?
    288 3: are
      1 3: are you
     39 3: are you?
    766 3: how
     36 3: how are you?
    482 3: (null)
    573 3: world
      1 3: world, how are you
     32 3: world, how are you?
    154 3: you
     23 3: you?
```
and with strtok_r the expected (did not use join to wait, so numbers are not equal)
```
$ ./t | sort | uniq -c
   2798 1: hello
   2769 2: world
   2737 3: how
```

and certainly what LinuxCNC is doing is not threadsafe in the sense that it may not be prohibited that the tool table may be edited simultaneously to some thread reading it because some locking mechanism not implemented. What we have here is that someone pressing the button to reread the tool table may lead to some complete nonsense at some other part of LinuxCNC that is auto-updating the contents of an .ini file or whatever. Completely unrelated so it cannot be addressed by protecting critical areas with semaphores. And dangerous. Just because of those routines are using strtok instead of strtok_r..
 
Since I am for some weird reason enjoying this, I suggest you just let me perform those strtok to strtok_r transitions, and it shall not be to the disadvantage of LinuxCNC.

rmu75:
It is out of the question that strtok is not thread-safe and concurrent use will mess up the save-pointer or worse. In general, internal linuxcnc interfaces are not thread safe anyway, strtok is not the only problem.

All stuff that is written in python is implicitly mutexed by the global interpreter lock.

The interpreter canon interface has no way to identify any context of a calling interpreter, so it really doesn't make sense to run two interpreters in one process at the same time.

IME concurrency problems do show up in program usage rather sooner than later. The only reason no problems are known in emcrsh and emcsched is because nobody is/was using those with multiple connections doing stuff at the same time.

rmu75:
Editing and reloading tool table in the same process at the same time from two different threads seems a rather theoretical possibility, why would anybody do that and expect it to work? It won't make sense in any case. Also all stuff that goes through NML also is serialized auomatically, messages are put into a message queue, and one giant semi-endless loop and corresponding switch statement is processing those messages, no danger of anything happening concurrently there.

BsAtHome:
The question about thread safety is of course a serious one. As said, there are many places that are not thread safe by design or implementation and we have not (yet) seen disasters because most user/usage patterns do not exploit parallelisms that would expose the races.

It should be a long term goal to update and fix all programs to adhere to proper thread safe standards and fix things. But I see no reason to expedite this update process at this moment unless there is a use case that demands it or while we are fixing a particular program anyway. For now, fixing as we get along should be fine.

rmu75:
Fixing things is of course OK and necessary but we should not begin to redesign stuff to become threadsafe/reentrant just for the heck of it or making stuff multi-threaded needlessly, just the opposite, stuff like emcrsh should be redone to not neet multiple threads IMHO.

BsAtHome:
Well, I did previously suggest to use a simple select/poll loop to make it simpler.

smoe:
There is some subtle distinction between multithreading (what in my limited understanding we are mostly discussing) and functions being reentrant (which strtok is not but strtok_r is).

I'll prepare the strtok->strtok_r transition against 2.9. And I am happily anticipating your select/poll loop.

BsAtHome:
> I'll prepare the strtok->strtok_r transition against 2.9. And I am happily anticipating your select/poll loop.

Working on it :-)

smoe:
> > I'll prepare the strtok->strtok_r transition against 2.9. And I am happily anticipating your select/poll loop.
> 
> Working on it :-)

The select/poll loop, so I hope :-) Have the strtok_r transition prepared locally.

BsAtHome:
Actually,... I've been doing a rewrite on linuxcncrsh (well, emcrsh.cc) making it single threaded and more of a C++ program. It started out as a small patch, but I went a bit further when I found all the off-by-one and buffer problems. Oh, and strtok is now also history. It is almost done. Currently testing it, smoothing things out and fix inconsistencies.

That whole process uncovered a lot more of funny stuff that may be necessary to do in schedrmt.cc too.

smoe:
Do you have ideas for including your (some of your) tests as part of the CI ?

BsAtHome:
> Do you have ideas for including your (some of your) tests as part of the CI ?

Not yet, but yes, they need to be added.

BsAtHome:
An update on the linuxcncrsh rewrite: Most code now just works. I've also re-enabled the test in `test/linuxcncrsh` and fixed a lot of errors in the test itself.

The idea that you can make linuxcncrsh single threaded is _not_ supported by the code. The problem is in the fact that NML will stall the entire process when you run synchronously (WAIT_MODE DONE). That is a problem and can only be solved by having the session (i.e. a connected socket) run in its own thread.

The next thing is that linuxcncrsh _**mostly is**_ single threaded because it uses an `enableConn` variable tracking which session is allowed to talk to NML (which, of course, currently has a race condition). That means that only _**one**_ session can talk to the machine at any time. That session is the one which successfully executed `SET ENABLE ...`.

Now, a real question is whether it makes sense, at all, to have multiple sessions to linuxcncrsh that can control a single machine. There may be scenarios, but I'd need someone to tell me which scenario that would be. However, if so, then this probably needs to me multiplexed onto one single NML connection, but that is  a speculation I have at this time.
It would be easier to have only one session controlling the machine and then, if the interface allows it, allow for other session only to observe some few things. Or maybe you may want to kick one (stuck) session from another session (but may lead to other problems).

BTW, there is no guard against instantiating linuxcncrsh multiple times on different ports and connecting a session from each instance. I have no idea how that may lead to catastrophic machine pulp, broken iron, lost limbs and other unfortunate faults.

Interactions of multiple sessions needs to be discussed how they are to be handled.


FWIW, I'm currently stuck on issuing `SET JOG_STOP`. It apparently aborts the entire test process and I have no clue why.

smoe:
To help Andy a bit with his decision making towards accepting your work:
* Is anything working less reliably than in the prior implementation? You mention SET JOG_STOP as your end boss.
* To be on the safe side when a second access is requested: A warning that linuxcncrsh is already in use and terminate?

And .. maybe https://linuxcnc.org/docs/html/man/man1/linuxcncrsh.1.html is not meant to be ultimately safe in the first place? It can be started in parallel to a GUI, so you already can unknowingly give contradictory (limb-cutting) commands to what the machine is doing. It is a bit funny for us to care about thread-safety when that same thread-safe tool may indeed ruin any real-live thread (and limb) at ease when interfering with a running program?!

BsAtHome:
> Is anything working less reliably than in the prior implementation?

No, it is working much better now :-)
Well, it is not yet finished entirely, but the framework is there and most commands perform as they should. A lot of extra checks have also been implemented to tell the user about invalid input.

It is just that I've been digging and discovered all these issues as I went along. Many I could solve without discussion, but some need discussion and this is the time to ask some of those questions and share reflections.

The test in tests/linuxcncrsh-tcp works just fine. I am working on getting the test in tests/linuxcncrsh to work because it is much more comprehensive (currently disabled in master).

> You mention SET JOG_STOP as your end boss.

Yeah, that is a very odd situation. It appears as if something in LCNC terminates when the command is issued and that wraps it all up. Need to do some more digging.

It reminds me a bit of the failure of multiple spindles I was tracing. It took me a whole day to realize that the wrong hal-file was referenced in the INI. And the INI has the SPINDLES variable in the wrong place. Then, homing was again a problem. It needs a small sleep pause before issuing the next home command. Otherwise the homing process is still running and you get an error.

> To be on the safe side when a second access is requested: A warning that linuxcncrsh is already in use and terminate?

I'd not have it terminate immediately. There is an argument for being able to kill a session from another session. Whether that can be done safely and with valid machine-state needs to be determined. There already is a limited set of available commands when not in "enable" mode. This can be useful for more things.

> And .. maybe linuxcncrsh is not meant to be ultimately safe in the first place? It can be started in parallel to a GUI [snip]

Indeed, you have a very valid point. OTOH, it would be nice to prevent some obvious risk or at least warn about it. You can rebuild a crashed machine. Regrowing limbs is not (yet) common practice ;-)

rmu75:
> Now, a real question is whether it makes sense, at all, to have multiple sessions to linuxcncrsh that can control a single machine. There may be scenarios, but I'd need someone to tell me which scenario that would be. However, if so, then this probably needs to me multiplexed onto one single NML connection, but that is a speculation I have at this time.
> It would be easier to have only one session controlling the machine and then, if the interface allows it, allow for other session only to observe some few things. Or maybe you may want to kick one (stuck) session from another session (but may lead to other problems).

It can make sense in case you have multiple HMIs / operator stations. The only real problem is that error messages and user message can only be received by one endpoint and if you connect multiple guis it is more or less non-deterministic which GUi receives a particular message. The status OTOH is polled and while in theory there could be races where one gui receives the status another gui polled, that is harmless in my experience. I wrote a debug tool gui that polls with screen refresh rate, and it is no problem to run that parallel to some "proper" gui, in fact, I use that also as a kind of DRO on a second screen on a mill.. In a sense multiple sessions / guis / operator stations is like attaching a pendant that allows to jog and "cycle run". 

On a real machine, all safety-relevant stuff has to be external to the linuxcnc control anyways. Also linuxcncrsh is something that has to be explicitely enabled / started.

rmu75:
multiplexing multiple sessions onto one NML session is probably a bad idea.

BsAtHome:
> multiplexing multiple sessions onto one NML session is probably a bad idea.

Ah, but then shcom.{cc,hh} needs to become a proper class that can be instantiated for every connection to encapsulate separate NML sessions within the same process. That is no problem, I already had a feeling it would be necessary. There is also other work in those files because I detected static buffers that need to go anyway.

When that is done, then there should be no problem in accepting multiple sessions.

rmu75:
see https://github.com/rmu75/cockpit/blob/main/src/shcom.cpp and https://github.com/rmu75/cockpit/blob/main/include/shcom.hh

I also have variants that talk zeroMQ.

BsAtHome:
> see https://github.com/rmu75/cockpit/blob/main/src/shcom.cpp and https://github.com/rmu75/cockpit/blob/main/include/shcom.hh
> I also have variants that talk zeroMQ.

Thanks, that is a good start. I'll have a deeper look and see what I can "steal" ;-)
There are a few things that need refactoring, like getting rid of the shared char buffer(s) killing threaded applications. But I see that a lot of your changes fit in my thoughts.

rmu75:
You should probably connect to user and error channels only once. Or not connect at all -- it won't work correctly if there is a real GUI in parallel.

BsAtHome:
There are three NML channels: emcCommand(W), emcStatus(R) and emcError(R). All are connected to namespace `xemc`.

Which of these is the 'user' channel?

The problem is that if two processes send a command on emcCommand, then the first reader of the emcError will retrieve any current error. Interestingly, the emcStatus channel is only peek'ed at and never actually read.

I am a bit baffled how this is supposed to work at all...

rmu75:
It has been some time since I last looked at that stuff, seems I misrembered about a "user" channel.

I didn't want to look too closely, there are too many things that induce headache. Like https://github.com/LinuxCNC/linuxcnc/blob/master/src/libnml/nml/nml.cc#L1260 and https://github.com/LinuxCNC/linuxcnc/blob/master/src/libnml/nml/nml.cc#L1281. Maybe there is some stubtle stuff I miss, but to me, it looks like a copy/paste error. I assume that every code path that is not used by linuxcnc contains some subtle bugs (like messages containing long truncating those to 32bit or the really strange length calculations in the ascii and display updaters).

BsAtHome:
Yeah, I tried going down that rabbit hole...

At least I can explain the peek() in the status channel. It is because the nml-file shows the buffer as not queued. Both emcCommand and emcError are queued buffers, hence the write/read.

Unfortunately, I've not seen an unread(). Anyway, multiple processes may interfere with each other on basis of the serial number. It seems that a lot of synchronization between write command and read status are based on the serial number. No one is checking the serial number of the error, so I don't even know if any error is correlated to the last command.

andypugh:
>  the first reader of the emcError will retrieve any current error. 
...
> I am a bit baffled how this is supposed to work at all...

This bit has been known to be a problem for decades. The first thing to "look" gets the error status, and might or might not inform the user.

BsAtHome:
> > the first reader of the emcError will retrieve any current error.
> > I am a bit baffled how this is supposed to work at all...
> 
> This bit has been known to be a problem for decades. The first thing to "look" gets the error status, and might or might not inform the user.

Ok then. My suspicions confirmed.

The real problem is that there are separate uncorrelated queued read and write channels that share the backing store.

Analysis: The emc.emcCommand(R) (emctaskmain.cc) receives the commands from any of emcsrv.emcCommand(W) or xemc.emcCommand(W) (for the default nml file). However, it cannot reply to the appropriate ???.emcError channel because these are shared for both emcsrv and xemc.

To fix this, you need a back channel for emcError that matches the command connection. This is not a scalable solution because you need to create complex nml configs with an error buffer for each connection and also track that in the task handling, which is a near impossibility. The changes required are not worth the effort.

The way it looks is that NML is inappropriate for handling this particular case. For this you'd want a socket connection and use sendmsg()/recvmsg(). That means, zeroMQ will be much better than NML to handle communication, including status because it makes a simple bidirectional connection.

@rmu75 you said something about files using zeroMQ. Have you hacked a LCNC version, ripped out NML and replaced it with zeroMQ?

rmu75:
> @rmu75 you said something about files using zeroMQ. Have you hacked a LCNC version, ripped out NML and replaced it with zeroMQ?

I didn't rip out anything, but added all nml messages as flatbuf structures and played around a bit stuffing those into zeromq sockets. See https://github.com/rmu75/linuxcnc/tree/rs/zmq-experiments and https://github.com/rmu75/cockpit/tree/rs/zmq.

It is just a hack, nothing is integrated into the build system, and generated files are checked in. If you want to try that, you will need to install flatbuffer and zeromq dev libs and invoke flatc by hand so generated stuff matches your installed flatbuf libs. flatc needs "--gen-object-api" IIRC.

I did look at machinekit and their machinetalk stuff (google protobuf and zeromq), and I wondered why they didn't go the full length and ripped out NML but built this "bridge". Turned out it is not that hard to feed alternate message format into task. I didn't like protobuf though, flatbuf is more efficient and doesn't suffer from a problematic compatibility story and a 2-3 version-transition that reminds of python's.

IIRC in my zeromq implementation, if you connect the error "channel", you receive error messages. Also status updates are not "peeked" but messages are send to each connection of the status socket.

smoe:
The ROS community now has adopted Zenoh (https://docs.ros.org/en/kilted/Installation/RMW-Implementations/Non-DDS-Implementations/Working-with-Zenoh.html). Open Source and faster than ZMQ  - and would possibly help building bridges to the ROS community.

rmu75:
On So, 2026-02-08 at 05:21 -0800, Steffen Möller wrote:
> smoe left a comment (LinuxCNC/linuxcnc#3734)
> The ROS community now has adopted Zenoh
> (https://docs.ros.org/en/kilted/Installation/RMW-Implementations/Non-
> DDS-Implementations/Working-with-Zenoh.html). Open Source and faster
> than ZMQ  - and would possibly help building bridges to the ROS
> community.

At first glance it looks like zenoh is something like MQTT: "Zenoh is a
distributed service to define, manage and operate on key/value spaces."
It looks like stuff is organised in a hierarchical tree and there you
can store vlues like JSON or plain text. 

not sure how that compares to 0mq and in what sense it could be faster.

BsAtHome:
If I am reading the docs correctly, then Zenoh seems to be a key/value bases system that looks a bit like MQTT.
[ninja'ed :-)]

The part that is bothering me is that I do not know how to map the NML functionality. This is a very non-trivial problem because we have both queued (commands) and non-queued (status) communication. The odd one is the NML error channel, which is also queued but should be one-to-many mapped, but only to a point.

rmu75:
we need some kind of remote procedure call (maybe with useful return-values, that doesn't exist now) and a channel to report status and error.

error messages could be put into a lru structure that keeps the last 10 or some number error messages.

status could in principle be published into a key value structure like mqtt but I think per default we need something that could deal with 1kHz or even 4khz of updates.

https://zenoh.io/blog/2023-03-21-zenoh-vs-mqtt-kafka-dds/

smoe:
> status could in principle be published into a key value structure like mqtt but I think per default we need something that could deal with 1kHz or even 4khz of updates.
> 
> https://zenoh.io/blog/2023-03-21-zenoh-vs-mqtt-kafka-dds/

4 kHz - that is 250 µs between invocations. That suggests to look for the one with minimal latencies. MQTT and all other contestants are already distributed if I get this right, so some external error monitoring would be supported?

BsAtHome:
The fact that the realtime thread runs this fast does not necessarily mean that the non-realtime system must be updated this fast. Actually, when you look at the non-realtime, like GUIs, they only updates at about 10 times per second. For example, the [DISPLAY]CYCLE_TIME is set at 0.1 seconds (sim's axis.ini).

Another thing to remember is that status updates can be done in an incremental fashion. But, for the sake of argument, let a status update happen at 4kHz with a 1400 byte payload (about one eth frame size excl. overhead) and it needs to be handled by a network connection (very unlikely). That would mean we're sending ~5.4 MiB/s data. Well, my internet connection is 4 times faster and I manage to saturate it with little CPU efforts :-)

What is important is data sequence and correlation. Every client (non-realtime) needs to be able to track the server (realtime) and not get confused because multiple clients are acting on the same connection (like we have with the error channel). That is the big challenge to solve so that we have consistency, predictability and repeatability for all clients.

rmu75:
4kHz is probably on the extreme end, but it could also cover cases where you want to synchronise multiple linuxcnc instances or implement remote access to hal.

rmu75:
> What is important is data sequence and correlation. Every client (non-realtime) needs to be able to track the server (realtime) and not get confused because multiple clients are acting on the same connection (like we have with the error channel). That is the big challenge to solve so that we have consistency, predictability and repeatability for all clients.

I don't understand what you mean with this. Now NML just broadcasts a command into emctask. No result is communicated back except for changes in status. IMO it can remain that way.

Status can also be more or less broadcast to all interested parties. Error or user messages can be put in a list. Error message only has informal character and don't need to be acked per se. If something really goes wrong then machine state will reflect that.

So in principle, we can treat communication channels as a room where every client can shout in commands through a window and observe what is happening, who is shouting isn't interesting, state has to be handled by emctask anyways. It might make sense to limit acceptable commands depending on channel.

smoe:
> > What is important is data sequence and correlation. Every client (non-realtime) needs to be able to track the server (realtime) and not get confused because multiple clients are acting on the same connection (like we have with the error channel). That is the big challenge to solve so that we have consistency, predictability and repeatability for all clients.
> 
> I don't understand what you mean with this.

I got this as "among all the many independently created sensor or position or whatever status updates you need to know at what moment what error kicked in and who triggered it".

> Now NML just broadcasts a command into emctask. No result is communicated back except for changes in status. IMO it can remain that way.
> 
> Status can also be more or less broadcast to all interested parties. Error or user messages can be put in a list. Error message only has informal character and don't need to be acked per se. If something really goes wrong then machine state will reflect that.

There is the kind of error info that is printed to stdout/stderr.  That is lost to the system. Such messages are both informal and informative :-) But lost. 

> So in principle, we can treat communication channels as a room where every client can shout in commands through a window and observe what is happening, who is shouting isn't interesting, state has to be handled by emctask anyways. It might make sense to limit acceptable commands depending on channel.

It may be a constructive misunderstanding on my side, but once we have a bit of a more formal representation of errors, then those errors could be also be interpreted by the controller, beyond some "-1" return value, and affect the systems further action - and human monitoring alike. I think we should start out with a diagram about how NML is used today and ornament our documentation with it. And then think more about what we all want. My instant hunch is that we want to borrow more from what ROS2 is doing.

rmu75:
> It may be a constructive misunderstanding on my side, but once we have a bit of a more formal representation of errors, then those errors could be also be interpreted by the controller, beyond some "-1" return value, and affect the systems further action - and human monitoring alike. I think we should start out with a diagram about how NML is used today and ornament our documentation with it. And then think more about what we all want. My instant hunch is that we want to borrow more from what ROS2 is doing.

A typical error that is communicated via error channel is "probe move stopped without contact" or similar. That is a global situation that can and should be communicated to each operator station. But there is not much else to do there.

You only need to look at emctask.cc and emctaskmain.cc. In a nutshell, emccanon sends stuff from the interpreter as NML into task, and the other thing is the GUI that sends commands.

I would proceed step by step, new API can be bolted on later, for me it is a separate issue.

BsAtHome:
> > What is important is data sequence and correlation. Every client (non-realtime) needs to be able to track the server (realtime) and not get confused because multiple clients are acting on the same connection (like we have with the error channel). That is the big challenge to solve so that we have consistency, predictability and repeatability for all clients.
> 
> I don't understand what you mean with this. Now NML just broadcasts a command into emctask. No result is communicated back except for changes in status. IMO it can remain that way.

No, NML does not broadcast. It consists of two queues (command and error) and one (virtually) shared memory region (status).

There is, at this time, no way to correlate the command issued with the error generated. Even worse, the error channel carries a mix of operator error, text and display messages. You have no way to relate that to what the machine was executing at the time the message was generated. And more disastrous, multiple clients reading the error queue will prevent consistency between readers because only the first-to-read will actually get the information.

When you connect two DROs that can display messages, then you _really_ want to have them display the _same_ thing, don't you?

> Status can also be more or less broadcast to all interested parties. Error or user messages can be put in a list. Error message only has informal character and don't need to be acked per se. If something really goes wrong then machine state will reflect that.

Status is a non-queued broadcast. It is the virtually shared memory segment containing the machine's state.

Errors are not necessarily a list nor merely informal. These "error" messages may be broadcast to all clients but correlating this with the current status remains a problem because status is asynchronous from the queues. There needs to be more information. Maybe the status channel could contain _some_ of the actual status, display and error messages and the "error" channel is a broadcast that "something important happened", which listeners can use to act upon. But there also needs to be (possibly historic) information on the error channel which client/command caused what situation.

> So in principle, we can treat communication channels as a room where every client can shout in commands through a window and observe what is happening, who is shouting isn't interesting, state has to be handled by emctask anyways. It might make sense to limit acceptable commands depending on channel.

I disagree that clients/servers just (should) shout at each other. It must be a very coordinated communication that guarantees both consistency and repeatability.

I do agree that we might want to limit commands on some channel depending function/intent of the client.

rmu75:
> > I don't understand what you mean with this. Now NML just broadcasts a command into emctask. No result is communicated back except for changes in status. IMO it can remain that way.
> 
> No, NML does not broadcast. It consists of two queues (command and error) and one (virtually) shared memory region (status).

Maybe I formulated a bit too informal. The effect is the same, commands are fed into a big switch statement, no distinction is made who sent the command.

> There is, at this time, no way to correlate the command issued with the error generated. Even worse, the error channel carries a mix of operator error, text and display messages. You have no way to relate that to what the machine was executing at the time the message was generated. 

That is a design issue of linuxcnc / emc task, not a comms interface issue. 

> And more disastrous, multiple clients reading the error queue will prevent consistency between readers because only the first-to-read will actually get the information.

This is a comms issue and can be fixed (but probably not with NML and acceptable dev-effort)

> When you connect two DROs that can display messages, then you _really_ want to have them display the _same_ thing, don't you?

yes absolutely.

rmu75:
> Errors are not necessarily a list nor merely informal. These "error" messages may be broadcast to all clients but correlating this with the current status remains a problem because status is asynchronous from the queues. There needs to be more information. Maybe the status channel could contain _some_ of the actual status, display and error messages and the "error" channel is a broadcast that "something important happened", which listeners can use to act upon. 

It could make sense to include some sort of "state tag" to error messages created in the interpreter that contains information like filename, linenumber etc... in a non-sprintf-way.

> But there also needs to be (possibly historic) information on the error channel which client/command caused what situation.

Why. If something happens, machine will go either into "Idle", "Off" or "ESTOP", depending on severity. Some message should indicate the cause of course. I agree that if g-code or remapped code is involved, the file/linenumber could/should be indicated. Historic error information IMO needs to go into a separate component that logs problems, I wouldn't put that into emctask.

What would actually be helpful: a message that says "machine can't be enabled because cover X is not closed" or "machine can't be enabled because spindle coolant is out of safe temperature range" or "machine can't be enabled because vacuum pump motor protection tripped", but those things usually are in the HAL domain and would need customization. AFAIK there is currently no possibility to communicate such info from HAL to a GUI. Usually in these situations you just get an enable button that doesn't enable.

smoe:
I try to find a good example for the need of some archeological research in error messages - or just a context-sensitive error interpretation as a start. My imagination describes a scenario with errors that are non-fatal, so the machine continues to operate but was unhappy - in parts (pun intended).  So if there is a robot feeding raw stock in over night and puts the finished parts on a pile, then likely some QA report for each part is also performed, not necessarily directly after the part being produced. Maybe some parts fail in their QA. And then suddenly it is all fine again because a dull end mill was swapped out, which deteriorated earlier than expected. And then you want to understand what was happening when returning in the morning.

I guess any sort of supervision of a process wants to see such context - be it human or an automation. It guides to the cause of the problem as the first step torwads a recovery. And if I understand the exchange between the two of you correctly then you ask what the role of LinuxCNC in any such more complex scenario may be and to what degree it can behave constructively already or if some additional mechanisms are needed.

rmu75:
On Mo, 2026-02-09 at 15:40 -0800, Steffen Möller wrote:
> smoe left a comment (LinuxCNC/linuxcnc#3734)
> I try to find a good example for the need of some archeological
> research in error messages - or just a context-sensitive error
> interpretation as a start. My imagination describes a scenario with
> errors that are non-fatal, so the machine continues to operate but
> was unhappy - in parts (pun intended).  So if there is a robot
> feeding raw stock in over night and puts the finished parts on a
> pile, then likely some QA report for each part is also performed, not
> necessarily directly after the part being produced. Maybe some parts
> fail in their QA. And then suddenly it is all fine again because a
> dull end mill was swapped out, which deteriorated earlier than
> expected. And then you want to understand what was happening when
> returning in the morning.

IMO stuff like this belongs in a layer outside of machine controllers,
like some production orchestration system that interfaces with the ERP.
I'm lacking imagination what kind of "non-fatal" error there could be
in linuxcnc. To me, the definition of "error" means that user
intervention is needed.

Things like bad quality because of dull tools or air-milling stock
because of brocken endmill is nothing that the machine controller can
prevent. Yes there could be tool breakage detection but it probably
only makes sense in a very high volume regime to deal with that
automatically.


> I guess any sort of supervision of a process wants to see such
> context - be it human or an automation. It guides to the cause of the
> problem as the first step torwads a recovery. And if I understand the
> exchange between the two of you correctly then you ask what the role
> of LinuxCNC in any such more complex scenario may be and to what
> degree it can behave constructively already or if some additional
> mechanisms are needed.

For me, I'm more concerned about distributed operator stations that are
somewhat on equal footing. Like main screen, a phone/tablet UI and some
sort of remote web interface, where it is possible to control machine
from each of those, at least in principle and after possibly
authentication. I think those things could be very useful in a hobby
and in a professional context, and currently, it would be hard to
implement correctly.

It could be done like in machinekit with a bridge between NML and
something more modern / supported, but IMO it would be better to
directly address the messaging and not introduce a translation layer.

rmu75:
We should move that discussion to some more appropriate place.

smoe:
> IMO stuff like this belongs in a layer outside of machine controllers, like some production orchestration system that interfaces with the ERP.

Yes, but the controller should pass all information needed.

> I'm lacking imagination what kind of "non-fatal" error there could be in linuxcnc. To me, the definition of "error" means that user intervention is needed.

This depends on the complexity of the mill and the number of sensors it is equipped with. Imagine the mill being a car. "No fuel" is an error in your sense. An unexpected "low fuel" (traffic jam, going to fast, whatever) will grant you a chance to react. Same for your motor losing performance for an unknown reason. You want to continue driving to safety/petrol station with what you have.

If your glass scales give a different position than you are expecting to be at, then at a lower level this may be an error. At a higher level you may change the acceleration or maybe the depth of cut ? 

I think I would want to understand "Error" as "Unexpected". And there may be no user around to deal with the situation but the controller (or some supervising agent) gives commands to react. So you do not want to halt the machine immediately, but a halt (and switching of the coolant) to be one way for the machine to react to the unexpected situation.

rmu75:
> This depends on the complexity of the mill and the number of sensors it is equipped with. Imagine the mill being a car. "No fuel" is an error in your sense. An unexpected "low fuel" (traffic jam, going to fast, whatever) will grant you a chance to react. Same for your motor losing performance for an unknown reason. You want to continue driving to safety/petrol station with what you have.

Everything you describe is pure customization. 

The logic that uses those sensors and generates something actionable can hardly be part of linuxcnc, at best you could implement or configure HAL components and connect some warning lamps. But HAL doesn't have a notion of "error", it is just a bunch of virtual components and virtual wires, there is no or not much state to HAL. If there is an error condition in HAL (like lost comms with a mesa card), the machine is dead. 

> 
> If your glass scales give a different position than you are expecting to be at, then at a lower level this may be an error. At a higher level you may change the acceleration or maybe the depth of cut ?

This also would be a very machine specific configuration. Machine specific components and configuration has to implement this logic.

The NML error channel communicates text messages (either "operator error", "operator text" or "operator display"). That should be augmented to include a source (component, source-file, linenumber, stuff like that) and in the future we will need something more "broadcast style" so it is not one sender talking to one non-deterministic receiver (out of potentially many)

Higher levels are outside, we should provide useful information, relevant stuff is available via status and HAL, not sure what can/should be added there.

PR #3736, opened 2026-01-23
Ensure Limit Axis Works Properly

sfroome:
This was fixed on Master awhile ago, but was never backported. 

The enable pins of limit_axis do not function properly with the if statement being nested in another if statement.


sfroome:
I might have the wrong branch

sfroome:
Apologies. The correct pull request is here: https://github.com/LinuxCNC/linuxcnc/pull/3737

PR #3737, opened 2026-01-23
Fix Bug in Limit Axis Hal Component

sfroome:
Backport bugfix from master that prevents the enable pins on the axis ranges from being enabled/disabled. This results in the hal component defaulting to potentially disabled axis ranges and only checking the ranges rather than the enable statuses.

Replaced #3736 as my source branch was based off the wrong branch.


rmu75:
original pr #2863

PR #3739, opened 2026-01-25
Fix running with tcl/tk 9

### LINKED ISSUE #3713: Tcl 9 breaks build
The new Tcl/Tk 9 release, included in Fedora 42 and 43, breaks the LCNC build. Python on the system is compiled using Tcl/Tk 9 and can therefore not be downgraded to 8.6 development packages. Additionally, it seems that the `tclX` package cannot be found as it is apparently an 8.6 version package, but this just be a symptom of "extension-package-no-longer-needed".

There are a lot of "kludges" in the sources handling many different versions older than 8.6 too. Since master no longer supports old Debians anyway, it would be prudent to clean up the sources and remove the old kludges. Both Bookworm and Trixie use 8.6, so dropping older versions should not be a problem for master. It can be expected that newer Debians will move to the tcl 9 too (Trixie already has a tcl9.0 deb), so this needs to be fixed sooner than later.

BsAtHome:
This PR addresses and fixes #3713.

It should fix the remaining Tcl/Tk 9 problems. Tested to run axis and qtvcp on Fedora 43 (both user interfaces use the togl extension for Tcl/OpenGL view). Also compiled on Trixie and Bookworm (debian 12 and 13) in a VM and confirmed that both will run axis properly.

A few additionally found tilde expansions were addressed too. One was added via the configure script and another was still present in halshow.

Notes:
* The changes lock Tcl/Tk version to >= 8.4 as a configure check. That should be no problem because Debian has been on 8.6 since Buster (debian 11). It might even be tightened a bit.
* The compile will also lock the binary to run with the system's version of Tcl/Tk, which should be no problem because you never downgrade Tcl/Tk on a running system.
* A binary compiled for Tcl 8.6 will not work on a system with Tcl 9.0 (and vice versa). They are binary incompatible and you must recompile for the correct version.

### DIFF
--- a/lib/python/rs274/OpenGLTk.py
+++ b/lib/python/rs274/OpenGLTk.py
@@ -380,8 +380,8 @@ http://www.yorvic.york.ac.uk/~mjh/
         self.activate()
 
     def zoomwheel(self, event):
-        if event.delta > 0: self.zoomin(event)
-        else: self.zoomout(event)
+        if event.delta > 0: self.zoomin()
+        else: self.zoomout()
 
     def tkStartZoom(self, event):
         self.startZoom(event.y)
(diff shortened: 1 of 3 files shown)


andypugh:
Thanks.

PR #3741, opened 2026-01-26
Fix tool mark disappearing with small tool diameter in g-code preview widget

RomSunZ:
Tool marker disappears if its diameter less then some value but not equal zero (for example if diameter from tool table is used as wear corrector)

### DIFF
--- a/configs/sim/axis/axis.ini
+++ b/configs/sim/axis/axis.ini
@@ -51,6 +51,9 @@ INTRO_TIME = 5
 #EDITOR = geany
 TOOL_EDITOR = tooledit
 
+#Prevent tool mark disappearing with small tool diameter in g-code preview widget
+GCODE_VIEW_TOOL_MIN_DIA = 0
+
 INCREMENTS = 1 in, 0.1 in, 10 mil, 1 mil, 1mm, .1mm, 1/8000 in
 
 [FILTER]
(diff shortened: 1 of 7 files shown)


andypugh:
I don't think that it makes sense to update the sample configs to have the same value as the default (ie zero). I feel that it should be some appropriate value dependent on the units of the config. 

Also, the new parameter would need adding to the documentation (docs/src/config/ini-config.adoc)

RomSunZ:
> I don't think that it makes sense to update the sample configs to have the same value as the default (ie zero). I feel that it should be some appropriate value dependent on the units of the config.
> 
> Also, the new parameter would need adding to the documentation (docs/src/config/ini-config.adoc)

I put in ini 0.08 inch for imperial and 2mm for metric. Also I am not familiar with github, should I create additional pull request for documentation or delete this pull request and than create another?

rmu75:
> > I don't think that it makes sense to update the sample configs to have the same value as the default (ie zero). I feel that it should be some appropriate value dependent on the units of the config.
> > Also, the new parameter would need adding to the documentation (docs/src/config/ini-config.adoc)
> 
> I put in ini 0.08 inch for imperial and 2mm for metric. Also I am not familiar with github, should I create additional pull request for documentation or delete this pull request and than create another?

I think you can just add commits to the branch you PRd and push that, it should update the PR, so just add documentation to this PR.

andypugh:
@RomSunZ 
We looked at this in the LinuxCNC video meet-up / PR review and it looks like this is still waiting for some further changes from you?

snowgoer540:
It also needs to be documented in docs/src/config/ini-config.adoc

andypugh:
I merged some docs, and pushed:
https://github.com/LinuxCNC/linuxcnc/commit/4a22a67b664567c0af04269eed52b09a41a658cf

Issue #3720, opened 2026-01-20
Unused inline function


BsAtHome:
Adding the S-curve trajectory planner with jerk limiting (https://github.com/LinuxCNC/linuxcnc/pull/3683) introduced an inline function in `emc/tp/tp.c` line 359 `STATIC inline double tpGetScaledAccel(...)` that is never used in the code.

It may be an oversight, a typo or development leftovers. Someone with knowledge of this code should review it and determine what should be done with it.

Sigma1912:
@grandixximo

grandixximo:
This will double check with YangYang, probably not needed, but we intend on improving arcs/parabolas interpolation, maybe we can still reuse it, pending...

mika4128:
We will remove this code. The function in question was not created by us—it was originally introduced by Robert W. Ellenberg on November 8, 2013, and later removed on January 10, 2019, when it was moved to blendmath.c. However, due to a divergence between our previous version and the upstream master branch, the code was inadvertently reintroduced.

grandixximo:
in #3744 unused function is gone

BsAtHome:
This unused function got re-added in:
https://github.com/LinuxCNC/linuxcnc/pull/3802/commits/1931b859b38fa828369593aebdf202ed23c65acb

Oversight?

grandixximo:
I think so, not sure how it popped up again, weird, I will check with YangYang on monday

grandixximo:
PR #3853 removes the function yet again, sorry about that guys...

Issue #3719, opened 2026-01-20
Overlapping comparisons always evaluate to true


BsAtHome:
Adding the S-curve trajectory planner with jerk limiting (#3683) introduced a rather interesting warning that definitely points towards an error in the code.

In `emc/tp/tp.c` line 2475 there is a comparison like:
```C
  if (cos_theta > min_cos_theta && (maxjerk > 0 || maxjerk < 99999)) {
...
  }
```
Resulting in (using clang 21):
```
emc/tp/tp.c:2475 col 55: warning: overlapping comparisons always evaluate to true [-Wtautological-overlap-compare]
```

The problem is that the `maxjerk` comparison is always true. When it is less than 0, then it is also less than 99999. When it is more than 99999 then it is also larger than 0.

My <i>guess</i> is that it was meant to determine the condition `0 < maxjerk < 99999`, but that is not what the code does. If so, then the `||` should be `&&`. However, that also begs the question what the real boundary values of `maxjerk` are and what happens when you are outside the seemingly arbitrary values of 0 and 99999.

Sigma1912:
@grandixximo

grandixximo:
This was YangYang's Idea, in his opinion jerk of 0 would create no motion, so instead of that we turn off the calcualtions.
A jerk over 10000k is practically equal to not having jerk, so why waste resources, just use original planner, it is basically the same. That was the thinking, i think 1Million would be more reasonable, will discuss with him, probably clean this up.

mika4128:
yes, it's bug. this need remove. At that moment i want if jerk  too big then use old tp.  but this never happened.

rmu75:
This max jerk cutoff limit really should depend on max acceleration of the machine and maybe on the servo period.

max jerk is bounded implicitely by max acceleration times 2 over servo period anyways (ignoring the cubic output resampler which smears things a bit). Max allowed jerk by the s-curve speed profile should be a fraction of that that is not too small. Nico suggested something like 5 - 10 servo periods to reach full acceleration should be reasonable.

grandixximo:
in #3744 the comparison is gone

PR #3744, opened 2026-01-27
S-Curve refinements

grandixximo:
fixed the two bugs
#3719
#3720
and general improvements, most spikes in the halscope are gone.

### DIFF
--- a/lib/python/qtvcp/widgets/detach_tabs.py
+++ b/lib/python/qtvcp/widgets/detach_tabs.py
@@ -12,6 +12,21 @@ class DetachTabWidget(QTabWidget):
 
         self.setTabBar(self.tabBar)
 
+        self.btn = QPushButton(self)
+        self.btn.setEnabled(True)
+        #self.btn.setMinimumSize(64, 40)
+        #self.btn.setIconSize(QSize(38, 38))
+        self.btn.setIcon(QIcon(':/qt-project.org/styles/commonstyle/images/up-32.png'))
+        self.btn.clicked.connect(self.invokeDetachTab)
+        self.setCornerWidget(self.btn)
+
+        self._lastgeometry = None
+
+    @pyqtSlot(bool)
+    def invokeDetachTab(self, **args):
+        x = self.btn.geometry().x()
+        y = self.btn.geometry().y()
+        self.detachTab(self.currentIndex(), QPoint(x,y))
 
     ##
     #  The default movable functionality of QTabWidget must remain disabled
@@ -58,62 +73,14 @@ class DetachTabWidget(QTabWidget):
         detachedTab.setWindowTitle(name + '  - Close to reattach')
         detachedTab.setWindowIcon(icon)
         detachedTab.setObjectName(name)
-        detachedTab.setGeometry(contentWidgetRect)
+        if self._lastgeometry is None:
+            detachedTab.setGeometry(contentWidgetRect)
+            detachedTab.move(point)
+        else:
+            detachedTab.setGeometry(self._lastgeometry)
+
         detachedTab.onCloseSignal.connect(self.attachTab)
-        detachedTab.move(point)
         detachedTab.show()
-
-    def attachTab1(self, contentWidget, name, icon):
-
-        # Make the content widget a child of this widget
-        contentWidget.setParent(self)
-
-
-        # Create an image from the given icon
-        if not icon.isNull():
-            tabIconPixmap = icon.pixmap(icon.availableSizes()[0])
-            tabIconImage = tabIconPixmap.toImage()
-        else:
-            tabIconImage = None
-
-
-        # Create an image of the main window icon
-        if not icon.isNull():
-            windowIconPixmap = self.window().windowIcon().pixmap(icon.availableSizes()[0])
-            windowIconImage = windowIconPixmap.toImage()
-        else:
-            windowIconImage = None
-
-
-        # Determine if the given image and the main window icon are the same.
-        # If they are, then do not add the icon to the tab
-        if name  == 'Model Selection':
-            index = 0
-        elif name  == "Model' Parameters":
-            index = 1
-        elif name  == 'Stim settings':
-            index = 2
-        elif name  == 'Parameter evolution settings':
-            index = 3
-        elif name  == 'LambdaE':
-            index = 4
-        elif name  == 'Simulation settings':
-            index = 5
-        elif name  == 'LFP + PPS + Pulse Results':
-            index = 6
-        if tabIconImage == windowIconImage:
-            index = self.insertTab(index,contentWidget, name)
-            # index = self.addTab(contentWidget, name)
-        else:
-            # index = self.addTab(contentWidget, icon, name)
-            index = self.insertTab(index,contentWidget, icon, name)
-
-
-
-        # Make this tab the current tab
-        if index > -1:
-            self.setCurrentIndex(index)
-
 
     ##
     #  Re-attach the tab by removing the content from the DetachedTab dialog,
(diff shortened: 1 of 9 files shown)


Sigma1912:
I cannot reach the forum at the moment so I'll put this here:

1. Maybe the halpin 'ini.N.jerk' could be called 'ini.N.max_jerk' to be more consistent with '.max_acceleration' and '.max_velocity' 

2. Testing with a 20x20mm square using 'axis_mm_scurve.ini' with G61.1 indicates jerk spikes when the end of the straight segment is reached:
Screenshot from 2026-01-27 09-58-36

3. Testing with real hardware indicates that homing breaks with planner_type=1 (joint will run into hard limits). So it is probably a bad idea to set 'PLANNER_TYPE = 1' in the [TRAJ] section of your ini file when running real hardware. Instead the scurve planner should be activated using the hal pin after the machine has been homed.

4. I don't seem to be able to get any scurve planning when using my real hardware config despite setting  planner_type and jerk limits:
Screenshot from 2026-01-27 10-31-48

My [TRAJ] section looks like this:

```
[TRAJ]
COORDINATES = XXYZ
LINEAR_UNITS = mm
ANGULAR_UNITS = degree
MAX_LINEAR_VELOCITY = 200.0
NO_FORCE_HOMING = 1

MAX_LINEAR_JERK = 1000.0
DEFAULT_LINEAR_JERK = 500.0
```

This is using your current maser branch at https://github.com/grandixximo/linuxcnc/tree/master

grandixximo:
Make your homing speeds slower, download my latest PR most spikes are mostly fixed, check other jerk parameters. Check all planner values. Compare with my configs

Sigma1912:
> Make your homing speeds slower,

You might want to put that information into the example ini file comments. Having the machine run into the hard limits is not a good start here.

> download my latest PR most spikes are mostly fixed

Not sure I understand, I'm running a current build of your master branch:
https://github.com/grandixximo/linuxcnc/tree/master

andypugh:
I will merge this, but we do seem to have work-in-progress?

grandixximo:
I did an update just few hours ago. How fresh is the code?

Sigma1912:
Built it half an hour ago.

Just pulled and built current linuxcnc master (with this PR ):
Screenshot from 2026-01-27 11-29-59

$ make -j4
$ sudo setuid
$ linuxcnc 

Machine configuration file is 'axis_mm_scurve.ini'

MDI: G61.1

I run my gcode file:

```
G21 
F5000
G0 x0 y0
g1 x20
y20
x0
y0
m2
```
 This is what I see.  I'm not saying it's bad, I'm just stating that there seems to be a jerk spike at the end of the segment:

 
Screenshot from 2026-01-27 11-33-47

grandixximo:
This is jerk-cmd, please try the ddt of ddt of ddt of your position, it should actually show what your machine is experiencing, commanded jerk is not reflective of what the axia experience, at the end of the segment there are some tricks we are doing internally to terminate the segment.

What do you suggest I add in the sample ini file? I do not have homing fields at the moment, should I add some comment out home search and home latch vel, with warning about slowing down from non scurve? Is that acceptable? You are referencing the axis_mm_scurve.ini correct?

grandixximo:
sorry, I was replaying by email, I did not understand the messages were here, what a dummy lol

I also found that it does not like to be activated from planner 0, and I don't really get why, I will debug more

grandixximo:
loadrt ddt names=j0_vel,j0_acc,j0_jerk,j1_vel,j1_acc,j1_jerk,j2_vel,j2_acc,j2_jerk,j3_vel,j3_acc,j3_jerk,j4_vel,j4_acc,j4_jerk


addf motion-command-handler   	servo-thread
addf motion-controller        	servo-thread
addf j0_vel						servo-thread
addf j0_acc						servo-thread
addf j0_jerk					servo-thread
addf j1_vel						servo-thread
addf j1_acc						servo-thread
addf j1_jerk					servo-thread
addf j2_vel						servo-thread
addf j2_acc						servo-thread
addf j2_jerk					servo-thread
addf j3_vel						servo-thread
addf j3_acc						servo-thread
addf j3_jerk					servo-thread
addf j4_vel						servo-thread
addf j4_acc						servo-thread
addf j4_jerk					servo-thread



net x-pos-fb		joint.0.motor-pos-cmd	=>	joint.0.motor-pos-fb	j0_vel.in	J0_switch.cur-pos
net x-vel			j0_vel.out				=>	j0_acc.in
net x-acc			j0_acc.out				=>	j0_jerk.in

net y-pos-fb		joint.1.motor-pos-cmd	=>	joint.1.motor-pos-fb	j1_vel.in	J1_switch.cur-pos
net y-vel			j1_vel.out				=>	j1_acc.in
net y-acc			j1_acc.out				=>	j1_jerk.in

net z-pos-fb		joint.2.motor-pos-cmd	=>	joint.2.motor-pos-fb	j2_vel.in	J2_switch.cur-pos
net z-vel			j2_vel.out				=>	j2_acc.in
net z-acc			j2_acc.out				=>	j2_jerk.in

net a-pos-fb		joint.3.motor-pos-cmd	=>	joint.3.motor-pos-fb	j3_vel.in	J3_switch.cur-pos
net a-vel			j3_vel.out				=>	j3_acc.in
net a-acc			j3_acc.out				=>	j3_jerk.in

net c-pos-fb		joint.4.motor-pos-cmd	=>	joint.4.motor-pos-fb	j4_vel.in	J4_switch.cur-pos
net c-vel			j4_vel.out				=>	j4_acc.in
net c-acc			j4_acc.out				=>	j4_jerk.in


try to test with these, I will have to have a look at how the jerk-cmd actually gets to the output

grandixximo:
I did another push, it should be easier to set jerk now even if you did not set anything to begin with, but mind that all joint and axis you want scurve on need to have a non zero value, you must ini.traj_max_jerk and ini.traj_default_jerk to non zero values, before changing planner_type to 1
In this #3744 PR there is a bug where you MUST set ini.traj_max_jerk first, in the new #3746 you can set in different order, but still you have to set them if you want jerk control.

rmu75:
> I cannot reach the forum at the moment so I'll put this here:
[...]
> ```
> [TRAJ]
> COORDINATES = XXYZ
> LINEAR_UNITS = mm
> ANGULAR_UNITS = degree
> MAX_LINEAR_VELOCITY = 200.0
> NO_FORCE_HOMING = 1
> 
> MAX_LINEAR_JERK = 1000.0
> DEFAULT_LINEAR_JERK = 500.0
> ```

you probably need much higher max and default jerk values. assuming you have acceleration set to 1000, a default jerk of 500 means it would take 2 whole seconds until machine is reaching full acceleration.

grandixximo:
the problem is finding the sweet spot, where testers will feel the difference, but the machines will not jolt, I went for 1000 just cause it's a round number, and you really can see the smooth curves, and can feel the difference, on and off, it is a testing in vitual environment kind of ini, not a run on the machine ini, I can change it to anything, what do you propose is sweet spot? you see curves in the scope, machine moves well, the limit sensors are not triggered? it is very dependent on machine weight and rigidity, I am not sure what most tester are running, I use 4000 on my machines, but they are heavy, 5000? 10000? what's a good default value that works for most users?

rmu75:
A linuxcnc machine using the traditional trapezoidal speed profile has inherent jerk limit of max acceleration over servo period (possibly times 2).  We discussed that in the bi-weekly video dev meeting some weeks ago, the opinion was a reasonable value would be a not too small fraction of that inherent max jerk, like something between 5%-25%. That would translate to reaching full acceleration in 4 to 20 servo periods.

The ini above does not show max acceleration, assuming it was 1000, i would set jerk limit to 100000. (Assuming jerk limit is in mm/s³)

rmu75:
On your machines, what acceleration and max velocity do you use? I assume that cutting stone with diamond saw and similar processes are on the slow side compared to typical modern CNC mills.

grandixximo:
Edit:

MAX_VELOCITY =  500
MAX_ACCELERATION =  3500
MAX_JERK = 4000


We run them smooth, no hurry

grandixximo:
I think changing from 2ms to 20ms it's not something a human can feel, I think something around 200-300ms, which is around human reaction time, can actually be felt as a difference, I maybe wrong, I think 10000 is more reasonable, as a sweet spot, thoughts?

PR #3746, opened 2026-01-27
S-curve jerk control fixes as requested in #3744

grandixximo:
Following discussion with Sigma1912 in #3744
Added ini warning and changed pin names

### DIFF
--- a/src/emc/usr_intf/axis/scripts/axis.py
+++ b/src/emc/usr_intf/axis/scripts/axis.py
@@ -123,12 +123,16 @@ ap = AxisPreferences()
 # Handle repeated key press events
 pressed_keys_list = []
 def key_pressed(ev):
+    if None == ev:
+        return False
     if ev.keysym in pressed_keys_list:
         return True
     pressed_keys_list.append(ev.keysym)
     return False
 
 def key_released(ev):
+    if None == ev:
+        return
     # KeyRelease without KeyPress may happen when a modifier is active when
     # the key is pressed without a KeyPress handler. No KeyPress event is
     # generated, but releasing the actual key while still holding the modifier
@@ -2702,21 +2706,21 @@ class TclCommands(nf.TclCommands):
 
     # The next three don't have 'manual_ok' because that's done in jog_on /
     # jog_off
-    def jog_plus(event):
+    def jog_plus(event=None):
         if key_pressed(event):
             return  # Ignore repeated press events
         a = ja_from_rbutton()
         speed = get_jog_speed(a)
         jog_on(a, speed)
 
-    def jog_minus(event):
+    def jog_minus(event=None):
         if key_pressed(event):
             return  # Ignore repeated press events
         a = ja_from_rbutton()
         speed = get_jog_speed(a)
         jog_on(a, -speed)
 
-    def jog_stop(event):
+    def jog_stop(event=None):
         key_released(event)
         a = ja_from_rbutton()
         jog_off(a)


grandixximo:
last commit fixes the jerk-cmd output for XYZ

configs/sim/axis/axis_mm_scurve.ini
Sigma1912: I would strongly recommend to NOT activate the new planner in the ini file.
I dropped my homing velocities to 20% and my real hardware setup still runs into the hard stops on homing. 
This is  going to annoy a LOT of  people.
Tell users to leave ini.planner_type = 0, home the machine and use halshow to set  ini.planner_type = 1.

smoe:
The current master and your master branch from which you request to be pulled from show following warnings triggered by cppcheck (executed within src as ../scripts/cppcheck.sh) refering to ini_maxjerk not being initialized.

```
/home/moeller/GitHub/linuxcnc/include/emc_nml.hh:731:5: warning: Member variable 'EMC_TRAJ_LINEAR_MOVE::ini_maxjerk' is not initialized in the constructor. [uninitMemberVar]
    EMC_TRAJ_LINEAR_MOVE()
    ^
/home/moeller/GitHub/linuxcnc/include/emc_nml.hh:756:5: warning: Member variable 'EMC_TRAJ_CIRCULAR_MOVE::ini_maxjerk' is not initialized in the constructor. [uninitMemberVar]
    EMC_TRAJ_CIRCULAR_MOVE()
    ^
/home/moeller/GitHub/linuxcnc/include/emc_nml.hh:909:5: warning: Member variable 'EMC_TRAJ_PROBE::ini_maxjerk' is not initialized in the constructor. [uninitMemberVar]
    EMC_TRAJ_PROBE()
    ^
/home/moeller/GitHub/linuxcnc/include/emc_nml.hh:932:5: warning: Member variable 'EMC_TRAJ_RIGID_TAP::ini_maxjerk' is not initialized in the constructor. [uninitMemberVar]
    EMC_TRAJ_RIGID_TAP()
```

No idea if this is related to your work, but anyway you may have an immediate idea how to best address this.

grandixximo:
warnings should be fixed, yes were due to my changes, forgot to add in NML

grandixximo:
This is to improve default behavior when max_jerk is not set and planner type is set to 1

grandixximo:
I don;t think it fails cause of me
should I merge upstream again?

grandixximo:
I think definitely something with this 24.04 arm dependencies, not something I did

@andypugh is it something you can take a look at?

rmu75:
it seems there are tests in the testsuite that expect
```
SET_JERK jerk=0
```
but in fact
```
SET_JERK jerk=10000
```
happens.

* interp/m98m99/12-M99-endless-main-program
* motion-logger/basic
* motion-logger/mountaindew
* motion-logger/startup-gcode-abort

grandixximo:
I see, actually 0 is not very good as default, because the users like PCW don't set a value then planner_type 1 cannot function proper, so I gave 10k which is what I thought was a sweet spot, or we could set 1e99 ? but YangYang said that may also cause issues, we never tested with such high values

Can I change the tests then?

rmu75:
> Can I change the tests then?

Yes the tests should be changed in lockstep with the code. You can run the tests locally before committing, that usually is way faster than waiting for the CI.

grandixximo:
Tests are fixed, not sure why doesn't build on arm

rmu75:
> Tests are fixed, not sure why doesn't build on arm

That looks like a temporary hickup in debian sid, should go away by itself. Having sid in the CI always is a bit of a gamble.

grandixximo:
All green, that's what I like to see 🤭

rmu75:
> I see, actually 0 is not very good as default, because the users like PCW don't set a value then planner_type 1 cannot function proper, so I gave 10k which is what I thought was a sweet spot, or we could set 1e99 ? but YangYang said that may also cause issues, we never tested with such high values

I think I commented elsewhere. People are used to tune max acceleration on their machines, and IMO recommended default max jerk limit should not be too low so max accel can be reached between a few to maybe 100 servo periods. After all machine is supposed to run more or less fine with that accel limit and trapezoidal speed.

Testing with my experimantal 5th order bezier speed profile let me set max acceleration much higher (like ×2) on a stepper machine that would stall out with the trapezoidal speed profile at that acceleration.

I didn't try yet on real hardware because machine is in use currently, but I think even high jerk limits like 5 servo periods will make a noticeable difference, especially when doing something like jog around or milling an adaptive toolpath with small feeds and rapids intertwined.

Going to 1e99 should result in more or less trapezoidal speed profile without any negative side effects, if it doesn't, that needs to be fixed IMO.

grandixximo:
1e99 will break the calculations nothing will move.

double T1 = amax / jerk;        // Nearly zero when jerk = 1e+99
T1 = sqrt(v / jerk);           // Nearly zero when jerk = 1e+99  
When jerk is 1e+99:

distance / 1e+99 becomes extremely small almost zero
Cube roots and square roots of near-zero values cause numerical instability
Division operations result in values too small for double precision
The motion planner can't compute valid S-curve parameters, so nothing moves

1e9 is more reasonable, and will get something like trapez, I will set it as default, and cap it as max.

smoe:
> I see, actually 0 is not very good as default, because the users like PCW don't set a value then planner_type 1 cannot function proper, so I gave 10k which is what I thought was a sweet spot, or we could set 1e99 ? but YangYang said that may also cause issues, we never tested with such high values

If 0 is no good max jerk, and the NML initialisation possibly being used from what cppcheck states, should it then also be initialised to a different value than 0 in https://github.com/LinuxCNC/linuxcnc/pull/3746/changes/34578c3537fea23f1a932dc8ee11568ce073e669?

grandixximo:
@smoe Thanks you for bringing this to my attention, I think we should keep the NML initialization at 0.0 intentionally, it's really just a safety fallback since all NML messages get populated from system state before they're used (see emccanon.cc).

Having 0.0 as the struct default gives us a nice property: if something goes wrong and a message doesn't get populated, it safely falls back to trapezoidal mode rather than silently using S-curve. It's a clear "uninitialized" sentinel value.

The real user-facing default of 1e9 is set where it should be, at the system level in initraj.cc. I think keeping them different makes sense here, but let me know if you feel strongly about it, I can definitely initialize at 1e9 if you'd prefer to see that.

All cppchecks pass, if I'm thinking about this wrong let me know.

grandixximo:
could this be merged in the weekend? would love to have it in master, YangYang wants to rebase on new master after the merge, to make sure everything lines up before continue development, can that happen or will this be on hold for longer? Don't want to be pushy, just trying to figure out how to proceed on our end.

Sigma1912:
> could this be merged in the weekend?

Might want to ping @andypugh for that question.

PR #3749, opened 2026-01-28
Enable loading of Qt Dragon without opencv installed

jtpedersen:
I wanted to try the QtDragon ui and ran into this issue when not having opencv installed.

There was code to handle the case, but something most likely changed so it was no longer valid.

"works on my machine"


andypugh:
Looks OK to me.

PR #3750, opened 2026-01-29
Replace xset in axis with detectable auto repeat

BsAtHome:
This PR replaces #3730 by removing the use of xset in axis completely. It is replaced by setting the "detectable auto repeat" feature of the X server and ignore repeated press events where necessary. Tracking the press/release events with detectable auto repeat makes the jog keys work as intended.

### DIFF
--- a/src/emc/usr_intf/axis/Submakefile
+++ b/src/emc/usr_intf/axis/Submakefile
@@ -1,14 +1,18 @@
 
 EMCMODULESRCS := emc/usr_intf/axis/extensions/emcmodule.cc
 TOGLMODULESRCS := emc/usr_intf/axis/extensions/_toglmodule.c
-PYSRCS += $(EMCMODULESRCS) $(TOGLMODULESRCS)
+TKDARMODULESRCS := emc/usr_intf/axis/extensions/tkdarmodule.c
+PYSRCS += $(EMCMODULESRCS) $(TOGLMODULESRCS) $(TKDARMODULESRCS)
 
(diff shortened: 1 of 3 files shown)


andypugh:
This looks like it was quite a bit of effort! Personally I would just ban keyboard jogging..... :-)

Is there any increased danger of key-release events being "lost"? I have heard occasional reports of a key-release not stopping the jog. (hence my policy above).

BsAtHome:
The chance of not receiving a key release is equal or less than before. The problem of missing key (release) events is usually a problem of bad key-rollover implementations of the physical keyboards. Some keyboards can not cope with multiple/many keys pressed if the keys are in the "wrong" rows/columns of the keyboard matrix. Other keyboards just "forget" keys if you press too many at the same time (for an arbitrary value of "too many").

The effort put into this is very much compensated by eliminating the annoyances generated by xset killing my workflow. Also this fixes the annoyance that a local program alters global settings (at least for axis).

FWIW, the changes were a lot less invasive and smaller than anticipated. And, I learned something about python/tcl/tk/X interactions too :-)

grandixximo:
This breaks moving with the buttons in axis
I've commit a fix together with my PR #3746
hash is d07b95f164ad773eee15e405b57cb88962eb12e8

andypugh:
I was berating myself for not testing this before merging, but console myself that I would never have thought of testing continuous mouse jogging anyway...

BsAtHome:
Same here... Fixed with #3758.

PR #3752, opened 2026-01-29
cppcheck: fix missing field initializers in constructor

BsAtHome:
The new S-curve TP caused several missing constructor field init problems seen by cppcheck.
Also included is the removal of a #pragma to ignore missing initializers that was inappropriate.

### DIFF
--- a/src/emc/nml_intf/emc_nml.hh
+++ b/src/emc/nml_intf/emc_nml.hh
@@ -735,6 +735,7 @@ class EMC_TRAJ_LINEAR_MOVE:public EMC_TRAJ_CMD_MSG {
         vel(0.0),
         ini_maxvel(0.0),
         acc(0.0),
+        ini_maxjerk(0.0),
         feed_mode(0),
         indexer_jnum(0)
     {};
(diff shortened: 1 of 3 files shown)


smoe:
https://github.com/LinuxCNC/linuxcnc/pull/3752 is partially addressed by [4eef784](https://github.com/LinuxCNC/linuxcnc/pull/3752/commits/4eef784f8922b2afd76fcdfbb8e9bfac66ed1067).

BsAtHome:
Ehm, yes, that is actually the commit I created in this PR.

Why are you telling it is "partially addressed"? Rerunning cppcheck with the proposed changes removes the particular warnings. It does not remove all cppcheck warnings, but that was never the intention. I'd like to keep these types of change sets isolated if they ever need to be bisected.

smoe:
> Ehm, yes, that is actually the commit I created in this PR.
> 
> Why are you telling it is "partially addressed"? 

I pasted to the wrong browser tab. I meant to point to your PR from https://github.com/LinuxCNC/linuxcnc/pull/3746 .

BsAtHome:
Ah, yes, that explains it :-)

But now it seems to be an academic discussion because it is already merged. That should be no problem for #3746 as these hunks would simply be skipped when that is merged.

Issue #3645, opened 2025-11-24
enabling coredumps of rtapi_app uspace


rmu75:
To debug certain crashes in realtime components it would make sense to enable coredumps. Basically limits need to be set appropriately (e.g. via systemd-coredump), and you need to set sysctl `fs.suid_dumpable = 2` to allow dumping of setuid binaries. It looks like uspace_rtapi_app does explicitely enable coredumps.

But there is something puzzling (to me at least): to get the thing to actually dump to disc, I also need to replace https://github.com/LinuxCNC/linuxcnc/blob/bd96ca87585d4b460e4ef1cac417823927eedcfc/src/rtapi/uspace_rtapi_app.cc#L692 "raise(sig)" with "kill(getpid(), sig);".

Man page says that for singlethreaded apps these 2 instructions are equivalent. rtapi_app usually consists of >=2 threads AFAIU, one running with non-critical priority, and one or more with realtime priority. 

I'm not sure where this "kill(getpid(), sig)" is sending the signal, I suspect to the main non-realtime thread, and there the actual core dump can happen.

So if anybody knows what's going on, please chime in.

rene-dev:
in the last meeting we decided to just change this, as it should not affect anything.

PR #3759, opened 2026-01-30
change how signal handler raises fatal signal to enable core dump files

rmu75:


andypugh:
Is this related to https://github.com/LinuxCNC/linuxcnc/issues/3645 ?

rmu75:
> Is this related to #3645 ?

yes sorry the issue number is only in my local branch name and got lost in the PR.

PR #3771, opened 2026-02-01
libnml,emc: strtok -> strtok_r

smoe:
The strtok function is not reentrant. It has an internal static char* memorizing where the function has last split the string when it is called again. As a consequence, its calls cannot be nested (as in an outer and inner loop) and also it cannot be used at the same time within two different threads.

The strtok_r function expects the programmer to take care of that internal variable, instead. This is what this patch is all about. The remnant strtok usage of LinuxCNC are for the binaries that implement remote access to LinuxCNC. Those would be more intrusive to transition to strtok_r and are not expected to be of immediate relevance, so this is not addressed in this stable branch.

### DIFF
--- a/src/emc/rs274ngc/interp_remap.cc
+++ b/src/emc/rs274ngc/interp_remap.cc
@@ -368,11 +368,12 @@ int Interp::parse_remap(const char *inistring, int lineno)
     if ((s = strchr(iniline, '#')) != NULL) {
 	*s = '\0';
     }
-    s = strtok((char *) iniline, " \t");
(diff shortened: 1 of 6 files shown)


BsAtHome:
<i>Pedantic mode enable</i>: isn't it "`saveptr`" and not "`safeptr`" (v verses f)?

smoe:
> Pedantic mode enable: isn't it "`saveptr`" and not "`safeptr`" (v verses f)?

Addressed. Thank you!

PR #3772, opened 2026-02-01
change emcmot motion id to struct containing hash of filename

rmu75:
as a first step, change type of motion id from int to a struct / union consisting of line_number and hash of source file name.


smoe:
The printf("...%s...", filename) I would feel like flanking with quotes, just in case some folks have blanks or other wild characters, so they become printf("...  '%s' ...", filename).
Liked to see the working with unions - has been a while for me.

BsAtHome:
What is the status of this PR?

PR #3777, opened 2026-02-01
Weblate changes merged into new branch to manually resolve conflict

smoe:
Let's see how this goes.


smoe:
Resolved conflict (it is only the date in the es.po file), but somehow these commits that appear redundant appeared. Don't merge, yet.

smoe:
Tried again - same thing - will investigate tomorrow again. Expecting no harm from merging this PR. Weblate could not update its repository for a while, which may have caused what looks weird to me.

PR #3780, opened 2026-02-02
d/configure: Prepare for execution on other OS

smoe:
Somehow I feel like it may possibly not be completely pointless to consider a compilation of LinuxCNC or parts of it on non-Linux Operating Systems. At least we should prepare debian/configure for the event that someone is trying it. Here is what I came up with. My personal immediate motivation is to support writers of the documentation who are likely to have a the source tree on whatever OS their machine may be.

My attempt on MacOS did not complete a configure run. But that is just because of missing packages in brew - nothing technical. Conda may also be worthwhile to have a look, but the libgpiod-dev equivalent for either I fail to find. It may be of interest to come up with a configure option that prepared for the documentation only.

### DIFF
--- a/debian/configure
+++ b/debian/configure
@@ -1,6 +1,7 @@
 #!/bin/bash
 # dpkg configuration script for linuxcnc
-# Copyright (C) 2006 Jeff Epler
+# Copyright (C) 2006      Jeff Epler
+#               2025-2026 Steffen Moeller
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


rene-dev:
Afaik it does build and run on freebsd, and on macos I got stuck getting the linker to create hal modules.

andypugh:
I did look into this and MacOS does have some form of RT API. But I think that Rene said is wasn't useful for us. 
https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/KernelProgramming/scheduler/scheduler.html

smoe:
At the very moment, starting of with what brew provides, LinuxCNC does not configure on MacOS. I do not know where any sort of direction for users of other platforms should be gatherning, so I must admit. And d/configure should eventually be removed in favour of a distribution-independent setup, I guess. So this PR is a bit of an experimental nature. I happily change it any way you want, am just uncertain if that URL should be in the d/configure sources or rather in some new section of our documentation - or maybe there should be a forum section for porting LinuxCNC.

andypugh:
Do you want to merge this?

smoe:
> Do you want to merge this?

Yes. Less because of its immediate functional impact than for its signal to strive at more than Debian (and that is me as a Debian Developer stating that).

andypugh:
As a note, the lack of libgpiod on Darwin is extremely unlikely to be a drawback. Whilst you could attempt to run Darwin on a Pi, it would be an eccentric choice for a LinuxCNC machine, and I don't know of any Apple hardware with exposed GPIO pins.

smoe:
Ah - the missing libgpiod has blocked src/configure for me. I'll see that I prepare a patch to make that optional.

PR #3787, opened 2026-02-04
tooledit: save current text when in edit mode

### LINKED ISSUE #2818: Gmoccapy - confirm value in Tool table

hansu:
Resolves #2818

This saves the current text when clicking on save while editing a cell.
Normally you have to commit that by hitting enter or click on another cell.

tooltable_save_edit-mode


zz912:
Thank you.

PR #3788, opened 2026-02-05
halscope: clarify restart message, dont execute comment lines

snowgoer540:


grandixximo:
Thank you, was on my to do list, just didn't get to it, I also need check when loading 16 channels, it seems to be missing some from autosave.

PR #3789, opened 2026-02-05
Gladevcp calculator_widget: Fix remnant locale handling

Sigma1912:
Ignore locale in functions 'get_value()' and 'set_value' when using 'use_localization' = False (default)

Missed this in https://github.com/LinuxCNC/linuxcnc/pull/3691


hansu:
Much better :smiley:  , Thanks!

PR #3790, opened 2026-02-05
Fix an endless loop.

mark-v-d:
The following program created an infinite loop, this patch fixes that

-----------------------------------------------------------------------------------------------------------------
#<_initial>=20
#<_final>=19.726
#<_champher>=0.5
#<_zmax>=0
#<_zmin>=-25
#<doc>=2

G54
G7
G18
G21
T2M6G43

o100 sub
        o101 if [#<_champher> gt 0]
                G1 X[#<_final>-2.0001*#<_champher>]
                Z#<_zmax>
                X#<_final>C#<_champher>
        o101 else
                X#<_final>
        o101 endif
        W-5
        U0.5
        Z#<_zmin>
        X#<_initial>
o100 endsub

M3 G96 S200 d1400
G95 F0.12
G52 z[0.1-#<doc>]
;m7
G71 Q100 Z[#<_zmax>+0.8+#<doc>] X#<_initial> D#<doc> I#<doc> R0.3
G52 z0
;m7
g42
G70 Q100 Z[#<_zmax>+1] X#<_initial> D#<doc> P1
g40
M9
G0X[#<_initial>+5]

M2


andypugh:
There seems to be a conflict with other changes to interp_g7x.cc

Your change:
```
	if(std::abs(imag(location)-x)>tolerance || (*p)->ep()==location) {
```
Existing code in Master:
```
        // We have added our move commands for the last path, we end the
        // recursive function before calculating and adding intersections.
	if (p==std::prev(end()))
	    return;

	if(std::abs(imag(location)-x)>tolerance) {
>>>>>>> master
```

Is it possible that this is already fixed?

mark-v-d:
I'll have a look soon, but I'm not sure it solves the same problem.
I will also improve my patch, since == on floating point numbers is not
a good idea.

andypugh ***@***.***> wrote:

	andypugh left a comment (LinuxCNC/linuxcnc#3790)
	
	There seems to be a conflict with other changes to interp_g7x.cc
	
	```
	<<<<<<< 2.9
		if(std::abs(imag(location)-x)>tolerance || (*p)->ep()==location) {
	=======
	
	        // We have added our move commands for the last path, we end the
	        // recursive function before calculating and adding intersections.
		if (p==std::prev(end()))
		    return;
	
		if(std::abs(imag(location)-x)>tolerance) {
	>>>>>>> master
	```
	
	Is it possible that this is already fixed?

petterreinholdtsen:
I corrected the base for this pull request to 2.9, and got rid of the conflict.

petterreinholdtsen:
Given that the submitters branch was based on 2.9, I assume it was ment to be targeting 2.9, not master.

Given the example program, I recommend adding a test to demonstrate the problem and hence also demonstrate that the patch fixes the problem.  Can you provide code in tests/ to demonstrate the endless loop?

PR #3791, opened 2026-02-06
Don't crash milltask when no tool data file is specified

BsAtHome:
This PR addresses a crash in milltask that happens when the ini-file lacks an entry for [EMCIO]TOOL_TABLE and a request is made to set a tool offset. The filename argument to tooldata_save() would be NULL and was naively dereferenced.

### DIFF
--- a/src/emc/tooldata/tooldata_common.cc
+++ b/src/emc/tooldata/tooldata_common.cc
@@ -388,6 +388,13 @@ int tooldata_save(const char *filename)
         if (!is_random_toolchanger) {return 0;}
         filename = DB_SPINDLE_SAVE; //one entry tbl (nonran only)
     } else {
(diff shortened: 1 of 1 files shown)


src/emc/tooldata/tooldata_common.cc
```diff
@@ -384,6 +384,11 @@ int tooldata_save(const char *filename)
     FILE *fp;
     int start_idx;
 
+    // filename == NULL happens without ini entry [EMCIO]TOOL_TABLE
+    // and a Task::emcToolSetOffset is performed.
+    if (!filename)
```
smoe: A bit later the filename may be assigned the value of DB_SPINDLE_SAVE.  Your check may be a bit too early.

I suggest to always perform the filename[0]== 0 check (empty string test) rather than having it conditional  for the !DB_ACTIVE branch and to prepend your test on NULL.

PR #3792, opened 2026-02-07
Gmoccapy: make Tool Table exit more verbose

zz912:
LinuxCNC core behavior:
Activating G43 with zero offsets deactivates tool compensation.
In this situation, G43 behaves like G49.
If the CNC operator is not aware of this, it can have dangerous consequences.

@hansu was not aware of this in the past. See:
https://github.com/LinuxCNC/linuxcnc/pull/3706#issuecomment-3777527229
I struggled with this as well.

Therefore, I made exiting the Tool Table in Gmoccapy more verbose and added the option to activate G43 when the active offset changes from zero to non-zero.

I would like to ask @Sigma1912 for their opinion on whether this is OK.

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -4744,11 +4744,20 @@ class gmoccapy(object):
             new_offset = (tt.xoffset, tt.yoffset, tt.zoffset,
                           tt.aoffset, tt.boffset, tt.coffset,
                           tt.uoffset, tt.voffset, tt.woffset)
+            message = None
+            new_offset_is_nonzero = any(v != 0.0 for v in new_offset)
+            old_offset_is_zero = all(v == 0.0 for v in self.stat.tool_offset)
             if (new_offset != self.stat.tool_offset) and ("G43" in self.active_gcodes):
                 message = _("Offset values for the tool in the spindle\n" \
                             "have been changed with tool compensation (G43) active.\n\n" \
                             "Do you want the new values to be applied as the currently\n" \
                             "active tool offset?")
+            elif new_offset_is_nonzero and old_offset_is_zero and ("G49" in self.active_gcodes):
+                message = _("Offset values for the tool in the spindle\n" \
+                            "have been changed from zero to non-zero.\n\n" \
+                            "Do you want to activate tool compensation (G43)\n" \
+                            "using the currently active tool offset?")
+            if message:
                 result = self.dialogs.yesno_dialog(self, message, _("Attention!"))
                 if result: # user says YES
                     self.command.mode(linuxcnc.MODE_MDI)


Sigma1912:
I have no problem with this addition.

src/emc/usr_intf/gmoccapy/gmoccapy.py
```diff
@@ -4744,11 +4744,27 @@ def on_btn_back_clicked(self, widget, data=None):
...
                           tt.aoffset, tt.boffset, tt.coffset,
                           tt.uoffset, tt.voffset, tt.woffset)
-            if (new_offset != self.stat.tool_offset) and ("G43" in self.active_gcodes):
-                message = _("Offset values for the tool in the spindle\n" \
-                            "have been changed with tool compensation (G43) active.\n\n" \
-                            "Do you want the new values to be applied as the currently\n" \
+            message = None
+            new_offset_is_nonzero = any(v != 0.0 for v in new_offset)
+            new_offset_is_zero = all(v == 0.0 for v in new_offset)
+            old_offset_is_zero = all(v == 0.0 for v in self.stat.tool_offset)
```
hansu: I don't think it's a good idea to compare float with "==" ...

zz912: You're right — comparing floats with "==" is not ideal.

This part was drafted with the help of ChatGPT, and I don't insist on this exact implementation. The EPS-based comparison below is also ChatGPT's suggestion, not my own Python expertise:

    all(abs(v) < 1e-9 for v in self.stat.tool_offset)

I'm happy to change the code accordingly, or to follow whatever approach you consider more appropriate for LinuxCNC.

Please feel free to treat this PR as a starting point for discussion. If you prefer, it can also be closed and used as input for a new PR with a better solution.

BsAtHome:
Please resolve the review if the changes are OK. Then we can merge it.

rmu75:
IMO this PR needs a proper commit message, I have no idea what a "Tool Table exit" is and why / how it affects (only?) gmoccapy.

zz912:
Current LinuxCNC behavior:
gm-tooltable-exit-001
1) Change the parameter of the selected tool
2) Save the tool table
3) Leave (exit) tooltable page
4) You will get popup message
5) If you give Yes, G43 will be executed and the corrections will be updated

This PR adds another popup message (4.) for situations when the operator selects offset == 0, or previously was offset == 0 and the operator changed the offset != 0.

rmu75:
> Activating G43 with zero offsets deactivates tool compensation.

I wonder what the rationale is justifying this behaviour or if it is simply a bug in the interpreter. It certainly isn't documented here https://linuxcnc.org/docs/devel/html/gcode/g-code.html#gcode:g43

Playing around with this a bit in qtvcp and axis, it seems there are bugs like this in other GUIs too. 

Maybe the proper place to fix this is somewhere else?

zz912:
> ... or if it is simply a bug in the interpreter.

That's an interesting question. I just took it for properties. It didn't occur to me that it could be a bug.
Does anyone know how it is with other systems (Fanuc, Heidenhain, Mach, Siemens ...)?

> Maybe the proper place to fix this is somewhere else?

If this is a bug, it should be fixed elsewhere. But that's beyond my capabilities.

Sigma1912:
> Activating G43 with zero offsets deactivates tool compensation.

Looking at this a bit more closely makes me think that this is actually a phantom problem as I'm not sure there IS an internal state like 'activated' / 'deactivated' tool offset as such. All we seem to do with G43/G49 is change the values of the internal tool length compensation. For G49 they are all set to zero.
See:
https://github.com/LinuxCNC/linuxcnc/blob/f82c475ac13b2919466a99a8570dc42866c564d6/src/emc/rs274ngc/interp_convert.cc#L6183

The displayed 'G43' / 'G49'  in the 'active gcodes' only reflects whether the current internal tool length values are all zero ('G49') or not ('G43')   
See:
https://github.com/LinuxCNC/linuxcnc/blob/f82c475ac13b2919466a99a8570dc42866c564d6/src/emc/rs274ngc/interp_write.cc#L110 


So  'activating G43 with zero offsets' does not actually 'deactivate' tool compensation it only makes 'G49' appear in the active gcodes list because all internal tool offset values are zero.

zz912:
Thank you for looking into this.

I agree that for LCNC developers this sentence "Activating G43 with zero offsets deactivates tool compensation." is inaccurate. However, for a CNC machine operator it is correct.

For LCNC developers, the sentence should be "Activating G43 with zero offsets makes 'G49' appear in the active gcodes list." However, a CNC machine operator will not even know what a "gcode list" is.


We are currently in a situation where LCNC developers should decide whether:
"Activating G43 with zero offsets makes 'G49' appear in the active gcodes list."
 is **correct behavior** or whether it is a **bug.**

If this is correct behavior, it should be added to the documentation and this PR should be merged.
If this is a bug, it should be fixed and this PR shouldnt be merged

I consider myself an LCNC tester, not a developer. Both options are acceptable to me.


BTW:
I searched for these lines for many hours with ChatGPT and couldn't find it:
https://github.com/LinuxCNC/linuxcnc/blob/f82c475ac13b2919466a99a8570dc42866c564d6/src/emc/rs274ngc/interp_write.cc#L110-L113
These lines prove that LCNC really works with zero values ​​and not with 10E-9.

Sigma1912:
> These lines prove that LCNC really works with zero values ​​and not with 10E-9.

What made you think that it used 10E-9 for tool offsets?

zz912:
> > These lines prove that LCNC really works with zero values ​​and not with 10E-9.
> 
> What made you think that it used 10E-9 for tool offsets?

https://github.com/LinuxCNC/linuxcnc/pull/3792#discussion_r2778136620

rmu75:
> Activating G43 with zero offsets deactivates tool compensation.
> In this situation, G43 behaves like G49.
> If the CNC operator is not aware of this, it can have dangerous consequences.

I wonder what dangerous consequences that could have? If tool offsets are 0, then there are no offsets, and it doesn't matter if machine displays "G43" or "G49".

zz912:
> I wonder what dangerous consequences that could have? If tool offsets are 0, then there are no offsets, and it doesn't matter if machine displays "G43" or "G49".

This situation can occur when you use a CNC machine as a conventional machine. It is the case when you use manual mode for machining as well.

You have for example:
Tool number 1 offset = 0
Tool number 2 offset = 5.256
Tool number 3 offset = 0.5
Tool number 4 offset = -3
Tool number 5 offset = -50

1) You have g43 activated and work with tool 3. You are watching DRO, everything is OK.
2) In MDI you change the tool M6T4. Gmoccapy has automatic G43. So the offsets are updated, you don't worry about anything and the offsets are updated. Everything is OK.
3) In tooltable page you change the offset for tool 4 to -3.5. You leave the tooltable page and get a message asking if you want to update the offset. You click Yes, the offsets are updated. Everything is OK.
4) In MDI you change the tool M6T1. Gmoccapy has automatic G43. However, it turns out that G49 appears in gcode.list.
5) In tooltable you change the offset of tool 1 to -45. You leave the tooltable page and you don't get any message, because G49 appears in gcode.list (state before this PR)
6) You don't notice that G49 is active. You look at the DRO and then you crash into the vice.

Another situation:
You are LinuxCNC beginner or student.  You are learning work with tooloffset.
1) You work only with tool number 5
2) In tooltable page you change the offset for tool 5 to -3.545. You leave the tooltable page and get a message asking if you want to update the offset. You click Yes, the offsets are updated. Everything is OK.
3) You have found that you have measured the offsets incorrectly. You will return to the tooloffset page.
4) In tooltable page you change the offset for tool 5 to 0. You want to try what happens. You leave the tooltable page and get a message asking if you want to update the offset. You click Yes, it turns out that G49 appears in gcode.list.
5) You will return to the tooloffset page.
6) In tooltable you change the offset of tool 5 to -45. You leave the tooltable page and you don't get any message. (state before this PR)
7) You don't notice that G49 is active. You look at the DRO and then you crash into the vice.

I think this is a problem mainly for Gmoccapy users who are spoiled by automatic G43.
This PR makes it so that if you keep pressing Yes, you will always have correct tool compensation.
"Activating G43 with zero offsets deactivates tool compensation." is not dangerous if you know about it **and** if the GUI warns you about it.

rmu75:
Thanks for the explanation, I understand how that can be surprising and annoying. Still not sure if there is something in the interpreter that should be changed, allowing an active G43 in case tool offsets are zero would be an easy change.

Sigma1912:
>  allowing an active G43 in case tool offsets are zero would be an easy change.

I would be in favor of changing it.

IMO it IS a bit problematic to present G43, G43.1, G49 as being modal and then silently change it to G49 when the tool offset values happen to be all zero. This is indeed rather unexpected behavior. 
https://www.linuxcnc.org/docs/html/gcode/overview.html#gcode:modal-groups

gmoccapy:
I have a different opinion.
IMHO a tool change with an active G43 should lead to a change and apply after the change directly the new tool offsets!

Tool 1 is in spindle with G43 active and Z-Offset = 22 mm
change to tool 2 with Z-Offset of 75 mm should bring after the tool change
Tool 2 is in spindle with G43 active and Z-Offset = 75 mm

Using only AUTO Mode with generated code by Cam will add a G43 after each tool change, but how often do we change a tool in Manual mode and than we need to take care to enter again G43 to get the correct values.

I do not know, if my wishes are utopia, but I like to dream ;-)
I did implement this behavior in gmoccapy and it did work for years without problems, until one user got problems mixing Halui commands with gmoccapy code. I did never expect that this would lead to random race conditions

Imho G43 can be active, even if the tool offsets are "zero", My touch probe have all offsets to zero and all my tools have there offsets.

Norbert.

zz912:
Hello Norbert.

> I have a different opinion.
IMHO a tool change with an active G43 should lead to a change and apply after the change directly the new tool offsets!

I think you misunderstood the functionality of this PR.

Several Gmoccapy users had problems with race conditions caused by the integrated AUTOMATIC_G43 in gmoccapy.py, not just me.
The integrated AUTOMATIC_G43 in gmoccapy.py was replaced by AUTOMATIC_G43 using the REMAP function. REMAP with AUTOMATIC_G43 was added to all sim configurations and was even added to the PNCconf Wizard. The warning about the end of the integrated AUTOMATIC_G43 in gmoccapy.py was also implemented. I think we did everything to continue your idea of ​​automatic offsets.

I think we have SAME opinion.
- I don't think anyone here is questioning the benefits of AUTOMATIC_G43
- I don't think anyone here wants to disturb AUTOMATIC_G43
- I don't question the benefits of AUTOMATIC_G43
- I don't want to disturb AUTOMATIC_G43

This PR does not apply to AUTOMATIC_G43 when changing a tool. This PR applies to AUTOMATIC_G43(popup message) when changing the offset for the same tool in the tooltable page. We continue with AUTOMATIC_G43 in other situations.

> Imho G43 can be active, even if the tool offsets are "zero", My touch probe have all offsets to zero and all my tools have there offsets.

Could you please send me the version of LCNC you are using and the tool.tbl you are using? I can double check, but I think it is impossible. I made video in the past: https://github.com/LinuxCNC/linuxcnc/issues/2613#issuecomment-1689913426

rmu75:
> Could you please send me the version of LCNC you are using and the tool.tbl you are using? I can double check, but I think it is impossible. I made video in the past: [#2613 (comment)](https://github.com/LinuxCNC/linuxcnc/issues/2613#issuecomment-1689913426)

Norbert probably means that there is no reason this behaviour would be harmful, not that it currently is that way.

I don't want to add further to the confusion and am going to implement that G43 can be active with all offsets identical zero.

Sigma1912:
See https://github.com/LinuxCNC/linuxcnc/pull/3816

[edit]
update link to new PR

rmu75:
@Sigma1912  was faster it seems ;)

hansu:
@zz912 I guess this PR needs to be updated because of #3816, right?

zz912:
We have 2 choice:
1) close this PR
2) add only 1 new message:
```
            elif new_offset_is_nonzero and old_offset_is_zero and ("G49" in self.active_gcodes):
                message = _("Offset values for the tool in the spindle\n"
                            "have been changed from zero to non-zero.\n\n"
                            "Do you want to activate tool compensation (G43)\n"
                            "using the currently active tool offset?")
```

https://github.com/LinuxCNC/linuxcnc/pull/3816 solved source of this problem.

So, it is up to you. I can modify it.

hansu:
> 2. add only 1 new message:

Please modify this PR

zz912:
PR is ready for merge with 1 new message.

PR #3808, opened 2026-02-15
Ruckig-motivated adjustments to debian/copyright

smoe:
Also some partial alphabetical ordering and removal of redundant copyright notices.


smoe:
This needs to be backported for the next upload to Debian.

andypugh:
As far as I know the Ruckig stuff is not in 2.9, which is the version on Debian. So why does that need the update?

smoe:
> As far as I know the Ruckig stuff is not in 2.9, which is the version on Debian. So why does that need the update?

It does not. Corrected. Thank you!

PR #3809, opened 2026-02-16
Fix heartbeats from task and motion

BsAtHome:
The motion controller has a heartbeat that was apparently addressed in task, but that heartbeat counter was never updated or propagated. Task's status line, when LCNC exists, always printed `task: 0 cycles, ...` because of this. Using motion controller's heartbeat in task was wrong anyway because it is supposed to track task's main loop.

This PR fixes this by adding a `taskbeat` to task's main loop and that is used to print the status line at the end at exit. Additionally, the taskbeat is exported in the NML shared emcStatus in the `task` structure.

Secondly, the motion controller's `heartbeat` is now correctly exported in the NML shared emcStatus in the `motion` structure. Motion controller's heartbeat is necessary to track the RT loop (servo-thread) because status changes can be delayed by one or more RT loops, causing potential for races between RT and non-RT. Non-RT needs to be able to track RT's progression. Example: setting TELEOP mode automatically happens after homing all joints, but that takes one extra loop of the motion controller for it to become set. Sending the next command can therefore fail if you do no wait appropriately.

Also, the heartbeats are now 64-bit (which became possible after the NML update() fixes). That should leave time for about half a billion years of uptime at 1 kHz rates.

### DIFF
--- a/src/emc/motion/motion.h
+++ b/src/emc/motion/motion.h
@@ -60,6 +60,8 @@ to another.
 #ifndef MOTION_H
 #define MOTION_H
 
+#include <stdint.h>
+
 #include "posemath.h"		/* PmCartesian, PmPose, pmCartMag() */
 #include "emcpos.h"		/* EmcPose */
 #include "cubic.h"		/* CUBIC_STRUCT, CUBIC_COEFF */
@@ -623,7 +625,7 @@ Suggestion: Split this in to an Error and a Status flag register..
 /*! \todo FIXME - all structure members beyond this point are in limbo */
 
 	/* dynamic status-- changes every cycle */
-	unsigned int heartbeat;
+	uint64_t heartbeat;     /* Incremented every time the motion controller is done. */
 	int config_num;		/* incremented whenever configuration
 				   changed. */
 	int id;			/* id for executing motion */
--- a/src/emc/motion/usrmotintf.cc
+++ b/src/emc/motion/usrmotintf.cc
@@ -370,7 +370,7 @@ void usrmotPrintEmcmotStatus(emcmot_status_t *s, int which)
 	    );
 	printf("cmd:          \t%d\n", s->commandEcho);
 	printf("cmd num:      \t%d\n", s->commandNumEcho);
-	printf("heartbeat:    \t%u\n", s->heartbeat);
+	printf("heartbeat:    \t%lu\n", s->heartbeat);
 /*! \todo Another #if 0 */
 #if 0				/*! \todo FIXME - change to work with joint
 				   structures */
--- a/src/emc/nml_intf/emc.cc
+++ b/src/emc/nml_intf/emc.cc
@@ -881,7 +881,7 @@ void EMC_AUX_STAT::update(CMS * cms)
 */
 void EMC_TASK_STAT_MSG::update(CMS * cms)
 {
-    cms->update(heartbeat);
+    cms->update(taskbeat);
 }
 
 /*
@@ -1847,6 +1847,7 @@ void EMC_MOTION_STAT::update(CMS * cms)
     EmcPose_update(cms, &eoffset_pose);
     cms->update(numExtraJoints);
     cms->update(jogging_active);
+    cms->update(heartbeat);
 }
 
 /*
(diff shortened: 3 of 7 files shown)


snowgoer540:
Well done.  I had some time to play with this evening.  In order to use it, I needed to added heartbeat and taskbeat to emcmodule.cc.  Were you planning to add it later?  If not, do you want to update this pull request, or I am happy to add it later after this is merged..  

Also, do you plan to update the documentation surrounding this?

BsAtHome:
I guess that it would be alright if you make a PR after this has been committed, if you can solve the problem described below.
Anyhow, these heartbeat counters are only relevant (especially motion controller's) to prevent some specific RT vs. non-RT races afaics. Adding these to emcmodule.cc needs a use case. I guess you have a use case in mind?

With respect to the documentation,... is the rest of EMC_STAT documented?

The emcmodule.cc interface does expose the status. Whether this is a complete mapping of EMC_STAT, I doubt it. If you want to add it, then you must map it to an unsigned integer of exact 64-bit size in python. That you'll have to find out how to do in a <i>portable</i> way. Python does not expose T_* types for uint64_t explicitly afaik and uses a generic T_UINT and T_ULONG and is a bit vague about the actual size.
Note: We just changed NML to use only explicit size integers ([u]int{8,16,32,64}_t) in the update() routines and let the compiler sort it out how that maps because we had problems with 64-bit values.

snowgoer540:
> I guess that it would be alright if you make a PR after this has been committed, if you can solve the problem described below. Anyhow, these heartbeat counters are only relevant (especially motion controller's) to prevent some specific RT vs. non-RT races afaics. Adding these to emcmodule.cc needs a use case. I guess you have a use case in mind?

Speaking candidly, I am not sure I will get around to implementing anything, I was talking loosely with Phill about it, and the initial thoughts surrounded interactions that were influenced by a halui pin.  Keeping track of the elapsed number of cycles for determining the exact position when an arc_ok signal was lost could be another use case.  Rapid jog direction changes, state/mode changes, internal MDI commands with offset changes and motion, etc.  Overall it seems like a pretty good tool to help with race conditions between rt and non-rt.

> 
> With respect to the documentation,... is the rest of EMC_STAT documented?

Sorry, I was speaking as if you were going to add it to emcmodule, I was thinking of the documentation surrounding the linuxcnc.stat attributes here: https://linuxcnc.org/docs/devel/html/config/python-interface.html#_linuxcnc_stat_attributes

That said, it would be up to me to add :)
> 
> The emcmodule.cc interface does expose the status. Whether this is a complete mapping of EMC_STAT, I doubt it. If you want to add it, then you must map it to an unsigned integer of exact 64-bit size in python. That you'll have to find out how to do in a portable way. Python does not expose T_* types for uint64_t explicitly afaik and uses a generic T_UINT and T_ULONG and is a bit vague about the actual size. Note: We just changed NML to use only explicit size integers ([u]int{8,16,32,64}_t) in the update() routines and let the compiler sort it out how that maps because we had problems with 64-bit values.

In my quick testing tonight I used T_ULONGLONG, but I will plan to use getter functions with PyLong_FromUnsignedLongLong() instead.  Unless you object?

BsAtHome:
You have thought about it, let the code flow.

> PyLong_FromUnsignedLongLong

I can see you can simply add it as a getter function in `Stat_getsetlist` and add the appropriate function.

For older python versions it seems appropriate to use ulonglong, but from python 3.14 onward you should use `PyLong_FromUInt64()` as an exact match for the source type.

snowgoer540:
> For older python versions it seems appropriate to use ulonglong, but from python 3.14 onward you should use `PyLong_FromUInt64()` as an exact match for the source type.

That is good to know.  

Looks like master currently supports 3.8-3.14, are you suggesting that when we get to 3.14 as the lowest version this change to FromUInt64, or something else?

If so, I’m not sure how to keep this from getting lost then (other than add a comment about it).  Open to suggestions.

BsAtHome:
Support must be from python 3.8 (older debians). Current debian uses 3.13. Fedora 43 is at 3.14. The `PyLong_FromUInt64()` function was added to the stable ABI since 3.14.
You can simply test the version:

```c++
#if PY_VERSION_HEX >= 0x030e00f0
  return PyLong_FromUInt64(...)
#else
  return PyLong_FromULongLong(...)
#endif
```

Issue #2237, opened 2023-01-02
xhc-whb04b-6 code is sprinkled with assert() and fails occasionally due to that. Please remove assert() with NDEBUG because the code does the right thing without assert()


HakanBastedt:
## Here are the steps I follow to reproduce the issue:

 1. Download and install linuxcnc on debian as a package in the standard way
 2. Start linuxcnc and use the xhc-whb04b-6. After a random time, hours perhaps, it stops responding
 3. I don't have the error message here but I have searched and found that this is the line that triggers the xhc-whb04b-6 module to exit:
     assert(0 == r); 
line 460 in https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/user_comps/xhc-whb04b-6/usb.cc
in the function 
bool Usb::setupAsyncTransfer()

## This is what I expected to happen:
If it was unable to setup async usb transfer, it should do some retrying or so. 

## This is what happened instead:
The hal module stops. It isn't possible to unload or in any way repair this (afaik).

## It worked properly before this:
(If the behavior changed after making a particular change in hardware or
software, describe the change you think is responsible.  E.g., "after upgrading
from LinuxCNC 2.7.3 to 2.7.4")
I got a new xhc whb04b-4 the other day, so I have always had this problem with the new pendant.

## Information about my hardware and software:
I am running 2.8 from apt archive.
I have also downloaded the developer source code and searched files and recompiled.
The issue is in both those versions.
The solution is simple. Disarm the assert() function with the NDEBUG directive.
Because when this is done, the code does the right thing. It tries to reconnect and re-establish
functionality. 
I have gotten an error in setupasynctransfer, the code handles this fine, and re-established functionality.
I am now running a version with NDEBUG and, yes this is how it should be.

burak1000:
thank you very much for this . i will try these and write here the results. 
have a nice day

petterreinholdtsen:
[HakanBastedt]
> The solution is simple. Disarm the assert() function with the NDEBUG
> directive.  Because when this is done, the code does the right
> thing. It tries to reconnect and re-establish functionality.

For the record, disabling bogus asserts with NDEBUG is the wrong
approach.  If the assert should not trigger, the assert should be
removed or perhaps rewritten as a test logging something instead.  The
behavour of the code should be the same way in debug and no-debug builds.

-- 
Happy hacking
Petter Reinholdtsen

burak1000:
[...]

PR #3810, opened 2026-02-16
xhc-whb04b-6 fixes

hdiethelm:
Hello

I'm pretty new to LinuxCNC, just finalizing my new CNC mill. I hope this PR matches your expectations. Otherwise, just tell me.

I fixed some issues in the xhc-whb04b-6 component:
- The -B argument was not working. I had 5 and 10 mm steps always active which is a bit dangerous
- I had strange movements after changing from MPG to Step or Continuous mode
- Jogging while a program is running gives unnecessary errors
- Rename example-configuration.md to example-configuration.hal (This is a hal file, not an md file)

Somehow, also the Spindle and Feed values are not shown on the display. I was not able to fix this until now. If I find the issue, I will create a separate PR.

### DIFF
--- a/src/hal/user_comps/xhc-whb04b-6/pendant.cc
+++ b/src/hal/user_comps/xhc-whb04b-6/pendant.cc
@@ -631,7 +631,7 @@ void FeedRotaryButton::update()
         mStepSize    = mStepSizeMapper.getStepSize(second);
         mIsPermitted = mStepSizeMapper.isPermitted(second);
         
-        if (!mIsStepMode_5_10 && mStepSize > 2) {mStepSize    = 0;}
+        if (!mIsStepMode_5_10 && mStepSize > 2) {mStepSize    = 1.0;}
         
     }
     else if (mStepMode == HandwheelStepmodes::Mode::CON)


hdiethelm:
I just fixed an other issue: Jogging while a program is running always gave an error. There is no reason to try to change to manual mode when the mode is auto.

hdiethelm:
I tested in Mach3 in a Windows VM with the original plugin from https://www.cdxhctech.com/html/edownloads/view_420.html and the display also doesn't show any spindle speed or feed. So I think might be a firmware bug of my device. So I leave that for now.

I created a back-port to the latest release if someone (like me) doesn't want to test all new changes on his new CNC: https://github.com/hdiethelm/linuxcnc-fork/tree/xhc-whb04b-6-fix-v2_backport-v2.9.8

andypugh:
The documentation seems to suggest that there is something odd in the firmware that means that Lead mode just inherently acts strangely. However I have not used an XHC pendant, so I can't be sure. 
I just find myself wondering if the original author had good reasons for not making the functions use the enable parameter.

I think that halui-based jogging requires manual mode with some user interfaces. Which ones have to tested with? Again, this is not something that I have tested recently. Axis springs to mind.

hdiethelm:
> The documentation seems to suggest that there is something odd in the firmware that means that Lead mode just inherently acts strangely. However I have not used an XHC pendant, so I can't be sure. 

I did not find any reference to a not working lead mode in the doc?

It took me some time to understand how this code works, at least parts of it. It seams the MPG mode is only used together with the -f flag for feed and the lead mode together with the -s flag for spindle override. This matches also the doc.

These overrides are not well implemented in the whb04b code, just every second wheel click is used and only up to a certain speed. Using halui.feed-override.counts and halui.spindle.N.override.counts would be better but I think I won't implement that right now and instead just use the feed/spindle +- knobs.

The lead mode in Mach3 is used to go forward or backward in steps in the g-code. Is this even possible in LinuxCNC?

The firmware is a bit strange. Some parts are done in the device, for example showing the asterisk for the selected axis or the % for continuous / step size for step mode. And some parts are commanded from the PC, like changing reset/step/mpg/lead mode. There are some bugs in this logic. Step is shown as 1, even if 60% or 100% are selected which are an undefined step size. Or the X axis has an asterisk, even if switched to off.

>I just find myself wondering if the original author had good reasons for not making the functions use the enable parameter.

I don't think so, the only reference I found is from main over WhbComponent to Pendant. It looks more like a mistake not found in testing due to this functions are only used with "true" as an argument in the above path.

> 
> I think that halui-based jogging requires manual mode with some user interfaces. Which ones have to tested with? Again, this is not something that I have tested recently. Axis springs to mind.

The jogging uses the axis pins from https://linuxcnc.org/docs/html/man/man9/motion.9.html and some halui pins. Manual mode is activated using halui.mode.manual but this doesn’t work when in auto mode (program running) and gives errors. That’s the reason for one of my changes. As soon as the program ended, manual mode is activated and you can continue jogging. When running an MDI command, the mode changes to mdi and reverts back to manual when jogging like before my change.

I tested two variants: 
- A simulation version of my CNC mill, Axis UI using the config found in https://raw.githubusercontent.com/LinuxCNC/linuxcnc/refs/heads/master/src/hal/user_comps/xhc-whb04b-6/example-configuration.md with some small changes for my setup. This file is not really md format, I guess that should be renamed to .txt or .hal?
- A simple hal file together with halshow to poke signals and for fast testing
```
#loadusr -W xhc-whb04b-6 -HsfB
#loadusr -W xhc-whb04b-6 -HsfB
loadusr -W xhc-whb04b-6 -H

setp whb.halui.machine.is-on 1
setp whb.halui.mode.is-teleop 1
setp whb.halui.mode.is-manual 1
setp whb.halui.max-velocity.value 40
```
I have no access to my real mill this week, but I will test it there next week with the back-port to 2.9.8

hdiethelm:
Hmm, just testing a bit more, I found there is still an issue. Might be not whb04b related. The following signals can also be set without the whb04b loaded.

1. Just after start and homing
2. axis.x.jog-vel-mode = false / axis.x.jog-scale = 0.0 / axis.x.jog-enable = true / axis.x.jog-counts = 0 
3. With whb04b, counts increases to a value but for reproducing, you can just set it
4. axis.x.jog-vel-mode = false / axis.x.jog-scale = 0.0 / axis.x.jog-enable = true / axis.x.jog-counts = -260 
5. axis.x.jog-vel-mode = true / axis.x.jog-scale = 0.12 / axis.x.jog-enable = true / axis.x.jog-counts = -260
6. axis.x.jog-counts changes to -261
7. Machine moves with full speed from x=325 to x=10

That happens exactly once per axis after init. After that, it is fine. Any clue why this is?

I figured I fixed this issue by removing the step from 0 to -260 at the first wheel click by sending the counter always, but I obviously did not test it well enough. The order of steps to reproduce it with the mpg is a bit cumbersome.

hdiethelm:
After some printf debugging, I found that axis->pos_cmd is 0.0 in axis.c line 421 and following. https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/motion/axis.c#L421 in this special case due to axis->teleop_tp.curr_pos is 0.0 a bit below.
This results in the axis moving to 10 mm instead of 315 mm.

Might be axis->teleop_tp.pos_cmd should be used instead, but I really don't know this code.

I found an easier way to reproduce. Note that on my machine, the home switch is at 330 mm.

1. Just after start and homing -> x is at 325 mm
2. axis.x.jog-vel-mode = true / axis.x.jog-scale = 0.12 / axis.x.jog-enable = true
3. Set axis.x.jog-counts = -1 -> Machine moves with full speed from x=325 to x=10

If I do any other move before jogging, all is fine.

This works but I won't bet that there are no unwanted side effects: https://github.com/hdiethelm/linuxcnc-fork/commit/b52a755e1a30e27ee2d9148c76fab4e70978ecc2

Note:
This is only about continuous mode. My fix with always sending the counter is still needed for step mode. Without it, the machine moves unexpectedly when the wheel is moved in mpg mode and then step mode is activated, after the first wheel click.

hdiethelm:
I went a bit further in the velocity jog run away issue. It seams the reason behind is in https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/motion/axis.c#L677 returning true due to the axis is at the limit, so axis->pos_cmd is not updated after switching to manual mode.
Now I can even reproduce the issue in axis_mm example after homing in any state.

1. G53 G0 X-254 (Or X254 or any other of the of the machine limits)
2. Switch to manual mode and jog in velocity mode in the same axis +1 or -1:
- axis.x.jog-vel-mode = true / axis.x.jog-scale = 0.12 / axis.x.jog-enable = true
- Set axis.x.jog-counts = +1 or -1
3. The machine will move back to center with full speed

Should I create a separate issue for this?

Sigma1912:
> Should I create a separate issue for this?

Yes, please.

hdiethelm:
> > Should I create a separate issue for this?
> 
> Yes, please.

Done: https://github.com/LinuxCNC/linuxcnc/pull/3820

andypugh:
> The lead mode in Mach3 is used to go forward or backward in steps in the g-code. Is this even possible in LinuxCNC?

I am not sure what you are describing. It is possible to run the G-code backwards in LinuxCNC by setting a negative feed-override. Is that what you mean? 

Thinking about this in general, I am wondering how many versions of the XHC pendant there are, and whether they all use the same firmware (or even processor). I am wary of chasing a moving target with a succession of XHC owners fixing the LinuxCNC driver to work with their particular version. 

I have considered buying an XHC purely for testing with, but so far have baulked at the expense for something that I don't need. 

In the absence of anything to test with I am happy to merge your changes if you are confident that they work, and have a reasonable expectation that they are compatible with however many variants there are out there.

hdiethelm:
> > The lead mode in Mach3 is used to go forward or backward in steps in the g-code. Is this even possible in LinuxCNC?
> 
> I am not sure what you are describing. It is possible to run the G-code backwards in LinuxCNC by setting a negative feed-override. Is that what you mean?

Ah that would be possible to do also in LinuxCNC? I just quickly tested the XHC in Mach3 to figure out what this lead mode is due to the manual doesn't say anything detailed. And ^^ is what it does there.

In the XHC code, the lead counter is available but I did not find any real use of it, it is not even sent to the HAL.

> 
> Thinking about this in general, I am wondering how many versions of the XHC pendant there are, and whether they all use the same firmware (or even processor). I am wary of chasing a moving target with a succession of XHC owners fixing the LinuxCNC driver to work with their particular version.
> 
> I have considered buying an XHC purely for testing with, but so far have baulked at the expense for something that I don't need.
> 
> In the absence of anything to test with I am happy to merge your changes if you are confident that they work, and have a reasonable expectation that they are compatible with however many variants there are out there.

I understand that. At least there are four versions (3/6 axis and wired/wireless). And it seams mine (wired, 6 axis) doesn't show the F/S fields properly while others sometimes do.

That's why I tried to correct only the obvious issues without doing anything else better that could be improved like for example exposing the lead counter, changing the -s and -f options to use counters instead of pulses and so on.

I am pretty confident my changes work also for other people due to they are mostly pretty basic but I can not be 100% sure about that. Because the -B option never worked before and now does, that could generate some confusion.

Now I just use A/B/C to control feeds and the spindle which works pretty nice with some small HAL changes: https://forum.linuxcnc.org/38-general-linuxcnc-questions/58322-xhc-whb04b-development?start=70#343571

In my opinion, the best option would just be replacing the whole firmware on the device and create a new linuxcnc module. But that's to much effort for me for not enough gain.

Sigma1912:
Not sure if this is relevant here but there is a recent post on the forum about a fix for the 'lead' function (whatever that is): 
https://forum.linuxcnc.org/42-deutsch/58442-handrad-xhc-whb04b-6-braucht-9b-fuer-lead-funktion

andypugh:
After yesterdays online developer discussion, I have decided to buy one of the pendants to test with. 

Thanks for your work, but there will be a week or more delay before I feel I can merge it.

andypugh:
> In my opinion, the best option would just be replacing the whole firmware on the device and create a new linuxcnc module. But that's to much effort for me for not enough gain.

I did find this, though it seems to be a complete replacement rather than alternate firmware:
https://github.com/moonglow/openxhc
Which gives this as a link to the "original" XHC:
https://www.cdxhctech.com/

hdiethelm:
> Not sure if this is relevant here but there is a recent post on the forum about a fix for the 'lead' function (whatever that is): https://forum.linuxcnc.org/42-deutsch/58442-handrad-xhc-whb04b-6-braucht-9b-fuer-lead-funktion

This is not really relevant to my changes. https://www.linuxcnc.org/docs/2.8/html/man/man1/xhc-whb04b-6.1.html -s option describes the same. However, internally the increase/decrease pins instead of counts is used. counts would be better. I could do that if you think it's useful, it should not be that difficult but it will be a breaking change.

> After yesterdays online developer discussion, I have decided to buy one of the pendants to test with.
> 
> Thanks for your work, but there will be a week or more delay before I feel I can merge it.

Thanks for your effort. Just tell me if you have some issues. It took me some time to get everything up and running and also to properly reproduce the issues.

hdiethelm:
BTW: If you want to back port this to 2.9, you need pieces of three other commits from BsAtHome or it won't work. See: https://github.com/hdiethelm/linuxcnc-fork/commits/xhc-whb04b-6-fix-v2_backport-v2.9.8/

hdiethelm:
> I did find this, though it seems to be a complete replacement rather than alternate firmware: https://github.com/moonglow/openxhc Which gives this as a link to the "original" XHC: https://www.cdxhctech.com/

That's for the HB04 which is an other device. It might be similar but I have no clue. It's a bit a mess. Might be one reason why some people have issues. I also first tried the HB04 driver without sucess... ;-)
There are: 


- XHC WHB04B (Wireless, works with this module) https://www.cdxhctech.com/html/eProduct/view_363.html
- XHC LHB04B (Wired, works with this module) https://www.cdxhctech.com/html/eProduct/view_426.html
- XHC HB04 (Uses: https://linuxcnc.org/docs/devel/html/man/man1/xhc-hb04.1.html) I think it's that one: https://www.eaglecnctec.com/products/181-mach3-pendant-xhc-hb04.html

finngineering:
I also have a XHC WHB04B-6 wireless pendant, and have put quite much time and effort into understanding how it works and improving it. Here is a link to the forum thread:
[https://forum.linuxcnc.org/38-general-linuxcnc-questions/58322-xhc-whb04b-development](https://forum.linuxcnc.org/38-general-linuxcnc-questions/58322-xhc-whb04b-development)
And here is a link to my open source firmware for the wireless dongle:
[https://github.com/finngineering/whb04b](https://github.com/finngineering/whb04b)

I have also just now reviewed the changes by hdiethelm in this PR. Of course, I have no affiliation with LinuxCNC as such, but I think the comments will anyway be useful. Based on the commit messages:

**whb04b: Pass arguments properly**
**whb04b: Fix inversion of -B argument**
These changes are all good. The 5 and 10 steps were always enabled in the past, and now it requires the -B argument to enable them. This only limits the maximum step side if specified, and should therefore be safe for everybody. The -f and -s arguments worked also in the past, because the mIsLeadModeFeed and mIsLeadModeSpindle were initialized to false. And the setLeadMode[Feed,Spindle] were only called in case the corresponding command line options were specified. The changes are anyway an improvement to the code in general.

**whb04b: Don't change from auto to manual when jogging**
If you search the forums, you see than many people have had problems with the RF pendant disconnecting. I believe ~90% of those disconnects are because we try to change mode and wait a long time before "realizing" that the mode change was unsuccessful. During that time, the pendant interrupt endpoint is not polled, which causes it to disconnect. So I think this is a very important change to implement. However, I think it would be better to use memory->in.isProgramIdle as a requirement for mode change rather than memory->in.isModeAuto. Axis seems to not be in auto mode normally, but gmoccapy is if you are in the "program" tab. Using the isProgramIdle as a requirement for change to manual means you can start jogging from the "program" tab also in gmoccapy, unless a program is running or paused.

**whb04b: Fix example-configuration file ending**
Great. That has been bugging me when trying to view the example configuration on Github.

**whb04b: Remove unused code
whb04b: Fix machine moving after change from MPG to STEP or CON**
I have not really checked these, but would believe both are good changes.


One more comment is that the "Lead" mode did not work on my pendant. My pendant also uses the 0x9b key code for the lead mode. It appears that the key mapping is missing here:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/user_comps/xhc-whb04b-6/pendant.cc#L345
So just add another line with `{lead.code9B,        &lead9B}`, and it started working. How the lead mode feed override works is kinda crazy to me, but that is all from the "original" implementation and in agreement with the documentation.

Great to see work on this! There are many other parts that could be improved (for instance we could fix the checksum calculation). But one step at a time.

hdiethelm:
Thanks for your feedback!

> ...would be better to use memory->in.isProgramIdle as a requirement for mode change rather than memory->in.isModeAuto...

Good point. I changed this. It also improves the behavior in MDI mode: The command is not aborted any more when you jog. This helps especially when wiring ABC to the feeds.

> One more comment is that the "Lead" mode did not work on my pendant. My pendant also uses the 0x9b key code for the lead mode. It appears that the key mapping is missing here: https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/user_comps/xhc-whb04b-6/pendant.cc#L345 So just add another line with `{lead.code9B, &lead9B}`, and it started working. How the lead mode feed override works is kinda crazy to me, but that is all from the "original" implementation and in agreement with the documentation.

I also pushed a fix for this. I tested it using the following code in usb.cc line 530 and removing some const. Mine sends 0x1C.
```
if(inputPackageBuffer.asFields.rotaryButtonFeedKeyCode==0x1c){
   inputPackageBuffer.asFields.rotaryButtonFeedKeyCode=0x9B;
}
```
Feel free to test the changes. I also updated the back port branch:
https://github.com/hdiethelm/linuxcnc-fork/tree/xhc-whb04b-6-fix-v2_backport-v2.9.8

finngineering:
Thank you for the changes. I tested them and they work as expected!

MetalMusings:
The issues I had https://github.com/LinuxCNC/linuxcnc/issues/2237 seems to be fixed with this PR.

hdiethelm:
That's nice. My main attempt was not to fix this issue, I was just annoyed of the long list of error messages popping up in the axis UI that I had to click away manually.

Do you have a wired or wireless device? Mine is wired, marked LHB04B-6 / TX V3 on the back and I did not have any issues with disconnects so far. I just checked the asserts. A few are really bad, especially that one:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/user_comps/xhc-whb04b-6/usb.cc#L465
I can not really test it, because mine is working well, but I can at leas try to do better error handling.

As a basic principle, you should never do asserts() for error handling. assert() should only be used for debug purpose to check for programming mistakes.

The issue with waiting for signals in the HAL is not 100% fixed with my PR. There are still other usleep() there for toggling signals and a timeout wait can still happen. However, the way the code is structured, it's not that easy to really solve that. One option would be to add states to the hal.cc and another one to add a separate USB thread. But I won't do that in this PR, so it doesn't get to big. Might be an other one after this is merged.

Can you add an usleep() after this line inside the while loop? https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/user_comps/xhc-whb04b-6/xhc-whb04b6.cc#L515
And slowly increase the delay until you have issues to see if sleeping is really the root cause of the issues you are having? I would go the binary search way: Start with usleep(100*1000) and the halve or double it to find the delay which creates issues. That will help do determine if the usleep() is really an issue.

hdiethelm:
@andypugh
I tried to improve error handling: https://github.com/hdiethelm/linuxcnc-fork/compare/xhc-whb04b-6-fix-v2...hdiethelm:linuxcnc-fork:xhc-whb04b-6-fix-v2-errorhandling
Just tell me if you are fine with this changes and I will merge it into this PR. It went a bit out of hands, but the existing code was not build with error handling in mind. It's still not perfect but at least I hope it should do the job.

andypugh:
I have been playing about with the XHC today.
It's weird. Mainly this is due to the hardware, and trying to make sensible choices from that. 

How hard a job would it be to have 5 and 10 positions on the jog speed give 1 in the case that the B flag is not used? It seems odd that those positions become useless. 

"lead" is spindle override (I don't know what the manufacturers intended it to be, but fair enough). But I have't figured out how you are _meant_ to get from "lead" (spindle) to "MPG" (feed) override and back to jogging. 
With your changes it seems to be being "not in lead" on the right-side selector, with a program running. And then you get back to jogging by pressing Continuous or Step when the program is stopped. I guess that makes sense, and is at least likely to avoid accidental jogs to some extent. 

The whole Continuous / Step thing is odd. I have never felt the need for anything but Step mode on my machines. (I realise that these choices were not made by you)

hdiethelm:
> I have been playing about with the XHC today. It's weird. Mainly this is due to the hardware, and trying to make sensible choices from that.

Yes,  it took me also some time to get used to. But it's the only MPG working mostly plug&play for a good price as far as I know.

> 
> How hard a job would it be to have 5 and 10 positions on the jog speed give 1 in the case that the B flag is not used? It seems odd that those positions become useless.

Should be easy, just change the 0 to 1.0 in:
`if (!mIsStepMode_5_10 && mStepSize > 2) {mStepSize    = 0;}`
`if (!mIsStepMode_5_10 && mStepSize > 2) {mStepSize    = 1.0;}`
I can do that and test it.

> 
> "lead" is spindle override (I don't know what the manufacturers intended it to be, but fair enough). But I have't figured out how you are _meant_ to get from "lead" (spindle) to "MPG" (feed) override and back to jogging. With your changes it seems to be being "not in lead" on the right-side selector, with a program running. And then you get back to jogging by pressing Continuous or Step when the program is stopped. I guess that makes sense, and is at least likely to avoid accidental jogs to some extent.

I changed nothing concerning lead/mpg, it is the same on master. You have to set -s and -f to have this connected to spindle / feed. I only changed that the component does not try to go to manual mode while a program is running. When jogging while a program is running, this just give you a ton of annoying warnings on master and this warnings generate a delay in the whb code resulting in disconnects for people using the wireless version as mentioned in the forum.

Trying Mach3, the manufacturer intended that the Lead mode to control the program flow: One step forward goes to the next line in gcode, one step back to the previous. MPG is the feed.

This feels all a bit crucked:

- To go to MPG: Go to lead and then back.
- Then click cont/step if you want to jog

This and the crooked pulse implementation in whb is why I am now just using axis A,B and C for feeds and speeds:
[xhc_whb04b_6.hal.txt](https://github.com/user-attachments/files/26321455/xhc_whb04b_6.hal.txt)

The mpg and lead mode would have to be changed to counters but this is not that easy with the existing code and I won't need it anyway due to it is annoying "go to lead then back" just to have feed mode active. Might be one day when I need the A/B/C axis for something else.

> 
> The whole Continuous / Step thing is odd. I have never felt the need for anything but Step mode on my machines. (I realise that these choices were not made by you)

I don't think so. At least for me it feels good.

- Step: When you want to move precisely: Move 5mm forward -> 1mm, 5 clicks
  - However, when you want to move fast, it doesn't stop right away. When you do 200 clicks, it moves 200mm and you have to press stop to stop.
- Cont: When you want to move coarse: Just jog as fast as you wish and it will halt after the deceleration ramp.
  - However, moving exactly 5mm is difficult.
 - These two modes are implemented in axis.c pin axis.x.jog-vel-mode and probably also used by other MPG's.

MetalMusings:
Just want to add that the issues with crashes I had after removing the assert() statements are related to the gcc version.
gcc 12.2 - component crashes after usb comm fault
gcc 14.2 - no crash. The usb can sometimes fail, but the code will reset and restart the usb comm and the mpg is fully working after a few seconds delay.

hdiethelm:
> Just want to add that the issues with crashes I had after removing the assert() statements are related to the gcc version. gcc 12.2 - component crashes after usb comm fault gcc 14.2 - no crash. The usb can sometimes fail, but the code will reset and restart the usb comm and the mpg is fully working after a few seconds delay.

On which branch did you have this result?

MetalMusings:
Both on master, and your xhc-whb04b-6-fix-v2-errorhandling

andypugh:
> Should be easy, just change the 0 to 1.0 in: `if (!mIsStepMode_5_10 && mStepSize > 2) {mStepSize = 0;}` `if (!mIsStepMode_5_10 && mStepSize > 2) {mStepSize = 1.0;}` I can do that and test it.

If you could, please.

hdiethelm:
Done, it works. I re-based also on master, hope that is fine. If not, I can undo it.
There was also the axis fix that already got already merged. Without that, it confused me a bit.

andypugh:
Thanks

PR #3813, opened 2026-02-19
Fix G43.x - G49 in active gcodes

Sigma1912:
Current behavior shows active gcode 'G49' if current tool offset values are all zero and 'G43' if not. This leads to incorrect information if we use G43 on a tool with all zero offsets and shows 'G43' for all g43.x commands.

With this change we show
'G43', 'G43.1', 'G43.2' after the respective command
'G49' after g49 command 

Note that nothing is shown at startup unless it (eg  'G49')  is found in RS274NGC_STARTUP_CODE.

### DIFF
--- a/src/emc/rs274ngc/interp_write.cc
+++ b/src/emc/rs274ngc/interp_write.cc
@@ -107,10 +107,11 @@ int Interp::write_g_codes(block_pointer block,   //!< pointer to a block of RS27
     (settings->origin_index <
      7) ? (530 + (10 * settings->origin_index)) : (584 +
                                                    settings->origin_index);
-  settings->active_g_codes[9] =
-    (settings->tool_offset.tran.x || settings->tool_offset.tran.y || settings->tool_offset.tran.z ||
-     settings->tool_offset.a || settings->tool_offset.b || settings->tool_offset.c ||
-     settings->tool_offset.u || settings->tool_offset.v || settings->tool_offset.w) ? G_43 : G_49;
+  if (block !=  NULL){
+    if (block->g_modes[gees[G_43]] > 0){
+    settings->active_g_codes[9] =  block->g_modes[gees[G_43]];
+    }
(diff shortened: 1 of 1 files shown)


Sigma1912:
Relates to https://github.com/LinuxCNC/linuxcnc/pull/3792

Sigma1912:
Note that there is similar code for state flags which is not modified in this PR:
https://github.com/LinuxCNC/linuxcnc/blob/f82c475ac13b2919466a99a8570dc42866c564d6/src/emc/rs274ngc/interp_write.cc#L281
https://github.com/LinuxCNC/linuxcnc/blob/f82c475ac13b2919466a99a8570dc42866c564d6/src/emc/rs274ngc/rs274ngc_pre.cc#L2212

Sigma1912:
@rmu75 
I had to remove the 'G49' on startup as this gets also executed on Abort which would lead to 'G49' being shown even with tool offset applied.   
This happens because (block == NULL) is True on startup AND on Abort. 

If you can think of a better approach please let me know.

rmu75:
My initial idea was to add a flag to the offsets, i.e. a "fake" offset.

Sigma1912:
> My initial idea was to add a flag to the offsets, i.e. a "fake" offset.

That was my initial idea as well but then I thought it would be nice to have 'active gcodes' reflect G43.1 and G43.2 as well. 

Do you know of a way to tell if we have a Startup vs Abort event?

rmu75:
That code here https://github.com/Sigma1912/linuxcnc/blob/4b2e51c12bcf470fba63e9d9070295c9f66cb9ab/src/emc/rs274ngc/interp_write.cc#L133 suggests block should be set if it is an abort... if it is not, sorry, no idea.

Sigma1912:
Replaced by https://github.com/LinuxCNC/linuxcnc/pull/3816

Issue #3814, opened 2026-02-19
Bookworm missing C++20 <format> header


BsAtHome:
There is a problem in Debian 12 (Bookworm) where gcc-12 is missing the C++20 `<format>` header. That means that things like the following is not possible:
```C++
#include <format>
....
std::string s = std::format("This string {} is at least {} times better.", "formatter", 10);
// formats into: "This string formatter is at least 10 times better."
....
```
This type of formatting  is also the default in python nowadays and is _much_ better than any snprintf or other bodgeworks.

The solution is to add a dependency `libfmt-dev` for building LinuxCNC to include libfmt (https://github.com/fmtlib/fmt), which is available as packages on all platforms afaik (libfmt7 on bullseye, libfmt9 on bookworm, libfmt10 on trixie; even buster has it).
Using libfmt means that `#include <format>` with `std::format()` just needs to be rewritten as `#include <fmt/format.h>` and use `fmt::format()`.

What do you guys say, is this acceptable?

rmu75:
`std::format` only covers part of what is in libfmt -- IME libfmt is a really high quality library that is easy and intuitive to use and really brings benefit to the code.

BsAtHome:
Fixed with merge of #3815.

PR #3815, opened 2026-02-20
Major overhaul/rewrite of linuxcncrsh

BsAtHome:
This PR addresses the many deficiencies and races in linuxcncrsh. Some of them have been discussed in #3734 and additional fixes have been added in previous PRs (#3809, #3803, #3800, #3791).

The entire linuxcncrsh program is now single-threaded and uses a poll loop to handle input/output. Multiple clients may be connected at the same time and _all_ can be in the ENABLED state and issue SET commands. The limitation is that only _**one**_ command can be in transit via NML for all clients (i.e. only one SET command can be processed at any given time). There are three exceptions; any enabled client can issue 'SET ABORT', 'SET MACHINE OFF' and 'SET ESTOP ON'. Any client can always issue GET commands.

Fixes and additions:

* C-style string handling has been replaced with std::string and C-style arrays now use std::vector (that should fix the buffer overflows ;-) ).
* Many commands have been streamlined in what they allow as arguments so that similar commands use similar arguments (like -1 for 'all' and defaults)
* Output of commands has been streamlined to use identical format for same type of information
* Axis and joint indicators are now checked whether they actually are available and result in an error if they are not
* Error checking is consistently done on arguments and messages returned when something is amiss
* Heartbeat from the servo-thread (motion controller) is used to prevent RT <--> non-RT races
* GET INI - now works and you can read any INI file entry
* GET INIFILE - works and returns path of INI file
* GET PROBE now issues an MDI G38.2 because it [cw]ould not work otherwise.
* GET CLIENTS returns who is connected
* SET DISABLE lets you downgrade own connection from being enabled and also allows _other_ clients to be downgraded
* SET JOINT_HOME (alias: SET HOME) now works in conjunction with SET JOINT_WAIT_HOMED to wait for completion
* SET JOINT_UNHOME lets you unhome any joint
* SET TIMESTAMP lets you enable date+time stamped responses
* Files read/loaded are now using a _functional_ path (as set by the `-d` command-line option). The default is `$HOME/nc_files`. You can pass your own ':' separated list to the `-d` option

Several GET/SET commands were removed or deprecated:

* COMM_MODE - removed - the communication is always ASCII.
* CONFIG - removed - was never implemented
* SET_WAIT - removed - was already deprecated (replaced by WAIT_MODE)
* COMM_PROT - deprecated - API versioning is overrated and is non-functional anyway
* UPDATE - deprecated - emcStatus updates are better left to be automatic
* UPDATE_STATUS - deprecated - emcStatus updates are better left to be automatic

The man-page has been updated to reflect all changes, removals and additions.

There are probably some more details that I forgot to mention here... Well, c'est la vie.

The diff of the source may be inadequate to look at because essentially every line got changed (and that is not because of whitespace changes). See the entire file: https://github.com/BsAtHome/linuxcnc/blob/fix_linuxcncrsh/src/emc/usr_intf/emcrsh.cc for review.

The new version also uses `fmt::format()` to do string formatting, as discussed in #3814, because gcc-12 in Debian 12 (bookworm) does not support std::format (need gcc-14 or newer for that). A configure test and package dependency is added.


grandixximo:
@BsAtHome look at line 2378 in /src/emc/usr_intf/emcrsh.cc

BsAtHome:
> @BsAtHome look at line 2378 in /src/emc/usr_intf/emcrsh.cc

Nice one... s/[joint]/[i]/

But I do appreciate it being less cryptic :-) The bug is in line 2379 according to my editor.
These bugs make you stare for ages before you see them.

However, one bug in 3800 lines of code? That would be my new record. There is this old saying: <i>"One error in every 500 lines of code, regardless language. Fix it and you introduce or uncover a new one to average it out in perpetuity."</i>. I think others have analyzed it to about one bug per 1000 lines, but that is just an insignificant scaling factor. And formal verification becomes rather impossible for anything complex (and linking to other [unverified] code).

BsAtHome:
> > @BsAtHome look at line 2378 in /src/emc/usr_intf/emcrsh.cc

>  The bug is in line 2379 according to my editor.

You were right. Better be on the right branch when checking...

grandixximo:
Sorry, I thought you'd enjoy the hunt, maybe too early in the morning for that 😅

BsAtHome:
> Sorry, I thought you'd enjoy the hunt, maybe too early in the morning for that 😅

Yeah, it is early in the morning...

BsAtHome:
You know, it is actually interesting to see that different compiler versions generate completely different warnings. The bug in emcrsh.cc is actually detected on trixie with gcc 14.2.0 but not on Fedora 43 with (newer) gcc 15.2.1 and not on bookworm with older gcc 12.2.0.

andypugh:
A recent build failed a test for linuxcncrsh. I suspect it isn't directly related to the changes, but rather a subsequent change has altered some behaviour. 
https://github.com/LinuxCNC/linuxcnc/actions/runs/24210647813/job/70678181120

BsAtHome:
> A recent build failed a test for linuxcncrsh. I suspect it isn't directly related to the changes, but rather a subsequent change has altered some behaviour.

Yes, it is in the homing procedure. That has always been a picky and tricky part.

But why it only fails on ARM is a strange case.

I pulled the changes of the revert (#3918) into my tree and that did not fail CI. I guess there is yet another RT/non-RT timing race problem somewhere that I missed. I restarted the abandoned CI and may restart the failed one a bit later to see if it indeed is timing. However, CI is testing on a non-realtime kernel and there may be (unknown) side effects not considered.

(BTW, CI runners have been very "laggy" for the past two days and some processes just stalled for ages)

BsAtHome:
Indeed, restarting the first one aborted almost immediately with an error from the CI infrastructure. A second restart of both missing runs (trixie and sid on arm) was successful.

It suggests there still is a timing race. My guess is that is surfaces when the RT thread (rtapi_app) gets delayed much more than the non-RT thread (task). Theory: this might cause a timeout or sync-mismatch while waiting.

PR #3819, opened 2026-02-21
mesa_modbus.adoc - missing .00 in function name

zz912:
addf my_file servo-thread
is not working

but

addf my_file.00 servo-thread
works

I found it here:
Mesa_modbus

### DIFF
--- a/docs/src/drivers/mesa_modbus.adoc
+++ b/docs/src/drivers/mesa_modbus.adoc
@@ -83,7 +83,7 @@ Load and activate the module with HAL commands such as
 
 ----
 loadrt my_file ports=hm2_7i69.0.pktuart.0
-addf my_file servo-thread
+addf my_file.00 servo-thread
 ----
 
 == Modbus ==


BsAtHome:
You may be right. However, you should not be using the mesa_modbus driver anymore. You should move to `hm2_modbus`.

There is a new driver that is much more capable. See https://linuxcnc.org/docs/stable/html/man/man9/hm2_modbus.9.html and https://linuxcnc.org/docs/stable/html/man/man1/mesambccc.1.html for docs on the new driver.

zz912:
This information should be listed here:
https://linuxcnc.org/docs/devel/html/drivers/mesa_modbus.html
I spent quite a bit of time studying and setting up Mesa Modbus.

Is Mesa Modbus unreliable?
Mesa Modbus looks simpler and has support.

Has HM2_MODBUS manual page with example configuration like Mesa Modbus?
Why mbccbs need an absolute path to prevent nasty surprises?

BsAtHome:
> This information should be listed here: https://linuxcnc.org/docs/devel/html/drivers/mesa_modbus.html I spent quite a bit of time studying and setting up Mesa Modbus.

You are right, it should have been made clear in the old docs that the old driver has been superseded (for about 10 months now).

> Is Mesa Modbus unreliable? Mesa Modbus looks simpler and has support.

The old driver is not very capable with respect to how data is represented to/from the modbus device. The hm2_modbus driver uses a mapping from modbus data to hal pins with proper data sizes, data types and translations (as specified in the mbccs). The new driver is also capable of device initialization and gives you full control over timing of the individual commands.

> Has HM2_MODBUS manual page with example configuration like Mesa Modbus?

Manual (man) page hm2_modbus.9 and mesambccc.1as in the links previously provided. There are example in configs in `configs/by_interface/mesa/hm2-modbus/*`.

> Why mbccbs need an absolute path to prevent nasty surprises?

The problem is that the driver has no control over from where it (linuxcnc) was started. So, the "present working directory" (PWD) may change between program starts and results in relative file paths not being found. To prevent that, use absolute paths. If you have files with identical names in multiple places, then you even risk opening the _wrong_ file is you have relative paths.

BsAtHome:
It may be better to take a look at the adoc sources and let github do the adoc formatting. It looks like the homepage's docs are not properly formatted.

https://github.com/LinuxCNC/linuxcnc/blob/master/docs/src/man/man1/mesambccc.1.adoc
https://github.com/LinuxCNC/linuxcnc/blob/master/docs/src/man/man9/hm2_modbus.9.adoc

zz912:
Thank you for response.

> > Why mbccbs need an absolute path to prevent nasty surprises?
> 
> The problem is that the driver has no control over from where it (linuxcnc) was started. So, the "present working directory" (PWD) may change between program starts and results in relative file paths not being found. To prevent that, use absolute paths. If you have files with identical names in multiple places, then you even risk opening the _wrong_ file is you have relative paths.

As a layman, this seems strange to me.
The INI file supports relative paths, for example:
```
PATH_PREPEND = ./python
TOPLEVEL = ./python/toplevel.py
SUBROUTINE_PATH = ./macros
```

I assume that the Modbus configuration file should be in the configuration folder:
Mesa_modbus_folder

BsAtHome:
> As a layman, this seems strange to me.
> The INI file supports relative paths, for example:

You are right, but as a systems engineer, I never trust relative paths unless _I_ control everything.

hansu:
> It may be better to take a look at the adoc sources and let github do the adoc formatting. It looks like the homepage's docs are not properly formatted. 

The homepage docs are properly formatted according to the (old) asciidoc-py. But Github seems to use the (state of the art) asciidoctor syntax.
(Even though there was introduced (new) asciidoctor syntax at some locations which of course doesn't work with asciidoc-py)

andypugh:
Yes, mesa_modbus has been superseded, but it still works (as well as it ever did). The docs should be correct.

Issue #3822, opened 2026-02-22
mesambccc.1.adoc - wrong name of port


zz912:
I spent almost a whole day on this.

In the documentation it says:
`loadrt hm2_modbus ports="7i96s.0.pktuart.0" mbccbs="/path/to/myfile.mbccb"`
and it should be:
`loadrt hm2_modbus ports="hm2_7i96s.0.pktuart.0" mbccbs="/path/to/myfile.mbccb"`

https://github.com/LinuxCNC/linuxcnc/blob/e2f089dbfdc541b87f2930feba767f8588b2f28e/docs/src/man/man9/hm2_modbus.9.adoc?plain=1#L58

The problem is that I couldn't figure out from the documentation where to find the port name.

I tried:
```
zdenek@cnc:~$ sudo mesaflash --device 7i96s --addr 10.10.10.10 --readhmid
Configuration Name: HOSTMOT2

General configuration information:

  BoardName : MESA7I96
  FPGA Size: 20 KGates
  FPGA Pins: 256
  Number of IO Ports: 3
  Width of one I/O port: 17
  Clock Low frequency: 100.0000 MHz
  Clock High frequency: 160.0000 MHz
  IDROM Type: 3
  Instance Stride 0: 4
  Instance Stride 1: 64
  Register Stride 0: 256
  Register Stride 1: 256

Modules in configuration:

  Module: DPLL
  There are 1 of DPLL in configuration
  Version: 0
  Registers: 7
  BaseAddress: 7000
  ClockFrequency: 100.000 MHz
  Register Stride: 256 bytes
  Instance Stride: 4 bytes

  Module: WatchDog
  There are 1 of WatchDog in configuration
  Version: 0
  Registers: 3
  BaseAddress: 0C00
  ClockFrequency: 100.000 MHz
  Register Stride: 256 bytes
  Instance Stride: 4 bytes

  Module: IOPort
  There are 3 of IOPort in configuration
  Version: 0
  Registers: 5
  BaseAddress: 1000
  ClockFrequency: 100.000 MHz
  Register Stride: 256 bytes
  Instance Stride: 4 bytes

  Module: OutM
  There are 1 of OutM in configuration
  Version: 0
  Registers: 1
  BaseAddress: B000
  ClockFrequency: 100.000 MHz
  Register Stride: 256 bytes
  Instance Stride: 4 bytes

  Module: PWM
  There are 1 of PWM in configuration
  Version: 0
  Registers: 5
[...]

PR #3823, opened 2026-02-22
docs: Fix typo in hm2_modbus.9 example 

BsAtHome:
This PR fixes a typo as noted in #3822. It also adds a note about pktuart node creation.


zz912:
I agree with merge.

PR #3824, opened 2026-02-23
Fix printing hm2_modbus messages at startup

BsAtHome:
Fix printing errors/warnings/messages at startup by using rtapi_print() instead of rtapi_print_msg(). The latter does not actually output anything in the rtapi_app_main() function.
This fixes a problem mentioned in #3822 of not finding the PktUART port and hm2_modus not telling the user of the issue.


smoe:
Should this also go into 2.9?

BsAtHome:
> Should this also go into 2.9?

Probably. But I can't test-run 2.9 branch because I haven't backported the tcl/tk9 stuff. That said, the same patch should work for both branches.

andypugh:
I would guess that if it passes CI as a 2.9 fix it's probably good?

PR #3825, opened 2026-02-24
canterp: fix parsing of ORIENT_SPINDLE

smoe:
Did not feel like trusting my eyes, but while the parser is parsing into d1, the value passed is that of d2, with the value of d1 lost. This makes no sense, right?

Can anyone confirm that this patch is correct? I would then prepare the same against 2.9.

### DIFF
--- a/src/emc/canterp/canterp.cc
+++ b/src/emc/canterp/canterp.cc
@@ -592,11 +592,11 @@ int Canterp::execute(const char *line) {
 	    return INTERP_ERROR;
 	}
 	if (!strcmp(s1, "CANON_CLOCKWISE")) {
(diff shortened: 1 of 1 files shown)


BsAtHome:
Correct, the use of `d2` is wrong and it should be `d1`.

smoe:
Thank you, @BsAtHome .  Backported to the 2.9 branch. Feels a bit like this should also go into 2.8 and 2.7.

andypugh:
It could maybe be backported to 2.8 (2.7 never got the multispindle update) but I don't anticipate any further 2.7 or 2.8 releases, and I rather doubt that anyone has ever used canterp in anger, especially not with multiple spindles. It really only exists as a least-effort alternative interpreter to check pluggable interpreters, as far as I can tell.

PR #3826, opened 2026-02-24
canterp presumed missinterpretation of scanf return value

smoe:
A comment should be added if the comparison against 6 is correct - 6 <= scanf() if there are some values not expected to be set, or 9, since all values are used downstream.

### DIFF
--- a/src/emc/canterp/canterp.cc
+++ b/src/emc/canterp/canterp.cc
@@ -299,7 +299,7 @@ int Canterp::execute(const char *line) {
     }
 
     if (!strcmp(the_command_name, "STRAIGHT_PROBE")) {
-	if (6 != sscanf(the_command_args, "%lf %lf %lf %lf %lf %lf %lf %lf %lf",
+	if (9 != sscanf(the_command_args, "%lf %lf %lf %lf %lf %lf %lf %lf %lf",
 			&d1, &d2, &d3, &d4, &d5, &d6, &d7, &d8, &d9)) {
 	    return INTERP_ERROR;
 	}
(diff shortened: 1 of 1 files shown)


BsAtHome:
Also correct. The test on sscanf is supposed to be exact and requires to match the number of conversions requested. Otherwise it makes no sense.

That said, canon's STRAIGHT_PROBE also has a probe-type argument, but that is apparently not supported/implemented?

smoe:
Checked the man page - "These functions return the number of input items assigned.  This can be fewer than provided for, or even zero, in the event of a matching failure.". So, there may be something hidden by that magic omission of 3 arguments that I fail to grasp.

We need some tests for this. And for other bits :-)

smoe:
> That said, canon's STRAIGHT_PROBE also has a probe-type argument, but that is apparently not supported/implemented?

Feels like we have some problem on a meta level that allows for any such inconsistencies.

BsAtHome:
There is a lack of automatic quality control because there is a lack of one or another form of unit tests (primarily internal tests of APIs). Using the static analysis of cppcheck uncovered many bugs and inconsistencies, but fails to detect logic errors (like the sscanf issue or using the wrong variable). There are many more lurking in the code. Ideas are welcome to extend (static) testing on that.

Internal tests need to be implemented to check both consistency and function of those parts that the user normally never sees. And that is not always a simple task<sup>(*)</sup>. There is afaik no standard framework to cope with the complexities and intricacies of LCNC, so it would need to be written. A heavy task that we can put on the TODO pile...

<sub><sup>(*)</sup> and they need to be parallelizable to speed things up</sub>

andypugh:
It feels like it would make sense to be able to omit ABCUVW if Canterp was a real interpreter. 
But that would be 3, not 6. 
All the example commands in canterp_example.can use all 9 axis values.

smoe:
@andypugh , is it worthwhile to redirect that patch to 2.9? Or would you just manually fix this and close this PR?

PR #3829, opened 2026-02-24
motion-logger: Do not use value passed unchecked to access array

smoe:
Please kindly have an extra eye on me setting <,<=,> correctly. Maybe these checks are not required because those checks are already performed elsewhere. Still, it feels weird to not perform such checks. The major question though is if enough is done if any such invalid number is observed.

### DIFF
--- a/src/emc/motion-logger/motion-logger.c
+++ b/src/emc/motion-logger/motion-logger.c
@@ -488,8 +488,16 @@ int main(int argc, char* argv[]) {
                     "SET_JOINT_POSITION_LIMITS joint=%d, min=%.6g, max=%.6g\n",
                     c->joint, c->minLimit, c->maxLimit
                 );
-                joints[c->joint].max_pos_limit = c->maxLimit;
-                joints[c->joint].min_pos_limit = c->minLimit;
+                if (c->joint < 0) {
+                    log_print("ERROR: SET_JOINT_POSITION_LIMITS ignored since joint number %d is < 0.\n",c->joint);
+                }
+                else if (c->joint < EMCMOT_MAX_JOINTS) {
+                    joints[c->joint].max_pos_limit = c->maxLimit;
+                    joints[c->joint].min_pos_limit = c->minLimit;
+                }
+                else {
+                    log_print("ERROR: SET_JOINT_POSITION_LIMITS ignored since joint number %d beyond technical limit %d.\n",c->joint,EMCMOT_MAX_JOINTS-1);
+                }
                 break;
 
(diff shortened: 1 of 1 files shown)


BsAtHome:
IIRC, axis checks should also check the mask because you have, for example,  XZA configs where Y-axis is effectively invalid. But you need to have access to the axis_mask in trajectory stat, `emcStatus->motion.traj.axis_mask`, which holds the valid axis as a bit mask. Don't know if that is exposed somehow, somewhere. If not, then the index check should suffice. Writing to unused axes should have no effect if the kinematics play nice.

The checks could also be slightly simplified as:
```C
// valid: zero inclusive, max exclusive
if (joint >= 0 && joint < EMCMOT_MAX_JOINTS) {
  // valid, use joint as index
} else {
  log_print("...invalid joint index %d. Valid range [0,%d]\n", joint, EMCMOT_MAX_JOINTS-1);
}
```

And, why do you split lines on closing curly brace and an else clause? You'd normally write that as:
```C
if (bla) {
  something();
} else {  // <-- this line
  something_else();
}
```

smoe:
Wrt formatting I still hope this will soon get resolved in an automated manner.  In R, I do as you suggested, in C/C++ not :-)

I do not know how that logger is ultimately used. But from looking over that file I tend to get the impression that it should be autogenerated from some abstract description of all these commands.

BsAtHome:
> I do not know how that logger is ultimately used.

There are at least a couple of tests that check using motion-logger (see `tests/motion-logger/*`).

> But from looking over that file I tend to get the impression that it should be autogenerated from some abstract description of all these commands.

While I agree in principle, the practicalities are not necessarily on our side. The whole emcmot* stuff  is used in RT/non-RT communication. You can't just change the interface without changing the code on both sides of the interface. So, any generated stuff needs careful manual work on both sides to fit again.

smoe:
The translations apparently do not build on bullseye because of an outdated po4a. 
"configure: WARNING: po4a version too old, need version 0.67 or newer, not building translated docs"

If it is not ultimately urgent then I suggest to backport the latest po4a to bullseye-backports. And then we see - should be doable.

rmu75:
> And, why do you split lines on closing curly brace and an else clause? You'd normally write that as:
> 
> ```c
> if (bla) {
>   something();
> } else {  // <-- this line
>   something_else();
> }
> ```

"normally" is debatable ;-) I prefer the else on a separate line, same with try/catch statements.

in linuxcnc you can find both styles, even in the same file, but `} else {` seems to be predominant.

andypugh:
Limits already checked in task (calling function). Further tests not needed.

PR #3831, opened 2026-02-25
fix: Error detection after open(2)

smoe:
The patch speaks for itself. I guess there was some confusion between open and fopen.

### DIFF
--- a/src/emc/tooldata/tooldata_common.cc
+++ b/src/emc/tooldata/tooldata_common.cc
@@ -83,7 +83,8 @@ int tooldata_read_entry(const char *input_line,
         return -1;
     }
     if (input_line[0] == ';') {return 0;} //ignore leading ';'
-    strcpy(work_line, input_line);
+    strncpy(work_line, input_line, sizeof(work_line)-1);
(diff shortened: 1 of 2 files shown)


BsAtHome:
Why do you need to comment on why the includes are included? Isn't that obvious?

And, nice catch, that fd value bug.

smoe:
I had "man 3 open" but that failed, so I was looking for a place to direct to the right one. That is typically after the #include line. For fopen it would have been correct, so I saw some value in the ref to the manual.

BsAtHome:
That is "man 2 open" :-) It is a syscall/kernel function and many make the same mistake.

However, it does not, IMNSHO, mean you should add the comment. The comment is a distraction. Or do you also want to add the comment in the fcntl.h include that is pulls in pid_t, mode_t, off_t and off64_t? Or what about the S_I* macros it defines? How about the locking functions? We can add the entire header as comment? Where do you draw the line? That is the problem.

Another question is that there are probably <i>redundant</i> includes in the source tree. That needs to be checked too and that is a hard job to do right.

PR #3832, opened 2026-02-25
fix: tooldata mutex init - give before try

smoe:
explicitly initialize the mutex, instead of calling tool_mmap_mutex_give() without a prior lock.

Mutex is “given” without an initial successful “try/get” In src/emc/tooldata/tooldata_mmap.cc:163, tool_mmap_creator() calls tool_mmap_mutex_give() right after setup without any prior lock acquisition. According to the man page this is unusual and potentially undefined depending on rtapi_mutex semantics.

I am not sure about how much of a bug this is, but there should be an initial assignment of mutex, right?

### DIFF
--- a/docs/man/man3/rtapi_mutex.3rtapi
+++ b/docs/man/man3/rtapi_mutex.3rtapi
@@ -36,4 +36,4 @@ init/cleanup, and realtime code.
 \fBrtapi_mutex_try\fR returns 0 for if the mutex was claimed, and nonzero
 otherwise.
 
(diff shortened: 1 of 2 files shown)


BsAtHome:
This is a mutex that can be "given" when not held.

That said, the solution is not to remove the mutex give, but to _add_ the mutex get in line 157 to protect the actual block that needs protecting.

BsAtHome:
Strike my last... You should just leave it as is.

The 'give' actually initializes the mutex. Is simple clears the mutex unconditionally because you do not need to hold it to run 'give' on it.

andypugh:
Closed after developer chat 1/3/2026

PR #3838, opened 2026-03-02
cppcheck: fix casts and copy constructors

BsAtHome:
This PR addresses most remaining warnings.
Missing class member initialization is fixed. All the "dangerous old" casts are now changed into `reinterpret_cast<>()`. Classes that use dynamic memory, as detected by cppcheck, have their copy constructors and copy assignment removed. These classes are not meant to be duplicated anyway.

The remaining cppcheck warnings deal with `emc/usr_intf/schedrmt.cc` stuff. However, the scheduler program needs a complete overhaul like linuxcncrsh recently had. That is on the todo list.

### DIFF
--- a/src/emc/ini/emcIniFile.cc
+++ b/src/emc/ini/emcIniFile.cc
@@ -43,7 +43,7 @@ EmcIniFile::ErrorCode
 EmcIniFile::Find(EmcJointType *result,
                  const char *tag, const char *section, int num)
 {
-    return(IniFile::Find((int *)result, jointTypeMap, tag, section, num));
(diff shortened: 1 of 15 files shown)


rmu75:
Maybe the coding style document should be updated with a hint to not use C-style casts in C++? https://github.com/LinuxCNC/linuxcnc/blob/master/docs/src/code/style-guide.adoc#c-conventions

BsAtHome:
> Maybe the coding style document should be updated with a hint to not use C-style casts in C++?

Maybe... But then again, the coding style of much of the C++ code is more like "C with classes".

PR #3842, opened 2026-03-03
Interpreter: Fix Allow G43 for tools with zero offsets

Sigma1912:
Fixes errors that I seem to have inadvertently introduced while preparing the final PR for 0673e3f :

- Fix wrong logic (current version sets the flag when an offset value is unequal zero)
- Check tool table offsets not the currently active tool offsets

Also includes 
- Initialization for the flag in interp_setup as suggested by @BsAtHome


zz912:
What this fix do? I tested Allow G43 ... and I didnt find problem.

Sigma1912:
> What this fix do? I tested Allow G43 ... and I didnt find problem.

Thanks for the prompt, I forgot to add the relevant information in the commit message. Updated now, so please see above.

hansu:
@zz912 I guess this was the error:

tooltable-g43-flag

zz912:
When the video loop begin?

hansu:
> When the video loop begin?


https://github.com/user-attachments/assets/31bf7f7c-a337-4542-a498-1d95168abe59

PR #3851, opened 2026-03-08
Fix python crash due to missing Py_None reference counting increase

### LINKED ISSUE #3850: LCNC 2.10-pre1: Crash (none_dealloc) in gladevcp/hal_widgets.py with HAL_RadioButton
**Summary:**
A Python error occurs in Gmoccapy 3.51 (LinuxCNC 2.10~pre1) when a custom Glade panel containing HAL_RadioButton groups is embedded. The UI crashes consistently after approximately 2 minutes and 29 seconds of idle time, even in a simulation environment.

**Environment:**
OS: LMDE 6 (Faye)
LinuxCNC Version: 2.10~pre1 (Run-In-Place / linuxcnc-dev)
Python Version: 3.11.x
GUI: Gmoccapy 3.51 (Simulated)

**Traceback:**

```
Python

Fatal Python error: none_dealloc: deallocating None: bug likely caused by a refcount error in a C extension
Python runtime state: initialized

Current thread 0x00007fb11f2f6040 (most recent call first):
  File "/home/cnc/linuxcnc-dev/lib/python/gladevcp/hal_widgets.py", line 50 in hal_update
  File "/home/cnc/linuxcnc-dev/lib/python/gladevcp/makepins.py", line 68 in update
  File "/usr/lib/python3/dist-packages/gi/overrides/Gtk.py", line 1689 in main
  File "/home/cnc/linuxcnc-dev/bin/gladevcp", line 396 in main
  File "/home/cnc/linuxcnc-dev/bin/gladevcp", line 410 in <module>
```
**How to reproduce:**

1.    Save the attached left-vel-frame.txt and rename it to left-vel-frame.glade.
2.    Create or navigate to a standard Gmoccapy Sim Machine configuration folder.
3.    Place the .glade file inside that folder.
4.    Edit the .ini file. Under the [DISPLAY] section, add the following lines:

**Glade File as .TXT:**
[left-vel-frame.txt](https://github.com/user-attachments/files/25818626/left-vel-frame.txt)

Edit the .ini file. Under the [DISPLAY] section, add the following lines:

   
```
    EMBED_TAB_NAME = left_vel_frame
    EMBED_TAB_LOCATION = box_custom_1
    EMBED_TAB_COMMAND = gladevcp -x {XID} left-vel-frame.glade
```

5.    Start LinuxCNC from the terminal with debug and verbose flags:
    linuxcnc -d -v gmoccapy.ini
6.    Wait for exactly **_2 minutes and 30 seconds_** without any user interaction.
7.   The application will crash with the none_dealloc error.

BsAtHome:
This fixes #3850 by adding reference count increase before returning Py_None.

### DIFF
--- a/src/hal/halmodule.cc
+++ b/src/hal/halmodule.cc
@@ -544,6 +544,7 @@ static PyObject *pyhal_ready(PyObject *_self, PyObject * /*o*/) {
     EXCEPTION_IF_NOT_LIVE(NULL);
     int res = hal_ready(self->hal_id);
     if(res) return pyhal_error(res);
+    Py_INCREF(Py_None);
     return Py_None;
 }
 
@@ -553,12 +554,14 @@ static PyObject *pyhal_unready(PyObject *_self, PyObject * /*o*/) {
     EXCEPTION_IF_NOT_LIVE(NULL);
     int res = hal_unready(self->hal_id);
     if(res) return pyhal_error(res);
+    Py_INCREF(Py_None);
     return Py_None;
 }
 
 static PyObject *pyhal_exit(PyObject *_self, PyObject * /*o*/) {
     halobject *self = reinterpret_cast<halobject *>(_self);
     pyhal_exit_impl(self);
+    Py_INCREF(Py_None);
     return Py_None;
 }
 
@@ -597,8 +600,10 @@ static PyObject *pyhal_get_prefix(PyObject *_self, PyObject *args) {
     if(!PyArg_ParseTuple(args, "")) return NULL;
     EXCEPTION_IF_NOT_LIVE(NULL);
 
-    if(!self->prefix)
-	return Py_None;
+    if(!self->prefix) {
+        Py_INCREF(Py_None);
+        return Py_None;
+    }
 
     return PyUnicode_FromString(self->prefix);
 }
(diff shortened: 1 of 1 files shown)


andypugh:
I can't even pretend to understand this bug, or the fix. But thanks.

BsAtHome:
> I can't even pretend to understand this bug, or the fix. But thanks.

Quite simple... I introduced the bug when fixing cppcheck errors a long while ago.

Background, every object in python has a reference count. When you get an object, you increase its reference count (informing that the object is in use). When you no longer need it, you decrease the count. Once the count reaches zero, python will automatically discard the object and free the memory. This is true for all objects, including the `None` object (in C it is called `Py_None`).

Halmodule had originally return statements using `Py_RETURN_NONE`, which is a macro that cppcheck is very disagreeable with. These got changed into `return Py_None`, but that forgot that the object is referenced (in use) one more time at return time and should have had its reference count increased. When the returned `None` object becomes unused, somewhere else in the code, then the reference count gets decreased again. At some stage, the `None` object gets discarded because it has experienced more reference decreases than increases, which is, to say it mildly, a bad thing to happen. When it happens, python correctly asserts a <i>programmer too stupid</i> error and crashes the program.

Issue #3849, opened 2026-03-06
hm2_modbus stops sending commands, increments


JTrantow:
I'm using latest master rip and seeing hm2_modbus lose communications with a CommanderSK VFD via a 7i96 pkuart.

The communications will be fine for several vfd on/off cycles but eventually communications won't turn the VFD on/off.

When this occurs, I see a constantly increasing count (approx 1000 counts per second) on one of the modbus.0.commands.XX.errors counts. It is not stopping at 5. The XX has always been a read command but which one seems random. I started with 11 commands (9 reads, 2 writes). 

modbus.0.commands.XX.error_code = 42 (ENOMSG)
modbus.0.commands.XX.disabled = true
modbus.0.commands.XX.reset doesn't seem to have any effect. 
modbus.0.commands.XX.disabled doesn't seem to have any effect.  

All the other commands have error count =0, but I can't disable or reset the command with a problem.

When this occurs, 
modbus.0.last-error-code = 0
modbus.0.fault_command = 0
modbus.0.suspend has no effect.
modbus.0.reset has no effect.

By "has no effect", I see the particular command.XX.error code continue to increment, approx every 1msec, and it doesn't clear or stop incrementing.
Communications are 38.4K baud and there are currently 9 read commands (bunch of register reads to debug what's going on) 

Docs indicate suspend is probed at end of <commands> list, maybe it is stuck in a 1msec loop incrementing the command error and never makes it to the end of the list???

Docs indicate the .errors count should max out at 5. This is not happening. 

Using the 2.9.8, I remember seeing .errors count up to 5 and I could reset them, may have been a different error code. Master version does NOT stop counting at 5 with error_code=42.

JTrantow:
I examined mesa_modbus.c and changed the MSG_WARN to 

`			MSG_ERR("%s: warning: reply to command %u had too long inter-character delay (%u > %u), dropping\n",
					inst->name, inst->cmdidx,
					HM2_PKTUART_RCR_ICHARBITS_VAL(frsize), inst->maxicharbits);

When I ran this change, the message reported inter character delay of 36 when it was automatically set to 29. I increased icdelay="40" and I have not locked up in the last few hours working on the lathe.

The CommanderSK is a fairly old VFD, maybe it just needs a longer icdelay? Even increasing icdelay fixes this case,  there is still a bug handling the ENOMSG error. This condition might be repeatable by setting small icdelay values.`

BsAtHome:
[...]

PR #3852, opened 2026-03-09
Issue #3849 Fix

JTrantow:
Fix for calc_ifdelay() and ENOMSG message handling.
Without the ENOMSG fix, the modbus would lock up upon a single ENOMSG error.
The calc_ifdelay() fix corrects the delay calculation for baud rates > 19.2K. Without this fix, you are more likely to get into the ENOMSG condition.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/hm2_modbus.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_modbus.c
@@ -387,7 +387,7 @@ static unsigned calc_ifdelay(hm2_modbus_inst_t *inst, unsigned baudrate, unsigne
 	}
 
 	// calculation works for baudrates less than ~24 Mbit/s
-	if(19200 < baudrate)
+	if(baudrate > 19200)
 		return (175u * baudrate + 99999u) / 100000u;
 	unsigned bits = 1 + 8 + (parity ? 1 : 0) + (stopbits > 1 ? 2 : 1);
 	return (bits * 35 + 9) / 10;	// Ceil of bits in 3.5 characters.


BsAtHome:
This statement:
```C
#define HARDWARE_MAX_DELAY (1020)	//!< Maximum ifdelay in number of bits limited by hardware.
```
Is only partly correct. The max delay (in hardware) is 255. However, PktUART v3 has a x4 clock divider on it that v2 does not have (older than v2 is not supported). So, stating that the max delay is 1020 is only valid for one version of the hardware and not the other. The difference is accounted for elsewhere (and is clamped in the hm2_pktuart_config() implementation).

JTrantow:
What do you think about removing the 1020 check from the calc_if() completely and have the hardware limitation get enforced when it gets applied at the PktUART level? That would simplify calc_if() to modbus bit timing without the hardware implementation constraint and allow the PktUART version to clamp/max the value appropriately.

BsAtHome:
Well, you made a four-line diff I suggested into a larger issue (see https://github.com/LinuxCNC/linuxcnc/issues/3849#issuecomment-4019870655). I'm not sure that is wise.

If you want to change how the component is written, commented and documented, then I'd be fine with that and you are very welcome to do so and create a PR for that. But the extra changes have no value in addressing the bug identified because it conflates two different issues. That is my point.

JTrantow:
Agreed, I had strayed off topic. 

I submitted PRs for master and 2.9. 
I left my comments as is. I see your points, feel free to change. When the PRs are complete this issue can be closed.

BsAtHome:
Thank you. However, now there is another issue with you patch:  The comparison is wrong because 19200 baud must use the 3.5 char mark. The original change, replacing `<=` with `>`, was the correct one.

PR #3853, opened 2026-03-09
tp.c: remove unused tpGetScaledAccel (fixes #3720)

### LINKED ISSUE #3720: Unused inline function

grandixximo:
Function was never called; dead code

### DIFF
--- a/src/emc/tp/tp.c
+++ b/src/emc/tp/tp.c
@@ -359,25 +359,6 @@ STATIC inline double tpGetRealFinalVel(TP_STRUCT const * const tp,
 /**
  * Get acceleration for a tc based on the trajectory planner state.
  */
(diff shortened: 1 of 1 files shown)


hdiethelm:
Just pulled: You forgot to delete the function's comment.
https://github.com/grandixximo/linuxcnc/blob/fc7c5c9303bd7875ec7a67eb3e28e48627cebe45/src/emc/tp/tp.c#L360

PR #3857, opened 2026-03-10
Issue#3849 fix backported to 2.9

JTrantow:
Fixes calc_ifdelay() for baud > 19200 that was setting bit timing delays incorrectly leading to a variety of communication problems. Fixes ENOMSG error handling that was locking up modbus communication.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/hm2_modbus.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_modbus.c
@@ -387,7 +387,7 @@ static unsigned calc_ifdelay(hm2_modbus_inst_t *inst, unsigned baudrate, unsigne
 	}
 
 	// calculation works for baudrates less than ~24 Mbit/s
-	if(19200 < baudrate)
+	if(baudrate > 19200)
 		return (175u * baudrate + 99999u) / 100000u;
 	unsigned bits = 1 + 8 + (parity ? 1 : 0) + (stopbits > 1 ? 2 : 1);
 	return (bits * 35 + 9) / 10;	// Ceil of bits in 3.5 characters.
@@ -1049,7 +1049,7 @@ fetch_more_data:
 			break;
 		}
 		if(inst->maxicharbits && HM2_PKTUART_RCR_ICHARBITS_VAL(frsize) > inst->maxicharbits) {
-			MSG_WARN("%s: warning: reply to command %u had too long inter-character delay (%u > %u), dropping\n",
+			MSG_WARN("%s: warning: reply to command %u had too long inter-character delay (%u > %u), trying to interpret\n",
 					inst->name, inst->cmdidx,
 					HM2_PKTUART_RCR_ICHARBITS_VAL(frsize), inst->maxicharbits);
 			set_error(inst, ENOMSG);


BsAtHome:
The same objection as in #3852. You are changing form and do not constrain yourself to substance (aka just fixing the bug).

If you object to the form, then please submit that as a separate PR where we can discuss that with a broader audience and decide whether the style of coding needs to be adjusted. But, if so, then you should not change one very tiny corner, but adjust the _entire_ source.

BsAtHome:
Thank you. However, now there are two issues with this backport:

1. [edit] [edit], no,... I was right. the original `baudrate <= 19200` was changed into `19200 < baudrate`, which translates into `baudrate >= 19200` and that is wrong. It should read `baudrate > 19200` and that would have translated into `19200 <= baudrate`. Reversing order, while regularly used in "defensive programming", is utterly poisonous for the mind because you make exactly this mistake when you are not dealing with `==` or `!=`.
2. you are missing the change in the message (change from "dropping" to "trying to interpret")

JTrantow:
(19200 < baudrate) and (baudrate > 19200) are both false when baudrate equals 19200.

Correct, missed the "trying to interpret" message in one PR. I will fix that tomorrow.

BsAtHome:
You are right about the comparison; when something like that my mind goes into condition reversal mode.

However, my statement stands, it is pure poison and contrary convention. It doesn't help understanding, doesn't improve readability and doesn't qualify in defensive programming (preventing the single '=' (assignment) problem).

JTrantow:
"pure poison and contrary convention", I'm guessing you never had an HP
postfix calculator?

Placing the constant first in relational ops is completely valid and once
you use that order for avoiding assignment vs relational mistakes the mind
easily flips for all relational operators.

I have reverted both PRs to conform to your preference and corrected the
MSG_WARN.

Let's get this issue closed and make the improved code available!
Beyond the ENOMSG lockup, I suspect the baud rate problem created more
widespread problems for anyone using hm2_modbu with baud > 19200. I was
seeing occasion messages about incorrect length (that recovered). All
communications problems went away once I applied the calc_ifdelay() fix.
The ENOMSG bug could have remained in the code for a long time if
calc_ifdelay() was correct.

A lot of these discussions would go better over a coffee or beer. Where are
you located? I'm near Madison, WI, USA.

BsAtHome:
> "pure poison and contrary convention", I'm guessing you never had an HP postfix calculator? Placing the constant first in relational ops is completely valid and once you use that order for avoiding assignment vs relational mistakes the mind easily flips for all relational operators.

Actually, I had an HP RPN calculator and loved it (beautiful red small 7-segment display). However, this work is not postfix (nor prefix as in Lisp). C is infix.

> I have reverted both PRs to conform to your preference and corrected the MSG_WARN. Let's get this issue closed and make the improved code available!

Agreed!

> Beyond the ENOMSG lockup, I suspect the baud rate problem created more widespread problems for anyone using hm2_modbu with baud > 19200. I was seeing occasion messages about incorrect length (that recovered). All communications problems went away once I applied the calc_ifdelay() fix. The ENOMSG bug could have remained in the code for a long time if calc_ifdelay() was correct.

Right. The problem with - especially older - equipment is that it may not be able to saturate the serial data line and actually leave inter-character delays in the stream. And those will be picked up as inter-packet delay if that is set too low. Good this particular bug was caught.

> A lot of these discussions would go better over a coffee or beer. Where are you located? I'm near Madison, WI, USA.

Love to chat over coffee, but I'm not even close. About 6700 km east (and a bit north) on the other side of the pond in that small country that has about one third of its land mass under sea level and has lots of windmills and levees. And, I wont be traveling to the USA again, ever (unfortunate but necessary).

JTrantow:
HP postfix comment was in reference to mindset and not C language in particular. Different way of doing the same thing. People love it or hate it. Change opinion after they see benefits for particular cases.

No travel plans (in that direction) in my immediate future. Spending the summer in Alaska.

Who signs off on the PRs?

Later today, I intend to initiate a related issue/PR on testing hm_modbus. My policy is to add a test case for every bug that would have captured the problem earlier and to ensure the same bug never made it's way into a release again. I already have a working tests/mod-bus that allows me to load various .mccb's with different com settingsand verify the delays. I would appreciate your comments on that. It's easy to test initial settings, I need to think about what would be necessary to emulate actual communications to catch ENOMSG type problems.

BsAtHome:
> Who signs off on the PRs?

Usually Andy merges. Merges in the 2.9 tree are done more carefully.

But, @andypugh, this PR should be merged as it fixes a real bug (you may want to squash the commits into one).

PR #3861, opened 2026-03-14
SEDM : a hal comp that emulates Sink EDM orbiting

tjtr33:
SEDM:
SEDM emulates Dink EDM orbiting.
That is mainly used to make injection molds. 
Orbiting is a way to increase precison 
and simply making the tools.

	PATH PLANNING
The entire forward path is pre-calculated 
and stored as a list of (X,Y,Z) tuples.
(often thousaands of tuples)
I store tuples on a 1 micron grid. 
These tuples are relative to the tool's starting position (SPO). 

	PARAMETERS
SEDM parameters are set by Mcodes, 
and by files called Etables..

	ROCESS CONTROL
Real EDM is driven by lthe voltage difference 
between tool and work (GapV orGV), 
to determine tool direction. 

There is no feed or speed input.
Those are results of thhe process, 
not inputs to the ontrol strategy. 

SEDM simply moves towrds th correct position
as soon as possible.

Good Sink EDM machines are stiff 
yet able to quickly move small distances. 

	REAL EDM vs EMULATED EDM
SEDM creates a fake GV using 
python's random:uniform().
The result is evaluated by 
a software ''window comparator', 
resulting FWD HOLD BWD. 

These evaluations determin thge next tool position.  

If a real EDM gap sensor is used, this code
bevomes a real Sink EDM Orbiting conrol
( needs a smapler, and a programmable generator0

	How FWD HOLD BWD control Tool Position
FWD simply incremennts the tool position, 
an the stored offsets are given to the 
external offsets of the axis involved. 
HOLD samples do not change position,. 
The comp continues and the GV wil eventuially 
move outside of the window. 
BWD samples will create a new DETOUR path 
on the fly. 

	DEOURS changes to the precalculated path
Detour paths  begin and end on the main path. 
BWD samples move further away from stock, 
and FWD samples return towards the main path.
The edgees of the window comparator are live 
time adjustable (GVHI GVLO).  

	Cuts can FAIL or SSUCCEED
If too many contiguous BWDs occur, 
then MAXXBWED parm is exceede 
and the program aborts. 
SSEDM always returns the tool 
to SPO, and turns off power,
and SEDM sleps til caalled again.

	OPERATOR INTERACTION
The operator can change Off Time, the winow comparator edges,\
ans the peck edm-in ( enable, cut time, jump distance)

	PEEK RESUME QUIT
Durinmg use, the operator can 
PEEK ( turn off powe andd withdraw the tool to SPO_
thwn RESUME (if inspection of tool and woek are ok)
or WUIT ( NGV is aborted, machine is ready for use, SEDM instanding by)

Please see README for more.


tjtr33:
there's a lot to EDM.
this a lot of code
I've dome similar for
other control systems
so there's a lot of old ideas here.
I hop[e someone cna use it.
It should do hole andsink and wire edm,

andypugh:
It's fine, the PR has been submitted. 
To get out again you just click on something at the top of the page.

Issue #3839, opened 2026-03-02
G2 doesn't do proper limit checks


hdiethelm:
At least I think that's not supposed to happen:

LinuxCNC 2.9.8 or master branch / axis_mm:

- Start and home
- Go to MDI and enter "G2 I1000 F2000"
- The machine just runs into the soft stops with full speed and overshoot the limits
  - In my real setup, it would probably crash due to the distance between soft limits and hard crash not enough to stop with full speed
- "G0 X1000" for example is checked and not executed if it will go off limits

If I find time, I might look into the code.

andypugh:
Well the good(?) news is that it's always been like this.  I have tested back to LinuxCNC v2.5 from 2012 and it's just the same. 

The same code inside a G-code program does bring up a warning, but if you override then it runs out of bounds too.

hdiethelm:
I did some future testing by writing a tiny ngc file and and pressing continue anyway when I'm asked.

- G0 out of bounds
  - GUI doesn't complain 
  - Interpreter doesn't execute it
- G1 out of bounds
  - GUI does complain
  - Interpreter doesn't execute
- G2/G3  out of bounds
  - GUI does complain 
  - Interpreter executes it and then it goes out of bounds
- G5 out of bounds
  - GUI does complain 
  - Interpreter doesn't execute it

It's not that big of an issue, because in most cases, you probably use G2/G3 only in .ngc files but you can crash you machine if you use it in MDI mode. Luckily I tried that first in simulation because I'm still learning and I don't want to crash my machine... ;-)

Also if you write a macro containing G2/G3 and execute it in MDI, you have the same issue. Probably also when you execute it using HALUI->MDI_COMMAND (I did not test that)

hdiethelm:
Some debugging later I found that inRange does only check the position of the circle, not the diameter:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/motion/command.c#L1076

Copying over some code from motion-logger.c results in:
```
SET_CIRCLE:
    pos: x=45.04, y=8.7672, z=-68.49, a=0, b=0, c=0, u=0, v=0, w=0
    center: x=1045.04, y=8.7672, z=-68.49
    normal: x=0, y=0, z=1
    id=0, motion_type=3, vel=33.3333, ini_maxvel=80, acc=200, turn=-1
```
a few seconds later:
`Exceeded POSITIVE soft limit (405.00000) on joint 1`

[...]

PR #3866, opened 2026-03-18
Fix command.c debug messages

hdiethelm:
While working on https://github.com/LinuxCNC/linuxcnc/issues/3839, I activated debug and found that the output had random newlines.

The reason is that at the end of emcmotCommandHandler_locked(), a newline is printed and not every command had a rtapi_print_msg().

ABORT doesn't use joint_num, so I see no reason to print it.

To test, I added rtapi_set_msg_level(RTAPI_MSG_ALL); in control.c emcmotController(). This is temporary, not in this PR. Is there any other way to activate debug per module?

Alternative: Remove the rtapi_print_msg(RTAPI_MSG_DBG, "\n"); at the end and add a newline to all other rtapi_print_msg(); Advantage: Less issues in the future. Disadvantage: You don't see if someone forgot to add prints and there will be more changed lines.

### DIFF
--- a/src/emc/motion/command.c
+++ b/src/emc/motion/command.c
@@ -503,7 +503,6 @@ void emcmotCommandHandler_locked(void *arg, long servo_period)
 	       does yet), and if in free mode, it disables the free mode traj
 	       planners which stops joint motion */
 	    rtapi_print_msg(RTAPI_MSG_DBG, "ABORT");
-	    rtapi_print_msg(RTAPI_MSG_DBG, " %d", joint_num);
 	    /* check for coord or free space motion active */
 	    if (GET_MOTION_TELEOP_FLAG()) {
                 axis_jog_abort_all(0);
@@ -536,6 +535,7 @@ void emcmotCommandHandler_locked(void *arg, long servo_period)
 	case EMCMOT_JOG_ABORT:
 	    /* abort one joint number or axis number */
 	    /* can happen at any time */
+	    rtapi_print_msg(RTAPI_MSG_DBG, "JOG_ABORT: %i", joint_num);
 	    if (GET_MOTION_TELEOP_FLAG()) {
 	        /* tell teleop planner to stop */
 	        if ((emcmotCommand->axis >= 0) && (emcmotCommand->axis < EMCMOT_MAX_AXIS)) {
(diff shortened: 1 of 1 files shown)


BsAtHome:
> Is there any other way to activate debug per module?

Unfortunately, no. The print message/debug interface is an all-or-nothing proposition. For real module/component/level differentiating debug messages we need to build a different interface. Maybe using a "channels" idea to select what kind of debug is enabled. And, IMO, we should be able to switch debug off and not being build/included at all to see what that does in performance (instead of executing the many if(level >...) statements).

andypugh:
You can change the debug level in the HAL file. This is something that I added, and I am not sure how well it works. 
It doesn't really address your question, except possibly allowing you to selectively debug the loading of a particular module at a different level. 

Prior to adding the HAL command the way to change debug level on the fly was "echo 7 > /proc/rtapi/debug".  But that had stopped working by the time I added the HAL command. Possibly it only ever worked in RTAI?

hdiethelm:
> You can change the debug level in the HAL file. This is something that I added, and I am not sure how well it works. It doesn't really address your question, except possibly allowing you to selectively debug the loading of a particular module at a different level.

Like `halcmd debug 5`? For me it doesn't work but I am also not using rtlinux on my dev machine due to ZFS is incompatible it seams. Might be a simulated RT issue.

hdiethelm:
I just did a bit digging: At least without rtlinux, do_set_debug_cmd() can not work because it is called in the halcmd process and sets the msg_level of this process. So not inside the rtapi_app process where it would have an effect. 
The command should be sent to rtapi_app and executed in there to have an effect.

I hacked something together which seams to work. I will make it nicer and create yet an other PR in a few days.

BsAtHome:
In uspace, there is a UNIX socket linking the master and slave `rtapi_app` program. The first to start becomes the master and all other instantiations of `rtapi_app` will become a slave.

The slave will send a command down the socket for the RT part to execute (like load an RT component). This command simply seems to be the `argv[1...argc-1]` list of the program. Changing the message level needs to be done in the master that should receive it automatically from the slave. But, there does not seem to be a "debug" command recognized in the `handle_command()` function. That is why the level is never changed in uspace.

hdiethelm:
Yes, i figured that already out and it is already working well. I just have to remove my debug code and then I will create a PR.

hdiethelm:
[BsAtHome](https://github.com/BsAtHome): Thanks, anyway. I think while you where writing, I already figured that out by adding some debug prints and reading the code.

PR is ready: https://github.com/LinuxCNC/linuxcnc/pull/3870

hdiethelm:
https://github.com/LinuxCNC/linuxcnc/pull/3870 was merged, so I think this can be merged to.

BsAtHome:
> #3870 was merged, so I think this can be merged to.

Agreed.

Issue #3779, opened 2026-02-02
Re-homing fails in axis sim


BsAtHome:
Running 2.10-pre directly from git master on Trixie.

A repeatable way to fail homing was discovered resulting in the message: `Cannot home while shared home switch is closed j=2`.

Scenario to repeat:
1. Start axis sim and set Estop=off, Machine=on
2. Home all axes (press "Home All" button) and wait for all axes to be homed
3. Unhome all axis (menu "Machine/Unhoming/Unhome All Axes")
4. Again try to home all axes (press "Home All" button)
5. --> error pops up

Other scenario to repeat:
1. Start axis sim and set Estop=off, Machine=on
2. Manually home X (menu "Machine/Homing/Home Axis X")
3. Manually home Z (menu "Machine/Homing/Home Axis Z")
4. --> error pops up

Manual order homing Y, Z, X in second scenario does work, but you still cannot unhome and repeat the homing as in the first scenario.

Sigma1912:
That is because we have 'HOME_IS_SHARED = 1' for joints 0 and 2 in the ini file.
To rehome you need to move the X axis in the negative direction so the simulated shared home switch signal goes False . Not sure if this ever worked otherwise but this could probably be fixed by setting 'HOME_OFFSET = 1'
https://github.com/LinuxCNC/linuxcnc/blob/master/configs/sim/axis/axis.ini#L172

andypugh:
Yes, this is a long-standing irritation with the Axis sim. I have often found it when trying to debug other things, but have always been more interested in the other thing....

hansu:
The same for the gmoccapy sim configs. At least when you abort homing and try to re-home.

BsAtHome:
Setting HOME_OFFSET of JOINT_0 to 1.0 does indeed solve the immediate problem. But you can still trigger the problem.

So, why is there a shared homing signal? Is this necessary for the sim? I can see an application in real machines, but in a sim?

Sigma1912:
I don't think it is necessary.  It may just be like that because somebody found a way to simulate it many years ago.
https://github.com/LinuxCNC/linuxcnc/blob/master/lib/hallib/simulated_home.hal

andypugh:
I imagine it is to demonstrate the concept. 
However, with a real machine you would be able to see which axis was on the home switch / end stop. With the sim it's a bit mysterious.

hdiethelm:
[...]

PR #3869, opened 2026-03-19
Remove use of HOME_IS_SHARED in sim configs

Sigma1912:
Using HOME_IS_SHARED causes unnecessary error messages when rehoming in sim configs.

For discussion:
https://github.com/LinuxCNC/linuxcnc/issues/3779


BsAtHome:
Shouldn't there be at least one example config (with an appropriate name) that demonstrates shared home?

Maybe one of the simple ones duplicated and shared home enabled so that people know what to do and expect.

andypugh:
> Shouldn't there be at least one example config (with an appropriate name) that demonstrates shared home?

This was mentioned in the issue report too. Maybe it's in here and we haven't spotted it?

Sigma1912:
Thanks for the feedback. I've added the config 'axis/home_is_shared'.

hdiethelm:
Looks good! You can remove:
loadrt or2 names=or2_0 in simulated_home.hal (https://github.com/Sigma1912/linuxcnc/blob/27ccff23f440d1ef23dea08118df8bdf2873e82b/lib/hallib/simulated_home.hal#L2C1-L2C23)
That module is not needed any more I think.

andypugh:
> Looks good! You can remove: loadrt or2 names=or2_0 in simulated_home.hal (https://github.com/Sigma1912/linuxcnc/blob/27ccff23f440d1ef23dea08118df8bdf2873e82b/lib/hallib/simulated_home.hal#L2C1-L2C23) That module is not needed any more I think.

Was this change made?

Sigma1912:
> Was this change made?

Yes

Sigma1912:
This is ready to be merged as far as I'm concerned.

PR #3870, opened 2026-03-19
Fix halcmd debug (uspace only)

hdiethelm:
As described in https://github.com/LinuxCNC/linuxcnc/pull/3866 the `hallcmd debug` was not working.

I fixed this by forwarding the command to `rtapi_app` and add the debug command also in there.

Caviat: This works only in uspace mode.

ToDo: Test with rtlinux. I tested only without so far.

### DIFF
--- a/src/hal/utils/halcmd_commands.cc
+++ b/src/hal/utils/halcmd_commands.cc
@@ -241,12 +241,18 @@ int do_unlinkp_cmd(char *pin)
 }
 
 int do_set_debug_cmd(char* level){
-    int new_level = atoi(level);
-    if (new_level < 0 || new_level > 5){
-        halcmd_error("Debug level must be >=0 and <= 5\n");
-        return -EINVAL;
-    }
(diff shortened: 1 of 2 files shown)


hdiethelm:
My testing version with some additional printf's is here if someone wants to see how this works:
https://github.com/hdiethelm/linuxcnc-fork/tree/fix_debug_hacking

If I find time, i will test it also on rtlinux. If someone else has time, feel free to test it also.

Sadly, I can not correct this for RTAI. I do not have a setup to test it. But I added an error message to notify others that this is not yet implemented.

hdiethelm:
How I tested (from src folder):
- Open LinuxCNC, axis_mm example: ../scripts/linuxcnc
- G0 X10 Y0
- No output
- ../bin/halcmd debug 5
- G0 X0 Y0
- Terminal shows: SET_LINE
- ../bin/halcmd debug 1
- G0 X10 Y0
- No output

So far I tested in a VM.
RTLinux and Xenomai work fine.

PR #3871, opened 2026-03-20
CFLAGS: Replace instances of -Os with -O2

NTULINUX:
-Os or -O2 should have no effect on doubles.

### DIFF
--- a/src/Makefile.inc.in
+++ b/src/Makefile.inc.in
@@ -77,7 +77,7 @@ RTAI = @RTAI@
 RTFLAGS = @RTFLAGS@ @EXT_RTFLAGS@
 KERNELDIR = @KERNELDIR@
 RTFLAGS := -I. -I@RTDIR@/include $(RTFLAGS) -DRTAPI -D_GNU_SOURCE -Drealtime
-ULFLAGS = -Wall -g -I. -I@RTDIR@/include -DULAPI -D_GNU_SOURCE -Os -DLOCALE_DIR=\"$(localedir)\" -DPACKAGE=\"$(package)\"
+ULFLAGS = -Wall -g -I. -I@RTDIR@/include -DULAPI -D_GNU_SOURCE -O2 -DLOCALE_DIR=\"$(localedir)\" -DPACKAGE=\"$(package)\"
 MODULE_EXT = @MODEXT@
 BUILD_SYS = @BUILD_SYS@
 CC = @CC@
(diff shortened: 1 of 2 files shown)


BsAtHome:
Good catch. I thought I caught them all some time ago, but apparently, not so...

NTULINUX:
Question, what are your thoughts on adding `-fno-unwind-tables -fno-asynchronous-unwind-tables` or at least `-ffunction-sections -fdata-sections` in CFLAGS with `-Wl,--gc-sections` in LDFLAGS to further reduce binary size?

Maybe tie the unwind tables flags to a configure option to disable a debug build? The \*section\* flags though can be default with no harmful consequences I think.

BsAtHome:
I don't think that the size of the binaries should be any concern. They are peanuts when stripped and running. The bin, lib and rtlib directories are <i>combined</i> less than 25MByte stripped (less than 80MByte unstripped). This is not a multi-GigiByte executable project. The GUI stuff is probably the biggest memory hog while running, which is for parts out of our control. Anything else is mostly transient usage. Even running RT is measured in MegaBytes and not GigaBytes.

The question is how much "dead code" you can get rid of in the link phase using `-ffunction-sections` and `-fdata-sections`. My guess is not too much (much library code dyn-linked via .so files anyway). But you are welcome to try and and share some statistics with us all.

NTULINUX:
There's so many spots where the CFLAGS and LDFLAGS go, should I sub them all or any ideas how to force them tree-wide?

BsAtHome:
Well, there is the real problem... You managed to find the forest and are asking for the location of the trees. I'd like that answer too ;-)

There are different places for different code. The RT stuff is generally separate from the userland stuff. You should be able to hook into this for the userland code from configure (see example in `configure.ac` function `add_to_cflags()` on line 1657). You should be able to influence LDFLAGS there too, but I've never tried that.

NTULINUX:
> Well, there is the real problem... You managed to find the forest and are asking for the location of the trees. I'd like that answer too ;-)
> 
> There are different places for different code. The RT stuff is generally separate from the userland stuff. You should be able to hook into this for the userland code from configure (see example in `configure.ac` function `add_to_cflags()` on line 1657). You should be able to influence LDFLAGS there too, but I've never tried that.

I'M ON IT! :D

edit: Definitely not worth the trouble, hardly any change in size.

PR #3876, opened 2026-03-23
Port to PyQt6

bgermann:
This is a completely mechanical conversion from PyQt5 to PyQt6. This is not tested to run successfully yet.

### DIFF
--- a/lib/python/qtvcp/designer/install_script
+++ b/lib/python/qtvcp/designer/install_script
@@ -82,35 +82,9 @@ fi
 echo -e '\ncreate plugin link at:\n~/.designer/plugins/python/\n'
 ln -s "$pifile" ~/.designer/plugins/python/
 
-arch=$(uname -m)
-sopath=/usr/lib/$arch-linux-gnu/qt5/plugins/designer/
-p2sofile=libpyqt5_py2.so
-p3sofile=libpyqt5.so
-
-# ensure we have a python3 .so file
-if [ ! -f "$sopath/$p3sofile" ]; then
-    # check if we have a python3 .so backup file from a python2 installation
-    if [ -f "$sopath/$p3sofile.old" ]; then
-        echo -e '\nrename libpyqt5.so.old file to libpyqt5.so'
-        sudo mv "$sopath/$p3sofile.old" "$sopath/$p3sofile"
-    # otherwise report an error
-    else
-        echo -e '\nlibpyqt5.so is missing from:'
-        echo -e '/usr/lib/x86_64-linux-gnu/qt5/plugins/designer/'
-        echo -e '\ndesigner installation aborted...\n'
-        exit
-    fi
-fi
-
-# remove python2 .so file if it exists
-if [ -f "$sopath/$p2sofile" ]; then
-    echo -e '\nremove python2 .so file'
-    sudo rm "$sopath/$p2sofile"
-fi
-
 #clear
 echo -e '\nInstallation complete, designer can be started with:'
-echo -e '\ndesigner -qt=5\n'
+echo -e '\ndesigner -qt=6\n'
 if [ "$response" -eq 2 ]; then
     echo -e 'After setting the rip-environment\n'
 fi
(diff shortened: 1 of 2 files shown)


Sigma1912:
> This is not tested to run successfully yet.

In that case you might want to mark this PR as 'Draft''

snowgoer540:
I am curious, did you write this, or did a LLM write this for you?

I am running into several issues here.

1. Creating a .diff of this pull request fails to apply.  

error: patch failed: configs/sim/woodpecker/compensate.py:21
error: configs/sim/woodpecker/compensate.py: patch does not apply

I removed the changes to compensate.py from the diff and the rest applies. 

2. python3-pyqt6.qtwebkit does not exist.  At best it's python3-pyqt6-qtwebengine.  The rest seemed to install.
3. Trying to run qtplasmac, I run into:
ImportError: cannot import name 'Q_ENUM' from 'PyQt6.QtCore' (/usr/lib/python3/dist-packages/PyQt6/QtCore.abi3.so)

So I tried:

from PyQt6.QtCore import QEnum as Q_ENUM

That failed and further research tells me that QEnum was not included into pyqt6 until later (6.5+) versions. 

I am currently running Debian 12 - Bookworm on this VM.  PyQt6.QtCore version tops out at 6.4.2.  

I am not sure if that stops your efforts, LinuxCNC is notoriously bad at drawing a line in the sand for supported Debian distros, but I think Boomworm is still very much supported.  

Obviously there are ways around all of this, but it's going to be quite an effort I think.

What are you hoping to gain in the shift from pyqt5 to pyqt6?

andypugh:
> I am not sure if that stops your efforts, LinuxCNC is notoriously bad at drawing a line in the sand for supported Debian distros, but I think Boomworm is still very much supported.

Here is the sand, with some lines in it:

snowgoer540:
> Here is the sand, with some lines in it:
> 

Thanks Andy.  Admittedly, I didn't know this existed.  If I read it right, looks like Master min is Debian 12 - Bookworm.

bgermann:
Yes, I have used a LLM, which is why the QVariant type is not dealt with in a proper manner.
Basically, I want to provide a way for this to keep the PDF functionality with [python-poppler-qt5 gone from forky](https://bugs.debian.org/1129685). I took this as an opportunity to upgrade to PyQt6 but as you want to keep bookworm support, it is probably a better idea to implement the following: Use QtPy to support both PyQt5 and PyQt6 (and PySide6 if you want). Keep the poppler-qt5 support for Qt5 and use QtPdf for Qt6. With this, Debian unstable/forky can move to PyQt6 and drop poppler-qt5 while you can support bookworm.

This approach will also make the diff tiny because the QtPy API is PyQt5-based, so basically only the imports change.

snowgoer540:
> Yes, I have used a LLM, which is why the QVariant type is not dealt with in a proper manner. Basically, I want to provide a way for this to keep the PDF functionality with [python-poppler-qt5 gone from forky](https://bugs.debian.org/1129685). I took this as an opportunity to upgrade to PyQt6 but as you want to keep bookworm support, it is probably a better idea to implement the following: Use QtPy to support both PyQt5 and PyQt6 (and PySide6 if you want). Keep the poppler-qt5 support for Qt5 and use QtPdf for Qt6. With this, Debian unstable/forky can move to PyQt6 and drop poppler-qt5 while you can support bookworm.
> 
> This approach will also make the diff tiny because the QtPy API is PyQt5-based, so basically only the imports change.

Ok, I understand the situation a bit better now.  

@c-morley and I are the two developers actively working on GUIs that depend on pyqt5.  Hopefully Chris will weigh in here, but I think it best to do as you suggested, that makes testing for that specific case pretty easy and allows you to move forward with dropping poppler from Forky.  It also gives us some time to work on porting our GUIs over to pyqt6, since it seems like that's coming sooner rather than later.

c-morley:
I thank you for looking at this issue - PyQy6 was on my agenda.
I was hoping linuxcnc 210 would be released soon then switch to PyQt6.
Otherwise some sort of automatic qt5/6 mechanism would be needed.
I'll have to look at Qtpy in more depth.

bgermann:
Closing this in favour of #3881

PR #3881, opened 2026-03-27
Port QtVCP from PyQt5 to QtPy

bgermann:
Replace all PyQt5-specific imports and idioms with their QtPy equivalents, making the codebase compatible with PyQt5 and PyQt6.

Import changes:
- Replace `from PyQt5.XXX import` with `from qtpy.XXX import`
- Replace bare `from PyQt5 import XXX` with `from qtpy import XXX`
- Replace `pyqtSignal`, `pyqtSlot`, `pyqtProperty` with the qtpy canonical names `Signal`, `Slot`, `Property`
- Port `PyQt5.QtDesigner` imports to `qtpy.QtDesigner`
- Port `PyQt5.Qsci` imports to `qtpy.Qsci`

PDFViewer: add QtPdf backend as Qt6 alternative to popplerqt5
- Try `qtpy.QtPdf.QPdfDocument` first; fall back to popplerqt5
- Use `QPdfDocumentRenderOptions` with Annotations flag for render quality matching the popplerqt5 Antialiasing/TextAntialiasing hints

sys_notify: make DBus main loop integration backend-agnostic
- Try dbus.mainloop.pyqt5, then pyqt6.

qtplasmac handlers: replace `find_spec("PyQt5.QtWebEngineWidgets")`
- Use a module-level `_WEBENGINE_AVAILABLE` flag set by try/import of `qtpy.QtWebEngineWidgets` in the qtplasmac handler

plugins/status_label_plugin: fix pyqtProperty descriptor check
- Replace fragile `'PyQt5.QtCore.pyqtProperty' in str(...)` with `isinstance(..., Property)` to work across all Qt bindings


andypugh:
@c-morley The description sounds like a good approach. What do you think?

snowgoer540:
Made some time to apply this to the project and test it out. 

After installing the necessary pyqt dependency, I get these in the terminal window when running the qtplasmac GUI:

```
QSocketNotifier: Can only be used with threads started with QThread
QSocketNotifier: Can only be used with threads started with QThread
QSocketNotifier: Can only be used with threads started with QThread
QSocketNotifier: Can only be used with threads started with QThread
```

bgermann:
> QSocketNotifier: Can only be used with threads started with QThread

I have removed the dbus.mainloop.glib fallback (which was added for PySide which we do not need to support) so this should be fixed.

snowgoer540:
It still gave the same QSocketNotifier message.  In order to fix it, I had to change the order of the import test to prefer testing pyqt5 first. 

From:
`for _mod in ('dbus.mainloop.pyqt6', 'dbus.mainloop.pyqt5'):`

To:
`for _mod in ('dbus.mainloop.pyqt5', 'dbus.mainloop.pyqt6'):`

bgermann:
> It still gave the same QSocketNotifier message. In order to fix it, I had to change the order of the import test to prefer testing pyqt5 first.

Applied to the PR.

c-morley:
hmm on my system I get:

    from qtpy.Qsci import QsciScintilla
ModuleNotFoundError: No module named 'qtpy.Qsci'

c-morley:
if in gcode_editor, I change to: 


try:
    from qtpy.Qsci import QsciScintilla, QsciLexerCustom, QsciLexerPython
except ImportError as e:
    from PyQt5.Qsci import QsciScintilla, QsciLexerCustom, QsciLexerPython

then dragon runs.
More testing needed...

bgermann:
That is interesting because I have actually verified that module to work on Debian bookworm. Support for qtpy.Qsci was introduced in [QtPy 2.3.0](https://github.com/spyder-ide/qtpy/blob/master/CHANGELOG.md#version-230-2022-11-07). So probably you are using an older version.

snowgoer540:
If the goal of this pull request is to allow Debian to drop python3-poppler-qt5, then I think we should limit the changes to only what is necessary to support the necessary updates in PDFViewer.py.

I understand and appreciate the groundwork being done with this commit to support our future conversion to pyqt6, but that introduces additional changes that require extensive testing and review.  Given that our master branch is approaching a release state, this may not be the right time to include the broader refactoring.

At this point, I recommend we focus only on the changes necessary to remove poppler from the codebase so Debian can proceed with dropping the package from Forky.  The rest can be handled by Chris and myself in a separate effort where the changes can be reviewed and tested thoroughly (after 2.10 is released).

bgermann:
That is okay. The Debian package can just drop the optional dependency. There is nothing in this PR that suggests to integrate it very soon.

c-morley:
Yes I'm using qtpy 2.0

adding something like this should cover old versions:

# load this after Logging set up so we get a nice dialog.
try:
    from qtpy.Qsci import QsciScintilla, QsciLexerCustom, QsciLexerPython
except ImportError as e:
    from qtpy import QT_VERSION as qtv

    if '6.' in qtv:
        msg = "Can't import QsciScintilla - is package python3-pyqt6.qsci installed?"
        LOG.critical(msg, exc_info=e)
        sys.exit(1)

    else:
        # for older versions of qtpy check pyqt5 explicitly
        try:
            from PyQt5.Qsci import QsciScintilla, QsciLexerCustom, QsciLexerPython
        except:
            msg = "Can't import QsciScintilla - is package python3-pyqt6.qsci installed?"
            LOG.critical(msg, exc_info=e)
            sys.exit(1)

And more testing didn't show any other problems/
Did you confirm that using pyqt6 works?

bgermann:
Yes, I have confirmed the pyqt6 version to start. However, I am not a regular user, so I have not verified the expected program behavior/design.

NTULINUX:
This will fix LinuxCNC's Qt interfaces on Gentoo! LOVE IT!!

c-morley:
@snowgoer540, I think we should take the plunge and pull this in.
Do you wish to do more QtPlasmac testing first?

snowgoer540:
> @snowgoer540, I think we should take the plunge and pull this in. Do you wish to do more QtPlasmac testing first?

You've convinced me.  Plus there's nothing here another pull request couldn't fix :-)

c-morley:
@bgermann thank you !

c-morley:
> Yes, I have confirmed the pyqt6 version to start. However, I am not a regular user, so I have not verified the expected program behavior/design.

Are you sure it was using pyqt6? I have tried it on a newer system and it needed tons of fixes and I never did get it to show a screen yet.
What linuxcnc system did you use?

bgermann:
You are right, probably I had both installed. Anyway, I have now verified it with a subsequent PR.

Issue #3875, opened 2026-03-22
Ruckig broke RTAI


BsAtHome:
The inclusion of the type 1 jerk limited trajectory planner, with inclusion of the ruckig code, completely break compiling RTAI. The problem is that ruckig is C++ code and that you cannot have in the kernel. Also, the code uses libc calls not available in kernel land.

The solution is one of:
1. remove the ruckig code for RTAI (removing this TP)
2. rewrite the ruckig code as plain C
3. take the trajectory planner out of the kernel and RT

Option 3 would be best, but either way, RTAI remains broken until this issue is solved.

[edit]
Or option 4: remove RTAI before the bitrot infects other parts.

NTULINUX:
As long as the classic TP works. I wouldn't remove support for RTAI just yet, I think Andy still uses it. Some systems do not play nice with PREEMPT_RT as well due to BIOS or motherboard issues.

BsAtHome:
> As long as the classic TP works. I wouldn't remove support for RTAI just yet, I think Andy still uses it.

Yes, but he surely hasn't compiled master for RTAI recently or the trouble would have been obvious. And he himself merged the ruckig code :-)

The problem is that it needs to be _compiled out_ for RTAI to function. That is a strategically losing proposition in master and makes things utterly ugly and complex. There is supposed to be plugability of the TP, but afaics, the way it is compiled suggests otherwise.

> Some systems do not play nice with PREEMPT_RT as well due to BIOS or motherboard issues.

That is the first time I hear about that... Can you be more specific? Do you mean hardware no longer supported by newer kernels or what?

NTULINUX:
> That is the first time I hear about that... Can you be more specific? Do you mean hardware no longer supported by newer kernels or what?

Latency will be over a million microseconds with PREEMPT_RT but under 10 with RTAI. It's obviously not every system but having RTAI as a fallback will enable systems that absolutely cannot get low latency with PREEMPT_RT regardless of BIOS settings can still use RTAI and get quite reasonable scores. The Intel system I had a few months ago was an i5-1135g7, speed stepping, removing all the power management/savings things, nothing worked.

BsAtHome:
> Latency will be over a million microseconds with PREEMPT_RT...

One million microseconds is one second. That number you mention here is either very wrong or hyperbole.

And did you boot the same kernel for RTAI and RT_PREEMPT testing? Kernel versions do matter.

NTULINUX:
[...]

PR #3884, opened 2026-03-28
Replace C++ Ruckig with pure C cruckig to fix RTAI builds

grandixximo:
The C++ Ruckig library cannot compile in RTAI kernel space (no C++ in kernel modules). Replace it with cruckig, a pure C99 port that produces identical trajectories (validated to 5.9e-12 position error).

- Vendor cruckig source (MIT license) from github.com/mika4128/cruckig
- Rewrite ruckig_wrapper.c to use cruckig API (same ruckig_wrapper.h interface, zero changes needed in tp.c, tc.c, sp_scurve.c)
- Add cruckig_compat.h for RTAPI kernel/userspace portability (rtapi_slab.h for malloc/free, rtapi_math.h for math functions)
- Update Makefile and motion-logger Submakefile for C-only build
- Remove C++ Ruckig source tree and ruckig_wrapper.cc

Fixes: https://github.com/LinuxCNC/linuxcnc/issues/3875

-------------------

Can someone try and see if it builds and work on RTAI?


BsAtHome:
Fast review...

You can not use standard libc includes (like f.ex. stddef and stdbool) or calls in your code. They only work in userland. You must use the RTAPI library because that was specifically created to work in both kernelland and userspace.

BTW, have you tested whether you need to have the includes almost everywhere? F.ex. you include cruckig/profile.h, which includes stddef and stdbool and that gets included may other places that _also_ include stddef and stdbool. I think you need to review whether or not includes are necessary or implicit by the use of other includes.

Your include spec using `<>` suggests that these are global includes (those the get exported in the dev package). However, this code is local to the TP and should therefore be addressed differently. You then also can drop the "include" part of the directory (rename `include/cruckig` to `cruckig`) and include all files using `#include "cruckig/whatever.h"`. That works as it is relative to the current directory where your includes will live. Then we don't need special treatment in the Makefile (remove the extra INCLUDE addition).

Function `fabs()` is not available in kernel space (nor userland) if you do not include rtapi_math.h.

Your diversion in include/cruckig/cruckig_compat.h is wrong. You must always thunk to `rtapi_*()` and include rtapi_slab.h for memory allocation.

Are you sure no memory allocations/frees are called dynamically? I see a lot of `*_create()` and `*_destroy()` functions that allocate and free dynamic memory. That is not what you what to run on every cycle.

Thats it for a first glance... ;-)

grandixximo:
> Fast review...
>
> You can not use standard libc includes (like f.ex. stddef and stdbool) or calls in your code. They only work in userland. You must use the RTAPI library because that was specifically created to work in both kernelland and userspace.

Fixed. All handled through a single `cruckig_compat.h` header that every cruckig source includes.

> BTW, have you tested whether you need to have the includes almost everywhere? F.ex. you include cruckig/profile.h, which includes stddef and stdbool and that gets included may other places that _also_ include stddef and stdbool. I think you need to review whether or not includes are necessary or implicit by the use of other includes.

Cleaned up. Removed redundant `stdbool`/`stddef` includes from `.c` files that already get them transitively through `cruckig_compat.h`. The headers themselves include `cruckig_compat.h` once, which provides `bool`, `size_t`, math, and string functions.

> Your include spec using `<>` suggests that these are global includes (those the get exported in the dev package). However, this code is local to the TP and should therefore be addressed differently. You then also can drop the "include" part of the directory (rename `include/cruckig` to `cruckig`) and include all files using `#include "cruckig/whatever.h"`. That works as it is relative to the current directory where your includes will live. Then we don't need special treatment in the Makefile (remove the extra INCLUDE addition).

Done. Flattened `include/cruckig/*.h` into `cruckig/*.h`, fixes all includes, fixed the Makefile.

> Function `fabs()` is not available in kernel space (nor userland) if you do not include rtapi_math.h.

Fixed. `cruckig_compat.h` always includes `<rtapi_math.h>` unconditionally, so all cruckig sources get it.

> Your diversion in include/cruckig/cruckig_compat.h is wrong. You must always thunk to `rtapi_*()` and include rtapi_slab.h for memory allocation.

Fixed. Removed the `#ifdef RTAPI` split entirely. `cruckig_compat.h` now always uses `rtapi_slab.h`, which already handles the kernel vs userspace dispatch internally. Same for `ruckig_wrapper.c` itself, which now uses `<rtapi_slab.h>` directly.

> Are you sure no memory allocations/frees are called dynamically? I see a lot of `*_create()` and `*_destroy()` functions that allocate and free dynamic memory. That is not what you what to run on every cycle.

Verified. The allocation pattern is init-only, not per-cycle:

- `ruckig_create()` is called once per TC_STRUCT (in `tp.c` when `!tc->ruckig_planner`) or once globally in `sp_scurve_init()`. Not per servo cycle.
- `ruckig_destroy()` is called only in `tcCleanupRuckig()` at segment end or `sp_scurve_cleanup()` at shutdown.
- `cruckig_calculate()` (called from `ruckig_plan_position`/`ruckig_plan_velocity` on replanning) does NOT allocate memory for single-segment trajectories. All work buffers (`blocks`, `pd`, `possible_t_syncs`, etc.) are pre-allocated inside `cruckig_calculator_create()` at init time and reused on every call.
- The only per-planner allocation in the wrapper is `min_velocity` (one `double`), allocated once on first use (guarded by `if (impl->input->min_velocity == NULL)`).

The waypoint code path does have dynamic allocations, but LinuxCNC only uses single-segment planning, so those paths are never hit.

> Thats it for a first glance... ;-)

Thanks for the review! Force-pushed with all fixes applied. Build passes clean.

BsAtHome:
Lets continue review then...

Just a comment, you flattened the directory structure one level more than I suggested... This also works, so no prob.

The double inclusion protection in cruckig.h is different from the others. You need to use `CRUCKIG_CRUCKIG_H` to be consistent. I.e. always and consistently prepend `CRUCKIG_` to the name.

You include cruckig_compat.h in both headers and c-files. That is double inclusion by definition and you should choose on cmethod consistently. So, you need to review your includes (again). Also, you'd normally include such a file as the first in the C-file followed by the include for the C-files. The alternative is to include the compat header consistently in the includes. You need to pick a method. You also need to consider what _internal_ headers describe and what _interface_ headers describe for linking against the code. You need to make a clean split and I'm not sure I'm seeing it. Or maybe I just cannot see the trees through the forest ;-)

BTW, if cruckig_config.h is such an important universal include, then why is it not part of cruckig_compat.h? Then the name suffix "compat" is also a misnomer. Compatibility to what exactly is this code pretending to be? Maybe you don't need two include files? Just thinking out loud (well, writing out loud). That thought would result in a more natural split that (cleanly) exposes your API in cruckig.h and does your (dirty) internals in cruckig_internal.h (and then your local internal per-file headers).

Why do you define `_POSIX_C_SOURCE` in cruckig.c? That does not work for kernelland.

In cruckig_compat.h you make distinction using `__KERNEL__`. However, including rtapi.h as the first should take care of that whole section.

In cruckig_config.h you have a comment at the beginning about optimization that makes no sense and has no actual content to go with the comment. Also, does anybody suggest that a test for `_MSC_VER` would make sense for this code?

You have a define for M_PI in roots.c (and `_USE_MATH_DEFINES`). That should not be there (handled by rtapi_math.h). And, FWIW, running with C99 is overrated because the oldest supported version distro already, at least, uses C11 and C17 has been available for a long time when we look at the master branch dependencies. It would be very safe to assume C11 and we'd probably should set it to C17 (just like we require C++20).

In trajectory.c you have a conditional if 16 DOF to do allocation (line 206). If that is never hit, then why is there code to support it? Wouldn't it be simply safer to say that DOF <= 16 is the implementation limit?

So far my second look.

grandixximo:
> Just a comment, you flattened the directory structure one level more than I suggested... This also works, so no prob.

Noted. Headers now live directly in `cruckig/` alongside the `.c` files.

> The double inclusion protection in cruckig.h is different from the others. You need to use CRUCKIG_CRUCKIG_H to be consistent. I.e. always and consistently prepend CRUCKIG_ to the name.

Fixed. `cruckig.h` now uses `CRUCKIG_CRUCKIG_H`, same pattern as all others.

> You include cruckig_compat.h in both headers and c-files. That is double inclusion by definition and you should choose one method consistently.

Fixed. The new approach: `cruckig_internal.h` is included only by the headers (as their first include). C files do NOT include it directly, they get it transitively through their own header. One method, consistently applied.

> Also, you'd normally include such a file as the first in the C-file followed by the include for the C-files. The alternative is to include the compat header consistently in the includes. You need to pick a method.

Picked the second method: internal header included consistently in every `.h` file. C files include only their own header.

> BTW, if cruckig_config.h is such an important universal include, then why is it not part of cruckig_compat.h? Then the name suffix "compat" is also a misnomer.

Agreed. Merged `cruckig_config.h` and `cruckig_compat.h` into a single `cruckig_internal.h`. It provides the RTAPI type/math/string/memory portability layer plus the compiler hint macros (LIKELY/UNLIKELY, FORCE_INLINE, HOT, RESTRICT, PREFETCH). Both old files are deleted.

> Why do you define _POSIX_C_SOURCE in cruckig.c? That does not work for kernelland.

Removed. It was there for `clock_gettime()` in `get_time_us()`, which is only used by `cruckig_update()` (the online real-time loop). LinuxCNC uses the offline `cruckig_calculate()` + `cruckig_trajectory_at_time()` path instead. Stubbed `get_time_us()` to return 0.0 unconditionally.

> In cruckig_compat.h you make distinction using __KERNEL__. However, including rtapi.h as the first should take care of that whole section.

Fixed. `cruckig_internal.h` now includes `<rtapi.h>` as the first include. The `__KERNEL__` / `<stddef.h>` section is gone, rtapi.h handles the kernel/userspace dispatch.

> In cruckig_config.h you have a comment at the beginning about optimization that makes no sense and has no actual content to go with the comment. Also, does anybody suggest that a test for _MSC_VER would make sense for this code?

Both removed. The optimization level comment and MSVC fallback are gone from `cruckig_internal.h`. GCC/Clang only, with a plain fallback for anything else.

> You have a define for M_PI in roots.c (and _USE_MATH_DEFINES). That should not be there (handled by rtapi_math.h).

Removed both `_USE_MATH_DEFINES` and the `M_PI` fallback define from `roots.c`. `rtapi_math.h` (included via `cruckig_internal.h`) provides `M_PI`.

> And, FWIW, running with C99 is overrated because the oldest supported version distro already, at least, uses C11 and C17 has been available for a long time when we look at the master branch dependencies. It would be very safe to assume C11 and we'd probably should set it to C17.

Agreed. The cruckig code itself doesn't depend on any C99-specific features that aren't in C11/C17. The RESTRICT macro already has a `__STDC_VERSION__` check but defaults to the GCC extension which works regardless.

> In trajectory.c you have a conditional if 16 DOF to do allocation (line 206). If that is never hit, then why is there code to support it? Wouldn't it be simply safer to say that DOF <= 16 is the implementation limit?

Done. Replaced the heap allocation fallback with a hard cap: `const size_t ndofs = (dofs > 16) ? 16 : dofs;`. Stack-only, no dynamic allocation in the hot path. LinuxCNC uses 1 DOF through the wrapper, so 16 is more than enough.

Force-pushed with all fixes. Build passes clean.

BsAtHome:
Shouldn't all files start with a license reference? You have a `LICENSE` file, but you may want to refer to that and mark the copyright dates/authors in each source. Maybe something like:
```
<short one-liner description project/file>
Copyright (c) <year>  <name>
Copyright (c) <otheryear> <othername>

License MIT, see the LICENSE file in this directory.
```

BTW, this license is (correctly as per Debian standard) referred to as Expat in the debian/copyright reference. You will need to update the directory name and author(s) there as well (line 154). And also the wrapper ref (line 159).

grandixximo:
> Shouldn't all files start with a license reference? You have a `LICENSE` file, but you may want to refer to that and mark the copyright dates/authors in each source. Maybe something like:
>
> ```
> <short one-liner description project/file>
> Copyright (c) <year>  <name>
> Copyright (c) <otheryear> <othername>
>
> License MIT, see the LICENSE file in this directory.
> ```
>
> BTW, this license is (correctly as per Debian standard) referred to as Expat in the debian/copyright reference. You will need to update the directory name and author(s) there as well (line 154). And also the wrapper ref (line 159).

Good tips, thanks. Both done:

Every `.c` and `.h` file in `cruckig/` now starts with:

```c
/*
 * cruckig - Pure C99 port of the Ruckig trajectory generation library
 * Copyright (c) 2025 Yang Yang <mika-net@outlook.com>
 * Copyright (c) 2021 Lars Berscheid (original C++ Ruckig)
 *
 * License: MIT, see the LICENSE file in this directory.
 */
```

And `debian/copyright` updated -- `src/emc/tp/ruckig/*` changed to `src/emc/tp/cruckig/*` with both authors listed, wrapper entry copyright extended to 2026.

Force-pushed.

BsAtHome:
> Removed. It was there for `clock_gettime()` in `get_time_us()`, which is only used by `cruckig_update()` (the online real-time loop). LinuxCNC uses the offline `cruckig_calculate()` + `cruckig_trajectory_at_time()` path instead. Stubbed `get_time_us()` to return 0.0 unconditionally.

There is `rtapi_get_time()` that maps to a monotonic clock and returns time in nanoseconds. If that would be any help...
Not sure why you'd need a clock reference in the planner, unless you start to iterate for convergence. But if you do that, then you should not be in realtime as you generally only want constant time calculations (O(1)) in realtime (yes, there are exceptions, but you know what I mean).

grandixximo:
> There is rtapi_get_time() that maps to a monotonic clock and returns time in nanoseconds. If that would be any help...
> Not sure why you'd need a clock reference in the planner, unless you start to iterate for convergence. But if you do that, then you should not be in realtime as you generally only want constant time calculations (O(1)) in realtime (yes, there are exceptions, but you know what I mean).

The clock is not used by LinuxCNC at all. It exists for two features in the cruckig/ruckig library that we don't use:

1. **`cruckig_update()` the online real-time loop:** this is a different API pattern where you call `cruckig_update()` every cycle and it both (re)plans and samples the trajectory in one call. The clock measures computation time and reports it in `output->calculation_duration`. Think of it as the robot-arm use case where targets change dynamically every cycle.

2. **Calculation interruption budget:** you can set `input->interrupt_calculation_duration` to a microsecond budget. If the waypoint solver exceeds it, computation is paused and resumed on the next `cruckig_update()` call. This is for multi-waypoint trajectory planning where the solver may need multiple iterations, and you want to bound worst-case cycle time.

LinuxCNC uses neither. We use the offline path, `cruckig_calculate()` runs the solver to completion (single-segment, 1 DOF, O(1) as you'd expect), then `cruckig_trajectory_at_time()` samples it. No iteration, no interruption, no clock needed. That's why the stub returning 0.0 is correct: it effectively means "infinite budget, never interrupt," which is the right behavior for an offline calculation that always completes in bounded time.

So no need to wire up `rtapi_get_time()`. The feature is there because cruckig is a general-purpose library, but LinuxCNC only exercises the simple path.

grandixximo:
**Fixed kernel stack frame size** in `calculator.c`: `CRuckigPositionThirdOrderStep1` is 3.6KB (contains 6 profile structs for the polynomial solver), which triggers `-Wframe-larger-than=2048` in kbuild. Moved the Step1 workspace into the heap-allocated `CRuckigCalculator` struct as a union, so the large structs stay off the kernel stack.

Also ran clang build and `scripts/cppcheck.sh`, both clean.

Force-pushed.

NTULINUX:
Pure genius right here!

NTULINUX (approved):

grandixximo:
@andypugh 

can you merge this? kind of hard to keep working on RTAI fixes without this in upstream, I don't want to sneak it in with another PR, but those PR won't build without this in.

Issue #3880, opened 2026-03-27
nested subroutines silently ignored


djdelorie:
Consider the following named subroutine in a file nested.ngc:

o<nested>sub
o100 sub
(DEBUG, #1 #2 #3)
o100 endsub
o100 call [4] [5] [6]
o<nested>endsub
M2

Now if you call this with this MDI command:

o<nested>call [7] [8] [9]

One would expect the DEBUG popup to have "4 5 6" in it.  However, it has "7 8 9" in it.  FYI this cost me a probe tip :-P

If the o100 sub is before the o<nested>sub, the parser doesn't see it at all and you get "EOF in nested.ngc seeking o-word".

If the o100 sub is after the o<nested>endsub, it functions correctly.

As far as I can tell, the o100 sub is ignored but the o100 endsub ends the o<nested>sub - the rest of the file is ignored.

THE BUG IS:

1. No errors are emitted if a nested sub is detected, or if a numbered sub comes before a named sub.
2. The requirement that numbered subs follow named subs is not documented in the named sub docs, and only briefly hinted ("no nesting") in the o-call example.  Those docs say "a subroutine may not be called until it is defined" but that isn't true for a numbered sub preceeding a named sub.
3. The named sub docs say "A named subroutine file can contain only a single subroutine definition." but this excludes numbered subs, which seem to be supported, but must follow the named sub.

Reproduced with 2.8, 2.9, and 2.10.0-git

Sigma1912:
> osub
...
oendsub

I'm confused. IIRC the syntax is
```

o<somename> sub 
...
o<somename> endsub
```

Sigma1912:
Ah, I see, the editor ate it.

Should be:
```
o<nested> sub
o100 sub
(DEBUG, #1 #2 #3)
o100 endsub
o100 call [4] [5] [6]
o<nested> endsub
M2
```

and 

`o<nested> call [7] [8] [9]`

Sigma1912:
Apparently the nested sub is only caught if the 'nested.ngc' file is loaded directly in the gui but not if it is called by `o<nested> call` .

grandixximo:
AFAIK sub definitions are first-level only. The docs explicitly state "Subroutine bodies may not be nested." So your code is invalid and produces unexpected behavior.

[...]

PR #3886, opened 2026-03-29
Detect and error on nested/invalid subroutine definitions in named subs

grandixximo:
The parser silently ignored sub definitions inside called subroutines, causing wrong parameters and premature returns. Numbered subs after a named endsub polluted the global offset map and conflicted across files.

- Add nesting checks in read and execute phases at call_level > 0
- Block forward-seek for subs at call_level > 0 when no file exists
- Let control_back_to errors propagate without generic overwrite
- Update o-code docs with nesting example and new error conditions

Ref: LinuxCNC/linuxcnc#3880


andypugh:
Can I suggest adding a test to the testsuite for this?

grandixximo:
> Can I suggest adding a test to the testsuite for this?

@andypugh tests were added

PR #3888, opened 2026-03-29
debian: Move python3-poppler-qt5 to Suggests

bgermann:
The package is being removed from Debian. Having it in Suggests is a way to signal that you might need it for optional functions.
The code can already deal with the package not being installed.


andypugh:
Your PR and my change doing exactly the same thing overlapped. 
I have addressed the issue in the 2.9 branch as that is the production branch, being uploaded to Debian.

Issue #3895, opened 2026-04-02
deprecate uses_fp functionality in hal_create_thread


rmu75:
Uses_fp proves to be problematic in case the compiler decides to generate SSE code, it silently corrupts FP / SSE state. This only affects RTAI.

I propose to get rid of realtime kernel threads that declare not to use FP (so every thread should be uses_fp=1), saving/restoring FP context shouldn't create measureable increase in latencies or max frequencies achievable on contemporary CPUs.

https://linuxcnc.org/docs/html/man/man9/motion.9.html
https://linuxcnc.org/docs/html/man/man3/hal_create_thread.3hal.html

grandixximo:
@andypugh @BsAtHome @smoe

Need your input on this. Pretty big change, docs, addf, .comp and .c components. I think this touches a bit of everything if it was to be done proper, where we strip out FP/non-FP differentiation out of the codebase altogether.

Hit this bug on RTAI 5.4.280 with the etch-servo parport config. System was hard-crashing (heap corruption, libc segfaults, black screen). Tracked it down to `pwmgen.make-pulses` in the base-thread.

GCC was optimizing struct zeroing in `make_pulses()` into SSE instructions (`pxor %xmm0; movups %xmm0`). Function is exported with `uses_fp=0`, so RTAI never saves/restores XMM state. Every base-thread cycle it silently corrupts whatever Linux process was using those registers.

Scanned all 217 RT modules. Right now pwmgen is the only non-FP function where GCC actually emits SSE. The others are clean today. But that's basically luck, a compiler upgrade or a small code change could break any of them and you'd never know until someone's machine starts crashing.

IMO Robert is right, just make everything `uses_fp=1` and be done with it. The ~100ns overhead for fxsave/fxrstor is nothing on modern hardware and it removes an entire class of silent bugs that are nearly impossible to diagnose.

grandixximo:
Just made a small PR to at least fix the pwmgen, this Bertho's PR and cruckig are all to be merged for RTAI to be somewhat functional, the PR does not close the issue, it is a temporary patching up IMO

BsAtHome:
On 4/2/26 1:17 PM, grandixximo wrote:
> GCC was optimizing struct zeroing in `make_pulses()` into SSE 
> instructions (`pxor %xmm0; movups %xmm0`). Function is exported with 
> `uses_fp=0`, so RTAI never saves/restores XMM state. Every base-
[...]

PR #3896, opened 2026-04-02
pwmgen: prevent GCC from emitting SSE in base-thread function

grandixximo:
make_pulses() is exported with uses_fp=0 for the base-thread, but GCC optimizes struct zeroing into SSE instructions (pxor/movups on XMM registers). Since RTAI skips FPU/SSE save/restore for non-FP threads, this silently corrupts XMM state of whatever Linux process was running, causing heap corruption, segfaults, and hard crashes.

Add __attribute__((target("general-regs-only"))) to make_pulses() to force GCC to use only general-purpose registers.

Confirmed via objdump that the compiled function contains zero SSE instructions after this change. Tested on RTAI 5.4.280 with the etch-servo parport config which previously crashed reliably.

See: https://github.com/LinuxCNC/linuxcnc/issues/3895

### DIFF
--- a/src/hal/components/pwmgen.c
+++ b/src/hal/components/pwmgen.c
@@ -135,7 +135,7 @@ static long periodns;		/* makepulses function period in nanosec */
 ************************************************************************/
 
 static int export_pwmgen(int num, pwmgen_t * addr, int output_type);
-static void make_pulses(void *arg, long period);
+static void __attribute__((target("general-regs-only"))) make_pulses(void *arg, long period);
 static void update(void *arg, long period);
 
 /***********************************************************************
@@ -231,7 +231,7 @@ void rtapi_app_exit(void)
     added and subtracted.
 */
 
-static void make_pulses(void *arg, long period)
+static void __attribute__((target("general-regs-only"))) make_pulses(void *arg, long period)
 {
     pwmgen_t *pwmgen;
     int n;


andypugh:
Is this fix still needed, or are we going for the meta-fix?

grandixximo:
I can work on ripping out non-FP in the weekend, this will not be needed once that is done, it is just a quick fix, if anyone wants to run RTAI with pwmgen in the meantime, Though I doubt there is anyone else but me testing this.

grandixximo:
superseded by #3901

andypugh:
I was about to ask for clarification about this, how do these three PRs fit together?

grandixximo:
Oh, this forcibly flagged the pwm step making function to prevented the compiler to use floating point operations in the compiled assembly. The problem was the PWM step making function is usually added to base-thread which has `uses_fp=0`, RTAI would skip operations for `uses_fp=0` because they are not supposed to be there for performance (which is not needed anymore in todays hardware), but compiler puts them in there even if unnecessary (performance landscape changed with new hardware). But since we decided (wisely) to deprecate `uses_fp` we don't need to check for this anymore, we can happily compile safely, and if the compiler thinks FP operations are faster it can compile them, and we don't crash since everything is  `uses_fp=1` with #3901 without breaking API. In the near future `uses_fp` should be completely ripped out of the API with another PR, which I can draft if you want. But this will break the API, so we are waiting for all the other stuff that does break the HAL API as well to come along aka `long_int` with getter\setter and `noparam`

PR #3900, opened 2026-04-03
Deprecate uses_fp: always enable FPU state save on all threads

grandixximo:
GCC on x86-64 may emit SSE instructions (e.g. for struct zeroing) even in functions that don't explicitly use floating point. On RTAI, threads created with uses_fp=0 skip FPU/SSE state save/restore, silently corrupting XMM registers of other Linux processes and causing heap corruption and system crashes.

Fix by always enabling FPU state save regardless of the uses_fp parameter:
- rtai_rtapi.c: always pass 1 to rt_task_init_cpuid
- hal_lib.c: override uses_fp=0 to 1 in hal_create_thread
- hal_lib.c: remove addf FP compatibility check (now meaningless)

Emit deprecation warnings when uses_fp=0 is requested.

Ref: linuxcnc/linuxcnc#3895

### DIFF
--- a/docs/man/man3/hal_create_thread.3hal
+++ b/docs/man/man3/hal_create_thread.3hal
@@ -18,7 +18,8 @@ The name of the thread
 The interval, in nanoseconds, between iterations of the thread
 
 .IP \fIuses_fp\fR
-Must be nonzero if a function which uses floating-point will be attached to this thread.
+\fBDeprecated.\fR All threads now unconditionally save and restore
+floating-point state.  This parameter is accepted but ignored.
 
 .SH DESCRIPTION
 \fBhal_create_thread\fR establishes a realtime thread that will execute one or more HAL functions periodically.
--- a/docs/man/man3/hal_export_funct.3hal
+++ b/docs/man/man3/hal_export_funct.3hal
@@ -20,8 +20,8 @@ The pointer to the function
 The argument to be passed as the first parameter of \fIfunct\fR
 
 .IP \fIuses_fp\fR
-Nonzero if the function uses floating-point operations, including assignment
-of floating point values with "=".
+\fBDeprecated.\fR All threads now unconditionally save and restore
+floating-point state, so this flag has no effect.
 
 .IP \fIreentrant\fR
 If reentrant is non-zero, the function may be preempted and called again
--- a/docs/man/man3/rtapi_task_new.3rtapi
+++ b/docs/man/man3/rtapi_task_new.3rtapi
@@ -19,7 +19,8 @@ started
 .IP \fIprio\fR
 A task priority value returned by \fBrtapi_prio_xxxx\fR
 .IP \fIuses_fp\fR
-A flag that tells the OS whether the task uses floating point or not.
+\fBDeprecated.\fR All tasks now unconditionally save and restore
+floating-point state.  This parameter is accepted but ignored.
 .IP \fItask_id\fR
 A task ID returned by a previous call to \fBrtapi_task_new\fR
 .SH DESCRIPTION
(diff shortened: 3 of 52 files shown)


grandixximo:
@andypugh I was thinking to have minimal docs and code changes for 2.9 just to address the issue directly, minimal impact, small hack but gets 2.9 working on RTAI again.
For 2.10 I will dig deeper in the docs, and the code, to completely rip out uses_fp, this will break compatibility with out of tree .comp and .c components, which we are already planning to do for the 64/32bit refactor.

BsAtHome:
I think this should also include the following (patch is against master; currently cannot easily build 2.9):
[halcompile-nofp-fp.diff.txt](https://github.com/user-attachments/files/26458637/halcompile-nofp-fp.diff.txt)

This patch changes halcompile to add a warning when you use `nofp` or `fp` in function statements. It also fixes all components I could find that did emit the warning. Please adapt to 2.9 branch.

BsAtHome:
A small typo in the warn function. It should always return 1:
```python
def optfp_warn(state):
    s = "nofp" if 0 == state else "fp"
    Warn("'{}' is not longer supported. Threads are always floating point threads.".format(s))
    return 1
```

The change in the parser rule `OptFP` should remain in place to call the warning function when the keyword is used. However, to ease the burden on 2.9 users, this may be changed slightly by keeping silent but still force the `fp` state:
```python
def optfp_warn(state):
    return 1
```

grandixximo:
Done with your patch as reference, you are doing 2.10 then? @BsAtHome

BsAtHome:
The man-pages are wrong. You are changing the generated files and not the adoc man-page source in docs/src/man/...

BsAtHome:
> Done with your patch as reference, you are doing 2.10 then? @BsAtHome

It is all your pleasure :-)

BsAtHome:
> The man-pages are wrong. You are changing the generated files and not the adoc man-page source in docs/src/man/...

Hm, yes, for master branch it is wrong... 2.9 is still of groff life support.

grandixximo:
> > The man-pages are wrong. You are changing the generated files and not the adoc man-page source in docs/src/man/...
> 
> Hm, yes, for master branch it is wrong... 2.9 is still of groff life support.

But this is 2.9, I was searching docs/src/man/
cannot find the adoc in the source.

BsAtHome:
> > > The man-pages are wrong. You are changing the generated files and not the adoc man-page source in docs/src/man/...
>>
> > Hm, yes, for master branch it is wrong... 2.9 is still of groff life support.
>
> But this is 2.9, I was searching docs/src/man/ cannot find anything in the source.

That is indeed what I said :-)
Groff is the man-page formatter for the [t]roff format used in 2.9 and earlier. Everything got converted to adoc after 2.9, including components documentation.

grandixximo:
> > > > The man-pages are wrong. You are changing the generated files and not the adoc man-page source in docs/src/man/...
> > > 
> > > 
> > > Hm, yes, for master branch it is wrong... 2.9 is still of groff life support.
> > 
> > 
> > But this is 2.9, I was searching docs/src/man/ cannot find anything in the source.
> 
> That is indeed what I said :-) Groff is the man-page formatter for the [t]roff format used in 2.9 and earlier. Everything got converted to adoc after 2.9, including components documentation.

So we are good, I did it right for 2.9 😌

grandixximo:
I did not rebuild the .po
should I do that? or not?

andypugh:
> I did not rebuild the .po should I do that? or not?

That's one for @smoe  I think.

grandixximo:
just rebased, ready for review

PR #3916, opened 2026-04-09
Fix python site-packages dir

NTULINUX:
Fixes the need for PYTHONPATH on Gentoo

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1657,7 +1657,7 @@ fi
 AC_MSG_RESULT([$PYTHON_TK_VERSION])
 
 AC_MSG_CHECKING(for site-package location)
-SITEPY="$($PYTHON -c 'import sysconfig; s = sysconfig.get_scheme_names(); m=list(set(("deb_system", "rpm_prefix")) & set(s)); print(sysconfig.get_path("platlib", m.__getitem__(0))) if m else print("/usr/lib/python3/dist-packages");')"
+SITEPY=`$PYTHON -c "import sysconfig; s=sysconfig.get_scheme_names(); print(sysconfig.get_path('platlib', 'deb_system' if 'deb_system' in s else 'rpm_prefix' if 'rpm_prefix' in s else sysconfig.get_default_scheme()))"`
 AC_MSG_RESULT($SITEPY)
 
 AC_MSG_CHECKING(for working GLU quadrics)


NTULINUX:
Looks like this needs to be fixed to account for Debian..

`dh_install: error: missing files, aborting`

Ah, looks like various parts of the tree are also hard-coded to use `/usr/lib/python3/dist-packages`

```
docs/src/plasma/qtplasmac.adoc:/usr/lib/python3/dist-packages/qtvcp/designer/install_script
lib/python/qtvcp/designer/install_script:    if [ -f /usr/lib/python3/dist-packages/qtvcp/plugins/qtvcp_plugin.py ]; then
lib/python/qtvcp/designer/install_script:        pifile=/usr/lib/python3/dist-packages/qtvcp/plugins/qtvcp_plugin.py
lib/python/qtvcp/designer/install_script:        echo -e '\n'"$pifile "'not found in /usr/lib/python3/dist-packages/qtvcp/plugins/'
scripts/setup_designer.in:    if [ -f /usr/lib/python3/dist-packages/qtvcp/plugins/qtvcp_plugin.py ]; then
scripts/setup_designer.in:        pifile=/usr/lib/python3/dist-packages/qtvcp/plugins/qtvcp_plugin.py
scripts/setup_designer.in:        echo -e '\n'"$pifile"' not found in /usr/lib/python3/dist-packages/qtvcp/plugins/'
```

Did not see these. Any tips? The issue is that depending on the exact version of python that debian uses will affect the installation path, which will vary from each debian release, that's the tricky part to make this work across all versions. Not sure on a clean way to avoid the hard-coded variables either, as those will also change.

BsAtHome:
Is there a distro independent way to find out where python packages need to be installed?

NTULINUX:
Well that's what this should do:

https://github.com/LinuxCNC/linuxcnc/blob/master/src/m4/ax_python_devel.m4#L340

That's where I got PYTHON_SITE_PKG from. There's also `python -c "import site; print(site.getsitepackages())"`

That's going to change though distro to distro... The commands themselves are universal, the result is not.

PYTHON_SITE_PKG works but `debian/linuxcnc.install.in` needs to account for it. It is hard-coded to `usr/lib/python3` as well as other parts of the tree.

`git grep -e "usr/lib/python3"`

Using PYTHON_SITE_PKG works for both a system-wide install though, and run-in-place.

I just checked in my Trixie VM and PYTHON_SITE_PKG is working as intended (both rip and system-wide install.)

Another problem is accounting for Fedora. PYTHON_SITE_PKG works there too but the hard-codes need to be fixed with a simple one-linger, not some 10 line script from autotools. Luckily, this works:

`python3 -c "import site; print([p for p in site.getsitepackages() if p.startswith('/usr/lib') and '/local/' not in p and '/lib64/' not in p][0])"`

Now I need to fix the other hard-codes but the upcoming code might be wrong..

NTULINUX:
This might be ready now. Not 100% certain.

NTULINUX:
Fixed typo, trimmed SITEPY a little. Fixed echo lines as $SITEPY was literal.

src/configure.ac
```diff
@@ -1657,7 +1657,7 @@ fi
 AC_MSG_RESULT([$PYTHON_TK_VERSION])
 
 AC_MSG_CHECKING(for site-package location)
-SITEPY="$($PYTHON -c 'import sysconfig; s = sysconfig.get_scheme_names(); m=list(set(("deb_system", "rpm_prefix")) & set(s)); print(sysconfig.get_path("platlib", m.__getitem__(0))) if m else print("/usr/lib/python3/dist-packages");')"
```
BsAtHome, line 1660: On Fedora 43, this returns:
  `/usr/lib64/python3.14/site-packages`.
The other instances, in install_script and setup_designer.in, using `import site; print([p for p in site.getsitepackages() if p.startswith('/usr/lib') and '/lib64/' not in p][0])` returns:
  `/usr/lib/python3.14/site-packages`.

lib/lib64 difference... very strange.

NTULINUX: As ugly as it may be, I can copy the logic from `m4` and put it in install_script and setup_designer.in

I was just trying to avoid that..

NTULINUX: Nevermind, you're talking about the old one, not PYTHON_SITE_PKG. I misunderstood. Currently, on fedora with my latest forced push, PYTHON_SITE_PKG and SITEPY in those files match.

BsAtHome: When I look in those two directories, then both are populated, but with different packages. I have no clue what goes into where. Maybe fedora docs can help?

NTULINUX: Yeah I noticed that too, I figured since PYTHON_SITE_PKG uses /lib instead of /lib64 to use that, plus /lib is more heavily populated.

```
fedora@fedora:~/dev/linuxcnc/src$ ls /usr/lib64/python3.14/site-packages | wc -l
120
fedora@fedora:~/dev/linuxcnc/src$ ls /usr/lib/python3.14/site-packages | wc -l
200
```

BsAtHome: There must be some documentation which distro wants what where. Better do it right than guess wrong?

NTULINUX: /usr/lib64/python3.14/site-packages/: This directory is used for architecture-specific (binary) packages.

/usr/lib/python3.14/site-packages/: This directory is used for architecture-independent (pure Python) packages. These "noarch" packages consist only of .py files and byte-compiled .pyc files, which are not tied to a specific hardware architecture.

BsAtHome: And that applies to Debian, Ubuntu, Gentoo and other distros too?

NTULINUX: Debian uses /usr/lib/python3/dist-packages for everything

Gentoo uses  /usr/lib/python3.14/site-packages for everything

Fedora... :P

NTULINUX: I just realized that this fix will break probe basic because that is hard-coded too.. I need to make minor adjustments to SITEPY in the files outside configure.ac for this to work properly 100% but then probe basic will need an update as well. I will work with Turboss on this though.

I think I found the magic line: `python3 -c "import site; print(next((p for p in site.getsitepackages() if any(x in p for x in ['site-packages', 'dist-packages']) and p.startswith('/usr/lib'))))"`

Works in Gentoo, Fedora (properly, using /usr/lib64) and Debian!

edit: But breaks on Alpine Linux... This is getting way too ugly, too fast. Scratch this commit?

grandixximo:
@NTULINUX thanks for catching the Gentoo gap. Suggest salvaging by keeping `sysconfig` as the detector and dropping only the hardcoded fallback. The original code only failed because it requested distro-specific schemes (`deb_system`, `rpm_prefix`) that Gentoo does not ship. The default scheme always works:

```sh
SITEPY="${PYTHON_SITE_PKG:-$($PYTHON -c 'import sysconfig; print(sysconfig.get_path("platlib"))')}"
```

- Debian: returns `/usr/lib/python3/dist-packages` (their patched default)
- Gentoo: returns `/usr/lib/python3.13/site-packages`
- Fedora/RPM: returns the correct versioned path
- Source/venv/prefix builds: also correct
- `PYTHON_SITE_PKG` still honored as packager override for cross-compile

This avoids the regression for Debian/Fedora packagers (who lose autodetect in the current diff) and removes the `site.getsitepackages()[0]` filter in the shell scripts, which is fragile against `/usr/local`, multilib, and list ordering.

NTULINUX:
One potential issue is that even with your SITEPY, Fedora returns:

checking for site-package location... /usr/lib/python3.14/site-packages

Shouldn't these be /usr/lib64 instead, since linuxcnc is not a `noarch` package?

To be fair, I don't care so much that it's maybe not in the absolute best spot, but even on a 64-bit system, at least /usr/lib/python3.14/site-packages is in a path that python searches for. I'll take it, honestly.

grandixximo:
`platlib` is the correct key for arch-dependent packages (which linuxcnc is, agreed). On stock Fedora it should already return lib64 because their python is built with `platlibdir=lib64` and the default `posix_prefix` scheme expands `{platlibdir}`.

Can you share the output of:

```sh
python3 -c 'import sysconfig; print(sysconfig.get_path("platlib")); print(sysconfig.get_config_var("platlibdir"))'
```

on the Fedora box where you saw `/usr/lib/python3.14/site-packages`? If `platlibdir=lib64` but the path drops it, that is a Fedora python 3.14 packaging issue worth reporting upstream. If `platlibdir=lib`, the interpreter was not built with multilib and lib is correct for that install.

Either way, `platlib` is still the right semantic choice over `purelib`. The fallback can stay as `PYTHON_SITE_PKG` for the rare case where sysconfig disagrees with the distro layout.

NTULINUX:
Output:

```
/usr/local/lib64/python3.14/site-packages
lib64
```

grandixximo:
Good catch, you are right. The `/usr/local` prefix is Fedora patching the default scheme to point at `/usr/local` so user installs do not collide with system RPMs. `rpm_prefix` exists specifically to force `/usr` during RPM builds, which is why the original detection code used it. My earlier snippet was wrong for Fedora.

The actual bug is only the hardcoded `/usr/lib/python3/dist-packages` fallback for distros without a custom scheme. Drop just that, fall back to the default scheme:

```sh
SITEPY="${PYTHON_SITE_PKG:-$($PYTHON -c '
import sysconfig
schemes = sysconfig.get_scheme_names()
for s in ("deb_system", "rpm_prefix"):
    if s in schemes:
        print(sysconfig.get_path("platlib", s))
        break
else:
    print(sysconfig.get_path("platlib"))
')}"
```

- Debian: `deb_system` returns `/usr/lib/python3/dist-packages`
- Fedora: `rpm_prefix` returns `/usr/lib64/python3.14/site-packages`
- Gentoo: default scheme returns `/usr/lib/python3.13/site-packages` (prefix=/usr, no /usr/local indirection)
- `PYTHON_SITE_PKG` still honored as packager override

NTULINUX:
Fedora with your latest snippet:

`checking for site-package location... /usr/lib/python3.14/site-packages`

I think the rpm_prefix got messed up.

What does work on Fedora with it using /usr/lib64 is:

`SITEPY="${PYTHON_PLATFORM_SITE_PKG}"` But this will break other distros as-is.

grandixximo:
Two things going on.

**1. Can you confirm Fedora 14's scheme list?**

```sh
python3 -c 'import sysconfig; print(sysconfig.get_scheme_names())'
python3 -c 'import sysconfig; print(sysconfig.get_path("platlib", "rpm_prefix"))'
```

You saw `/usr/lib/python3.14/...` (no `/usr/local`, no `lib64`), which suggests `rpm_prefix` is still matched but now returns the wrong path. The scheme list will pin it down.

**2. `PYTHON_PLATFORM_SITE_PKG` is the right answer for Fedora RPM builds.** That env var is set by Fedora's RPM macros during `rpmbuild` to `%{python3_sitearch}` (= `/usr/lib64/python3.14/site-packages`). Outside `rpmbuild` it is unset and sysconfig autodetect is what you want.

A layered approach covers all cases without breaking other distros:

```sh
SITEPY="${PYTHON_PLATFORM_SITE_PKG:-${PYTHON_SITE_PKG:-$($PYTHON -c 'import sysconfig; print(sysconfig.get_path("platlib"))')}}"
```

- Fedora RPM build: `PYTHON_PLATFORM_SITE_PKG` set by spec, wins.
- Generic packager override: `PYTHON_SITE_PKG`.
- Debian: default `platlib` already returns `/usr/lib/python3/dist-packages` (Debian patches the default scheme).
- Gentoo: default returns `/usr/lib/python3.X/site-packages`.
- Fedora user install (no rpmbuild): default returns `/usr/local/lib64/...`, correct for that context.

This drops the `rpm_prefix`/`deb_system` scheme dance entirely. Each distro that needs a non-default path injects it via env var at build time.

NTULINUX:
```
fedora@fedora:~/dev/linuxcnc/src$ python3 -c 'import sysconfig; print(sysconfig.get_scheme_names())'
('nt', 'nt_user', 'nt_venv', 'osx_framework_user', 'posix_home', 'posix_prefix', 'posix_user', 'posix_venv', 'rpm_prefix', 'venv')

fedora@fedora:~/dev/linuxcnc/src$ python3 -c 'import sysconfig; print(sysconfig.get_path("platlib", "rpm_prefix"))'
/usr/lib64/python3.14/site-packages
```

grandixximo:
A few things from digging into this:

The `rpm_prefix` scheme is supposed to return `/usr/lib64/...` per Fedora's documented behavior since F36. The fact that it returns `/usr/lib/python3.14/site-packages` on your box looks like a Fedora 3.14 packaging regression. Could you check:

```sh
python3 -c 'import sysconfig; print(sysconfig.get_paths("rpm_prefix"))'
```

and report it upstream to fedora-python if `platlib` there is wrong?

`PYTHON_PLATFORM_SITE_PKG` is not a standard autoconf variable upstream. Where is it being defined in your build environment? Fedora RPM macro export? Local m4 include?

The GNU-standard answer for this is automake's `AM_PATH_PYTHON`, which exports `pyexecdir` (= platlib, arch-dependent, lib64 on multilib) and `pythondir` (= purelib). For an arch-dependent package like linuxcnc, `pyexecdir` is the correct variable. That avoids the scheme-name fight entirely and respects distro patches.

References:
- [Fedora 36 install scheme writeup](https://hackmd.io/@python-maint/BkqScKJW5)
- [Automake AM_PATH_PYTHON](https://www.gnu.org/software/automake/manual/html_node/Python.html)
- [sysconfig docs](https://docs.python.org/3/library/sysconfig.html)

NTULINUX:
```
fedora@fedora:~/dev/linuxcnc/src$ python3 -c 'import sysconfig; print(sysconfig.get_paths("rpm_prefix"))'
{'stdlib': '/usr/lib64/python3.14', 'platstdlib': '/usr/lib64/python3.14', 'purelib': '/usr/lib/python3.14/site-packages', 'platlib': '/usr/lib64/python3.14/site-packages', 'include': '/usr/include/python3.14', 'platinclude': '/usr/include/python3.14', 'scripts': '/usr/bin', 'data': '/usr'}
```

PYTHON_PLATFORM_SITE_PKG comes from:

https://github.com/LinuxCNC/linuxcnc/blob/master/src/m4/ax_python_devel.m4#L369

grandixximo:
Good, that pins it. `rpm_prefix` works correctly in isolation, so the multi-line for/else snippet is getting mangled by m4 in `configure.ac` (autoconf eats `("a", "b")` tuples and indentation).

Two ways out:

**A)** Single-statement form, m4-safe:

```m4
SITEPY=$($PYTHON -c "import sysconfig; s=sysconfig.get_scheme_names(); print(sysconfig.get_path('platlib', 'deb_system' if 'deb_system' in s else 'rpm_prefix' if 'rpm_prefix' in s else sysconfig.get_default_scheme()))")
```

**B)** Use `AM_PATH_PYTHON` and reference `$pyexecdir` (= platlib, lib64-aware). Upstream-blessed for arch-dependent installs and sidesteps the scheme question entirely.

Could you try (A) to confirm the diagnosis? If it returns `/usr/lib64/python3.14/site-packages` on Fedora and `/usr/lib/python3/dist-packages` on Debian, the m4-mangling theory is right and the PR can land with that one line.

grandixximo:
You are right, I went on a wild goose chase. `ax_python_devel.m4` already handles this correctly:

- Detects `posix_local` (Debian) and swaps to `posix_prefix` to avoid `/usr/local`
- Calls `sysconfig.get_path('platlib', scheme, vars={'platbase': prefix})` with the actual configure `--prefix`

With `--prefix=/usr`:
- Fedora: `/usr/lib64/python3.14/site-packages`
- Debian: `/usr/lib/python3/dist-packages`
- Gentoo: `/usr/lib/python3.X/site-packages`

The only real bug in the PR is using `PYTHON_SITE_PKG` (purelib, lib) instead of `PYTHON_PLATFORM_SITE_PKG` (platlib, lib64). linuxcnc has compiled C extensions so platlib is correct.

Three-letter salvage:

```diff
-SITEPY="$PYTHON_SITE_PKG"
+SITEPY="$PYTHON_PLATFORM_SITE_PKG"
```

That should resolve Gentoo without breaking Fedora/Debian. Apologies for the noise on the scheme detection digression.

grandixximo (commented):
Should fix debian fedora and gentoo installs

NTULINUX:
@BsAtHome We figured this out on IRC, the latest force push is good to go!

Issue #3915, opened 2026-04-09
[FATAL] /usr/lib/libuspace-posix.so.0 is hard-coded now?


NTULINUX:
Some commit in the past day or two just broke LinuxCNC:

```
LINUXCNC - 2.10.0~pre1
Machine configuration directory is '/home/lcnc/$/usr.share.doc.linuxcnc.examples.sample-configs.sim.axis'
Machine configuration file is 'axis.ini'
Starting LinuxCNC...
linuxcnc TPMOD=tpmod HOMEMOD=homemod EMCMOT=motmod
dlopen: /usr/lib/libuspace-posix.so.0: cannot open shared object file: No such file or directory
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Could not load rtapi dll
<commandline>:0: /usr/bin/rtapi_app exited without becoming ready
<commandline>:0: insmod for tpmod failed, returned -1
dlopen: /usr/lib/libuspace-posix.so.0: cannot open shared object file: No such file or directory
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Could not load rtapi dll
<commandline>:0: /usr/bin/rtapi_app exited without becoming ready
<commandline>:0: insmod for homemod failed, returned -1
Found file(lib): /usr/share/linuxcnc/hallib/core_sim.hal
dlopen: /usr/lib/libuspace-posix.so.0: cannot open shared object file: No such file or directory
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Could not load rtapi dll
/usr/share/linuxcnc/hallib/core_sim.hal:5: /usr/bin/rtapi_app exited without becoming ready
/usr/share/linuxcnc/hallib/core_sim.hal:5: insmod for trivkins failed, returned -1
Shutting down and cleaning up LinuxCNC...
USRMOT: ERROR: command 32 timeout (seq: 1)
emcMotionInit: emcTrajInit failed
```

```
lcnc@localhost ~ $ find /usr -name "libuspace-posix.so.0" 2> /dev/null
/usr/lib64/libuspace-posix.so.0
```

NTULINUX:
`git log -S "libuspace-posix.so.0"` points to:

commit 913c7ad31b9ffc92b4209f874e032efb19fb2320
Date:   Mon Apr 6 16:53:08 2026 +0200

    Cleanup: Renaming

commit 62e5ea6457af35ef4d7380168b962528f348a5f6
[...]

PR #3917, opened 2026-04-09
WIP: Fix: Rtapi cleanup: Library not found after install

hdiethelm:
This might fix the bug https://github.com/LinuxCNC/linuxcnc/issues/3915 introduced in https://github.com/LinuxCNC/linuxcnc/pull/3908

According to the manual page of dlopen(), this variant should search the library using ld.so.

~Caviat: I was not able to reproduce the issue until now. I am working on it. For me, both variant's work but this one is nicer anyway.~

Also https://github.com/LinuxCNC/linuxcnc/pull/3908#issuecomment-4212735732 is still open. Depending on the urgency, you can merge this already if it solves the bug.

@NTULINUX: Can you check if this solves the issue for you?

### DIFF
--- a/src/rtapi/uspace_rtapi_main.cc
+++ b/src/rtapi/uspace_rtapi_main.cc
@@ -769,17 +769,17 @@ static RtapiApp *makeDllApp(std::string dllName, int policy) {
 static RtapiApp *makeApp() {
     RtapiApp *app;
     if (euid != 0 || harden_rt() < 0) {
-        app = makeDllApp(EMC2_HOME "/lib/libuspace-posix.so.0", SCHED_OTHER);
+        app = makeDllApp("libuspace-posix.so.0", SCHED_OTHER);
     } else {
         WithRoot r;
         if (detect_xenomai_evl()) {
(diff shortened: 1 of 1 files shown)


hdiethelm:
So, I was able to reproduce the issue. The debian package is built withouth `--libdir=/usr/lib64`.
Using:
```
./autogen.sh
./configure --with-realtime=uspace --prefix=/usr --sysconfdir=/etc --libdir=/usr/lib64
make -j16
sudo make install
```

Shows on master:
```
latency-test 
dlopen: /usr/lib/libuspace-posix.so.0: cannot open shared object file: No such file or directory
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Could not load rtapi dll
lat.hal:1: /usr/bin/rtapi_app exited without becoming ready
lat.hal:1: insmod for threads failed, returned -1
```
And after this change:
```
latency-test 
Note: Using POSIX realtime
Unexpected realtime delay on task 0 with period 25000
This Message will only display once per session.
Run the Latency Test and resolve before continuing.
```

linuxcnc with axis_mm works also fine.

So it's fixed at least with the test above.

hdiethelm:
So was for nothing I guess...

PR #3918, opened 2026-04-09
Revert "Rtapi cleanup"

andypugh:
Reverts LinuxCNC/linuxcnc#3908


hdiethelm:
Shure that's needed? Already fixed in https://github.com/LinuxCNC/linuxcnc/pull/3917

hdiethelm:
Anyway, already merged, so I call it a day and look into it a later or also not...

andypugh:
I think we still want it, but as a single commit that actually works properly. 
Assuming that you have a "good" branch, make a new PR.

hdiethelm:
How do you mean single commit? Just on top of my branch + fix or all commits squashed to one commit?

BsAtHome:
I also agree that the rtapi RTAI/Xenomai/EVL cleanup and work done is valuable and should go into the tree. But we have apparently failed at review and have not spotted any problems in the original PR before it was merged.

BTW, @andypugh, you were a bit fast to merge, I guess. There were still some in-progress analyses on the proposed changes (like the mutex, which I mentioned to look into in https://github.com/LinuxCNC/linuxcnc/pull/3908#issuecomment-4194831493).

Never mind past problems, @hdiethelm, if you care to make a new PR, then we all together should get it reviewed and tested for any problems and looks.

hdiethelm:
Ok, fine for me. I was a bit surprised that it was merged that fast and then confused that it was reverted while my fix was already there. But no big deal.

Issue #3191, opened 2024-11-26
ld.lld: error: version script assignment of 'global' to symbol 'rtapi_app_exit' failed: symbol not defined


NTULINUX:
# The issue tracker is not a support forum

The LinuxCNC issue tracker is to report bugs in the software.
If you have a question about how to use the software, use one of the other methods detailed on our community support page: http://linuxcnc.org/community/

(delete this section before submitting your bug report)

## Here are the steps I follow to reproduce the issue:

 1. Build with Clang/LLD 19
 2. Wait

## This is what I expected to happen:

A successful compilation.

## This is what happened instead:

ld.lld: error: version script assignment of 'global' to symbol 'rtapi_app_exit' failed: symbol not defined

## It worked properly before this:

LLD recently changed their behavior to: --no-undefined-version by default:

https://reviews.llvm.org/D135402

Passing:

```LDFLAGS="-Wl,--undefined-version" make``` works around the problem, however there should be a more proper fix.

## Information about my hardware and software:

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`): 
 * I am using this kernel version (shown by `uname -a`): Linux ntu-gentoo 6.12.1-BUILD SMP Fri Nov 22 18:35:05 CST 2024 x86_64 AMD Ryzen 7 2700X Eight-Core Processor AuthenticAMD GNU/Linux
 * I am running ...
   * [ ] A binary version from linuxcnc.org (including buildbot.linuxcnc.org)
   * [X] A binary I built myself
   * [ ] A binary version from some other source besides linuxcnc.org
 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`): Latest 2.9 or master checkout

rmu75:
I'm unsure what this error is trying to say?

NTULINUX:
Can confirm, https://github.com/LinuxCNC/linuxcnc/pull/3925 fixes this issue. Axis now loads as well.

PR #3919, opened 2026-04-09
Rtapi cleanup v2

hdiethelm:
Continuation of https://github.com/LinuxCNC/linuxcnc/pull/3908 reverted in https://github.com/LinuxCNC/linuxcnc/pull/3918

Target: Move some classes out of the huge uspace_rtapi_app.cc

uspace_rtapi_main: Contains the main function and helpers
uspace_rtapi_app: Contains the RtapiApp class
uspace_posix: Contains the PosixApp class

Other fixes:
- Don't start master just for exit: https://github.com/LinuxCNC/linuxcnc/pull/3908/changes/371793cdd9187a8872c9ed1dea6e76ea1d6ac59c
- Remove unused rtapi_task::ratio: https://github.com/LinuxCNC/linuxcnc/pull/3908/changes/309757829b986bc60817f646c8813d124f0ea01d
  - Set but only read to force a fixed ratio which probably was not the intention
- Slightly different lock to avoid needing reference to instance: https://github.com/LinuxCNC/linuxcnc/pull/3908/changes/c8af827762238ce5c85c83c34bdc6d0c22f97b17 
- All real time implementations are now library's and handled the same way: https://github.com/LinuxCNC/linuxcnc/commit/62e5ea6457af35ef4d7380168b962528f348a5f6
 - No hard-coded library paths https://github.com/LinuxCNC/linuxcnc/commit/e1bbaf4f57166b093e0abe739d9b48a8f7de575c
 - ... and many more that came up during review

### DIFF
--- a/src/rtapi/uspace_rtapi_app.cc
+++ b/src/rtapi/uspace_rtapi_app.cc
@@ -48,6 +48,8 @@ rtapi_task::rtapi_task()
 {
 }
 
+struct rtapi_task *RtapiApp::task_array[MAX_TASKS];
+
 /* Priority functions.  Uspace uses POSIX task priorities. */
 
 int RtapiApp::prio_highest() const {
--- a/src/rtapi/uspace_rtapi_app.cc
+++ b/src/rtapi/uspace_rtapi_app.cc
@@ -23,7 +23,7 @@
 #include <string.h>
 
 std::atomic_int WithRoot::level;
-uid_t euid, ruid;
+uid_t WithRoot::ruid, WithRoot::euid;
 
 WithRoot::WithRoot() {
     if (!level++) {
@@ -39,6 +39,11 @@ WithRoot::~WithRoot() {
         setfsuid(ruid);
 #endif
     }
+}
+
+void WithRoot::init(uid_t ruid_ini, uid_t euid_ini){
+    ruid=ruid_ini;
+    euid=euid_ini;
 }
 
 rtapi_task::rtapi_task()
--- a/src/rtapi/uspace_rtapi_app.cc
+++ b/src/rtapi/uspace_rtapi_app.cc
@@ -41,9 +41,9 @@ WithRoot::~WithRoot() {
     }
 }
 
-void WithRoot::init(uid_t ruid_ini, uid_t euid_ini){
-    ruid=ruid_ini;
-    euid=euid_ini;
+void WithRoot::init(uid_t ruid_ini, uid_t euid_ini) {
+    ruid = ruid_ini;
+    euid = euid_ini;
 }
 
 rtapi_task::rtapi_task()
--- a/src/rtapi/uspace_rtapi_app.cc
+++ b/src/rtapi/uspace_rtapi_app.cc
@@ -234,7 +234,7 @@ out:
     fclose(fp);
 }
 
-int find_rt_cpu_number() {
+int RtapiApp::find_rt_cpu_number() {
     if (getenv("RTAPI_CPU_NUMBER"))
         return atoi(getenv("RTAPI_CPU_NUMBER"));
 
@@ -269,7 +269,7 @@ int find_rt_cpu_number() {
 #endif
 }
 
-void set_namef(const char *fmt, ...) {
+void RtapiApp::set_namef(const char *fmt, ...) {
     char *buf = NULL;
     va_list ap;
 
--- a/src/rtapi/uspace_rtapi_app.cc
+++ b/src/rtapi/uspace_rtapi_app.cc
@@ -15,56 +15,18 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-#include "config.h"
-#include <linuxcnc.h>
-
-#ifdef __linux__
-#include <sys/fsuid.h>
-#endif
-#include <sys/types.h>
-#include <sys/socket.h>
-#include <sys/un.h>
-#include <sys/stat.h>
-#include <fcntl.h>
-#include <unistd.h>
-#include <errno.h>
-#include <dlfcn.h>
-#include <signal.h>
-#include <vector>
-#include <string>
-#include <map>
-#include <algorithm>
-#include <sys/time.h>
-#include <time.h>
-#include <stdlib.h>
-#include <stdio.h>
-#ifdef HAVE_SYS_IO_H
-#include <sys/io.h>
-#endif
-#include <sys/resource.h>
-#include <sys/mman.h>
-#ifdef __linux__
-#include <malloc.h>
-#include <sys/prctl.h>
-#endif
-#ifdef __FreeBSD__
-#include <pthread_np.h>
-#endif
-
-#include <boost/lockfree/queue.hpp>
-
 #include "rtapi.h"
-#include <hal.h>
-#include "hal/hal_priv.h"
-#include "rtapi_uspace.hh"
+#include "uspace_rtapi_app.hh"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
 
 std::atomic_int WithRoot::level;
-static uid_t euid, ruid;
-
-#include "rtapi/uspace_common.h"
+uid_t euid, ruid;
 
 WithRoot::WithRoot() {
-    if(!level++) {
+    if (!level++) {
 #ifdef __linux__
         setfsuid(euid);
 #endif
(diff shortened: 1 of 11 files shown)


hdiethelm:
Done:
- libuspace-xxx.so.0 -> liblinuxcnc-uspace-xxx.so.0
  - Longest name is liblinuxcnc-uspace-xenomai-evl.so.0

Still open:
- Check if slightly different lock to avoid needing reference to instance: https://github.com/LinuxCNC/linuxcnc/commit/c8af827762238ce5c85c83c34bdc6d0c22f97b17 is a good idea
- Review

And
> In the Submakefile it reads:
> 
> ```makefile
> $(call TOOBJSDEPS, $(USPACE_POSIX_SRCS)): EXTRAFLAGS += -pthread -fPIC
> ```
> 
> Shouldn't that be (not adding to EXTRAFLAGS):
> 
> ```makefile
> $(call TOOBJSDEPS, $(USPACE_POSIX_SRCS)): EXTRAFLAGS = -pthread -fPIC
> ```

Hmm, I don't understand makefiles in depth. I just copied what was already there a few lines below and edited it to match my lib. As much as I understand this, it just adds this flags to the global EXTRAFLAGS in Makefile for one compile command only. I did not see any duplicated flags.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,946 @@
...
+        if (r != 4) {
+            rtapi_print_msg(RTAPI_MSG_WARN, "failed to write to /dev/cpu_dma_latency: %s\n", strerror(errno));
+        }
+        // deliberately leak fd until program exit
+    }
+#endif /* __linux__ */
+    return 0;
+}
+
+static RtapiApp *makeDllApp(std::string dllName, int policy) {
```
BsAtHome: Argument `std::string` should be `const std::string &`

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,946 @@
...
+
+#define MAX_MODULES 64
+#define MODULE_OFFSET 32768
+
+static void signal_handler(int sig, siginfo_t * /*si*/, void * /*uctx*/) {
+    switch (sig) {
+    case SIGXCPU:
+        // should not happen - must be handled in RTAPI if enabled
+        rtapi_print_msg(RTAPI_MSG_ERR, "rtapi_app: BUG: SIGXCPU received - exiting\n");
+        exit(0);
```
BsAtHome: You cannot call `exit()` in a signal handler. You must call `_exit()`

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,946 @@
...
+        break;
+
+    case SIGTERM:
+        rtapi_print_msg(RTAPI_MSG_ERR, "rtapi_app: SIGTERM - shutting down\n");
+        exit(0);
```
BsAtHome: `exit()` -> `_exit()`

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,946 @@
...
+    default: // pretty bad
+        rtapi_print_msg(RTAPI_MSG_ERR, "rtapi_app: caught signal %d - dumping core\n", sig);
+        sleep(1); // let syslog drain
+        signal(sig, SIG_DFL);
+        // for reasons unknown raise(sig); doesn't lead to core dump file
+        // but this will
+        kill(getpid(), sig);
+        break;
+    }
+    exit(1);
```
BsAtHome: `exit()` -> `_exit()`

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,946 @@
...
+        break;
+
+    case SIGTERM:
+        rtapi_print_msg(RTAPI_MSG_ERR, "rtapi_app: SIGTERM - shutting down\n");
+        exit(0);
+        break;
+
+    default: // pretty bad
+        rtapi_print_msg(RTAPI_MSG_ERR, "rtapi_app: caught signal %d - dumping core\n", sig);
+        sleep(1); // let syslog drain
```
BsAtHome: calling `sleep()` in a signal handler may be problematic because it can be implemented using SIGALRM. Signals in signal is a very bad concept.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,946 @@
...
+    setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, &enable, sizeof(enable));
+    struct sockaddr_un addr;
+    memset(&addr, 0x0, sizeof(addr));
+    addr.sun_family = AF_UNIX;
+    if ((len = get_fifo_path(addr.sun_path, sizeof(addr.sun_path))) < 0)
+        exit(1);
+
+    // plus one because we use the abstract namespace, it will show up in
+    // /proc/net/unix prefixed with an @
+    int result = ::bind(fd, (sockaddr *)&addr, len + sizeof(addr.sun_family) + 1);
```
BsAtHome: The size of the sockadd_un is defined by `sizeof(addr)` (effectively `sizeof(struct sockaddr_un)`) and not by some math.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,946 @@
...
+
+    if (result == 0) {
+        //If called in master mode with exit command, no need to start master
+        //and exit again
+        if (args.size() == 1 && args[0] == "exit") {
+            return 0;
+        }
```
BsAtHome, line 614: If you exit here, then why perform the socket() and bind() calls?

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,946 @@
...
+        struct timeval t0, t1;
+        gettimeofday(&t0, NULL);
+        gettimeofday(&t1, NULL);
+        for (int i = 0; i < 3 || (t1.tv_sec < 3 + t0.tv_sec); i++) {
+            result = connect(fd, (sockaddr *)&addr, len + sizeof(addr.sun_family) + 1);
+            if (result == 0)
+                break;
+            if (i == 0)
+                srand48(t0.tv_sec ^ t0.tv_usec);
+            usleep(lrand48() % 100000);
```
BsAtHome: Can there be unforeseen interaction with SIGALRM here? I don't think you want to involve signals. Also, wouldn't there be a _minimum_ wait wanted here? A random wait can return close to zero all the time and you only loop three times.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,946 @@
...
+        s = std::string(getenv("HOME")) + "/.rtapi_fifo";
+    } else {
+        rtapi_print_msg(
+            RTAPI_MSG_ERR, "rtapi_app: RTAPI_FIFO_PATH and HOME are unset.  rtapi fifo creation is unsafe.\n"
+        );
+    }
+    return s;
+}
+
+static int get_fifo_path(char *buf, size_t bufsize) {
```
BsAtHome: Why use char* and size_t? Why not have a std::string& as argument?
And why are there two get_fifo_path() functions with different arguments? Can't they be replaced by one?

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,946 @@
...
+            goto out;
+        if (force_exit || instance_count == 0)
+            goto out;
+    }
+    sim_rtapi_run_threads(fd, callback);
+out:
+    pthread_cancel(queue_thread);
+    pthread_join(queue_thread, nullptr);
+    rtapi_msg_queue.consume_all([](const message_t &m) {
+        fputs(m.msg, m.level == RTAPI_MSG_ALL ? stdout : stderr);
```
BsAtHome, line 509: Why fputs()?

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,946 @@
...
+
+static void write_strings(int fd, const std::vector<std::string> &strings) {
+    std::string buf;
+    write_number(buf, strings.size());
+    for (unsigned int i = 0; i < strings.size(); i++) {
+        write_string(buf, strings[i]);
+    }
+    if (write(fd, buf.data(), buf.size()) != (ssize_t)buf.size())
+        throw WriteError();
+}
```
BsAtHome: This is a socket communication we work with. Why isn't this passed through and handled by the standard sendmsg() and recvmsg() system calls? Wouldn't they be much better at handling this? Or is there and advantage to all this code?

NTULINUX:
Works here on Gentoo, rip, system install, clang and gcc builds, whole shebang!

edit: Have not yet tested with RTAI.

hdiethelm:
Thanks for testing!

I renamed the library's to:
```
liblinuxcnc-uspace-posix.so.0 liblinuxcnc-uspace-xenomai-evl.so.0
liblinuxcnc-uspace-rtai.so.0    liblinuxcnc-uspace-xenomai.so.0
```
A bit long names, but I think it is fine. But i'm open for other suggestions.

Additionally, I reduced the globals.

I tested all 5 configurations in a VM and they all work. There are two issues but I don't think they are due to this PR:

 - LXRT doesn't start on an isolated CPU. Workaround: RTAPI_CPU_NUMBER=1 linuxcnc
 - RTAI doesn't unload the modules in the correct order and fails to unload all

hdiethelm:
@NTULINUX: Do you have a test setup you can share or is it all just manually setup? A series of docker files would be nice, so different OS can be tested if something starts. I messed my VM up slightly by using `make install` and just after the fact figured out that there is no `make uninstall` but I was able to restore it.
If there is nothing yet, I can create some, should not be to involved. I often use podman for similar things where I don't want to mess up my host.  Podman is similar to docker but it doesn't need root.

NTULINUX:
I have VM images up right now but they're in flux. I'm going to post new VMs with all the right fixes in a bit. Currently tracking this PR here:

https://github.com/LinuxCNC/linuxcnc/pull/3925

My ebuilds at the moment are broken but we're in the middle of sorting it all out for good. Will share link to new VM soon with everything tied together, cleanly.

andypugh:
Some discussion in the Sunday video meet-up has suggested that we should look at incorporating this into the rtapi cleanup:

https://github.com/LinuxCNC/linuxcnc/pull/918

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Good catch, done

src/rtapi/uspace_rtapi_main.cc
BsAtHome: Also,using printf or puts in signal handlers is not allowed.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: This code is a bit funny: Bind is used to detect if a master is running. It returns 0 if no master is already running, so a new one is started.

Now I had the case when exiting, everything was initialized again due to master exits automatically as soon as there is no instance any more: Line 417.

Due to that, master was started again just to call exit and "Note: Using POSIX realtime" was shown a second time before exiting when I closed latency test.

BsAtHome: But isn't that is just as race prone as any other method?

The concept "the first becomes master" is flawed when you try to exit.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: The first function gets the path as std::string. The second copies it to a char pointer and checks the size.
That was the original implementation, I found that somewhat convoluted and the return NULL was also not good to read, so I reordered it a bit without changing to much:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/rtapi/uspace_rtapi_app.cc#L503

BsAtHome: But the whole cascade of get_fifo_path feels wrong IMO. We should get rid of char buffers as much as we can (everywhere in the code).

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,944 @@
...
+        rtapi_print_msg(
+            RTAPI_MSG_ERR,
+            "rtapi_app: rtapi fifo path is too long (arch limit %zd): %s\n",
+            sizeof(sockaddr_un::sun_path),
+            s.c_str()
+        );
+        return -1;
+    }
+    len = snprintf(buf + 1, bufsize - 1, "%s", s.c_str());
+    return len;
```
BsAtHome: And then we're back to snprintf(). fmt::format would fit better. Length check can be done where it is relevant.

hdiethelm:
I commented and corrected the things I have changed.
Now do we really want do do all of that in this MR?

I always prefer to have refactoring (moving code around) and bug fixing/functional changes as separated as possible. That makes review and testing easier, you just check that the moved code arrived as it was and so avoids having merge conflicts due to the branch staying open for to long.

But I have to admit, I also did some changes that I just was not able to leave it as it was and which simplified moving code due to removed dependency's and of course created a bug doing so.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Do you have a reference why? I now understand the difference between the two but I did not find anything why _exit() is needed in signal handlers.

BsAtHome:
See: https://www.man7.org/linux/man-pages/man7/signal-safety.7.html for the callyou are allowed to make in signal handlers.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: That's clearly a bug in the existing code, fixed two times, also on line 559.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Thanks for the reference. Corrected.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Corrected

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Corrected

BsAtHome:
> I commented and corrected the things I have changed. Now do we really want do do all of that in this MR?

Well, that is a good question. You are doing "cleanup" and that would imply refactor and fixes, IMO.

> I always prefer to have refactoring (moving code around) and bug fixing/functional changes as separated as possible. That makes review and testing easier, you just check that the moved code arrived as it was and so avoids having merge conflicts due to the branch staying open for to long.

That is a possibility too, but there are soooo (key got stuck) many problems with this code that it is a good question why not hit those <s>two birds with one stone</s> two lines with one keypress.

> But I have to admit, I also did some changes that I just was not able to leave it as it was and which simplified moving code due to removed dependency's and of course created a bug doing so.

And I appreciate the changes! They are very necessary. Before we're done it needs to be tested,of course. But when the code gets better structured, then that should also become easier, I hope.

But, if you want to split it, then I think we need to have a very clear split where you move without actual changes and then refactor. My opinion is that moving code also classifies as refactoring and therefore it would be a missed opportunity if not fixed in one go.

BsAtHome:
BTW, I've been wanting to vacuum this code for a long time but have not gotten arround to that point yet. You know, INI-file reader just done, HAL types/access revamp in the pipeline, tcl9 stuff that still needs fixing, build system cleanup, and so on.

(and also got CI's -Werror merged)

hdiethelm:
> > I commented and corrected the things I have changed. Now do we really want do do all of that in this MR?
> 
> Well, that is a good question. You are doing "cleanup" and that would imply refactor and fixes, IMO.
> 
Might be I did not specify my (initial) intentions well enough. I just found the rtapi_uspace hard to manage while implementing xenomai and wanted to split it up without changing the existing code if not needed. But that diverged anyway already a bit.

With the new structure, it should also be easier to implement rtapi_udp_sendto() or something similar needed for RTNet.

> > I always prefer to have refactoring (moving code around) and bug fixing/functional changes as separated as possible. That makes review and testing easier, you just check that the moved code arrived as it was and so avoids having merge conflicts due to the branch staying open for to long.
> 
> That is a possibility too, but there are soooo (key got stuck) many problems with this code that it is a good question why not hit those two birds with one stone two lines with one keypress.
> 
> > But I have to admit, I also did some changes that I just was not able to leave it as it was and which simplified moving code due to removed dependency's and of course created a bug doing so.
> 
> And I appreciate the changes! They are very necessary. Before we're done it needs to be tested,of course. But when the code gets better structured, then that should also become easier, I hope.
> 
> But, if you want to split it, then I think we need to have a very clear split where you move without actual changes and then refactor. My opinion is that moving code also classifies as refactoring and therefore it would be a missed opportunity if not fixed in one go.

I have to look into it a bit more next week if I find time. But from my point of view it would make sense to split the more involved changes in a new PR or also two.

But there is also always some testing effort behind which I don't know how much you have to do from your side.

Also there is: https://github.com/LinuxCNC/linuxcnc/pull/3925
Just randomly, I found also that DLSYM halpr_find_comp_by_name https://github.com/hdiethelm/linuxcnc-fork/blob/rtapi_cleanup_v2/src/rtapi/uspace_rtapi_main.cc#L112 does not match the signature (char* vs const char*), 

I was now wondering if there is not a better solution for this. Often I use the following pattern to define the function and the matching type in the same header and then from then on only FUN_T* for function pointers. So you naturally change both. By some define magic, it should be possible do do both in one, but then it gets hard to read.
```
typedef int FUN_T(void *, void *);
int fun(void *, void *);
```
You know of a good way to test if both definitions match?

It would also be nicer to check at compile time if a .so has the needed main/exit functions.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: I think it doesn't make a difference. send() / recv() could improve the readability.  sendmsg() / recvmsg() is a bit annoying to use for this application due to you have first to fill the struct. What I found so far:
https://www.man7.org/linux/man-pages/man2/send.2.html
https://www.man7.org/linux/man-pages/man2/recv.2.html
`The only difference between send() and [write(2)](https://www.man7.org/linux/man-pages/man2/write.2.html) is the presence of flags. With a zero flags argument, send() is equivalent to [write(2)](https://www.man7.org/linux/man-pages/man2/write.2.html).`

src/rtapi/uspace_rtapi_main.cc
hdiethelm: It writes out all pending messages before exiting.
Same on line 81 where the normal message writer task function is. It's a bit risky when the 0 termination is missing but otherwise fine. Any sugestions?
 `m.level == RTAPI_MSG_ALL ? stdout : stderr` sends to stdout if the message is from rtapi_print() -> RTAPI_MSG_ALL, otherwise stderr. Looks also fine.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: You are right, it feels wrong. I tried to improve it by renaming the second to get_fifo_path_to_addr() and do what it says.

src/rtapi/uspace_rtapi_main.cc
BsAtHome: As long as it isn't in a signal handler it should be fine. Just wondering.

However, now you mention rtapi_print (and friends), there are rtapi_print_msg in the signal handler. That may pose a real problem. The default handler prints using stdio and that is a nono in a signal handler.

src/rtapi/uspace_rtapi_main.cc
BsAtHome: Just thought that you could use scatter/gather I/O on that you were writing instead of manual serializing.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Which kind or race do you have in mind? I think `bind` is probably atomic, I would have to research it.

The issue I fixed is just: "Don't start a master if none is running just to immediately close it due to the command was exit"

The only that came to my mind is if the there are multiple clients at the same time, one sends exit while the other sends a loadrt for example. In that case, it is random if the master exits and then starts again to execute loadrt or the other way around, loadrt and then exit.

But this "If there is no master, I will become master" feels a bit wrong in general. Better would be: "A master is started at start of linuxcnc and closed at the end" but this is to much for this PR i think and needs probably changes in other places to.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Do you have a good solution for this? I don't really understand what the idea behind this code is.

BsAtHome: The whole "exit" procedure done in a signal handler is faulty and wrong. You normally run your main handler in a select/poll loop where one of the input descriptors is the read-end of a pipe. You write (one byte) into this pipe from the signal handler and you know what to do outside the signal handler in the main handler loop. Then you can use a simple switch outside the signal handler to determine what to do.

The code "tries" to do a flush, but does it in the wrong place. Anything called in there is bound to be wrong or not guaranteed to work as you need it to. BTW, dumping core should not be preceded by printing a message because all you know is that the printing of a message cause the signal (SEGV for example) and everything is corrupt. Dumping code needs to be just that, dump core.

src/rtapi/uspace_rtapi_main.cc
BsAtHome: By the time you determine that there is no master, someone else can have become master. By the time you determine that there is a master, it may have exited before sending your command.

But you are right, this needs to be rethought and redone from the ground up. Lets leave it for now.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: That one looks funny. Signal will break it, the for loop is only 0.3s (100000 != 1000000) and gettimeofday can fail if you change the system time. Improved.

I did not find a function for diff of timespec, so I added one above main.

hdiethelm: BTW: That one could also be simplified by just using `(now->tv_sec - start->tv_sec) < 3` but this would be something between 2 and 4s.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: This one is not in a signal and the standard handler for uspace_rtapi writes to a fifo. Only uspace_ulapi prints directly which is fine due to this has no signal handler.

This "just link a different function with the same name" is hard to track.

But motion.c uses rtapi_set_msg_handler() to change it. I have to track this. I think this is a real time thread.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: It would be even nicer to use a real protocol instead of numbers/strings but not for now I think.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: > By the time you determine that there is no master, someone else can have become master.

As long as bind is atomic, this can not happen. If bind is not atomic, two applications could be server on the same socket which would be a linux bug.

>By the time you determine that there is a master, it may have exited before sending your command.

That can happen only if the application removes everything and then starts to add new things due to the master exits on exit or if there are no instances any more. Probably no linuxcnc app should do that but you can do that manually in halcmd. I have to test this.

> But you are right, this needs to be rethought and redone from the ground up. Lets leave it for now.

Ok, we do that later. I think it's not a bad issue, just hard to read code with possible issues.

src/rtapi/uspace_rtapi_main.cc
BsAtHome: Yes, motion.c is part of motmod and is part of the interface layer in realtime from commands from non-realtime.

src/rtapi/uspace_rtapi_main.cc
BsAtHome: Proper serialization, now that's a thought ;-)
Agreed, next future round of fixes can address that.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: I think this is solved by https://github.com/hdiethelm/linuxcnc-fork/commit/9dfe1e2eedcec7645ac46fbb3546f289c7c5402b + https://github.com/hdiethelm/linuxcnc-fork/commit/2692a61318b475325d1895f9c8611651d6d8c2a7

src/rtapi/uspace_rtapi_main.cc
hdiethelm: I tracked the motion.c part a bit more. I don't fully understand why this is done that way, but it goes to a fifo, so seams to be fine.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: See: https://github.com/hdiethelm/linuxcnc-fork/commit/9dfe1e2eedcec7645ac46fbb3546f289c7c5402b + https://github.com/hdiethelm/linuxcnc-fork/commit/2692a61318b475325d1895f9c8611651d6d8c2a7
I was not happy after the first commit, so I changed it a bit more.

src/rtapi/uspace_rtapi_main.cc
BsAtHome: Agreed

src/rtapi/uspace_rtapi_main.cc
BsAtHome: Agreed

BsAtHome:
I'll need to make a new look over it after all these changes to see if I missed stuff.

hdiethelm:
> I'll need to make a new look over it after all these changes to see if I missed stuff.

Yes, I changed again a lot based on your review...

I rewrote the socket serialize/deserialize anyway. After looking for to long at this read_strings, I could not let it be that way. It's still a bit handwavy but I also did not want to import a new library just for this.

Still open from your review:

- The signal handling. I don't really see the intent behind. Due to it is in harden_rt() it could have some influence on real time. But otherwise, I would just delete it all together. I see no point in a graceful exit if a signal is received except for debug purpose. And for this there are better solutions.

I would like to postpone this to separate PR's: 
 - Master/slave issue
 - Rootless
 - Might be better solution for function pointers?

Of course except you see something else/new.

hdiethelm:
I quickly tested the signal handler by doing a segfault on purpose in the load command handler:
```
../bin/rtapi_app load test
Note: Using POSIX realtime
rtapi_app: caught signal 11 - dumping core
```
Of course there is no core dump what so ever, also not on master.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: I have to correct my self:
https://github.com/hdiethelm/linuxcnc-fork/blob/rtapi_cleanup_v2/src/rtapi/uspace_rtapi_main.cc#L927
rtapi_print_msg() uses vfprintf() in the main thread. So it should also not be used in the signal handler.

hdiethelm:
So lacking better options, I went for: https://github.com/LinuxCNC/linuxcnc/commit/cd41d503eb9d99d53c46050c79a241f1a22a16ca
Do you think that's acceptable?
An option would be to add:
``` patch
@@ -661,7 +662,10 @@ static void signal_handler(int sig, siginfo_t * /*si*/, void * /*uctx*/) {
         WRITE_STDERR_STR("rtapi_app: UNKNOWN - shutting down\n");
         break;
     }
-    
+    rtapi_msg_queue.consume_all([](const message_t &m) {
+        WRITE_STDERR_STR(m.msg);
+    });
+
     _exit(-1);
 }
```
so the last messages are shown. But I am not sure if this is signal signal-safety conform.

However, it works. I blocked the normal consumer thread and then did `killall rtapi_app` to get messages printed in the signal handler:
```
Note: Using POSIX realtime
rtapi_app: SIGTERM - shutting down
Unexpected realtime delay on task 0 with period 25000
This Message will only display once per session.
Run the Latency Test and resolve before continuing.
```

BsAtHome:
> I quickly tested the signal handler by doing a segfault on purpose in the load command handler:
> rtapi_app: caught signal 11 - dumping core
> Of course there is no core dump what so ever, also not on master.

If you run on a system with systemd (most likely) you may want to use `coredumpctl` to access the coredump (run `coredumpctl debug` to start gdb with the dump).

src/rtapi/uspace_rtapi_main.cc
BsAtHome: The "simple" alternative for writing static/fixed messages in a signal handler are something like:
```C++
static void sighandler(int sig) { 
...
  static const char msg[] = "This is a message from a sighandler to stderr\n";
  ::write(2, msg, sizeof(msg)-1);  // minus one to prevent the \0 from printing
...
}
```

BsAtHome:
> WRITE_STDERR_STR("rtapi_app: UNKNOWN - shutting down\n");

That should work too. As long as the string is a constant and strlen is replaced by the compiler builtin, then the length expands as a constant too. Just a minor detail, you should put the expansion of a macro in parentheses.


> However, it works. I blocked the normal consumer thread and then did `killall rtapi_app` to get messages printed in the signal handler:
> [snip]

You are now catching more signals than the original. That is wrong. You actually want to coredump on "bad" signals (anything not with explicit meaning for the running program). The system should behave predictable and if that does not happen you want to be able to backtrack why it didn't behave. That is why it codedumps, so you can backtrack.

hdiethelm:
> If you run on a system with systemd (most likely) you may want to use `coredumpctl` to access the coredump (run `coredumpctl debug` to start gdb with the dump).

Ah now I get a coredump. It did work on master and I broke it. But I also told you I am not sure what I am doing. A fix will follow.

If not setid: `ulimit -c unlimited`

If setuid, `prctl(PR_SET_DUMPABLE, 1)` is used in `harden_rt()` but `echo 1 | sudo tee /proc/sys/fs/suid_dumpable` is needed. I don't have coredumpctl installed but just to test if it works, I don't need it.

 References:
https://linuxvox.com/blog/linux-enable-core-dump/ 
https://access.redhat.com/solutions/4896

src/rtapi/uspace_rtapi_app.cc
```diff
@@ -15,899 +15,65 @@
...
-}
-/* data for all tasks */
-struct rtapi_task *task_array[MAX_TASKS];
+struct rtapi_task *RtapiApp::task_array[MAX_TASKS];
 
 /* Priority functions.  Uspace uses POSIX task priorities. */
 
-int RtapiApp::prio_highest() const
-{
+int RtapiApp::prio_highest() const {
```
BsAtHome, line 60: Pedantic mode: Adding the open curly brace is a method/function is normally on the next line.
(several other places too)

hdiethelm:
So, next try. If you want  to take over for the signal part, feel free. The last time I used coredumps is like 20 years ago and for signals like 10.

I now just call abort() inside the handler which works and is allowed. But I don't know if it is good practice.

I added the message queue write out also. As long as there is no malloc/free in consume_all(), that is fine. And I don't think there is due to the queue is fixed size but I won't bother reviewing the boost library. I can remove it again if you think it's to dangerous.

For testing, I just run `killall -SIGFPE rtapi_app` in a second terminal.
rtapi_app itselv behaves fine now:
```
../bin/rtapi_app 
Note: Using POSIX realtime
rtapi_app: SIGFPE - aborting
Aborted (core dumped)
```
Coredump is there.

However, latency-test not. No clue why but it's the same on master. :
```
../scripts/latency-test 
Note: Using POSIX realtime
Unexpected realtime delay on task 0 with period 25000
This Message will only display once per session.
Run the Latency Test and resolve before continuing.
rtapi_app: SIGFPE - aborting
```
No coredump.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,1007 @@
...
+        /// XXX handle arguments
+        int (*start)(void) = DLSYM<int (*)(void)>(module, "rtapi_app_main");
+        if (!start) {
+            rtapi_print_msg(RTAPI_MSG_ERR, "%s: dlsym: %s\n", name.c_str(), dlerror());
+            dlclose(module);
+            modules.erase(name);
+            return -1;
+        }
+        int result;
+
```
BsAtHome, line 247: It was determined that when a component gets loaded it should ensure that both `rtapi_app_main` and `rtapi_app_exit` exist. See #3191 and #3925.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,1007 @@
...
+    return res == sizeof(int);
+}
+
+static void set_uint16(std::vector<char> &buf, uint16_t value, size_t idx) {
+    buf[idx] = 0xff & (value >> 0);
+    buf[idx + 1] = 0xff & (value >> 8);
+}
+
+static uint16_t get_uint16(std::vector<char> &buf, size_t idx) {
+    return ((uint16_t)buf[idx] << 0) | ((uint16_t)buf[idx + 1] << 8);
```
BsAtHome: Should use `const std::vector &`

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,1007 @@
...
+}
+
+static bool recv_args(int fd, std::vector<std::string> &args) {
+    //Get size
+    uint16_t tmp;
+    ssize_t res = recv(fd, &tmp, sizeof(uint16_t), 0);
+    if (res != sizeof(uint16_t)) {
```
BsAtHome, line 345: This assumes a SOCK_STREAM connection. Is that appropriate?

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,1007 @@
...
+        return false;
+    }
+
+    return true;
+}
+
+static bool send_args(int fd, const std::vector<std::string> &args) {
+    //Calculate size
+    size_t buff_size = 0;
+    buff_size += 2 * sizeof(uint16_t);
```
BsAtHome, line 379: Why 2*sizeof(uint16_t)?

I think you need to add a comment to explain the exact format of the serialized data and how it is read.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,1007 @@
...
+        buf.insert(buf.begin() + idx, args[i].begin(), args[i].end());
+        idx += args[i].size();
+    }
+    if (idx != buff_size) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "rtapi_app: Bug send_args: idx %li != buff_size %li\n", idx, buff_size);
+        return false;
+    }
+
+    //Send
+    ssize_t res = send(fd, buf.data(), buf.size(), 0);
```
BsAtHome, line 410: This may return -1 and errno == EAGAIN. In that case you need to retry.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,1007 @@
...
+    uint16_t tmp;
+    ssize_t res = recv(fd, &tmp, sizeof(uint16_t), 0);
+    if (res != sizeof(uint16_t)) {
+        return false;
+    }
+    size_t buff_size = tmp;
+
+    //Get data
+    std::vector<char> buf(buff_size);
+    res = recv(fd, buf.data(), buff_size, 0);
```
BsAtHome, line 352: May return -1 and errno == EAGAIN. You need to retry in that case.

src/rtapi/uspace_rtapi_app.cc
hdiethelm: That's from clang-format. So it's not several other places but all other places I formatted... ;-)

src/rtapi/uspace_rtapi_main.cc
hdiethelm: I can integrate https://github.com/LinuxCNC/linuxcnc/pull/3925 in my branch or wait for https://github.com/LinuxCNC/linuxcnc/pull/3925 to get merged and then rebase on top of master. How to move forward?

BsAtHome:
> So, next try. If you want to take over for the signal part, feel free. The last time I used coredumps is like 20 years ago and for signals like 10.

Most of it is now fine :-)

> I now just call abort() inside the handler which works and is allowed. But I don't know if it is good practice.

Well, you are allowed to call abort() in the signal handler. The existing solution (using kill()) re-raises the caught signal with the handler to default. Using abort() seems the better choice.

src/rtapi/uspace_rtapi_main.cc
BsAtHome: Just merged #3925. Go ahead and rebase/ff.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Yes: uspace_rtapi_main.cc: 584 `int fd = socket(PF_UNIX, SOCK_STREAM, 0);`

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Good catch. Done.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: I added a description. `uint16_t n_args` could be left away, it is redundant, but it makes parsing slightly simpler.
The arguments are limited to 64k bytes. I was thinking to go to uint32_t but I don't think this is necessary.

BsAtHome: In the "old days" there used to be a 32k limit for cmd-line args. You only hit that limit when your shell expansion was a bit large. I guess 64k is good enough for the youngsters ;-)

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Done.

src/rtapi/uspace_rtapi_app.cc
hdiethelm: It's this option:
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakbeforebraces
Now this option is commented out in the file from the PR you mentioned. But it is active anyway duet to:
`BasedOnStyle: LLVM` and llvm has braces always on the same line.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Is this needed? https://www.man7.org/linux/man-pages/man2/send.2.html says that happens only on non blocking sockets. And the socket used in here is blocking.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Is this needed? https://www.man7.org/linux/man-pages/man2/recv.2.html says that happens only on non blocking sockets. And the socket used in here is blocking.

BsAtHome: You are right, my mind was probably asleep or walkabout when I wrote that...

The error you need to check is EINTR (a signal interrupted the operation before it was executed).

src/rtapi/uspace_rtapi_main.cc
hdiethelm: This should be solved now.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: I don't really like having two lines and a lot of copy paste. But I also did not like my define so much, so I replaced it by a simple function. strlen() and write() is fine according to signal-safety(7).

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Hmm, I think interruption by a signal should be handled in a specific way if signals are used in a meaningful way, like setting a flag what to do and only retry in specific cases. This has to be done right trough the whole HAL like in hm2_eth.c or not at all.

rtapi_app just catches a few signals to exit in all cases, just with a different message and explicit coredump. Even ALRM just exits the process.
SIGTSTP is ignored according to the code: https://github.com/hdiethelm/linuxcnc-fork/blob/rtapi_cleanup_v2/src/rtapi/uspace_rtapi_main.cc#L816

So I think it only makes sense to add EINTR error handling if the signal handler does something usefull and in that case, it is specific to what you want to do. Just adding a generic while(EINTR) error handling could even make it more difficult in the future if someone wants to add a specific signal for example to nicely exit the process.

I tried `killall -SIGTSTP rtapi_app` to get accept https://github.com/hdiethelm/linuxcnc-fork/blob/rtapi_cleanup_v2/src/rtapi/uspace_rtapi_main.cc#L461 to return EINTR but this doesn't happen due to the signal is ignored.

Context: 
https://www.man7.org/linux/man-pages/man7/signal.7.html
https://unix.stackexchange.com/questions/253349/what-is-the-rationale-behind-eintr

BsAtHome: A standard way to handle EINTR is:
```C++
retry_recv:
    ssize_t err = recv(...); // Also for send(), read(), write(),...
    if(err < 0) {
        if(EINTR == errno)
            goto retry_recv;
        // else handle the error
    }
```
The question is whether it happens in practice. What parts in the code allow, handle or depend on signals that can cause an interrupted syscall.

Another problem that can occur with reads and writes is that there are only partial reads/writes. But this is usually ignored and in practice does not happen on Linux [pipes do have a PIPE_BUF limit for atomicity]. As long as you write/read under a page size, then it should never occur afaik, so lets ignore it ;-)

src/rtapi/uspace_rtapi_main.cc
BsAtHome: Fine with a function. It looks nicer and cleaner. It is `static inline` and will be gone anyway. The strlen() call will use __builtin_strlen, which sees it as a constant and be replaced with a number.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: > The question is whether it happens in practice. What parts in the code allow, handle or depend on signals that can cause an interrupted syscall.

It seams to not happen in practice. I did not manage to achieve an EINTR to test it and I also did not find anything in the LinuxCNC issues. So I think it is save to just not check for EINTR, it was never checked in the 20 years the code uses sockets in a similar way in rtapi_usespace: https://github.com/LinuxCNC/linuxcnc/commit/50e259cf91c61b76c4d6f3d1cd5697fb21eb3f31

Otherwise, EINTR should be handled also for connect/accept and might be more. So for consistency, I prefer to not check this at all as long as it never happens. Of course, if someone starts to use signals for specific actions, it should be checked all over the place and specific actions should be taken. This then would look like:

```C
retry_recv:
    ssize_t err = recv(...); // Also for send(), read(), write(),...
    if(err < 0 && !exit_flag) {
        if(EINTR == errno)
            goto retry_recv;
        // else handle the error
    }
 ```

Comment (hdiethelm) [reply src/rtapi/uspace_rtapi_main.cc:350]: BTW: My only change was to replace write by send and read by recv for readability. read/write on sockets just feels wrong even though it is equal. The man page says:

The only difference between send() and [write(2)](https://www.man7.org/linux/man-pages/man2/write.2.html) is the presence of flags. With a zero flags argument, send() is equivalent to [write(2)](https://www.man7.org/linux/man-pages/man2/write.2.html).

and

The only difference between recv() and [read(2)](https://www.man7.org/linux/man-pages/man2/read.2.html) is the presence of flags.  With a zero flags argument, recv() is generally equivalent to [read(2)](https://www.man7.org/linux/man-pages/man2/read.2.html) (but see NOTES)`

Comment (hdiethelm) [reply src/rtapi/uspace_rtapi_main.cc:350]: Now I manage to get EINTR for accept():
Comment out https://github.com/hdiethelm/linuxcnc-fork/blob/rtapi_cleanup_v2/src/rtapi/uspace_rtapi_main.cc#L712 so the signal handler doesn't stop the program.
Terminal 1: `../bin/rtapi_app`
Terminal 2: `killall rtapi_app`
Result:
```
rtapi_app: SIGTERM - shutting down
rtapi_app: failed to accept connection from slave: Interrupted system call
```
But I still think this is a use-case that should be properly implemented implemented if needed in the future.

Never the less, while looking trough the code, I found the run_threads() function to be pretty useless and removed it all together. See commit https://github.com/LinuxCNC/linuxcnc/pull/3919/commits/f52ef214c84521e1800e9cebdb6b67e495d43653.

**Comment by hdiethelm** [comment]
I tested the actual state of this PR both with latency-histogram and on my machine. No issues so far except https://github.com/LinuxCNC/linuxcnc/issues/3946.

**Comment by hdiethelm** [comment]
@BsAtHome Thanks already for the detailed review.

Still open points:
- https://github.com/LinuxCNC/linuxcnc/pull/3919#issuecomment-4217381296
- Socket error EINTR handling: I would ignore it but if you insist on it, I can add it, no big deal

Anything else?

**Comment by BsAtHome** [comment]
My guess is that were almost there. I'd like to have another look at the complete picture to see if I overlooked something.

The EINTR handling is a completeness thing. It is probably not a big deal, at least, it hasn't been so far.

There are other correctness and completeness problems. Like reading from a socket/file/pipe may actually return less than requested under some circumstances. It is very pronounced when dealing with serial ports, but known to happen with sockets and pipes too when atomicity no longer can be guaranteed and also very pronounced when non-blocking IO plays a role. But, since the data written and read are very small blocks (only 20..100 bytes or so) you'd probably never ever see the problem happen. And, I don't think this code will have to run on another non-Linux kernel any time soon.

**Comment by hdiethelm** [comment]
Ok, I do now all checking, even if it probably never fails, it doesn't hurt. To avoid master freezing forever, I added a timeout of 10s.

Additionally, I found a mistake in send_args() and corrected it. See last commits.

Long ago, I implemented tcp / udp using sockets. There, you have a bunch of errors and send() / recv() only process the requested amount up to a certain size. I think it depends on the socket type but better be save than sorry.

I tested it using the following code:
Terminal 1: `../bin/rtapi_app`
Terminal 2: 
```
echo -n > testfile ; for((i=0;i<65523;i++)); do echo -n x >> testfile ; done
../bin/rtapi_app load $(cat testfile)
#All ok
echo -n > testfile ; for((i=0;i<65524;i++)); do echo -n x >> testfile ; done
../bin/rtapi_app load $(cat testfile)
#To much data:
rtapi_app: send_args: args to big, size = 65536!
rtapi_app: failed to write to master
```

Up to 65533 bytes, there is no retry, i added some debug printf's for this after send/recv:
Client: 
send 65535
Master:
recv 2
recv 65533

File: src/rtapi/uspace_rtapi_main.cc
Comment (BsAtHome) [inline src/rtapi/uspace_rtapi_main.cc:408]: Wouldn't you need to make sure that `idx+1` exists?

Comment (hdiethelm) [reply src/rtapi/uspace_rtapi_main.cc:408]: Valid point. Added bounds checking there and also for extracting argument.

I went for buf.at() / try + catch mainly due to a separate check before access has the risk that someone later changes this code for example to buf[idx + 2] and forgets to update the check right above.

**Comment by grandixximo** [comment]
Hi @hdiethelm, I have been working on a follow-up to PR #918 (rootless preempt-rt, plus fixes for issue #3928) on a local branch. Since your rtapi cleanup touches the same files, I would like to avoid adding to your workload.

I just opened it as a draft PR (#3964) so you can see the shape of it. Whatever is least burden for you is fine with me: I can wait for PR #3919 to land and rebase afterwards, or target your branch directly. Just tell me which you prefer and I will do the rebasing. No rush on my side.

**Comment by hdiethelm** [comment]
Hi @grandixximo 
Thanks for informing me and taking regard to my PR. I'm busy today, i can look deeper into your PR in the next days.
It's probably the least effort for both of us if you base your changes on my branch and as soon as mine is merged, rebase yours to master.
It shoud not take to long any more from my side. All important parts are done, i am just waiting for additional feedback.

**Comment by grandixximo** [comment]
Done rebased #3964 on this PR.

**Comment by hdiethelm** [comment]
Just rebased to master again.
@grandixximo You can rebase your branch back to mine, should be straight forward.
@BsAtHome Anything still open from your side? Otherwise I will do some final testing in the next few days and then it is ready to merge.

**Comment by andypugh** [comment]
I have converted this to draft, until it isn't WIP any more.

File: src/rtapi/uspace_rtapi_main.cc [hunk 20]
```diff
@@ -0,0 +1,1135 @@
...
+        return -1;
+    }
+    return 0;
+}
+
+static void remove_quotes(std::string &s) {
+    s.erase(remove_copy(s.begin(), s.end(), s.begin(), '"'), s.end());
+}
+
+static int do_comp_args(void *module, std::vector<std::string> args) {
```
BsAtHome: Shouldn't the vector args be const &?

src/rtapi/uspace_rtapi_main.cc
BsAtHome: Where will it be caught?

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,1135 @@
...
+            rtapi_print_msg(
+                RTAPI_MSG_ERR, "rtapi_app: send_args failed, sent only %li of %li bytes\n", res, buf.size()
+            );
+        }
+        return false;
+    }
+    return true;
+}
+
+static int handle_command(std::vector<std::string> args) {
```
BsAtHome: Shouldn't args vector be const &?

BsAtHome:
I guess the just added comments are the last I have to say for this review. There might be other things, but that should not matter.

The final question is who has been testing this. Are there any things caught in running this?

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Done, was not straight forward but solved an other strange implementation: https://github.com/hdiethelm/linuxcnc-fork/commit/69c0eb82a3cf5123c7b6d9dae8ea07001c9afe66

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Done, also not straight forward.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: It is caught here: https://github.com/hdiethelm/linuxcnc-fork/blob/rtapi_cleanup_v2/src/rtapi/uspace_rtapi_main.cc#L464
If desired, I can also use return true/false for error handling.

BsAtHome:
If other people can confirm proper function... I'm fine with it.

NTULINUX:
I don't have a CNC to test on but I have LinuxCNC's sims working in Fedora and two VMs (Gentoo and Debian, both with PREEMPT_RT and RTAI available)

BsAtHome:
Just one more thing...

Can you run from the src directory: `../scripts/cppcheck.sh rtapi/*.cc rtapi/*.c`
and see if there are any comments, warnings or errors that way around.

hdiethelm:
> Can you run from the src directory: `../scripts/cppcheck.sh rtapi/*.cc rtapi/*.c` and see if there are any comments, warnings or errors that way around.

It shows all green.

One (second) last commit from my side: cherry-picked https://github.com/LinuxCNC/linuxcnc/commit/00d534c87464a3ed446656998aa02b8abc74b391 from 2.9

I will do some final testing in 1-2 days and it is also additionally tested in https://github.com/LinuxCNC/linuxcnc/pull/3964 which bases of this PR.

What I can test in Debian Trixie:
- All different real time implementations
- Real CNC with mesa card

hdiethelm:
> I don't have a CNC to test on but I have LinuxCNC's sims working in Fedora and two VMs (Gentoo and Debian, both with PREEMPT_RT and RTAI available)

This sounds good, so together we have already good coverage.

hdiethelm:
So, i went trough what I can test. This is:

Config: Debian Trixie / axis UI / started from source folder / Machine: Ethernet + 7i96s:
- PREEMT_RT
- Xenomai3
- Xenomai4

for each:

- latency-test
- latency-histogram
- linuxcnc on my CNC + axis + air milling the linuxcnc logo.

In my VM, Debian package installed:

- PREEMT_RT
- Xenomai3
- Xenomai4

for each:

- latency-test
- latency-histogram
- linuxcnc + axis + air milling the linuxcnc logo

for PREEMT_RT:
- linuxcnc + gmocappy + air milling the linuxcnc logo

I was not able to test RTAI.

There where no issues created by this branch, also the timing is still good. However:
- https://github.com/LinuxCNC/linuxcnc/issues/3978
- https://github.com/LinuxCNC/linuxcnc/issues/3979
- https://github.com/LinuxCNC/linuxcnc/pull/3980
  - I had to go back to 2.9 to check if https://github.com/LinuxCNC/linuxcnc/issues/3979 is only on master and was not able to build
- In Xenomai3 / Xenomai4 i sometimes have timeouts while starting the real machine, both on 2.9 and master, I will investigate that later.
- In master this branch bases on, the config examples are copied to the wrong place. This was fixed in latest master.

So from my side it is save to merge. @NTULINUX Where you able to test it?

hdiethelm:
After reverting locally https://github.com/LinuxCNC/linuxcnc/pull/3936  + https://github.com/LinuxCNC/linuxcnc/pull/3968 I was able to test RTAI in a VM, started from source folder:
- latency-test
- latency-histogram
- linuxcnc + axis + air milling the linuxcnc logo

NTULINUX:
Awaiting https://github.com/LinuxCNC/linuxcnc/issues/3978 to get fixed so I can try this on RTAI. So far, this commit (cleanup v2) works with uspace on Fedora.

edit: did not see your comment, will try again in a bit with those reverts.

hdiethelm:
Re-based onto master with https://github.com/LinuxCNC/linuxcnc/issues/3978 merged and performed a few tests again.
@NTULINUX Now you should be able to test it.

NTULINUX:
Looks good!

NTULINUX (approved):

hdiethelm:
> Looks good!

Thanks for testing! So I mark this as ready.

Of course there might still be issues, I did a lot of changes. Just tag me if there is a bug related to this and I will investigate.

NTULINUX:
The only thing I noticed is that the latency-histogram takes a little while to get going and sometimes throws dots in the terminal for awhile, or the graph freezes initially for a few seconds and then goes back to being normal without user intervention.

grandixximo:
I think that is new histogram behavior caused by Bertho's component, or possibly my rewrite of BLT, run with --legacy flag to check old behavior

NTULINUX:
Well adding `--legacy` to the histogram does certainly change the behavior but it's still not identical to the old behavior. In either case, it's not related to this commit. I believe the rtapi clean-up is good to go!

hdiethelm:
@NTULINUX Is this using RTAI? In PREEMT_RT, it starts immediately on my PC.

Can you check this also on master? I remember slow or even no starts on RTAI from time to time but not connected to this branch and also before the latency-histogram improvement. And now I just had a full VM freeze on RTAI while trying to test it.

I did not touch anything for RTAI. However, trough includes, something might has changed. To verify, I built master this PR bases on and this branch. There are binary identical except some timestamps in generated files.

But there is one mistake I just discovered: `lib/liblinuxcnc-uspace-posix.so` is also build for RTAI even thought not needed there. Should not be an issue but is also not nice. I will go back to draft and correct the Submakefile.

hdiethelm:
So, I found it. I did not see the `if` in the Submakefile. Now the RTAI build in 791f9c0bb445541 and [rtapi_cleanup_v2](hdiethelm:rtapi_cleanup_v2) are binary identical except generated timestamps in non-binary files.

The last commit doesn't change anything for uspace build. For the RTAI build, `lib/liblinuxcnc-uspace-posix.so` is not built any more.

I doubt the unneeded .so has any influence on RTAI, so I don't except any behavioral changes.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,1145 @@
...
+    }
+}
+
+/*
+ * Fully checked send/recv
+ * Will retry on EINTR, so to abort a send_data/recv_data on a
+ * signal, change to something like while(remaining > 0 && !exit_flag)
+ * and set exit_flag in signal handler
+ */
+static int send_data(int fd, const void *buf, size_t n, int flags) {
```
BsAtHome: You return `n`of type `size_t` as an `int` return value.

You may want to use `ssize_t` as argument type, return value and counter `n_rem`.

You actually assign to a ssize_t line 379.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,1145 @@
...
+            return (n - n_rem); // No more data
+        } else {
+            ptr += n_ret;
+            n_rem -= n_ret;
+        }
+    }
+    return n; // All sent
+}
+
+static int recv_data(int fd, void *buf, size_t n, int flags) {
```
BsAtHome: Same `size_t` vs. `int` problem. Convert to `ssize_t` seems more appropriate.

You assign to a ssize_t in line 395.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Good catch. I changed a lot from int to size_t but of course I missed my own new function. I left size_t as argument, this is standard in the c library for example for send / recv.
This could be cached with -Wconversion, so I activated it temporary and corrected other issues in this code. However, there are many more in other included code I did not want to touch at this stage.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Fixed.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: And pedantic mode: https://stackoverflow.com/questions/5859707/rationale-for-system-calls-that-allow-request-of-size-t-but-result-of-only-ssize
Force pushed.

hdiethelm: ...not that this could have triggered any bug with the actual implementation due to the size is limited to 64k but who knows what this code is used later for. So better save than sorry.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,1158 @@
...
+    sigaction(SIGFPE, &sig_act, (struct sigaction *)NULL);
+    sigaction(SIGTERM, &sig_act, (struct sigaction *)NULL);
+    sigaction(SIGINT, &sig_act, (struct sigaction *)NULL);
+
+#ifdef __linux__
+    int fd = open("/dev/cpu_dma_latency", O_WRONLY | O_CLOEXEC);
+    if (fd < 0) {
+        rtapi_print_msg(RTAPI_MSG_WARN, "failed to open /dev/cpu_dma_latency: %s\n", strerror(errno));
+    } else {
+        size_t r = write(fd, "\0\0\0\0", 4);
```
BsAtHome: This is wrong. write(2) has return type ssize_t.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -0,0 +1,1158 @@
...
+        }
+    }
+    return n; // All sent
+}
+
+static ssize_t recv_data(int fd, void *buf, size_t n, int flags) {
+    if(n > SSIZE_MAX){
+        errno=EFBIG;
+        return -1;
+    }
```
BsAtHome, line 344: I think you can spare yourself the trouble when you declare the argument `n` as ssize_t too.

If you run out of space with a ssize_t, then your system is about to crash and burn anyway. Or do you have 8 exibyte stuffed or hidden in your system?

(also in send_data())

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Shure it is, stupid ssize vs size... ;-) Sorry.

src/rtapi/uspace_rtapi_main.cc
hdiethelm: There is a reason mentioned in https://stackoverflow.com/questions/5859707/rationale-for-system-calls-that-allow-request-of-size-t-but-result-of-only-ssize. I tested it:
```c++
#include <sys/types.h>
#include <vector>

void stest(size_t x) {
}

void sstest(ssize_t x) {
}

int main(void){
    std::vector<int> x;
    stest(x.size());
    sstest(x.size());
}
```
Result with `-Wsign-conversion`:
```
<source>: In function 'int main()':
<source>:15:18: warning: conversion to 'ssize_t' {aka 'long int'} from 'std::vector<int>::size_type' {aka 'long unsigned int'} may change the sign of the result [-Wsign-conversion]
   15 |     sstest(x.size());
      |            ~~~~~~^~
Compiler returned: 0
```

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Fixed, force pushed.

src/rtapi/uspace_rtapi_main.cc
BsAtHome: Right. But when a value of > SSIZE_MAX is being written, then the write will fail anyway afaik.

hdiethelm: 8 exibyte would be nice, with today's ram prices... :-D

With ssize_t, I would have to check < 0 for pedantic mode, also not better.

In https://www.man7.org/linux/man-pages/man2/write.2.html, the behavior is specified under notes for linux.

I could just leave the check away arguing that nobody has this much ram in the near future. But having `ssize_t recv_data(int fd, void *buf, size_t n, int flags)` signature wise equal to `ssize_t recv (int fd, void *buf, size_t n, int flags)` is just nice in my opinion.

The whole code is anyway not `-Wsign-conversion` compatible, there are like 20 warnings just in uspace_rtapi_main.cc including in this two functions. I could fix them in my code but at this point, 2^63 vs 2^64 bytes doesn't really matter.

BsAtHome: Agreed.

hdiethelm: So we leave it as it is now?

BsAtHome: If all the size_t/ssize_t/int confusion is fixed, sure.

hdiethelm: In my code, yes. In code I did not touch, just moved + formatted, there are still some non-perfect types.

I corrected one more I felt save to do so. task_id should be a size_t but it goes trough the whole RT stack, so I did not want to touch it. Testing is already trough and I don't want to retest everything.

hdiethelm:
Rebased to master quick retest: latency-historgram / latency-test / linuxcnc axis_mm
All fine

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Last real change: https://github.com/LinuxCNC/linuxcnc/compare/02c2cf696b3c12eb4e4f58d760968440a3712593..777c8d9c48cce30dab474c4b6c5b09b321b0949d

BsAtHome:
Then I think this is finally done, cooked and ready :-)

Issue #991, opened 2020-11-22
End of line comments in .ini lead to hanging / errors in 2.8


Hans470:
With Linuxcnc 2.7.x it was no problem to have comments in the ini-files and the end of the line like
``` ini
[AXIS_X]
MIN_LIMIT = 0.0 # comment
```
But in version 2.8 it seems to to be supported any longer. The problem is that Linuxcnc then nearly hangs up.

## Here are the steps I follow to reproduce the issue:

 1.
 2.
 3.

## It worked properly before this:
(If the behavior changed after making a particular change in hardware or
software, describe the change you think is responsible.  E.g., "after upgrading
from LinuxCNC 2.7.3 to 2.7.4")

## Information about my hardware and software:

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`):
 * I am using this kernel version (shown by `uname -a`):
 * I am running ...
   * [ ] A binary version from linuxcnc.org (including buildbot.linuxcnc.org)
   * [ ] A binary I built myself
   * [ ] A binary version from some other source besides linuxcnc.org
 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`):
 * I am using this user interface (GUI) (e.g., AXIS, Touchy, gmoccapy, etc):
 * I am using this interface hardware vendor and chipset (e.g., parallel port, ethernet port, FPGA card):

andypugh:
I seem to remember this as a deliberate change, but I don't recall the reasoning behind it.

phillc54:
It is noted in the documents:
http://linuxcnc.org/docs/2.8/html/config/ini-config.html#_comments

Hans470:
Ok, but in 2.7 it was still accepted even though the documentation says not.
Maybe this can be checked by the update script to prevent later errors?
It took me too much time to find that the source of the error was only the stupid comment.

rene-dev:
Is there any technical reason for the new behavior?

andypugh:
Dewey found the commit:

https://github.com/LinuxCNC/linuxcnc/commit/c3ac4f5bf2f76b97

Hans470:
Ok but when an ERR_CONVERSION is thrown - why is Linuxnc still starting?

amitSA:
[linuxcnc version 2.9.3]

I'm new to linuxcnc, I was adding in-line comments and `linuxcnc my-mill.ini` would hang after printing this:
```
LINUXCNC - 2.9.3
[...]

Issue #3863, opened 2026-03-15
Master does not correctly recognize boolean ini values


Sigma1912:
As per documentation
https://linuxcnc.org/docs/devel/html/config/ini-config.html#sub:ini:sec:rs274ngc
> Boolean values can be on one of TRUE, YES or 1 for true/enabled and one of FALSE, NO or 0 for false/disabled. The case is ignored.

Examples using 'configs/sim/axis/axis_9axis.ini' 

1. default behavior (ie no 'WRAPPED_ROTARY ' entry)      
      MDI:  'G0 A400'   -> A-axis moves to '400'                      

2. Insert [AXIS_A]  'WRAPPED_ROTARY = 0' (expected is same behavior as with no 'WRAPPED_ROTARY ' entry )
      MDI command : 'G0 A400'  ->  We get no error and  movement to A: 40 
                
3. insert [AXIS_A]  'WRAPPED_ROTARY = 1' 
      MDI:  'G0 A400'   -> We get the expected interpreter error and no movement

4. change to [AXIS_A]  'WRAPPED_ROTARY = YES' (expected is same behavior as 'WRAPPED_ROTARY = 1' )
      MDI command : 'G0 A400'  ->  We get no error and  movement to A: 40

**It seems that anything other than no entry or '1' leads to unexpected behavior.**

This used to work in version 2.9

andypugh:
There certainly are discrepancies. 
Taking the example of [DISPLAY]LATHE. 

Axis:
LATHE = 1 -> lathe
LATHE = 0 -> lathe
\# LATHE = 1 -> not lathe

Gscreen:
LATHE = 1 -> lathe
LATHE = YES -> not lathe

The best way to handle this would be to move the boolean parsing into ini_find, with a method to specify that the returned value should be parsed as a boolean.

Sigma1912:
Unfortunately it's not just the python code, it's also inconsistent internally:

' WRAPPED_ROTARY = YES'  -> **not** wrapped_rotary

but 
'HOME_USE_INDEX = YES' -> homing uses index

Sigma1912:
Two examples with error output:

'NO_FORCE_HOMING' seems to only accept '1' or '0', anything else defaults to '0' and will output an error message to terminal (the config still starts):
`invalid [TRAJ] NO_FORCE_HOMING in /home/user/git/mylinuxcnc-master/configs/sim/axis/axis_9axis.ini (false); using default 0 `

[...]

PR #3921, opened 2026-04-10
ini: Implement a new ini-file parser and adapt the code to use it.

BsAtHome:
This PR replaces all INI-file handling with a new reader and parser that uses internal type conversion for consistent handling of variables throughout the sources, as long as you use the same typed reading method for the same variable.

The problems with the old INI-file reader, parser and usage is exemplified in #3863 where it was noted that boolean values are handled very inconsistently between UIs and also in the C/C++ sources. Many UIs use python where there never was established a typed interface to read INI-file values and code used manual casting to get (hopefully correct) results.

All LinuxCNC sources using INI-files have been updated in the PR to use the new API. Porting was done with care, but mistakes can happen... Please try this branch and verify that everything is working for you with several UIs.

The INI-file content is now parsed more strictly where both section and variable names may only contain letters, digits and underscore (where the first character may not be a digit). Comments using `#` and `;` are allowed everywhere, also in lines with values. You can now simply quote the value string if you need to or want to embed these characters. Quoted values can use single or double quotes and multiple quoted fragments are merged into one string. A double quoted string may contain the usual escape sequences, as well as hex literal escapes (`\xHH`), UTF-16 (`\uHHHH`) and UTF-32 (`\UHHHHHHHH`) escapes. Both UTF-16 and UTF-32 escapes are converted to UTF-8 characters. Embedding the NUL (`\0`) character is not allowed. Any resulting value is checked for being valid UTF-8.

A few examples:
```ini
[section]
value = value in plain sight # this is a comment
other = "a comment character # in the value" # and this is a comment
smile = "a smile \U0001F600"
string= "segmented" 'fragments' # that gives: "segmentedfragments"
boola = true # or "1" or "on" or "yes" (case insensitive)
boolb = false # or "0" or "off" or "no" (case insensitive)
int = 1234
real = 1.76e-3
prefix = 0xabcdef # hex
prefix = 0o765432 # octal
prefix = 0b101010 # binary
```

Include files using `#INCLUDE` are now part of the new parser and handled automatically. Previously, INI-file includes were merged in the `linuxcnc` (startup) script, but that made general usage difficult and could not be used or properly checked by outside programs. The existence of includes is also the reason why you _cannot_ use python's configparser, or others, because they do not support the include mechanism. The `inivar` program has been updated to call the new `inivalue` program, which provides much more functionality (including handling the includes properly).

There are now established several methods to read values and convert them into the following types:
- string (always UTF-8)
- signed integer (64 bit)
- unsigned integer (64 bit)
- real (double floating point)
- boolean

The C++ API also provides query operations like `hasSection` and `hasVariable`.  A new method is provided to allow any value->value maps to be easily implemented to support enumerated types in INI-file values (either case sensitive or insensitive).

All methods allow for defaults to be specified if the specific [section]variable is not found. This reduces the amount of repeated code significantly and make it easier to read. Convenience methods are provided to use standard integers and simply truncates the 64 bit signed integer version.

The old parser sources have been removed from `src/libnml/inifile/*` and the new parser resides in `src/emc/ini/inifile.*`. The shared library has a new version number and is now `liblinuxcncini.so.1`.

The python API has been extended greatly to support the same typed system using:
- getstring
- getsint
- getuint
- getreal
- getbool

Convenience methods are provided for compatibility and resemblance to python's configparser (which should never be used for LinuxCNC's main INI-file):
- getfloat - maps to getreal
- getint - maps to getsint
- find - maps to getstring

Additional methods:
- findall - gets all values (as string) of a specific name
- hassection - to query the existence of a section
- hasvariable - to query the existence of a variable

There are also additional methods implemented for support. You can try yourself by printing the builtin doc:
```python
import linuxcnc
print(linuxcnc.ini.__doc__) # ini module info
print(linuxcnc.ini.getbool.__doc__) # getbool() info
...
```


grandixximo:
Are you aware of the `setlocale()` possible race in `convertReal()/convertSInt()/etc` if two thread were to call `findReal()`?
Intentional, or should it move to use newlocale()/uselocale() (POSIX) or thread-local locale?

Should we have a deprecation alias for the removed Python `.find()` for out of tree callers?

Impressive commit by the way 👍

BsAtHome:
You are right, there is an MT problem. Not that it normally happens, but that is no excuse.

Created a local static mutex and a lock_guard surrounding the strto{d,ll,ull} conversions. And also, actually added the "C" locale switch to the integer conversions too. They too <i>may</i> have locale specific forms and we only want one single form across all languages for portability. The mutex should be fine here, I guess. The locale is returned to its previous value within the locked sequence.

The python method linuxcnc.ini.find() has not been removed or deprecated. It simply maps to linuxcnc.ini.findstring(). I did not see any real harm to leave it in place. Otherwise I'd need to clean/fix a lot more instances in the python code (and I may have become a bit lazy after several weeks of staring at this code).

Deprecating find() would cause a semi-hard break for all local user code and I did not absolutely see the need for that. We just need to ensure that they use the typed interface if/when users start contributing their code after this gets merged. However, there might be a reason to condition users to adapt by deprecation, but I am open to suggestions. So, any (other) compelling reason(s)?

grandixximo:
My bad, I saw `find()` in the red while reviewing, and assumed it was gone, never mind, I found the remap, no compelling reason.

BsAtHome:
Places where I needed to add `fallback="blabla"` it got rewritten from find() to getstring() because I was already typing ;-)

BsAtHome:
@grandixximo Strike my last... the race on setlocale has not been solved properly. The local mutex is wrong and even a mutex within the IniFileContent or inifile global instance will not solve it because other threads may still interfere. The only workable solution is a thread-local locale setting. Needs to be fixed.

BsAtHome:
> ... should it move to use newlocale()/uselocale() (POSIX) or thread-local locale?

Yes it should and it now does. The uselocale() is thread local, so that should actually fix this problem completely.

I'm not sure it would be a good idea to share a newlocale() created locale between threads, which would allow for caching (saves a newlocale()/freelocale() call on each invocation). Don't know the internal details, so I can't be sure. However, this stuff is not speed critical at all and most of it is only called at startup. Which is why I don't think we should bother.

hansu:
Gmoccapy  fails to start after this merge:
```

  File "/home/cnc/linuxcnc-master/bin/gmoccapy", line 40, in <module>
    from common import hal_glib # needed to make our own hal pins
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/cnc/linuxcnc-master/lib/python/common/hal_glib.py", line 5, in <module>
    import linuxcnc
ImportError: /home/cnc/linuxcnc-master/lib/python/linuxcnc.so: undefined symbol: _ZN8linuxcnc7IniFileC1ERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE

```

BsAtHome:
I've seen this one happen too. It is caused by old code remaining from an old build (something is linked against the old version library that not exports the new function). You need to do a full rebuild. Use at least make clean (maybe even git clean) and rebuild.

hansu:
`make clean` isn't enough.  `git clean -dfx` on the root dir did it. Thanks!

hansu:
Current master fails running max5kins.ini. Worked before this merge. Looks like a INI-parse issue.


```
LINUXCNC - 2.10.0~pre1
Machine configuration directory is '/home/cnc/linuxcnc-master/configs/sim/axis/vismach/5axis/max5'
Machine configuration file is 'max5kins.ini'
check_config: Unchecked: [KINS]KINEMATICS=maxkins
Starting LinuxCNC...
linuxcnc TPMOD=tpmod HOMEMOD=homemod EMCMOT=motmod
Note: Using POSIX realtime
Found file(REL): ./max5kins.hal
No isolated CPU's found, expect some latency or set RTAPI_CPU_NUMBER to select CPU
emcTrajSetUnits failed to set [TRAJ]LINEAR_UNITS or [TRAJ]ANGULAR_UNITS
emcMotionInit: emcTrajInit failed
emcTrajSetUnits failed to set [TRAJ]LINEAR_UNITS or [TRAJ]ANGULAR_UNITS
emcMotionInit: emcTrajInit failed
emcTrajSetUnits failed to set [TRAJ]LINEAR_UNITS or [TRAJ]ANGULAR_UNITS
emcMotionInit: emcTrajInit failed
emcTrajSetUnits failed to set [TRAJ]LINEAR_UNITS or [TRAJ]ANGULAR_UNITS
emcMotionInit: emcTrajInit failed
emcTrajSetUnits failed to set [TRAJ]LINEAR_UNITS or [TRAJ]ANGULAR_UNITS
emcMotionInit: emcTrajInit failed
emcTrajSetUnits failed to set [TRAJ]LINEAR_UNITS or [TRAJ]ANGULAR_UNITS
emcMotionInit: emcTrajInit failed
emcTrajSetUnits failed to set [TRAJ]LINEAR_UNITS or [TRAJ]ANGULAR_UNITS
emcMotionInit: emcTrajInit failed
emcTrajSetUnits failed to set [TRAJ]LINEAR_UNITS or [TRAJ]ANGULAR_UNITS
emcMotionInit: emcTrajInit failed
emcTrajSetUnits failed to set [TRAJ]LINEAR_UNITS or [TRAJ]ANGULAR_UNITS
emcMotionInit: emcTrajInit failed
emcTrajSetUnits failed to set [TRAJ]LINEAR_UNITS or [TRAJ]ANGULAR_UNITS
emcMotionInit: emcTrajInit failed
ini.traj_max_velocity still 0.0 after 14 seconds
emc/task/emctaskmain.cc 3008: can't initialize motion

Warning: Missing required specifier (has angular joint or axis):
    [DISPLAY]MAX_ANGULAR_VELOCITY
or  [TRAJ]MAX_ANGULAR_VELOCITY
See the 'INI Configuration' documents

note: MAXV     max: 346.000 units/sec 20760.000 units/min
note: LJOG     max: 346.000 units/sec 20760.000 units/min
note: LJOG default: 200.000 units/sec 12000.000 units/min
note: AJOG     max: 1.000 units/sec 60.000 units/min
note: AJOG default: 1.000 units/sec 60.000 units/min
waiting for s.joints<0>, s.kinematics_type<0>
waiting for s.joints<0>, s.kinematics_type<0>
waiting for s.joints<0>, s.kinematics_type<0>
waiting for s.joints<0>, s.kinematics_type<0>
waiting for s.joints<0>, s.kinematics_type<0>
waiting for s.joints<0>, s.kinematics_type<0>
waiting for s.joints<0>, s.kinematics_type<0>
waiting for s.joints<0>, s.kinematics_type<0>
waiting for s.joints<0>, s.kinematics_type<0>
A configuration error is preventing LinuxCNC from starting.
More information may be available when running from a terminal.
Note: Using POSIX realtime
    /home/cnc/linuxcnc_debug.txt
and
    /home/cnc/linuxcnc_print.txt
```

BsAtHome:
> Current master fails running max5kins.ini. Worked before this merge. Looks like a INI-parse issue.
> emcTrajSetUnits failed to set [TRAJ]LINEAR_UNITS or [TRAJ]ANGULAR_UNITS
> emcMotionInit: emcTrajInit failed

The problem is that in `configs/sim/axis/vismach/5axis/max5/max5_base.inc` you have the following:
```ini
[TRAJ]
LINEAR_UNITS = 1.0
ANGULAR_UNITS = 1.0
```

These entries are wrong. Both LINEAR_UNITS and ANGULAR_UNITS are <i>enumerated types</i> and are not a numerical value.

See #3935, which describes the situation in detail.

BsAtHome:
And, BTW, using a quick grep, also wrong in:
* configs/by_interface/parport/plasma-thc-sim/plasma_mm.ini
* configs/by_interface/parport/stepper-gantry/stepper_mm.ini
* configs/by_interface/pluto/lathe-pluto/lathe-pluto.ini
* configs/sim/axis/vismach/5axis/max5/max5_base.inc
* configs/sim/axis/axis_9axis_scurve.ini
* configs/sim/axis/axis_mm_scurve.ini

They all need to be fixed.

hansu:
Ah ok I didn't notice that this config also uses those wrong entries.

andypugh:
Does this fix https://github.com/LinuxCNC/linuxcnc/issues/991 ?

BsAtHome:
I do think so, yes.

The new parser will convert the number and bug the user (warning message to console stderr) that there is trailing text. However, the converted value is still representative of that was there in the first place. Thus `1.234xyz` will return 1.234 and issue a warning. However, if there is a space after the number, then all trailing text is simply ignored.

If the value is invalid then the conversion fails and the returns is as if the variable was not found. An error message will be printed to console, depending on how the conversion was initiated. For variable searches that fail conversion, an error will be emitted. Only if you call the conversion API directly (rarely happens), then only warnings are emitted.

hansu:
> Does this fix #991 ?

It depends on where you put it and which GUI do you use. Like #991 it seems to be no problem for axis, but for the other parser(s) used in touchy and gmoccapy e.g.

Tested some cases for gmoccapy:

No problem:
```ini
[AXIS_X]
MIN_LIMIT = -400.0 # comment
MAX_LIMIT = 400.0
MAX_VELOCITY = 166
MAX_ACCELERATION = 1500.0
```

This
```ini
[TRAJ]
COORDINATES =      X Y Z
LINEAR_UNITS =     mm # comment
ANGULAR_UNITS =    degree
DEFAULT_LINEAR_VELOCITY = 35
MAX_LINEAR_VELOCITY = 234
POSITION_FILE =    position.txt
```
gives just this: 
```
LINUXCNC - 2.10.0~pre1
Machine configuration directory is '/home/cnc/linuxcnc-master/configs/sim/gmoccapy'
Machine configuration file is 'gmoccapy.ini'
/home/cnc/linuxcnc-master/configs/sim/gmoccapy/gmoccapy.ini: error: [TRAJ]LINEAR_UNITS: Must be one of [mm,metric,in,inch,imperial]
linuxcnc_check_ini validation failed
    /home/cnc/linuxcnc_debug.txt
and
    /home/cnc/linuxcnc_print.txt
```


And this
```ini
[TRAJ]
COORDINATES =      X Y Z  # comment
LINEAR_UNITS =     mm
ANGULAR_UNITS =    degree
DEFAULT_LINEAR_VELOCITY = 35
MAX_LINEAR_VELOCITY = 234
POSITION_FILE =    position.txt
```

gives this

```
[Gmoccapy.COMMON.INIINFO][CRITICAL]  MISSING [AXIS_#] MAX VELOCITY or MAX ACCELERATION entry in INI file. (iniinfo.py:202)
[Gmoccapy.COMMON.INIINFO][CRITICAL]  MISSING [AXIS_C] MAX VELOCITY or MAX ACCELERATION entry in INI file. (iniinfo.py:202)
[Gmoccapy.COMMON.INIINFO][CRITICAL]  MISSING [AXIS_O] MAX VELOCITY or MAX ACCELERATION entry in INI file. (iniinfo.py:202)
[Gmoccapy.COMMON.INIINFO][CRITICAL]  MISSING [AXIS_M] MAX VELOCITY or MAX ACCELERATION entry in INI file. (iniinfo.py:202)
[Gmoccapy.COMMON.INIINFO][CRITICAL]  MISSING [AXIS_M] MAX VELOCITY or MAX ACCELERATION entry in INI file. (iniinfo.py:202)
[Gmoccapy.COMMON.INIINFO][CRITICAL]  MISSING [AXIS_E] MAX VELOCITY or MAX ACCELERATION entry in INI file. (iniinfo.py:202)
[Gmoccapy.COMMON.INIINFO][CRITICAL]  MISSING [AXIS_N] MAX VELOCITY or MAX ACCELERATION entry in INI file. (iniinfo.py:202)
[Gmoccapy.COMMON.INIINFO][CRITICAL]  MISSING [AXIS_T] MAX VELOCITY or MAX ACCELERATION entry in INI file. (iniinfo.py:202)

  File "/home/cnc/linuxcnc-master/bin/gmoccapy", line 6551, in <module>
    import gladevcp.makepins
  File "/home/cnc/linuxcnc-master/lib/python/gladevcp/__init__.py", line 1, in <module>
    from .hal_pythonplugin import *
  File "/home/cnc/linuxcnc-master/lib/python/gladevcp/hal_pythonplugin.py", line 17, in <module>
    from .status_label import Status_Label
  File "/home/cnc/linuxcnc-master/lib/python/gladevcp/status_label.py", line 27, in <module>
    from gladevcp.core import Status as GStat
  File "/home/cnc/linuxcnc-master/lib/python/gladevcp/core.py", line 66, in <module>
    from gladevcp.gtk_action import _Lcnc_Action as _ActionParent
  File "/home/cnc/linuxcnc-master/lib/python/gladevcp/gtk_action.py", line 20, in <module>
    INFO = Info()
           ^^^^^^
  File "/home/cnc/linuxcnc-master/lib/python/common/iniinfo.py", line 56, in __init__
    self.update()
  File "/home/cnc/linuxcnc-master/lib/python/common/iniinfo.py", line 211, in update
    axisnum = "XYZABCUVW".index(let)
              ^^^^^^^^^^^^^^^^^^^^^^
ValueError: substring not found
 (gmoccapy:75)
```

BsAtHome:
All INI-file entries that are numerical (floating point or integer) will be parsed properly and all trailing context is ignored, as long as it is separated by one or more spaces before a non-space character is encountered. This is compatible with other places in LCNC where such lines are parsed.

All other entries are <i>text</i> based, including all enumerated types, and will have the '#' and all that follows be part of that text. That happens because comments cannot be part of the value. These entries will fail to parse because they are text-matched.

And that you get presented with an unhandled exception is a problem of the python code, not the INI-file parser. Some of these errors are caught by the pre-flight ini-file check, but not everything is caught in there.

Issue #2331, opened 2023-02-09
Hostmot2 encoder counts and HAL encoder counts are 32 bits and can wrap. 


andypugh:
The stepgens hold a step accumulator internally in an s64 (long long) and the smart-serial encoder counter does the same. 

HAL now supports 64-bit pins, so we could simply change the type of counts and rawcounts pins, but this is likely to break existing configs. 

Possibly the answer is to create counts64 and rawcounts64 pins, use those to calculate (64-bit float) position and then truncate the result into the existing pins. 

Or we could simply hide the internal accumulator, like the smart-serial driver does:

andypugh:
See https://forum.linuxcnc.org/38-general-linuxcnc-questions/48181-correct-handling-of-a-wrapping-s32-number-in-a-custom-comp#264182 for the context that brought this up.

petterreinholdtsen:
Staying backwards compatible for a few releases seem like a good idea,
perhaps even with a runtime warning if the old 32-bit counters are used.
This would mean new names for the 64 bit values, perhaps 'totalcounts'
something similar instead of adding type information in the name?

-- 
Happy hacking
Petter Reinholdtsen

pcw-mesa:
Do utilities like halshow know how to display 64 bit integers?

andypugh:
> Do utilities like halshow know how to display 64 bit integers?

Probably not, I don't see any changes to them in this commit:
https://github.com/LinuxCNC/linuxcnc/commit/7a4c0908567e27f034c7cb2fe2f6c5b1341c933c

hansu:
Yes it does not:
![grafik](https://github.com/LinuxCNC/linuxcnc/assets/67957916/2cdaf81b-bcf7-4db3-863f-aa0fa3efa9bb)
But the problem is here that even halcmd cannot handle 64 bit types:
 ```
$ halcmd show pin abs-s64
Component Pins:
Owner   Type  Dir         Value  Name
    40  undef IN      undef      abs-s64.0.in
    40  bit   OUT         FALSE  abs-s64.0.is-negative
    40  bit   OUT         FALSE  abs-s64.0.is-positive
    40  undef OUT     undef      abs-s64.0.out
    40  bit   OUT         FALSE  abs-s64.0.sign
    40  s32   OUT           434  abs-s64.0.time

$ halcmd getp abs-s64.0.out
unknown_type
```

hansu:
[...]

PR #3932, opened 2026-04-12
Expand hardware drivers to 64 bits (address issue #2331)

andypugh:
Mainly a re-working of the register wrapping to 64 bits. 
Note that this does not yet include Hostmot2, though that has already been at least partially patched in https://github.com/LinuxCNC/linuxcnc/commit/8a3ac3f336129c489b825a3253132ee14304e09a


src/hal/drivers/hal_gm.c
```diff
@@ -1,9 +1,10 @@
 #include <rtapi_pci.h>
 #include <rtapi_io.h>
 
-#include <rtapi.h>		// RTAPI realtime OS API
-#include <rtapi_app.h>		// RTAPI realtime module decls
-#include <hal.h>		// HAL public API decls
+#include "rtapi.h"		// RTAPI realtime OS API
+#include "rtapi_app.h"		// RTAPI realtime module decls
+#include "rtapi_stdint.h"	// portable ints
+#include "hal.h"		// HAL public API decls
```
BsAtHome: These need to be <> includes.

src/hal/drivers/hal_motenc.c
```diff
@@ -124,9 +125,10 @@
 
 #include <linux/pci.h>
 
-#include <rtapi.h>			// RTAPI realtime OS API.
-#include <rtapi_app.h>			// RTAPI realtime module decls.
-#include <hal.h>			// HAL public API decls.
+#include "rtapi.h"			// RTAPI realtime OS API.
+#include "rtapi_app.h"			// RTAPI realtime module decls.
+#include "rtapi_stdint.h"		// rtapi_s32 etc
+#include "hal.h"			// HAL public API decls.
```
BsAtHome, line 131: These need to be <> includes.

src/hal/drivers/hal_ppmc.c
```diff
@@ -73,10 +73,11 @@
...
 #include <rtapi_io.h>		/* kmalloc() */
-#include <rtapi.h>		/* RTAPI realtime OS API */
-#include <rtapi_app.h>		/* RTAPI realtime module decls */
-#include <hal.h>		/* HAL public API decls */
-#include <rtapi_parport.h>
+#include "rtapi.h"		/* RTAPI realtime OS API */
+#include "rtapi_app.h"		/* RTAPI realtime module decls */
+#include "rtapi_stdint.h"	// portable ints
+#include "hal.h"		/* HAL public API decls */
+#include "hal_parport.h"
```
BsAtHome, line 80: Again <> includes.

Except, hal_parport.h is not accessible in the driver directory and was eliminated by replacing the calls to rtapi calls.

src/hal/drivers/hal_ppmc.c
BsAtHome, line 1087: A rtapi_s64 has 8 bytes, not 4.

src/hal/drivers/hal_ppmc.c
```diff
@@ -1106,16 +1108,16 @@ static void read_encoders(slot_data_t *slot)
...
 	pos.byte.b2 = (signed char)slot->rd_buf[byteindex++];
-        pos.byte.b3 = oldpos.byte.b3;
         /* check for - to + transition */
         if ((oldpos.byte.b2 & 0xc0) == 0xc0 && (pos.byte.b2 == 0))
-            pos.byte.b3++;
+            pos.l += 0x01000000;
         else
             if ((oldpos.byte.b2 == 0) && (pos.byte.b2 & 0xc0) == 0xc0)
-                pos.byte.b3--;
+                pos.l -= 0x01000000;
```
BsAtHome, line 1120: Byte-testing is a bad option. Better to look at the value as a whole and mask unwanted bits using a binary and (&). Then the compiler can sort it out properly.

src/hal/drivers/hal_stg.c
```diff
@@ -616,7 +624,7 @@ static void stg_counter_capture(void *arg, long period)
 		stg->pos_scale[n] = 1.0;
 	}
 	/* scale count to make floating point position */
-	*(stg->pos[n]) = *(stg->count[n]) / stg->pos_scale[n];
+	*(stg->pos[n]) = *(stg->count64[n]) / stg->pos_scale[n];
```
BsAtHome, line 627: Writing a hal pin and reading the same pin just a bit later is very inefficient because the compiler must physically read the value from memory. The volatile modifier prevents caching. However, the value does not change here between the two lines. Therefore, a local intermediary allows the compiler to be more efficient.

This issue is seen all over this and other code, writing and then shortly thereafter again reading same pin.

src/hal/drivers/hal_ppmc.c
```diff
@@ -1106,16 +1108,16 @@ static void read_encoders(slot_data_t *slot)
     for (i = 0; i < 4; i++) {
       slot->encoder[i].indrescnt++;  /* increment counter each servo cycle */
         oldpos.l = slot->encoder[i].oldreading;
+	pos.l = *(slot->encoder[i].count); // init the higher-order bytes
 	pos.byte.b0 = (signed char)slot->rd_buf[byteindex++];
 	pos.byte.b1 = (signed char)slot->rd_buf[byteindex++];
 	pos.byte.b2 = (signed char)slot->rd_buf[byteindex++];
-        pos.byte.b3 = oldpos.byte.b3;
         /* check for - to + transition */
         if ((oldpos.byte.b2 & 0xc0) == 0xc0 && (pos.byte.b2 == 0))
```
BsAtHome, line 1116: What exactly is the counter's width and signedness? Masking two bits that must be set looks strange. If this two's complement, one's complement or another format?

BsAtHome:
There are probably some more issues that need to be addressed, but my first list is probably a start ;-)

BsAtHome:
BTW, the CI fail is due to a warning from the ini-file merge. That just got a fix which I merged in #3931. Rebasing/FF your code should fix that.

src/hal/drivers/hal_gm.c
```diff
@@ -1472,7 +1477,7 @@ encoder(void *arg, long period)
     card	*pCard = device->pCard;
 
     int		i;
-	hal_s32_t	temp1 = 0, temp2;
+	rtapi_s32	temp1 = 0, temp2, delta_counts;
 	hal_float_t	vel;
```
BsAtHome, line 1481: Local variables must not use HAL types.

src/hal/drivers/hal_gm.c
BsAtHome, line 1520: Wrong type in cast

BsAtHome, line 1524: Wrong type in cast

src/hal/drivers/hal_gm.c
```diff
@@ -1529,12 +1534,15 @@ encoder(void *arg, long period)
 		  }
 		}
 		device->encoder[i].last_index_latch = temp2;
-		
-		*(device->encoder[i].rawcounts) =  temp1 - device->encoder[i].raw_offset;
+		delta_counts = temp1 - device->encoder[i].old_counts;
+		device->encoder[i].old_counts = temp1;
+		*(device->encoder[i].rawcounts) += delta_counts;
+		*(device->encoder[i].rawcounts64) += delta_counts;
```
BsAtHome, line 1540: This looks fine. You do:
```C
s32_deltacount = s32_counter - s32_saved_counter;
s32_saved_counter = s32_counter;
s64 += s32_deltacount;
```
I assume that the s32_counter is two-s complement. However, it should also work for an unsigned counter because the deltacount should always be result in correct two's complement.

BsAtHome: The only assumption is that you cannot wrap a 32-bit counter in <i>one</i> servo period. But I guess we should be save there?

src/hal/drivers/hal_ppmc.c
BsAtHome, line 1104: The `timebase` variable seems to be a 16-bit value and needs to be read from a stream in an array(`rd_buf[]`). The data looks to be in little endian (LSB in the lower address). We are running this code on a little endian machine. Then this method seems better (easier for the compiler to optimize):
```C
unsigned short val;
val = slot->rd_buf[byteindex + 1] & 0xff;
val <<= 8;
val += slot->rd_buf[byteindex + 0] & 0xff;
// The byteindex seems to be overwritten subsequently,
// so this line may be dead code
byteindex += 1; // Only increments once in the code...
```

Alternate formulation:
```C
val = (slot->rd_buf[byteindex + 0] & 0xff) + ((slot->rd_buf[byteindex + 1] & 0xff) << 8);
// Same comment...
byteindex += 1; // Only increments once in the code...
```

src/hal/drivers/hal_ppmc.c
```diff
@@ -1106,16 +1108,16 @@ static void read_encoders(slot_data_t *slot)
     for (i = 0; i < 4; i++) {
       slot->encoder[i].indrescnt++;  /* increment counter each servo cycle */
         oldpos.l = slot->encoder[i].oldreading;
+	pos.l = *(slot->encoder[i].count); // init the higher-order bytes
 	pos.byte.b0 = (signed char)slot->rd_buf[byteindex++];
 	pos.byte.b1 = (signed char)slot->rd_buf[byteindex++];
 	pos.byte.b2 = (signed char)slot->rd_buf[byteindex++];
-        pos.byte.b3 = oldpos.byte.b3;
```
BsAtHome, line 1112: Deletion of `b3` leaves it undefined and at a random value.

src/hal/drivers/hal_ppmc.c
```diff
@@ -1106,16 +1108,16 @@ static void read_encoders(slot_data_t *slot)
     for (i = 0; i < 4; i++) {
       slot->encoder[i].indrescnt++;  /* increment counter each servo cycle */
         oldpos.l = slot->encoder[i].oldreading;
+	pos.l = *(slot->encoder[i].count); // init the higher-order bytes
 	pos.byte.b0 = (signed char)slot->rd_buf[byteindex++];
 	pos.byte.b1 = (signed char)slot->rd_buf[byteindex++];
 	pos.byte.b2 = (signed char)slot->rd_buf[byteindex++];
```
BsAtHome, line 1114: You are reading a value from a stream. The better method is to shift the value in byte-by-byte into the larger type instead of using a unonized type.

If the value you stream-read is a two's complement value, then there is a second advantage. You can do proper sign extension in one go by casting the MSB to a signed char and then to the signed larger type. Shifting left will preserve the sign.

src/hal/drivers/hal_gm.c
andypugh: I think so. The Mesa counters max out at 20MHz, the internet suggests that 500MHz is possible. 
Wrapping 32 bits in 1ms is a few Teraherz.

BsAtHome: At 500MHz you have a comfortable 8.5 <i>seconds</i> before you wrap :-)

src/hal/drivers/hal_ppmc.c
andypugh: The value read from the FPGA is an unsigned 16 bit number. So the positive wrap looks like:
0xFFFF
0x0000
0x0001
...
And the reverse wrap is
0x0001
0x0000
0xFFFF
....

andypugh:
Note that in all cases I have attempted to touch the existing code as little as possible. I am not sure that fixing long-standing inelegancies is within the scope of this change.

BsAtHome:
> Note that in all cases I have attempted to touch the existing code as little as possible. I am not sure that fixing long-standing inelegancies is within the scope of this change.

Inelegancies besides, there are a lot of problems with the code. "Just touching the counter wrap" may actually break stuff in ways you didn't imagine because of the way it is written. The code is (very) fragile in my opinion. I see it as a pest verses cholera choice, you choose how to die ;-)

src/hal/drivers/hal_ppmc.c
BsAtHome: But that means a simple unsigned or two's complement signed 16 bit value. Then the proper procedure would be the same as for the 32 bit case, but simply with a 16 bit sign-extended starting point:
```C
rtapi_s32 oldval = _initval_;
rtapi_s64 bigcounter = _initval_;
...
rtapi_s32 newval = (((rtapi_s32)(signed char)MSB) << 8) + (LSB & 0xff);
rtapi_s32 delta = oldval - newval;
oldval = newval;
bigcounter += delta;
```

The first MSB cast makes the value a signed type and then the next rtapi_s32 cast forces sign-extension.

And the assumption is that it does not wrap within a servo period (should be safe with just over 65 MHz at 1ms servo thread).

rmu75:
Given a 16bit hardware counter, suppose i have a reading of 16383, followed by a reading of -16384, how is the code supposed to know if a wrap occured in this situation or not? Obviously that wrapping can only work if the maximum allowed change between sampling is less than half of the total encoder counts. That should probably be documented somewhere, although it is probably unlikely to occur on real hardware.

src/hal/drivers/hal_ppmc.c
rmu75: That could be interpreted as signed two's complement or unsigned.

BsAtHome:
> Given a 16bit hardware counter, suppose i have a reading of 16383, followed by a reading of -16384, how is the code supposed to know if a wrap occured in this situation or not?

That is a very good observation. The code has a problem at half the range disregarding directional information.

What about this:
```C
// your example
rtapi_s32 oldval = 16383;
rtapi_s32 newval = -16384;
rtapi_s32 delta = newval - oldval; // delta is now -32767 (within range)

// Full 16 bit range forward
rtapi_s32 oldval = 32767;
rtapi_s32 newval = -32768; // moved one position
rtapi_s32 delta = newval - oldval; // delta is now -65535, should be +1

// Full 16 bit range reverse
rtapi_s32 oldval = -32768;
rtapi_s32 newval = 32767; // moved one position
rtapi_s32 delta = newval - oldval; // delta is now +65535, should be -1

// half-way problem correction
if(delta > 32768)  // -32768 -> 0 is valid
    delta -= 65536;
else if(delta < -32768) // 0 -> -32768 is valid
    delta += 65536;
```

It also means that 32767 -> -1 (or 32766 -> -2) will become -32768.

rmu75:
I don't like that bit manipulation stuff. IMHO it would be better to make that stuff explicit. Then we also would not have to assume anything about endianness or how exactly stuff is represented and could deal with arbitrary number of max counts, e.g. 18 bit absolute encoders.

lets assume a counter with 4000 counts going from 0 to 3999. midpoint would be 1999.  

calculate delta of new value and old value, this would be in the range [-3999 to +3999]

wrap around occured if new value <= 1999 and old value > 1999 (or vice versa) and abs(delta) >= 2000. 

in case abs(delta) < 2000 no wraparound. else

in case new < old, the delta will be negative, e.g. new = 100, old = 3900, delta = -3800. The correct delta in this case would be +200, so we need to add 4000.

in case old < new, the delta will be positive, e.g. old = 3900, new = 100, delta = 3800. The correct delta in this case would be -200, so we need to subtract 4000.

that same logic should apply no matter if counts are treated as signed or unsigned. the delta should be calculated in something large enough.

andypugh:
> I don't like that bit manipulation stuff. IMHO it would be better to make that stuff explicit.

Bear in mind that this is the PPMC driver, which gets the bytes one at a time through the EPP parallel port. 
Endian-ness isn't an issue in that regard. 

~~And the value is definitely unsigned 16 bit, this has been confirmed with Jon @ Pico Systems.~~
Sorry, I misremembered, it is _24_ bit, unsigned. But definitely unsigned. 

In the case of the other drivers, the values captured from the hardware _are_ signed, and just doing "counts += delta_counts" works without any conditional execution even through the wrap, with the correctly-sized signed datatypes.

andypugh:
I discussed the wrap-management techniques with Jeff Epler during the work on this PR. 

"Here's the implementation from pluto, which takes a bit count argument

```
static inline int64_t extend(int64_t old, int newlow, int nbits) {
    int64_t mask = (1<<nbits) - 1;
    int64_t maxdelta = mask / 2;
    int64_t oldhigh = old & ~mask;
    int64_t oldlow = old & mask;
    int64_t candidate1, candidate2;

    candidate1 = oldhigh | newlow;
    if(oldlow < newlow) candidate2 = candidate1 - (1<<nbits);
    else                candidate2 = candidate1 + (1<<nbits);

    if (llabs(old-candidate1) > maxdelta)
        return candidate2;
    else
        return candidate1;
}
```
Way back in the day I tested it to my satisfaction. However it has a conditional branch in it which is not ideal.

It's better to do like you suggest, computing a delta value .. the trick to do it on an odd size like 24 is to *shift the bits up so that the 24 bits are at the top of the integer* to find the correct delta value from the new low bits.. I have to credit my time working with MicroPython for teaching me this idea, they need to do it for efficient math on 30 and 31 bit signed values.


```
#include <stdint.h>
#include <limits.h>
#include <assert.h>

#ifdef __CPROVER__
#define assume(x) __CPROVER_assume((x))
#define assert(x) __CPROVER_assert((x), #x)
#else
#define assume(x)
#include <assert.h>
#endif

int32_t nondet_s32();
int64_t nondet_s64();

static inline int64_t extend(int64_t old, int newlow, int nbits) {
    int nshift = 64 - nbits;
    uint64_t oldlow_shifted = ((uint64_t)old << nshift);
    uint64_t newlow_shifted = ((uint64_t)newlow << nshift);
    int64_t diff_shifted = newlow_shifted - oldlow_shifted;
    return (uint64_t)old + (diff_shifted >> nshift);
}

int main() {
    int nbits = nondet_s32();
    assume(nbits > 8 && nbits < 32);

    int mask = UINT_MAX >> (32 - nbits);
    int64_t old = nondet_s64();

    int delta = nondet_s32() & (mask >> 1); // Can handle deltas up to this big

    int64_t new1 = (uint64_t)old + (uint64_t)delta;
    int64_t res1 = extend(old, new1 & mask, nbits);
    assert(new1 == res1);

    int64_t new2 = (uint64_t)old - (uint64_t)delta;
    int64_t res2 = extend(old, new2 & mask, nbits);
    assert(new2 == res2);
}
```
This program is an input to `cbmc`  which proves that the assertions hold for _ALL_ values that match the assume() clauses. And this version is branch free and free of undefined behavior due to signed integer overflow: `cbmc --signed-overflow-check extend2.c`

Double check it for yourself because it's late here on a holiday night and I shouldn't be writing low level C code. But you successfully nerd-sniped me."

rmu75:
hmm. am i missing something? in case of two's complement the sign is convention. in 24 bit signed 8388607 + 1 wraps to -8388608, in unsigned it becomes +8388608, both have representation 10000000 00000000 00000000. extension to 32bit would differ though in the uppermost byte.

(claification: only considering addition/subtraction)

rmu75:
The first thing from jeff epler seems to be exactly what I had in mind. It is a bit odd that he calls 24 odd ;-) The second thing is quite clever, shifting avoids the need for sign extension...

I wonder given MicroPython targets microcontrollers if a branch on such architectures is really worse than 3 shifts... The problem with clever code is it is clever ;) and 5 years from now nobody knows anymore.

andypugh:
> The second thing is quite clever, shifting avoids the need for sign extension...

The conversation started with me asking him if it was a dependable solution. 
I have had to handle this more than a few times, writing the various hardware drivers. 

> The problem with clever code is it is clever ;) and 5 years from now nobody knows anymore.

Maybe, but with adequate comments, it should be fairly clear. 
The fun is when you don't know the bit-width at compile time (like with smart-serial) and have to deal with it at runtime.

rmu75:
On Mo, 2026-04-13 at 14:31 -0700, andypugh wrote:
> Maybe, but with adequate comments, it should be fairly clear.
> The fun is when you don't know the bit-width at compile time (like
> with smart-serial) and have to deal with it at runtime.

don't understand -- both functions take the bit width as argument?

we should put this extend function into hal lib.

andypugh:
> we should put this extend function into hal lib.

This was also a suggestion from Jeff. 

However, I don't think that it would help with ppmc and its 24-bit unsigned counter.

In the function shown, why is "old" cast to uint before the addition?

rmu75:
> However, I don't think that it would help with ppmc and its 24-bit
> unsigned counter.

I'm feeling stupid. Maybe I'm missing something obvious, but I don't think there is any meaningful distinction between a "signed" counter and an "unsigned" counter  (assuming two's complement, but that is mandatory with C23 and hardware counting in signed one's complement would be very stupid). -- the bit patterns and their sequence are exactly the same.

example of a 4 bit counter:

| bit pattern  | as unsigned | as signed |
| ------------- | ------------ | -------- |
| 0000 | 0 wrap to 15 | 0|
| 0001 | 1 | 1 |
| 0010 | 2 | 2 |
| 0011 | 3 | 3 |
| 0100 | 4 | 4 |
| 0101 | 5 | 5 |
| 0110 | 6 | 6 |
| 0111 | 7 | 7  wrap to -8 |
| 1000 | 8 | -8 wrap to 7 |
| 1001 | 9 | -7 |
| 1010 | 10 | -6 |
| 1011 | 11 | -5 |
| 1100 | 12 | -4 |
| 1101 | 13 | -3 |
| 1110 | 14 | -2 |
| 1111 | 15 wrap to 0 | -1 |

the only difference is where the "wrap" occurs. except for the wrap counting is strictly monotonous. 

counter at count "4" counting "5" steps up would result in "9" or "-7", counting "5" steps down would result in "15" or "-1" -- same bit pattern.

treating the 24bit number as signed and feeding it into jeff eplers extend function(s) should just work.

rmu75:
> In the function shown, why is "old" cast to uint before the addition?

I think to avoid overflowing a signed int which in C is undefined behaviour and can tickle the compiler the wrong way. E.g. some optimization steps (sometimes) remove everything that depends on the result of such undefined behaviour.

https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Signed-Overflow.html

PR #3936, opened 2026-04-14
add hal_extend_counter function (extend limited width counters to 64 bit)

rmu75:
helper function to deal with wrap around and extension of lower-width counters to 64bit ints.

### DIFF
--- a/src/hal/hal.h
+++ b/src/hal/hal.h
@@ -692,6 +692,7 @@ extern int hal_get_signal_value_by_name(
 extern int hal_get_param_value_by_name(
     const char *name, hal_type_t *type, hal_data_u **data);
 
+
 /***********************************************************************
 *                   EXECUTION RELATED FUNCTIONS                        *
 ************************************************************************/
@@ -982,6 +983,101 @@ extern bool hal_stream_writable(hal_stream_t *stream);
 #ifdef ULAPI
 extern void hal_stream_wait_writable(hal_stream_t *stream, sig_atomic_t *stop);
 #endif
+
+
+/***********************************************************************
+*                    MISC HELPER FUNCTIONS                             *
+************************************************************************/
+
+
+/** HAL_STATIC_ASSERT wrapper for compile time asserts
+*/
+#if __STDC_VERSION__ >= 202311L || __cplusplus
+#define HAL_STATIC_ASSERT(expression, message) static_assert((expression), message)
+#else
+/* _Static_assert: GCC extension, in C standard since C11, deprecated in favour of
+   static_assert (like C++) from C23 onwards*/
+#define HAL_STATIC_ASSERT(expression, message) _Static_assert((expression), message)
+#endif
+
+
+/** hal_extend_counter() extends a counter value with nbits to 64 bits.
+    
+    For some hardware counters or encoders it may be desireable to
+    extend their range beyond their native width.
+
+    This function maintains a 64bit counter value and counts wrap
+    arounds. It may be useful to e.g. keep track of full rotations on
+    a gray disc absolute encoder.
+    
+    Changes of hardware encoder between calls to this function need to
+    be less than 2**(nbits-1).
+
+    Usage:
+
+    Call with current 64bit counter value to be updated as @param old,
+    new low-width counter value read from hardware as @param newlow 
+    and width of counter as @param nbits.
+    @returns new 64bit counter value which should be stored and
+    supplied as old in the next call.
+     */
+static inline rtapi_s64 hal_extend_counter(rtapi_s64 old, rtapi_s64 newlow, int nbits)
+{
+    /* Extend low-bit-width counter value to 64bit, counting wrap arounds resp.
+       "rotations" in additional bits. 
+
+       see https://github.com/LinuxCNC/linuxcnc/pull/3932#issuecomment-4239206615
+       This code avoids branches and undefined behaviour due to signed overflow.
+       Idea from MicroPython.
+
+       The tricky part is how to efficiently calculate the difference between
+       two counter values that cross from minimum to maximum or vice versa.
+       (underflow/overflow).
+
+       To calculate this difference and avoid costly branches and explicit 
+       verbose code, the difference is calculated from values shifted to the
+       right, the MSB of the counter values are shifted to the MSB of the
+       processor registers (64 bit). This difference is automatically
+       truncated to 64bit and then shifted back right to its original position
+       while preserving the sign (arithmetic right shift). This difference
+       is added to the large counter value.
+       
+       The limit when using N-bit bit-width limited counters is that the maximum
+       count difference between subsequent invocation may not be larger than
+       2**(nshift-1)-1. Otherwise it is impossible to determine the direction of the
+       counter.
+
+       Heuristically, if your encoder can do more than half a rotation between
+       calls to this function, it is impossible to deduce which direction it
+       went.
+
+       Code contributed by Jeff Epler.
+
+       Example with 3bit absolute hardware encoder and 8bit extended counter
+       (nshift would be 5):
+
+       counter at 7, transition from 111 (7) to 000 (0) should increment the
+       extended counter to 8.
+
+       call hal_extend_counter(7, 0, 3)
+       oldlow_shifted = 7<<5 = 224 (1110 0000)
+       newlow_shifted = 0<<5 = 0
+       delta_shifted = 0 - 224 = -224 (1 0010 0000) = 32 (0010 0000) truncated to 8 bits 
+       old + (32 >> 5) = 7 + 1 = 8
+    */
+
+    /* tripwire if somebody tries to use this code on a Cray with wrong
+       compiler flags */
+    HAL_STATIC_ASSERT((-2 >> 1) == -1, 
+        "hal_extend_counter impl only works with arithmetic right shift");
+
+    int nshift = 64 - nbits;
+    rtapi_u64 oldlow_shifted = ((rtapi_u64)old << nshift);
+    rtapi_u64 newlow_shifted = ((rtapi_u64)newlow << nshift);
+    rtapi_s64 diff_shifted = newlow_shifted - oldlow_shifted;
+    return (rtapi_u64)old + (diff_shifted >> nshift); // unsigned to avoid signed overflow
+}
+
 
 RTAPI_END_DECLS
 


src/hal/hal_lib.c
```diff
@@ -1872,6 +1872,24 @@ int hal_get_param_value_by_name(
 }
 
 
+/***********************************************************************
+*                    MISC HELPER FUNCTIONS                             *
+************************************************************************/
+
+int64_t hal_extend_int(rtapi_s64 old, int newlow, int nbits)
```
BsAtHome: You cannot use "int64_t"

BsAtHome: And why is the "newlow" arg an int? Wouldn't it be more appropriate to use rtapi_s64?

src/hal/hal.h
```diff
@@ -692,6 +692,18 @@ extern int hal_get_signal_value_by_name(
...
+/***********************************************************************
+*                    MISC HELPER FUNCTIONS                             *
+************************************************************************/
+
+/** hal_extend_int() extends a counter value with nbits to 64 bits.
+    Increments between updates need to be less than 2**(nbits-1).
+    Call with current 64bit counter value to be updated, new
+    lower-bit counter value and number of bits of counter.
+    @returns new counter value.
+    Code by Jeff Epler. */
```
BsAtHome, line 710: What is the license?

src/hal/hal_lib.c
```diff
@@ -1872,6 +1872,24 @@ int hal_get_param_value_by_name(
...
+{
+    /* see https://github.com/LinuxCNC/linuxcnc/pull/3932#issuecomment-4239206615
+       This code avoids branches and undefined behaviour due to signed overflow
+       Idea from MicroPython.
```
BsAtHome: To remind you of https://github.com/LinuxCNC/linuxcnc/pull/3932#issuecomment-4239767880, You should have a detailed explanation of how this code works. Otherwise you end up exactly in the trouble that you warned about with "smart code".

src/hal/hal_lib.c
rmu75: > You cannot use "int64_t"

Missed this one

rmu75: > And why is the "newlow" arg an int? Wouldn't it be more appropriate to use rtapi_s64?

in the original patch nbits should be between 8 and 32. it should also work for larger and smaller counters, but won't make much sense imo.

src/hal/hal.h
rmu75: no idea. is this copyrightable in the first place?

src/hal/hal_lib.c
rmu75: yes i will put some more explanation.

src/hal/hal.h
BsAtHome: Yes it is.

src/hal/hal_lib.c
BsAtHome: What about 40 bit or 48 bit counters? You may not see them today, but I can't see a reason to create an artificial constraint.

src/hal/hal.h
rmu75: Then @andypugh needs to get perm from @jepler.

jepler: Hi! You're free to use this code under the GPL v2.0 or newer license. I wrote it with the intent that it would be contributed to LinuxCNC.

rmu75: Thanks.

BsAtHome:
The explanation does not tell us what the function is supposed to do, why it exists and how it accomplished the desired outcome. Neither does it describe arguments, like what values old, newlow and nbits represent.

Also, IMO, the name `hal_extend_int()` is a misnomer for what is actually performed.

rmu75:
> The explanation does not tell us what the function is supposed to do, why it exists and how it accomplished the desired outcome. Neither does it describe arguments, like what values old, newlow and nbits represent.
> 
> Also, IMO, the name `hal_extend_int()` is a misnomer for what is actually performed.

```c++
/** hal_extend_int() extends a counter value with nbits to 64 bits.
    Increments between updates need to be less than 2**(nbits-1).
    Call with current 64bit counter value to be updated, new
    lower-bit counter value and number of bits of counter.
    @returns new counter value.
    Code by Jeff Epler. */
```

hmm.  I guess my description presupposes some context. It could certainly be more verbose. Regarding the name, I didn't and still don't have a good idea how to name it. Feel free to suggest something. @andypugh opinion re. name?

BsAtHome:
> hmm. I guess my description presupposes some context. It could certainly be more verbose.

Yes, and that is exactly the problem you yourself pointed out and warned about when adding "smart code".


> Regarding the name, I didn't and still don't have a good idea how to name it. Feel free to suggest something. @andypugh opinion re. name?

What is its primary function? Usually, name follows function.

BsAtHome:
Just one more thing... This whole routine is being added because the usual algorithm has one or two conditionals in performing a fixup.

Then <i><b>why</b></i> is this being added as a <i>subroutine</i> involving two non-local jumps (call and ret), disregarding possible PLT indirection, and at least two stack operations, assuming the arguments are register passed? I fail to see the advantage in that.

rmu75:
I think the "cleverness" is adequately addressed in the code comment.

I propose this code to be added to have one (and only one) correct implementation that can deal with all cases, not to avoid branches or whatever. Correcctness trumps speed.

Unconditional jumps are cheap. I don't think a macro is warranted here, neither pollution of the header file with implementation of an inline function or a ".inc" code include file.

BsAtHome:
> I think the "cleverness" is adequately addressed in the code comment.

We apparently have very different ideas of "adequately addressed".


> I propose this code to be added to have one (and only one) correct implementation that can deal with all cases, not to avoid branches or whatever. Correcctness trumps speed.

I can accept correctness, in code. But I fail to see that the implementation is actually correct in the greater context.

> Unconditional jumps are cheap.

No, they are not cheap. These are not simple jumps and involve stack frames. If this code is PLT indirected, then there is one more indirect jump involved too.

> I don't think a macro is warranted here, neither pollution of the header file with implementation of an inline function or a ".inc" code include file.

It shouldn't be a macro as such. It should be an <i>inline</i> function that is always inlined. Then the compiler has a chance of performing optimizations and is not forced to create and tear down a stack frame construct (requiring flushing local register scheduling).

src/hal/hal.h
```diff
@@ -692,6 +692,24 @@ extern int hal_get_signal_value_by_name(
...
+
+    This function maintains a 64bit counter value and deals with wrap
+    arounds. Increments between updates need to be less than 2**(nbits-1).
+    Call with current 64bit counter value to be updated as @param old,
+    new lower-bit counter value read from hardware as @param newlow 
+    and width of counter as @param nbits.
+    @returns new counter 64bit counter value.
+    Code by Jeff Epler. */
+extern rtapi_s64 hal_extend_int(rtapi_s64 old, rtapi_s64 newlow, int nbits);
+
```
BsAtHome, line 712: The prototype is effectively superfluous unless you try to link in a system that does both not inline and not define that static function locally as static. But, since we are using a C compiler that actually does inline, I do not see the point of the prototype.

Or is there something that I missed?

src/hal/hal.h
```diff
@@ -983,6 +1001,41 @@ extern bool hal_stream_writable(hal_stream_t *stream);
...
+       To avoid messy code to sign extend a lower-width value to 64bits, shift
+       the counter value such that the sign bit gets into the MSB.
+       Calculate the delta from the stored value in the shifted base.
+       Shifting back the delta assures the sign is extended properly.
+       Addition as unsigned avoids signed overflow, which would be undefined
+       behaviour.
+
+       Attention has to be paied that the magnitude of increments / decrements
+       of the counter stay within 1<<(nshift-1) between calls to this function,
+       else the wrap around will be counted in the wrong direction.
```
BsAtHome, line 1027: You may think that I am pedantic here, but I'm just reading from a blanked mind (like one from 5 years in the future) and try to fit that into why I would need this function.

It still does not explain what the arguments `old` and `newlow` actually are in the context of extending a counter. Which of these arguments is the counter?
The names suggest that there is something "old", but it is unclear what is old. There also is something "new" and apparently also "low" at the same time. But I cannot fathom, from the description, how these names fit into the algorithm.

BTW, the name hal_extend_counter() does seem much more appropriate, but there is more going on than just extending the counter. It is calculating a counter increment from the difference of the arguments and adding that difference to one of the arguments. This exploits the behaviour of unsigned verses two's complement calculations. It uses smaller values as source into a larger register set as if the calculation was performed on a "nbits" register in an unsigned operation (exploiting overflow behaviour) casted to a two's complement result. The final calculation is furthermore (ab)using the unsigned/signed duality. How exactly is this an "easy" algorithm to follow?

rmu75: See comment next to prototype. This comment here in the impl explains what the code is doing and is inside. IDE will display the other comment if oyu hover over a function call.

In fact it doesn't matter if you work with signed or unsigned counter values, the important thing is that the difference has the proper sign and is calculated with width nbits. instead of figuring out the high bits, shifting it up gets proper overflow behaviour and the "extra (0) bits" towards LSB don't matter.

rmu75: before i force push and trigger yet another senseless CI run what about

```
    /* Extend low-bit-width counter value to 64bit, counting wrap arounds resp.
       "rotations" in additional bits. 

       see https://github.com/LinuxCNC/linuxcnc/pull/3932#issuecomment-4239206615
       This code avoids branches and undefined behaviour due to signed overflow
       Idea from MicroPython.

       To avoid messy code to sign extend and overflowa lower-width value to
       64bits, this code shifts the counter value to the end of a 64bit int.
       Thus the calculated delta value overflows correctly and gets the correct
       sign.

       Shifting back the delta assures the sign is extended properly.
       Addition as unsigned avoids signed overflow, which would be undefined
       behaviour.

       Attention has to be paid that the magnitude of increments / decrements
       of the counter stay within 1<<(nshift-1) between calls to this function,
       else the wrap around will be counted in the wrong direction.

       Heuristically, if your encoder can do more than half a rotation between
       calls to this function, it is impossible to deduce which direction it
       went.

       Code contributed by Jeff Epler.

       Example with 3bit absolute hardware encoder and 8bit extended counter
       (nshift would be 5):

       counter at 7, transition from 111 (7) to 000 (0) should increment the
       extended counter to 8.

       call hal_extend_counter(7, 0, 3)
       oldlow_shifted = 7<<5 = 224 (1110 0000)
       newlow_shifted = 0<<5 = 0
       delta_shifted = 0 - 224 = -224 (1 0010 0000) = 32 (0010 0000) truncated to 8 bits 
       old + (32 >> 5) = 7 + 1 = 8
    */
```

BsAtHome: My suggestion:
hal_extend_counter(): Calculate the actual counter value from one reading to the next while compensating for counter wrapping.
Counters with a limited bit-width will wrap while counting up or down. You need an unwrapped counter value to calculate the actual counter difference between readings. This function calculates the new reading's actual value based on the current counter value (newlow), the bit-width (nbits) of the counter and the previous (old) counter value. This algorithm uses a branchless approach.

For any set of bits in a counter, we extend its width to 64 bits by moving the most significant bit of the counter into the most significant bit of the processor register. This creates a counter as if it were a full register's width and the difference between the counter positions will have the proper sign, as calculated by the processor. The difference is then corrected back, preserving sign, to be of the original counter's bit-width. This calculated difference is added to the old counter value and returned as the actual new counter's value. The returned value thus mitigates any counter wrapping and preserves counting direction.

The limit when using N-bit bit-width limited counters is that the maximum count difference between subsequent invocation may not be larger that 2**(N-1)-1. Otherwise it is impossible to determine the direction of the counter.


Other comments:

> Shifting back the delta assures the sign is extended properly.

That is very poorly defined behaviour. Right shifting may use arithmetic shift or logic shift. Depending on implementation specific behaviour at this level may not be wise. Right shifting signed values is generally bad practice and should be replaced by division (`diff_shifted >> nshift` => `diff_shifted / (1LL << nshift)`).

> Addition as unsigned avoids signed overflow, which would be undefined behaviour.

You are saying that "signed overflow is undefined" is problematic and asserting that "unsigned overflow is defined". They are both problematic. On most architectures (all in practice) they work exactly the same way because there is only one adder. It is the interpretation of status flags afterwards that introduces the problem.

Replacing the code with only signed values produces exactly the same result:
```C
    rtapi_s64 oldlow_shifted = old << nshift;
    rtapi_s64 newlow_shifted = newlow << nshift;
    rtapi_s64 diff_shifted = newlow_shifted - oldlow_shifted;
    return old + (diff_shifted >> nshift);
```
The reason for this to work is that both the signed and unsigned 64-bit values behave the same in addition and subtraction, as well as left shift. The code was created to be branch-free and it is only in conditional branches that the difference between signed and unsigned becomes potentially problematic. Therefore, I fail to see the argument in using casts to unsigned.

BsAtHome: BTW, I think that `hal_unwrap_counter()` would be a better name as it describes what actually is happening.

rmu75: > You are saying that "signed overflow is undefined" is problematic and asserting that "unsigned overflow is defined". They are both problematic. On most architectures (all in practice) they work exactly the same way because there is only one adder. It is the interpretation of status flags afterwards that introduces the problem.

To quote https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Signed-Overflow.html

> For signed integers, the result of overflow in C is in principle undefined, meaning that anything whatsoever could happen. Therefore, C compilers can do optimizations that treat the overflow case with total unconcern. (Since the result of overflow is undefined in principle, one cannot claim that these optimizations are erroneous.) 

and https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Unsigned-Overflow.html

> Unsigned arithmetic in C ignores overflow; it produces the true result modulo the nth power of 2, where n is the number of bits in the data type. We say it “truncates” the true result to the lowest n bits. 

the former is a problem, the latter not. Also, we can safely assume two's complement. 

Jeff Epler put his code through cbmc to check for undefined behaviour. 


right shift of signed ints is in theory implementation defined but in practice every implementation does arithmetic shift.

I looked a bit at the assembly generated at compiler explorer, it seems all those shifts etc... are eliminitated in the end.


without cast to unsigned, the compiler may determine in some specific circumstances that some values can overflow the addition and eliminate everything that depends on the result. such bugs are very hard to find, speaking from experience. we can't avoid possible overflow because "old" value can be arbitrary. signed overflow -> potential catastrophe, unsigned overflow -> everything ok (just truncates).

andypugh: > Counters with a limited bit-width will wrap while counting up or down.

Maybe "Hardware counters typically communicate a limited number of bits and will wrap when counting up or down"

BsAtHome: The phenomenon of wrapping is not limited to hardware counters. It cal also happen in interfacing in general.

BsAtHome: > without cast to unsigned, the compiler may determine in some specific circumstances that some values can overflow

Which circumstances?

rmu75: > > Which circumstances?
> please look at the whole comment and the link to the documentation of gcc.

That is hand waving. I did read the docs and cannot find the "specific conditions" you are referring to.

BsAtHome: > right shift of signed ints is in theory implementation defined but in practice every implementation does arithmetic shift.

https://cmu-sei.github.io/secure-coding-standards/sei-cert-c-coding-standard/recommendations/integers-int/int13-c/ (risk assessment: severity high)

rmu75:
> > right shift of signed ints is in theory implementation defined but
> > in practice every implementation does arithmetic shift.
> https://cmu-sei.github.io/secure-coding-standards/sei-cert-c-coding-standard/recommendations/integers-int/int13-c/
>   (risk assessment: severity high)

https://www.gnu.org/software/c-intro-and-ref/manual/html_node/Bits-Shifted-In.html

> For a signed data type, GNU C performs “arithmetic shift,” which keeps the number’s sign unchanged by duplicating the sign bit.

risk: non existent. but of course we could also use division. 


https://cmu-sei.github.io/secure-coding-standards/sei-cert-c-coding-standard/recommendations/integers-int/int16-c

> INT16-C. Do not make assumptions about representation of signed integers

https://gcc.gnu.org/onlinedocs/gcc/Integers-implementation.html

> GCC supports only two’s complement integer types, and all bit patterns are ordinary values. C23 requires signed integer types to be two’s complement.



We could also require  MISRA-C standards, but then nobody will be left
to contribute anything.

BsAtHome:
Lets me offer a compromise.

The cast to unsigned is fine. It doesn't hurt as such.

The right shift of a signed type does bother me. However, gcc supports it. You need to verify and make sure that clang also handles the right shift case like gcc. Additionally, you need to add a preprocessor test to check for gcc/clang (and their respective lowest version if applicable) and bail out with an error message if that is not what is used.

The advantage with the shift is that the whole algorithm is reduced to six assembly instructions.

src/hal/hal.h
rmu75: > > > Which circumstances?
> > > please look at the whole comment and the link to the documentation of gcc.
> 
> That is hand waving. I did read the docs and cannot find the "specific conditions" you are referring to.

It is hand waving because it really depends on the code, on the compiler, on optimization flags and what kind of code analysis the compiler is doing.

I will refer to https://predr.ag/blog/falsehoods-programmers-believe-about-undefined-behavior/

I have seen it happen with GCC on ARM that it replaced a variable where the compiler somehow was of the opinion that signed overflow will happen with 0, and eliminated statements like "if (variable > 0)".

rmu75:
> The right shift of a signed type does bother me. However, gcc supports it. You need to verify and make sure that clang also handles the right shift case like gcc. Additionally, you need to add a preprocessor test to check for gcc/clang (and their respective lowest version if applicable) and bail out with an error message if that is not what is used.

I googled a bit and it seems compilers that would right shift signed integers "logically" are a very rare minority and target "exotic" platforms like vintage Cray supercomputers or DSP platforms with fixed point hardware, even there you can usually switch behaviour with a compiler flag. GCC, CLANG, MSVC, Keil C and even SDCC always use two's complement and arithmetic shift. IMO compilers other than GCC and CLANG are of no immediate concern. 

FWIW I added a static assert in the code as a trip wire.

BsAtHome:
And what about the name `hal_unwrap_counter()`? Then the function name describes what the function does.

And the explanation I provided what the code does and how it actually works?

rmu75:
> And what about the name `hal_unwrap_counter()`? Then the function name describes what the function does.
> 
> And the explanation I provided what the code does and how it actually works?

need to catch up with all comments.

grandixximo:
my name preference

hal_counter_unwrap(prev_wide, new_low, nbits)

andypugh:
``` C
    /* tripwire if somebody tries to use this code on a Cray with wrong
       compiler flags */
    HAL_STATIC_ASSERT((-2 >> 1) == -1, 
        "hal_extend_counter impl only works with arithmetic right shift");
```

:-)

andypugh:
Are we still arguing about the name. or is this ready to go in?

BsAtHome:
I still think that hal_extend_counter is a misnomer that confuses its function.

rmu75:
On Mo, 2026-04-20 at 03:05 -0700, andypugh wrote:
> andypugh left a comment (LinuxCNC/linuxcnc#3936)
> Are we still arguing about the name. or is this ready to go in?

Some changes to comments and function name are pending, I hope I can
get to it today at night. Would be good if there was consensus
regarding the name. I don't really care, just want to get rid of the PR
😉

grandixximo:
🚀 hal_counter_unwrap
👀 hal_unwrap_counter
❤️hal_extend_counter

Vote on this post with a reaction matching the name preference....

rmu75:
So @BsAtHome  says `hal_unwrap_counter`, @andypugh  `hal_extend_counter` and @grandixximo `hal_counter_unwrap`.

BsAtHome:
There are two votes for "unwrap" in the name. I actually don't care too much about whether it is counter unwrap or unwrap counter.

rmu75:
:heart: from @pcw-mesa in irc.

BsAtHome:
Can we please just paint the bikeshed in <i>any</i> color?

rmu75:
I adapted / improved the explanation a bit (so I hope) and the rest including naming I defer to the native english speakers. As I am out of time currently, I will leave it at that, if someone wants to improve further pls. open separate PR.

grandixximo:
Two votes for unwrap, and two votes for extend, @jepler want to break the tie?

jepler:
No strong opinions.

grandixximo:
Changed to heart, split the difference, lets just paint the bike-shed ;-)

BsAtHome:
Cppcheck now has this warning:
```
hal/hal.h:1071:5: portability: Shifting a negative value is technically undefined behaviour [shiftNegativeLHS]
    HAL_STATIC_ASSERT((-2 >> 1) == -1,
```

rmu75:
cppcheck is wrong.

https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf page 95:

> The result of E1 >> E2 is E1 right-shifted E2 bit positions. If E1 has an unsigned type or if E1 has a signed type and a nonnegative value, the value of the result is the integral part of the quotient of E1 / 2E2. If E1 has a signed type and a negative value, the resulting value is **implementation-defined**.

(emphasis added)

Also it is kinda silly to complain about a check in a static assert ;-)

BsAtHome:
It generates a _**portability**_ warning and a portability problem it is. You can quote any standard, but fact is that it is a portability problem for all practical purposes and intents because compilers are different. And remember, cppcheck wasn't created to check our code, but to check code in general.

And, it doesn't matter if cppcheck is right or wrong in this case. We have this warning and it needs to be dealt with. In fact, there are several false positives from cppcheck in other places and we've dealt with them too. You need to disable the specific trigger locally with a special comment to quell the message.

rmu75:
ok I added comment to suppress portability warning about check for arithmetic right shift. #3986

PR #3941, opened 2026-04-16
add component 'output_buffer'

hansu:
This adds a component that behaves like a tristate buffer, but the outputs will be zero instead of high-Z.

I haven't found a better name. Could also be named 'output_switch' or 'multiple_output_buffer' ...

```
                            ┌───────────┐
  output-buffer.N.in-0 ─────┤  ──[>]──  ├───── output-buffer.N.out-0
                            │   ┌─┘     │
  output-buffer.N.in-1 ─────┤  ─┼[>]──  ├───── output-buffer.N.out-1
                            │   ├─┘     │
  output-buffer.N.in-2 ─────┤  ─┼[>]──  ├───── output-buffer.N.out-2
                            │   ├─┘     │
                   ... ─────┤   :       ├───── ...
                            │   :       │
  output-buffer.N.in-M ─────┤  ─┼[>]──  ├───── output-buffer.N.out-M
                            │   ├─┘     │
                enable ─────┤  ─┘       │
                            └───────────┘
```

### DIFF
--- a/src/hal/components/output_buffer.comp
+++ b/src/hal/components/output_buffer.comp
@@ -1,0 +1,48 @@
+component output_buffer "Feed through multiple bits when enable pin is set";
+pin in bit enable "Enable input";
+pin in bit off-level = 0 "Signal level when output is off (enable=0)";
+pin in bit in-#[32 : personality] "Inputs";
+pin out bit out-#[32 : personality] "Outputs (Follow inputs when enable=1)";
+
+function _;
+description """
+Feed through up to 32 bit inputs to their outputs when enable set,
+otherwise outputs will be the value of _off-level_ (default 0).
+
+                            ┌───────────┐
+  output-buffer.N.in-0 ─────┤  ──[>]──  ├───── output-buffer.N.out-0
+                            │   ┌─┘     │
+  output-buffer.N.in-1 ─────┤  ─┼[>]──  ├───── output-buffer.N.out-1
+                            │   ├─┘     │
+  output-buffer.N.in-2 ─────┤  ─┼[>]──  ├───── output-buffer.N.out-2
+                            │   ├─┘     │
+                   ... ─────┤   :       ├───── ...
+                            │   :       │
+  output-buffer.N.in-M ─────┤  ─┼[>]──  ├───── output-buffer.N.out-M
+                            │   ├─┘     │
+                enable ─────┤  ─┘       │
+                            │           │
+             off-level ─────┤ [0/1]     │
+                            └───────────┘
+
+The number of channels are determined by the value of 'personality'.
+""";
+examples """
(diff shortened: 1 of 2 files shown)


zz912:
Maybe "Enable binary signals" en-bin-sig

andypugh:
The same can be achieved with mux_generic in bit mode, with a little more flexibility in "disabled" value. 
This is simpler and easier to wire, though. Possibly also lighter-weight in terms of executed code. 
What do you see as the use-case?

hansu:
>The same can be achieved with mux_generic in bit mode

Not really - at least not with one.

Alternatively I could use an **and2** for each channel. But this would bloat my hal file unnecessarily...
I thought this component could be also useful for someone.

My use case is the following: I have an operating panel with illuminated keys. I simply want to turn off the illumination when the machine is not enabled. Otherwise the keys should get their appropriate illumination.

andypugh:
> > The same can be achieved with mux_generic in bit mode
> 
> Not really - at least not with one.

Indeed, you would need one for each channel. 

Is it worth considering a second set of "off" inputs (that default to zero)?

hansu:
>Is it worth considering a second set of "off" inputs (that default to zero)?

I thought about that too, but couldn't really imagine a case where this would be useful.
But yes I can add it  - this makes it more flexible.

BsAtHome:
Shouldn't you be able to set what level "off" actually is? I can imagine that someone would need negative logic.

hansu:
Added a parameter to set the "off-level". Could also be a pin...

Regarding the name: the correct name would be something like "buffer with enable". But shortenings like 'enable_buffer' or 'buffer_enable' sound weird.

andypugh:
It probably makes sense to not create new parameters at the moment, as they are (probably) being phased out.

hansu:
Ok, changed to pin.

src/hal/components/output_buffer.comp
```diff
@@ -0,0 +1,47 @@
...
+FUNCTION(_) {
+    int i;
+    for(i=0; i < personality; i++) {
+        if(enable) {
+            out(i) = in(i);
+        }
+        else {
+            out(i) = off_level;
+        }
+    }
```
BsAtHome: May I suggest a slightly different construct? The condition is part of the loop where volatile data is handled and cannot be optimized very much. The following construct reads the volatile values `enable` and `off_level` only once:
```C
if(enable) {
    for(int i = 0; i < personality; i++)
        out(i) = in(i);
} else {
    bool lvl = off_level;
    for(int i = 0; i < personality; i++)
        out(i) = lvl;
}
```

hansu: Agree, checking the enable pin outside the loop is more effective.

PR #3942, opened 2026-04-16
Docs: add info about jogging in unhomed state

hansu:
Add information in which state it is possible to jog in which mode.
I would like to have some more information about that in the docs and maybe a bit more detailed. So suggestions where to add information are welcome.

### DIFF
--- a/docs/src/man/man1/halui.1.adoc
+++ b/docs/src/man/man1/halui.1.adoc
@@ -271,6 +271,8 @@ _N_ = joint number (0 ... num_joints-1)
 
 === Joint jogging (*N* = joint number (0 ... num_joints-1))
 
+Joint jogging is possible in unhomed state except for joints having a negative HOME_SEQUENCE value. See Section "Homing Configuration" of the documentation.
+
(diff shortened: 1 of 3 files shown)


docs/src/man/man1/halui.1.adoc
```diff
@@ -271,6 +271,8 @@ _N_ = joint number (0 ... num_joints-1)
 
 === Joint jogging (*N* = joint number (0 ... num_joints-1))
 
+Joint jogging is possible in unhomed state except for joints having a negative HOME_SEQUENCE value. See Section "Homing Configuration" of the documentation.
```
andypugh, line 274: I think that this way of phrasing it inverts the cause and effect of negative HOME_SEQUENCE. 
Possibly it should say "except for axes with shared joints/actuators (as inidicated by a negative HOME_SEQUENCE)"

hansu: But can't you have shared joints with a non-negative HOME_SEQUENCE even though it makes not much sense?

PR #3943, opened 2026-04-16
qtvcp: fix PyQt5 ≥ 5.15 and PyQt6 compatibility

bgermann:
This is a follow-up on #3881


c-morley:
I suggest adding:

        LOG.info('green<Qt version: {}>'.format(QtCore.qVersion()))

on about line 171 in qtvcp.py, so we know what bindings are being used.

you can force pyqt6 by adding:

os.environ['QT_API'] = 'pyqt6'

on about line 11 in qtvcp.py

c-morley:
In indicatorMixin.py I get erros related to using toPointF() need to switch to QPointF instead eg:

gradient = QtGui.QLinearGradient((top_bot + QtCore.QPointF(size, 0)), top_bot)

bgermann:
Okay. The toPointF() calls should be resolved.

NTULINUX:
How would one go about using `pyside6-rcc`/`pyside6-uic` instead of `rcc` or `uic` if Qt6, without breaking compatibility with Qt5? Without the pyside6 variants, those are specific to Qt5 and do not exist on a pure Qt6 system. I can force it to use the pyside6 ones but then that will break LinuxCNC on Qt5 systems.

bgermann:
I do not know what is supposed to be a "pure Qt6" system but Debian has the two programs in the [qt6-base-dev-tools](https://packages.debian.org/trixie/amd64/qt6-base-dev-tools/filelist) package, built from src/tools directory of the source distribution.

NTULINUX:
Is it possible to add a `try:` thing to python so if rcc is not found, it tries to use pyside6-rcc?

Also, `/usr/lib/qt6/libexec/rcc` is not in the default path in debian.

c-morley:
Thank you. pushed manually.

Issue #3286, opened 2025-01-15
RFC: Switch 32 bit int HAL pins to 64 bit, rather than having both types in parallel. 


andypugh:
See also: https://github.com/LinuxCNC/linuxcnc/issues/2331

A long time ago HAL float pins were single-precision. These were switched to double, even though at the time many CPUs were 32 bit. 

Much more recently, now that most CPUs are 64-bit the 64-bit signed and unsigned HAL pins were added, along with the various conversion functions. 

I think we should simply change the width of integer HAL pins rather than retain two parallel types. 

As of now, LinuxCNC 2.9 does not have u64 and s64 HAL pins. We can fairly easily make a clean switch at this point in development without affecting any configs running the stable version.  

As things stand almost no HAL components have 64-bit inputs or outputs, and this includes the rawcounts pins of the encoders and stepgens, which is one place where 32-bit ints may well be inadequate. 

Rather than replace some instances of HAL component pins with 64-bit equivalents (which will cause incompatibilities in HAL files) I propose switching all integer HAL pins to the 64-bit equivalent simply by redefining the integer types. This way there need be no changes to existing HAL files. 

Some changes to certain HAL components can be made. For example "bitwise" will gain a number of new pins, but I would anticipate that this would not cause any functional changes in existing HAL configs. 

Similarly "LUT5" would gain one more  input, and should probably be renamed to LUT6. This name-change can be handled by the update_ini script. For a given config and input setup the operation would be unaffected.

andypugh:
Indeed, I had always anticipated it being a bit of a job, and was never considering redefining hal_u32 to be 64 bits 

There would have to be some leeway in halcompile. The docs suggest "signed" and "unsigned" as the type specifier, but u32 and s32 have been allowed, and should still be allowed to work for users with custom HAL components. (or we could actually error out with a helpful suggestion to switch to "signed" and "unsigned" as halcompiling is a one-time thing.

BsAtHome:
I don't think that using all-floats will do performance any good, regardless resolution.

And, FWIW, using doubles indiscriminately in calculations will reduce the accuracy of the value contained in the double to approx log10(sqrt(2^53)), or about 8 decimal digits, due to error accumulation and propagation.

rmu75:
[...]

PR #3944, opened 2026-04-17
refactor(hm2): drop _64 encoder pins, keep 64-bit internal

grandixximo:
RFC #3286 narrow option 1: fix float position wrap on high-count encoders via 64-bit internal counters, without exposing new HAL pin types. s32 count/rawcounts/rawlatch/ count-latched remain as truncated views of the internal 64-bit counters. Float position is computed from the full 64-bit value so it does not wrap.

Avoids committing four integer pin types (s32/u32/s64/u64) into 2.10 ABI before the getter/setter direction in #3286 is decided. The overflow fix from 8a3ac3f3 is preserved; only the pin exposure is reverted. Unused
prev_event_rawcounts_64 dropped.

BREAKING CHANGE: count_64, rawcounts_64, rawlatch_64 and count-latched_64 pins added post-2.9 are removed. HAL surface matches 2.9.

Refs #3286 #2331

### DIFF
--- a/docs/src/man/man9/hostmot2.9.adoc
+++ b/docs/src/man/man9/hostmot2.9.adoc
@@ -322,9 +322,7 @@ following pins and parameters:
 *Pins:*
 
 count (s32 out)::
- Number of encoder counts since the previous reset.
-count_64 (s64 out)::
- Number of encoder counts since the previous reset (64 bit).
+ Number of encoder counts since the previous reset. 32-bit truncation of the 64-bit internal counter; position is computed from the full-width internal value so it does not wrap.
 position (float out)::
  Encoder position in position units (count / scale).
 position-interpolated (float out)::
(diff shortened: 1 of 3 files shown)


rmu75:
sounds like this should be application of #3936

grandixximo:
> sounds like this should be application of #3936

Easy enough to rebase this on #3936 if that lands first

grandixximo:
@pcw-mesa please have a look at this, the `_64` bit pins will be deprecated eventually, this is IMO the correct stop gap, till getter/setter is ready, see #3286

pcw-mesa:
It this a step on the way to moving all 32 bit hal ints to 64 bit?

So dropping back to 32 bits temporarily?

grandixximo:
Yes, but the encoder 64bit should work if you connect to the 32bit pin, it is just the 32bit count view that is truncated, position will be correct, it's a stop gap to the getter setter discussed on #3286

pcw-mesa:
I think that's fine as a stop gap, the integer pins are also just used to some edge applications
(simulated index/mod etc)

PR #3949, opened 2026-04-18
motion: axis_sync_teleop_tp_to_carte_pos resets velocity/acceleration

is-primary-dev:
`axis_sync_teleop_tp_to_carte_pos` in `src/emc/motion/axis.c` updates `teleop_tp.curr_pos` and `pos_cmd` to the current cartesian command but leaves `curr_vel` and `curr_acc` untouched. Both callers semantically represent "joint at rest at the current cartesian position" — enabling motion from `DISABLED`, and entering `TELEOP` after homing. Leaving residual velocity in the trajectory planner lets `simple_tp_update_normal` integrate one cycle of motion on the first pass after the sync, drifting `curr_pos` away from the synced value. The patch zeros `curr_vel` and `curr_acc` alongside the existing position sync.

Reproduction on an XYZZ setup where Z homes at `MAX_LIMIT`: home all, jog Z in the minus direction, release the jog key (`axis_jog_abort(immediate=0)` runs with residual velocity), re-home. Pre-fix: Z lands 1–3 mm below `HOME` depending on jog-settle timing. Post-fix: Z lands at `HOME` within machine resolution.

No upstream test harness exercises the `FREE -> TELEOP` transition after re-homing, so verification is by manual reproduction.

### DIFF
--- a/src/emc/motion/axis.c
+++ b/src/emc/motion/axis.c
@@ -440,10 +440,22 @@ void axis_sync_teleop_tp_to_carte_pos(int extfactor, double *pcmd_p[])
 {
     int n;
     // expect extfactor =  -1 || 0 || +1
+    //
+    // This function initializes the teleop trajectory planner to a REST
+    // state at the current cartesian position. Both callers (enabling
+    // motion, entering teleop mode) expect the joint to be stationary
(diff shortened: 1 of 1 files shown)


andypugh:
Thanks. We assume that someone will shout if it breaks something.

Issue #3899, opened 2026-04-03
Gmoccapy scaling up each time you exit tool table - 2.9.8 trixie


rotarysmp:
Hi Guys,

I updated my lathe to 2.9.8 on Trixie with Preemt-RT. Now Gmoccapy appears to have a bug causing it to scale up each time you exit the tool table or the max preview. window size setting also doesn't work correctly, as the screen reduces in height when commanded to increase in either width or height.

I made a video to show the behavior. 
https://www.youtube.com/watch?v=H-MpyDQSuOo

Like this, the lathe is nearly unusable, as I have to keep shutting down and restarting LinuxCNC to reset it. Any suggestions how to correct this greatly appreciated.
Cheers,
Mark

hansu:
Can you try the gmoccapy sample config 'lathe' and check if it shows the same behaviour?
Are you running X11 or Wayland (check with `echo $XDG_SESSION_TYPE`)?

rotarysmp:
Hi Hans,

Thanks for taking a look at it.

It reports X11.

I loaded the Gmoccapy Lathe simulator. It does not display this scaling behaviour. I works normally.

I loaded the Gmoccapy lathe_macros simulator, as that is the one used. It has the same scaling behaviour.

Thanks for that data point. Any ideas what I should do next?
Cheers,
Mark

hansu:
Do you use the lathemacro like in the lathe_macros sim config?
If yes, try this as a temporary workaround:
```diff
EMBED_TAB_NAME = Cycles
-EMBED_TAB_LOCATION = ntb_preview
+EMBED_TAB_LOCATION = ntb_user_tabs
EMBED_TAB_COMMAND = halcmd loadusr -Wn gladevcp gladevcp -c gladevcp -U notouch=1 -U norun=0 -u lathehandler.py -x {XID} lathemacro.ui
```

If not, please post your INI config file.

rotarysmp:
Yes, the same macros from the sim.

I cut/paste those lines into the Macro section of the INI, but the screen still scales when I exit the tool holder screen. 

[schaublin_125-CNC.ini.txt](https://github.com/user-attachments/files/26470982/schaublin_125-CNC.ini.txt)

I have attached my INI file.
Oh wait, I just noticed I already had those lines under Display. I removed them from Macros, and just added the missing 
+EMBED_TAB_LOCATION = ntb_user_tabs.

It still has the scaling problem.

hansu:
[...]

PR #3950, opened 2026-04-18
Fix Gmoccapy screen resizing issue #3899

andypugh:
Also trigger redraw on a different event, this massively reduces CPU/GPU load.


hansu:
Thanks!

PR #3952, opened 2026-04-19
Integrate gantry (XYYZ) into matrixkins.comp

hdiethelm:
matrixkins did not support gantry machines and mine is one. So I added support for gantry.

Additionally, the setup was done in the first call to `kinematicsType()` but there is the `EXTRA_SETUP()` function dedicated to do the setup.

`coordinates` and `kinstype` arguments added, so this is a drop in replacement for trivkins for simple setups. If a setup is not supported, there is an error message.

### DIFF
--- a/src/hal/components/matrixkins.comp
+++ b/src/hal/components/matrixkins.comp
@@ -19,6 +19,7 @@ component matrixkins "Calibrated kinematics for 3-axis machines";
 description
 """
 The matrixkins component implements custom kinematics for 3-axis
+Use coordinates=XYZ for standard and coordinates=XYYZ for gantry
 Cartesian machines that allows compensating minor alignment inaccuracies
 in software.
 
@@ -30,6 +31,11 @@ By default identity matrix is used, which is equal to trivial kinematics:
 | X_joint |   | 1 0 0 |   | X_axis |
 | Y_joint | = | 0 1 0 | * | Y_axis |
 | Z_joint |   | 0 0 1 |   | Z_axis |
+or
+| X_joint  |   | 1 0 0 |   | X_axis |
+| Y1_joint | = | 0 1 0 | * | Y_axis |
+| Y2_joint | = | 0 1 0 | * | Y_axis |
+| Z_joint  |   | 0 0 1 |   | Z_axis |
 ....
 
 Adjusting the calibration matrix allows compensating out many
@@ -46,6 +52,11 @@ For 3 axis machine, the equations become:
 ....
 X_joint = C_xx * X_axis + C_xy * Y_axis + C_xz * Z_axis
 Y_joint = C_yx * X_axis + C_yy * Y_axis + C_yz * Z_axis
+Z_joint = C_zx * X_axis + C_zy * Y_axis + C_zz * Z_axis
+or:
+X_joint = C_xx * X_axis + C_xy * Y_axis + C_xz * Z_axis
+Y1_joint = C_yx * X_axis + C_yy * Y_axis + C_yz * Z_axis
+Y2_joint = C_yx * X_axis + C_yy * Y_axis + C_yz * Z_axis
 Z_joint = C_zx * X_axis + C_zy * Y_axis + C_zz * Z_axis
 ....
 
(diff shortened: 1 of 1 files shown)


src/hal/components/matrixkins.comp
```diff
@@ -220,28 +278,24 @@ static int matrixkins_setup(void) {
 
     if (res) goto error;
 
-    res = hal_ready(comp_id);
```
andypugh, line 223: I don't see a replacement for hal_ready() but .comp files don't normally need one. Is this now automatically included with the EXTRA_SETUP macro?

src/hal/components/matrixkins.comp
```diff
@@ -19,6 +19,7 @@ component matrixkins "Calibrated kinematics for 3-axis machines";
 description
 """
 The matrixkins component implements custom kinematics for 3-axis
+Use coordinates=XYZ for standard and coordinates=XYYZ for gantry
```
andypugh, line 22: This sentence seems to be out of place.

andypugh:
This seems to be able to handle only XYYZ. It would ideally handle XXYZ and XXYYZ too. And maybe ZZ variants. 
I have even seen ZZZ used on some 3D printers. (Voron Trident, though that is also CoreXY so out of scope for these kins)

src/hal/components/matrixkins.comp
hdiethelm: I removed also a hal_set_unready(). I think so but i have to check the generated c code.

hdiethelm:
Yes. Handling other coordinate lists would require to copy paste a lot of code from trivkins and make it complex.
There is no generic calibration available yet except writing it in hal or kins, right?
A better option would be to do a it in a generic way before the various kins:
XYZAB* -> calibrate -> XYZAB*_cal -> kins -> joints
and the whole chain also backwards for inverse kinematics. That would then work for all kins types.
What do you think?

andypugh:
Pondering further on this, I am wondering why a 2-motor gantry would be out-of-square. One feature of a gantry (using trivkins) is that it can be squared-up by adjusting the HOME_OFFSET for the paired joints. 

I think that correcting for out-of-square is a bit of a bodge, but if it has to be done, then the kinematics seems like a reasonable place. It's basically just admitting that it is a non-cartesian geometry.

hdiethelm:
Every machine is crucked in all axis if you measure precise eneught. YY can be corrected with HOME_OFFSET. But mine was not streight in Z by 0.08 mm. I to corrected it in software, the reason for this PR, works perfect. Thanks LinuxCNC for making this possible. :-)
Without this change, this module is not usable on gantrys. But if you think it's not needed I can just close this PR, i just wanted to share it.
Z is also bowed by 0.02 mm which I can not correct with this module. A general calibration module could make this also possible provided backlash and resolution is low enought.
The described way above would make it possible to create such a module that works for all kins, not only for specific joint configurations.
I can start a thread in the forum to see if enought people are interested in such a solution.

hdiethelm:
@PetteriAimonen Looks like the initial module was written by you. What do you think about this change?

PetteriAimonen:
In principle this seems ok, but I think is we want to generalize it to gantries, then it shouldn't be limited to having the dual axis just on Y.

So it should then support any coordinates string like "XXYZ" or "XXYYZZABC", and initialize some array of indices which is then used to permute the joints array in the kinematics functions.

hdiethelm:
This is a valid point and I was thinking about this while changing the code. I already went trough the code in trivkins.c and the related kins_util.c.

If identityKinematicsSetup(), identityKinematicsForward() and identityKinematicsInverse() can be reused, this would be one way to go.

After some greping trough the code, I found in the top level makefile:
```
obj-m += trivkins.o
trivkins-objs := emc/kinematics/trivkins.o
trivkins-objs += emc/kinematics/kins_util.o
```
I think this is how the modules are linked together, right? Until now, if I use them in matrixkins.comp, I just got `matrixkins: dlopen: .../linuxcnc/rtlib/matrixkins.so: undefined symbol: identityKinematicsForward`.

So probably the way to go would would be to convert this module to C code, move it to emc/kinematics, adapt the Makefile and then reuse identityKinematics*().

An even better option would be to add a generic nD calibration hook:
XYZABCUVW -> calibrate -> XYZABCUVW_cal -> kins -> joints
The advantage would be that the nD calibration is independent of the kinematics afterwards, so you can use it for trivkins but also for any other kins type like corexykins.

Now LinuxCNC is still a hobby for me. I might or might not do one of these two options in the near future.

Feel free to close this PR if it is not general enough, I understand it. It works for me and a few others but for sure not everybody.

src/hal/components/matrixkins.comp
hdiethelm: Yes, hal_ready() is called in rtapi_app_main() after EXTRA_SETUP(). Otherwise, it probably would not have worked anyway.
The previous implementation used hal_set_unready(), additional init, hal_ready() at the first call of kinematicsType() which seamed a bit odd to me.

Issue #3818, opened 2026-02-21
Start page - My configurations - missing


zz912:
RIP instalation
```
zdenek@cnc:~$ uname -a
Linux cnc 6.1.0-43-rt-amd64 #1 SMP PREEMPT_RT Debian 6.1.162-1 (2026-02-08) x86_64 GNU/Linux
```

In master branch "My configurations" is missing:

<img width="790" height="514" alt="Image" src="https://github.com/user-attachments/assets/89cc2d19-a184-4564-b656-c690df9f3b8f" />

Old branch:

<img width="790" height="514" alt="Image" src="https://github.com/user-attachments/assets/94a94f5f-bbc5-4037-a7df-4004ac585bfb" />

grandixximo:
@zz912 please check #3962 if it fixes this issue

zz912:
I can confirm: https://github.com/LinuxCNC/linuxcnc/pull/3962 fixes this issue.

grandixximo:
Thank you

PR #3962, opened 2026-04-24
Fix #3924: linuxcnc creates '$' folder in cwd

### LINKED ISSUE #3924: RIP build now makes a temp file or folder (I think called $) in the working directory where linuxcnc 2.10pre is invoked from the command line
This seems to be a new issue in the last week or so. If you run the rip-environment script and then move to a system folder (eg /etc), 
and run a linuxcnc sim, Linuxcnc crashes becasue it does not have permissions to create the temporary folder.

From my memory where am building and testing  linuxcnc at least once a week, this is new behaviour.

If you run Linuxcnc from within your home folder, the error does not occur. Presumably because you have write access.

grandixximo:
## Summary
- Fixes the `$`-folder regression reported in #3924.
- Regression introduced by 6f0bf8b845 (@BsAtHome, #3739) when swapping tilde expansion for `$::env(HOME)` to support Tcl 9.
- `tcl/linuxcnc.tcl.in` wrapped `@LINUXCNC_CONFIG_PATH_TCL@` in `{}` braces, which blocked Tcl substitution. After autoconf substituted the literal `$::env(HOME)/linuxcnc/configs:...`, `split` on `:` chopped `::` into empty tokens, the first element became literal `$`, and `USER_CONFIG_DIR` resolved to `[pwd]/$`.
- Switched to `[subst -nocommands -nobackslashes {...}]` so `$::env(HOME)` expands while `[...]` command substitution stays disabled (safer than plain double-quoting in case a path ever contains brackets).

With the fix `USER_CONFIG_DIR` correctly resolves to `~/linuxcnc/configs`, and existing user configs reappear under "My Configurations".

## Test plan
- [x] Verified `USER_CONFIG_DIR` resolves to `/home/<user>/linuxcnc/configs` from `cwd=/etc` under RIP build (`RUN_IN_PLACE=yes`).
- [x] Verified same resolution under non-RIP build (`--prefix=/usr`, matching .deb config).
- [x] Verified no `$` folder created under RIP.
- [ ] Third-party retest on .deb install from `/etc` (cc @rodw-au @NTULINUX @thomam04).
- [x] Retest with Tcl 9 once available.

### DIFF
--- a/debian/rules.in
+++ b/debian/rules.in
@@ -88,16 +88,16 @@ override_dh_auto_install-arch:
 	rm -f $(DESTDIR)/usr/share/doc/@MAIN_PACKAGE_NAME@/examples/sample-configs/*/*position*.txt
 
 override_dh_installdocs-arch:
-ifeq (,$(filter nodocs,$(DEB_BUILD_OPTIONS)))
-	# Sample configs go in `usr/share/doc/linuxcnc` (not `.../doc/@MAIN_PACKAGE_NAME@`)
-	# because that's where the `linuxcnc` launcher script looks for them,
-	# and that's inconvenient to change.
+	# Sample configs are runtime data read by the launcher, not documentation,
+	# so they must be installed unconditionally (i.e. regardless of
+	# DEB_BUILD_OPTIONS=nodocs).  They go in `usr/share/doc/linuxcnc` (not
+	# `.../doc/@MAIN_PACKAGE_NAME@`) because that's where the `linuxcnc` launcher
+	# script looks for them, and that's inconvenient to change.
 	dh_installdocs --doc-main-package=@MAIN_PACKAGE_NAME@ --package=@MAIN_PACKAGE_NAME@
 	mkdir -p debian/@MAIN_PACKAGE_NAME@/usr/share/doc/linuxcnc
 	mv debian/@MAIN_PACKAGE_NAME@/usr/share/doc/@MAIN_PACKAGE_NAME@/examples debian/@MAIN_PACKAGE_NAME@/usr/share/doc/linuxcnc
 
 	dh_installdocs --doc-main-package=@MAIN_PACKAGE_NAME@ --package=@MAIN_PACKAGE_NAME@-dev
-endif
 
 override_dh_auto_install-indep:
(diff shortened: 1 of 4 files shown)


NTULINUX:
On both Fedora (Tcl9, rip build) and Gentoo (Tcl8, rip build and system-wide install) LinuxCNC does not ask me to copy files to my home directories still (same behavior as: https://github.com/LinuxCNC/linuxcnc/issues/3924#issuecomment-4229465787 ) but otherwise works fine without adding a `$` directory. Thank you!

grandixximo:
@thomam04 @NTULINUX, pushed three more commits, please re-test:

- `f191ac9298` makes `make_ini_for_tcl` fall back to a per-user cache dir (`$XDG_RUNTIME_DIR/linuxcnc-$UID/`) when the inifile's directory is read-only, and exposes that path to Tcl via `LINUXCNC_INI_EXPANDED`. This should clear the startup error @thomam04 hit after applying the original fix when launching an installed sample config directly.
- `00c61bfa3d` makes `ok_to_copy_config` force the copy prompt for non-RIP builds when the selected file is outside `myconfigs_node`, regardless of filesystem writability. This addresses @NTULINUX's "no copy prompt on Gentoo system-wide install": the writability heuristic doesn't hold on distros where installed sample-configs land in a path the running user can write to.
- `526ce653e7` is a tangential packaging fix found while testing. `debian/rules` was gating the sample-configs install on `nodocs`, so a `DEB_BUILD_OPTIONS=nodocs` build produced a `.deb` with no built-in configurations. Sample configs are runtime data, not docs, so they were moved out of the guard. Happy to drop this one if reviewers want #3924 narrowly scoped.

For RIP, the copy prompt is intentionally suppressed (Sigma1912's design: RIP samples live under your checkout, customizing them is fine). To exercise the prompt path in RIP for testing, use `debug_pickconfig=1 linuxcnc`.

System installs (Debian/Gentoo/Fedora system-wide) should now prompt to copy when you pick a sample, and starting an installed sample directly via path or desktop shortcut should no longer fail on the read-only `.expanded` write.

grandixximo:
This seems to also fix #3818

PR #3964, opened 2026-04-24
Rootless uspace preempt-rt (rebase of #918, fixes for #3928)

### LINKED ISSUE #3928: Why silently ignore the `LINUXCNC_FORCE_REALTIME` environment variable if some other `rtapi_app` binary doesn't have setuid?
I'm currently packaging LinuxCNC on NixOS, and I find some design choices rather appalling.

LinuxCNC tries very hard to detect whether it is real-time capable. And this logic is flawed in multiple ways, at least from my point of view.

1. `rt_api` explicitly checks for setuid as root, as opposed to, just the needed capabilities. I guess for uspace with PREEMPT_RT kernel `cap_net_admin,cap_sys_nice,cap_sys_resource+p` would suffice, but oh well.
2. It does not check this on `argv[0]` (the actual file executable), but on on `EMC2_BIN_DIR "/rtapi_app"`, regardless of whether that is the file which was called.
3. It does check this, **before** considering the `LINUXCNC_FORCE_REALTIME`, silently rendering that mechanism useless.

Now, on NixOS this causes a problem, because all applications reside in `/nix/store` and never have the setuid bit (or any file based capabilities) set. For things like `sudo` or `rtapi_app`, we usually create a wrapper with setuid set under `/run/wrappers/bin/`, than dispatching (I'm not sure how, I believe via exec) to the actual executable in `/nix/store`.

Regardless of my niche use-case on running LinuxCNC on NixOS, in general, I think at least the  `LINUXCNC_FORCE_REALTIME` should be honored in **all cases**.

https://github.com/LinuxCNC/linuxcnc/blob/568354c7ed26efce1fd76db783983038e1d82f55/src/rtapi/uspace_common.h#L412-L416

grandixximo:
Draft. Opening early for visibility while I coordinate merge order
with @hdiethelm since PR #3919 touches the same files.

Revives PR #918 (@dwrobel) rebased onto current master, addresses
the three concrete complaints in issue #3928 (@wucke13), and finishes
the items that #918 listed as TODO.

## What the branch does

* Four cherry-picks from PR #918, conflicts resolved.
* A build-fix commit restoring dispatch helpers and the `unlimited`
  rlimit constant that #918 removed but master still references.
* New `make setcap` target with a post-build warning that accepts
  either setuid or file capabilities as "ready".
* `hm2_eth` auto-skips iptables when it cannot be invoked, plus a
  `no_iptables=1` module parameter as an explicit override.
* `rtapi_is_realtime()` reworked as a SCHED_FIFO success probe,
  matching the convention in JACK, PipeWire, rtkit, Xenomai and
  Klipper. `LINUXCNC_FORCE_REALTIME` is honored before any other
  check. Closes #3928.
* udev rules for Mesa HostMot2 PCI cards (direct-Mesa 0x2718 and
  PLX-bridged Mesa boards listed in `hm2_pci.h`) and for
  `/dev/cpu_dma_latency`, both using the `plugdev` group.

## Tested

Debian 13.4, stock non-RT kernel 6.12.57. With `make setcap` the
RT thread runs SCHED_FIFO priority 98 unprivileged. Without caps the
code falls back to SCHED_OTHER with a clear warning.

## Help wanted: hardware testing

I do not have Mesa hardware on hand, so the following paths need a
second pair of eyes with a real card before merge:

* Mesa PCI / PCIe cards (5i2x, 6i25, 5i24 family). The udev rule is
  validated by `udevadm verify` only, the chmod / chgrp runtime effect
  on `/sys/bus/pci/devices/*/{config,resource*}` is not.
* Mesa ethernet cards (7i76e, 7i80, 7i92, 7i96). The iptables auto-skip
  path under rootless needs confirmation that packet reception and
  interface isolation still behave as expected.
* Parallel-port setups (pluto, hm2_7i43 via EPP). `iopl(3)` path with
  `CAP_SYS_RAWIO` only.

@pcw-mesa, if you happen to see this, any feedback on the udev vendor
and subsystem ID list in `99-linuxcnc-hm2-pci.rules` would be very
welcome. Alternatively, @andypugh or anyone else with hardware and
time, please let me know and I can iterate.

## Not in this PR

Debian packaging setcap (`dh_setcap` or postinst hook). The existing
`sudo make setuid` path is unchanged so there is no regression.
Happy to add a follow-up if reviewers prefer.

## Coordination

Per the PR #3919 discussion, rootless was agreed to be a follow-up.
I am happy to target whichever base is least effort for you, and will
do the rebasing work.

cc @dwrobel, @hdiethelm, @BsAtHome, @andypugh, @jepler


hdiethelm:
Hello
Thanks for taking this on and regarding my pending PR.
To be able to see the changes compared to rtapi_cleanup_v2, you can use this link: https://github.com/hdiethelm/linuxcnc-fork/compare/rtapi_cleanup_v2...grandixximo:linuxcnc:nosetuid

I did a short review and testing round with latency test in the VM, what I found so far with different kernels:
none:
- Standard-Kernel: non-realtime, Ok
- Preemt RT: non-realtime, Ok
- Xenomai3: Fail: `iopl() failed: Operation not permitted`
- Xenomai4: Fail: `iopl() failed: Operation not permitted`

setcap:
- Standard-Kernel: Fail, shows `POSIX realtime`, I think the PREEMT_RT check should still be used here.
- Preemt RT: Ok
- Xenomai3:  `WARNING: [main] cannot open RTDM device /dev/rtdm/memdev-private: Permission denied`:  I have to check if there are cap's for this.
- Xenomai4: `evl_sleep_until: Operation not permitted`: I have to check if there are cap's for this.

setuid:
- Standard-Kernel: Fail, shows `POSIX realtime`, I think the PREEMT_RT check should still be used here.
- Preemt RT: Ok
- Xenomai3: Ok
- Xenomai4: Ok

Issue for the standard kernel: can_set_sched_fifo() returns true also when kernel has PREEMPT_DYNAMIC and it will generate issues for people later on wondering why their latency is so bad.

It is not basically wrong due to you can use SCHED_FIFO on PREEMPT_DYNAMIC kernels to be better than no RT but it will be worse than with PREEMT_RT. A warning would make sense in this case. I would leave the can_set_sched_fifo() logic equal otherwise so you can use FIFO on PREEMPT_DYNAMIC from now on.

Issue for Xenomai and probably also RTAI variants: In the actual state, setuid should be checked or if there is a cap, it should be set and checked instead. I would just add a setuid() check in these detect_, return false if not setuid and show a warning that this is not supported. Xenomai / RTAI is not used that often and it can be improved if Xenomai gets some traction. This will solve the issue with "none" and "setcap" i had in these cases.

I can help you with testing of different RT variants and a 7I96S MESA Ethernet card.

hdiethelm:
Ah, I got it for xenomai3:
`sudo adduser $USER xenomai` so detect_xenomai() should check if the user is in this group as long as setuid is not set.
https://v3.xenomai.org/tips/regular-user/ Warning: They say this is probably exploitable.

xenomai4:
https://evlproject.org/core/user-api/index.html
`sudo chown -R $USER:$USER /dev/evl/` does the trick but is not the way to do it. Creating an evl group in my libevl debian package, setting up udev rules and then using `sudo adduser $USER evl` would be nicer.

If you don't want to bother about this, I am fine if you just add the setuid check for  Xenomai and RTAI and I can tackle the setcap / group variant in a later PR.

hdiethelm:
BTW: To check in the code if the cap's are correctly set: https://linux.die.net/man/3/libcap / https://linux.die.net/man/3/cap_get_proc
This is gives you an additional check to make sure the cap's are set correctly.
The following code + `-lcap` in the Makefile does the job for cap_net_admin:
```C
cap_t caps;
cap_flag_value_t flag_value;
caps = cap_get_proc();
if (caps == NULL){
    perror("cap_get_proc");
    return false;
}

if (cap_get_flag(caps, CAP_NET_ADMIN, CAP_EFFECTIVE, &flag_value) == -1) {
    perror("cap_get_flag");
    return false;
}

if (cap_free(caps) == -1){
    perror("cap_free");
    return false;
}

printf("Flag is %i\n", flag_value);
return flag_value == CAP_SET;
```

grandixximo:
Thanks for the test matrix and the libcap pointer @hdiethelm. Pushed `aa0f9d43de` addressing both concerns.

**1. PREEMPT_DYNAMIC vs PREEMPT_RT.** Restored `detect_preempt_rt()` (uname based) but use it for a one shot warning at `makeApp()` rather than as a gate. Reasoning: SCHED_FIFO on a PREEMPT_DYNAMIC stock kernel is genuinely better than SCHED_OTHER, just not bounded the way PREEMPT_RT or Xenomai is. Gating it false would make the rootless path useless on the most common laptop/desktop install. JACK, PipeWire, rtkit do not check kernel type; Klipper warns and runs anyway. Took the Klipper approach. Output now:

```
Note: SCHED_FIFO available but kernel is not PREEMPT_RT.
Latency may be unbounded; install a PREEMPT_RT kernel for hard
realtime guarantees.
```

`rtapi_is_realtime()` itself stays a pure capability probe so `hal.is_rt` keeps its existing binary semantics for downstream callers.

**2. Xenomai/RTAI fallthrough.** Added `geteuid()==0` gate inside `detect_rtai()`, `detect_xenomai()`, `detect_xenomai_evl()` with a FIXME comment naming you and your follow up plan. An unprivileged user on a Xenomai kernel will now fall through to the SCHED_FIFO probe and the POSIX path instead of hitting `iopl() failed: Operation not permitted`. Setuid users keep current behavior.

This is the band aid you suggested, not the proper fix. Could you tag me on the follow up PR that adds udev rules + `xenomai`/`evl` group membership? I will rebase and drop the setuid gate at that point.

**3. libcap.** Skipping the precheck for now. The syscall return value is the ground truth (caps can be present but ambient/inheritable wrong, file caps can be set but not effective on the running thread), and JACK/PipeWire/rtkit follow the same try and handle EPERM convention. Happy to add a libcap diagnostic dump as a separate optional helper if you find it useful for triage. The `cap_get_proc` snippet is saved for that.

Yes please on testing the 7I96S, that closes the biggest gap I have. Let me know what you need from me to make that easy.

Diff against `rtapi_cleanup_v2`: https://github.com/hdiethelm/linuxcnc-fork/compare/rtapi_cleanup_v2...grandixximo:linuxcnc:nosetuid

hdiethelm:
^Now you had in a similar issue I had with the uspace_common.h header. I think this should be split in a .c and .h and the make files adapted. I did not do that in the rtapi_cleanup_v2 just because it is already a big enough PR for my taste.

I did a quick test, your last commit does the job. But I would change to RTAPI_MSG_ERR for the note about PREEMPT_RT, default level is error and with warning, you normally don't see it.

>This is the band aid you suggested, not the proper fix. Could you tag me on the follow up PR that adds udev rules + xenomai/evl group membership? I will rebase and drop the setuid gate at that point.

My roadmap for Xenomai is approximately: 
1. POC for RT network, I got a supported card a few days ago
2. Add network the proper way (After rtapi_cleanup_v2 got merged, so I don't have to much re-basing to do)
3. Better Debian packages
4. Non-root, drop the setuid() check
5. Documentation
6. Backport to 2.9 and try to get some feedback about the performance from the forum and if Xenomai is desired and useful
7. TBD: Xenomai kernel module version (This will be even lower latency but probably not that simple)

So I think this PR will get merged before I tackle the non-root support for Xenomai. For LXRT, I think it is not worth doing it because it was already broken for years and no one complained. LRXT is also really unstable on my PC.

I can tag you as soon as 4. is done. Xenomai 4 has no Debian packaging support, so I did a minimalist package myself but this has to be improved before I can do the proper udev + evl group support. The udev rules should be in there so it is general, not linuxcnc specific. But my knowledge of proper Debian packaging not that big, so I have to read in first which will take some time. Except someone is willing to help.

>3. libcap

I think also this is not strictly necessary. It can be added if people are having issues for additional diagnostics.

> Yes please on testing the 7I96S, that closes the biggest gap I have. Let me know what you need from me to make that easy.

You will get feedback on this soon.

hdiethelm:
I tested using 7I96S / Preemt RT:
none:
Failure: `hm2_eth: ERROR: ioctl SIOCSARP failed: Operation not permitted`
Probably nothing new.

setcap:
```
hm2_eth: loading Mesa AnyIO HostMot2 ethernet driver version 0.2
hm2_eth: iptables is not available to this process (running unprivileged?); skipping automatic rule installation.  Configure firewall externally to isolate the hm2-eth interface.
hm2_eth: WARNING: Unable to restrict other access to the hm2-eth device.
This means that other software using the same network interface can violate
realtime guarantees.  See hm2_eth(9) for more information.
```
In this case, it would be nice to show to the user the commands needed to activate the isolation. The man page does not say how to do this. This could be done by modifying install_iptables_rule() and other functions to either show or activate the rule. This would look like this:
```
iptables -N hm2-eth-rules-output
iptables -A OUTPUT -j hm2-eth-rules-output
iptables -A hm2-eth-rules-output -s 192.168.1.120/32 -d 192.168.1.121/32 -p udp -m udp --sport 42357 --dport 27181 -j ACCEPT
iptables -A hm2-eth-rules-output -o enp4s0 -p icmp -j DROP
iptables -A hm2-eth-rules-output -o enp4s0 -j REJECT --reject-with icmp-admin-prohibited
```

setuid:
All works as expected.

While testing, I discovered another issue, also not related to your change:
The iptables are not fully cleaned up after exiting linuxcnc. Each time I start/stop linuxcnc, there is one new `-A OUTPUT -j hm2-eth-rules-output` appended. So after 3 start/stop I get:
```
sudo iptables -S
-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT
-N hm2-eth-rules-output
-A OUTPUT -j hm2-eth-rules-output
-A OUTPUT -j hm2-eth-rules-output
-A OUTPUT -j hm2-eth-rules-output
```
Feel free to cherry-pick this commit to your PR: https://github.com/LinuxCNC/linuxcnc/commit/c226577a886b6cad02b060f02652a3e9d84d46bf
I can also create a separate PR if you prefer.

With this change, it is cleaned up properly:
```
sudo iptables -S
-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT
```

grandixximo:
Pushed `71bec2860a` addressing both rounds.

* `RTAPI_MSG_WARN` to `RTAPI_MSG_ERR` on the PREEMPT_RT note (folded into the gate commit). Will now show at default level.
* `hm2_eth` print-or-exec mode for iptables. When unavailable, install_iptables_rule() prints the command instead of skipping. The probe also emits the chain setup commands. User gets a complete copy paste recipe.
* Cherry-picked your `c226577a886` (Fully cleanup iptables) with original authorship preserved.

Header split for `uspace_common.h` deferred. Agree it should be done, but as you noted, scope creep for this PR.

Roadmap looks reasonable. Tag me when step 4 lands and I will rebase off the proper Xenomai cap path and drop the setuid gate.

hdiethelm:
There are some issues with the "print-or-exec":

`--sport` is assigned automatically and changes on each start of linuxcnc. If I set the rule from the first start, at the second start, I get an error: `hm2_eth: ERROR: Could not retrieve hardware address (MAC) of 192.168.1.121: Operation not permitted` due to the source port has changed and is blocked now. Just leaving `--sport` away works but doesn't do full port isolation. Ping is blocked but udp requests from other ports to 27181 not. This should not be a big deal, it's not windows throwing out tons of packages on each interface continuously... ;-)

The output is mangled between other init code and `env "PATH=/usr/sbin:/sbin:${PATH}"` works but it is not what you normally enter as command.
```
hm2_eth: loading Mesa AnyIO HostMot2 ethernet driver version 0.2
hm2_eth: iptables is not available to this process (running unprivileged?); automatic rule installation will be skipped.  To enable hm2-eth isolation manually, run the following as root, plus the per-board and per-interface rules logged below as boards come up:
    env "PATH=/usr/sbin:/sbin:${PATH}" iptables -N hm2-eth-rules-output
    env "PATH=/usr/sbin:/sbin:${PATH}" iptables -I OUTPUT 1 -j hm2-eth-rules-output
hm2_eth: WARNING: Unable to restrict other access to the hm2-eth device.
This means that other software using the same network interface can violate
realtime guarantees.  See hm2_eth(9) for more information.
hm2_eth: 192.168.1.121: INFO: Hardware address (MAC): 00:60:1b:16:8b:d3
hm2_eth:     env "PATH=/usr/sbin:/sbin:${PATH}" iptables -A hm2-eth-rules-output -p udp -m udp -d 192.168.1.121 --dport 27181 -s 192.168.1.120 --sport 37028 -j ACCEPT
hm2_eth: discovered 7I96S
hm2/hm2_7i96s.0: Low Level init 0.15
hm2/hm2_7i96s.0: Smart Serial Firmware Version 43
hm2/hm2_7i96s.0: 51 I/O Pins used:
hm2/hm2_7i96s.0:     IO Pin 000 (TB3-01): InM Input Module #0, pin in0,enca0 (Input)
...all the pins are listed, for me 50 lines...
hm2/hm2_7i96s.0:     IO Pin 050 (P1-25/DB25-13): IOPort
hm2/hm2_7i96s.0: registered
hm2_eth:     env "PATH=/usr/sbin:/sbin:${PATH}" iptables -A hm2-eth-rules-output -o enp4s0 -p icmp -j DROP
hm2_eth:     env "PATH=/usr/sbin:/sbin:${PATH}" iptables -A hm2-eth-rules-output -o enp4s0 -j REJECT --reject-with icmp-admin-prohibited
hm2_eth:     /sbin/sysctl -q net.ipv6.conf.enp4s0.disable_ipv6=1
Found file(REL): ./manualtoolchange.hal
Found file(REL): ./xhc_whb04b_6.hal
```

It might be more comprehensive updating the man page with some guidance. Something like that:
```
If you don't use suid binarys, you need to setup port isolation rules manually.
Assuming your host IP is 192.168.1.120, your device IP is 192.168.1.121 and your Ethernet card is enp4s0, use the following commands:
iptables -N hm2-eth-rules-output
iptables -I OUTPUT 1 -j hm2-eth-rules-output
iptables -A hm2-eth-rules-output -p udp -m udp -d 192.168.1.121 --dport 27181 -s 192.168.1.120 -j ACCEPT
iptables -A hm2-eth-rules-output -o enp4s0 -p icmp -j DROP
iptables -A hm2-eth-rules-output -o enp4s0 -j REJECT --reject-with icmp-admin-prohibited
sysctl -q net.ipv6.conf.enp4s0.disable_ipv6=1
```

hdiethelm:
A better but way more demanding solution would be to use a library instead of executing iptables. This way, the cap's are inherited and it might work directly. What I found so far:
- Cap's https://unix.stackexchange.com/questions/385109/can-you-list-iptables-as-a-non-root-user-and-why
- nftables: https://netfilter.org/projects/libnftnl/index.html
- BPF Filter: https://docs.kernel.org/bpf/libbpf/libbpf_overview.html / https://linuxvox.com/blog/how-to-install-bpftool-command-in-linux/
- iptables: https://tldp.org/HOWTO/Querying-libiptc-HOWTO/qfunction.html but it is not recommended: https://www.netfilter.org/documentation/FAQ/netfilter-faq-4.html#ss4.5

hdiethelm:
Or might be this one helps also, explicitly launch iptables with inherited cap's:
- https://stackoverflow.com/questions/78980096/process-capabilities-inheritance
- https://man7.org/linux/man-pages/man3/cap_launch.3.html

rmu75:
Re. the iptables situation, the commands could be put into it's own executable that gets CAP_NET_ADMIN capabilities. It will only need a few parameters like ip-addresses and network interfaces so should be relatively easy to get right.

grandixximo:
Pushed `be46bab12a..4ca555ebed` addressing the `--sport` and log-mangling issues with a small privileged helper.

**Approach:** new `hm2_eth_iptables` binary with `cap_net_admin,cap_dac_override+ep` file caps. It raises `cap_net_admin` into the ambient set and execs `/sbin/iptables`, so the cap actually reaches the kernel call. `hm2_eth.c` invokes it via `posix_spawnp` with structured argv (no shell, no `env PATH=...`, no log interleaving). The helper validates IPs (`inet_pton`), interface names (`if_nametoindex`), and dport range before exec.

**Per your feedback:**
- `--sport` dropped (ephemeral, broke second start). Port isolation now `dport`-only as you suggested.
- `iptables-restore` style was tempting but `iptables-nft` makes the legacy interface less attractive long-term; the helper exec keeps the diff small and works on both legacy and nft backends today. libnftnl can come in your step 2 follow-up if you want.
- The man page now carries the explicit recipe (with shell variables for IP/dport/iface) so users without setcap on the helper still have a copy-paste path. No `--sport` there either.

Also picked up your iptables cleanup commit `c226577` in the prior round.

**Tested locally:** probe/init/board/iface/cleanup full lifecycle, idempotent re-init, idempotent cleanup, IPv6 sysctl write via `cap_dac_override`, input validation rejects bad IP/iface/port. Not tested with hardware. If the offer for 7I96S testing still stands, that would close the loop.

Trade-off note: this re-introduces one setcap binary, which partially walks back the rootless goal. Scope is narrow (single helper, validated args) and `cap_net_admin` only goes ambient inside the helper, not in `rtapi_app`. Open to feedback if you want a different shape.

hdiethelm:
Just short from my side: Is this helper binary needed? hm2_eth is linked to rtapi_app when started which has (nearly) the same caps as the new helper. So all the code in the helper should also work if placed inside hm2_eth.
This aproach will need less changes, can still use --sport and doesn't need a helper, just execute iptables with inherited caps: https://github.com/LinuxCNC/linuxcnc/pull/3964#issuecomment-4322997224.
Caviat: I don't know if this works, i havn't tested it. If desired, I can create a branch testing it.

grandixximo:
Fair point. The helper was over-engineered for the iptables case. Two questions before I rip it out:

1. IPv6 disable currently uses `sysctl net.ipv6.conf.<if>.disable_ipv6=1`. That writes a 644 root-owned proc file, which needs `cap_dac_override` (rtapi_app does not have it). I think replacing it with `ip6tables -A OUTPUT -o $IFACE -j DROP` is the cleanest alternative, since it stays inside the cap_net_admin envelope. OK with that?
2. `--sport` should be safe to keep now that cleanup-on-exit (`c226577`) is in place: each run installs its own rule and removes it on exit. Agreed?

If yes to both, I will switch to in-process ambient cap raise + drop the helper binary, shorten the man-page recipe to a brief "rules are auto-installed" note, and keep the `LINUXCNC_FORCE_REALTIME`/diagnostic message work from #3928.

grandixximo:
Apologies for abusing the CI

hdiethelm:
I don't know the exact reason for the sysctl call. If it is only to block ipv6, where iptables has no effect, ip6tables should do the job. If there is an additional rason, telling the user to add a line to /etc/sysctl.conf is also an option.

grandixximo:
Agreed. Plan:

- Replace the runtime sysctl call with `ip6tables -A OUTPUT -o $IFACE -j DROP` (cap_net_admin only).
- Document the one-time `net.ipv6.conf.<if>.disable_ipv6=1` in `/etc/sysctl.conf` as an optional belt-and-braces step in the man page (prevents the stack from generating the packets in the first place; not strictly needed since ip6tables drops them).

With that I will:
1. Drop the `hm2_eth_iptables` helper binary, its Submakefile rule, the `cap_dac_override` setcap line, and the `libcap-dev` build dep added for it.
2. Move the iptables/ip6tables calls back into `hm2_eth.c`, with the ambient `cap_net_admin` raise done once at module load.
3. Restore `--sport` (cleanup-on-exit makes per-run port pinning safe).
4. Trim the man page recipe accordingly.

Will keep the `LINUXCNC_FORCE_REALTIME` / cap diagnostic message commit (`5c9fc7`); that one is orthogonal.

hdiethelm:
Sounds good to me.
I can test it again later.
BTW: Is there a mesa card simulator? It would help a lot, also for me implementing rtnet. If there is none, I could look into what would be needed to do this. A fancy aproach would be GHDL to use the VHDL code. A simple aproach just capturing some packages and replaing them with minimal logic.

grandixximo:
These are all I found, not sure any will cover exactly what you are looking for, worth a check?

- **HAL mock component (7i76e):** https://github.com/pkl42/linuxcnc_components_sim
- **litehm2 (full HM2 protocol on LiteX/RISC-V):** https://github.com/sensille/litehm2
- **hm2-rp2040 (HM2 on RP2040 microcontroller):** https://github.com/SebKuzminsky/hm2-rp2040
- **Forum thread on the HAL mock:** https://forum.linuxcnc.org/27-driver-boards/56722-simulating-a-hm2-eth-driver-e-g-mesacard-7i76e
- **HostMot2 register map:** https://github.com/LinuxCNC/hostmot2-firmware/blob/master/src/regmap
- **hm2_eth.c source (LBP16 packet structure):** https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/drivers/mesa-hostmot2/hm2_eth.c

dwrobel:
>Revives PR https://github.com/LinuxCNC/linuxcnc/pull/918 (@dwrobel) rebased onto current master, addresses

@grandixximo  You might consider also cherry-picking/rebasing/improving:
- https://github.com/dwrobel/linuxcnc/commit/1aae7667b0be0b6b8290371f934b0c5d25b3f55e _(Allow the incr_mem_usage() to be used without root privileges.)_
- https://github.com/dwrobel/linuxcnc/commit/79d0cda42c8c91b4b67f370f57e1042bac58e6f5 _(Allow the incr_io_usage() to be used without root privileges.)_
- https://github.com/dwrobel/linuxcnc/commit/dd7f9bc448db282c6d9d13487f6b4bc8fe03c80d _(Remove direct usage of inb(), outb())_

I haven't push them as I don't have any PCI based card for testing them.

grandixximo:
Cherry-picked all three: `724784b` (incr_mem_usage), `d523e0a` (incr_io_usage), `b849e1d` (remove direct inb/outb).

Builds OK locally on uspace (where `upci.c` is not compiled), but the PCI changes are non-uspace and I have no PCI Mesa card to test the runtime path. @pcw-mesa, would you be able to validate `pci_read` / `pci_write` and a PCI HostMot2 card (e.g. 5I25 / 6I25 / 7I80 / 7I92) under both setuid root and `sudo make setcap` (rootless) installs? The intent of dwrobel's three commits is that `/dev/mem` open and `iopl(3)` succeed via `cap_sys_rawio` file caps without needing `seteuid(0)`, falling back to setuid only when the cap is absent.

@hdiethelm if you also have a moment to add the PCI sanity to your retest, that would help close the loop.

Branch is `nosetuid` on grandixximo/linuxcnc; HEAD = `b849e1d`.

pcw-mesa:
I can test a PCI card sometime this week. 
Is the rootless uspace stuff in master (with sudo make setcap) or a specific branch?

hdiethelm:
> These are all I found, not sure any will cover exactly what you are looking for, worth a check?

Thanks, I have to look into these links. What would be nice is a software that behaves like a mesa card and communicates over Ethernet.

CAP_SYS_RAWIO /dev/mem access: I have just seen this now.
If you have this cap, you are only one step away from root. This allows to read/write the whole memory space. So you can even inject code into the kernel: https://blackhat.com/presentations/bh-europe-09/Lineberry/BlackHat-Europe-2009-Lineberry-code-injection-via-dev-mem.pdf

Probably there is no way around this and this PR fixes the packaging issue with setuid. But it does not improve security by much, it just makes it a bit harder to gain root access. Probably there is somewhere code that gives you root using /dev/mem.

>@hdiethelm if you also have a moment to add the PCI sanity to your retest, that would help close the loop.

I can only test ethernet mesa, I don't have a PCI mesa card.

> I can test a PCI card sometime this week. Is the rootless uspace stuff in master (with sudo make setcap) or a specific branch?

It is not yet merged (in master). It is available here: https://github.com/grandixximo/linuxcnc/tree/nosetuid

BsAtHome:
Completely basic security... If you have access to the hardware of the machine (from software), then the machine is pwned.

In the case of LinuxCNC; The issue of running rootless is not one of security as such, but it is a practice of minimizing access privilege. That way bugs will be (oh so very) slightly less impact severe to crash the system and may coredump more easily to track the error.

hdiethelm:
> Completely basic security... If you have access to the hardware of the machine (from software), then the machine is pwned.

Setuid or also setcap have the issue that any random bad person that can access any PC with linuxcnc installed (in person, over ssh, otherwise...) can exploit this to gain root if they are any issues anywhere in the code.

Due to this, I removed the suid flags on my dev workstation where I don't have real time anyway.

This PR makes it at least more difficult to gain root, so a good step in the right direction. Dropping CAP_SYS_RAWIO would be nicer but hard to impossible, so out of scope for this PR in my opinion.

Adding a linuxcnc group and only allowing users from this group running linuxcnc would future reduce the risks. But also out of scope for this PR.

rmu75:
Introducing a linuxcnc group and requiring users to be part of that group is probably a good idea. Currently, any method of remoting into a linuxcnc machine can give you root.

I'm not sure something can be done about the /dev/mem issue. Networked hardware doesn't need access to /dev/mem, pci based stuff doesn't need CAP_NET_ADMIN so capabilities that are not needed could be removed. Maybe these capabilites can be dropped after the "setup phase" of the hardware so components added later can't exploit these?

Maybe stuff that needs "priviledged" access could be separated into it's own executable / thread but that would probably introduce additional latencies.

The main problem with all these things is it will create confusion among users.

hdiethelm:
I tested again on my machine: Debian Trixie / PREEMT_RT / Mesa Ethernet
suid or setcap. Both work fine.

grandixximo:
@rodw-au Could you build a new image with this patch included, and see that it works within the live image?

NTULINUX:
I'm working on Gentoo images for LinuxCNC atm, I can include this in there. (bare-metal, not VM)

rodw-au:
> @rodw-au Could you build a new image with this patch included, and see that it works within the live image?

Actually, this test is superflous, The deployed live image is just a method of installing the Linuxcnc Debs from our repository.  So as long as the Deb build works, it will be OK. Whilst it is possible to build a live image with a RIP install or build and install the Debs from source, testing that process is not adding any value because it will work and is not different from testing from  the ~/linuxcnc-dev folder.

BsAtHome:
The test is not supposed to test the .deb as such. It is supposed to test whether the rootless design of the build works. There may be things in the install procedure that need to be adapted that are not visible unless you try to install a new system without any prior data on the drive. There can be all kinds of details lacking that our "established" systems take for granted but need some tweaking on a virgin system when rootless must work.

rodw-au:
ISO here: https://vmnet.com.au:5001/sharing/KqSeStneO 
Password: 3964 (will expire in a few days)
This build has cloned, built and installed this PR.

Be prepared for a surprise on first boot.
Just wait at the first logon screen for a while (a minute or less) for the system to reboot to apply my kernel tuning optimisations. Be warned! Its a bit unnerving!

It works on a VM. I used qemu

Note my ISO's include installation and configuration of ethercat. This one uses grandixximo's latest release of the linuxcnc-ethercat driver (need to change that when he sorts out his new repo at linuxcnc-ethercat)

There is a script `sudo purge-ethercat` which will remove ethercat.

`sudo verify` will produce a pdf report in the current folder that is a report card of the optimisations applied. This also runs a quick 5 second latency test using cyclictest.

If further ISO are needed they will be published at the same location

grandixximo:
Thank you, downloaded, but have not had the time to install it yet, haven't forgot about this, will try to get to it tomorrow

grandixximo:
@rodw-au thanks for the ISO. Boots clean in qemu after install. One surprise: `linuxcnc` reports **2.9.8** inside the live system, but this PR targets master (2.10-pre). Was that expected (ISO layers the PR rebuild on top of 2.9 packages from the vmnet repo), or did the rebuild not actually replace the installed debs? If the latter, the live system is not exercising any rootless paths under test, and the run only confirms the image boots. Could you check, and if needed rebuild so the 3964 branch debs override the 2.9 ones?

---

@BsAtHome picking up Sunday's thread on `rtapi_spawn_as_root` and the iptables path, now that I have re-read the code.

State today on this branch:

- Setuid stays the default install. `make setcap` is opt-in; the Debian path still drops `rtapi_app` and `linuxcnc_module_helper` as 4755.
- `rtapi_spawn_as_root` / `rtapi_spawnp_as_root` are bare `posix_spawn` / `posix_spawnp`. Under setuid the child is genuinely root; under filecap it inherits the user's UID plus the ambient cap set (currently CAP_NET_ADMIN only).
- The iptables path no longer uses `rtapi_spawn_as_root`. `hm2_eth.c` calls `posix_spawn("/sbin/iptables", ...)` directly and relies on the ambient CAP_NET_ADMIN raise done at rtapi_app startup. That part is settled.
- Only remaining caller of `rtapi_spawn_as_root` is `eshellf.c`, used by `hm2_rpspi.c` for `modprobe spi-bcm2835` / `rmmod`. Those need CAP_SYS_MODULE, which is not in the setcap target, so they silently fail under filecap. Under setuid they work as before.

Since setuid is the standard install, the function name is still accurate for the supported path. It only misleads under filecap. Two options:

1. **Keep the name**, update the `rtapi.h` doc comment to spell out the two-mode behaviour and note that callers needing extra caps must arrange them. Smallest diff.
2. **Rename** to `rtapi_spawn_helper` (or similar) with the same doc note. More honest but more churn across `rtapi.h`, `uspace_rtapi_main.cc`, `eshellf.c`, `tests/uspace/spawnv-root/`.

I lean option 1 since setuid is the default and a rename risks signalling a semantic change that did not happen. Happy to do option 2 if you prefer, or a third shape (split into two functions, drop the wrapper, etc.). Let me know which way to go.

rodw-au:
Sorry thats odd. This is what you should have got. Maybe I stuffed up the deployment to the download folder.
```
# This clones only the PR #3964 code into a folder named 'linuxcnc-pr3964'
git clone --depth 1 --single-branch --branch pull/3964/head https://github.com/LinuxCNC/linuxcnc.git "$TARGET_DIR"
cd "$TARGET_DIR"
```

I'll rebuild and check this end before sharing again

BsAtHome:
I think it would be easier to remove the modprobe, rmmod and insmod.
The only test really required is test for the module(s) loaded at startup and then it should simply fail hard when detected. Then just write a message that having kernel spi enabled with hm2_rpspi/hm2_spix is not compatible and should be disabled using raspi-config.

grandixximo:
@BsAtHome agreed, the detect-and-fail approach is cleaner than trying to plumb CAP_SYS_MODULE through. Following that, the scope cascades nicely:

- `hm2_rpspi.c`, `spix_rpi3.c`, `spix_rpi5.c`: replace the `modprobe`/`rmmod` calls with a one-shot check (fopen `/proc/modules`, grep for `spi_bcm2835` / `dw_spi_mmio`). Hard-fail at load with a message pointing the user at `raspi-config` (or the equivalent module blacklist for the rpi5 case).
- `eshellf.c` / `eshellf.h`: no remaining callers, delete.
- `rtapi_spawn_as_root` / `rtapi_spawnp_as_root` in `uspace_rtapi_main.cc` and `rtapi.h`: the only non-test caller goes away with `eshellf`; the API can be removed along with `tests/uspace/spawnv-root/`. That also removes the "spawn as root" footgun under filecap entirely.

One detail before I cut: the rpi5 path (`spix_rpi5.c`) blacklists `dw_spi_mmio`, which I don't think `raspi-config` exposes directly. Are you OK with the message there pointing at `/etc/modprobe.d/blacklist-linuxcnc.conf` (or similar) instead of raspi-config? Otherwise I'll word it as "disable the kernel SPI driver, see the rpi5 section of the hm2_spix(9) man page" and document the blacklist recipe there.

If that direction looks right, I'll push the cascade in one commit per driver plus a cleanup commit for `eshellf` + `rtapi_spawn_as_root`.

BsAtHome:
Make a reference to the manual page(s) in the message and fix the docs :-)

I don't think that removing the spawn as root completely would be prudent, yet. Maybe it can be declared deprecated and wait a moment for people to complain. I'm not sure if there are folks with local components that call out as root.

A differentiator may be to set rootless build in the configure stage and enable deprecation of the spawn as root functions when building rootless. That would give locally built components a way out, for now. This needs to be thought through a bit more... Suggestions?

BTW, isn't there also a "open as root" file operation?

grandixximo:
@BsAtHome all four points taken.

1. **Man pages and docs**: I'll update `hm2_rpspi(9)` and `hm2_spix(9)` with the disable-kernel-SPI recipe (raspi-config for the rpi3 path, `/etc/modprobe.d` blacklist for rpi5 `dw_spi_mmio`), and have the load-time failure message in each driver point users at those pages by name. Same treatment for the AsciiDoc integrators manual where the drivers are listed.

2. **Deprecate, don't remove**: agreed. I'll mark `rtapi_spawn_as_root` and `rtapi_spawnp_as_root` with `__attribute__((deprecated("...")))`, keep the implementation, and leave `tests/uspace/spawnv-root/` in place so the UID semantics test still runs. Same treatment for `rtapi_open_as_root`, which I missed in the first pass: 8 callers (hal_bb_gpio, hm2_rpspi, spix_rpi3, spix_rpi5, hm2_spix), all opening `/dev/mem` or sysfs clock files. Under filecap the `WITH_ROOT` fsuid lift is a no-op since euid is already the user, so the function silently leans on file chmod/chown, it needs the same deprecation note and eventually the same rewire.

3. **Configure-time rootless gate for the deprecation**: I think the cleanest shape is a single `--enable-rootless` switch in `configure.ac` that defines `LINUXCNC_ROOTLESS_BUILD` and is consumed in `rtapi.h` to gate the `deprecated` attribute on the three `_as_root` prototypes. Out-of-tree components built against a setuid install see no warnings, the moment a packager flips to rootless (and the day a distro defaults to it), they get a compile-time nudge. The runtime binary is unchanged so we don't fork ABI. Side benefits: same flag can later gate the install rule choice (`make install` picks setcap target when rootless build, setuid target otherwise) instead of the user remembering `sudo make setcap`. Does that match what you had in mind, or were you thinking of something finer-grained?

4. **`rtapi_open_as_root`**: covered above. Adding it to the deprecation set in the same commit.

Net plan for this PR if the above lands:

- Driver cleanup: hm2_rpspi, spix_rpi3, spix_rpi5 detect-and-fail on conflicting kernel SPI module, with man-page-referenced messages.
- Delete `eshellf.c` / `eshellf.h` (no callers left after the driver change).
- Add `--enable-rootless` to configure, define `LINUXCNC_ROOTLESS_BUILD`.
- Mark `rtapi_spawn_as_root`, `rtapi_spawnp_as_root`, `rtapi_open_as_root` deprecated when `LINUXCNC_ROOTLESS_BUILD` is defined.
- Update `hm2_rpspi(9)`, `hm2_spix(9)`, and the integrators manual sections that cover those drivers.

I'll hold off pushing until you confirm the `--enable-rootless` shape.

BsAtHome:
The first two points can go ahead without further discussion, I think.

Point 3 I need to think about for a moment.

I'm actually not sure how point 4 is to be solved. Linux has a separate FS uid and that gets toggled (using the `WithRoot` construct). The docs state that setuid also sets the fsuid. I don't know why it is necessary when the binary is setuid root and the code actually does a setresuid(), then the filesystem should be all accessible. The docs for setfsuid(2) actually mark it deprecated. Very confusing.

@hdiethelm, you were several places in that code and actually added a WithRoot use to Xenomai EVL. Any specific reason why?

grandixximo:
@BsAtHome will work on 1 and 2.

On point 4, AFAIK `WithRoot` is just toggling fsuid, not euid. Under setuid `main()` does `setresuid(0, 0, user)` then `setfsuid(user)`, so the process stays root for syscalls but acts as the user for FS ops (keeps PID/log files user-owned). `WithRoot` flips fsuid to 0 for the few cases that need root FS access (`/dev/mem`, `/dev/evl/...`).

Under filecap there's no fsuid to toggle, so it's a no-op and `rtapi_open_as_root` falls through to DAC (chmod/chown permissions). That's what the deprecation note should say.

The deprecation note I want to add on the `_as_root` APIs is essentially "this function only does what the name says under a setuid build, under a filecap build it falls through to DAC and may quietly succeed-or-fail based on file permissions".

@hdiethelm tagging on the EVL `WithRoot` in `wrapper()`. AFAIK it's there because `evl_attach_self` opens root-owned files under `/dev/evl/`. Correct me if it's another reason.

BsAtHome:
Check. I clearly had my mind another place when I read it. Thanks for the detail.

So, if I understand correctly, then you'd need CAP_SYS_RAWIO for /dev/mem. Is that enough for /dev/evl? Otherwise you'd need CAP_SETUID to use set[fs]uid and friends.

Besides, CAP_IPC_LOCK for mmap/mlockall and friends .

grandixximo:
@BsAtHome The way I think about it: two layers per file, both must pass.

1. **DAC** (mode bits, owner, group) gates `open()`.
2. **In-driver cap check** gates what you can do through the open fd.

Cap alone does not bypass DAC; DAC alone does not bypass the driver check.

Applied:

- `/dev/mem` (`crw-r----- root kmem`): DAC blocks first (no `cap_dac_override` in setcap, by design). Even past DAC, the driver wants `CAP_SYS_RAWIO`, and `CONFIG_STRICT_DEVMEM` blocks SoC peripheral ranges anyway. Dead end for rootless on SoC GPIO.
- `/dev/evl`: pure DAC, no cap involved. Needs an `evl` group + udev rule, per hdiethelm's libevl Debian work.
- `set[fs]uid` / `WithRoot`: filecap path is a no-op. We deliberately don't grant `CAP_SETUID` since `setresuid(0,0,0)` from it is root-equivalent.
- `CAP_IPC_LOCK`: needed for `mlockall` past 64 MB `RLIMIT_MEMLOCK`. Already in setcap.

**Scope I had in mind for this PR**: the cases where DAC+cap can be made to work today without sliding back toward root. iptables via ambient `cap_net_admin`, PCI BARs via `/sys/bus/pci/.../resource*` + plugdev udev rule, `cpu_dma_latency` via udev. Deprecate `_as_root` APIs and document the mode-dependent behaviour. Keep setuid as the unchanged default install path.

**Out of scope here, candidates for follow-up PRs**:

1. SoC GPIO/SPI off `/dev/mem` onto `/dev/gpiomem` and/or device-specific mmap, for `hal_bb_gpio`, `hm2_rpspi`, `spix_rpi3`. This is the bulk of the remaining `rtapi_open_as_root` callers.
2. Xenomai/EVL rootless: evl group + libevl udev rules (hdiethelm's roadmap step 4).
3. RTAI / Xenomai 3 rootless: same shape, separate group + udev. Probably lowest priority.
4. Once 1 + 2 + 3 land, the runtime `geteuid()==0` gates in `detect_xenomai_evl/xenomai/rtai` come out and `_as_root` can move from deprecation to removal.

**Final destination**: rootless is the default install, setuid is a legacy fallback for the RTAI path until that gets its own rootless story. `rtapi_app` runs unprivileged with the four caps we have now plus whatever group memberships the per-driver udev rules expose.

If you'd rather rescope this PR to also tackle item 1 (so we ship a path where SoC GPIO works rootless out of the box), say so and I'll fold it in.

hdiethelm:
> @rodw-au thanks for the ISO. Boots clean in qemu after install. One surprise: `linuxcnc` reports **2.9.8** inside the live system, but this PR targets master (2.10-pre). Was that expected (ISO layers the PR rebuild on top of 2.9 packages from the vmnet repo), or did the rebuild not actually replace the installed debs? If the latter, the live system is not exercising any rootless paths under test, and the run only confirms the image boots. Could you check, and if needed rebuild so the 3964 branch debs override the 2.9 ones?

There is a bug inside `scripts/get-version-from-git` which reports every branch as 2.9.8. I noticed this while working on the CI but did not crate an issue yet. To much issues always popping up... ;-)

hdiethelm:
^ So, I created an issue: https://github.com/LinuxCNC/linuxcnc/issues/4030

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -840,13 +846,23 @@ static void signal_handler(int sig, siginfo_t * /*si*/, void * /*uctx*/) {
...
+    // Best-effort raise of the soft cap to the hard cap.  Fails on
+    // unprivileged processes without CAP_SYS_RESOURCE or without a
+    // matching setrlimit; CAP_IPC_LOCK alone lets mlockall succeed
+    // regardless of the rlimit, so ignoring the failure is safe.
+    struct rlimit limit;
+    if (getrlimit(RLIMIT_MEMLOCK, &limit) == 0) {
+        limit.rlim_cur = limit.rlim_max;
+        if (setrlimit(RLIMIT_MEMLOCK, &limit) < 0)
+            rtapi_print_msg(RTAPI_MSG_DBG,
+                "setrlimit(RLIMIT_MEMLOCK) failed: %s\n", strerror(errno));
```
BsAtHome: Is this a debug message? Looks more like a genuine error or is it just a warning in this context?

BsAtHome:
This seems to be heading in the right direction.

hdiethelm:
> @hdiethelm tagging on the EVL `WithRoot` in `wrapper()`. AFAIK it's there because `evl_attach_self` opens root-owned files under `/dev/evl/`. Correct me if it's another reason.

Yes, both xenomai3 and 4 need access to certain files. Both can be done by a group membership. To do that, I first have to update my debian packages and then add a group membership test there. Still WIP, i got into a few side projects. I think xenomai is not-(yet) used that often. So this branch can be merged anyway and I can later create a rootless xenomai branch as soon as my xenomai packages support this.

src/rtapi/uspace_rtapi_main.cc
grandixximo: I think `RTAPI_MSG_DBG` is the right pick here.

The call is a best-effort soft-cap raise. Its failure has zero functional consequence as long as `CAP_IPC_LOCK` is held, because `mlockall` ignores `RLIMIT_MEMLOCK` entirely with that cap. Under filecap (no `cap_sys_resource` in the setcap line) the setrlimit is *expected* to fail; promoting it to WARN would print every startup on a correctly configured rootless install.

The real check is `mlockall` itself a few lines down, which already logs at `RTAPI_MSG_WARN`:

```
mlockall failed: %s. Realtime latency may suffer.
```

That covers the case the user actually cares about. The setrlimit line is diagnostic for the rare case where someone is debugging why mlockall failed and wants to know whether the rlimit raise was also denied.

Happy to bump to INFO if you'd rather have it visible at the default level, but WARN feels like it would cry wolf on the rootless path.

BsAtHome: But when it is expected to fail, then the message is very misleading. It should then state "setrlimit.... failed as expected:...".

grandixximo: @BsAtHome fair point on the misleading text. Before I rephrase, want to check if a different fix would land better.

We already use libcap in this file (`raise_net_admin_ambient`, the fallback diagnostic block in `makeApp`), so the cleaner move is to query `CAP_IPC_LOCK` upfront and only log when the failure actually matters:

```c
static void configure_memory() {
    bool have_ipc_lock = false;
#ifdef __linux__
    cap_t caps = cap_get_proc();
    if (caps) {
        cap_flag_value_t v;
        if (cap_get_flag(caps, CAP_IPC_LOCK, CAP_EFFECTIVE, &v) == 0 && v == CAP_SET)
            have_ipc_lock = true;
        cap_free(caps);
    }
#endif
    struct rlimit limit;
    if (getrlimit(RLIMIT_MEMLOCK, &limit) == 0) {
        limit.rlim_cur = limit.rlim_max;
        if (setrlimit(RLIMIT_MEMLOCK, &limit) < 0 && !have_ipc_lock)
            rtapi_print_msg(RTAPI_MSG_WARN,
                "setrlimit(RLIMIT_MEMLOCK) failed and CAP_IPC_LOCK not held: %s. "
                "mlockall will likely fail.\n",
                strerror(errno));
    }
    ...
```

- Filecap + `cap_ipc_lock`: silent (expected, harmless).
- Setuid root: rlimit raise succeeds, silent.
- Plain user, no caps: genuine WARN with actionable info.
- Setuid that somehow dropped the cap: same WARN, surfaces the regression.

This drops the misleading-when-expected log and adds real signal when actually broken. Also: cap-presence check is now in three places in this file, so a small `static bool cap_held(cap_value_t)` helper would DRY all three. Want me do that as part of the change?

If you'd rather I keep it minimal and just rephrase the existing DBG message as you suggested (`"setrlimit(...) denied (expected without CAP_SYS_RESOURCE): %s"` or similar), say so and I'll do that instead.

BsAtHome: Clear diagnostics when something is broken is always preferable. The simplest solution is often (not always) the best, you know, KISS. That said, the <i>correct</i> solution gets you multiple KISSes ;-)

grandixximo: @BsAtHome done. WARN now fires only when the raise fails and `CAP_IPC_LOCK` is not held; silent on the expected paths.

Tried a `cap_held()` helper first but reverted: only two pure-query sites (`raise_net_admin_ambient` mutates, can't share), not worth the indirection.

grandixximo:
Forgot to mention, rod did give me linuxcnc ISO built on this, and everything was working as expected.

hdiethelm:
Was this tested with PCI cards?
I would go for merge. This is open since some time and if new issues arrive, there is still the suid mode.

jepler:
Thanks, this looks like a lot of great work. I am out of touch, so please bear in mind that what I'm saying here could be out of date or incorrect.

One small note I had in looking at the diff: The old iptables would prevent all other app access to the hm2-eth connection, by making a rule on the **source port** of the outgoing packets. The intent is to prevent e.g., mesaflash from also communicating with the card during a LinuxCNC session. Unless I misunderstand, the 'static' firewall instructions do not provide this protection.

I also didn't see what the "fate" of the SIOCSARP ioctl was. Without this, packet delivery (transmission from Linux to hm2-eth board) can be delayed due to the device's entry disappearing from the ARP cache.  A ctrl-f search of the diff view didn't show any hits, but without net admin I think the call will fail.

NTULINUX:
I really think `nftables` support should be added since not all systems use `iptables`. I don't even have `iptables` installed on any of my systems.

An `nftables` conversion looks like so:

```
table inet filter {
    chain OUTPUT {
        type filter hook output priority filter; policy accept;
        ip saddr $HOST_IP ip daddr $BOARD_IP udp dport $BOARD_DPORT accept
        oifname $IFACE ip protocol icmp drop
        oifname $IFACE reject with icmp type admin-prohibited
        oifname $IFACE ip6 version drop
    }
}
```

Commands are as such:

```
nft add table inet filter
nft add chain inet filter OUTPUT '{ type filter hook output priority filter; policy accept; }'
nft add rule inet filter OUTPUT ip saddr "$HOST_IP" ip daddr "$BOARD_IP" udp dport "$BOARD_DPORT" accept
nft add rule inet filter OUTPUT oifname "$IFACE" ip protocol icmp drop
nft add rule inet filter OUTPUT oifname "$IFACE" reject with icmp type admin-prohibited
nft add rule inet filter OUTPUT oifname "$IFACE" ip6 version drop
```

Not tested on mesa hardware..

grandixximo:
Thanks both, picked these up in two follow-ups.

@NTULINUX, nftables support is #4111. New `firewall=auto|iptables|nft|none` parameter; `auto` prefers iptables when usable and falls back to `nft`, so iptables-less systems get isolation out of the box. The nft rules live in a private `table inet hm2_eth` (flush/delete never touches your other rules), and `inet` covers v4 and v6 in one chain. The man page gets a manual `nft` recipe too. I have no Mesa ethernet card, so a test on real hardware would be very welcome.

@jepler, both notes were on point.

Source port: the runtime auto-install still matches it (`udp sport` for nft, `--sport` for iptables), so a live session keeps the same-board isolation against e.g. mesaflash. The gap is only in the manual static recipe, and it is inherent: the source port is ephemeral and assigned after connect, so a hand-written rule cannot know it. The iptables static recipe omits it for the same reason. Pinning it would mean binding a fixed source port in the driver, which seemed like more than it is worth for a fallback. Happy to revisit.

SIOCSARP: real rootless regression. Without `CAP_NET_ADMIN` the ioctl returned `-errno` and aborted the whole board load, making rootless-without-setcap unusable for ethernet boards. #4112 makes it best-effort like the firewall path: warn, fall back to dynamic ARP, continue. The pinned entry and your latency point are restored once setcap or setuid is in place.

hdiethelm:
> I also didn't see what the "fate" of the SIOCSARP ioctl was. Without this, packet delivery (transmission from Linux to hm2-eth board) can be delayed due to the device's entry disappearing from the ARP cache. A ctrl-f search of the diff view didn't show any hits, but without net admin I think the call will fail.

You don't see the fate due to nothing was changed there. setcap sets `CAP_NET_ADMIN`, so this call works as before.

Issue #3922, opened 2026-04-11
Missing files while creating docs in a RIP install


rodw-au:
This is not a big issue because I never read the generated docs and go to the web site instead.
But: When compiling for RIP with documents included,  I see the following warnings reported

```
cp: cannot stat '../docs/src/drivers/:images/pluto-pinout.png': No such file or directory
../docs/html/gui/gmoccapy.html:649: parser error : Opening and ending tag mismatch: figure line 647 and a
</a>
    ^
../docs/html/gui/gmoccapy.html:651: parser error : Opening and ending tag mismatch: a line 646 and figure
</figure>
         ^
unable to parse ../docs/html/gui/gmoccapy.html
cp: cannot stat '../docs/src/gui/:images/tkemc-interp.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/halshow-watch-tab-1.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/stepgen-block-diag.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/stepgen-type2-4.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/stepgen-type5-10.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/stepgen-type11-14.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/encoder-block-diag.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/pid-block-diag.png': No such file or directory
```

I am guessing there are going to be some online docs missing their images.

grandixximo:
@rodw-au see if my PR fixes your issue

hansu:
Resolved by #3965

PR #3965, opened 2026-04-25
docs: fix leading-colon image src from index-entry/image:: paragraph fusion (#3922)

grandixximo:
## Summary

Fixes warnings reported in #3922:

```
cp: cannot stat '../docs/src/drivers/:images/pluto-pinout.png': No such file or directory
cp: cannot stat '../docs/src/gui/:images/tkemc-interp.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/halshow-watch-tab-1.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/stepgen-block-diag.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/stepgen-type2-4.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/stepgen-type5-10.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/stepgen-type11-14.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/encoder-block-diag.png': No such file or directory
cp: cannot stat '../docs/src/hal/:images/pid-block-diag.png': No such file or directory
```

## Root cause

When an asciidoc index macro `(((entry)))` sits directly above a block `image::path[...]` macro with no intervening blank line, asciidoc parses the two lines as a single paragraph. The block macro then degrades to inline `image:` syntax, and the second `:` of `image::` is consumed as part of the target, producing `src=":images/foo.png"` in the rendered HTML.

`docs/src/Submakefile` (the `.html-images-stamp` rule) extracts those `src` values with `xsltproc` and feeds them to `cp -f`, which is why the warnings show the literal `<dir>/:images/<file>.png` paths.

## Fix

Insert a blank line after each offending `(((...)))` so the following `image::` line is recognized as a block macro. Nine call sites across four files:

- `docs/src/drivers/pluto-p.adoc` (pluto-pinout)
- `docs/src/gui/tklinuxcnc.adoc` (tkemc-interp)
- `docs/src/hal/halshow.adoc` (halshow-watch-tab-1)
- `docs/src/hal/rtcomps.adoc` (stepgen-block-diag, stepgen-type2-4, stepgen-type5-10, stepgen-type11-14, encoder-block-diag, pid-block-diag)

## Note

The `gmoccapy.html` parser error (`figure`/`a` tag mismatch around lines 647-651) reported in the same issue is a separate sub-bug with a different cause and is filed as a follow-up issue.

## Test plan

- [x] Rebuilt all four affected `.adoc` files locally with `asciidoc -b xhtml11`; no `src=":...` strings in output.
- [x] Confirmed all nine images resolve to `src="images/<file>.png"` after the fix.
- [ ] Reviewer: full doc build (`make` in `src/`) should no longer print the nine `cp: cannot stat ':images/...'` warnings.


docs/src/hal/halshow.adoc
```diff
@@ -142,6 +142,7 @@ The following figure shows this canvas with several pins.
 [[fig:halshow-watch-tab-1]]
 .Halshow: Watch Tab
 (((Halshow: Watch Tab)))
+
 image::images/halshow-watch-tab-1.png["Halshow Watch Tab",align="center"]
```
hansu: This change makes the caption to a headline. I suggest this fix in conjunction with https://github.com/LinuxCNC/linuxcnc/issues/3966

The same for the other changes.



```suggestion
(((Halshow: Watch Tab)))

[[fig:halshow-watch-tab-1]]
.Halshow: Watch Tab
image::images/halshow-watch-tab-1.png["Halshow Watch Tab",align="center"]

```

grandixximo: Good catch. Pushed 2a047362ec: moved the index entry above the `[[fig:..]]` + `.caption block` at all 7 figure sites. Verified locally: caption now renders as figure title (Figure N. ...) instead of a headline; image src and anchor cross-references stay clean. The pluto-pinout and tkemc-interp sites already followed the pattern after the original blank-line fix, so no further change there.

grandixximo: Pulled #3966 into this PR per your suggestion: commit 3c34067574 drops the custom `xhtml11-image.conf` so the shipped asciidoc `[image-blockmacro]` takes over. That eliminates the `<figure>`/`<a>` cross-nesting at the gmoccapy `link=` image and gives well-formed `<a>` handling for all other image sites. Captions still render as figure titles, cross-references still resolve, and the original `src=":images/..."` fix from #3922 is unaffected (verified by rebuilding all four touched files plus gmoccapy.adoc with `xmllint --noout` clean).

hansu:
Looks good, thanks!

PR #3718, opened 2026-01-20
Architecture LinuxCNC-block-diagram - update

zz912:
I tried to update the block diagram of the LinuxCNC architecture. I would like the image to include both the ideal state and the current unwanted state that has arisen.

LinuxCNC-block-diagram

I am currently observing that some problems are being solved at the frontend level, instead of being solved at the background level.

When I started with LinuxCNC, I was missing a display where the HAL is. So I drew it.

If anyone knows how other parts of LCNC should work that would be good to draw. Feel free to send them to me drawn with a pencil on paper and photographed with a mobile phone. I can then draw them in the image.

I will be happy if there is a discussion about my proposal.

PR #3967, opened 2026-04-25
docs: correct HAL caption in motion controller diagram

### LINKED ISSUE #3843: Docs: Incorrect info in architecture diagram
https://github.com/LinuxCNC/linuxcnc/blob/master/docs/src/code/LinuxCNC-motion-controller.png

Suggests that the HAL code is in "EXTINTF.C", and it's not, it's in various hal*.* files.

grandixximo:
Diagram referenced obsolete "EXTINTF.H" / "EXT????.C" filenames. HAL is implemented across multiple hal_*.c files; updated caption in both full and -small PNG variants.

Closes #3843


zz912:
I wanted repair this picture too:
https://github.com/LinuxCNC/linuxcnc/pull/3718

Unfortunately I found that it is beyond my capabilities. I think I could do it with AI, but at least this year I will not have time for it.

This image is very outdated, for example it does not contain JOINTS and AXIS. It only shows AXIS. So the question is whether it is worth maintaining it.

smoe:
Closed upon acceptance of #3718.

Issue #3793, opened 2026-02-07
TclX issue on Fedora 43


NTULINUX:
The `2.9` branch didn't want to configure due to a missing TclX error for me but `master` did.

Upon installing LinuxCNC `master` via my custom-made RPM (a system install, not rip) I noticed that axis and gmoccapy are both working, latency-test works but not the histogram:

can't find package Tclx
    while executing
"package require Tclx"
    (procedure "load_packages" line 2)
    invoked from within
"load_packages"
    invoked from within
"if ![info exists ::LH(start)] {
  set_defaults
  config
  progress "Loading packages"
  load_packages
  signal trap SIGINT finish
  progress "Making g..."
    (file "/usr/bin/latency-histogram" line 842)

rmu75:
after chatting on IRC, seems to be related to tcl 9.0 (no tclx yet for tcl 9.0). but doesn't work for tcl8.x either because it doesn't find blt. probably won't fix.

BsAtHome:
The issue is that Fedora moved to Tcl/Tk 9. That includes the python tkinter interface, which is linked with Tcl/Tk 9 and that _must_ match LinuxCNC's.

I have not (yet) backported the changes required to run 2.9 again on Fedora. I thought I could get away with it... However, this issue is a confirmation that it probably is a good idea that I do backport the changes :-)

Edit: and I will look into the Tclx issue too.

NTULINUX:
Thanks, yeah I was going to rebuild blt and bwidget against tcl/tk8.6 from fedora instead but ending up going down a rabbit hole. I'm going to wait until things settle down before trying again.

Another problem..
```
fedora@fedora-kde:~$ tclsh
% package require Tclx     
can't find package Tclx
% exit
```
```
fedora@fedora-kde:~$ tclsh8
% package require Tclx
8.4
% exit
```
```
fedora@fedora-kde:~$ tclsh
% package require BWidget
1.10.1
% exit
```
```
fedora@fedora-kde:~$ tclsh8
% package require BWidget
can't find package BWidget
% exit
```

TclX only works with tcl8.6, but BWidget only works with tcl9...

I opened a bug report for upstream:

https://bugzilla.redhat.com/show_bug.cgi?id=2437570

NTULINUX:
Pushed the issue along further by making my own bwidget package, now hitting:

```
checking python version... OK
checking match between tk and Tkinter versions... TCL mismatch: 8.6 vs 9.0
[...]

PR #3968, opened 2026-04-25
latency-histogram: port BLT chart to Tk canvas for Tcl 9

grandixximo:
## Summary

- BLT is dead since 2002 with no Tcl/Tk 9 port; this blocks `latency-histogram` on Fedora 42+.
- Replaces `blt::barchart` with a small in-script `lh_chart` namespace on plain Tk canvas; call sites stay identical.
- Also stubs TclX (`signal` no-op) and replaces `blt::winop snap` with a Tk 8.7+/9 native photo, ImageMagick fallback.

Refs #3793.

## Test plan

- [x] Debian 13 / Tcl 8.6 RIP build, both threads with setuid, side-by-side compared with system BLT version.
- [x] Fedora 43 / Tcl 9.x (would appreciate @NTULINUX testing this).

### DIFF
--- a/scripts/latency-histogram
+++ b/scripts/latency-histogram
@@ -26,6 +26,500 @@
 # Mu sign is Unicode 00b5
 set ::MICROSEC \u00b5s
 
+#-----------------------------------------------------------------------
+# lh_chart: minimal canvas-based bar chart implementing the subset of
+# blt::barchart used by this script. Goal: visual + behavioral parity
+# without depending on BLT (which has no Tcl/Tk 9 port).
+#
+# Public API (used at $w via command rename + dispatch):
+#   $w axis    configure x|y -min -max -majorticks -logscale -hide -showticks
+#   $w element create|configure NAME -xdata -ydata -fg -bg -barwidth -stipple
+#   $w element exists NAME
+#   $w legend  configure -hide 0|1
+# Constructor:
+#   lh_chart::create $w -title T -width W -height H \
+#                       -plotbackground COLOR -cursor C
+#-----------------------------------------------------------------------
+namespace eval lh_chart {
+    variable state
+    variable stipple_map
+    array set stipple_map {}
+}
+
+proc lh_chart::install_stipples {} {
+    variable stipple_map
+    if {[info exists stipple_map(_installed)]} return
+    set dir /tmp/lh_chart_stipples_[pid]
+    catch {file mkdir $dir}
+    # Tk's XBM reader is picky: needs static char (signed, not unsigned)
+    # and the data block in K&R-style continuation form with a trailing
+    # closing-brace-semicolon on its own line. One-liners are rejected.
+    foreach {name bits} {
+        pbmap {0xe3 0xf1 0xf8 0x7c 0x3e 0x1f 0x8f 0xc7}
+        nbmap {0xc7 0x8f 0x1f 0x3e 0x7c 0xf8 0xf1 0xe3}
+    } {
+        set fp [open $dir/$name.xbm w]
+        puts $fp "#define ${name}_width 8"
+        puts $fp "#define ${name}_height 8"
+        puts $fp "static char ${name}_bits\[\] = \{"
+        puts $fp "  [join $bits {, }]\};"
+        close $fp
+        set stipple_map($name) "@$dir/$name.xbm"
+    }
+    set stipple_map(_installed) 1
+}
+
+proc lh_chart::create {w args} {
+    variable state
+    install_stipples
+    array set opts {
+        -title           ""
+        -width           480
+        -height          384
+        -plotbackground  honeydew1
+        -cursor          arrow
+    }
+    array set opts $args
+    canvas $w \
+        -width  $opts(-width) \
+        -height $opts(-height) \
+        -bg     "#d9d9d9" \
+        -bd 0 -highlightthickness 0 \
+        -cursor $opts(-cursor)
+    set state($w,title)        $opts(-title)
+    set state($w,width)        $opts(-width)
+    set state($w,height)       $opts(-height)
+    set state($w,plotbg)       $opts(-plotbackground)
+    set state($w,xmin)         -1.0
+    set state($w,xmax)          1.0
+    set state($w,ymin)          0.0
+    set state($w,ymax)          1.0
+    set state($w,ylogscale)     0
+    set state($w,xticks)        {}
+    set state($w,elements)      {}
+    set state($w,legend_hide)   1
+    set state($w,dirty)         1
+    set state($w,redraw_pending) 0
+    bind $w <Configure> [list lh_chart::on_configure $w]
+    rename ::$w ::lh_chart::_orig_$w
+    proc ::$w {args} "::lh_chart::dispatch $w {*}\$args"
+    lh_chart::schedule_redraw $w
+    return $w
+}
+
+proc lh_chart::dispatch {w args} {
+    set sub [lindex $args 0]
+    set rest [lrange $args 1 end]
+    switch -- $sub {
+        axis    { return [lh_chart::cmd_axis    $w {*}$rest] }
+        element { return [lh_chart::cmd_element $w {*}$rest] }
+        legend  { return [lh_chart::cmd_legend  $w {*}$rest] }
+        default { return [::lh_chart::_orig_$w {*}$args] }
+    }
+}
+
+proc lh_chart::cmd_axis {w sub which args} {
+    variable state
+    if {$sub ne "configure"} { error "lh_chart axis: unsupported subcommand $sub" }
+    array set opts $args
+    if {$which eq "x"} {
+        if {[info exists opts(-min)]}        { set state($w,xmin)   $opts(-min) }
+        if {[info exists opts(-max)]}        { set state($w,xmax)   $opts(-max) }
+        if {[info exists opts(-majorticks)]} { set state($w,xticks) $opts(-majorticks) }
+    } elseif {$which eq "y"} {
+        if {[info exists opts(-logscale)]}   { set state($w,ylogscale) $opts(-logscale) }
+        if {[info exists opts(-min)]}        { set state($w,ymin) $opts(-min) }
+        if {[info exists opts(-max)]}        { set state($w,ymax) $opts(-max) }
+    }
+    schedule_redraw $w
+}
+
+proc lh_chart::cmd_element {w op name args} {
+    variable state
+    variable stipple_map
+    switch -- $op {
+        exists {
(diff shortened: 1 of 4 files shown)


NTULINUX:
This commit fixes: https://github.com/LinuxCNC/linuxcnc/issues/3793

Can confirm on Tcl/Tk 9 setup. Removed TclX and blt packages, still works!

NTULINUX (approved):

BsAtHome:
I think you will need to adapt the code to prevent the RT/non-RT race of the data transfer from RT into non-RT.

Both latencybins and timedelta were dropped from my version and rewrote them as timedeltalatch.comp and latencybinstream.comp (see attached zip). These new versions solve the problem by introducing data latching. This also solves the (exorbitant) delay you get when you wanted to read the bins from RT because you had to wait one cycle for each +/- bin pair after setting the index (and in that time, the histogram changed every cycle).
[latchcomps.zip](https://github.com/user-attachments/files/27096758/latchcomps.zip)

The latency bins are conveyed using the HAL stream mechanism, which is why I needed to fix that first. The data is pushed into the stream in RT as one set when you set the latch pin. Non-RT can then look and wait until the right amount of values are in the stream and read them all. Below are the code snippets I used to get the data (python). You'll have to adapt it to Tcl...

Getting the latency bin data from the stream:
```python
    def readBins(self):
        if not self.stream:
            return
        mod = self.name + "hist."    # 'basehist', 'servohist'
        dly = self.period / 1e9
        if dly < 1e-3: dly = 1e-3
        hal.set_p(mod+"stream", "1")   # Signal to stream
        cnt = 0;
        req = 2*self.nbins + 1 + 2
        # Wait for the data to arrive
        while True:
            time.sleep(dly)
            depth = self.stream.depth
            if depth >= req:
                break
            cnt += 1
            if cnt >= 100:
                err = hal.get_value(mod+"stream-error")
                if err:
                    warning("Stream data for {} was not sent".format(self.name))
                    return
                else:
                    error("Failed to get enough stream data ({} < {}) for {} thread".format(depth, req, self.name))

        # Get the data from the steam
        cnt = 1
        maxi = -1
        bins = []
        for i in range(0, req):
            if not self.stream.readable:
                error("Stream became not readable while reading element {} from {} elements".format(i, req))
            tup = self.stream.read()[0]
            if tup > maxi:
                maxi = tup
            if tup <= 0:
                tup = 1
            bins.append(tup)
        if cnt <= 0:
            error("No samples counted")
#        cnt /= 100  # We want range 1...100
        cnt = maxi / 100
        self.bins = [bins[i] / cnt for i in range(0, req)]
        self.timeval = hal.get_value(mod+"stream-time")
        if self.timeval >= self.period:
            self.timewarn += 1
        # Get the statistics
        self.latavg  = hal.get_value(mod+"latency-avg")
        self.latmin  = hal.get_value(mod+"latency-min")
        self.latmax  = hal.get_value(mod+"latency-max")
        self.stdev   = math.sqrt(math.fabs(hal.get_value(mod+"variance")))
        hal.set_p(mod+"stream", "0")   # Reset for next rising edge
```

Getting the time delta values:
```python
    def readStats(self):
        mod = self.name + "time."
        dly = self.period / 1000000000
        if dly < 1e-3:  # Don't sleep less than one millisecond
            dly = 1e-3
        hal.set_p(mod+"latch", "1")
        cnt = 0
        while cnt < 100:
            time.sleep(dly)
            if hal.get_value(mod+"latched"):
                break
        if cnt >= 100:
            error("Failed to retrieve {} statistics, 'latched' never became active".format(self.name))
        hal.set_p(mod+"latch", "0")
        self.count= hal.get_value(mod+"count")
        self.pmin = hal.get_value(mod+"period-min")
        self.pmax = hal.get_value(mod+"period-max")
        self.psum = hal.get_value(mod+"period-sum")
        self.pssq = hal.get_value(mod+"period-ssq")
        self.jmin = hal.get_value(mod+"jitter-min")
        self.jmax = hal.get_value(mod+"jitter-max")
        self.jsum = hal.get_value(mod+"jitter-sum")
        self.jssq = hal.get_value(mod+"jitter-ssq")
        self.amin = hal.get_value(mod+"absjit-min")
        self.amax = hal.get_value(mod+"absjit-max")
        self.asum = hal.get_value(mod+"absjit-sum")
        self.assq = hal.get_value(mod+"absjit-ssq")
        self.pstdev = self.stdev(self.psum, self.pssq, self.count)
        self.jstdev = self.stdev(self.jsum, self.jssq, self.count)
        self.astdev = self.stdev(self.asum, self.assq, self.count)
        if self.count > 0:
            self.pavg = self.psum / self.count
            self.javg = self.jsum / self.count
            self.aavg = self.asum / self.count
        else:
            self.pavg = self.javg = self.aavg = 0.0
            # The min/max would otherwise show +/- MAXINT
            self.pmax = self.pmin = self.jmax = self.jmin = self.amax = self.amin = 0
        if self.pmax > self.pmaxall: self.pmaxall = self.pmax
        if self.pmin < self.pminall: self.pminall = self.pmin
        if self.jmax > self.jmaxall: self.jmaxall = self.jmax
        if self.jmin < self.jminall: self.jminall = self.jmin
        if self.amax > self.amaxall: self.amaxall = self.amax
        if self.amin < self.aminall: self.aminall = self.amin
```

grandixximo:
Pushed [f600c1d](https://github.com/LinuxCNC/linuxcnc/pull/3968/commits/f600c1d828).

- Brings in @BsAtHome `latencybinstream.comp` as the default. Confirmed the race: bins transferred one cycle at a time get smeared across hundreds of ms, stats desync from bins.
- Tcl has no `hal_stream` binding, so a small Python helper (`scripts/latency-bin-reader`) drains the FIFO on demand over a stdin/stdout pipe. One helper per thread, kept open for the session.
- Old `latencybins` path kept behind `--legacy` for direct comparison.

Did not pull `timedeltalatch.comp` in this round, the script's existing min/max/sdev come from `latencybinstream` pins (latched on stream rising edge) so no extra comp needed for the GUI. Happy to add it in a follow-up if there's appetite for period/jitter readouts.

Tested on Debian 13 / Tcl 8.6 with both default and `--legacy`.

grandixximo:
Force-pushed [0ff9d2f](https://github.com/LinuxCNC/linuxcnc/pull/3968/commits/0ff9d2f822) (squashed).

Replaced the Python sidecar with a native Tcl hal_stream binding in halsh.c (attach/read/drain/depth/detach). Pure Tcl now, no extra runtime dep. latency-histogram drains the latencybinstream FIFO directly.

src/hal/utils/halsh.c
```diff
@@ -86,6 +89,155 @@ static int halCmd(ClientData cd, Tcl_Interp *interp, int argc, const char **argv
...
+                            hal_stream_t *s, int idx,
+                            hal_stream_data_u v) {
+    char buf[64];
+    switch(hal_stream_element_type(s, idx)) {
+        case HAL_BIT:   snprintf(buf, sizeof(buf), "%d", v.b ? 1 : 0); break;
+        case HAL_FLOAT: snprintf(buf, sizeof(buf), "%g", v.f); break;
+        case HAL_S32:   snprintf(buf, sizeof(buf), "%lld", (long long)v.s); break;
+        case HAL_U32:   snprintf(buf, sizeof(buf), "%llu", (unsigned long long)v.u); break;
+        case HAL_S64:   snprintf(buf, sizeof(buf), "%lld", (long long)v.s); break;
+        case HAL_U64:   snprintf(buf, sizeof(buf), "%llu", (unsigned long long)v.u); break;
```
BsAtHome: This is wrong. The values for HAL_S64 is `v.l` and for HAL_U64 is `v.k`. See `hal_stream_data_u` in hal/hal.h for details.

.gitignore
```diff
@@ -43,7 +43,8 @@ cscope.*
 oprofile*
 *.log
 position.txt
-# *.9
+# Generated by halcompile from .comp files
+docs/man/man9/*.9
```
BsAtHome: Aren't there a few non-generated files in here?

src/hal/utils/halsh.c
```diff
@@ -19,6 +19,9 @@
 #include <stdio.h>
 #include <tcl.h>
 #include "halcmd.h"
+#include "hal.h"
+
```
BsAtHome: This should address the global hal header using <> because it is not an internal part of hal, but a layer on top.

scripts/latency-histogram
```diff
@@ -635,45 +1177,88 @@ proc make_chart {} {
...
+    set tries 0
+    while {[hal_stream depth $h] < $need} {
+      after 1
+      incr tries
+      if {$tries > 250} {
+        hal setp $::LH($thd,name).stream 0
+        puts "$thd: stream timeout (depth [hal_stream depth $h] < $need)"
+        return
+      }
+    }
```
BsAtHome, line 1244: You should wait to see if there are enough values in the stream before reading. The queue is large enough (should be). When the right amount is in the queue, then you also have a better guarantee to be in sync. You can then also check that you are at depth level zero after reading.

grandixximo:
Thanks for the review @BsAtHome. [8d0c092](https://github.com/LinuxCNC/linuxcnc/pull/3968/commits/8d0c092805) addresses the four points:

- halsh.c: HAL_S64 / HAL_U64 now use `v.l` / `v.k`. Good catch, the s32/s64 truncation would have bitten any wider stream.
- halsh.c: `#include <hal.h>` for the public layer.
- script: in addition to the depth-reach poll before reading, also verify `hal_stream depth == 0` after `drain` so we abort on out-of-phase reads, and check `stream-error` pin on timeout.
- .gitignore: I went through all 204 `.9` files in `docs/man/man9/` and each has either a `.comp`, a `.adoc`, or a kinematics `.c` source. None are tracked. Happy to switch to ignoring the directory if you prefer.

grandixximo:
Reverted root .gitignore change. Added `man9/latencybinstream.9` to `docs/man/.gitignore` matching the existing per-file convention. [a3271af](https://github.com/LinuxCNC/linuxcnc/pull/3968/commits/a3271af9d2)

There are no non-generated files, maybe another PR could do the switch, from write one by one to a simple *.9

src/hal/utils/halsh.c
grandixximo: fixed, will try to keep an eye on it.

.gitignore
grandixximo: No, but probably another PR should fix this, I just don't like to see the file as untracked in my file tree when I switch between branches...
I'll survive...

src/hal/utils/halsh.c
grandixximo: Fixed, thanks.

Issue #662, opened 2019-11-22
Queue is not empty


grandixximo:
## Here are the steps I follow to reproduce the issue:

 1. add custom.hal to 9axis configuration
 2. open 9axis configuration and load the probe.ngc
 3. start and let it run may take a while for the error to come up

## This is what I expected to happen:

code runs forever without errors

## This is what happened instead:

after a while (may take minutes to hours) error "Queue is not empty after probing" comes up

## It worked properly before this:
I think this is a persistent bug since older versions of Linuxcnc

## Information about my hardware and software:

 * I am using this Linux distribution and version (often, shown by `lsb_release -a`):Ubuntu 18.04.3 LTS
 * I am using this kernel version (shown by `uname -a`): 4.18.0-15-generic
 * I am running ...
   * [ ] A binary version from linuxcnc.org (including buildbot.linuxcnc.org)
   * [x] A binary I built myself
   * [ ] A binary version from some other source besides linuxcnc.org
 * I am using this LinuxCNC version (shown in package manager or, for git versions, `scripts/get-version-from-git`): 2.8.0~pre1 
 * I am using this user interface (GUI) (e.g., AXIS, Touchy, gmoccapy, etc): Axis
 * I am using this interface hardware vendor and chipset (e.g., parallel port, ethernet port, FPGA card): simulator

Issue #263, opened 2017-04-19
ERROR: emcStatus->motion.traj.queue during probing


Cromaglious:
Getting an error during probing using smartprobe.ngc and a flat file using (PROBEFILE...

I get further the slower I go

Error message is

Queue is not empty after probing
is the actual error
emc/task/emctask.cc 397: interp_error: Queue is not empty after probing

LOG(G1 X21.000000 Y2.300000 Z-5.156492)  was last output before error
add something so you know how far through it was
Y 24th loop and X at 240*22+210

using F360 for smartprobe and F240 for flat file.

Motherboard ITX atom1.6Ghz quadcore 2GB RAM
Linuxcnc 2.8.0-Pre1-2905
Linux Kernel 4.9.13 AMD64 PreemptRT
Debian 8.7 Jesse 64bit

Parallel port BOB
JMC-motor.com iHSS57-36-20 Integrated Closed Loop Stepper
Max Jog 6500mm/m 200steps per mm
stepgen at 20Khz
SmartProbe.ngc
(Configuration section)
G21   (Metric)
F360   (probe speed)

#<X_Left>=0  (X start)
#<StepX>=0.1 (X increment)
#<X_right>=240  (X count)

#<Y_Front>=0 (Y start)
#<StepY>=0.1 (Y increment)
#<Y_Rear>=240  (Y count)

#<Z_Clear>=0 (Z safety)
#<Z_Probe>=-15 (Z probe)
(End configuration section)

grandixximo:
I can confirm this.
I have my own custom probing code too, and sometimes i get this error too, it's weird because it may happen once, and then never happen again, even running the same exact code, really weird and hard to debug

rmu75:
seems to be duplicate of #3650

PR #3971, opened 2026-04-26
task: tighten AUTO read-ahead wait-clear to match MDI (#3650)

### LINKED ISSUE #3650: Queue is not empty after probing
Version 2.9.7.
When cyclically calling the command queue

G38.3 Z1
G38.5 Z30
G38.3 Y[#<_y> + 1]
G38.3 Z1

I constantly get an error "Queue is not empty after probing"
The topic is discussed on the forum - https://forum.linuxcnc.org/38-general-linuxcnc-questions/51443-how-to-fix-queue-is-not-empty-after-probing?start=10#338958

grandixximo:
Fixes the long-standing "Queue is not empty after probing" error.

`mdi_execute_hook` has, since 2012, gated wait-clear on `execState==DONE && motion.traj.queue==0 && io.status==DONE`. `readahead_reading` (AUTO mode) checks only `execState==DONE`, on the assumption that the prior `EMC_TASK_PLAN_SYNCH` precondition (`WAITING_FOR_MOTION_AND_IO`) had already drained motion.

That assumption breaks across cycles: `emcTaskIssueCommand` writes `EMC_TRAJ_PROBE` to motion via shared memory, then the same task cycle's `emcMotionUpdate` may snapshot `emcmotStatus` before the servo loop has added the new segment. `motion.status` reads `DONE` (stale), the SYNCH precondition transitions immediately, and the next `Interp::_read` runs `read_inputs` against a queue that is now non-empty. CHKS fires.

Patch: add the same two conjuncts to the AUTO check.

Closes #3650, #662, #263.

Complementary to #3537 (already merged): that fix handled probe re-trip during deceleration. This one handles the inter-command snapshot-lag race that lets the same family of bugs leak through.

cc @andypugh, @DauntlessAq, @rmu75, @BsAtHome: would value review on the wait-clear semantics.
cc @tiket18, @Cromaglious: could you test against your reproducing workload? Can use built artifacts for testing

### Test plan

- [ ] Cyclic G38.3/G38.5 sim loop (>20k iterations), error gone
- [ ] Single-probe + tool-setter still latch \`#5061..#5069\` correctly
- [ ] Reporter validation against XXYYZ scan + smartprobe workloads
- [ ] 2.9 backport candidate after master soak

### DIFF
--- a/src/emc/task/emctaskmain.cc
+++ b/src/emc/task/emctaskmain.cc
@@ -494,11 +494,18 @@ void readahead_reading(void)
                     int count = 0;
 interpret_again:
 		    if (emcTaskPlanIsWait()) {
-			// delay reading of next line until all is done
+			// delay reading of next line until all is done.
+			// Match the stricter check used by mdi_execute_hook:
+			// trusting only execState==DONE can race when motion
+			// hasn't yet processed a just-issued command (e.g. a
+			// G38 probe), letting the next read see a non-empty
+			// queue.
 			if (interp_list.len() == 0 &&
 			    emcTaskCommand == 0 &&
 			    emcStatus->task.execState ==
-			    EMC_TASK_EXEC::DONE) {
+			    EMC_TASK_EXEC::DONE &&
+			    emcStatus->motion.traj.queue == 0 &&
+			    emcStatus->io.status == RCS_STATUS::DONE) {
 			    emcTaskPlanClearWait();
 			 }
 		    } else {


rmu75:
Not sure I really qualify to comment, but explanation makes sense.

grandixximo:
Need testers, I will try to post in the forum as well

grandixximo:
I'll send a 2.9 PR just so I can have debs built, users on the forum want 2.9 to test the fix

DauntlessAq:
I'm slightly confused, does this issue occur when the probe move starts, or when it ends?

Because the error message implies it is at the end, that is to say that the probe move (`EMC_TRAJ_PROBE`) is still in the queue when it should have been removed?

Aka, currently it is possible that a state sync check at the end of the probe move passed when it should not do so. This would be because `execState==DONE` was set before the move was REMOVED from the queue, due to a race condition.
So just checking `execState==DONE` isn't completely reliable, hence your more stringent checks being a fix.

This makes sense to me as an explanation that also matches the error message (but of course it might not be correct)

&nbsp;

But when you say that:

> emcTaskIssueCommand writes EMC_TRAJ_PROBE to motion via shared memory, then the same task cycle's emcMotionUpdate may snapshot emcmotStatus before the servo loop has added the new segment

Are you saying that the error actually occurs during a state sync BEFORE the probe move takes place, instead of AFTER? Because you're talking about a sync where `EMC_TRAJ_PROBE` is the move which is written?

Or are you talking about the END of the probe move, and the following move just happens to be `EMC_TRAJ_PROBE` in your testing because you are using consecutive probe moves in your testing (e.g. both towards and away from the part)?

But either way, you're saying that the race condition is that the move is ADDED prematurely, not that the condition passed prematurely because the move has not been fully REMOVED?

Surely the move wouldn't be added prematurely because the addition of the move to the interpreter queue would itself be gated behind the state sync check of `execState==DONE` so that the moves aren't added to the queue until the sync has occurred?
Else the next move would always be added during a probe move, so this "Queue is not empty after probing" error would happen every probe move?

&nbsp;

Or are you saying that _is_ the race condition, that upon completion of the probe move there is a brief period where moves can be added to the queue before the sync is commanded?

And this can allow the next move to be added to the queue before the sync takes place, causing the current sync logic to fail because the next move has been added to the queue despite the machine otherwise being in a state where the sync should pass?

In which case, while tightening up this logic might be a fix, we'd ideally also want to fix the issue closer to its orgin by preventing moves which require a sync beforehand from being added to the queue before a sync is confirmed to have occurred?

&nbsp;

Just to show where the error message originates:

This is where the "Queue is not empty after probing" message is defined:
https://github.com/LinuxCNC/linuxcnc/blob/2378edb2ab7b8eeeb9470780b5adb8dad647efbf/src/emc/rs274ngc/rs274ngc_return.hh#L133

And this is how the message is called:
https://github.com/LinuxCNC/linuxcnc/blob/2378edb2ab7b8eeeb9470780b5adb8dad647efbf/src/emc/rs274ngc/rs274ngc_pre.cc#L1448

    if (settings->probe_flag) {
	CHKS((GET_EXTERNAL_QUEUE_EMPTY() == 0),
	     NCE_QUEUE_IS_NOT_EMPTY_AFTER_PROBING);
	set_probe_data(&_setup);
	settings->probe_flag = false;
    }

The message is triggered when the interpreter is commanded to read an input, if the probe_flag is true and the queue is not empty then the error occurs. So the error definitely seems to occur AFTER a probe move.

grandixximo:
Good catch, my wording conflated two phases. Let me restate, with the caveat that this is my **theoretical** model of the race and still needs validation against the reporter workloads.

The CHKS fires at the END of probe N, on the next `read_inputs` call when reading line N+1. You're right that `execState==DONE` was set before motion truly drained.

The mechanism I think is upstream of that. Trace:

1. Cycle K: task pulls `EMC_TRAJ_PROBE` from `interp_list`, issues to motion via shared memory in step 3. End of cycle K, `emcMotionUpdate` runs `usrmotReadEmcmotStatus`.
2. If the servo loop hasn't run since the NML write, `emcmotStatus.depth` and the INPOS bit still reflect the pre-probe state. Snapshot says `queue==0, motion.status==DONE`.
3. Cycle K+1: task pulls `TASK_PLAN_SYNCH` (queued by the probe's `INTERP_EXECUTE_FINISH` path). Precondition \`WAITING_FOR_MOTION_AND_IO\` reads the stale snapshot and transitions to DONE immediately, then issues SYNCH same cycle.
4. Cycle K+2: \`interp_list==0, cmd==0, execState==DONE\`. AUTO wait-clear releases. By now motion is processing the probe, queue grows.
5. Cycle K+3: read line N+1, \`read_inputs\` CHKS reads queue from latest snapshot, sees probe N still in flight, fires \`NCE_QUEUE_IS_NOT_EMPTY_AFTER_PROBING\`.

So the proposed race: SYNCH's precondition raced through using stale "motion idle" data, and AUTO wait-clear didn't double-check. MDI's wait-clear has, since 2012, included \`motion.traj.queue==0 && io.status==DONE\` and doesn't hit this. Patch brings AUTO to the same standard.

Probe N+1 in the user's program is incidental, the race would fire on whatever line N+1 happens to be. User's cyclic G38.x just makes it more visible because every line is queue-busted.

> Surely the move wouldn't be added prematurely because the addition of the move to the interpreter queue would itself be gated behind the state sync check of execState==DONE so that the moves aren't added to the queue until the sync has occurred? Else the next move would always be added during a probe move, so this "Queue is not empty after probing" error would happen every probe move?

Right, normally the SYNCH precondition handles it. The hypothesis is that the race only fires when servo hasn't run between NML write and snapshot, which is timing-sensitive on \`[TASK]CYCLE_TIME\` vs servo period. Most cycles align fine, occasionally don't. Matches the "tens of thousands of probes pass, then random fail" reports across #3650, #662, #263.

To stress: this is theory until someone with a reproducing setup confirms the patch fixes it. If reporters here (#3650 tiket18, #263 Cromaglious) can rerun their failing workloads against this patch and the error stops firing, that's strong evidence the model is right and this is the fix. If it still fires, the model is wrong and we go deeper.

> we'd ideally also want to fix the issue closer to its origin by preventing moves which require a sync beforehand from being added to the queue before a sync is confirmed to have occurred

Agreed there's a deeper fix available. Could tighten the SYNCH precondition itself, or the \`EMC_TRAJ_PROBE\` postcondition, to require explicit queue+io drain rather than just \`motion.status==DONE\`. This patch is the minimum intervention matching the 2012 MDI gate. Happy to explore the deeper refactor once we have empirical confirmation that this surface fix actually closes the reported failures.

grandixximo:
cc @mhaberler: you wrote the 2012 MDI wait-clear gate ([81f105b477](https://github.com/LinuxCNC/linuxcnc/commit/81f105b477)) that this PR mirrors into the AUTO path. Do you remember whether the `motion.traj.queue==0 && io.status==DONE` conjuncts were added for a specific reproducer, or as defense-in-depth? Any context you have on why AUTO was left looser would be valuable for understanding whether the same fix is sufficient there.

mhaberler:
sorry, no recollection

andypugh:
Trying to solve this problem I have already found a race condition, though I am not sure it is the same one. 
I am not sure at this point if that ever made it to a public branch, so can't at this point tell if it is the same thing in the same place as yours. 
That stalled when the reporter said that he had fixed it in hardware...

PR #3974, opened 2026-04-28
docs/getting-started -document RIP vs install copy behavior in pickconfig

grandixximo:
## Summary

Documents the Configuration Selector copy-on-launch behavior in
`docs/src/getting-started/running-linuxcnc.adoc`:

- copy fires on package installs (deb, rpm) where sample configs live
  under a non-writable system path
- RIP builds run sample configs in place because the source tree is
  writable by the user who built it
- `debug_pickconfig=1` env var forces a copy from a RIP build

Behavior is unchanged since 2006 (commit 03ebae23, "prompt the user
to copy a read-only configuration"); this was previously only
documented in `tcl/bin/pickconfig.tcl` source comments.

Closes the documentation gap raised in #3924 by @rodw-au.

## Test plan

- [x] asciidoc renders the new sections without errors
- [x] visual check of rendered output

### DIFF
--- a/docs/src/getting-started/running-linuxcnc.adoc
+++ b/docs/src/getting-started/running-linuxcnc.adoc
@@ -153,18 +153,18 @@ that is used to describe a specific configuration.
 
 === When the copy happens
 
-The Configuration Selector decides whether to copy a sample
-configuration to ~/linuxcnc/configs based on where the source
-configuration lives:
+For the following cases the Configuration Selector copies the chosen sample
+configuration to ~/linuxcnc/configs:
 
 * *Package install (deb, rpm, distro packages)*: sample configurations
-  live under a system path. The selector copies the configuration to
-  ~/linuxcnc/configs before running it, so edits never touch the
-  system-wide files.
-* *Run-In-Place (RIP) build*: the source tree is normally writable by
-  the user who built it, so the selector runs the configuration
-  directly from the source tree without copying. Edits made through
-  the configuration apply to the files in the RIP tree.
+  live under a system path and are not writabe.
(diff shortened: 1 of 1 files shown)


rodw-au:
Looks good!

docs/src/getting-started/running-linuxcnc.adoc
```diff
@@ -151,6 +151,35 @@ for a configuration to a new subdirectory of ~/linuxcnc/configs
...
+
+* *Package install (deb, rpm, distro packages)*: sample configurations
+  live under a system path. The selector copies the configuration to
+  ~/linuxcnc/configs before running it, so edits never touch the
+  system-wide files.
+* *Run-In-Place (RIP) build*: the source tree is normally writable by
+  the user who built it, so the selector runs the configuration
+  directly from the source tree without copying. Edits made through
+  the configuration apply to the files in the RIP tree.
+
```
hansu:
```suggestion
For the following cases the Configuration Selector copies the chosen sample
configuration to ~/linuxcnc/configs:

* *Package install (deb, rpm, distro packages)*: sample configurations
  live under a system path and are not writabe.
* *The file is not in the directory tree of `CONFIG_DIR`* (environmental variable)
* *The file is in a directory included in `LINUXCNC_AUX_EXAMPLES`* (environmental variable)

For *Run-In-Place (RIP) builds* the source tree is normally writable by
the user who built it, so the selector runs the configuration
directly from the source tree without copying. Edits made through
the configuration apply to the files in the RIP tree.

```

hansu:
Good idea to add the info there :+1: 
But then I would like to mention all cases like suggested above.

Issue #3859, opened 2026-03-12
hm2_modbus needs testing


JTrantow:
(I have questions at the bottom, I would appreciate input.)

Issue#3849 was caused by two bugs in the hm2_modbus driver.  

- BUG1 Was caused by detection and improper handling of ENOMSG during communication.
- BUG2 calc_ifdelay() was wrong for baudrate > 19200. Fixing this probably avoids BUG1 in normal operating conditions.

The fix for these bugs is contained in PR#3857 (2.9), PR$3852 (master).
**However**, there is no automatic test for hm2_modbus to ensure these problems or similar problems never creep in again.

Existing tests:

- tests/hm2-idrom Shows how to use the hm2-test module and test patterns to partially emulate a mesa board.
- tests/mb2hal Similar driver to hm2_modbus but looks like the tests simply check the pin names???

Adding tests for hm_modbus is complicated by the fact that it is a driver that works with the hostmot2, hm2_eth or similar driver, and a fpga configuration that supports pktuart.  

I used mesaflash --rpo to create a file with all that hostmot2 variables from a 7i96(7i96d_7i85p_pktv3.bit).
I added a test_pattern 15 to use this memory in hm2_test. 
    loadrt hostmot2
    loadrt hm2_test test_pattern=15
    loadrt hm2_modbus  ports=hm2_test.0.pktuart.0 mbccbs=CommanderSK.mbccb

BUG2 can tested by loading .mccb files with various communication settings and looking at the results:
Parameters:
Owner   Type  Dir                 Value  Name
    10  u32   RO             0x00009600  hm2_modbus.0.baudrate
    10  u32   RO             0x00000002  hm2_modbus.0.drivedelay
    10  u32   RO             0x0000001D  hm2_modbus.0.icdelay
    10  u32   RO             0x00000000  hm2_modbus.0.parity
    10  s32   RW                      0  hm2_modbus.0.process.tmax
    10  bit   RO                  FALSE  hm2_modbus.0.process.tmax-increased
    10  u32   RO             0x00000043  hm2_modbus.0.rxdelay
    10  u32   RO             0x00000002  hm2_modbus.0.stopbits
    10  u32   RO             0x00000045  hm2_modbus.0.txdelay

BUG1 was caused by communication delays violating constraints. This can be spoofed in hm2_test by modifying hm2_test queue read and write functions to return dummy communication values at different rates controllable by halcmds. This should be a different test than the BUG2 tests. 

QUESTIONS:
[...]

PR #3975, opened 2026-04-28
2.9 issue#3859

JTrantow:
This PR extends the hm2_test component and provides unit tests for the BUG2 portion of #3859. To open a modbus component for simulation the hm2_test component emulates enough of a 7i96 with pktuart to load the modbus component. 

A more general solution would be to implement a user component that could load an arbitrary fpga dump into the test pattern memory. But this is enough to test the modbus delays for various bauds and parity conditions.

Further development will be required before hw2_test can be used to test the BUG1 portion of $3859.


src/hal/drivers/mesa-hostmot2/hm2_test.h
```diff
@@ -31,3 +31,786 @@ typedef struct {
     hm2_lowlevel_io_t llio;
 } hm2_test_t;
 
+/*
+    This is the memory dump from a 7i96 fpga with pktuart.
+    This is a temporary fix to allow testing of hm2-modbus/modbus-Params.
+    Replace this with ability to load an arbitrary fpga image (using user mode???).
+*/
+const uint32_t config_memory_dump[] = {
+0xFFFFFFFF,0xFFFFFFFF,0xFFFFFFFF,0xFFFFFFFF,0xFFFFFFFF,0xFFFFFFFF,0xFFFFFFFF,0xFFFFFFFF,
```
BsAtHome, line 40: Headers should not be used to embed data. Use a separate C-file you link in to define the data.

Secondly, can this data not be derived from the original bit-file(s) to program the card?

tests/hm2-modbus/modbus-Params/test.sh
```diff
@@ -0,0 +1,131 @@
...
+#
+# Test pattern 15 uses a memory dump from a 7i96 as hostmod2 variables.
+# This will get replaced using hm_test2_usr for any mesa firmware variables.
+#
+TEST_PATTERN=15
+export TEST_PATTERN
+#
+# Convert each mbcc source file to binary.
+#
+for MBCCS in $(ls *.mbccs)
```
BsAtHome, line 14: Why call `ls`?

tests/hm2-modbus/modbus-Params/test.sh
```diff
@@ -0,0 +1,131 @@
...
+do
+    MBCCB=$(echo "$MBCCS" | sed 's/mbccs$/mbccb/')
```
BsAtHome, line 16: This is usually done with `basename` (and `dirname` if necessary) constructs.

tests/hm2-modbus/modbus-Params/test.sh
```diff
@@ -0,0 +1,131 @@
...
+# Override icdelay.
+#
+# test5 
+#
+echo "hm2_modbus delay test." >> halrun-stdout 
+#
+# Run HAL with each test .mbccb
+#
+TEST_PATTERN=15
+for MBCCB in $(ls *.mbccb)
```
BsAtHome, line 36: Again superfluous `ls`

tests/hm2-modbus/modbus-Params/test.sh
```diff
@@ -0,0 +1,131 @@
...
+    TEST_MBCCB=$MBCCB
+    export TEST_MBCCB
+    export TEST_PATTERN
+
+    halrun -f modbus_test.hal >halrun-stdout 2>halrun-stderr
+
+    arr=($(grep "hm2_modbus.0.baudrate" halrun-stdout))
+    baudrate=$(echo "${arr[3]}" | sed 's/0x0\+/0/') 
+    baudrate=$((16#$baudrate))
+    echo "    baudrate = $baudrate"
```
BsAtHome, line 49: Wouldn't this be easier done using `awk`? Then you can do all in a simple script.

.gitignore
```diff
@@ -63,3 +63,6 @@ docs/help/vi/*
 docs/help/zh_CN/*
 configs/**/*.tbl.bak
 configs/**/halshow.preferences
+/tests/hm2-modbus/modbus-Params/*.mbccb
+/tests/hm2-modbus/modbus-Params/halrun-stderr
+/tests/hm2-modbus/modbus-Params/halrun-stdout
```
BsAtHome, line 68: These do not belong in the top-level .gitignore file. These need to go into a local file at `tests/hm2-modus/.gitignore` Also, they are absolute paths that should be relative.

tests/hm2-modbus/modbus-Params/test.sh
```diff
@@ -0,0 +1,131 @@
+#!/bin/bash
+
```
BsAtHome, line 2: Why are all these files located in a subdirectory `modbus-Params`? There is only one test and those are generally not put into subdirectories. Are you planning on adding more? which?

(and directory/file names should generally be lower case).

tests/hm2-modbus/modbus-Params/dump_config.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+#
+# This script will dump the current configuration memory to the command line.
+# Use ./dump_config.sh >> test_pattern15.h to capture your configuration to a file. Replace 7i96 with your firmware name.
+
+# Want to use --rpo from 0x0000 - 0x6100.
```
BsAtHome, line 7: This and the other dump script will only work if you have the specific card and it is on the network.

The more versatile way is to create the config from the bit-file and make that loadable by the hm2_test component. Then you don't have any static config in the test component either.

andypugh:
@JTrantow Are you able to address these comments? Or maybe mark this PR as draft?

JTrantow:
Away from computers until Sept.

PR #3983, opened 2026-04-29
CI Improvements: Add RTAI build

hdiethelm:
In this PR, I will try to improve the CI

There will be a few pushes, including some that should fail on purpose to test if everything works as desired. Just tell me if I abuse the CI to much and I will find a different solution.

Edit: I think it makes sense to merge this and continue in a separate PR so RTAI is not broken again in a few weeks.

Issue #3756, opened 2026-01-29
Add tests starting GUIs, likely falling back to xvfb for it


andypugh:
See discussion in this separate issue: https://github.com/LinuxCNC/linuxcnc/issues/3391#issuecomment-3709882799

smoe:
When starting LinuxCNC, this typically comes with a GUI. But anything that needs to open a Window typically terminates when there is no X display server running. However, when executed after "xvfb-run", there is no such complaint. This way it should be possible to get our GUIs up, perform various actions, and possibly also to terminate again in a controlled manner.

I think it was @petterreinholdtsen who prepared such tests (without Xvfb) in debian/tests, which I would very much like to see developed further.

grandixximo:
Picking this up. Plan:

**Phase 1 (smoke):** xvfb-run wrapper test per GUI (axis, qtdragon, qtvcp variants, gmoccapy, touchy). Launch with a sim config, wait for main window, send Estop release via `linuxcnc.command`, assert running state via `linuxcnc.stat`, clean shutdown. ~10s per test.

**Phase 2 (functional):** load a small G-code file via `linuxcnc.command.program_open` + `auto(RUN)`, verify final position via `linuxcnc.stat.position`. Cross-check GUI state via `xdotool` or AT-SPI where needed. ~30s per test.

**Phase 3 (optional):** screenshot or short video on failure, uploaded as CI artifact.

Starting point: petterreinholdtsen's `debian/tests` scripts (per @smoe's note). Will adapt them to run under `tests/ui/` so they go through the standard `runtests` harness.

Coordinating dep additions with @hdiethelm in #3984 (xvfb, xauth, xdotool, pytest-qt, mesa-utils). Branching from master so PRs land independently.

Bertho's open question (smoke vs functional): I'd start with smoke to prove the infra works, then layer functional on top. Object?

PR #3984, opened 2026-04-30
CI improvemens: General improvements

hdiethelm:
This is the follow up to https://github.com/LinuxCNC/linuxcnc/pull/3983

It will be more in depth changes so riskier and will take some time.

There will be a few pushes, including some that should fail on purpose to test if everything works as desired. Just tell me if I abuse the CI to much and I will find a different solution.

It is experimental for now but I need a PR so the CI runs.

### DIFF
--- a/.github/scripts/add-linuxcnc-repository.sh
+++ b/.github/scripts/add-linuxcnc-repository.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+set -eu #Needed so CI fails when anything is wrong
+set -x
+
+DIST=$1
+
+echo "deb [arch=amd64,arm64 signed-by=/etc/apt/keyrings/linuxcnc.gpg] https://www.linuxcnc.org/ $DIST base" | sudo tee /etc/apt/sources.list.d/linuxcnc.list > /dev/null
+case $DIST in
+	'buster' | 'bullseye' | 'bookworm')
+                gpg --homedir "${PWD}/gnupg" --export 3CB9FD148F374FEF | sudo tee /etc/apt/keyrings/linuxcnc.gpg > /dev/null
+                ;;
+        *)
+                GPGTMP=$(mktemp -d /tmp/.gnupgXXXXXX)
+                gpg --homedir "$GPGTMP" --keyserver hkp://keyserver.ubuntu.com --recv-key e43b5a8e78cc2927
+                gpg --homedir "$GPGTMP" --export 'LinuxCNC Archive Signing Key' | sudo tee /etc/apt/keyrings/linuxcnc.gpg > /dev/null
+                ;;
+esac
+sudo apt-get --quiet update
--- a/.github/scripts/build-doc.sh
+++ b/.github/scripts/build-doc.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+set -eu #Needed so CI fails when anything is wrong
+set -x
+
+cd src
+./autogen.sh
+./configure --disable-check-runtime-deps --enable-build-documentation=html
+make -O -j$((1+$(nproc))) manpages
+make -O -j$((1+$(nproc))) translateddocs
+make -O -j$((1+$(nproc))) docs
+# Note that the package build covers html docs
--- a/.github/scripts/build-package-arch.sh
+++ b/.github/scripts/build-package-arch.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+set -eu #Needed so CI fails when anything is wrong
+set -x
+
+debian/configure
+debian/update-dch-from-git
+scripts/get-version-from-git | sed -re 's/^v(.*)$/\1/' >| VERSION; cat VERSION
+git diff
+apt-get --yes build-dep --arch-only .
+debuild -us -uc --build=any
--- a/.github/scripts/build-package-indep.sh
+++ b/.github/scripts/build-package-indep.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+set -eu #Needed so CI fails when anything is wrong
+set -x
+
+debian/configure
+debian/update-dch-from-git
+scripts/get-version-from-git | sed -re 's/^v(.*)$/\1/' >| VERSION; cat VERSION
+git diff
+apt-get --yes build-dep --indep-only .
+debuild -us -uc --build=source,all
--- a/.github/scripts/build-rip.sh
+++ b/.github/scripts/build-rip.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+set -eu #Needed so CI fails when anything is wrong
+set -x
+
+#Any arguments are passed to configure
+
+cd src
+./autogen.sh
+./configure "$@" --disable-check-runtime-deps --enable-werror
+make -O -j$((1+$(nproc))) default pycheck V=1
--- a/.github/scripts/install-deps.sh
+++ b/.github/scripts/install-deps.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+set -eu #Needed so CI fails when anything is wrong
+set -x
+
+sudo apt-get --quiet update
+sudo apt-get install --yes --no-install-recommends devscripts equivs build-essential lintian clang
+debian/configure
+sudo apt-get --yes build-dep .
--- a/.github/scripts/install-rtai.sh
+++ b/.github/scripts/install-rtai.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+set -eu #Needed so CI fails when anything is wrong
+set -x
+
+DIST=$1
+
+if false; then
+    #To install the RTAI deb's from linuxcnc base
+    .github/scripts/add-linuxcnc-repository.sh "$DIST"
+    sudo apt-get --yes install linux-headers-5.4.279-rtai-amd64 linux-image-5.4.279-rtai-amd64 rtai-modules-5.4.279
+else
+    #To install the RTAI deb's from NTULINUX git
+    TMPDIR=$(mktemp -d)
+    (
+        cd "$TMPDIR"
+        curl --no-progress-meter -fLO https://github.com/NTULINUX/RTAI/releases/download/v5.3.4/linux-headers-5.4.302-rtai-amd64_5.4.302-rtai-amd64-1_amd64.deb
+        curl --no-progress-meter -fLO https://github.com/NTULINUX/RTAI/releases/download/v5.3.4/linux-image-5.4.302-rtai-amd64_5.4.302-rtai-amd64-1_amd64.deb
+        curl --no-progress-meter -fLO https://github.com/NTULINUX/RTAI/releases/download/v5.3.4/rtai-modules-5.4.302_5.3.4-linuxcnc_amd64.deb
+    )
+    sudo dpkg -i \
+    "$TMPDIR/linux-headers-5.4.302-rtai-amd64_5.4.302-rtai-amd64-1_amd64.deb" \
+    "$TMPDIR/linux-image-5.4.302-rtai-amd64_5.4.302-rtai-amd64-1_amd64.deb" \
+    "$TMPDIR/rtai-modules-5.4.302_5.3.4-linuxcnc_amd64.deb"
+    rm -rf "$TMPDIR"
+fi
--- a/.github/scripts/verify-clean-repo.sh
+++ b/.github/scripts/verify-clean-repo.sh
@@ -1,0 +1,37 @@
+#!/bin/sh
+
+set -eu #Needed so CI fails when anything is wrong
+
+#To let the script fail if git status has any issues
+#If a command in $() fails, the script just continues
+git status -u --porcelain -- "$@" > /dev/null
+
+if [ $# -gt 0 ]; then
+    #Arguments are pathspecs for git status, used to exclude files for this test
+    if [ -n "$(git status -u --porcelain -- "$@")" ]; then
+        echo "Build produced untracked or modified files:----------------------------------------"
+        git status -u --porcelain -- "$@"
+        echo "-----------------------------------------------------------------------------------"
+        echo "Pathspec is: \"$*\", withouth pathspec:--------------------------------------------"
+        git status -u --porcelain
+        echo "-----------------------------------------------------------------------------------"
+        exit 1
+    else
+        echo Repo is clean
+        echo "Pathspec is: \"$*\", withouth pathspec:--------------------------------------------"
+        git status -u --porcelain
+        echo "-----------------------------------------------------------------------------------"
+        exit 0
+    fi
+else
+    #No arguments: Test all files
+    if [ -n "$(git status -u --porcelain)" ]; then
+        echo "Build produced untracked or modified files:----------------------------------------"
+        git status -u --porcelain
+        echo "-----------------------------------------------------------------------------------"
+        exit 1
+    else
+        echo Repo is clean
+        exit 0
+    fi
+fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,9 @@ on:
 permissions:
   contents: read #  to fetch code (actions/checkout)
 
+env:
+  DEBIAN_FRONTEND: noninteractive
+
 jobs:
 
   rip-and-test:
@@ -157,10 +160,7 @@ jobs:
         GITHUB_CONTEXT: ${{ toJson(github) }}
       run: echo "$GITHUB_CONTEXT"
     - name: Install pre-dependencies
-      env:
-        DEBIAN_FRONTEND: noninteractive
       run: |
-        set -e
         set -x
         apt-get update
         apt --yes upgrade
@@ -173,32 +173,11 @@ jobs:
         # "fetch-depth: 0" fetches all of history, this is needed by
         # our build system to determine the version from tags
         fetch-depth: 0
-
-    - name: Add linuxcnc.org deb archive
-      env:
-        DEBIAN_FRONTEND: noninteractive
-      run: |
-        case "${{matrix.image}}" in
-            debian:sid|debian:bookworm|debian:trixie)
-                exit 0
-                ;;
-            *)
-                ;;
-        esac
-        set -e
-        set -x
-        apt-get --yes install --no-install-recommends gpg software-properties-common
-        gpg --homedir="${PWD}/gnupg" --output /etc/apt/trusted.gpg.d/linuxcnc-deb-archive.gpg --export 3CB9FD148F374FEF
-        DIST=$(echo ${{matrix.image}} | cut -d : -f 2)
-        add-apt-repository "deb http://linuxcnc.org $DIST base"
-        apt-get update
     - name: Build architecture-specific Debian packages
       env:
         DEBEMAIL: emc-developers@lists.sourceforge.net
         DEBFULLNAME: LinuxCNC Github CI Robot
-        DEBIAN_FRONTEND: noninteractive
       run: |
-        set -e
         set -x
         git config --global --add safe.directory "${PWD}"
         debian/configure
@@ -208,10 +187,7 @@ jobs:
         apt-get --yes build-dep --arch-only .
         debuild -us -uc --build=any
     - name: Test debian packages
-      env:
-        DEBIAN_FRONTEND: noninteractive
       run: |
-        set -e
         set -x
         apt-get --yes install ../*.deb
         apt-get --yes install sudo # some tests run sudo...
@@ -222,7 +198,6 @@ jobs:
         su -c "./scripts/runtests -p ./tests" testrunner
     - name: Gather build artifacts
       run: |
-        set -e
         set -x
         ARCH=$(dpkg --print-architecture)
         DIST=$(echo ${{ matrix.image }} | cut -d : -f 2)
@@ -244,7 +219,6 @@ jobs:
         path: artifacts/${{ steps.meta.outputs.dist }}/${{ steps.meta.outputs.arch }}
         if-no-files-found: error
 
-
   package-indep:
     runs-on: ubuntu-24.04
     strategy:
@@ -261,10 +235,7 @@ jobs:
         GITHUB_CONTEXT: ${{ toJson(github) }}
       run: echo "$GITHUB_CONTEXT"
     - name: Install pre-dependencies
-      env:
-        DEBIAN_FRONTEND: noninteractive
       run: |
-        set -e
         set -x
         apt-get update
         apt --yes upgrade
@@ -277,32 +248,11 @@ jobs:
         # "fetch-depth: 0" fetches all of history, this is needed by
         # our build system to determine the version from tags
         fetch-depth: 0
-
-    - name: Add linuxcnc.org deb archive
-      env:
-        DEBIAN_FRONTEND: noninteractive
-      run: |
-        case "${{matrix.image}}" in
-            debian:sid|debian:bookworm|debian:trixie)
-                exit 0
-                ;;
-            *)
-                ;;
-        esac
-        set -e
-        set -x
-        apt-get --yes install gpg software-properties-common
-        gpg --homedir="${PWD}/gnupg" --output /etc/apt/trusted.gpg.d/linuxcnc-deb-archive.gpg --export 3CB9FD148F374FEF
-        DIST=$(echo ${{matrix.image}} | cut -d : -f 2)
-        add-apt-repository "deb http://linuxcnc.org $DIST base"
-        apt-get update
     - name: Build architecture-independent Debian packages
       env:
         DEBEMAIL: emc-developers@lists.sourceforge.net
         DEBFULLNAME: LinuxCNC Github CI Robot
-        DEBIAN_FRONTEND: noninteractive
       run: |
-        set -e
         set -x
         git config --global --add safe.directory "${PWD}"
         debian/configure
@@ -312,15 +262,11 @@ jobs:
         apt-get --yes build-dep --indep-only .
         debuild -us -uc --build=source,all
     - name: Test install debian packages
-      env:
-        DEBIAN_FRONTEND: noninteractive
       run: |
-        set -e
         set -x
         apt-get --yes install ../*.deb
     - name: Gather build artifacts
       run: |
-        set -e
         set -x
         DIST=$(echo ${{ matrix.image }} | cut -d : -f 2)
         ARCH=all
@@ -358,7 +304,7 @@ jobs:
         path: release_artifacts
     - name: Prepare upload assets
       run: |
-        set -e
+        set -x
         mkdir -p upload
         echo "Downloaded artifacts layout:" && find release_artifacts -maxdepth 3 -print | sed 's/^/  /'
         for d in release_artifacts/*; do
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,22 +24,30 @@ jobs:
     - name: Dump GitHub context
       env:
         GITHUB_CONTEXT: ${{ toJson(github) }}
-      run: echo "$GITHUB_CONTEXT"
-    - uses: actions/checkout@v2
+      run: |
+        echo Number of CPUs: $(nproc)
+        echo "$GITHUB_CONTEXT"
+    - name: Checkout repository
+      uses: actions/checkout@v6
       with:
-        submodules: true
-        fetch-depth: 0
-    - run: git fetch --recurse-submodules=no https://github.com/linuxcnc/linuxcnc +refs/tags/*:refs/tags/*
-    - name: Build RIP & test
+        submodules: false
+        fetch-depth: 1
+    - name: Install dependencies
       run: |
         set -x
         ./scripts/travis-install-build-deps.sh
+    - name: Build RIP GCC
+      run: |
+        set -x
         cd src
         ./autogen.sh
         ./configure --disable-check-runtime-deps --enable-werror
         make -O -j$((1+$(nproc))) default pycheck V=1
         # Note that the package build covers html docs
-        ../scripts/rip-environment runtests -p
+    - name: Test
+      run: |
+        set -x
+        ./scripts/rip-environment runtests -p
 
   rip-rtai:
     runs-on: ubuntu-24.04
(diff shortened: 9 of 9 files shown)


hdiethelm:
[Remove eatmydata](https://github.com/LinuxCNC/linuxcnc/pull/3984/commits/490b4aec3c5edba435b00d405b48674ce864b60e) tested:
[Before](https://github.com/LinuxCNC/linuxcnc/actions/runs/25149231313/usage): 4h 42m 44s
[After](https://github.com/LinuxCNC/linuxcnc/actions/runs/25150463132/usage?pr=3984): 4h 34m 41s + 15m estimated of a stalled process -> 4h 49m 41s
So no big impact and better readable.

BsAtHome:
Did you see this message at the "Complete job" stage:
```
Warning: Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v2. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```

hdiethelm:
> Did you see this message at the "Complete job" stage:
> 
> ```
> Warning: Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v2. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
> ```

Yes, commit is already there, I just wait for the last CI to pass before the next push.

hdiethelm:
@BsAtHome Feel free to cancel any pipeline that has failed jobs. I think I don't have the access rights to do so.

hdiethelm:
So first success, with 4 CPU's, down to 3h 32
[Before](https://github.com/LinuxCNC/linuxcnc/actions/runs/25155265119/usage): 4h 38m usage / 38m runtime
[After](https://github.com/LinuxCNC/linuxcnc/actions/runs/25157067613/usage): 3h 42m usage / 24m runtime

hdiethelm:
Now the bigger change will be to run all in prepared dockers or runners. It probably won't reduce the runtime a lot, estimated: 3-5 min. However, it is just nice for the debian to not hammer their package repo just for fun. But let's see how well that works.

@BsAtHome Debian is the main target, right? So running rip-* under a debian container is also fine? At the moment, they run with an ubuntu-24.04 runner.

BsAtHome:
Debian is the primary target, yes.

Nice to see improvement and it is primarily in the independent packages building the documentation. Those were the slowest all the time.

The tests are not going to be significantly faster because they run in sequence. We've been discussing parallel execution, but that requires #2722 to be fully implemented. There are a significant number of issues not addressed in that PR (some noted, others implied, still others to be discovered).

andypugh:
#3983 is in.

hdiethelm:
> #3983 is in.

Thanks, re-based on top of master.

> Debian is the primary target, yes.
> 
> Nice to see improvement and it is primarily in the independent packages building the documentation. Those were the slowest all the time.
> 
> The tests are not going to be significantly faster because they run in sequence. We've been discussing parallel execution, but that requires #2722 to be fully implemented. There are a significant number of issues not addressed in that PR (some noted, others implied, still others to be discovered).

It depends what the target is. For local usage, #2722 is the most comfortable way for users. But it can be also done another way:
- One Build
- Run multiple tests in parallel, each testing one part. Can also be separated using:
  - Multiple runners (This needs an artifact, the way linuxcnc is built right now, this can be cumbersome bit it might work)
  - Might be multiple namespaces in the same runner. If this works, this can be used also locally. Namespaces is one part how docker separates itself from the host but it can be also used directly. I would have to try it out.

Due to the test runner will only shorten the CI if the package-indep gets below 12min, I see this as low prio for the CI. 

@andypugh Objections if I push docker images to the linuxcnc github? They will appear somewhere here: https://github.com/LinuxCNC/linuxcnc/packages I can probably do that in CI without any additional rights. However, I might need you if I mess something up and packages have to be deleted. I will try it in my account first with the free credits but to do something meaningful, I have to do it in the linuxcnc CI.

BsAtHome:
Maybe the curl progress meter can be shut off at download.

@NTULINUX, when do these kernels move to linuxcnc.org? Is there a procedure?

grandixximo:
Opened #3992 as a small stopgap for the firefox snap flake (issue #3991). It adds one `apt remove firefox` line before each upgrade. If you would prefer the larger CI cleanup in #3984 to land first I can close mine; otherwise rebasing #3984 over my change should be a one-line conflict at most. Whichever lands first works for me.

hdiethelm:
@grandixximo This is fine for me. The main target here is to not have to do all this installs at all. But this will take some time.

I could extract the --cpu improvements (25% faster package build) and if desired:
- Actions update (A bit risky, I can not test the release target, but I don't expect that there is an issue)
- Split into dependency / build / test (I don't see any risks)
- Remove apt-get upgrade
- Remove eatmydata

in a separate PR, so this can be already merged while I am busy here.

hdiethelm:
So, I have a container based prototype running in my gitlab:
https://github.com/hdiethelm/linuxcnc-fork/tree/ci_docker_prototyping
https://github.com/hdiethelm/linuxcnc-fork/compare/ci_improvemens...hdiethelm:linuxcnc-fork:ci_docker_prototyping
The containers are visible here and in the project page, will be similar on the original repo:
https://github.com/hdiethelm?tab=packages&repo_name=linuxcnc-fork
Docker build:
https://github.com/hdiethelm/linuxcnc-fork/actions/runs/25216626662
LinuxCNC rip-and-test:
https://github.com/hdiethelm/linuxcnc-fork/actions/runs/25216626671

Do you think such an approach is viable? If yes, I will try how to automatically update the docker images from time to time and then port all other targets.

Advantage:
- Faster, rip-and-test uses 3m30 less time, will be even more for the debian package build
- Will not download any packages except if new dependency's are added until the docker images are updated
- Also rip-and-test targets could use a matrix to test arm / all Debian variants
- You can use the build docker images also locally

Disadvantage:
- Due to preinstalled package, missing dependency's might be discovered later
- Possibly a bit more difficult to maintain
- Will use storage for images but depending on your contract, this is for free

I will not yet merge this it this PR. As soon as I do this, packages will most probably appear in the original linuxcnc repo.

hdiethelm:
@grandixximo 
> might want to look at: ROS2, Yocto, buildroot for inspiration

Let's continue the discussion here. LinuxCNC is a bit special due to it need's a ton of depency's to build. The container needed is like 3.4GB in size.

- buildroot: No meaningful CI in github
- Yocto: self-hosted runner / Might be i am looking in the wrong project, but they do not much here: https://github.com/yoctoproject/vscode-bitbake/actions/runs/24083919003/job/70257898769
- ROS2: Also not much in Github CI

If you find something, just tell me. I had some inspiration from https://github.com/open-webui/open-webui due to I know they build containers from CI and many different manuals / articles.

BsAtHome:
The size of the container should be acceptable and manageable if it is in a local repo for quick download. If there need to be many different versions, that could be a problem.

However, I have no clue what deals have been made with github. @andypugh, do you know of any deals with github?

BTW, you need to rebase after I merged the snapped firefox killer.

hdiethelm:
> The size of the container should be acceptable and manageable if it is in a local repo for quick download. If there need to be many different versions, that could be a problem.
> 
> However, I have no clue what deals have been made with github. @andypugh, do you know of any deals with github?
> 
> BTW, you need to rebase after I merged the snapped firefox killer.

Looks like it is no issue due to the docker images are on github. I could not let it be to finalize my prototype. Now there is the same build like before but all using docker containers and I am still not out of free credits:

https://github.com/hdiethelm/linuxcnc-fork/actions/runs/25235809082
https://github.com/hdiethelm/linuxcnc-fork/actions/runs/25234109241

The time is now down by the estimated 3 min / job. The build-arch could be reduced by future 30s due to they install dependency packages (257MB / 268 packages). However, I don't have an idea how to pipe this list back to the docker build yet. And I am not sure if it is worth the effort.

Master CI: 4h 41m / 38m runtime
Without docker: 3h 42m usage / 24m runtime
With docker: 3h01 usage /  21m duration

Nearly a factor of 2 is already nice.

grandixximo:
> If you find something, just tell me.

I was thinking auto installation of missing dependency, whenever one is added, with warnings to request addition to the CI.

I thought one of those projects had it, guess I was wrong, I will look again.

grandixximo:
Found something. rust-lang/rust does this nicely: [src/ci/docker/run.sh](https://github.com/rust-lang/rust/blob/master/src/ci/docker/run.sh) hashes the Dockerfile plus every file it `COPY`s (plus arch and a manual cache-version int) and uses the hash as the image tag. PR that touches deps flips the tag, registry lookup misses, script rebuilds transparently with `--cache-from type=registry` warmed by the previous image. No auto-install fallback needed because a stale image cannot be picked up by construction, and the weekly rebuild becomes a cache warm-up rather than a correctness gate.

For linuxcnc the hash inputs would be `debian/control` plus the Dockerfile. LLVM pairs the same idea with a `paths:` filter on those files in [build-ci-container.yml](https://github.com/llvm/llvm-project/blob/main/.github/workflows/build-ci-container.yml) so the image rebuild and the dep change land atomically in the same PR.

grandixximo:
While you are restructuring the CI flow, would it be in scope to add a working-tree-clean gate after the build steps?

```yaml
- name: Verify no untracked or modified files after build
  run: |
    if [ -n "$(git status --porcelain)" ]; then
      echo "Build produced untracked or modified files:"
      git status --porcelain
      exit 1
    fi
```

Motivation: `docs/man/.gitignore` enumerates ~450 generated man pages by hand. The list goes stale silently when a new component lands without a matching ignore entry. Recent recurrences include `demux_generic.9` (fixed in 073f3a5022) and `output_buffer.9` (currently untracked on master after 3820604cd0).

`output_buffer.9` would actually make a convenient test case for the gate: drop the check into the workflow without fixing the gitignore first, and master should fail. Then a one-line addition to `docs/man/.gitignore` makes it green. Confirms the gate works end-to-end on a real existing oversight.

Happy to send it as a separate PR after this one merges if preferred.

hdiethelm:
> While you are restructuring the CI flow, would it be in scope to add a working-tree-clean gate after the build steps?

Makes sense. But as of now, the linuxcnc folder is quite a mess after build, especially RTAI. It will probably fail. But I can add it as a soft failing step, so more a reminder something should be cleaned up, like cppcheck / shellcheck. It can be changed to hard fail after the full cleanup is done. The same should be done one time for the above tasks.

hdiethelm:
> Found something. rust-lang/rust does this nicely: [src/ci/docker/run.sh](https://github.com/rust-lang/rust/blob/master/src/ci/docker/run.sh)...

Good idea but I have to see how I can use this in Gitlab CI. At the moment, there are three places where packages are installed in my docker variant with size: (download/disk MB)

1. Base dependency's, hash of dockerfile will do it (207/902MB): https://github.com/hdiethelm/linuxcnc-fork/blob/ci_docker_prototyping/.github/docker/Dockerfile#L14
2. Debian build dependency's, hash of Debian folder will do it (870/2241MB): https://github.com/hdiethelm/linuxcnc-fork/blob/ci_docker_prototyping/.github/docker/Dockerfile#L21
3. At test install, which file cover these dependency's? Are they automatically generated by the package creation? (257/1158MB): https://github.com/hdiethelm/linuxcnc-fork/blob/ci_docker_prototyping/.github/workflows/ci-docker.yml#L209

Additionally, I left all build-dep steps still in so if a new package is added later, it doesn't fail. However as of now, nothing is installed there due to everything is already in docker.

The 3. step is a bit annoying to pack in docker. I see some options:
- Find a way to get all packages for install without actually building the package.
- Build Debian package also before creating docker, so it can be installed with all dep's and then only the linuxcnc removed again.
- Get the Debian package of an other build and do ^^. But I haven't yet found how to get these from artifacts. I could push them as packages.
- Save a package list manually created

BsAtHome:
`xauth` needs to be installed, it seems, for some tests to complete. It does not seem to be listed as explicit dependency (or I'm not seeing it).

hdiethelm:
> `xauth` needs to be installed, it seems, for some tests to complete. It does not seem to be listed as explicit dependency (or I'm not seeing it).

Thanks, this got lost with --no-install-recommends. Fixed.
Both docker and non-docker variants are now quite similar. So they can be easily compared.

However, compared to the master CI, there are not a lot of lines I did not change. I think before setting this ready I will swash a few commits and write why I changed some parts or even removed them completely.

But first some more changes and then we have to decide if we go the prebuilt docker way or not.

hdiethelm:
> While you are restructuring the CI flow, would it be in scope to add a working-tree-clean gate after the build steps?

Done, was not so bad, only one man was missing, so I was able to enforce it. But I had to add an exclude option so I can exclude the auto updated VERSION / debian/changelog file. The regexp is not perfect, it would also exclude VERSION123 but better than nothing. Any good idea to do this?

The doc step still fails with changed .po files. Should I ignore these?

grandixximo:
For the regex false-positives (VERSION123 etc.), git pathspec excludes give exact-path matching with no regex anchoring needed:

```sh
git status -u --porcelain -- ':(exclude)VERSION' ':(exclude)debian/changelog'
```

That replaces the `grep -v -E` step entirely. If you want to keep the regex form for flexibility, anchoring against the porcelain `XY filename` format works: `^.{3}(VERSION|debian/changelog)$`.

For the `.po` files, those get touched by `msgmerge` during the doc build (line-number refs in source comments shift whenever any translatable string moves), so failing CI on them would basically mean any docs change blocks the gate. I'd lean toward excluding them too, but that's a doc/i18n call. @BsAtHome any preference, or should we ping the translation maintainers (Hans Unzner / Steffen Möller from recent .po history)?

hdiethelm:
> For the regex false-positives (VERSION123 etc.), git pathspec excludes give exact-path matching with no regex anchoring needed:

Thanks for the hint with pathspec. I tend towards this solution. Regexp is the thing that many people don't really understand and get it wrong (including me).
For the doc's, `':(exclude)docs/po/*.po' ':(exclude)docs/po/documentation.pot'` does the trick. For now, I just allow failure until we know what to do.

hdiethelm:
BTW: With the dockers from my repo, you can easily test parts of the CI locally with this command run from the local source directory and then copy-past snippets from ci.yml:
`podman run --rm -it -v .:/linuxcnc ghcr.io/hdiethelm/linuxcnc/build-container:trixie-amd64-latest`



I prefer podman over docker due to podman runs without setuid so the created files are not owned by root.

BsAtHome:
> For the `.po` files, those get touched by `msgmerge` during the doc build (line-number refs in source comments shift whenever any translatable string moves), so failing CI on them would basically mean any docs change blocks the gate. I'd lean toward excluding them too, but that's a doc/i18n call. @BsAtHome any preference, or should we ping the translation maintainers (Hans Unzner / Steffen Möller from recent .po history)?

The docs build, integration and translations are fragile. Failing on translations/.po can lead to some self-inflicted wounds. That said, the most common problem is the newline consistency, where translations are missing or adding trailing newlines. Most of that gets caught now because msgmerge fails and that is good. This is not normally merged into the main tree, but just the weblate branch failing CI. That fail is a Good Thing<sup>TM</sup>.

The default man-pages should all function and there should not be any non-.gitignored remains in docs/man/\*. That bug has bitten many times. I think, in general, there should be no non-.gitignored remains anywhere after the build and test. That also goes for some of the junk left-over in config/\* when, f.ex. running RIP qtdragon from a sim config.

grandixximo:
> I think, in general, there should be no non-.gitignored remains anywhere after the build and test.

If I understand correctly, the new CI enforces this already

> That also goes for some of the junk left-over in config/* when, f.ex. running RIP qtdragon from a sim config.

There are no tests for the UI's, since they are not run, no left over can be caught, we'd have to make tests that run the UIs, and then check for leftovers. @hdiethelm I could work on a separate PR for UI tests, would that be ok?

BsAtHome:
> > I think, in general, there should be no non-.gitignored remains anywhere after the build and test.
>
> If I understand correctly, the new CI enforces this already

Great!

> > That also goes for some of the junk left-over in config/* when, f.ex. running RIP qtdragon from a sim config.
>
> There are no tests for the UI's, since they are not run, no left over can be caught, we'd have to make tests that run the UIs, and then check for leftovers. @hdiethelm I could work on a separate PR for UI tests, would that be ok?

As i understand it, there has been talk about test-running some/all/any of the sims in xvfb and do some basic functional testing. That would be a very great test target because the UIs also break. Some sim UIs haven't seen a test in ages (one ini-file has a very old version tag).

See #3756.

grandixximo:
@hdiethelm planning to start on #3756 (xvfb-based UI smoke tests) once this lands or in parallel. Quick coordination question on the dep side before I start: my tests will need at minimum `xvfb`, `xauth` (already added by you, thanks), `xdotool`, and `python3-pytest-qt`. Possibly `mesa-utils` / `libgl1-mesa-dri` for GLcanon under llvmpipe.

I think this is a different situation from your rtapi cleanup V2 (#3919) where I rebased: that one shares files in `src/rtapi/*` and conflicts were unavoidable. UI tests live under `tests/ui/*` and only touch `.github/` for the dep install list, so the only real shared edit is 3-5 package names. No code conflict, no rebase requirement on either side.

Two ways to handle the deps:

1. You add the extras to your dep list in this PR while it's open. Keeps the install step in one place.
2. I add them in the UI test PR. Lets this one land cleanly without scope creep.

Either works for me. Lean which way?

I'll branch from master so neither PR blocks the other. Whoever lands second adjusts the dep list (one-line edit).

hdiethelm:
>...so the only real shared edit is 3-5 package names. No code conflict, no rebase requirement on either side.
> 
> Two ways to handle the deps:
> 
>     1. You add the extras to your dep list in this PR while it's open. Keeps the install step in one place.
>     2. I add them in the UI test PR. Lets this one land cleanly without scope creep.

Probably you need 2. so the CI still runs on your branch.

I have to rebase to master anyway before merge to check if all is still fine. Just do your thing, as long as there are only a few packages more, i will see them, no issue or if mine gets in first this should also be no issue for you to rebase before merge.

hdiethelm:
So, I finally managed to get all packages inside the docker container:
https://github.com/hdiethelm/linuxcnc-fork/actions/runs/25278237458
https://github.com/hdiethelm/linuxcnc-fork/actions/runs/25278693726
To do this, I just compile the deb's, install them and remove linuxcnc_* afterwards, so the dependency's stay:
https://github.com/hdiethelm/linuxcnc-fork/blob/ci_docker_prototyping/.github/docker/Dockerfile#L34

I would not call it elegant but it does the job. Docker build goes up from 7 to 12 min but due to this is only needed from time to time to be up to date, I don't see any issue.

Or has anyone an idea how to install these packages without building / installing linuxcnc?

The linuxcnc pipeline stays at ~ 21m duration / 3h usage. The advantage is that the debian archives are not loaded for nothing and that if they are down, the build still works.

hdiethelm:
> > I think, in general, there should be no non-.gitignored remains anywhere after the build and test.
> 
> If I understand correctly, the new CI enforces this already

Actually, I check after build. Should I move the check to after test? It's easy to do, I can even test after build and after test, takes no time at all.

@BsAtHome What to do with the untracked po files? At the moment, I don't ignore them but I made this build step non-failing, similar to cppcheck / shellcheck: 
https://github.com/LinuxCNC/linuxcnc/actions/runs/25276731551
https://github.com/LinuxCNC/linuxcnc/actions/runs/25276731551/job/74107710597#step:6:17

BsAtHome:
> @BsAtHome What to do with the untracked po files? At the moment, I don't ignore them but I made this build step non-failing, similar to cppcheck / shellcheck: https://github.com/LinuxCNC/linuxcnc/actions/runs/25276731551 https://github.com/LinuxCNC/linuxcnc/actions/runs/25276731551/job/74107710597#step:6:17

The po-files should just be ignored. There is no point in failing on them. The whole i18n and l10n stuff needs to be cleaned up anyway. But that is subordinate to fixing the docs build in general (docs build is quite fragile). And all of that is subordinate to reorganizing the docs structure.

Currently, cppcheck and shellcheck must be ignored because the build is still not clean. FWIW, the ability to add -Werror took many, many months before all the kinks were ironed out.
The cppcheck part requires the remote scheduler to be fixed (well, rewritten). Then that could become terminal on any warning or error.
The shellcheck is even harder because the code expects a lot of dubious constructs to work and needs a lot more work. It was a lot worse but there is still a lot of work to get it clean.

hdiethelm:
> The po-files should just be ignored.

Done: 
https://github.com/LinuxCNC/linuxcnc/pull/3984/changes/0cce4aea12581fd685f33d9e207133e6d7add38a
https://github.com/LinuxCNC/linuxcnc/actions/runs/25287962668/job/74135296641#step:6:9

> Currently, cppcheck and shellcheck must be ignored because the build is still not clean. FWIW, the ability to add -Werror took many, many months before all the kinks were ironed out. The cppcheck part requires the remote scheduler to be fixed (well, rewritten). Then that could become terminal on any warning or error. The shellcheck is even harder because the code expects a lot of dubious constructs to work and needs a lot more work. It was a lot worse but there is still a lot of work to get it clean.

I know that, introducing this kind of checks takes a lot of patience. Best in my opinion is to warn only for certain files but make the test failing for everything that is already good. So at least no new issues are introduced and then go from there step by step.

hdiethelm:
Any clue why this fails? Might be a broken package went into sid just right now? I did not do any relevant changes...
https://github.com/LinuxCNC/linuxcnc/actions/runs/25287962668

hdiethelm:
@BsAtHome Thinking about this checks, of course I found some issues I created.

One option would be to do something like this once and commit the file:
`scripts/shellcheck.sh | grep -Po '^In \K[^ ]+' | sort -u > shellcheck_warn_only.txt`
Similar for cppcheck.
- Then warn only for all files in this list / hard fail for everything else. So at least no new issues come in.
- Remove corrected files from these lists to avoid new issues coming in on already fixed files.

hdiethelm:
So how to continue? So far:
[Master CI before branch](https://github.com/LinuxCNC/linuxcnc/actions/runs/25213719310): 4h 41m / 38m
[This PR](https://github.com/LinuxCNC/linuxcnc/actions/runs/25289622442?pr=3984): 3h 25m / 22m
[With Docker](https://github.com/hdiethelm/linuxcnc-fork/actions/runs/25424329671): 3h 0m / 22m

Options:

1. Finalize this PR optionally docker variant later in a separate PR
2. Go to docker variant
3. Add docker CI to this PR but deactivate one of them: Just move the file to like ci.yml.deactivated. Which one to deactivate? 

I will wait for feedback and then ToDo from my side in any case:

- Review
- Test
  - Do a few on purpose failing commits
  - Compare resulting artifacts before and after this PR
  - Tag a release and test the release stage (I can do that in my github CI)
- Cleanup commit history and add description why some changes are done

BsAtHome:
I think we should start with this PR. It already improves quality and run time significantly. Cleanup this PR (and resolve merge conflicts), then do some more tests. I should be able to copy this branch and test some failing commits.

From there we should develop the docker variant and probably make the container build a separate repo under the LinuxCNC account.

hdiethelm:
Agreed. I also don't like to much change in one PR.

Docker build in a separate repo has the issue that at the moment, I need the linuxcnc source to determine the needed packages. Options:
- Fetch source from main repo
- Manually maintain a package list: Manually is always what breaks from experience so not desired
- Figure out a way to do that automatically without needing the source:
  - Publish packages and use these
  - Figure out a way to get artifacts from main repo
  - Publish needed package list from main repo and use this
  - ... Probably more

The main issue I see with the docker variant is, that if you remove a dependency by accident that is needed, it can stay undetected due to the packages are still in the docker.

With the actual docker version where I install missing packages in the main CI, newly added dependency's are no issue. And I can add a warning if this happens easily.

I suppose to finalize this PR and start a discussion about the docker variant  in an issue after this one is merged.

hdiethelm:
@BsAtHome So, I have redone the CI improvements in a series of logical commits with some reasoning why in the commit message. Do you think this is understandable this way?

Are the scripts travis-build-test.sh /  travis-install-build-deps.sh used anywhere else than in the CI? I placed now all CI scripts in .github/scripts. If the travis* scripts are unused now, I could delete them.

So this is ready for review.

After this CI has finished, I will follow with a few on purpose failing commits to see if the failures are properly handled.

hdiethelm:
Fail build&doc: https://github.com/LinuxCNC/linuxcnc/actions/runs/25636338081
Fail test: https://github.com/LinuxCNC/linuxcnc/actions/runs/25636480466
Fail untracked: https://github.com/LinuxCNC/linuxcnc/actions/runs/25636582818
Seams to do the job.

BsAtHome:
> @BsAtHome So, I have redone the CI improvements in a series of logical commits with some reasoning why in the commit message. Do you think this is understandable this way?

Yes, but it could have been one commit. This is a big change one way or the other. That said, the dev-chain for the changes is nice to follow, so its fine.

> Are the scripts travis-build-test.sh / travis-install-build-deps.sh used anywhere else than in the CI? I placed now all CI scripts in .github/scripts. If the travis* scripts are unused now, I could delete them.

Well, I know I use them once in a while...

FWIW, CI is well isolated with all the required scripts under .github/scripts.

> So this is ready for review.
> After this CI has finished, I will follow with a few on purpose failing commits to see if the failures are properly handled.

When done, remove the draft tag.

hdiethelm:
So, the fail checks failed as they should, so it is ready from my side.

I downloaded the trixie indep and amd64 artifacts, extracted them and compared them to master. Most is binary identical and there are no new/missing files in the deb's except:
- Timestamps 
- All generated images, however they look equal
- All generated doc's which have a version number inside
- A few binary's, might be there is also a version number or timestamp inside

There is only one thing I don't yet fully understand. Sometimes, the package-indep sid job failed and I don't know why, for example here: https://github.com/LinuxCNC/linuxcnc/actions/runs/25428177491/job/74587110190

The log is really long and I did not find any relevant error message and I was also not able to reproduce it on my PC.

Today it did not happen until now but there might be a hidden issue. Or it might be also github being github... ;-)

hdiethelm:
> > Are the scripts travis-build-test.sh / travis-install-build-deps.sh used anywhere else than in the CI? I placed now all CI scripts in .github/scripts. If the travis* scripts are unused now, I could delete them.
> 
> Well, I know I use them once in a while...

Now that the CI scripts are there and there are dockers in my fork, I started using the dockers and the CI scripts for testing. For example:

``` sh
podman run --rm -it -v .:/linuxcnc ghcr.io/hdiethelm/linuxcnc/build-container:trixie-amd64-latest
cd linuxcnc/
.github/scripts/build-rip.sh --with-realtime=uspace
```

This is really nice to debug issues with a Debian variant you don't have a VM for. With some tricks you can even start the UI:

``` sh
xhost +"local:podman@linuxcnc"
podman run --net=host -e DISPLAY=$DISPLAY --rm -it -v .:/linuxcnc ghcr.io/hdiethelm/linuxcnc/build-container:trixie-amd64-latest
cd linuxcnc/
chown -R linuxcnc:linuxcnc .
su linuxcnc
.github/scripts/build-rip.sh --with-realtime=uspace
scripts/linuxcnc
```
Note: chown to linuxcnc user messes up your access rights under the host user for this directory. You need to fix them with sudo chown afterwards.

BsAtHome:
One problem is that the indep CI is using "make -j4" and is missing the -O option to keep the messages from each instance together. Now they may be all gobbled up.

It always seems to happen in "make docs". This has always been a bit fragile. Not sure where it comes from, but one of the usual suspects is a bad translation that trips something somewhere. When you look through the messages, then you can discover all sorts of problems that are not handled nor fixed. Fixing the docs build is another huge undertaking that is due.

BsAtHome:
BTW, one thing we need to get rid of is using (Xe)TeX for docs.

hdiethelm:
Thanks for merging!

For the docker variant: Just tell me if and when you like to do this.

The code is there and it is easy to merge changes over in case the CI files get changed on master, so no urgency from my side. The whole main CI structure is pretty much identical for the docker variant, it just uses containers.

Time wise, the docker variant has no big advantages. It just loads the gitlab servers instead of the debian mirrors, so it is nicer on the debian project.

If you decide against the docker variant, fine for me. It was a good learning anyway... ;-)

BsAtHome:
You are welcome, but it is all others who have to say thanks. You did a lot of work and now it works a lot faster.

The docker variant needs to be integrated into the project in a different way. We need a separate repo to auto-build images when specific updates are done (not sure which trigger). It may not be time-wise be better, but it does make us more independent of provided images and we can tune/test the way we'd like. The docker variant may also solve some of the (github) instabilities we see when the default images are run.

But for now, I think we should enjoy the current changes for a moment before going deeper into the CI pit ;-)

grandixximo:
Many many thanks for the time spent on the CI, the effort does not go unnoticed, keep it up 💪

hdiethelm:
You are welcome. One of this side projects that took longer than estimated... ;-) I know people that complain about CI's taking 5-10 min and this one took 40 min so I took a look...

Time wise, the doc part is still the blocking path but this is not CI related. Way's to improve:

- Buy runners with more cores at gitlab
- Improve the doc build
- Only build doc when needed like:
  - Changes detected (Hard to do with the actual way of collecting doc's from all over the place)
  - After merge to master (Will detect doc issues late)
  - Special tag
  - ...

But let it be for a while and see how the actual version behaves before doing more change. I also want to implement xenomai net and create some chips on the CNC.

BsAtHome:
> Improve the doc build

This is the way to proceed :-)

All other options are just dealing with symptoms.

PR #3987, opened 2026-04-30
Halshow: Add function to add items from HAL text

hansu:
This adds a functionality to add pins/signals/parameters from a text which is copied from a HAL file.

Currently it fills the text box with the clipboard. This is handy but could also be irritating. What do you guys think?

grafik

grafik

grafik

### DIFF
--- a/tcl/bin/halshow.tcl
+++ b/tcl/bin/halshow.tcl
@@ -704,6 +704,7 @@ proc makeWatch {} {
     bind $::cisp <Configure> {
         if {$::watchlist_len <= 20} reloadWatch
     }
+    bind $::cisp <Button-3> {popupmenu_watch_general %W %X %Y}
 }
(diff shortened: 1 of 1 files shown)


zz912:
What is it good for?

hansu:
> What is it good for?

If you want to debug a specific part of your HAL (in this example the tool change related pins), then you can copy this part of your HAL file and you have all relevant pins and signals with one click instead of adding all them manually.

zz912:
I was thinking about your PR. I don't know if it would be too complicated, but I would rather like an editor for loaded HAL files. I need to edit things in the HAL files I'm using. Not just add something. What do you think about it?

hansu:
You mean something like a live HAL editor? This would be a comprehensive task and has nothing to do with this PR.

PR #3988, opened 2026-04-30
ini: Fix Tcl interaction, fix missing updates and some improvements

### LINKED ISSUE #3863: Master does not correctly recognize boolean ini values

### LINKED ISSUE #3972: linuxcnc config folder is polluted with a 'ini expanded' file when a config is run
using master RIP
When I run a sim config, there is a file auto generated and saved as the same name but with 'expanded' added to the end.
Why does this need to be there?
There is already too many files in the typical config folder. can at least it be deleted when the config closes?
I certainly could see a user being confused about there being two similarly named  INI files.

BsAtHome:
This PR addresses the problem with the ini-parser and Tcl interaction. The Tcl side now maps onto the new ini-parser and gets the data from the same parser and data pool as all other parts of linuxCNC. There is no longer a necessity to predigest the ini-file. This PR fixes #3972.

There is a need to get rid of some Tcl code that uses the parse_ini method to add sections/variables to the global Tcl scope as associative arrays. But that can wait for another time because the fallout would be larger than a simple update. It works now by reading the data from the new parser instead of trying to read the ini-file manually.

LinuxCNC pre-flight config check got an update. A new ini-file checker was written in python, linuxcnc_check_ini, to replace the old Tcl version (check_config.tcl). Checks have been greatly expanded, it performs better and uses the new ini-parser for all reading.
The new pre-flight check tests all ini-file variables that are known to be boolean to be valid if they are present in the ini-file. It also tests known enumerations and some integer variables with known numerical limits. This PR fixes #3863. 

The python interface now exposes conversion routines for some commonly used enumerations (linear/angular units and joint type). This way no reimplementations are necessary and they will be handled consistently.

A previously missed python based user interface (gscreen.py) has been updated to use the proper typed conversions.

The documentation has been updated.


c-morley:
How would an INI line like this be interpreted? 
Note it uses semicolons, not to designate comments.
 
MDI_COMMAND_MACRO0 = G0 Z25;X0 Y0;Z0, Goto\nUser\nZero

BsAtHome:
> How would an INI line like this be interpreted? Note it uses semicolons, not to designate comments.
> MDI_COMMAND_MACRO0 = G0 Z25;X0 Y0;Z0, Goto\nUser\nZero

Hm, good one, I completely missed that...

Updated so that comments are again not allowed in values. I tested it using the qtdragon sim and it looks to be correct again.

c-morley:
It was a big job you took on. Very Nice.

IFAIK the character # is not used anywhere except for comments. Could we just use it to allow inline comments?

BsAtHome:
> It was a big job you took on. Very Nice.

Yes, it is a big job and knew there would be some fallout and setbacks. They should be fixed now.


> IFAIK the character # is not used anywhere except for comments. Could we just use it to allow inline comments?

That could be possible (that was what I tested and probably forgot about the semicolon).

But, I think that the current ini-file format is abysmal and insane. A major problem is that strings are implied instead of explicit. That makes the whole format vulnerable to abuse and inconsistencies because there is no clear and clean delimiters.

I originally made those strings explicit, just to discover that quotes are abused to be part of the content. That is just insane. And worse, the expansions are directly fed into shell-expansions and execution (f.ex. hal files). We are talking about enabling rootless operation and have this gaping hole of running configuration files as scripts.

Configuration should be that, configuration, not code to be executed. There is a fine line between the two domains, code and configuration. The current setup steps on and crosses over that line. I do not say that I have all the solutions, but I see the challenge in making any change. Just, the current situation is not good and any change for the better is, in effect, a breaking change.

c-morley:
One win at a time, try to make the project a little better each time. Hopefully you find the challenge interesting.
Thanks for the fix there.

PR #3992, opened 2026-05-01
ci: remove firefox before apt upgrade to avoid snap store flake

### LINKED ISSUE #3991: CI: apt upgrade triggers firefox snap install, blocks PRs intermittently
## Summary

The `sudo eatmydata apt --yes --quiet -o Acquire::Retries=5 upgrade` step in `.github/workflows/ci.yml` (lines 35, 61, 118, 225, ...) triggers a firefox snap install/upgrade on `ubuntu-24.04` runners. When the snap store is slow or unreachable from the runner's network, the install retries for ~50 minutes and then fails with exit code 100, killing every job that runs that step before any LinuxCNC compile happens.

This is at least the second time firefox snap has held PRs back. Recently hit on PR #3990 three times in a row, all jobs running the apt upgrade step failed identically:

```
ERROR: ld.so: object 'libeatmydata.so' from LD_PRELOAD cannot be preloaded ...
==> Checking connectivity with the snap store
===> Unable to contact the store, trying every minute for the next 30 minutes
===> Still unable to contact the store, trying for another 20 minutes
===> Still unable to contact the store, trying for another 10 minutes
===> Still unable to contact the store, aborting
dpkg: error processing archive .../firefox_1%3a1snap1-0ubuntu5_amd64.deb (--unpack):
E: Sub-process /usr/bin/dpkg returned an error code (1)
##[error]Process completed with exit code 100.
```

`package-arch` and `package-indep` (Debian-based) jobs were unaffected because they do not run on ubuntu-24.04 with `apt upgrade`.

## Affected jobs

- `rip-and-test`
- `rip-and-test-clang`
- `rip-rtai`
- `htmldocs`

## Root cause

GitHub Actions `ubuntu-24.04` runner images ship firefox as a snap transition package. `apt upgrade` triggers snap store contact to refresh firefox, which is unreliable from runner VNet.

## Proposed fix

Two options:

1. **Drop the `apt upgrade` lines.** Build deps install via `scripts/travis-install-build-deps.sh` already provides what is needed; the upgrade adds little value on ephemeral runners.

2. **Sledgehammer**: remove firefox before the upgrade.
   ```yaml
   sudo apt-get remove -y firefox firefox-locale-en firefox-locale-* || true
   sudo eatmydata apt --yes --quiet -o Acquire::Retries=5 upgrade
   ```

Option 2 is robust and minimally invasive (preserves the upgrade for any genuinely needed package updates). Option 1 is cleaner but risks missing security patches that may matter for a niche test job.

Either fix is ~5 to 10 lines across `.github/workflows/ci.yml`. Happy to send a PR if a maintainer confirms the preferred approach.

grandixximo:
Fixes #3991.

The firefox snap transition package contacts the snap store from the runner VNet during apt upgrade. When the store is slow or unreachable the upgrade step retries for ~50 minutes then fails, killing the whole job before any LinuxCNC build runs.

Removing firefox before the upgrade sidesteps the snap path. Applied to the four ubuntu-24.04 jobs that run apt upgrade. Container-based package-arch / package-indep jobs are not affected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,7 @@ jobs:
         set -x
         sudo apt-get --yes -o Acquire::Retries=5 install eatmydata
         eatmydata ./scripts/travis-install-build-deps.sh
+        sudo apt-get remove -y firefox firefox-locale-* 2>/dev/null || true
         sudo eatmydata apt --yes --quiet -o Acquire::Retries=5 upgrade
         cd src
         eatmydata ./autogen.sh
@@ -71,6 +72,7 @@ jobs:
         sudo dpkg -i linux-headers-5.4.302-rtai-amd64_5.4.302-rtai-amd64-1_amd64.deb linux-image-5.4.302-rtai-amd64_5.4.302-rtai-amd64-1_amd64.deb rtai-modules-5.4.302_5.3.4-linuxcnc_amd64.deb
         #-----
         eatmydata ./scripts/travis-install-build-deps.sh
+        sudo apt-get remove -y firefox firefox-locale-* 2>/dev/null || true
         sudo eatmydata apt --yes --quiet -o Acquire::Retries=5 upgrade
         cd src
         eatmydata ./autogen.sh
@@ -96,6 +98,7 @@ jobs:
         sudo apt-get --yes -o Acquire::Retries=5 install eatmydata
         eatmydata ./scripts/travis-install-build-deps.sh
         sudo eatmydata apt-get -y -o Acquire::Retries=5 install clang
+        sudo apt-get remove -y firefox firefox-locale-* 2>/dev/null || true
         sudo eatmydata apt --yes --quiet -o Acquire::Retries=5 upgrade
         cd src
         eatmydata ./autogen.sh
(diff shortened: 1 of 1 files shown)


hdiethelm:
Is there any reason to do apt upgrade at all? I somehow expect that the github VM should be more or less up to date?

BsAtHome:
> Is there any reason to do apt upgrade at all? I somehow expect that the github VM should be more or less up to date?

Apparently not, considering the amount of packages pulled in at the upgrade.

grandixximo:
Wasn't sure if dropping the upgrade entirely was safe. VM age and edge cases I'm not familiar with could matter, or it might just be wasted bandwidth, but my confidence wasn't high enough to commit to that. The sledgehammer was a quick fix to take Firefox out of the snap-store path. Feel free to roll it back in #3984 and drop the upgrade altogether. I'm fine with not updating on Ubuntu stable.

hdiethelm:
My target in https://github.com/LinuxCNC/linuxcnc/pull/3984 is to go a completely different route:

1. Create docker images with all packages installed. Update them like once a week or on request. (TBD how)
2. Use these during CI runs, so no packages have to be installed at all expect if a new dependency was added. In this case, this is needed until the docker images got updated.

I don't now if or how well this works but no other way to discover this than to try it out. This is standard practice in gitlab, but I am new to github actions.

grandixximo:
might want to look at: ROS2, Yocto, buildroot for inspiration

Issue #3948, opened 2026-04-18
Error messages are not shown if motmod is loaded and not started


hdiethelm:
In 2.9, the message is:
```
hm2: loading Mesa HostMot2 driver version 0.15
Unknown parameter `num_pwmgens=1'
./hm2-stepper-eth.hal:38: waitpid failed /home/hannes/linuxcnc-src/bin/rtapi_app hm2_eth
./hm2-stepper-eth.hal:38: /home/hannes/linuxcnc-src/bin/rtapi_app exited without becoming ready
./hm2-stepper-eth.hal:38: insmod for hm2_eth failed, returned -1
```
In master, the message is:
```
hm2: loading Mesa HostMot2 driver version 0.15
./hm2-stepper-eth.hal:38: waitpid failed /home/hannes/linuxcnc-src/bin/rtapi_app hm2_eth
./hm2-stepper-eth.hal:38: /home/hannes/linuxcnc-src/bin/rtapi_app exited without becoming ready
./hm2-stepper-eth.hal:38: insmod for hm2_eth failed, returned -1
```

Note the missing `Unknown parameter 'num_pwmgens=1'`. This message is not printed any more which probably would have saved me 90% of the time debugging the issue https://github.com/LinuxCNC/linuxcnc/issues/3946.

The issue is introduced with this commit: https://github.com/LinuxCNC/linuxcnc/commit/47324a78687b6edf667129e5b261883ba6e8927e

In motion.c after init, all messages are routed through emc_message_handler. https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/motion/motion.c#L415

So in the case loading future modules fails and the error is for example printed in https://github.com/LinuxCNC/linuxcnc/blob/master/src/rtapi/uspace_rtapi_app.cc#L232, this is never shown due to the motion thread never starts.

Options:

- Only switch the message handler to emc_message_handler when motion is started
- Revert the commit and accept that there are duplicated messages
- ...probably more

BsAtHome:
This is probably the same bug that caused the long search in #3822 its workaround fix in #3824.

I'd rather see duplicate messages instead of non at all... But what is the actual root-cause of duplicate messages?

BTW, is it possible to switch the emc_message_handler <i>in </i> the startup procedure of the motion thread, as last action when all went fine?

hdiethelm:
[...]

PR #3993, opened 2026-05-02
motion: make emcmotError fifo lockfree MPSC (fixes #3951)

### LINKED ISSUE #3951: motion.c emc_message_handler() probably not thread save
At least in uspace mode, when motmod is loaded, the uspace_rtapi_app.cc:default_rtapi_msg_handler() is replaced by emc_message_handler().

emc_message_handler() is from then on called directly by all rtapi_print() and rtapi_print_msg() calls from all real time modules / threads.

Now looking in the code in emc_message_handler(), I don't see anything that guarantees thread safety between concurrent calls to emcmotErrorPutfv() and calls to emcmotErrorGet().

num++ / num-- looks like a possible race condition.

Mostly a theoretical issue, due to errors are not logged that often, but it can go wrong in theory due to concurrent calls to num++ / num-- can lead to num incremented only by one instead of two or not at all.

grandixximo:
Fixes #3951.

cc @BsAtHome @hdiethelm (from the discussion in #3948 / #3951).

`emc_message_handler()`, installed by `motmod` at the end of `rtapi_app_main`, can be invoked from any RT thread. `emcmotErrorPutfv()` and `emcmotErrorGet()` share plain `int` counters with no synchronisation: concurrent RT producers race on `num` / `start` / `end`, and the RT producer also races `num++` against the userspace consumer's `num--` across processes.

`rtapi_mutex_get` is unsuitable here: it spins with `sched_yield`, which is priority inversion when an RT producer waits on the non-RT consumer (the header explicitly says it should not be used in realtime code). `rtapi_mutex_try` would convert the race into a silent message drop on contention.

This patch replaces the counters with a lockfree multi-producer single-consumer ring using GCC `__atomic_*` builtins on plain `unsigned long long` fields, so the layout is identical in the C producer (`motmod`) and the C++ consumer (`milltask`) sharing the shmem region.

Producers CAS-claim a sequence number on `write_reserve`, write the payload to slot `S % EMCMOT_ERROR_NUM`, then spin until `write_commit == S` and release-store `S + 1`. The in-order publish step ensures the consumer never sees a committed slot whose payload from a lower seq is still being written. The consumer compares `read_seq` to acquire-loaded `write_commit`, copies the slot, and release-stores `read_seq + 1`. Buffer-full returns `-1`, same bound as before.

Properties: RT-safe, no `sched_yield`, no priority inversion, MPSC-correct, no silent drops outside the buffer-full case, no API change.

Files touched:

- `src/emc/motion/motion.h`: struct fields replaced (`head` / `tail` / `start` / `end` / `num` removed; `write_reserve` / `write_commit` / `read_seq` added).
- `src/emc/motion/emcmotutil.c`: `emcmotErrorInit` / `emcmotErrorPutfv` / `emcmotErrorGet` rewritten.

This does not change when `emc_message_handler` is installed (#3948); that symptom is orthogonal and intentionally left for the discussion in #3948 to converge. The `pushmsg` component proposed by @BsAtHome in #3948 would be a useful test bed once available.

Tested: clean RIP build, no warnings.

### DIFF
--- a/src/emc/motion/emcmotutil.c
+++ b/src/emc/motion/emcmotutil.c
@@ -7,7 +7,7 @@
 * Author:
 * License: GPL Version 2
 * System: Linux
-*    
+*
 * Copyright (c) 2004 All rights reserved.
 ********************************************************************/
 
@@ -16,17 +16,33 @@
 #include "dbuf.h"
 #include "stashf.h"
 
+/*
+ * Lockfree multi-producer single-consumer (MPSC) ring buffer.
+ *
+ * Producers (real-time threads invoked via the rtapi message handler) may
+ * race; the consumer (the userspace task) is single-threaded.
+ *
+ * Producer:
+ *   1. CAS-loop on write_reserve to claim seq S; bail if buffer full.
+ *   2. Write payload into slot S % EMCMOT_ERROR_NUM.
+ *   3. fetch_add(write_commit, 1) with release ordering.
+ * Consumer:
+ *   1. If write_reserve != write_commit, some producer is mid-write; skip
+ *      (return -1) so the RT producer is never blocked. Next consumer call
+ *      will retry.
+ *   2. Otherwise drain one slot at read_seq, increment read_seq.
+ */
+
 int emcmotErrorInit(emcmot_error_t * errlog)
 {
     if (errlog == 0) {
 	return -1;
     }
 
-    errlog->head = 0;
-    errlog->start = 0;
-    errlog->end = 0;
-    errlog->num = 0;
-    errlog->tail = 0;
+    errlog->write_reserve = 0;
+    errlog->write_commit  = 0;
+    errlog->read_seq      = 0;
+    atomic_thread_fence(memory_order_release);
 
     return 0;
 }
@@ -35,22 +51,38 @@ int emcmotErrorPutfv(emcmot_error_t * errlog, const char *fmt, va_list ap)
 {
     struct dbuf errbuf;
     struct dbuf_iter it;
+    unsigned long long my_seq;
 
-    if (errlog == 0 || errlog->num == EMCMOT_ERROR_NUM) {
-	/* full */
+    if (errlog == 0) {
 	return -1;
     }
 
-    errlog->head++;
+    {
+	unsigned long long w = atomic_load_explicit(&errlog->write_reserve,
+	                                            memory_order_relaxed);
+	for (;;) {
+	    unsigned long long r = atomic_load_explicit(&errlog->read_seq,
+	                                                memory_order_acquire);
+	    if (w - r >= (unsigned long long)EMCMOT_ERROR_NUM) {
+		return -1;
+	    }
+	    if (atomic_compare_exchange_weak_explicit(
+	            &errlog->write_reserve, &w, w + 1,
+	            memory_order_acquire, memory_order_relaxed)) {
+		my_seq = w;
+		break;
+	    }
+	}
+    }
 
-    dbuf_init(&errbuf, (unsigned char*)errlog->error[errlog->end], EMCMOT_ERROR_LEN);
+    dbuf_init(&errbuf,
+              (unsigned char*)errlog->error[my_seq % EMCMOT_ERROR_NUM],
+              EMCMOT_ERROR_LEN);
     dbuf_iter_init(&it, &errbuf);
     vstashf(&it, fmt, ap);
 
-    errlog->end = (errlog->end + 1) % EMCMOT_ERROR_NUM;
-    errlog->num++;
-
-    errlog->tail = errlog->head;
+    atomic_fetch_add_explicit(&errlog->write_commit, 1ULL,
+                              memory_order_release);
 
     return 0;
 }
@@ -72,16 +104,29 @@ int emcmotErrorPut(emcmot_error_t *errlog, const char *error)
 
 int emcmotErrorGet(emcmot_error_t * errlog, char *error)
 {
-    if (errlog == 0 || errlog->num == 0) {
-	/* empty */
+    unsigned long long r;
+    unsigned long long w_res;
+    unsigned long long w_com;
+
+    if (errlog == 0) {
 	return -1;
     }
 
-    errlog->head++;
-    memcpy(error, errlog->error[errlog->start], EMCMOT_ERROR_LEN);
-    errlog->start = (errlog->start + 1) % EMCMOT_ERROR_NUM;
-    errlog->num--;
-    errlog->tail = errlog->head;
+    w_com = atomic_load_explicit(&errlog->write_commit, memory_order_acquire);
+    w_res = atomic_load_explicit(&errlog->write_reserve,
+                                 memory_order_acquire);
+    if (w_res != w_com) {
+	/* a producer is mid-write; try again next cycle */
+	return -1;
+    }
+
+    r = atomic_load_explicit(&errlog->read_seq, memory_order_relaxed);
+    if (r >= w_com) {
+	return -1;
+    }
+
+    memcpy(error, errlog->error[r % EMCMOT_ERROR_NUM], EMCMOT_ERROR_LEN);
+    atomic_store_explicit(&errlog->read_seq, r + 1, memory_order_release);
 
     return 0;
 }
--- a/src/emc/motion/motion.h
+++ b/src/emc/motion/motion.h
@@ -82,6 +82,9 @@ to another.
 #define MOTION_INVALID_ID INT_MIN
 #define MOTION_ID_VALID(x) ((x) != MOTION_INVALID_ID)
 
+#include <rtapi.h>		/* must precede rtapi_atomic.h in kernel mode */
+#include <rtapi_atomic.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -739,14 +742,12 @@ Suggestion: Split this in to an Error and a Status flag register..
         int inhibit_probe_home_error;
     } emcmot_config_t;
 
-/* error structure - A ring buffer used to pass formatted printf strings to usr space */
+/* error structure - lockfree MPSC ring buffer. See emcmotutil.c. */
     typedef struct emcmot_error_t {
-	unsigned char head;	/* flag count for mutex detect */
 	char error[EMCMOT_ERROR_NUM][EMCMOT_ERROR_LEN];
-	int start;		/* index of oldest error */
-	int end;		/* index of newest error */
-	int num;		/* number of items */
-	unsigned char tail;	/* flag count for mutex detect */
+	rtapi_atomic_ullong write_reserve;
+	rtapi_atomic_ullong write_commit;
+	rtapi_atomic_ullong read_seq;
     } emcmot_error_t;
 
 
--- a/tests/lowlevel/emcmot-error-mpsc/test.c
+++ b/tests/lowlevel/emcmot-error-mpsc/test.c
@@ -1,0 +1,112 @@
+/*
+ * Concurrency / correctness test for the emcmotError MPSC ring buffer.
+ *
+ * Spawns NPROD producer threads pumping NMSG messages each, plus one
+ * consumer that drains. Verifies every message arrives exactly once
+ * and that messages from each individual producer arrive in producer-
+ * local sequence order. Across producers any interleaving is allowed.
+ */
+
+#include <pthread.h>
+#include <sched.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+
+#include <emcmotcfg.h>
+#include "motion.h"
+#include "dbuf.h"
+#include "stashf.h"
+#include <rtapi_atomic.h>
+
+#define NPROD 8
+#define NMSG  100000
+
+static emcmot_error_t errlog;
+static rtapi_atomic_int producers_done = 0;
+
+static void *producer_fn(void *arg) {
+    long id = (long)arg;
+    char buf[64];
+    for (int i = 0; i < NMSG; i++) {
+        snprintf(buf, sizeof(buf), "P%ld:%010d", id, i);
+        while (emcmotErrorPut(&errlog, buf) < 0) {
+            sched_yield();
+        }
+    }
+    atomic_fetch_add(&producers_done, 1);
+    return NULL;
+}
+
+static void *consumer_fn(void *arg) {
+    (void)arg;
+    long total = (long)NPROD * NMSG;
+    long received = 0;
+    int per_producer[NPROD] = {0};
+    char raw[EMCMOT_ERROR_LEN];
+    char msg[EMCMOT_ERROR_LEN];
+    int errors = 0;
+
+    while (received < total) {
+        struct dbuf d;
+        struct dbuf_iter di;
+        /* dbuf_init zero-fills the buffer, so call it BEFORE
+           emcmotErrorGet, which then writes the received bytes in. */
+        dbuf_init(&d, (unsigned char *)raw, EMCMOT_ERROR_LEN);
+        if (emcmotErrorGet(&errlog, raw) == 0) {
+            dbuf_iter_init(&di, &d);
+            if (snprintdbuf(msg, sizeof(msg), &di) < 0) {
+                fprintf(stderr, "snprintdbuf failed\n");
+                errors++;
+                received++;
+                continue;
+            }
+            long pid;
+            int seq;
+            if (sscanf(msg, "P%ld:%d", &pid, &seq) != 2
+                || pid < 0 || pid >= NPROD) {
+                fprintf(stderr, "bad msg: '%s'\n", msg);
+                errors++;
+            } else if (seq != per_producer[pid]) {
+                fprintf(stderr, "out-of-order P%ld: got %d, want %d\n",
+                        pid, seq, per_producer[pid]);
+                errors++;
+            } else {
+                per_producer[pid]++;
+            }
+            received++;
+        } else {
+            if (atomic_load(&producers_done) >= NPROD) {
+                struct timespec ts = {0, 1000000};
+                nanosleep(&ts, NULL);
+            } else {
+                sched_yield();
+            }
+        }
+    }
+
+    if (errors == 0) {
+        printf("test passed\n");
+    } else {
+        printf("test FAILED with %d errors\n", errors);
+    }
+    return errors == 0 ? (void *)0 : (void *)1;
+}
+
+int main(void) {
+    emcmotErrorInit(&errlog);
+
+    pthread_t prods[NPROD];
+    pthread_t cons;
+    pthread_create(&cons, NULL, consumer_fn, NULL);
+    for (long i = 0; i < NPROD; i++) {
+        pthread_create(&prods[i], NULL, producer_fn, (void *)i);
+    }
+    for (long i = 0; i < NPROD; i++) {
+        pthread_join(prods[i], NULL);
+    }
+    void *cret;
+    pthread_join(cons, &cret);
+    return cret ? 1 : 0;
+}
--- a/tests/lowlevel/emcmot-error-mpsc/test.sh
+++ b/tests/lowlevel/emcmot-error-mpsc/test.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+# Concurrency / correctness test for the emcmotError MPSC ring buffer.
+# RIP-only: builds the test driver against the in-tree motion sources.
+
+SCRIPTDIR="$(cd "$(dirname "$0")" && pwd)"
+TOPDIR="$(cd "$SCRIPTDIR/../../.." && pwd)"
+SRCDIR="${TOPDIR}/src/emc/motion"
+INCDIR="${TOPDIR}/include"
+
+if [ ! -f "${SRCDIR}/emcmotutil.c" ]; then
+    echo "skip: motion sources not found (RIP build required)"
+    exit 0
+fi
+
+gcc -O -pthread -DULAPI -Wall \
+    -I"${INCDIR}" \
+    -I"${SRCDIR}" \
+    "${SCRIPTDIR}/test.c" \
+    "${SRCDIR}/emcmotutil.c" \
+    "${SRCDIR}/dbuf.c" \
+    "${SRCDIR}/stashf.c" \
+    -o "${SCRIPTDIR}/test" || { echo "compile failed"; exit 1; }
+
+"${SCRIPTDIR}/test"
+exitval=$?
+rm -f "${SCRIPTDIR}/test"
+exit $exitval
--- a/src/rtapi/rtapi_atomic.h
+++ b/src/rtapi/rtapi_atomic.h
@@ -46,4 +46,36 @@
 
 #endif // defined(__cplusplus)
 
+/* Prefixed aliases for the C11 atomic typedefs. C++ pre-C++23 does not
+   expose the unqualified <stdatomic.h> typedefs at global scope, so use
+   these names when declaring atomic fields in headers shared between C
+   and C++ translation units. */
+#if defined(__cplusplus)
+typedef std::atomic_bool   rtapi_atomic_bool;
+typedef std::atomic_char   rtapi_atomic_char;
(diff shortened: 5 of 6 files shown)


BsAtHome:
I fixed the rtapi/rtapi_atomic.h some time ago to actually be useful and compile successfully when I fixed the hal streamer code.
I just never tested the kernel mode side. However, it seems that the RTAI code just compiles fine and it uses rtapi_atomic.h. Therefore, the default wrappers should be available in kernel mode and should be used (just include <rtapi_atomic.h>).

grandixximo:
Thanks @BsAtHome. Switched to `<stdatomic.h>` via `<rtapi_atomic.h>`, force-pushed the squash.

src/emc/motion/emcmotutil.c
```diff
@@ -7,26 +7,60 @@
...
-    errlog->end = 0;
-    errlog->num = 0;
-    errlog->tail = 0;
+    atomic_store_explicit(ATOMIC_FIELD(&errlog->write_reserve), 0ULL,
+                          memory_order_relaxed);
+    atomic_store_explicit(ATOMIC_FIELD(&errlog->write_commit),  0ULL,
+                          memory_order_relaxed);
+    atomic_store_explicit(ATOMIC_FIELD(&errlog->read_seq),      0ULL,
+                          memory_order_relaxed);
+    atomic_thread_fence(memory_order_release);
```
BsAtHome: I think you only need to fence at init. You just need to ensure that no writes are moved beyond the fence.

src/emc/motion/motion.h
```diff
@@ -739,14 +739,22 @@ Suggestion: Split this in to an Error and a Status flag register..
...
     typedef struct emcmot_error_t {
-	unsigned char head;	/* flag count for mutex detect */
 	char error[EMCMOT_ERROR_NUM][EMCMOT_ERROR_LEN];
-	int start;		/* index of oldest error */
-	int end;		/* index of newest error */
-	int num;		/* number of items */
-	unsigned char tail;	/* flag count for mutex detect */
+	unsigned long long write_reserve;	/* next slot seq to claim */
+	unsigned long long write_commit;	/* highest published seq */
+	unsigned long long read_seq;		/* next slot seq to consume */
```
BsAtHome: Shouldn't these be simply declared atomic? Then you can just read/write them without the awkward pointer/cast hack.

BsAtHome:
There does seem to be a problem with atomics in kernel compile mode...
Check the rip-rtai check.

grandixximo:
> There does seem to be a problem with atomics in kernel compile mode... Check the rip-rtai check.

I think it was because of the includes order, has to be last.

Edit:

It was the order, passes now.

src/emc/motion/emcmotutil.c
grandixximo: Just to make sure I read you right, are you saying the trailing atomic_thread_fence(memory_order_release) is OK to keep, or that I should drop it (since the caller publishes emcmotError itself, providing happens-before for any later observer)?

BsAtHome: The trailing fence is required so that no writes are carried beyond the fence. However, the order in which the values are initialized is unimportant.

All other atomicity requirements should be guaranteed when you declare the variables as atomic. (that is another comment I wrote)

BsAtHome:
You resolved the type declaration comment without saying why... So, why did you?

grandixximo:
> You resolved the type declaration comment without saying why... So, why did you?

I must have done it by mistake, wasn't intentional, I was wondering why it was compacted lol
I see now, it says I did mark it as resolved, must have clicked while scrolling.

src/emc/motion/motion.h
```diff
@@ -739,14 +740,18 @@ Suggestion: Split this in to an Error and a Status flag register..
...
+#ifdef __cplusplus
+	std::atomic<unsigned long long> write_reserve;
+	std::atomic<unsigned long long> write_commit;
+	std::atomic<unsigned long long> read_seq;
+#else
+	_Atomic unsigned long long write_reserve;
+	_Atomic unsigned long long write_commit;
+	_Atomic unsigned long long read_seq;
+#endif
     } emcmot_error_t;
```
BsAtHome: The type `atomic_ullong` should be used as they are generic and interchangeable between C/C++.

src/emc/motion/motion.h
grandixximo: I force pushed with atomic declaration.

src/emc/motion/motion.h
grandixximo: fixing

src/emc/motion/emcmotutil.c
```diff
@@ -7,26 +7,44 @@
...
-    errlog->end = 0;
-    errlog->num = 0;
-    errlog->tail = 0;
+    atomic_store_explicit(&errlog->write_reserve, 0ULL,
+                          memory_order_relaxed);
+    atomic_store_explicit(&errlog->write_commit,  0ULL,
+                          memory_order_relaxed);
+    atomic_store_explicit(&errlog->read_seq,      0ULL,
+                          memory_order_relaxed);
+    atomic_thread_fence(memory_order_release);
```
hdiethelm, line 45: I think atomic is not needed here. emcmotErrorInit() is not called from multiple threads or we have a serious issue.

src/emc/motion/motion.h
grandixximo: @BsAtHome `atomic_ullong` is not in scope from `<atomic>` or `<stdatomic.h>` under `-std=gnu++20` (libstdc++ gates the C-compat aliases on `__cpp_lib_stdatomic_h`, which requires C++23). How would you like me to proceed?

src/emc/motion/emcmotutil.c
grandixximo: right that init is single-threaded; the atomic stores aren't for thread safety. With the fields declared `_Atomic`, plain assignment becomes `seq_cst` per C11 anyway; the explicit `memory_order_relaxed` stores plus the trailing release fence are the pattern @BsAtHome asked for in the inline review (relaxed + one fence). Happy to switch to plain assignment if both of you prefer it.

src/emc/motion/emcmotutil.c
```diff
@@ -35,22 +53,42 @@ int emcmotErrorPutfv(emcmot_error_t * errlog, const char *fmt, va_list ap)
...
+              EMCMOT_ERROR_LEN);
     dbuf_iter_init(&it, &errbuf);
     vstashf(&it, fmt, ap);
 
-    errlog->end = (errlog->end + 1) % EMCMOT_ERROR_NUM;
-    errlog->num++;
-
-    errlog->tail = errlog->head;
+    while (atomic_load_explicit(&errlog->write_commit,
+                                memory_order_acquire) != my_seq) {
```
hdiethelm: This might deadlock. Reason: If a RT thread waits for a RT thread with lower prio on the same CPU, the lower thread newer gets the CPU and it is locked.

hdiethelm:
Some quick comments from my side. But I have to look into it in detail. This atomic stuff is always complicated and can have delays / deadlock if not done properly.

An RT thread should never spinnlock and even yield() can lock due to the cpu takes always the highest priority thread except priority inheritance is given but the way it is done now, it looks like it is not the case.

grandixximo:
Reworked. Dropped the centralized `write_commit` and the in-order publish spin. Each slot now has its own `slot_pub[i]` marker; producer release-stores `slot_pub[S % N] = S + 1` after writing payload, no producer-producer wait. Consumer at seq R checks `slot_pub[R % N] == R + 1`. Force-pushed for review.

src/emc/motion/emcmotutil.c
grandixximo: I think you are right, I recreated the race 😳

src/emc/motion/motion.h
BsAtHome: Hm,... https://en.cppreference.com/c/header/stdatomic states C11 and https://en.cppreference.com/cpp/atomic/atomic states C++11.

So, where does this go wrong? Is this maybe a kernel headers thingy?

src/emc/motion/emcmotutil.c
BsAtHome: Maybe placing the memory fence is just me playing it safe ;-)
<sub>(probably done too much embedded hacking where you are much more explicit and conservative.)</sub>

src/emc/motion/emcmotutil.c
BsAtHome: It also creates a situation where the fifo no longer is lock free... And that was the point, wasn't it?

src/emc/motion/motion.h
```diff
@@ -83,6 +83,7 @@ to another.
 #define MOTION_ID_VALID(x) ((x) != MOTION_INVALID_ID)
 
 #ifdef __cplusplus
+#include <atomic>
 extern "C" {
```
BsAtHome, line 89: Don't use `#include <atomics>`. That is why one reason rtapi_atomic.h was created. Rtapi's atomics should be valid across both C and C++.

src/emc/motion/motion.h
grandixximo: @BsAtHome triple-checked locally debian 13: `atomic_ullong` is not in scope under `-std=gnu++20` (libstdc++ gates the C-compat aliases on `__cpp_lib_stdatomic_h`, C++23 only). Project is C++20.

hdiethelm:
There is one thing we can assume for this queue which makes the algorithm simpler:
The queue is not getting to many messages, so we can poll in reader until all was written and all writers are finished.

So might be something like this would work, quick and dirty pseudo code:
```
push(in):
  atomic(write_reserve++ % SIZE)
  data[write_reserve] = in
  atomic(write_commit++ % SIZE)
  
pop(out):
  if(atomic(w = (write_reserve == write_commit))){ //I don't know if there is an atomic to do this
      if(read_ptr < w){
         out=data[read_ptr]
         read_ptr++ % size //No atomic needed, single consumer
      }
  }
  ```

File: src/emc/motion/motion.h (hunk 3 above)
Comment (grandixximo) [reply src/emc/motion/motion.h]: cppreference splits across two pages. C `<stdatomic.h>` provides global `atomic_ullong` since C11. C++ `<atomic>` provides `std::atomic_ullong` (namespaced) since C++11. The unqualified C-style names at global scope come from C++ `<stdatomic.h>`, which is C++23 onward (P0943): https://en.cppreference.com/w/cpp/header/stdatomic.h.

Locally on g++ 14.2:

```

#include <stdatomic.h>
int main() { atomic_ullong x{0}; (void)x; return 0; }

$ g++ -std=gnu++20 -c t.cc
error: 'atomic_ullong' was not declared in this scope
$ g++ -std=gnu++23 -c t.cc      # ok
```
libstdc++14's `<stdatomic.h>` confirms: using `std::atomic_ullong`; is inside `#ifdef __cpp_lib_stdatomic_h // C++ >= 23`. The macro is not defined under `-std=gnu++20`.

**Comment by grandixximo** [comment]
@hdiethelm thanks for the alternative design. Question: in your version the consumer drains only when `write_reserve == write_commit`, so if any one producer is preempted mid-write, the consumer stalls until that producer resumes and bumps `write_commit`. Other completed producers' messages also wait, even though their slots are ready. For an ERR-rate queue this should be fine, but I want to confirm: is that consumer stall acceptable to you, or should I keep the per-slot publish marker (current `d4c4d786b2`) which lets the consumer make progress as soon as the next contiguous slot is published?

File: src/emc/motion/motion.h (hunk 3 above)
Comment (BsAtHome) [reply src/emc/motion/motion.h]: Well, we could always do it the rtapi way and prefix the types and add this to rtapi/rtapi_atomic.h:
```C
#ifdef __cplusplus
typedef std::atomic_bool   rtapi_atomic_bool;
typedef std::atomic_char   rtapi_atomic_char;
typedef std::atomic_schar  rtapi_atomic_schar;
typedef std::atomic_uchar  rtapi_atomic_uchar;
typedef std::atomic_short  rtapi_atomic_short;
typedef std::atomic_ushort rtapi_atomic_ushort;
typedef std::atomic_int    rtapi_atomic_int;
typedef std::atomic_uint   rtapi_atomic_uint;
typedef std::atomic_long   rtapi_atomic_long;
typedef std::atomic_ulong  rtapi_atomic_ulong;
typedef std::atomic_llong  rtapi_atomic_llong;
typedef std::atomic_ullong rtapi_atomic_ullong;
#else
typedef atomic_bool   rtapi_atomic_bool;
typedef atomic_char   rtapi_atomic_char;
typedef atomic_schar  rtapi_atomic_schar;
typedef atomic_uchar  rtapi_atomic_uchar;
typedef atomic_short  rtapi_atomic_short;
typedef atomic_ushort rtapi_atomic_ushort;
typedef atomic_int    rtapi_atomic_int;
typedef atomic_uint   rtapi_atomic_uint;
typedef atomic_long   rtapi_atomic_long;
typedef atomic_ulong  rtapi_atomic_ulong;
typedef atomic_llong  rtapi_atomic_llong;
typedef atomic_ullong rtapi_atomic_ullong;
#endif
```

hdiethelm:
I think the simpler the code the better as long as it fulfills the purpose.

The consumer doesn't really stall, it only skips it until the next cycle. As long as there is no message storm, this should be no issue and happen like once in a year. 

Producers are all threads, consumer the emctaskmain which consumes only one message per cycle.

It can be an issue if no messages are printed any more due to for example the stepper thread writes a message at 25kHz but as long as the ringbuffer is not overwritten when it is full, as soon it is full, a message will be written again. But this is not in my pseudo code.

hdiethelm:
Ah but there is an edge case: If a producer task freezes or crashes mid write, the ring buffer is locked forever.

hdiethelm:
There is a reason why lock free queue algorithms have papers about it, it's hard to get it right:
https://www.cs.rochester.edu/u/scott/papers/1996_PODC_queues.pdf

grandixximo:
Thanks both. Pushed `be3497e196`: `rtapi_atomic_*` typedefs added per @BsAtHome, simpler counter design per @hdiethelm (CAS-claim + `fetch_add(write_commit)`, consumer gates on `reserve == commit`).

grandixximo:
Agreed, lock-free is treacherous. Walked through the race cases on the current design before pushing: producer-producer concurrency, mid-write skip, buffer wraparound, and the memory-ordering on `fetch_add(write_commit)` release vs consumer acquire. The quiescence gate is what makes it tractable: out-of-order commits are tolerated because the consumer only reads when `reserve == commit`, at which point all reserved slots are valid in seq order. Happy to dig into any specific scenario you want me to verify.

hdiethelm:
Thanks, looks good at first glance. Does read_seq needs to be an atomic? emcmotErrorGet() is anyway not thread save and emcmotErrorPutfv() only reads it.

BsAtHome:
You have again that header ordering problem. Where does that happen? What is it that needs to be included before the rtapi_atomic.h header?

grandixximo:
@hdiethelm yes, `read_seq` is single-writer (consumer) but read concurrently by RT producers for the buffer-full check `w_reserve - read_seq < N`; the producer's acquire-load pairs with the consumer's release-store so the slot copy happens-before reuse. C11 calls non-atomic concurrent access UB even when natural alignment would make it work in practice.

@BsAtHome the conflict is `<stdatomic.h>` (macro `atomic_fetch_add(PTR, VAL)`) vs `<linux/atomic.h>` via `<asm-generic/atomic-instrumented.h>` (function `atomic_fetch_add(int, atomic_t*)`). Macro defined first text-substitutes the function declaration. CI chain: `motion.h => rtapi_atomic.h => stdatomic.h` defines macros, then `dbuf.h => rtapi_string.h => rtapi.h => linux/atomic.h` tries to declare the functions and breaks. `hal_lib.c` already orders `rtapi.h` before `rtapi_atomic.h` manually; the motion.h change makes that order transitive for consumers.

Heading to sleep, will pick up tomorrow.

hdiethelm:
> @hdiethelm yes, `read_seq` is single-writer (consumer) but read concurrently by RT producers for the buffer-full check `w_reserve - read_seq < N`; the producer's acquire-load pairs with the consumer's release-store so the slot copy happens-before reuse. C11 calls non-atomic concurrent access UB even when natural alignment would make it work in practice.

Yes, that's true. It normally works assuming that the CPU can write the value in one op. But this assuming makes it probably UB when you look at it from a standards viewpoint.

For testing: Do you have a machine with good real time behavior? I got mine down to ~1us jitter as long as the GPU has not a lot to do.

Testing two scenarios would make sense:
- Latency impact of the atomic when you trigger a log message in an RT task. Atomic can trigger a cache line flush and stall the CPU a bit. Should not be an issue but better save than sorry. I remember that there was a discussion somewhere about a hal module to trigger logs using signals but I don't remember where.
- Concurrent producers and a single consumer, as many threads as you have cores, let it run for a while and check all received messages. Is there a framework for unit tests in linuxcnc?

src/emc/motion/motion.h
grandixximo: Done, this is in rtapi/rtapi_atomic.h now

src/emc/motion/emcmotutil.c
grandixximo: Should be lock free now, I'm testing performance as hdiethelm suggested.

grandixximo:
Latency A/B on a non-RT N100 (no_isolcpus), 5 min each, `latency-histogram` +/- `--pushmsg`:

| Metric | A base | A stress | Δ A | B base | B stress | Δ B |
|---|---|---|---|---|---|---|
| servo sdev µs | 2.1 | 3.6 | +1.5 | 2.0 | 3.5 | +1.5 |
| servo max µs | 32.7 | 50.4 | +17.7 | 26.8 | 38.2 | +11.4 |
| base sdev µs | 0.5 | 0.7 | +0.2 | 0.5 | 0.9 | +0.4 |
| base max µs | 35.8 | 54.3 | +18.5 | 48.0 | 47.5 | -0.5 |

Same delta on servo sdev. Servo max delta is smaller on B (+11.4) than A (+17.7), so the atomic fifo is at worst neutral and shows tighter servo-max under stress. Box is non-isolated so absolute numbers are inflated; the relevant signal is the delta within each package. No regression.

Concurrency test was added as a separate commit (`7a8f52f966`): pthread driver under `tests/lowlevel/emcmot-error-mpsc/`, 8 producers x 100k messages, verifies exactly-once delivery and per-producer order. Run with `scripts/runtests -p tests/lowlevel/emcmot-error-mpsc`.

hdiethelm:
Nice, even with a test! :-) Just checked, the test fails on master, so that's good.

Just a tiny thing: `tests/lowlevel/emcmot-error-mpsc/test.sh` should probably have the executable flag set.

How do you do `--pushmsg`?

tests/lowlevel/emcmot-error-mpsc/test.c
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * Spawns NPROD producer threads pumping NMSG messages each, plus one
+ * consumer that drains. Verifies every message arrives exactly once
+ * and that messages from each individual producer arrive in producer-
+ * local sequence order. Across producers any interleaving is allowed.
+ */
+
+#include <pthread.h>
+#include <sched.h>
+#include <stdatomic.h>
```
BsAtHome: `<rtapi_atomic.h>`

tests/lowlevel/emcmot-error-mpsc/test.c
```diff
@@ -0,0 +1,112 @@
...
+#include <emcmotcfg.h>
+#include "motion.h"
+#include "dbuf.h"
+#include "stashf.h"
+
+#define NPROD 8
+#define NMSG  100000
+
+static emcmot_error_t errlog;
+static atomic_int producers_done = 0;
```
BsAtHome: rtapi_atomic_int

tests/lowlevel/emcmot-error-mpsc/test.sh
```diff
@@ -0,0 +1,33 @@
...
+if [ -n "${HEADERS}" ] && [ -f "${HEADERS}/../src/emc/motion/emcmotutil.c" ]; then
+    SRCDIR="${HEADERS}/../src/emc/motion"
+    INCDIR="${HEADERS}"
+elif [ -f "${TOPDIR}/src/emc/motion/emcmotutil.c" ]; then
+    SRCDIR="${TOPDIR}/src/emc/motion"
+    INCDIR="${TOPDIR}/include"
+else
+    echo "skip: motion sources not found (need in-tree build)"
+    exit 0
+fi
```
BsAtHome: HEADERS is the global include dir IIRC. Why should you need to use it here? This can only be tested RIP anyway. Can't this then be made simpler?

tests/lowlevel/emcmot-error-mpsc/test.sh
```diff
@@ -0,0 +1,33 @@
...
+
+gcc -O -std=gnu11 -pthread -DULAPI -Wall \
+    -I"${INCDIR}" \
+    -I"${SRCDIR}" \
+    "${SCRIPTDIR}/test.c" \
+    "${SRCDIR}/emcmotutil.c" \
+    "${SRCDIR}/dbuf.c" \
+    "${SRCDIR}/stashf.c" \
+    -o "${SCRIPTDIR}/test" || { echo "compile failed"; exit 1; }
```
BsAtHome: -std=gnu11 ??

grandixximo:
> How do you do `--pushmsg`?

I made a modified version of latency-histogram with component proposed by Bertho here #3948

[pushmsg.zip](https://github.com/user-attachments/files/27326613/pushmsg.zip)
put in /src/hal/components/pushmsg.comp 
[latency-histogram.zip](https://github.com/user-attachments/files/27326624/latency-histogram.zip)
this put in /scripts

Then rebuild, and you should be able to spam messages with `--pushmsg` flag

> Just a tiny thing: tests/lowlevel/emcmot-error-mpsc/test.sh should probably have the executable flag set.

I believe I fixed it

tests/lowlevel/emcmot-error-mpsc/test.c
grandixximo: Added at the end

tests/lowlevel/emcmot-error-mpsc/test.c
grandixximo: Changed

tests/lowlevel/emcmot-error-mpsc/test.sh
grandixximo: RIP only, simplified

tests/lowlevel/emcmot-error-mpsc/test.sh
grandixximo: copy pasta, removed.

BsAtHome:
My pushmsg was a quick-and-dirty hack. Your version is pushing messages at the rate of the thread because it is level triggered. That may be very inadequate.

I suggest to make a version that can be used generically with load-time settable messages and edge trigger. Maybe something like:
```
loadrt pushmsg emsg="bla","more bla","ping pong" wmsg="oops","what about that..." imsg="told you so","telling you again"
addf pushmsg.0
...
# normaly done externally...
setp pushmsg.0.info.0 1
```
dmsg - debug
emsg - error
wmsg - warning
imsg - info

Could f.ex. use personality to set the amount of messages per level (to limit the loop).

grandixximo:
It was only used as a quick and dirty test to see if performance regressed, are you asking for a component to put in tree? Or you want me to run the test again with the more general component, but still keep it out of tree??

BsAtHome:
> It was only used as a quick and dirty test to see if performance regressed, are you asking for a component to put in tree? Or you want me to run the test again with the more general component, but still keep it out of tree??

This pushmsg was just a dirty hack for testing. But it may be interesting to be able to push custom messages. There is one component, message, but that one is rather inflexible afaics.

Need to think a bit about a generic message generator that is <i>flexible</i>. Most constructs are difficult, compute heavy or have problematic interfacing. Here a genuine KISS approach is sought after ;-)

grandixximo:
Got it, will leave it as a future separate PR. The dirty pushmsg stays a local test rig for this PR; the fifo race fix and concurrency test here don't depend on it. We should be good here right?

We can work on pushmsg over at #4003

PR #3995, opened 2026-05-02
Add headings and other geometric pins in motion using State_tags

rodw-au:
### Outline
create new **[motion.interp.xx ]** pins for heading and various geometric data (refer man motion for this PR). Note the following use cases behind this PR. 
- **heading** requested by Andy as it was trivial to add while working on this. Its been in Mach3 for years. Useful for tangential knives and saws.
- **normal-heading**  angle from current position to arc centre. Allow bevelled cutting in conjunction with heading. This could allow cuttings of bevels (eg with a plasma torch) with 2D gcode. While torch holders exist that keep the angled torch tip axial to the Z axis, external offsets could be used to calculate the desired XYZ coordinates to centre the torch tip
- **radius**   A theory of mine. Plasma cutters use torch voltage as a process contol variable to set torch height to obtain even kerf but this assumes a constant velocity (reducing velocity increases torch voltage). When the torch aproaches a centrepetal radius limit, the system could lock the height and switch to using current as the process control variable to keep the voltage constant.
- **iscircle**  allows a plasma cutter to detect bolt holes (where arc start position = end position) and apply custom hole processing rules more inteligently than the current state. Can be used in conjunction with normal-heading for countersinking holes.

I am sure the community will find many other use cases.

### Code notes

- Uses an extension of state_tags.
- New procedures tag_straight() and tag_arc() in interp_convert.cc apply our new tags for straight moves and arcs. 
- Because these tags are only applied at the beginning of a segment, heading calculations are applied in motion/control.c/update_status() in real time.
- The rest of the code is mechanical to transfer the tags added by the interpreter back to motion so the data can be used in real time.
- May provide a framework for using  additional tags in motion

### Caveats

We use the state tag motion_type (GM_FIELD_MOTION_MODE) to detect G0,G1,G2,G3 in motion. Currently there is no way to use existing interpreter equates for these values so the numbers are hard coded as 0,10,20,30.

### RunTests
These pass locally
```
Runtest: 278 tests run, 278 successful, 0 failed + 0 expected, 4 skipped, 0 shmem errors
rod@debian:~/devt/linuxcnc$ 
```
### Acknowledgements
Many thanks to Luca Toniolo (grandixximo) for his encouragement and help resolving errors with the tests.


BsAtHome:
Some observations:
* Why is there a file called `arc_heading`?
* I'm not sure that creating test-files in the root LCNC directory called `nc_files` is a good idea. Tests are supposed to be in the tests directory and have a real test associated. If they are not for testing, then you need to have a specific config and the should be created in the config/sim/... somewhere as an example.
* The line `docs/man/man9/output_buffer.9` is malplaced in the root .gitignore and should be in the `docs/man/.gitignore` file. However, it is probably malplaced in this PR as it seems unrelated to this PR.
* The diff of `src/emc/rs274ngc/interp_convert.cc` is 7000+ lines. Most, if not all, seems to be inappropriate (or stylistic wrong) white-space changes. What are the actual differences in this file?

rodw-au:
Thanks for the feedback.
1.  I will remove the spurious (empty) arc_heading
2. The nc_folder was created due to issue #3924 . I reported which was not resolved until after I commenced work on this PR. I will remove it as there is no need to demonstrate these pins with a custom sim. (I just used the normal axis sim for testing)
5. `docs/man/man9/output_buffer.9` I think appeared on acceptance of a PR from Luca so not sure how to remove it. It seems to be created  again in the build process and is not part of this feature. Added to `docs/man/.gitignore`
6. Sorry if the style has been broken. Changes in `interp_convert.cc` are two new procedures at the end `tag_straight() `and `tag_arcs() `which are each called once from `convert_straight() `and` convert_arc2()`.  New procedures are from line 7080 and on. Probably easiest to review in my fork; state_tags_arcs branch. https://github.com/rodw-au/linuxcnc/blob/state-tags-arcs/src/emc/rs274ngc/interp_convert.cc#L7080

Ideas to resolve Item 3 would be appreciated.

Just reviewing further, it appears #3924  that the ~/linuxcnc/configs folder contents have been added to the root folder on my PC. I don't think they have been included in this PR. Could someone please check and confirm?

hansu:
>docs/man/man9/output_buffer.9 I think appeared on acceptance of a PR from Luca so not sure how to remove it. It seems to be created again in the build process and is not part of this feature. Added to docs/man/.gitignore

Sorry my fault, I forgot to add the generated man page to gitignore (#3941)

BsAtHome:
Hm, I'm not sure I've been looking too good at the nc directory... Point <strike>1</strike>2 makes 300+ files disappear. That seems very, very wrong. (edit should read point 2)

Point 3 should be a separate PR, I guess.

Point 4 needs to be solved by you taking the original file and only change what is necessary.

BsAtHome:
Just to add a comment on nc files.

It is not that you cannot or should not add (example) files. The point is that there needs to be some kind of docs or procedures that use or reference them. Just having files without people knowing how to find them is kinda pointless.

BsAtHome:
> > docs/man/man9/output_buffer.9 I think appeared on acceptance of a PR from Luca so not sure how to remove it. It seems to be created again in the build process and is not part of this feature. Added to docs/man/.gitignore
> 
> Sorry my fault, I forgot to add the generated man page to gitignore (#3941)

You are not the first who gets bitten by this. There is a change in the works to let CI fail when this happens.

rodw-au:
> Just to add a comment on nc files.
> 
> It is not that you cannot or should not add (example) files. The point is that there needs to be some kind of docs or procedures that use or reference them. Just having files without people knowing how to find them is kinda pointless.

As mentioned in the original comment, these new pins are documented in the motion man file eg `man motion`  under I_Interpreter Metadata pins._ These should find hteir way to the web site shouldn't they?

Re point 2,  you now see the problem with the nc_files folder caused by #3924 . I too only saw it it when the files were removed by git add.  It was very wrong as it created the files that should have been under ~/linuxcnc in whatever was the current working directory. So  saving the sim to your desktop created this mess!

BsAtHome:
> As mentioned in the original comment, these new pins are documented in the motion man file eg `man motion` under I_Interpreter Metadata pins._ These should find hteir way to the web site shouldn't they?

The man page is a reference for the component. These man pages usually do not do a lot of examples. For that we have the other documentation in the docs/src directory, where specific scenarios are described and more in depth explanations are (should be) presented. This is usually also the place, while explaining and going into depth, to reference the examples provided.


> Re point 2, you now see the problem with the nc_files folder caused by #3924 . I too only saw it it when the files were removed by git add. It was very wrong as it created the files that should have been under ~/linuxcnc in whatever was the current working directory. So saving the sim to your desktop created this mess!

Yes, there was that unfortunate wrong side-effect while fixing the Tcl9 compatibility problem, sorry about that. However, it should be fixed now, isn't it?

BsAtHome:
> These should find hteir way to the web site shouldn't they?

Man pages should find their way, yes. So does the other documentation afaik.
(if auto updating works, don't know about that)

rodw-au:
Formatting issue in `interp_convert.cc `was caused by  an errant  autoformatting editor I had no idea about.
I reverted to a copy from master branch than reapplied the changes from this PR so the last commit 5461c2d1f7a712f364dde7a595d77d7ed64b07b8 now shows changes nice and clearly.

I can't see a clean way to correct the .gitignore issue by a PR from the oversight in #3941. Its been corrected in this PR.

Re documentation, when I look at a  the docs page on the web site, I can't see any chapter where additional documentation of a few extra pins is appropriate except for the man page.  Perhaps these gcode samples could be deployed and mentioned on the man page but where do I put them so they turn up in the users nc_file examples?
[nc_files.zip](https://github.com/user-attachments/files/27326164/nc_files.zip). 

Adding example gcode for read only pins just does not seem right to me. The man page should be sufficient (as it is for every other component).

All runtests pass
`Runtest: 278 tests run, 278 successful, 0 failed + 0 expected, 4 skipped, 0 shmem errors`

I think this is now OK for approval. It is  benign as it does not add any additional features, just exposes some useful information. Breakages are unlikely.  Any fixes following user feedback can be added.

BsAtHome:
I think you also need to restore all the files you deleted from nc_files. You removed much more than you probably intended by simply deleting the entire directory. (This may be partly my fault because I suggested something like it, but I was actually only referring to the added files.)

docs/man/.gitignore
```diff
@@ -463,6 +463,7 @@ man9/xor2.9
 man9/xyzab_tdr_kins.9
 man9/xyzacb_trsrn.9
 man9/xyzbca_trsrn.9
+man9/output_buffer.9
 # Ignore po4a-generated files for other languages
 ar/*
 bg/*
```
BsAtHome: This change needs to be removed. It is not part of this PR and is solved permanently in #4002.

rodw-au:
What a nightmare, nc-files restored. I thought the folder had been created by #3924 and did not realise that was a genine folder. So much to learn about this all!

src/.gitignore
```diff
@@ -22,3 +22,4 @@
 /configure~
 /.html-images-stamp
 /.htmldoc-stamp
+/nc_files
```
BsAtHome: This addition is wrong. I don't think you should add this here at all.

.gitignore
```diff
@@ -57,3 +57,4 @@ position.txt
 *.*-stamp
 # Ignore VSCode settings
 .vscode/settings.json
+
```
BsAtHome: This is a white-space change that should not be part of the PR.

BsAtHome (changes requested):

BsAtHome:
Thanks. You are still missing file `nc_files/3D_Chips.pdf`.

I also added some comments on the gitignore changes.

src/emc/rs274ngc/rs274ngc_pre.cc
```diff
@@ -2228,8 +2228,13 @@ int Interp::active_modes(int *g_codes,
...
 
-    // Copy float-type state
-    for (i=0; i<GM_FIELD_FLOAT_MAX_FIELDS; i++)
+    // Copy float-type state. settings[] is sized ACTIVE_SETTINGS (5);
+    // GM_FIELD_FLOAT_MAX_FIELDS may be larger when extra tag fields are
+    // appended for HAL pin output. Cap at the smaller of the two to avoid
+    // overrunning the caller's stack array.
+    int n = GM_FIELD_FLOAT_MAX_FIELDS < ACTIVE_SETTINGS
+	? GM_FIELD_FLOAT_MAX_FIELDS : ACTIVE_SETTINGS;
+    for (i=0; i<n; i++)
```
rmu75, line 2237: why not use `std::min`? or even just `for(i = 0, i < ACTIVE_SETTINGS, i++)` ?

rodw-au: This was added in a PR by Luca to resolve runtime tests. Without it, because we have added additional state_tags, he maintains the active_settings array would overflow (which made sense).

This worried me because I thought the codebase should be more robust than that so I queried it with him on Discord:

_Just get some feedback on a bug you found. In interp_base.hh we have  #define ACTIVE_SETTINGS 5
Shouldn't that be  #define ACTIVE_SETTINGS GM_FIELD_FLOAT_MAX_FIELDS to prevent future overflows of  state tags related code? Or maybe better #defined as the max of the 2  equates?_

```
grandixximo — Yesterday at 11:35 PM
no, that will break the API
compatibility with axis UI and other
If we want a guard against future drift, a one-liner like:
static_assert(ACTIVE_SETTINGS <= GM_FIELD_FLOAT_MAX_FIELDS,
              "ACTIVE_SETTINGS slot must exist in fields_float[]");

would catch anyone adding an ACTIVE_SETTINGS entry without growing the tag, without coupling GM_FIELD_FLOAT_MAX_FIELDS with ACTIVE_SETTINGS. 

Using max(5, 11) = 11 would re-create the exact bug my PR fixed
```
I did not understand the static_assert and was going to research it once I got this done and raise it here if necessary.

rodw-au:
> Thanks. You are still missing file `nc_files/3D_Chips.pdf`.

3D_chips.pdf exists but is in a .gitignore  file
rod@debian:~/devt/linuxcnc$ git add nc_files/3D_Chips.pdf --dry-run
The following paths are ignored by one of your .gitignore files:
nc_files/3D_Chips.pdf

I would not have added that!

rodw-au:
Local tests pass
`Runtest: 278 tests run, 278 successful, 0 failed + 0 expected, 4 skipped, 0 shmem errors`

src/emc/rs274ngc/rs274ngc_pre.cc
rodw-au: > static_assert is like assert (it checks an expression and only continues if expression is true) but it "runs" at compile time. So all quantities in the expression need to be known at compile time.
> 
> max would be the "wrong dircetion", you want to copy those settings that both structures can accomodate (the minimum).
> 
> static_assert like
> 
> ```
> static_assert(ACTIVE_SETTINGS == GM_FIELD_FLOAD_FEEDRATE); 
> ```
> 
> would probably be a good idead nonetheless. Frankly, that whole `state_tags` structure is a mess and should be converted into a proper struct with proper members, but that's a different story...
> 
> What is the difference between `GM_FIELD_FLOAT_FEED` and a `GM_FIELD_FLOAT_FEEDRATE`?

I will leave it to others with more knowledge to edit this code. Luca fixed it twice after I broke it when merging!

re `GM_FIELD_FLOAT_FEED` and a `GM_FIELD_FLOAT_FEEDRATE`
I'm not sure. I will review that another day. I do recall  Dewey added a feed_upm pin (name from memory) using state tags when they were first released but now I see a number of other feed rate pins. machine units per minute was not particularly useful because if using it you would normally  want to compare it with current_vel which is in machine units per second (for plasma cutting you want to disable torch height control if current_vel < original feederate to prevent diving.

I found the packed flags OK to work with but the messy bit is the data has to be repeated in a C++ structure in the interpreter. So we have a classic case of data redundancy. Using a union placing bit mapped fields over  the unsigned int might help but that's likely to be just as messy!

The other thing I learnt is that the tag is applied at the start of  the segment so to see arc headings in real time, you need to calculate them in motion. Thats also a bit messy as I mentioned in my initial comment.

rmu75: static_assert is like assert (it checks an expression and only continues if expression is true) but it "runs" at compile time. So all quantities in the expression need to be known at compile time.

max would be the "wrong dircetion", you want to copy those settings that both structures can accomodate (the minimum).

static_assert like
```
static_assert(ACTIVE_SETTINGS == GM_FIELD_FLOAD_FEEDRATE); 
```
would probably be a good idead nonetheless. Frankly, that whole `state_tags` structure is a mess and should be converted into a proper struct with proper members, but that's a different story...

What is the difference between `GM_FIELD_FLOAT_FEED` and a `GM_FIELD_FLOAT_FEEDRATE`?

src/emc/motion/state_tag.h
```diff
@@ -98,6 +99,12 @@ typedef enum {
     GM_FIELD_FLOAT_SPEED,
     GM_FIELD_FLOAT_PATH_TOLERANCE,
     GM_FIELD_FLOAT_NAIVE_CAM_TOLERANCE,
+    GM_FIELD_FLOAT_FEEDRATE,
```
rmu75: GM_FIELD_FLOAT_FEEDRATE isn't referenced anywhere it seems, the "feedrate" pins take the value from the value indexed by GM_FIELD_FLOAT_FEED.

BsAtHome:
> 3D_chips.pdf exists but is in a .gitignore file rod@debian:~/devt/linuxcnc$ git add nc_files/3D_Chips.pdf --dry-run The following paths are ignored by one of your .gitignore files: nc_files/3D_Chips.pdf

Yes, PDF files are ignored by the `*.pdf` rule in the main .gitignore file because docs generate them. Not sure whether we should make an exception or move the rule to the docs. Definitely sub-optimal, but letting it just stand as is seems fine for now, I guess. You can always argue in a separate PR that a change is necessary ;-)

src/emc/motion/state_tag.h
```diff
@@ -67,7 +68,7 @@ typedef enum {
  * WARNING:
  *
  * 1) Since these are used as array indices, they have to start at 0,
- * be monotonic, and the MAX_FIELDS enum MUST be last in the list.
+ * be monotonic, and the MAX_FIELDS GM_FIELD_FLOAT_MAX_FIELDS enum MUST be last in the list.
```
BsAtHome: The enum is called `StateFieldFloat`

src/emc/rs274ngc/interp_internal.hh
```diff
@@ -769,7 +777,7 @@ struct setup
   int defining_sub;                  // true if in a subroutine defn
   const char *sub_name;                    // name of sub we are defining (free this)
   int doing_continue;                // true if doing a continue
-  int doing_break;                   // true if doing a break
+  int doing_break;                   // true if doing a breakiscircle
```
BsAtHome: Why did this comment change? Is there something special here?

src/emc/motion/state_tag.h
rodw-au: This was incorrect when we started (in Master branch). Corrected to GM_FIELD_MAX_FIELDS to avoid any ambiguity as Line 66 states:
 * Enum for various fields of state info that are int type.

src/emc/rs274ngc/interp_internal.hh
rodw-au: An auto formatting  errror by vscode. It insisted on copying text into places it had no business to do so.
I stopped using it and reverted to Geany because of this behaviour very early in this project. `iscircle` removed.

src/emc/motion/state_tag.h
rodw-au: Agreed. reviewed at my end and removed

BsAtHome:
You still have the `nc_files/3D_Chips.pdf` marked as removed. You need to run git restore on that file and commit.

src/emc/rs274ngc/interp_internal.hh
BsAtHome: Using an editor that tries to write code for you will always fail you... ;-)

src/emc/rs274ngc/interp_internal.hh
```diff
@@ -528,7 +536,7 @@ struct block_struct
     // true if in a remap procedure the code being remapped was
     // referenced, which caused execution of the builtin semantics
     // reason for recording the fact: this permits an epilog to do the
-    // right thing depending on whether the builtin was used or not.
+    // right thing depending oniscircle whether the builtin was used or not.
```
BsAtHome: Is this another "editor glitch"?

rodw-au:
Corrected  typos in comments. (1 was pre existing)
Rebased to master branch to bring in .gitignmore entry for docs/man/man9/output_buffer.9
removed unused tag GM_FIELD_FLOAT_FEEDRATE which was not necessary in the end.

`Runtest: 278 tests run, 278 successful, 0 failed + 0 expected, 4 skipped, 0 shmem errors`

Thanks for everyone's help

src/emc/rs274ngc/interp_internal.hh
rodw-au: yes, fixed, searched  for `iscircle` using grep to check no more

rodw-au:
I had to force add Chips as it had been removed so restore failed.
```
rod@debian:~/devt/linuxcnc$ git restore ./nc_files/3D_Chips.pdf
error: pathspec './nc_files/3D_Chips.pdf' did not match any file(s) known to git
rod@debian:~/devt/linuxcnc$ git add ./nc_files/3D_Chips.pdf
The following paths are ignored by one of your .gitignore files:
nc_files/3D_Chips.pdf
hint: Use -f if you really want to add them.
hint: Disable this message with "git config advice.addIgnoredFile false"
rod@debian:~/devt/linuxcnc$ git add -f ./nc_files/3D_Chips.pdf
rod@debian:~/devt/linuxcnc$ git commit
[state-tags-arcs 4fc14bca89] force added missing 3D_Chips.pdf file
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 nc_files/3D_Chips.pdf
rod@debian:~/devt/linuxcnc$ git push
```

rodw-au:
I don't trust Githubs tests, all runtests still pass locally
`Runtest: 278 tests run, 278 successful, 0 failed + 0 expected, 4 skipped, 0 shmem errors`

src/emc/motion/control.c
```diff
@@ -2033,10 +2048,8 @@ static void output_to_hal(void)
 	/* point to joint struct */
 	joint = &joints[joint_num];
 	joint_data = &(emcmot_hal_data->joint[joint_num]);
-
 	/* apply backlash and motor offset to output */
-	joint->motor_pos_cmd =
-	    joint->pos_cmd + joint->backlash_filt + joint->motor_offset;
```
BsAtHome, line 2039: This also seems to be just whitespace changes. Any specific reason to do so?

src/emc/motion/motion.c
```diff
@@ -563,6 +578,7 @@ static int init_hal_io(void)
     CALL_CHECK(hal_param_s32_newf(HAL_RO, &(emcmot_hal_data->debug_s32_0), mot_comp_id, "motion.debug-s32-0"));
     CALL_CHECK(hal_param_s32_newf(HAL_RO, &(emcmot_hal_data->debug_s32_1), mot_comp_id, "motion.debug-s32-1"));
 
+
```
BsAtHome: Another just whitespace change...

src/emc/motion/state_tag.h
```diff
@@ -118,7 +124,9 @@ struct state_tag_t {
     float fields_float[GM_FIELD_FLOAT_MAX_FIELDS];
 
     // Any G / M code states that doesn't pack nicely into a single bit
-    // These are an array mostly because it's easier to pass an
+    // These are an array mostly because In function ‘output_to_hal’,
+
+    //it's easier to pass an
     // arbitrary-length array through NML than individual fields
```
BsAtHome: What happened to this comment?

src/emc/rs274ngc/interp_internal.hh
```diff
@@ -781,7 +789,6 @@ struct setup
   context sub_context[INTERP_SUB_ROUTINE_LEVELS];
   int call_state;                  //  enum call_states - indicate Py handler reexecution
   offset_map_type offset_map;      // store label x name, file, line
-
```
BsAtHome, line 784: Whitespace change... any reason?

src/emc/motion/control.c
rodw-au: there is a fair bit of added code in control.c. Those lines start at 2051 now but yest the blank line was removed. Will reinsert.

src/emc/motion/motion.c
rodw-au: Was at line 565, removed whitespace
Also noted a pin was commented out on line 566. restored

src/emc/motion/state_tag.h
rodw-au: Corrected by removing spurious text on line 127 and reformatting. Man, I'll never use vscode again!

BsAtHome: I never understood anyone using that crappy editor-trying-to-be-everything-and-getting-it-all-wrong thingy (/rant). Vim rocks (Emacs friends are accepted too) ;-)

src/emc/rs274ngc/interp_internal.hh
rodw-au: Looked untidy. No reason for whitespace in that group of variables. Other blank lines were accompanied by a comment.

BsAtHome:
The problem with purely whitespace changes are that they are detractors from what is being changed. As such, it may be fine to change whitespace, but in a complex change it can become hard to keep track. (/me being pedantic)

rodw-au:
Yeh, I know you are being pedantic but as you say this is complex so I don't mind. Helps me learn.
I'm getting lots of runtests errors like this here
```
rod@debian:~$ cd devt/linuxcnc
rod@debian:~/devt/linuxcnc$ . ./scripts/rip-environment
rod@debian:~/devt/linuxcnc$ runtests
Running test: /home/rod/devt/linuxcnc/tests/abort/feed-rate
*** /home/rod/devt/linuxcnc/tests/abort/feed-rate: XFAIL: test run exited with 1
*** SHMERR: Shared memory segment 0x00000064 (Emc motion key) was not removed. Removing...

```
I did a git clean and  rebooted but they are persisting. It can't be anything in that last commit surely. The only code I touched was adding back in the commented out pin.

Any ideas short of recloning?

andypugh:
> * **normal-heading**  angle from current position to arc centre.

I assume that this is just heading +/- 90 degrees. (I thought it might be pointless, but then realised that there is a +/- flip to consider)

BsAtHome:
Then there is a problem in the exit procedure of the motion process. That can be because you have a crash that goes undetected?

Run `runtests -s -n`. Then the process will stop at the first error and the generated files are not cleaned. You should then be able to look at f.ex. the generated `stderr` and other output files in the failed test's directory. They may give you an indication what happened.

rodw-au:
Thanks. Something odd. 44 tests fail and we only ever had 5 max before Luca helped me clean them up and I did runtests at my last commit and the output was still on the screen when I started tonight.
```
rod@debian:~/devt/linuxcnc$ runtests -s -n
Running test: /home/rod/devt/linuxcnc/tests/abort/feed-rate
*** /home/rod/devt/linuxcnc/tests/abort/feed-rate: XFAIL: test run exited with 1
Runtest: 1 tests run, 0 successful, 1 failed + 0 expected, 0 skipped, 0 shmem errors
Failed: 
/home/rod/devt/linuxcnc/tests/abort/feed-rate
```

Linuxcnc_debug.txt in home folder says
```
error: Cannot find instance 1 of '[DISPLAY]INTRO_GRAPHIC'
error: Cannot find instance 1 of '[DISPLAY]INTRO_TIME'
```

BsAtHome:
What is in file in `/home/rod/devt/linuxcnc/tests/abort/feed-rate/stderr`? There may be other during the test generated files in there.

rodw-au:
Hmm, so it says its an ini file thing but I have no idea where/if runtests has one. I check the env is set correctly for the rip path,

BsAtHome:
> Hmm, so it says its an ini file thing but I have no idea where/if runtests has one. I check the env is set correctly for the rip path,

You cannot depend on the files files in the home directory. Those are mostly not used in the tests.

If unsure about RIP setup, run the tests from the src directory like `../scripts/rip-environment runtests -s -n`, which should always work as intended.

You can specify the test you want to run like: `../scripts/rip-environment runtests -s -n ../tests/abort/feed-rate`, which will only run the test in that test directory. But you have to look for result and stderr files in that test directory.

rodw-au:
hmm, looks like it is to do with enabling that pin
```
rod@debian:~/devt/linuxcnc/tests/abort/feed-rate$ cat stderr
+ rm -f sim.var
+ linuxcnc -r test.ini
Note: Using POSIX realtime
emc/tooldata/tooldata_common.cc, 315: Failed to open tool table file 'simpockets.tbl': No such file or directory
emc/tooldata/tooldata_common.cc, 315: Failed to open tool table file 'simpockets.tbl': No such file or directory
HAL: ERROR: pin_new(motion.jog-is-active) called with already-initialized memory
HAL: ERROR: duplicate variable 'motion.jog-is-active'
MOTION: init_hal_io() failed
motmod: rtapi_app_main: Operation not permitted (-1)
./core_sim.hal:8: waitpid failed /home/rod/devt/linuxcnc/bin/rtapi_app motmod
./core_sim.hal:8: /home/rod/devt/linuxcnc/bin/rtapi_app exited without becoming ready
./core_sim.hal:8: insmod for motmod failed, returned -1
Note: Using POSIX realtime
```

rodw-au:
rod@debian:~/devt/linuxcnc/tests/abort/feed-rate$ cat stderr
+ rm -f sim.var
+ linuxcnc -r test.ini
Note: Using POSIX realtime
emc/tooldata/tooldata_common.cc, 315: Failed to open tool table file 'simpockets.tbl': No such file or directory
emc/tooldata/tooldata_common.cc, 315: Failed to open tool table file 'simpockets.tbl': No such file or directory
HAL: ERROR: pin_new(motion.jog-is-active) called with already-initialized memory
HAL: ERROR: duplicate variable 'motion.jog-is-active'
MOTION: init_hal_io() failed
motmod: rtapi_app_main: Operation not permitted (-1)
./core_sim.hal:8: waitpid failed /home/rod/devt/linuxcnc/bin/rtapi_app motmod
./core_sim.hal:8: /home/rod/devt/linuxcnc/bin/rtapi_app exited without becoming ready
./core_sim.hal:8: insmod for motmod failed, returned -1
Note: Using POSIX realtime

rodw-au:
duplicate pin, deleted the one I  uncommented; see what happens

rodw-au:
now passing. I must have copied that line down to use as a template when I had syntax errors with the pins

rodw-au:
> > * **normal-heading**  angle from current position to arc centre.
> 
> I assume that this is just heading +/- 90 degrees. (I thought it might be pointless, but then realised that there is a +/- flip to consider)

Yeh, normally this angle would be from the centre out (which is calced to get the tangent) but this points from the path back to the centre so we can cut a bevel. I think a component could use external offsets based on these formulas
image

So we could cut a countersunk hole. (the dxf that is used to generate the gcode would have the bolt hole and the outline of the countersink) so we'd use 2 operations and it would be all 2D gcode  for 5 axis operations. I've recruited  a couple of guys to help with this. Might end up with some components and sims once  this is squared away.

rodw-au:
`Runtest: 278 tests run, 278 successful, 0 failed + 0 expected, 4 skipped, 0 shmem errors`
Hopefully we are done

rodw-au:
Also had a brief look about tangential knives. a voicecoil actutator with 5mm stroke is reasonabley priced so hook that to the end of a stepper motor to hold the knife and we should be able to vary stroke frequency with velocity so as it slows, strokes become slower... would be a pretty simple machine to make...

andypugh:
> So we could cut a countersunk hole

I already looked into this a _long_ time ago for kerf correction:
https://sourceforge.net/p/emc/mailman/emc-users/thread/AANLkTin08wfVDXmAUO3y0FrHPGdJL91XKlENEMM4iOjv%40mail.gmail.com/#msg25474106

rodw-au:
> > So we could cut a countersunk hole
> 
> I already looked into this a _long_ time ago for kerf correction: https://sourceforge.net/p/emc/mailman/emc-users/thread/AANLkTin08wfVDXmAUO3y0FrHPGdJL91XKlENEMM4iOjv%40mail.gmail.com/#msg25474106

Well that was a long time ago! I have been told water jet machines adjust the cutting head angle to allow for perpendicular cutting. Even plasma has a tapered kerf. I was thinking of countersinking with something simple like
```
M68 E0 Q45 (Set 'Bevel Angle' variable to 45 degrees)
M64 P1     (Turn on 'Bevel Mode' via digital out)
G2 X10 Y10 I5 J0 (Cut a standard circle)
M65 P1     (Turn off Bevel Mode)
```
But external offsets would allow the cutting angle to be adjusted dynamically for kerf correction. QTPYVCP's Monokrom plasma GUI includes a material/machine SQL database I contributed containing high definition cutting parameters so it could be extended to include kerf correction for specific materials.

The beauty is that it can all be done in 2D  which avoids expensive 5 axis  post processing and keeps the operator in familiar 2D territory.

andypugh:
> Well that was a long time ago!

Indeed, but there may still be some valid points raised in that discussion.

rodw-au:
> > Well that was a long time ago!
> 
> Indeed, but there may still be some valid points raised in that discussion.

Agreed, but I think actioning those points is out of scope of this PR. This is an enabler for other derived enhancements which most likely can be delivered via components.

Is there anything else I need to do to finalize the PR so it can be accepted?
It shows there is still one requested change, but I can't locate what it is. I thought I had addressed all concerns yesterday.

rmu75:
I think my comment about ACTIVE_SETTINGS is still open.

The situation is somewhat ugly. If state_tags or settings was a "proper" structure with named members, you wouldn't use a for loop but explicit assignments. The code as it is is a bit fragile and will lead to a silent bug if somebody extends ACTIVE_SETTINGS without also inserting the same field into the float fields of state tag structure at the same offset. I would probably add a static_assert that ensures ACTIVE_SETTINGS is 5, but that also is kinda ugly.

rodw-au:
> I think my comment about ACTIVE_SETTINGS is still open.
> 
> The situation is somewhat ugly. If state_tags or settings was a "proper" structure with named members, you wouldn't use a for loop but explicit assignments. The code as it is is a bit fragile and will lead to a silent bug if somebody extends ACTIVE_SETTINGS without also inserting the same field into the float fields of state tag structure at the same offset. I would probably add a static_assert that ensures ACTIVE_SETTINGS is 5, but that also is kinda ugly.

I'll look at the assert tomorrow. 

In fairness, some decisions were made by the original dev to make the tag structure compact because it is attached to every interpreter segment.

Clearly from the code comments state_tags was designed to be extended. This is the first serious attempt to extend the float values and its identified some issues. The challenge will  be not to break existing usage now its been in place for 5 years or so. Its a bit of a chicken and egg problem now. It needs a seperate  PR but now it needs this to be committed before we can do that.

I have an idea that may solve your concern which may work. But I need to give it some serious thought.

rmu75:
AFAIK it was reluctance to explicitly write out the serialization stuff for NML but you are right that's a different cake.

docs/src/man/man9/motion.9.adoc
```diff
@@ -263,6 +263,49 @@ Note: feed-inhibit applies to G-code commands -- not jogs.
...
+  The XY plane heading of the current linear move in degrees. Measured counter-clockwise from the +X axis (0 to 360). This could be used to control a tangential knife.
+
+* `motion.interp.arc-radius` (float, out) +
+  The radius of the current circular move. This value is 0.0 during linear moves. This could be used to modify plasma cutting parameters based on the arc radius and when hole cutting.
+
+* `motion.interp.arc-center-x` (float, out) +
+  The absolute X-coordinate of the center point for the current arc.
+
+* `motion.interp.arc-center-y` (float, out) +
+  The absolute Y-coordinate of the center point for the current arc.
```
BsAtHome: What happens when you select another plane with G18/G19? There is apparently code, but why then is the Z-coordinate missing in this list?

The UVW planes (G1[789].1) do not support arcs according to the docs. But the Z in XZ and YZ planes does seem relevant.

docs/src/man/man9/motion.9.adoc
```diff
@@ -263,6 +263,49 @@ Note: feed-inhibit applies to G-code commands -- not jogs.
...
+
+* `motion.interp.normal-heading` (float, out) +
+  the heading from the current point back to the arc circlefor the current arc.
+  
+ * `motion.interp.iscircle` (bit, out) +
+  True if the current arc is a full circle.
```
BsAtHome: What about a spiral?

docs/src/man/man9/motion.9.adoc
rodw-au: I would have to check again when I'm in front of my PC but I thought in the interp::convert functions still treated the Z planes as XY coordinates hence the apparent oversight.

Options:

1. add motion.interp.arc-centre-z and copy the appropriate X or Y value in  tag_straight() and tag_arc() to it based on the active plane. 
2. publish a pin for the active plane and the consuming component maps the values as required
3. only populate the pins if in the XY plane

Option 2 is interp-centric, option 1 is user-centric, option 3 locks to the normal use case.

docs/src/man/man9/motion.9.adoc
rodw-au: Good question. Currently, I think it would be true if the XY start = end regardless of the number of turns.
I think it should also look at the delta z and limit the value to a 2D plane
The use case is for 2D CAM so I think that is reasonable.

src/emc/rs274ngc/rs274ngc_pre.cc
rodw-au: Is this code even necessary? We copy settings[] to state_tags in interp_write.cc and here we copy the same data back from state_tags to settings[] here in rs274ngc_pre.cc

Are we loading data for the GUI to see? Maybe someone with more knowledge can help here.

docs/src/man/man9/motion.9.adoc
rodw-au: Now fixed, docs updated. A helix will not trigger this pin becasue it unlikely you want this for plasma or tangential cutting

docs/src/man/man9/motion.9.adoc
rodw-au: Now fixed.  all features are supported on all planes  XY, XZ, YZ  (G17,G18,G19) eg option 1

rodw-au:
> AFAIK it was reluctance to explicitly write out the serialization stuff for NML but you are right that's a different cake.

Still todo. 
Today was a fairly big body of work enabling full supports for all planes.

rmu75:
> Still todo.
> Today was a fairly big body of work enabling full supports for all planes.

Don't bother with the NML stuff, that is a different thing, the int and float arrays in the state_tags should remain (for now).

rodw-au:
> Don't bother with the NML stuff, that is a different thing, the int and float arrays in the state_tags should remain (for now).
[nc_files.zip](https://github.com/user-attachments/files/27476533/nc_files.zip)

Thanks.
Then its complete!
I just pushed a correction for the failing RTAI test
`Runtest: 278 tests run, 278 successful, 0 failed + 0 expected, 4 skipped, 0 shmem errors`

Some simple gcode files in the attached .zip file if anybody wants to test it. Just put all the motion.interp.* pins into halshow and observe the values while running
**Files in Zip are:**
headings-test1.ngc -  arcs on XY plane (G17)
headings-test2.ngc  - straight lines on XY plane (G17)
headings-test1-g18.ngc   -  Arcs on  ZX plane (G18)
headings-test1-g19.ngc  - Arcs on YZ plane (G19)
 helix.ngc (simple helix to show **iscircle** pin will not go true becasue circular motion not restricted to a single plane)

BsAtHome:
Would it be possible to add a test under, lets say, `tests/motion/heading/` and verify that the code is and remains functional?

rodw-au:
I'll have a think about it tomorrow. You could probably test straight headings as they are set at the beginning of a segment but that's not really helpful as you really want to check the arcs which are dynamic. I wonder if you could use hal streamer? A fail would be where the heading does not change during the run. Could that be done?

rodw-au:
I'm a bit stuck here. I have the test mostly working, Linuxcnc opens, the gcode file is processed but I am not seeing any of the halsampler output in the results file. If someone could test and update or advise what to do, it would be much appreciated.

Steps Outstanding:
1. obtain halsampler output
2. write checkresult script

rodw-au:
OK, after a bit more testing with the hal file on a normal axis config, halsampler immediately outputs a buffer full of readings before the GUI opens which means that the buffer is fully depleted before the gcode file is even loaded!
I guess the reason for this is the enable pin is set by default which must be before the digital output I used to enable the sampler is set up.

Unless this can be resolved, its not possible to implement a runtest for the heading. Any ideas please let me know

BsAtHome:
It may be appropriate to add more pins to the streamer:
* motion.interp.line-number
* motion.interp.motion-type
* motion.interp.feedrate
* motion.interp.heading

Then instantiate the sampler in HAL (could be a normal HALFILE):
```
loadrt sampler depth=400 cfg=sfff
net linenum  motion.interp.line-number => sampler.0.pin.0
net mottype  motion.interp.motion-type => sampler.0.pin.1
net feedrate motion.interp.feedrate    => sampler.0.pin.2
net heading  motion.interp.heading     => sampler.0.pin.3

loadusr -w halsampler -n 10000
# Do not enable the sampler or it will start to fill up before we are actually doing something
```

In the UI script:
```python
...
# Load program
c.program_open("heading-test.ngc")

# Enable sampling
h.set_p("sampler.0.enable", "1");

# Run program
c.auto(linuxcnc.AUTO_RUN, 1)
...
```

The UI script should wait for the halsampler executable to finish or it must send a SIGTERM to halsample to instruct it to terminate.

BsAtHome:
Just came to think of it...

I actually fixed the python HAL stream interface. You don't need the halsampler program and could read the data directly in the UI wrapper in its loop.

```python
sampler = h.stream(h, h.sampler_base, "sfff")
...
# sampler.read() returns a tuple with the sample value
def read_samples():
  while sampler.readable:
    print(sampler.read()) # maybe this should do something else...
...

# Don't wait, loop with sampling
# l.wait_for_interp_state(linuxcnc.INTERP_IDLE)
while True:
  s.poll()
  if s.interp_state == linuxcnc.INTERP_IDLE:
    break
  if sampler.readable:
    read_samples()
  time.sleep(0.001)

h.set_p("sampler.0.enable", "0")
time.sleep(0.1)
read_samples() # drain rest
...
# Do other stuff
```

rodw-au:
Thanks, I decided to stick with halsampler for now, then try your new method but I am having a bit of trouble with this

```
  File "/home/rod/devt/linuxcnc/tests/motion/heading/./test-ui.py", line 47, in <module>
    h.set_p("sampler.0.enable", "1")
    ^^^^^^^
AttributeError: Pin 'set_p' does not exist
```
And yes, I saw your errant C  ; line ending. :)

I did have a lot of other pins which I had tested in a normal sim but was taking baby steps :)

```
loadrt sampler depth=10000 cfg=ffffffffb

net x-pos-fb joint.0.pos-fb             => sampler.0.pin.0
net y-pos-fb joint.1.pos-fb             => sampler.0.pin.1
net heading motion.interp.heading       => sampler.0.pin.2
net normal motion.interp.normal-heading => sampler.0.pin.3
net radius motion.interp.arc-radius     => sampler.0.pin.4
net center-x motion.interp.arc-center-x => sampler.0.pin.5
net center-y motion.interp.arc-center-y => sampler.0.pin.6
net center-z motion.interp.arc-center-z => sampler.0.pin.7
net iscircle motion.interp.iscircle     => sampler.0.pin.8
```
The line number is a good idea though.

BsAtHome:
You `import hal`, so it should read `hal.set_p(...)`.

Those semicolons... well, habits die hard ;-)

Using halsampler makes it more difficult because it is a program running asynchronous from the UI wrapper and you have to control its termination. Doing the sampler read in the UI wrapper eliminates that async and external program problem go away. But, sure, test it first ;-)

rodw-au:
using the old hal way failed as the test-ui never returned so I went to the python method.
I had a number of issues
1. I could not load the sampler using your syntax and resorted to some AI which suggested I use the numeric instance which worked (maybe)
2. The sampler is not returning data in Python
3. Homing was only doing one axis before failing so I changed how we do that

I ran up a normal linuxcnc  instance for testing and debugging and the script runs to completion and I could see it homing, loading and running the gcode file. But there is no sampler data arriving adding a few more print statements

I'm really out of my depth here. I've never used the python interface before and barely know the python language. I much prefer ;'s too :) 
 
If somebody could review this after my last commit, it would be great! If we could get data printing, I should be able to devise some tests

BsAtHome:
I'll have a look.

BsAtHome:
I have a sampling version of the test. No checks performed on it yet.

The attached zip is the content of the tests/heading directory: [heading.zip](https://github.com/user-attachments/files/27550628/heading.zip)

You need to `git rm` the `.gitignore` and `sampler.hal` files. There are not necessary.

Problems found:
* The DISPLAY program needs to be a shell findable path and must be specified with leading `./` (very annoying, I know)
* You had two HAL files for the sampler setup with different content. Very confusing. Now there is only one (heading.hal) and that one is referenced in the INI-file.
* The test-ui.py script needed some tweaking to make it work as expected.
* I added a feed move to the G-code, just to see what happened

The data is currently dumped in the `result` file. This may not be appropriate and the test-ui.py script needs some updating to manage the data.

Please note that the motion-type field lags one sample when the next command is executed.

rodw-au:
Thanks so much for this. Sorry, I thought I had deleted the extra files.
I have been out tonight so will look in the morning.
There could be a lot of data as currently we've just tested a single plane. I need to review the data set but I was thinking I might check say every 200th reading on each line of gcode and compare reported vs calculated data then write out "pass" or "fail" in results. For G18/G19, I probably only need to check the x,y,z centres are correct.

rodw-au:
So had some time today. I have the tests validating now but initially had issues with radius calculations when the machine had imperial units and my gcode was in metric so the tests found an issue before they were released! Updating the ini file in the tests fixed that and all tests passed. I take every 100th entry and perform tests on them as there are 1000s of entries. All lines assessed now pass including g2,g3,g1,g0
I will look into that. Ideally it should respect the gcode measurement system, not device units.

BsAtHome:
> So had some time today. I have the tests validating now but initially had issues with radius calculations when the machine had imperial units and my gcode was in metric so the tests found an issue before they were released! Updating the ini file in the tests fixed that and all tests passed.

Yes, the units were off. Also, the feed rate units do probably not match.

> I take every 100th entry and perform tests on them as there are 1000s of entries.

That is a problem. This is coming from RT and you do not know how many samples you will get and therefore not reproducible. You probably need to assess the situation whenever the state changes.

rodw-au:
[result.txt](https://github.com/user-attachments/files/27575419/result.txt)
Now checks every reading

I have the test operating succesfully see reults.txt
I don't understand how to use checkresult. I can't use expected because of the variable timing data linuxcnc returns
I thought I'd be able to open results and search for the "Completed sucessfully" but it is not found
I did see one example that seemed to look for
`Runtest: 1 tests run, 0 successful, 1 failed + 0 expected, 0 skipped, 0 shmem errors`
but I could not understand to apply it to my situation
I know test.sh will recieve a return code from linuxcnc but I did not see how to apply that
A bit of help to close this out would be great.

Please note that unexpected deviations from the expected path set in the gcode by joint.0.pos-fb joint.0.pos-fb are out of scope of this PR. I noted that around the transition between the G2 and the G1 you added, the radius gradually deviates up to 1.2mm at the end and no G64 or G61 settings would alter this.  This was for up to 40 readings at the end of the G2 and 5 readings at the start of the G1

In the end, I surmised that the planner must blend the trajectory regardless of the expected gcode controls so I ignore the deviations in the G2 unless they exceed 40 and skip the first 5 readings in the G1

I think this is a valid approach becasuse the remaining 2700 odd readings gave perfect results and we don't touch anything with the actual motion.

I still need to review gcode vs machine units but wanted to finish this first

rodw-au:
Runtests now works. 
Comments in the commit

tests/motion/heading/test-ui.py
```diff
@@ -0,0 +1,247 @@
...
+
+#
+# Sample collection routine
+#
+
+def read_sample():
+  global nsamples, lastsample
+  sample = sampler.read()
+  if sample != lastsample: # Don't print duplicates
+    if (nsamples  % SAMPLEN  == 0):
```
BsAtHome: SAMPLEN is one (1) and should always be one. This modulo/if code is therefore noise and should be removed

tests/motion/heading/test-ui.py
```diff
@@ -0,0 +1,247 @@
...
+            heading, normal-heading, iscircle, x-pos-fb, y-pos-fb
+    """
+    global lastsample
+    global lastrecnum
+   
+    #print("{:6d} {}".format(nsamples, sample))
+    if sample[0] != lastsample[0]:
+      return True
+    if sample[0] not in (4, 5, 6):
+      # We are only interested in lines 4,5,6. Do not change gcode file
```
BsAtHome: I'm not sure you should actually test on line-number. You have the motion-type field (trailing by one sample) that should be enough? If not, why not?

In principle you want to check heading for your motion types and could do so based on that field. It would certainly simplify this code if I'm not mistaking. You can then also change the g-code to suit you without having to hack this code as well.

andypugh:
> I have the test operating succesfully see reults.txt I don't understand how to use checkresult. I can't use expected because of the variable timing data linuxcnc returns I thought I'd be able to open results and search for the "Completed sucessfully"

I have seen at least one test that pre-processes the "result" file to leave only the pertinent data before comparing.
Here is one example:
https://github.com/LinuxCNC/linuxcnc/blob/751e0c1cb9ed4b5a64ebfb674b408beaff5d905f/tests/mb2hal/mb2hal.2a/checkresult#L10

tests/motion/heading/test-ui.py
rodw-au: yes, we can probably get rid of the line number test and only test G0,1,2,3 lines based on motion type. Lines was my first idea when I had no clue where to start. I've just completed an update to check gcode units vs machine units but have not tested it. I also want to add tests using imperial machine units. (and rebase to master).

One crazy Idea I had was to write back the current heading to a next heading field in the previous segment. This would allow a component to check  if a tangential knife needed to be lifted before the next segment and use external offsets before the knife is twisted while still in the material. Would there be a way to do that? (would make it a seperate PR)

rodw-au:
> > I have the test operating succesfully see reults.txt I don't understand how to use checkresult. I can't use expected because of the variable timing data linuxcnc returns I thought I'd be able to open results and search for the "Completed sucessfully"
> 
> I have seen at least one test that pre-processes the "result" file to leave only the pertinent data before comparing. Here is one example:
> 
> https://github.com/LinuxCNC/linuxcnc/blob/751e0c1cb9ed4b5a64ebfb674b408beaff5d905f/tests/mb2hal/mb2hal.2a/checkresult#L10

Thanks Andy, I've now done something similar this morning by overwriting the results file with success/failure messages from tests.sh. Thinking about it, I probably could enable verbose messaging so that if the test fails, runtests -n would leave the error messages and test data for review.

rodw-au:
This is now complete.  Changes include:
1. Changes to radius calculations to ensure G20/G21 units are respected regardless of machine units
2. Streamlined tests so now universal and not tied to line numbers
3. Now there are two tests based on machine units heading-mm (original metric) and heading-in (imperial) - both pass
4. Modified gcode file to run both in G20 and G21 units used in both tests
5. Merged upstream master into this branch and resolved conflicts. I made a mistake resolving conflicts but its resolved in a subsequent commit. (forgot to delete some lines)

All runtime tests pass locally.

rodw-au:
Last minute I realised I had an error in gcode (not using G20) and this was not picked up because I was not using the corrected radius in interp_conv.cc.
Both issues corrected.
Both heading-mm and heading-in tests pass

```
$ runtests tests/motion/heading-mm
Running test: tests/motion/heading-mm
Runtest: 1 tests run, 1 successful, 0 failed + 0 expected, 0 skipped, 0 shmem errors
$ runtests tests/motion/heading-in
Running test: tests/motion/heading-in
Runtest: 1 tests run, 1 successful, 0 failed + 0 expected, 0 skipped, 0 shmem errors

```

src/emc/motion/command.c
```diff
@@ -530,6 +530,15 @@ void emcmotCommandHandler_locked(void *arg, long servo_period)
 		SET_JOINT_FAULT_FLAG(joint, 0);
 	    }
 	    emcmotStatus->paused = 0;
+	    // Clear pins on abort so tests see a clean state
+		if (emcmot_hal_data) {
+				*(emcmot_hal_data->interp_arc_radius) = 0.0;
+				*(emcmot_hal_data->interp_arc_center_x) = 0.0;
+				*(emcmot_hal_data->interp_arc_center_y) = 0.0;
+				*(emcmot_hal_data->interp_straight_heading) = 0.0;
```
BsAtHome, line 539: Isn't the z-value missing here?

src/emc/rs274ngc/interp_write.cc
```diff
@@ -333,6 +333,16 @@ int Interp::write_state_tag(block_pointer block,
...
+    
+    // Pack new geometric data. block is NULL on the M70 save path
+    // (save_settings() in interp_convert.cc), so guard against null deref.
+    state.fields_float[GM_FIELD_FLOAT_STRAIGHT_HEADING] = (block == NULL) ? 0.0f : (float)block->arc_heading;
+    state.fields_float[GM_FIELD_FLOAT_ARC_RADIUS]       = (block == NULL) ? 0.0f : (float)block->arc_radius;
+    state.fields_float[GM_FIELD_FLOAT_ARC_CENTER_X]     = (block == NULL) ? 0.0f : (float)block->arc_center_x;
+    state.fields_float[GM_FIELD_FLOAT_ARC_CENTER_Y]     = (block == NULL) ? 0.0f : (float)block->arc_center_y;
+    state.fields_float[GM_FIELD_FLOAT_ARC_CENTER_Z]     = (block == NULL) ? 0.0f : (float)block->arc_center_z;
+    state.fields_float[GM_FIELD_FLOAT_NORMAL_HEADING]   = (block == NULL) ? 0.0f : (float)block->normal_heading;
+    state.flags[GM_FLAG_IS_CIRCLE]                      = (block == NULL) ? false : block->iscircle;
```
BsAtHome: These are explicitly cast to `float`. You should let the compiler take care of that. Also, I think that the `state.fields_float` should actually be a double, as all the rest of LCNC is double.

Wouldn't it be more natural to write this like:
```C++
if (null != block) {
  state.fields_float[GM_FIELD_FLOAT_STRAIGHT_HEADING] = block->arc_heading;
  state.fields_float[GM_FIELD_FLOAT_ARC_RADIUS] = block->arc_radius;
 ...
} else {
  state.fields_float[GM_FIELD_FLOAT_STRAIGHT_HEADING] = 0.0;
  state.fields_float[GM_FIELD_FLOAT_ARC_RADIUS] = 0.0;
 ...
}
```

tests/motion/heading-in/test-ui.py
```diff
@@ -0,0 +1,237 @@
...
+
+#
+# Sample collection routine
+#
+
+def read_sample():
+  global nsamples, lastsample
+  sample = sampler.read()
+  if sample != lastsample: # Don't print duplicates
+    if (nsamples  % SAMPLEN  == 0):
```
BsAtHome: Using a modulo with one makes no sense. You want to test all samples so this while condition should be removed.

tests/motion/heading-in/test-ui.py
```diff
@@ -0,0 +1,237 @@
...
+
+#
+# Sample collection routine
+#
+nsamples = 0
+lastsample = None
+lastrecnum = 0
+ctr = 0
+
+def is_arc_valid(sample, tolerance=0.11):
```
BsAtHome: You never call with empty tolerance. Therefore, you should remove the default from the argument.

(also, the default tolerance of 0.11 is a bit huge, don't you think?)

tests/motion/heading-in/test-ui.py
```diff
@@ -0,0 +1,237 @@
...
+
+import time
+import sys
+import ast
+import math
+
+
+INTERPTIMEOUT = 5   # Max 5 seconds to wait for the interpreter
+GCODETIMEOUT  = 120  # Max 60 seconds for the gcode file to run12
+SAMPLEN       = 1  # Sample ewvery n'th record
```
BsAtHome: The SAMPLEN construct needs to be removed

tests/motion/heading-mm/heading.hal
```diff
@@ -0,0 +1,22 @@
+# HAL file for sampling motion heading data
+
+loadrt sampler depth=400 cfg=ssfffffffbff
+
```
BsAtHome: This is code duplication and is a very bad strategy. You should separate the configs and data from the code and reuse things.

You can f.ex. create the directory structure:
* tests/motion/header/metric
* tests/motion/header/imperial

and you leave the common components under tests/motion/header/ or, if you prefer, in tests/motion/header/shared or something like that.

That also brings me to another issue. What about <i>mixed</i> units? When your machine is metric and the g-code is imperial or vice versa. Has that been tested?

tests/motion/heading-mm/test-ui.py
```diff
@@ -0,0 +1,237 @@
+#!/usr/bin/env python3
+
+import linuxcnc
+import linuxcnc_util
+import hal
+
```
BsAtHome: Code duplication. See other comment.

BsAtHome:
You have in the test g-code:
```ngc
G64 P0.1 Q0.1 	; Tolerance 0.1mm
G21             	; Set units to millimeters
...
```
Should these two be reversed? First set the machine into metric and then set the tolerance to 0.1mm?

tests/motion/heading-mm/heading.hal
BsAtHome: I've seen that you do mixed test via the g-code test file. So that is great.

Making an example how I'd do it. Will post later.

rodw-au:
I will review suggestions tomorrow.
My very last commit (which is quite small) introduced an error on /tests/interp/python/error and is also stopping Linuxcnc from opening with `undefined symbol emcStatus`. I think it is some kind of linker voodoo which I don't understand. It must be the changes in interp_convert.cc rather than the gcode edits because Linuxcnc errors before it even opens. What I am trying to do is to check if the gcode units = machine units
If you had any ideas on that, please let me know.
I was going to try using the units in units.h when I understand them but it was meal time!
I want to resolve that before tackling revisions.

BsAtHome:
Please find attached [heading2.zip](https://github.com/user-attachments/files/27709824/heading2.zip) that has the merged tests/heading/ content. It also updates the ui script with the points that I mentioned in review.

The CI problem is likely caused by your master branch being <i>divergent</i> from LinuxCNC master branch. You added some commits directly to your master branch and those changes should probably have been in your state-tags-arcs branch. You need to save all your changes, then revert/reset your master branch to match LinuxCNC's and then apply your saved changes to your state-tags-arcs branch. Then you can look and find where you made the error.

BsAtHome:
That should go into `tests/motion/heading`, that is.

rodw-au:
> That should go into `tests/motion/heading`, that is.
Thanks, the subfolders are a great idea. I tried to get it all in one folder but didn't know enough!
Sometimes, One of the great things about being an Aussie is people like you work our night shift!

You were correct about the divergence. It happened in the first batch of commits and it must have finally bitten. Now fixed. I noticed you corrected the tolerance. But I think its now looser on an inch machine thjan a metric machine. Not sure how to fix that. Can you pass a paramter to the ui file? I guess it won't really matter becasue if its off it will be a G20/G21 issue and it will be far more than that!

rodw-au:
> You have in the test g-code:
> 
> ```
> G64 P0.1 Q0.1 	; Tolerance 0.1mm
> G21             	; Set units to millimeters
> ...
> ```
> 
> Should these two be reversed? First set the machine into metric and then set the tolerance to 0.1mm?

I noticed you had fixed that.

tests/motion/heading-mm/test-ui.py
rodw-au: resolved with your contribution

tests/motion/heading-mm/heading.hal
rodw-au: Adopted your changes. I liked you ran each arc twice!
Yes, we do test mixed units on both imperial and metric machines.

tests/motion/heading-in/test-ui.py
rodw-au: Resolved in your changes thanks!
I just wasn't sure if the sample size would become too large so did not want to throw it away but I increased the feedrate to offset.

tests/motion/heading-in/test-ui.py
rodw-au: Removed and tested
As said elsewhere the tolerance of 0.001 mm used now is quite a lot tighter than 0.001 inches. Is that an issue?

tests/motion/heading-in/test-ui.py
rodw-au: resolved

src/emc/motion/control.c
```diff
@@ -2033,10 +2048,9 @@ static void output_to_hal(void)
 	/* point to joint struct */
 	joint = &joints[joint_num];
 	joint_data = &(emcmot_hal_data->joint[joint_num]);
-
+
```
BsAtHome: On this line you (only) add trailing whitespace.

I discovered that you have trailing whitespace on lines multiple places in your code. You should check for that and remove it.

src/emc/rs274ngc/interp_write.cc
rodw-au: corrected

src/emc/motion/command.c
rodw-au: these lines are uperflous as they are all set (including z) further down at around line 2249.
Removed

src/emc/rs274ngc/interp_write.cc
```diff
@@ -333,7 +333,29 @@ int Interp::write_state_tag(block_pointer block,
...
+    // Pack new geometric data. block is NULL on the M70 save path
+    // (save_settings() in interp_convert.cc), so guard against null deref.
+    if(nullptr != block){
+		state.fields_float[GM_FIELD_FLOAT_STRAIGHT_HEADING] = (float)block->arc_heading;
+		state.fields_float[GM_FIELD_FLOAT_ARC_RADIUS]       = (float)block->arc_radius;
+		state.fields_float[GM_FIELD_FLOAT_ARC_CENTER_X]     = (float)block->arc_center_x;
+		state.fields_float[GM_FIELD_FLOAT_ARC_CENTER_Y]     = (float)block->arc_center_y;
+		state.fields_float[GM_FIELD_FLOAT_ARC_CENTER_Z]     = (float)block->arc_center_z;
+		state.fields_float[GM_FIELD_FLOAT_NORMAL_HEADING]   = (float)block->normal_heading;
+		state.flags[GM_FLAG_IS_CIRCLE]                      = block->iscircle;
```
BsAtHome: Do **not** cast these values to `float`. At best the compiler must solve this issue automatically and, actually, types will be matched after #4020. You must not artificially reduce accuracy by casting.

src/emc/motion/command.c
BsAtHome: Well, its still here...

src/emc/rs274ngc/interp_write.cc
rodw-au: fixed

src/emc/motion/control.c
rodw-au: I will review in a seperate commit from code fixes

rodw-au:
Oops Fixed command.c. Was looking in the wrong file (control.c) and found some redundant code anyway
Will put formatting in a seperate commit

src/emc/motion/control.c
rodw-au: Seems OK now

src/emc/motion/command.c
rodw-au: Now fixed. Was originally looking in the wrong file (control.c) and found superflous code there!

src/emc/motion/state_tag.h
```diff
@@ -67,7 +68,7 @@ typedef enum {
  * WARNING:
  *
  * 1) Since these are used as array indices, they have to start at 0,
- * be monotonic, and the MAX_FIELDS enum MUST be last in the list.
+ * be monotonic, and the GM_FIELD_MAX_FIELDS enum MUST be last in the list.
```
rodw-au, line 71: This corrects an ambiguity in the original comment

src/emc/motion/control.c
```diff
@@ -2035,8 +2043,8 @@ static void output_to_hal(void)
 	joint_data = &(emcmot_hal_data->joint[joint_num]);
 
 	/* apply backlash and motor offset to output */
-	joint->motor_pos_cmd =
-	    joint->pos_cmd + joint->backlash_filt + joint->motor_offset;
+	joint->motor_pos_cmd = 
+             joint->pos_cmd + joint->backlash_filt + joint->motor_offset;
```
BsAtHome: Again added trailing whitespace after '=' (line 2046) and an unnecessary change in leading whitespace (line 2047).

src/emc/motion/control.c
```diff
@@ -2211,6 +2222,90 @@ static void update_status(void)
       emcmotStatus->stepping = 0;
       emcmotStatus->paused = 1;
     }
+	// State Tags handling    
+	// Get the current executing trajectory component (the "Source of Truth")
+	/* Update the HAL Output Pins from the active tag */
+	if (emcmot_hal_data) {			
+		// Line and Motion Type
+		if (emcmot_hal_data->interp_line_number) {
+			*(emcmot_hal_data->interp_line_number) = (int)emcmotStatus->tag.fields[GM_FIELD_LINE_NUMBER];
```
BsAtHome: Two issues:
1. This added code uses tab indenting, which looks fine here, but is not looking good elsewhere because this file is a mixed jumble of tabs and spaces. You should use 4 spaces per level of indent. This happens also in other files. <i>You should review all your additions</i> in all changed files to use the style of that particular file (yes, it is a messy business). If the file uses all tabs, you also use tabs. If the file uses spaces, you also use spaces.
2. You have trailing whitespace on many lines you added here. Please clean that up too. You should check <i>all</i> your additions for trailing whitespace.

src/emc/motion/control.c
```diff
@@ -2211,6 +2222,90 @@ static void update_status(void)
       emcmotStatus->stepping = 0;
       emcmotStatus->paused = 1;
     }
+	// State Tags handling    
+	// Get the current executing trajectory component (the "Source of Truth")
+	/* Update the HAL Output Pins from the active tag */
+	if (emcmot_hal_data) {
```
BsAtHome: Why are you testing for `emcmot_hal_data` to be non-null? 

This same function derefences that exact same variable and does not check. So, the check is superfluous and creates a false sense of guard. The whole containing if() construct is redundant.

src/emc/rs274ngc/interp_convert.cc
```diff
@@ -6363,3 +6398,156 @@ int Interp::update_tag(StateTag &tag)
...
+    {
+        // For Z-only moves, reuse the last calculated XY heading
+        block->arc_heading = last_valid_heading;
+    }
+
+    // Reset Arc data for safety
+    block->radius = 0.0;
+    block->arc_center_x = 0.0; 
+    block->arc_center_y = 0.0;
+    block->iscircle = false;
```
BsAtHome, line 6446: Shouldn't you also clear arc_center_z?

tests/motion/heading/heading-test.ngc
```diff
@@ -0,0 +1,15 @@
...
+G90                      ; Set to absolute positioning
+G53 G0 X75 Y50           ; Rapid move to start point (center X50 + radius 25)
+G2 X75 Y50 I-25 J0 F5000 ; Clockwise circle to start, center is -25mm in X
+G53 G1 X50 Y75 F5000     ; Feed move reversed X and Y
+G3 X50 Y75 I25 J0 F5000  ; Counter clockwise circle back, center is +25mm in X
+G20                      ; Set units to inches
+G90                      ; Set to absolute positioning
+G53 G0 X3 Y2             ; Rapid move to start point (center X2 + radius 1)
+G2 X3 Y2 I-1 J0 F5000    ; Clockwise circle to start, center is -1" in X
+G53 G1 X2 Y3 F5000       ; Feed move reversed X and Y
```
BsAtHome: Why are there G53 (machine coordinates) in the G0/G1? If the coordinate system were different, then surely the G2/G3 would be off?

The test starts with a clean set of variables, which should mean default coordinate system and no offsets.

src/emc/motion/control.c
rodw-au: I saw the report of white space and  edited that to agree with master https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/motion/control.c#L2038

rodw-au: just to clarify on my branch, this was on one line

BsAtHome: Normally I would not be too critical, but these small whitespace movements and changes distract from what is actually being changed. That's why I'm pedantic ;-)

rodw-au: thats alright, I did keep an eye on tabs and space but now I've turned on show white spaces in geany, this one comes up as all tabs

src/emc/motion/control.c
rodw-au: This is a mess of spaces and tabs, went with spaces on my stuff

src/emc/rs274ngc/interp_convert.cc
rodw-au: done

tests/motion/heading/heading-test.ngc
rodw-au: I've always left gcode to postprocessors so you want me to delete g53's?

BsAtHome: Well, the real question is "why is it there in the first place?". This is a test, not a full-fledged program to make an object on a mill, lathe, cutter or whatever. I think it should use the defaults and if those defaults change the test may fail because assumptions are no longer true.
So, yes, I think these G53 need to be removed.

src/emc/motion/control.c
```diff
@@ -2077,10 +2085,10 @@ static void output_to_hal(void)
 	if (IS_EXTRA_JOINT(joint_num) && get_homed(joint_num)) {
 	    // passthru posthome_cmd with motor_offset
 	    // to hal pin: joint.N.motor-pos-cmd
-	    extrajoint_hal_t *ejoint_data;
-	    int e = joint_num - NO_OF_KINS_JOINTS;
-	    ejoint_data = &(emcmot_hal_data->ejoint[e]);
-	    *(joint_data->motor_pos_cmd) = *(ejoint_data->posthome_cmd)
+	    extrajoint_hal_t *ejoint_data;    // Clear pins on abort so tests see a clean state
```
rodw-au: NOt sure what went wrong here. I deleted something in error and fixed after chekcing master

src/emc/motion/control.c
```diff
@@ -2131,6 +2139,9 @@ static void update_status(void)
 		// and the state machine is still active. The homing status deassertion
 		// must be delayed until the state machine is done.
 		joint_status->homing = get_homing(joint_num);
+	/* check to see if we should pause in order to implement
```
rodw-au: another odd one here

rodw-au: Still got the gcode to fix. but code cleaned up. Tests pass locally.

src/emc/motion/control.c
```diff
@@ -2036,7 +2044,7 @@ static void output_to_hal(void)
 
 	/* apply backlash and motor offset to output */
 	joint->motor_pos_cmd =
-	    joint->pos_cmd + joint->backlash_filt + joint->motor_offset;
+			joint->pos_cmd + joint->backlash_filt + joint->motor_offset;
```
BsAtHome: Your editor does not seem to like your intent... You still changed the four spaces to one tab.

src/emc/rs274ngc/interp_convert.cc
rodw-au: Completed

src/emc/motion/control.c
BsAtHome: The master branch does not have this comment. The comment also makes no sense.

src/emc/motion/control.c
```diff
@@ -2211,6 +2219,74 @@ static void update_status(void)
       emcmotStatus->stepping = 0;
       emcmotStatus->paused = 1;
     }
+    // State Tags handling
+    // Get the current executing trajectory component (the "Source of Truth")
+    /* Update the HAL Output Pins from the active tag */
+        // Line and Motion Type
+        *(emcmot_hal_data->interp_line_number) = (int)emcmotStatus->tag.fields[GM_FIELD_LINE_NUMBER];
+
+        // Performance Metadata
```
BsAtHome: Why doesn't this line up? Still space/tab trouble, I guess.

src/emc/motion/control.c
rodw-au: I've checked this against master and it appears to be byte for byte identical so I don't know why its being flagged as a change

BsAtHome: The problem is that the original line starts with a tab and then has 4 spaces (yes, horrible). Your tree/branch replaced it with two tabs.

src/emc/motion/control.c
rodw-au: master has one tab then spaces

Image

src/emc/motion/control.c
rodw-au: seems right now in my editor

BsAtHome: The diff clearly disagrees.

src/emc/motion/control.c
rodw-au: Looks like I missed formatting this block. Still had tabs, fixed

rodw-au:
Finally! All cleaned up. Sorry about the stream of commits.
There is now only 2 deletions. One by Luca dealing with Active-tags and the orther was me tidying up a comment in state-tag.h

I'm very relieved our revised gcode passes all tests
`Runtest: 281 tests run, 281 successful, 0 failed + 0 expected, 4 skipped, 0 shmem errors`

Thanks for all your help and suggestions. It will be great to see this committed now. Some users are keen for it as it was a forum post that spurred me on to get this done. I had an attempt about 6 years ago and gave up but it was good to start afresh with this commit as things had changed and I had done even more silly things than I did in this time!

BsAtHome (approved):

BsAtHome:
Looks fine to me ;-)

BsAtHome:
@andypugh you need to squash the commits.
There are many individual commits in Rod's tree and many are not relevant to the actual change, but cosmetics. Now, I don't want to cause yet another delay, so squashing this while merging should solve it.

rodw-au:
Should I rebase to master at this point?
Such a learning curve.

BsAtHome:
In principle, you could squash the commits locally into one commit with a new message and then force push the result. It is a bit tricky, so make a backup before you attempt and verify locally before you push. But the squash can also be done in the merge here.

rodw-au:
@andypugh Now rebased to master and squashed to a single commit. 

Ready to  merge into master.
All tests pass locally.
`Runtest: 281 tests run, 281 successful, 0 failed + 0 expected, 4 skipped, 0 shmem errors`

PR #3996, opened 2026-05-04
interp: make #5401-#5409 reflect the applied tool length offset (fixes #2994)

### LINKED ISSUE #2994: Tool Offset Parameters #5401-5409
 1.  Install Linuxcnc
 2.  Open any Linuxcnc configuration.
 3.  Load a tool offset using G43Hn, where "n" is not equal to the loaded tool number. 

I expected the values displayed by parameter #5401-5409 to equal the values saved in the tool table for "n".

What happened is that the values displayed for parameters #5401-5409 were either all zero if no tool is loaded, or equal to the values saved for the current loaded tool number.

If the developers feel the current behavior is correct, I would like to make this a request to add either a new set of number parameters or global named parameters that will display the current "applied" tool offset values, inducing any additional tool offsets from G43.2.  

I do not know if this has ever worked correctly, but it behaves as described above in 2.5, 2.9, and 2.10(Master).

I have checked this behavior on multiple systems and distributions, including virtual machine simulation.

grandixximo:
## Summary
Parameters #5401-#5409 now report the tool length offset actually
applied to motion. Matches existing docs ("Set by G43") and Fanuc
#5081-#5088 / Haas #5081-#5086 semantics.

## Changes
- `convert_tool_length_offset` writes `#5401-#5409` for G43, G43Hn
  (any n), G43.1, G43.2, and G49 (zero).
- `set_tool_parameters` / `default_tool_parameters` /
  `init_tool_parameters` no longer touch `#5401-#5409` (M6 changes
  the loaded tool but does not by itself apply its offset).
- `convert_setup_tool` (G10 L1/L10/L11) no longer touches them.
- `#5400` and `#5410-#5413` keep their existing M6 / G10 / startup
  update behavior.

## Behavior change (intentional)
Third-party g-code that reads `#5401-#5409` between an M6 / G10 and
the next G43 will now see the offset still in effect (or zero), not
the new tool's stored table values. This matches the docs and
industry precedent; the previous behavior is what issue #2994 calls
out as wrong.

## Test plan
- [x] No tool loaded, `G43H5` -> params reflect tool 5
- [x] Tool 3 loaded, `G43H7` -> params reflect tool 7
- [x] `G43H5; G43.2 X1.0` -> `#5401` = tool5.x + 1.0
- [x] `G49` -> params zero
- [x] `M6` alone does not change `#5401-#5409`
- [x] `G10 L1` on loaded tool does not change `#5401-#5409`
- [x] `#5400`, `#5410-#5413` unchanged behavior

Fixes #2994


grandixximo:
Modified the tests to check for the new behavior

Issue #3979, opened 2026-04-29
axis: Tool pointer not shown in master


hdiethelm:
2.9:
<img width="274" height="201" alt="Image" src="https://github.com/user-attachments/assets/dd6c4c27-ebdf-4906-a884-3f7c09bd6449" />

master:
<img width="303" height="221" alt="Image" src="https://github.com/user-attachments/assets/33b50bea-a32a-4c9e-a02f-b33cde5bb3f6" />

andypugh:
Possibly caused by 
https://github.com/LinuxCNC/linuxcnc/commit/4a22a67b664567c0af04269eed52b09a41a658cf
I had a bit of trouble merging this due to conflict with the new INI parser changes. 
@BsAtHome can you take a look at:
https://github.com/LinuxCNC/linuxcnc/blob/master/lib/python/rs274/glcanon.py#L571
Possibly I misunderstood "fallback"

BsAtHome:
> can you take a look at: https://github.com/LinuxCNC/linuxcnc/blob/master/lib/python/rs274/glcanon.py#L571 Possibly I misunderstood "fallback"

You refer to this line:
```python
self.tool_min_dia = self.inifile.getreal("DISPLAY", "GCODE_VIEW_TOOL_MIN_DIA", fallback=0.0)
```
When `[DISPLAY]GCODE_VIEW_TOOL_MIN_DIA` is <i>invalid or not set</i>, then it returns the fallback value of 0.0.

The problem, I guess, is that the value of 0.0 will make the tool vanish from view. But looking a bit further...

In glcanon.py:1501:
```python
  if current_tool is None or current_tool.diameter < self.view_tool_min_dia:
    ...
```
That condition was previously a comparison <i>with</i> zero (`if current_tool is None or current_tool.diameter == 0:`). But that condition is no longer triggered when you compare to <i>less than</i> and the value is zero.

My guess is that the condition must be <i>less than or equal</i> (`<=`) in the comparison.

BsAtHome:
BTW, in `lib/python/rs274/glcanon.py`, it just occurred to me that there are addressed two different variables `self.tool_min_dia` and `self.view_tool_min_dia`.

The member `self.view_tool_min_dia` is set to 0.0 on line 521 and never used elsewhere in the file. Grep'ing the member in the entire code base turns up only an assignment to zero and the compare in line 1501.

[...]

PR #3999, opened 2026-05-04
test: UI smoke tests for axis, touchy, gmoccapy, qtdragon

grandixximo:
Draft, opening for CI feedback. Refs #3756.

## Summary

Phase 1 of the GUI test work tracked in #3756. Each test launches a GUI under `xvfb-run` against an existing `configs/sim/<gui>/*.ini`, drives Estop reset / machine on / home all via NML, asserts the interpreter reaches IDLE, then shuts down cleanly. Verifies the GUI starts and accepts basic commands without crashing.

## Coverage

- axis
- touchy
- gmoccapy
- qtdragon (qtdragon_xyz/qtdragon_metric.ini)

## Mechanics

- `tests/ui-smoke/_lib/launch.sh`: `xvfb-run` wrapper, `setsid` so the linuxcnc process group can be signalled cleanly, falls back to `axis-remote --quit` then SIGTERM with grace then SIGKILL. Skips with exit 77 if `xvfb-run` is unavailable (matches `tests/tooledit` and `tests/pyvcp`).
- `tests/ui-smoke/_lib/drive.py`: NML driver. Tolerant of sim configs that come up already in `STATE_ON` via auto-estop-release HAL wiring. Falls back to per-joint serial homing if no `HOME_SEQUENCE` is configured.
- `tests/ui-smoke/_lib/checkresult.sh`: pass when `UI_SMOKE_OK` printed and no crash markers in captured logs.
- Reuses existing sim configs, no test-only INI files.

## Cleanup discipline

- `.gitignore` covers all runtime artifacts (`linuxcnc.{out,err,pid}`, `ui-smoke.{out,err}`, `result`, `stderr`)
- 4 consecutive runs locally: 4/4 pass, 0 shmem errors, working tree clean (no untracked files added beyond the committed test scripts). Aligns with the clean-tree gate Bertho asked for and that @hdiethelm is wiring up in #3984.

## Deps

`xvfb` is already declared in `debian/control` with the `<!nocheck>` profile so `apt-get build-dep` installs it on the existing CI without a workflow change. Coordinated with @hdiethelm in #3984: this PR adds no system deps; if his lands first, no rebase needed here.

## Out of scope (deferred)

- Phase 2: load a small G-code file via `linuxcnc.command.program_open` + `auto(RUN)`, verify final position via `linuxcnc.stat.position`. Per-GUI cross-checks via `xdotool` or AT-SPI where useful.
- Phase 3: screenshot or short video on failure, uploaded as CI artifact.

## Test plan

- [x] Local: 4/4 pass under `scripts/runtests tests/ui-smoke`, no shmem leaks
- [x] CI: rip-and-test passes
- [x] Reviewer feedback on scope: shipped smoke ("does it start, NML reachable, no startup crash"), per Bertho's framing in #3756. Functional behaviour (load G-code, verify position) tracked as Phase 2 follow-up.


tests/ui-smoke/_lib/launch.sh
```diff
@@ -0,0 +1,126 @@
...
+        # Run linuxcnc in its own process group so we can signal the whole
+        # group cleanly (linuxcnc forks task, motion, GUI, halrun).
+        setsid linuxcnc -r '$CONFIG_INI' >linuxcnc.out 2>linuxcnc.err &
+        LINUXCNC_PGID=\$!
+        echo \$LINUXCNC_PGID >linuxcnc.pid
+
+        # Give task time to come up before driver attaches. The GUI also
+        # needs time to register and home up to the point where it accepts
+        # commands; 8s is conservative for headless sim runs.
+        sleep 8
```
BsAtHome: Isn't there a way to detect what you are waiting for? Timed starts may fail when CI is busy and real-clock timeouts do not match the machine's activity.

This may also be a problem other places where a real-clock timeout is playing out.

tests/ui-smoke/_lib/launch.sh
```diff
@@ -0,0 +1,126 @@
...
+        # Last resort if Cleanup never finished. Use halrun -U to drop
+        # any HAL state we may have left behind, and remove any of
+        # linuxcnc's known shared memory keys explicitly so the next
+        # ui-smoke test does not inherit a corrupted environment.
+        if kill -0 \$LINUXCNC_PGID 2>/dev/null; then
+            echo 'WARN: linuxcnc did not exit on SIGTERM, escalating to KILL' >&2
+            kill -KILL -\$LINUXCNC_PGID 2>/dev/null || true
+            sleep 2
+            halrun -U 2>/dev/null || true
+            for key in 0x48414c32 0x48484c34 0x00000064; do
```
BsAtHome: Duplicate key list. Make an array once and index that. Then you can also add stuff when needed. (see also key list above)

tests/ui-smoke/_lib/launch.sh
```diff
@@ -0,0 +1,126 @@
...
+                shmid=\$(ipcs -m | awk -v k=\$key 'tolower(\$1)==k {print \$2}')
+                [ -n \"\$shmid\" ] && ipcrm -m \$shmid 2>/dev/null || true
```
BsAtHome: Duplication? --> function?

tests/ui-smoke/README
```diff
@@ -0,0 +1,21 @@
...
+
+Each test directory contains:
+  test.sh       launches the GUI under xvfb-run, runs drive.py, captures
+                exit code and stderr
+  checkresult   examines the captured output for crash markers
+
+Shared helpers live in _lib/.
+
+If xvfb-run is not available on the host, tests skip gracefully (matches
+the precedent set by tests/tooledit and tests/pyvcp).
```
BsAtHome: I think we need to discuss this. We'd want to run this on CI and then need the xvfb-run. However, if it skips gracefully, then CI does not fail and we don't know whether the code is sane. Something like "damned if you do and damned if you don't" situation?

hdiethelm: I would just fail the test. In CI, we can make sure that xvfb-run is there and locally on a PC, it won't hurt anybody.
If anytime there is an issue in CI that can not be easily corrected, just add "continue-on-error: true" until it is fixed.

hdiethelm:
> Phase 3: screenshot or short video on failure, uploaded as CI artifact.

If you manage to create consistent screenshots and want to go to pedantic mode:

- Store reference known good screenshots (TBD where, I often use submodules for test data storage so the main repo is not overfilled and it is still tracked)
- Take screenshots at certain points where everything is static, like before / after homing / at the end
- Compare to the reference and highlight any differences, fail if there are differences -> Artifact
- The dev can download the artifacts, check them manually and if the change was on purpose replace the known good ones, so the CI passes again

Probably over complicated and I don't know how deterministic LinuxCNC is but this way, bugs like this https://github.com/LinuxCNC/linuxcnc/issues/3979 can be easily avoided. Testing manually, these kind of bugs are just often overlooked.

tests/ui-smoke/_lib/launch.sh
grandixximo: Dropped the pre-driver sleep 8. drive.py already polls echo_serial_number for task readiness, so the wall-clock wait was redundant. The remaining time-bounded primitives are timeout on the linuxcnc/driver pair and a 60-second SIGTERM grace, both of which have to bound something to keep CI from hanging on a wedge.

tests/ui-smoke/_lib/launch.sh
grandixximo: extracted into `_lib/cleanup-runtime.sh`. Single source of truth for the daemon list and shared-memory key list, called both pre-launch and as the post-SIGTERM last resort.

tests/ui-smoke/README
grandixximo: policy now matches what hdiethelm proposed. We only skip on `xvfb-run` absence (rare local dev env). Python and gi typelib deps are declared in` debian/control` under `<!nocheck>` so `apt-get build-dep` installs them on CI. Missing deps now fail the test loudly instead of silently skipping.

tests/ui-smoke/_lib/launch.sh
grandixximo: now calling `bash $LIB_DIR/cleanup-runtime.sh`

grandixximo:
> > Phase 3: screenshot or short video on failure, uploaded as CI artifact.
> 
> If you manage to create consistent screenshots and want to go to pedantic mode:
> 
>     * Store reference known good screenshots (TBD where, I often use submodules for test data storage so the main repo is not overfilled and it is still tracked)
> 
>     * Take screenshots at certain points where everything is static, like before / after homing / at the end
> 
>     * Compare to the reference and highlight any differences, fail if there are differences -> Artifact
> 
>     * The dev can download the artifacts, check them manually and if the change was on purpose replace the known good ones, so the CI passes again
> 
> 
> Probably over complicated and I don't know how deterministic LinuxCNC is but this way, bugs like this #3979 can be easily avoided. Testing manually, these kind of bugs are just often overlooked.

The reference-screenshot diff approach is a good Phase 3 idea, will track it on #3756. For Phase 1 (this PR) I'm staying with NML state assertions only since they're deterministic; rendering will need the screen-stabilization tricks you mentioned.

grandixximo:
Round 2 pushed; CI now passes 282/282 with all 4 ui-smoke tests running.

Changes in this round:

- **NML connect-and-poll robustness ([drive.py](tests/ui-smoke/_lib/drive.py))**: linuxcncsvr's status buffer can be invalid (`emcStatusBuffer invalid err=3`) for the first ~30s after startup. Driver now retries the connect-and-poll cycle, recreating the `stat` object each iteration so a stale invalid buffer does not stick. `CONNECT_TIMEOUT_S=60s`.

- **Software OpenGL ([launch.sh](tests/ui-smoke/_lib/launch.sh))**: GitHub Actions runners have no GPU and qtdragon's GLcanon widget segfaulted under hardware GL. Set `LIBGL_ALWAYS_SOFTWARE=1`, `GALLIUM_DRIVER=llvmpipe`, `QT_QUICK_BACKEND=software`, `QSG_RHI_BACKEND=software`, `QT_OPENGL=software`.

- **`pyqt5-dev-tools` Build-Depends**: qtvcp compiles a QRC file via `pyrcc5` at first run; without the package qtdragon segfaulted with "No such file or directory: 'rcc'". Added with `<!nocheck>` profile.

- **Driver-trust checkresult ([checkresult.sh](tests/ui-smoke/_lib/checkresult.sh))**: replaced the crash-marker grep with a simple `UI_SMOKE_OK` present + `UI_SMOKE_FAIL` absent check. The driver only prints `UI_SMOKE_OK` after a successful NML round-trip plus a re-poll after the GUI settle, so it is the authoritative signal. The previous regex was catching shutdown-side Qt teardown races that are out of scope for a startup smoke test.

- **`$HOME/linuxcnc/nc_files` mkdir workaround**: touchy's [filechooser.py:29](lib/python/touchy/filechooser.py#L29) does `os.listdir($HOME/linuxcnc/nc_files)` with no try/except and crashes on a clean `$HOME`. launch.sh pre-creates the directory as a workaround. Filing a separate issue for the underlying touchy bug.

grandixximo:
@BsAtHome this is review ready, after the review, merge this and make another PR for phase 2? or just continue here, and merge after all is done?

tests/ui-smoke/.gitignore
```diff
@@ -0,0 +1,9 @@
+# Runtime artifacts left by launch.sh; per Bertho's clean-tree rule,
+# generated build/test outputs must be gitignored or committed.
+linuxcnc.out
+linuxcnc.err
+linuxcnc.pid
+ui-smoke.out
+ui-smoke.err
+result
+stderr
```
BsAtHome: These two are already ignored at a higher up level

grandixximo:
@BsAtHome did you force rerun the CI? I got an email it failed, but looking here it did not? Was it Ubuntu's servers?

Edit:

Never Mind, it failed again, some sudo issues, will check...

tests/ui-smoke/_lib/checkresult.sh
```diff
@@ -0,0 +1,27 @@
...
+# The driver only emits UI_SMOKE_OK after a successful NML round-trip
+# (linuxcnc task ready and stat.poll() still alive after settle), so
+# this is sufficient evidence that the GUI booted. We do not grep for
+# generic crash markers like "Segmentation fault" or "Traceback":
+# linuxcnc's own scripts/linuxcnc Cleanup may emit shutdown-side
+# segfaults (Qt/GTK teardown races) that are out of scope for a
+# startup smoke test, and the driver-signal approach catches the
+# startup-time failures we actually care about.
+set -u
+LOG="${1:-/dev/stdin}"
```
BsAtHome: I'm not sure if stdin fallback is a good idea. The checkresult script is called with the result file as argument. If it is missing, then the check should fail.

That would also make the greps simpler, I guess.

BsAtHome:
Also,...have you run `scripts/shellcheck.sh ...your_scripts...`? May find something more ;-)

BsAtHome:
CI keeps failing... did you rebase?

BsAtHome:
> did you force rerun the CI? I got an email it failed, but looking here it did not? Was it Ubuntu's servers?

Yes, I did. I saw the fail on eatmydata adduser and was guessing a transient failure (those we've seen many of). As a "member" I can restart runs.

tests/ui-smoke/.gitignore
grandixximo: cleaned

tests/ui-smoke/_lib/checkresult.sh
grandixximo: Ok, failing hard now, and simple grep.

grandixximo:
I did rebase, also squashed and fixed a few thing, including your review points, shellcheck on my scripts returns clean.
Hopefully this round it passes, I'll keep an eye on it...

grandixximo:
@BsAtHome this is not me, it is failing I think cause of something that is fixed with #3984 ??

Edit:

I'll file a tiny PR to quickly patch the CI

BsAtHome:
> @BsAtHome this is not me, it is failing I think cause of something that is fixed with #3984 ??

That is not yet merged and my newly created PR #4016 worked fine... I don't understand github's CI docker images... they are intermittently unstable it seems.

grandixximo:
will see if #4017 passes

hdiethelm:
I just checked a few past CI failures. 
Mostly, adduser was missing. Building with sid will need attention from time to time, it's called sid for a reason... ;-)
This one: https://github.com/LinuxCNC/linuxcnc/actions/runs/25674529024/job/75368570073 failed because github being github (Github has a reputation for being unreliable...)

BsAtHome:
> I just checked a few past CI failures. Mostly, adduser was missing. Building with sid will need attention from time to time, it's called sid for a reason... ;-)

That was why I merged your PR today (AP usually does the merging). Your improvements fixed the issue, improved speed and structure (very considerably) and replaced another PR that would have generated a conflict with yours. How many birds were <strike>hit</strike>unharmed with that stone...

> (Github has a reputation for being unreliable...)

And it is getting worse by the day.

hdiethelm:
Yes, I hit the not available adduser also in my PR but already fixed it. Just tag me if you have an unexplained CI failure and I can quickly investigate. No I know a bit how github works.

For example this: `Error: fatal: unable to access 'https://github.com/LinuxCNC/linuxcnc/': The requested URL returned error: 500` is most probably a github internal API being down.

> > (Github has a reputation for being unreliable...)
> 
> And it is getting worse by the day.

You can follow it live here:
https://damrnelson.github.io/github-historical-uptime/

grandixximo:
Well it passes now, waiting for confirmation on 
- Phase 2: load a small G-code file via linuxcnc.command.program_open + auto(RUN), verify final position via linuxcnc.stat.position. Per-GUI cross-checks via xdotool or AT-SPI where useful.

separate PR based on this one?

BsAtHome:
This PR concentrates on "start the UI and see if it runs". That is a good start and tests a lot of code.

The next PR can then focus on "run a program", however that is going to work. There will be some complexities to solve for this to work reliably.

The question is now, how far along is _this_ PR? It is called WIP in the title, so I guess there is still some work ;-)

grandixximo:
> The question is now, how far along is this PR? It is called WIP in the title, so I guess there is still some work ;-)

Phase 1 is complete and ready for review. WIP title was just leftover from the draft cycle; dropping it now.

Status:
- CI 15/15 green on the latest push (rip-and-test, clang, cppcheck, htmldocs, full package-arch matrix).
- All inline review comments from you and @hdiethelm addressed (timeout, dup key list, dup cleanup logic, skip vs fail policy, gitignore, checkresult stdin fallback).
- `scripts/shellcheck.sh` clean on the added scripts.
- 4 consecutive local `scripts/runtests tests/ui-smoke` runs: 4/4 pass, clean tree, no shmem leftovers.

Phase 2 (load g-code, `auto(RUN)`, assert final position) will go in a separate PR on top of this one, matching what you suggested above. That keeps the "does it start" surface reviewable on its own and lets the "does it run a program" complexity land independently.

grandixximo:
the CI is stuck on fetching from azure for 5 hours now ;-(

PR #4000, opened 2026-05-04
Convert qtpdf to poppler, convert poppler to pdfium

NTULINUX:
I cannot get qtpdf working in Debian however the poppler back-end is working fine. PDFium actually has better quality than poppler however the transparent background is grey as opposed to white. Does this matter, and if so, I can make transparent colors white instead.

LinuxCNC now prints out which back-end is being used to help with debugging.

Tested on Debian (poppler) and Gentoo (PDFium)

Poppler is the default back-end for Okular which is great as well.

Feedback is appreciated!

Thanks!

### DIFF
--- a/lib/python/qtvcp/lib/qt_pdf/PDFViewer.py
+++ b/lib/python/qtvcp/lib/qt_pdf/PDFViewer.py
@@ -7,24 +7,26 @@ from qtpy.QtCore import Qt
 from qtvcp import logger
 LOG = logger.getLogger(__name__)
 
-USE_QTPDF = False
 USE_POPPLER = False
+USE_PDFIUM = False
 
 try:
-    from qtpy.QtPdf import QPdfDocument, QPdfDocumentRenderOptions
-    _QTPDF_RENDER_OPTS = QPdfDocumentRenderOptions()
-    _QTPDF_RENDER_OPTS.setRenderFlags(
-        QPdfDocumentRenderOptions.RenderFlag.Annotations
-    )
-    USE_QTPDF = True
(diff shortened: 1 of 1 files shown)


NTULINUX:
Poppler:

poppler

PDFium:

pdfium

Not sure why the buttons disappeared but I don't think it's related to this. Looks like those buttons don't do anything anyway :P

BsAtHome:
Not knowing why the buttons disappeared might not be a good sign?

c-morley:
Colors are usually adjusted with style sheets to match the rest of the color scheme.
I don't feel strongly of the two choices shown myself.

Buttons are used with HTML page. I don't know if they work with a multi page PDF - not sure if I tried that - probably meant to get back to that.
If these sample screens were run on different linux distributions, that is probably why the button image is missing. The missing images are Qt builtin images.

Please switch the print statements to LOG statements: LOG.info if it's gonna be information that will commonly be referenced, INFO.debug if it's only needed if there is a problem. (or remove it if it was just for testing while developing)

NTULINUX:
Scratch this commit, sorry about that! QTPDF works as-is as long as you have PDF enabled in pyside (which most distributions don't, but Gentoo needs it as it does not have python-poppler-qt5.) Plus, the logic here does nothing useful besides print debug info. LOG.info doesn't seem to work either.

c-morley:
You need to use  qtvcp -i to see info comments

NTULINUX:
> Not knowing why the buttons disappeared might not be a good sign?

FWIW, that happens with QTPDF before this commit as well.

PR #4002, opened 2026-05-04
doc: Ignore man-page directories because all files are auto-generated

BsAtHome:
Remove the list of man-pages to ignore and replace it with the directories. All man-pages are generated from asciidoc nowadays and keeping track of mixed troff/adoc sources is no longer required.

This should fix the longstanding problem of forgetting to add a newly added component or program to the man-page ignore list.


hansu:
Yeah good idea, why not making things easier.
PS: Please rebase because of the merge conflict now.

BsAtHome:
> Yeah good idea, why not making things easier. PS: Please rebase because of the merge conflict now.

Rebased.

PR #4003, opened 2026-05-04
hal: add pushmsg component for RT message generation

grandixximo:
KISS RT-to-non-RT message generator, replaces the dirty test-rig hack used in the #3993 latency comparison.

Each rising edge on `pushmsg.{error,warning,info,debug}-N` emits the corresponding string from the load-time `{e,w,i,d}msg` arrays through `rtapi_print_msg()`. 16 slots per level. Slots without a configured message are inert. Singleton instance.

Example:

```
loadrt pushmsg emsg="boom","other boom" wmsg="careful"
addf pushmsg servo-thread
setp pushmsg.error-00 1
```

cc @BsAtHome who proposed this in #3993.

Draft for discussion. Open to flexibility tweaks (personality bits for per-level counts, multi-instance, etc.).

### DIFF
--- a/src/hal/components/pushmsg.comp
+++ b/src/hal/components/pushmsg.comp
@@ -1,0 +1,383 @@
+component pushmsg "Push configurable RT messages to non-RT logging on rising edges";
+
+description """
+Each entry in the `msgs` modparam describes a level, pin name, and message
+text using `level|pinname|message` syntax. Level is one letter:
+`e` (error), `w` (warning), `i` (info), `d` (debug). On the rising edge of
+`pushmsg.<pinname>` the corresponding text is emitted via `rtapi_print_msg()`
+at the chosen level.
+
+Pin names use HAL-allowed characters (alphanumeric, dot, dash, underscore).
+Maximum 64 entries per loadrt, 2 substitutions per message, 128 chars per
+message.
+
+Message text may include `{t:refname}` placeholders. Each one creates a
+typed input pin `pushmsg.<refname>` that the user wires to the source
+signal. Type letter selects the pin type:
+`b` => bit, `s` => s32, `u` => u32, `l` => s64, `k` => u64, `f` => float.
+Use `\\{` to embed a literal curly brace.
+
+Example:
+[source,hal]
+----
+loadrt pushmsg msgs="e|nooil|No oil pressure","w|low|Oil {f:level} L low"
+addf pushmsg servo-thread
+net no-oil classicladder.0.out-21 pushmsg.nooil
+net oil-level oil.level pushmsg.level
+----
+
+Example INI expansion:
+[source,ini]
+----
+[PUSHMSG]
+# Note: no surrounding whitespace
+MSGS=\\
+"e|nooil|No oil pressure",\\
+"w|low|Oil {f:level} L low"
+----
+[source,hal]
+----
+loadrt pushmsg msgs=[PUSHMSG]MSGS
+----
+""";
+
+pin in bit enable = TRUE "Enable message generation";
+
+modparam dummy msgs "Comma separated list of custom messages in format t|pin|message";
+
+option period no;
+option extra_setup yes;
+option extra_cleanup yes;
+option singleton yes;
+
+function _;
+license "GPL"; // v2+
+author "LinuxCNC";
+
+include <rtapi_ctype.h>;
+include <rtapi_gfp.h>;
+include <rtapi_slab.h>;
+
+;;
+
+#define MSG_LEN_MAX     128 // Max size of a message at input and expanded
+#define MSG_N_MAX        64 // Max number of messages supported
+#define MSG_SUBST_MAX     2 // Max number of substitutions within one message
+
+// Command-line config parameter
+static char *msgs[MSG_N_MAX] = {};
+RTAPI_MP_ARRAY_STRING(msgs, MSG_N_MAX, "Message slots in 't|pin|message' format");
+
+// The HAL structure lives in HAL memory space
+typedef struct {
+    hal_bit_t  *trigger;
+    hal_data_u *substs[MSG_SUBST_MAX];
+} msg_hal_t;
+
+// The admin structure lives in normal memory space
+typedef struct {
+    bool        prev;                       // Previous pin state (to detect rising edge)
+    int         msgtype;                    // Rtapi message type (err, warn, info, debug)
+    char        msgbuf[MSG_LEN_MAX];        // Complete message with embedded NULs for substitution
+    int         msglen;                     // Length so we may find the end easily (stpcpy is not available)
+    hal_type_t  substtype[MSG_SUBST_MAX];   // Substitution types (hal type enum)
+    const char *parts[MSG_SUBST_MAX];       // Text parts index msgbuf for after substitution
+} msg_slot_t;
+
+//
+// The message structure is decomposed from the source:
+//   msgs[n] = "e|trigpin|Message {s:pin1} othertext {u:pin2} trailing"
+//
+// The source is copied into the slots->msgbuf and all pin refs are isolated:
+//   msgbuf = "Message \0s:pin1\0 othertext \0u:pin2\0 trailing"
+//                               ^                    ^
+//                               parts[0]             parts[1]
+// The slots->parts[] array is always the trailing part after the substitution.
+// The pins for the substitution variable are referenced in the pins->substs[]
+// array.
+//
+
+static msg_hal_t  *pins;    // Trigger and substitution pins
+static msg_slot_t *slots;   // Admin data for each message
+static int nslots;          // Number of active messages
+
+static inline char printable(char c)
+{
+    return isprint(c & 0xff) ? c : '?';
+}
+
+static int message_type(char c)
+{
+    switch (c) {
+    case 'e': case 'E': return RTAPI_MSG_ERR;
+    case 'w': case 'W': return RTAPI_MSG_WARN;
+    case 'i': case 'I': return RTAPI_MSG_INFO;
+    case 'd': case 'D': return RTAPI_MSG_DBG;
+    }
+    return -1;
+}
+
+static int check_pinname(const char *name)
+{
+    for (; *name; name++) {
+        if (!isascii(*name & 0xff) || (!isalnum(*name & 0xff) && NULL == strchr("._-", *name & 0xff)))
+            return -EINVAL;
+    }
+    return 0;
+}
+
+static int setup_message(const char *pfx, int idx, const char *msg, msg_slot_t *slot, msg_hal_t *pin)
+{
+    int pfxlen = strlen(pfx);
+    int l = strlen(msg);
+    if (l >= MSG_LEN_MAX) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d too long, more than %d characters\n", pfx, idx, MSG_LEN_MAX-1);
+        return -EMSGSIZE;
+    }
+    // Absolute minimum is "t|p|"
+    if (l < 4) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d too short, less than 4 characters\n", pfx, idx);
+        return -EMSGSIZE;
+    }
+    // Get the message type
+    if ((slot->msgtype = message_type(msg[0])) < 0) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d has invalid type '%c', expected one of {e,w,i,d}\n",
+                        pfx, idx, printable(msg[0]));
+        return -EBADMSG;
+    }
+    // Message type must be followed by a '|'
+    if ('|' != msg[1]) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d missing first '|' separator\n", pfx, idx);
+        return -EBADMSG;
+    }
+    // Find the second '|' in "t|name|message"
+    const char *pinend = strchr(msg+2, '|');
+    if (!pinend) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d missing pin name, no second '|' separator\n", pfx, idx);
+        return -EBADMSG;
+    }
+    // Make sure we can encompass the full name
+    int pinnamelen = pinend - msg - 2;
+    if (pinnamelen + pfxlen + 1 >= HAL_NAME_LEN) {  // +1 for the extra '.' character
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d pin name too long\n", pfx, idx);
+        return -EBADMSG;
+    }
+
+    int rv;
+    // Copy the pinname for later pin creation use
+    char pinname[HAL_NAME_LEN+1] = {};  // The init ensures termination
+    memcpy(pinname, &msg[2], pinnamelen);
+    if ((rv = check_pinname(pinname)) < 0) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d pin name contains bad characters\n", pfx, idx);
+        return rv;
+    }
+    // Create the trigger input pin
+    if ((rv = hal_pin_bit_newf(HAL_IN, &pin->trigger, comp_id, "%s.%s", pfx, pinname)) < 0) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d cannot create pin (duplicate?)\n", pfx, idx);
+        return rv;
+    }
+
+    strcpy(slot->msgbuf, msg + pinnamelen + 3);  // Copy message so we may take it apart
+
+    // Now see if there are any expansions in the string.
+    int subst = 0;  // Count number of substitutions
+    for (char *cptr = slot->msgbuf; *cptr; cptr++) {
+        // Check "\{..." escaped curly open brace
+        if ('\\' == *cptr) {
+            // The next char is escaped and cannot start a substitution
+            if (cptr[1])  // Or this is a backslash at the end
+                cptr++;
+            continue;
+        }
+        if ('{' == *cptr) {
+            // Need "{t:pinname}"
+            *cptr = 0;  // Terminate the previous string "foo {t:pin} bar" --> "foo "
+            cptr++;     // Move to subst internals
+            char *end = strchr(cptr, '}');
+            if (!end) {
+                rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d missing '}' in substitution\n", pfx, idx);
+                return -EBADMSG;
+            }
+            slot->parts[subst] = end + 1;  // This is the continuation position "t:pin} bar" --> " bar"
+            *end = 0;  // Terminate the pinname "t:pin} bar" --> "t:pin"
+
+            // Min content is 3 chars: type + ':' + 1-char-name. end-cptr counts those bytes.
+            if (end - cptr < 3) {
+                rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d substitution too small, need type and pinname\n", pfx, idx);
+                return -EBADMSG;
+            }
+            if (':' != cptr[1]) {
+                rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d substitution missing ':' after type character\n", pfx, idx);
+                return -EBADMSG;
+            }
+
+            char tchar = *cptr;
+            cptr += 2;  // Move to pinname
+            if ((end - cptr) + pfxlen + 1 >= HAL_NAME_LEN) {  // +1 for the extra '.' character
+                rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d substitution pin name too long\n", pfx, idx);
+                return -EBADMSG;
+            }
+            if ((rv = check_pinname(cptr)) < 0) {
+                rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d substitution pin name contains bad characters\n", pfx, idx);
+                return rv;
+            }
+
+            // Create the correct pin based on the type letter
+            // Unfortunately, we cannot determine whether a pin has a duplicate
+            // name when pin creation fails. The return value is not specific
+            // enough. We'd have to track all pin names ourselves if we wanted
+            // to merge pin substitution references into one pool.
+            switch (tchar) {
+            case 'b': case 'B':
+                slot->substtype[subst] = HAL_BIT;
+                if ((rv = hal_pin_bit_newf(HAL_IN, (hal_bit_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
+                    return rv;
+                break;
+            case 's': case 'S':
+                slot->substtype[subst] = HAL_S32;
+                if ((rv = hal_pin_s32_newf(HAL_IN, (hal_s32_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
+                    return rv;
+                break;
+            case 'u': case 'U':
+                slot->substtype[subst] = HAL_U32;
+                if ((rv = hal_pin_u32_newf(HAL_IN, (hal_u32_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
+                    return rv;
+                break;
+            case 'l': case 'L':
+                slot->substtype[subst] = HAL_S64;
+                if ((rv = hal_pin_s64_newf(HAL_IN, (hal_s64_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
+                    return rv;
+                break;
+            case 'k': case 'K':
+                slot->substtype[subst] = HAL_U64;
+                if ((rv = hal_pin_u64_newf(HAL_IN, (hal_u64_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
+                    return rv;
+                break;
+            case 'f': case 'F':
+                slot->substtype[subst] = HAL_FLOAT;
+                if ((rv = hal_pin_float_newf(HAL_IN, (hal_float_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
+                    return rv;
+                break;
+            default:
+                rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d bad substitution type '%c', expected one of {b,f,s,u,k,l}\n",
+                                pfx, idx, printable(tchar));
+                return -EBADMSG;
+            }
+
+            // Do not perform more substitutions than allowed
+            subst++;
+            if (subst >= MSG_SUBST_MAX)
+                break;
+
+            cptr = end;  // Move to end of substitution (loop will increment to next char)
+        }
+    }
+
+    // Finally set the first part's length for quick concat in print_slot()
+    slot->msglen = strlen(slot->msgbuf);
+    return 0;
+}
+
+EXTRA_SETUP()
+{
+    (void)__comp_inst;
+    (void)extra_arg;
+
+    // See how many messages are set
+    for (nslots = 0; nslots < MSG_N_MAX && msgs[nslots]; nslots++) {
+        // Just counting. Drops out at the first NULL (unset msgs)
+        rtapi_print_msg(RTAPI_MSG_DBG, "%s: msgs[%d]=%s\n", prefix, nslots, msgs[nslots]);
+    }
+
+    if (nslots < 1) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: No messages to handle\n", prefix);
+        return -EINVAL;
+    }
+
+    // Get memory for the HAL interface
+    pins = (msg_hal_t *)hal_malloc(nslots * sizeof(pins[0]));
+    if (!pins) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: No HAL memory for pins\n", prefix);
+        return -ENOMEM;
+    }
+    memset(pins, 0, nslots * sizeof(pins[0]));  // Ensure zeros because we check NULLs
+
+    // Get memory for the message slots
+    slots = (msg_slot_t *)rtapi_kzalloc(nslots * sizeof(slots[0]), RTAPI_GFP_KERNEL);
+    if (!slots) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: No kernel memory for slots\n", prefix);
+        return -ENOMEM;
+    }
+
+    // Interpret every message format and allocate pin resources
+    for (int i = 0; i < nslots; i++) {
+        int rv = setup_message(prefix, i, msgs[i], &slots[i], &pins[i]);
+        if (rv < 0) {
+            rtapi_kfree(slots);
+            slots = NULL;
+            return rv;
+        }
+    }
+    return 0;
+}
+
+EXTRA_CLEANUP()
+{
+    if (slots)
+        rtapi_kfree(slots);
+}
+
+static void print_slot(int s)
+{
+    char buf[MSG_LEN_MAX];
+    char *end = strcpy(buf, slots[s].msgbuf) + slots[s].msglen;
+    for (int i = 0; i < MSG_SUBST_MAX && pins[s].substs[i]; i++) {
+        int left = sizeof(buf) - (end - buf);
+        int n;
+        if (left <= 1)  // Need room for terminator
+            break;
+        switch (slots[s].substtype[i]) {
+        case HAL_BIT:
+            n = rtapi_snprintf(end, left, "%d%s", (int)!!pins[s].substs[i]->b, slots[s].parts[i]);
+            break;
+        case HAL_S32:
+            n = rtapi_snprintf(end, left, "%d%s", (int)pins[s].substs[i]->s, slots[s].parts[i]);
+            break;
+        case HAL_U32:
+            n = rtapi_snprintf(end, left, "%u%s", (unsigned)pins[s].substs[i]->u, slots[s].parts[i]);
+            break;
+        case HAL_S64:
+            n = rtapi_snprintf(end, left, "%ld%s", (long)pins[s].substs[i]->ls, slots[s].parts[i]);
+            break;
+        case HAL_U64:
+            n = rtapi_snprintf(end, left, "%lu%s", (unsigned long)pins[s].substs[i]->lu, slots[s].parts[i]);
+            break;
+        case HAL_FLOAT:
+            n = rtapi_snprintf(end, left, "%lf%s", (double)pins[s].substs[i]->f, slots[s].parts[i]);
+            break;
+        default:
+            n = 0;
+            break;
+        }
+        if (n < 0) {
+            rtapi_print_msg(RTAPI_MSG_ERR, "pushmsg: Printing slot %d failed rtapi_snprintf() call\n", s);
+            return;
+        }
+        end += n;
+    }
+    rtapi_print_msg(slots[s].msgtype, "%s\n", buf);
+}
+
+// Servo-thread cycle function
+FUNCTION(_)
+{
+    bool en = enable;  // Cache enable pin
+    // For each slot, test the trigger and save the trigger state
+    for (int i = 0; i < nslots; i++) {
+        bool trig = *(pins[i].trigger);
+        if (en && trig && !slots[i].prev) {
+            print_slot(i);  // Rising edge -> print message
+        }
+        slots[i].prev = trig;
+    }
+}
--- a/src/hal/components/pushmsg.comp
+++ b/src/hal/components/pushmsg.comp
@@ -3,18 +3,31 @@ component pushmsg "Push configurable RT messages to non-RT logging on rising edg
 description """
 Each entry in the `msgs` modparam describes a level, pin name, and message
 text using `level|pinname|message` syntax. Level is one letter:
-`e` (error), `w` (warning), `i` (info), `d` (debug). On the rising edge of
-`pushmsg.<pinname>` the corresponding text is emitted via `rtapi_print_msg()`
-at the chosen level.
+`e` (error), `w` (warning), `i` (info), `d` (debug). Each entry creates a
+HAL pin sub-tree `pushmsg.<pinname>.{trigger,force,edge}` and emits the
+corresponding text via `rtapi_print_msg()` at the chosen level.
+
+`pushmsg.<pinname>.trigger` (in bit): emits on transition to `edge` value.
+`pushmsg.<pinname>.force` (in bit): a FALSE->TRUE transition re-emits while
+trigger is at the active level.
+`pushmsg.<pinname>.edge` (in bit): selects edge polarity. The default value
+of FALSE makes a FALSE->TRUE trigger transition emit (rising-edge mode).
 
 Pin names use HAL-allowed characters (alphanumeric, dot, dash, underscore).
 Maximum 64 entries per loadrt, 2 substitutions per message, 128 chars per
 message.
 
 Message text may include `{t:refname}` placeholders. Each one creates a
-typed input pin `pushmsg.<refname>` that the user wires to the source
-signal. Type letter selects the pin type:
-`b` => bit, `s` => s32, `u` => u32, `l` => s64, `k` => u64, `f` => float.
+typed input pin `pushmsg.<pinname>.<refname>` that the user wires to the
+source signal. The type letter selects the pin type:
+
+* `b` - bit
+* `s` - s32
+* `u` - u32
+* `l` - s64
+* `k` - u64
+* `f` - float
+
 Use `\\{` to embed a literal curly brace.
 
 Example:
@@ -22,8 +35,8 @@ Example:
 ----
 loadrt pushmsg msgs="e|nooil|No oil pressure","w|low|Oil {f:level} L low"
 addf pushmsg servo-thread
-net no-oil classicladder.0.out-21 pushmsg.nooil
-net oil-level oil.level pushmsg.level
+net no-oil classicladder.0.out-21 pushmsg.nooil.trigger
+net oil-level oil.level pushmsg.low.level
 ----
 
 Example INI expansion:
@@ -71,12 +84,15 @@ RTAPI_MP_ARRAY_STRING(msgs, MSG_N_MAX, "Message slots in 't|pin|message' format"
 // The HAL structure lives in HAL memory space
 typedef struct {
     hal_bit_t  *trigger;
+    hal_bit_t  *force;
+    hal_bit_t  *edge;
     hal_data_u *substs[MSG_SUBST_MAX];
 } msg_hal_t;
 
 // The admin structure lives in normal memory space
 typedef struct {
-    bool        prev;                       // Previous pin state (to detect rising edge)
+    bool        prevtrig;                   // Previous trigger pin state (to detect rising edge)
+    bool        prevforce;                  // Previous force pin state (to detect rising edge)
     int         msgtype;                    // Rtapi message type (err, warn, info, debug)
     char        msgbuf[MSG_LEN_MAX];        // Complete message with embedded NULs for substitution
     int         msglen;                     // Length so we may find the end easily (stpcpy is not available)
@@ -158,22 +174,35 @@ static int setup_message(const char *pfx, int idx, const char *msg, msg_slot_t *
     }
     // Make sure we can encompass the full name
     int pinnamelen = pinend - msg - 2;
-    if (pinnamelen + pfxlen + 1 >= HAL_NAME_LEN) {  // +1 for the extra '.' character
+    if (pinnamelen + pfxlen + 9 >= HAL_NAME_LEN) {  // +9 for the two extra '.' characters and 'trigger' word (longest)
         rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d pin name too long\n", pfx, idx);
         return -EBADMSG;
     }
 
     int rv;
-    // Copy the pinname for later pin creation use
+    // Copy the pin name for later pin creation use.
+    // It already includes the prefix for our convenience.
     char pinname[HAL_NAME_LEN+1] = {};  // The init ensures termination
-    memcpy(pinname, &msg[2], pinnamelen);
-    if ((rv = check_pinname(pinname)) < 0) {
+    strcpy(pinname, pfx);  // "pushmsg"
+    pinname[pfxlen] = '.'; // "pushmsg."
+    memcpy(&pinname[pfxlen+1], &msg[2], pinnamelen);
+    // Only checks the actual pin name, not the prefix
+    if ((rv = check_pinname(&pinname[pfxlen+2])) < 0) {
         rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d pin name contains bad characters\n", pfx, idx);
         return rv;
     }
-    // Create the trigger input pin
-    if ((rv = hal_pin_bit_newf(HAL_IN, &pin->trigger, comp_id, "%s.%s", pfx, pinname)) < 0) {
-        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d cannot create pin (duplicate?)\n", pfx, idx);
+
+    // Create the trigger, force and edge input pins
+    if ((rv = hal_pin_bit_newf(HAL_IN, &pin->trigger, comp_id, "%s.trigger", pinname)) < 0) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d cannot create trigger pin (duplicate?)\n", pfx, idx);
+        return rv;
+    }
+    if ((rv = hal_pin_bit_newf(HAL_IN, &pin->force, comp_id, "%s.force", pinname)) < 0) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d cannot create force pin (duplicate?)\n", pfx, idx);
+        return rv;
+    }
+    if ((rv = hal_pin_bit_newf(HAL_IN, &pin->edge, comp_id, "%s.edge", pinname)) < 0) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d cannot create edge pin (duplicate?)\n", pfx, idx);
         return rv;
     }
 
@@ -213,7 +242,7 @@ static int setup_message(const char *pfx, int idx, const char *msg, msg_slot_t *
 
             char tchar = *cptr;
             cptr += 2;  // Move to pinname
-            if ((end - cptr) + pfxlen + 1 >= HAL_NAME_LEN) {  // +1 for the extra '.' character
+            if ((end - cptr) + pfxlen + pinnamelen + 2 >= HAL_NAME_LEN) {  // +2 for the two extra '.' characters
                 rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d substitution pin name too long\n", pfx, idx);
                 return -EBADMSG;
             }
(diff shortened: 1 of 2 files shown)


src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,73 @@
...
+RTAPI_MP_ARRAY_STRING(wmsg, MAX_MSG, "Warning messages")
+static char *imsg[MAX_MSG];
+RTAPI_MP_ARRAY_STRING(imsg, MAX_MSG, "Info messages")
+static char *dmsg[MAX_MSG];
+RTAPI_MP_ARRAY_STRING(dmsg, MAX_MSG, "Debug messages")
+
+static bool prev_e[MAX_MSG];
+static bool prev_w[MAX_MSG];
+static bool prev_i[MAX_MSG];
+static bool prev_d[MAX_MSG];
```
BsAtHome: This is wrong. You need to use variables because there can be multiple instances of components.

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,73 @@
...
+
+FUNCTION(_)
+{
+    int i;
+    for (i = 0; i < MAX_MSG; i++) {
```
BsAtHome: This loop was exactly what I wanted to avoid. You now have NxMx4 tests and assignments for every cycle.
Besides, the loop counter test is wrong because you declare the pins as a personality.

andypugh:
How does this compare with the existing "message" component? 

https://linuxcnc.org/docs/stable/html/man/man9/message.9.html

BsAtHome:
> How does this compare with the existing "message" component?
> 
> https://linuxcnc.org/docs/stable/html/man/man9/message.9.html

The message component only allows you to send error messages. You cannot use it for debug, warning or informational messages.

src/hal/components/pushmsg.comp
BsAtHome: Well, you made a singleton... I fell into the "wishful reading wrongly about my preconceptions" trap ;-)

grandixximo:
@ BsAtHome

While we are at it, what about,  .msg or similar file, INI style

```
[ERROR]
error-0 = the inverter had error <halpin.error.val>
[WARNING]
warning-0 = the oil pump is empty
```

and lift the 16 max limit?

Or is this too complex?

src/hal/components/pushmsg.comp
grandixximo: Changed to variables

BsAtHome:
Reading files in components is not necessary here. It should suffice to enter them as arguments in the loadrt.

The problem I have is the loop as such (I wrote exactly the same thing before deciding it was wrong). I think it is wasteful to loop over all <i>possible</i> triggers that are all ever so unlikely until you actually trigger. Similarly, having a single trigger and a setup pin to select the message makes it more difficult to select which message you want to emit.

Neither solution are enough KISS in my eyes. At least, I'm not convinced either solution is the right one.

src/hal/components/pushmsg.comp
grandixximo: revised, personality oriented, only loops on what user adds

grandixximo:
I don't think no loop is structurally possible if input is N HAL bit signals...

grandixximo:
> It should suffice to enter them as arguments in the loadrt.

I find that keeping track of which message is which on a single long string of loadrt is quite a bother, that is why I was proposing a file, where everything is forcibly formatted to be readable and properly indexed, knowing what message 7 of 15 is at a glance is convenient when you are debugging, rather than counting the commas in a long string.

BsAtHome:
That is a valid argument. However, you then need to parse a text file in the kernel...

grandixximo:
Fair, no kernel parsing needed, if what we want is just a better messaging tool for integrators. I was thinking of rescoping to a userspace daemon (loadusr): reads config file, creates HAL pins, polls at 20 Hz, emits via rtapi_print_msg (all 4 levels) plus NML for GUI ERR. RT untouched, file remains readable, <halpin.x> substitution available. What do you think?

BsAtHome:
> I was thinking of rescoping to a userspace daemon (loadusr): ...

Well, that would make this whole discussion redundant. The point of the original hack was to test rtapi_print_msg() propagation from <i>within</i> a component in <i>RT space</i>. Making a userspace daemon is kinda like a "problem solved, patient dead" solution.

I am looking for the actual use-case where you want several messages at several levels. The "old" component `message` is not used in any example. Just adding a new component 'because' is not very convincing.

grandixximo:
Fair, uspace message is another PR, not rescoping this one, `pushmsg` stays RT, do you want me to redraft it? make it similar to `message` component?

BsAtHome:
Well, answer my question: What is the use-case? There must be a clear reason why this is more than just a debug tool.

And then, why use a loop over all message types? Wouldn't it be better to use two args, one with strings and the other with message level? This complicates pins, but reduces cpu cycles. I have this cost/effort/benefit picture in my mind that does not want to fit.

grandixximo:
Use-case: integrator commissioning + field diagnostics. Real examples I hit:

- inverter fault bit => ERROR (operator stops, sees in GUI)
- coolant low / lube level => WARNING (pre-failure heads-up)
- spindle warmup incomplete / tool-change in progress => INFO (status, not error)
- state-machine transitions for ladder/classicladder debug => DEBUG (only when DEBUG set)

The `message` component covers only ERROR, so today integrators either:
- abuse error for warnings (operator fatigue, real errors lost in noise), or
- write a custom .comp per machine.

A 4-level RT push closes that gap. NML is GUI-only; rtapi_print_msg reaches dmesg/log too, which matters for headless/embedded retrofits.

On design: agreed, drop the per-level loop. Two parallel modparams, `messages` (strings) and `levels` (ints 0..3), single pin array `in-NN`, each pin emits its assigned message at its assigned level. One loop over N pins, edge-detect, fire. No 4x scan. Pins stay 1:1 with messages so wiring is unambiguous.

Bonus: HAL pin value substitution via `<pin.name>` or `<pin.name:%fmt>` syntax in the message string. Parsed once at `loadrt` (no kernel runtime string parsing): `halpr_find_pin_by_name` resolves to ptr + type, cached per slot. Fire time is just `rtapi_snprintf` with cached pointers. Constraint: referenced source comp must be loaded before pushmsg, standard HAL ordering.

Example:
  loadrt pushmsg messages="inverter fault <vfd.fault-code>","coolant low" levels=0,1
  net inverter-fault vfd.fault => pushmsg.0.in-00
  net coolant-low coolant.low => pushmsg.0.in-01

Want me to redraft `pushmsg.comp` on those lines?

andypugh:
Maybe it would be better to have a string of key-value pairs in the loadrt line? 
loadrt message w|oil|low_oil, e|nooil|no_oil at_all ...
ie, a key that determines warning/error etc and _also_ provides the pin name (needs uniqueness check at load time). 
Then there is no need to count commas and new messages can be insterted anywhere in the string. 
(In the above example I have also decided that we might need to replace spaces with underscores to keep the modparam system happy, but the system would replace with spaces for display)

andypugh:
I think that the "message" component is used quite often on real-world configs. Whether is should be used in a sample config is an interesting question.

BsAtHome:
And we can make it easier for people by using an INI expansion like:
```
loadrt pushmsg msgs=[PUSHMSG]MESSAGES
```

And then:
```ini
[PUSHMSG]
# Note: no surrounding whitespace
MESSAGES=\
"w|oil|Low oil",\
"e|nooil|No oil",\
"i|ping|Tool said hi"
```

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,223 @@
...
+option singleton yes;
+option extra_setup yes;
+option period no;
+
+pin in bit enable = TRUE "Master enable: when FALSE no messages are emitted";
+
+modparam dummy msgs """List of `level|pinname|message` entries.""";
+
+function _ "Edge-detect triggers, emit messages";
+license "GPL v2";
```
BsAtHome: Needs to state: `licence "GPL";` or it will not load into kernel.

You can add a comment that it is v2 (why not v2+ because you state "LinuxCNC" as author?).

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,223 @@
...
+        }
+        if (*bar2 != '|' || bar2 == pn) goto bad;
+        int pnlen = bar2 - pn;
+        if (pnlen > HAL_NAME_LEN - (int)sizeof("pushmsg.")) goto bad;
+        char pname[HAL_NAME_LEN + 1];
+        memcpy(pname, pn, pnlen); pname[pnlen] = 0;
+        hal_bit_t **trig = hal_malloc(sizeof(hal_bit_t *));
+        if (!trig) return -ENOMEM;
+        if (hal_pin_bit_newf(HAL_IN, trig, comp_id,
+                             "pushmsg.%s", pname) < 0) {
```
BsAtHome: This is very wrong. The new pin stores a pointer to the hal object created. You need no hal_malloc here.

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,223 @@
...
+
+typedef struct {
+    hal_bit_t **trigger;
+    hal_bit_t prev;
+    int level;
+    char literal[MAX_LIT];
+    int  litlen;
+    ref_t refs[MAX_REFS];
+    int  nrefs;
+} slot_t;
```
BsAtHome: There are two types of structures required. One where the hal reference lives (in hal memory) and one for internal admin can be simply allocated.

The hall pin in not a double pointer. It is a single pointer.
You _do not_ use hal-types to store the previous pin status. This is abuse of hal types and we've been battling this for a long time. You need the underlying type.

With these very basic mistakes, it looks like this code is LLM generated??

src/hal/hal_lib.c
```diff
@@ -4511,6 +4511,10 @@ EXPORT_SYMBOL(halpr_find_pin_by_sig);
 EXPORT_SYMBOL(hal_pin_alias);
 EXPORT_SYMBOL(hal_param_alias);
 
+EXPORT_SYMBOL(hal_get_pin_value_by_name);
+EXPORT_SYMBOL(hal_get_signal_value_by_name);
+EXPORT_SYMBOL(hal_get_param_value_by_name);
+
```
BsAtHome: You cannot use these function in running RT code. They use a lock and this cycling RT code is explicitly forbidden to use locks.

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,223 @@
...
+    hal_type_t type;
+    hal_data_u *data;
+    if (hal_get_pin_value_by_name(r->name, &type, &data, NULL) < 0)
+        return rtapi_snprintf(buf, bufsize, "<%s?>", r->name);
+    const char *fmt = r->fmt[0] ? r->fmt : (type == HAL_FLOAT ? "%g" : "%d");
+    switch (type) {
+        case HAL_BIT:   return rtapi_snprintf(buf, bufsize, fmt, (int)data->b);
+        case HAL_S32:   return rtapi_snprintf(buf, bufsize, fmt, (int)data->s);
+        case HAL_U32:   return rtapi_snprintf(buf, bufsize, fmt, (unsigned)data->u);
+        case HAL_FLOAT: return rtapi_snprintf(buf, bufsize, fmt, (double)data->f);
```
BsAtHome: Incomplete.

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,223 @@
...
+            "pushmsg: malformed entry '%s' (need L|pin|msg)\n", src);
+        return -EINVAL;
+    }
+    return 0;
+}
+
+static int format_ref(char *buf, int bufsize, ref_t *r) {
+    hal_type_t type;
+    hal_data_u *data;
+    if (hal_get_pin_value_by_name(r->name, &type, &data, NULL) < 0)
```
BsAtHome: You cannot call hal search functions. They use locking.

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,223 @@
...
+        s->nrefs++;
+        i = (end - src) + 1;
+    }
+    s->literal[s->litlen] = 0;
+    return 0;
+}
+
+EXTRA_SETUP() {
+    (void)__comp_inst; (void)prefix; (void)extra_arg;
+    for (int i = 0; i < MAX_MSGS && msgs[i]; i++) {
```
BsAtHome: It would be wise to determine how many pins you need to find out how much memory you need.

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,223 @@
...
+`pushmsg.<pinname>` the corresponding text is emitted via `rtapi_print_msg()`
+at the chosen level.
+
+Pin names must be unique and use HAL-allowed characters (alphanumeric, dot,
+dash, underscore). Maximum 64 entries per loadrt.
+
+Message text may contain `<sourcecomp.pin>` or `<sourcecomp.pin:%fmt>`
+references. Values are looked up at fire time, so source pins may be netted
+in any order. Default format from pin type if `%fmt` is omitted: bit/s32/u32
+use `%d`, float uses `%g`.
```
BsAtHome: You have no access to search pins.

What you could do is expose input pins that may be referenced and can be connected with signals.

[And you may want to use very basic {} formatting]
[And the types exposed are incomplete.]

BsAtHome (changes requested):

src/hal/components/pushmsg.comp
grandixximo: Changed to GPL, not good with licensing, don't want to plaster my name everywhere LinuxCNC is fine, and IANAL...

src/hal/components/pushmsg.comp
grandixximo: Was trying to make it automatic, dropped, user nets to exposed pins.

src/hal/hal_lib.c
grandixximo: Reverted, no longer needed

src/hal/components/pushmsg.comp
grandixximo: Changed, pre-count, single hal_malloc per array

src/hal/components/pushmsg.comp
grandixximo: Reworked, apologies for the rough first pass, not sure I fully understand why it is wrong :-(
I did have LLM help, but I don't understand why its wrong, I'll look at it again tomorrow after some sleep...

src/hal/components/pushmsg.comp
grandixximo: add 64 bit right?

src/hal/components/pushmsg.comp
grandixximo: revised to pre-counting

src/hal/components/pushmsg.comp
grandixximo: Spent the morning studying it, I think I get it now....

HAL pin storage and admin state in separate structs (paired by index). Mixing forces per-pin malloc instead of one alloc sized after pre-count, which should be preferred.
Internal state uses underlying C types (bool, int), not hal_bit_t.

I had the wrong mental model on the double pointer; thought it was a clever address cache. The legitimate `**` pattern (matrix_kb, classicladder, mux_generic) is pointer-to-array with one alloc and indexed access. Per-pin `**` is just single-pointer with extra steps.

Noticed several legacy components mix the two struct types (encoder, pid, pwmgen, stepgen, parport, debounce, watchdog) and mb2hal does the per-pin malloc, so I see the pattern you've been pushing back on. Not asking to clean those up, just confirming I get it?

src/hal/components/pushmsg.comp
grandixximo: Done, creating pins, and users will need to `net` them to get the values to the messages

BsAtHome:
I've been hacking a bit too (with my LBM<sup>†</sup> running on sugar and oxygen) and came up with this:

<sup>† <sub>Large Brain Model ;-)</sub></sup>
```C
component pushmsg "Push a message to non-RT";

pin in bit enable "Enable message generation";

description """
bla bla bla


"""
;
modparam dummy msgs "Comma separated list of custom messages in format t|pin|message";
option period no;
option extra_setup yes;
option extra_cleanup yes;
option singleton yes;
function _;
license "GPL";	// v2+
author "LinuxCNC";

include <rtapi_ctype.h>;
include <rtapi_gfp.h>;
include <rtapi_slab.h>;
;;
#define MSG_LEN_MAX	  128	// Max size of a message at input and expanded
#define MSG_N_MAX      64	// Max number of messages supported
#define MSG_SUBST_MAX   2	// Max number of substitutions within one message

// Command-line config parameter
const char *msgs[MSG_N_MAX] = {};
RTAPI_MP_ARRAY_STRING(msgs, MSG_N_MAX, "Message slots in 't|pin|message' format");

// The HAL structure lives in HAl memory space
typedef struct {
	hal_bit_t *trigger;
	hal_data_u *substs[MSG_SUBST_MAX];
} msg_hal_t;

// The admin structure lives in normal memory space
typedef struct {
	bool prev;      // Previous pin state (to detect rising edge)
	int  msgtype;   // Rtapi message type (err, warn, info, debug)
	char msgbuf[MSG_LEN_MAX]; // Complete message with embedded NULs for substitution
	int  msglen;    // Length so we may find the end easily (stpcpy is not available)
	hal_type_t substtype[MSG_SUBST_MAX]; // Substitution types (hal type enum)
	const char *parts[MSG_SUBST_MAX];    // Text parts index msgbuf for after substitution
} msg_slot_t;

//
// The message structure is decomposed from the source:
//   msgs[n] = "e|trigpin|Message {s:pin1} othertext {u:pin2} trailing"
//
// The source is copied into the slots->msgbuf and all pin refs are isolated:
//   msgbuf = "Message \0s:pin1\0 othertext \0u:pin2\0 trailing"
//                               ^                    ^
//                               parts[0]             parts[1]
// The slots->parts[] array is always the trailing part after the substitution.
// The pins for the substitution variable are referenced in the pins->substs[]
// array.
//

static msg_hal_t *pins;   // Trigger and substitution pins
static msg_slot_t *slots; // Admin data for each message
static int nslots;        // Number of active messages

static inline char printable(char c)
{
	return isprint(c & 0xff) ? c : '?';
}

static int message_type(char c)
{
	switch(c) {
	case 'e': case 'E': return RTAPI_MSG_ERR;
	case 'w': case 'W': return RTAPI_MSG_WARN;
	case 'i': case 'I': return RTAPI_MSG_INFO;
	case 'd': case 'D': return RTAPI_MSG_DBG;
	}
	return -1;
}

static int check_pinname(const char *name)
{
	for(;*name; name++) {
		if(!isascii(*name & 0xff) || (!isalnum(*name & 0xff) && NULL == strchr("._-", *name & 0xff)))
			return -EINVAL;
	}
	return 0;
}

static int setup_message(const char *pfx, int idx, const char *msg, msg_slot_t *slot, msg_hal_t *pin)
{
	int pfxlen = strlen(pfx);
	int l = strlen(msg);
	if(l >= MSG_LEN_MAX) {
		rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d too long, more than %d characters\n", pfx, idx, MSG_LEN_MAX-1);
		return -EMSGSIZE;
	}
	// Absolute minimum is "t|p|"
	if(l < 4) {
		rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d too short, less than 4 characters\n", pfx, idx);
		return -EMSGSIZE;
	}
	// Get the message type
	if((slot->msgtype = message_type(msg[0])) < 0) {
		rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d has invalid type '%c', expected one of {e,w,i,d}\n",
						pfx, idx, printable(msg[0]));
		return -EBADMSG;
	}
	// Message type must be followed by a '|'
	if('|' != msg[1]) {
		rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d missing first '|' separator\n", pfx, idx);
		return -EBADMSG;
	}
	// Find the second '|' in "t|name|message"
	const char *pinend = strchr(msg+2, '|');
	if(!pinend) {
		rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d missing pin name, no second '|' separator\n", pfx, idx);
		return -EBADMSG;
	}
	// Make sure we can encompass the full name
	int pinnamelen = pinend - msg - 2;
	if(pinnamelen + pfxlen + 1 >= HAL_NAME_LEN) {	// +1 for the extra '.' character
		rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d pin name too long\n", pfx, idx);
		return -EBADMSG;
	}

	int rv;
	// Copy the pinname for later pin creation use
	char pinname[HAL_NAME_LEN+1] = {};	// The init ensures termination
	memcpy(pinname, &msg[2], pinnamelen);
	if((rv = check_pinname(pinname)) < 0) {
		rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d pin name contains bad characters\n", pfx, idx);
		return rv;
	}
	// Create the trigger input pin
	if((rv = hal_pin_bit_newf(HAL_IN, &pin->trigger, comp_id, "%s.%s", pfx, pinname)) < 0) {
		rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d pin name too long\n", pfx, idx);
		return rv;
	}

	strcpy(slot->msgbuf, msg + pinnamelen + 3);	// Copy message so we may take it apart

	// Now see if there are any expansions in the string.
	int subst = 0;	// Count number of substitutions
	for(char *cptr = slot->msgbuf; *cptr; cptr++) {
		// Check "\{..." escaped curly open brace
		if('\\' == *cptr) {
			// The next char is escaped and cannot start a substitution
			if(cptr[1])	// Or this is a backslash at the end
				cptr++;
			continue;
		}
		if('{' == *cptr) {
			// Need "{t:pinname}"
			*cptr = 0;	// Terminate the previous string "foo {t:pin} bar" --> "foo "
			cptr++;		// Move to subst internals
			char *end = strchr(cptr, '}');
			if(!end) {
				rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d missing '}' in substitution\n", pfx, idx);
				return -EBADMSG;
			}
			slot->parts[subst] = end + 1;	// This is the continuation position "t:pin} bar" --> " bar"
			*end = 0;	// Terminate the pinname "t:pin} bar" --> "t:pin"

			if(end - cptr - 1 < 3) {	// Note: end points one char beyond (at a NUL)
				rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d substitution too small, need type and pinname\n", pfx, idx);
				return -EBADMSG;
			}
			if(':' != cptr[1]) {
				rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d substitution missing ':' after type character\n", pfx, idx);
				return -EBADMSG;
			}

			cptr += 2;	// Move to pinname
			// Note: end points to the trailing char
			if((end - cptr - 1) + pfxlen + 1 >= HAL_NAME_LEN) {	// +1 for the extra '.' character
				rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d substitution pin name too long\n", pfx, idx);
				return -EBADMSG;
			}
			if((rv = check_pinname(cptr+2)) < 0) {
				rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d substitution pin name contains bad characters\n", pfx, idx);
				return rv;
			}

			// Create the correct pin based on the type letter
			// Unfortunately, we cannot determine whether a pin has a duplicate
			// name when pin creation fails. The return value is not specific
			// enough. We'd have to track all pin names ourselves if we wanted
			// to merge pin substitution references into one pool.
			switch(*cptr) {
			case 'b': case 'B':
				slot->substtype[subst] = HAL_BIT;
				if((rv = hal_pin_bit_newf(HAL_IN, (hal_bit_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
					return rv;
				break;
			case 's': case 'S':
				slot->substtype[subst] = HAL_S32;
				if((rv = hal_pin_s32_newf(HAL_IN, (hal_s32_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
					return rv;
				break;
			case 'u': case 'U':
				slot->substtype[subst] = HAL_U32;
				if((rv = hal_pin_u32_newf(HAL_IN, (hal_u32_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
					return rv;
				break;
			case 'l': case 'L':
				slot->substtype[subst] = HAL_S64;
				if((rv = hal_pin_s64_newf(HAL_IN, (hal_s64_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
					return rv;
				break;
			case 'k': case 'K':
				slot->substtype[subst] = HAL_U64;
				if((rv = hal_pin_u64_newf(HAL_IN, (hal_u64_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
					return rv;
				break;
			case 'f': case 'F':
				slot->substtype[subst] = HAL_FLOAT;
				if((rv = hal_pin_float_newf(HAL_IN, (hal_float_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
					return rv;
				break;
			default:
				rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d bad substitution type, expected one of {b,f,s,u,k,l}\n", pfx, idx);
				return -EBADMSG;
			}

			// Do not perform more substitutions than allowed
			subst++;
			if(subst >= MSG_SUBST_MAX)
				break;

			cptr = end;	// Move to end of substitution (loop will increment to next char)
		}
	}

	// Finally set the first part's length for quick concat in print_slot()
	slot->msglen = strlen(slot->msgbuf);

	return 0;
}

EXTRA_SETUP()
{
	(void)__comp_inst;
	(void)extra_arg;

	// See how many messages are set
	for(nslots = 0; nslots < MSG_N_MAX && msgs[nslots]; nslots++) {
		// Just counting. Drops out at the first NULL (unset msgs)
		rtapi_print_msg(RTAPI_MSG_DBG, "%s: msgs[%d]=%s\n", prefix, nslots, msgs[nslots]);
	}

	if(nslots < 1) {
		rtapi_print_msg(RTAPI_MSG_ERR, "%s: No messages to handle\n", prefix);
		return -EINVAL;
	}

	// Get memory for the HAL interface
	pins = (msg_hal_t *)hal_malloc(nslots * sizeof(pins[0]));
	if(!pins) {
		rtapi_print_msg(RTAPI_MSG_ERR, "%s: No HAL memory for pins\n", prefix);
		return -ENOMEM;
	}
	memset(pins, 0, nslots * sizeof(pins[0]));	// Ensure zeros because we check NULLs

	// Get memory for the message slots
	slots = (msg_slot_t *)rtapi_kzalloc(nslots * sizeof(slots[0]), RTAPI_GFP_KERNEL);
	if(!slots) {
		rtapi_print_msg(RTAPI_MSG_ERR, "%s: No kernel memory for slots\n", prefix);
		return -ENOMEM;
	}

	// Interpret every message format and allocate pin resources
	for(int i = 0; i < nslots; i++) {
		int rv = setup_message(prefix, i, msgs[i], &slots[i], &pins[i]);
		if(rv < 0) {
			rtapi_kfree(slots);
			slots = NULL;
			return rv;
		}
	}
	return 0;
}

EXTRA_CLEANUP()
{
	if(slots)
		rtapi_kfree(slots);
}

static void print_slot(int s)
{
	char buf[MSG_LEN_MAX];
	char *end = strcpy(buf, slots[s].msgbuf) + slots[s].msglen;
	for(int i = 0; i < MSG_SUBST_MAX && pins[s].substs[i]; i++) {
		int left = sizeof(buf) - (buf - end);
		int n;
		if(left <= 1)	// Need room for terminator
			break;
		switch(slots[s].substtype[i]) {
		case HAL_BIT:
			n = rtapi_snprintf(end, left, "%d%s", (int)!!pins[s].substs[i]->b, slots[s].parts[i]);
			break;
		case HAL_S32:
			n = rtapi_snprintf(end, left, "%d%s", (int)pins[s].substs[i]->s, slots[s].parts[i]);
			break;
		case HAL_U32:
			n = rtapi_snprintf(end, left, "%u%s", (unsigned)pins[s].substs[i]->u, slots[s].parts[i]);
			break;
		case HAL_S64:
			n = rtapi_snprintf(end, left, "%ld%s", (long)pins[s].substs[i]->ls, slots[s].parts[i]);
			break;
		case HAL_U64:
			n = rtapi_snprintf(end, left, "%lu%s", (unsigned long)pins[s].substs[i]->lu, slots[s].parts[i]);
			break;
		case HAL_FLOAT:
			n = rtapi_snprintf(end, left, "%lf%s", (double)pins[s].substs[i]->f, slots[s].parts[i]);
			break;
		default:
			n = 0;
			break;
		}
		if(n < 0) {
			rtapi_print_msg(RTAPI_MSG_ERR, "pushmsg: Printing slot %d failed rtapi_snprintf() call\n", s);
			return;
		}
		end += n;
	}
	rtapi_print_msg(slots[s].msgtype, "%s\n", buf);
}

// Servo-thread cycle function
FUNCTION(_)
{
	bool en = enable;	// Cache enable pin
	// For each slot, test the trigger and save the trigger state
	for(int i = 0; i < nslots; i++) {
		bool trig = *(pins[i].trigger);
		if(en && trig && !slots[i].prev) {
			print_slot(i);	// Rising edge -> print message
		}
		slots[i].prev = trig;
	}
}
// vim: ts=4 sw=4 syn=c
```

src/hal/components/pushmsg.comp
BsAtHome: yes.

src/hal/components/pushmsg.comp
BsAtHome: Hal pins in a component are a pointer to a in-HAL-memory data unit. When you create a pin, then you must store the reference to that internal data unit. The reference is a pointer. To store a reference, you need the address of the reference, hence the double pointer.

Yes, there is a lot of confusion about types and memory usage. It needs to be cleaned up, but that is another issue.

grandixximo:
Your LBM beats my LLM 😅

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,397 @@
...
+} msg_slot_t;
+
+//
+// Source format:
+//   msgs[n] = "e|trigpin|Message {s:pin1} othertext {u:pin2} trailing"
+//
+// In-place decomposition into msgbuf with embedded NULs:
+//   msgbuf = "Message \0s:pin1\0 othertext \0u:pin2\0 trailing"
+//                                ^                   ^
+//                                parts[0]            parts[1]
```
BsAtHome: parts[0] is offset one to the right. It should point at the space after the NUL.

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,397 @@
...
+} msg_hal_t;
+
+// Admin structure: parsed state, lives in normal memory.
+typedef struct {
+    bool        prev;
+    int         msgtype;
+    char        msgbuf[MSG_LEN_MAX];
+    int         msglen;
+    hal_type_t  substtype[MSG_SUBST_MAX];
+    const char *parts[MSG_SUBST_MAX];
```
BsAtHome: Where did the comments go?

A lot more comments are just gone... Why did you strip all the comments in the code? That is a Bad Move<sup>TM</sup>.

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,397 @@
...
+                    (unsigned)pins[s].substs[i]->u, slots[s].parts[i]);
+                break;
+            case HAL_S64:
+                n = rtapi_snprintf(end, left, "%lld%s",
+                    (long long)pins[s].substs[i]->ls, slots[s].parts[i]);
+                break;
+            case HAL_U64:
+                n = rtapi_snprintf(end, left, "%llu%s",
+                    (unsigned long long)pins[s].substs[i]->lu,
+                    slots[s].parts[i]);
```
BsAtHome: Are you sure the kernel's version of rtapi_snprint supports %lld and %llu? By my glance it only supports %ld and %lu.

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,397 @@
...
+    for (; *name; name++) {
+        if (!isascii(*name & 0xff)
+            || (!isalnum(*name & 0xff) && NULL == strchr("._-", *name & 0xff)))
+            return -EINVAL;
+    }
+    return 0;
+}
+
+static int setup_message(const char *pfx, int idx, const char *msg,
+                         msg_slot_t *slot, msg_hal_t *pin)
```
BsAtHome: What is it with all the line breaks? Everywhere the code is smeared across lines.

We're living in the age where terminals have been 132 characters wide for more than 40 years...

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,397 @@
...
+            char *end = strchr(cptr, '}');
+            if (!end) {
+                rtapi_print_msg(RTAPI_MSG_ERR,
+                    "%s: Message %d missing '}' in substitution\n", pfx, idx);
+                return -EBADMSG;
+            }
+            slot->parts[subst] = end + 1;
+            *end = 0;
+
+            if (end - cptr < 3) {
```
BsAtHome, line 255: Why was this changed? If I'm not mistaking, end points beyond making end-cptr to be at least one (when no chars are in the string).

cptr points to the first character after '{' and end must therefore be larger to be able to find a '}'.

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,397 @@
...
+
+Pin names use HAL-allowed characters (alphanumeric, dot, dash, underscore).
+Maximum 64 entries per loadrt, 2 substitutions per message, 128 chars per
+message.
+
+Message text may include `{t:refname}` placeholders. Each one creates a
+typed input pin `pushmsg.<refname>` that the user wires to the source
+signal. Type letter selects the pin type:
+`b` => bit, `s` => s32, `u` => u32, `l` => s64, `k` => u64, `f` => float.
+Use `\\{` to embed a literal curly brace.
```
BsAtHome: I'm not sure you need double backslash to escape the backslash. Please check the adoc and the resulting troff file.

BsAtHome (changes requested):

src/hal/components/pushmsg.comp
grandixximo: For `{s:a}` after the setup code:
- cptr at 's' (positions: s=0, ':'=1, 'a'=2, '}'=3 from cptr)
- end = strchr(cptr, '}') points at '}' => end - cptr = 3
- Original `end - cptr - 1 < 3` evaluates to `2 < 3` => rejects valid `{s:a}`
- My `end - cptr < 3` evaluates to `3 < 3` => accepts

Min valid is 3 chars between braces: type + ':' + 1-char name.

For `{}`: cptr at '}' (after cptr++), strchr finds it at cptr,
end - cptr = 0 (not "at least one"). Original would also reject this
(`-1 < 3` true), so behaviour matches there. The divergence is at the
3-char boundary.

Confirmed empirically: pre-fix, `{s:a}` returned EBADMSG; post-fix it
loads and the pin is created.

BsAtHome: Check! I'll eat my brain ;-)

BsAtHome:
BTW, the mighty off-by-one gift keeps on giving.

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,383 @@
...
+            if (':' != cptr[1]) {
+                rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d substitution missing ':' after type character\n", pfx, idx);
+                return -EBADMSG;
+            }
+
+            char tchar = *cptr;
+            cptr += 2;  // Move to pinname
+            if ((end - cptr) + pfxlen + 1 >= HAL_NAME_LEN) {  // +1 for the extra '.' character
+                rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d substitution pin name too long\n", pfx, idx);
+                return -EBADMSG;
```
BsAtHome, line 268: Needs to be rewritten. 

The cptr += 2 was a late addition that is wrongly placed there because the following switch would address *cptr as the type char. That you changed, but the length calculations are now probably off.

The best way to handle this is to move the cptr+=2 down just before the switch and adapt the length calculation.

BsAtHome: Never mind... My mind checked out.

src/hal/components/pushmsg.comp
BsAtHome: BTW, maybe the type list should be a bullet list like in:
``` adoc
* 'b' - bit
* 's' - s32
* 'u' - u32
* ...
```

andypugh:
Should we add a deprecation message to the documentation for the "message" component?

BsAtHome:
> Should we add a deprecation message to the documentation for the "message" component?

That may be appropriate.

I think we need to have a look at some of the extra pins and trigger modes provided in the "message" component. I don't think they need to be forward ported, but it doesn't hurt to look another time.

BsAtHome:
Just had a quick look at message.

If we want to add trigger-edge selection and force-trigger, then we need to change the pin names to something like:
```
"e|pinname|message" gives:
pushmsg.enable (bit)
pushmsg.pinname.trigger (bit)
pushmsg.pinname.force (bit)
pushmsg.pinname.edge (bit)
```

If we decide that it would be appropriate to add this, then it may also be appropriate to add the substitution pins in the sub-tree like:
```
"e|pinname|message says {b:bitly} is {f:level}" gives:
pushmsg.enable (bit)
pushmsg.pinname.trigger (bit)
pushmsg.pinname.force (bit)
pushmsg.pinname.edge (bit)
pushmsg.pinname.bitly (bit)
pushmsg.pinname.level (float)
```

src/hal/components/pushmsg.comp
grandixximo: Verified: source `\\{` => man renders as `\e{` => displayed as `\{` (correct)

src/hal/components/pushmsg.comp
grandixximo: kept inline

src/hal/components/pushmsg.comp
grandixximo: you are right, rolled back rtapi_snprint does not support

src/hal/components/pushmsg.comp
grandixximo: Restored comments

src/hal/components/pushmsg.comp
grandixximo: offset one to the right pointing at space after `\0`

BsAtHome:
This diff (wrt current PR) should enable, for each pin, edge selection, force generation and isolate all pins into a pin name hierarchy. Please test it.
```diff
--- hal/components/pushmsg.comp.orig	2026-05-06 15:16:05.145630367 +0200
+++ hal/components/pushmsg.comp	2026-05-06 15:15:07.194801100 +0200
@@ -71,12 +71,15 @@
 // The HAL structure lives in HAL memory space
 typedef struct {
     hal_bit_t  *trigger;
+    hal_bit_t  *force;
+    hal_bit_t  *edge;
     hal_data_u *substs[MSG_SUBST_MAX];
 } msg_hal_t;
 
 // The admin structure lives in normal memory space
 typedef struct {
-    bool        prev;                       // Previous pin state (to detect rising edge)
+    bool        prevtrig;                   // Previous trigger pin state (to detect rising edge)
+    bool        prevforce;                  // Previous force pin state (to detect rising edge)
     int         msgtype;                    // Rtapi message type (err, warn, info, debug)
     char        msgbuf[MSG_LEN_MAX];        // Complete message with embedded NULs for substitution
     int         msglen;                     // Length so we may find the end easily (stpcpy is not available)
@@ -158,22 +161,35 @@
     }
     // Make sure we can encompass the full name
     int pinnamelen = pinend - msg - 2;
-    if (pinnamelen + pfxlen + 1 >= HAL_NAME_LEN) {  // +1 for the extra '.' character
+    if (pinnamelen + pfxlen + 9 >= HAL_NAME_LEN) {  // +9 for the two extra '.' characters and 'trigger' word (longest)
         rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d pin name too long\n", pfx, idx);
         return -EBADMSG;
     }
 
     int rv;
-    // Copy the pinname for later pin creation use
+    // Copy the pin name for later pin creation use.
+    // It already includes the prefix for our convenience.
     char pinname[HAL_NAME_LEN+1] = {};  // The init ensures termination
-    memcpy(pinname, &msg[2], pinnamelen);
-    if ((rv = check_pinname(pinname)) < 0) {
+    strcpy(pinname, pfx);  // "pushmsg"
+    pinname[pfxlen] = '.'; // "pushmsg."
+    memcpy(&pinname[pfxlen+1], &msg[2], pinnamelen);
+    // Only checks the actual pin name, not the prefix
+    if ((rv = check_pinname(&pinname[pfxlen+2])) < 0) {
         rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d pin name contains bad characters\n", pfx, idx);
         return rv;
     }
-    // Create the trigger input pin
-    if ((rv = hal_pin_bit_newf(HAL_IN, &pin->trigger, comp_id, "%s.%s", pfx, pinname)) < 0) {
-        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d cannot create pin (duplicate?)\n", pfx, idx);
+
+    // Create the trigger, force and edge input pins
+    if ((rv = hal_pin_bit_newf(HAL_IN, &pin->trigger, comp_id, "%s.trigger", pinname)) < 0) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d cannot create trigger pin (duplicate?)\n", pfx, idx);
+        return rv;
+    }
+    if ((rv = hal_pin_bit_newf(HAL_IN, &pin->force, comp_id, "%s.force", pinname)) < 0) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d cannot create force pin (duplicate?)\n", pfx, idx);
+        return rv;
+    }
+    if ((rv = hal_pin_bit_newf(HAL_IN, &pin->edge, comp_id, "%s.edge", pinname)) < 0) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d cannot create edge pin (duplicate?)\n", pfx, idx);
         return rv;
     }
 
@@ -213,7 +229,7 @@
 
             char tchar = *cptr;
             cptr += 2;  // Move to pinname
-            if ((end - cptr) + pfxlen + 1 >= HAL_NAME_LEN) {  // +1 for the extra '.' character
+            if ((end - cptr) + pfxlen + pinnamelen + 2 >= HAL_NAME_LEN) {  // +2 for the two extra '.' characters
                 rtapi_print_msg(RTAPI_MSG_ERR, "%s: Message %d substitution pin name too long\n", pfx, idx);
                 return -EBADMSG;
             }
@@ -230,32 +246,32 @@
             switch (tchar) {
             case 'b': case 'B':
                 slot->substtype[subst] = HAL_BIT;
-                if ((rv = hal_pin_bit_newf(HAL_IN, (hal_bit_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
+                if ((rv = hal_pin_bit_newf(HAL_IN, (hal_bit_t **)&pin->substs[subst], comp_id, "%s.%s", pinname, cptr)) < 0)
                     return rv;
                 break;
             case 's': case 'S':
                 slot->substtype[subst] = HAL_S32;
-                if ((rv = hal_pin_s32_newf(HAL_IN, (hal_s32_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
+                if ((rv = hal_pin_s32_newf(HAL_IN, (hal_s32_t **)&pin->substs[subst], comp_id, "%s.%s", pinname, cptr)) < 0)
                     return rv;
                 break;
             case 'u': case 'U':
                 slot->substtype[subst] = HAL_U32;
-                if ((rv = hal_pin_u32_newf(HAL_IN, (hal_u32_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
+                if ((rv = hal_pin_u32_newf(HAL_IN, (hal_u32_t **)&pin->substs[subst], comp_id, "%s.%s", pinname, cptr)) < 0)
                     return rv;
                 break;
             case 'l': case 'L':
                 slot->substtype[subst] = HAL_S64;
-                if ((rv = hal_pin_s64_newf(HAL_IN, (hal_s64_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
+                if ((rv = hal_pin_s64_newf(HAL_IN, (hal_s64_t **)&pin->substs[subst], comp_id, "%s.%s", pinname, cptr)) < 0)
                     return rv;
                 break;
             case 'k': case 'K':
                 slot->substtype[subst] = HAL_U64;
-                if ((rv = hal_pin_u64_newf(HAL_IN, (hal_u64_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
+                if ((rv = hal_pin_u64_newf(HAL_IN, (hal_u64_t **)&pin->substs[subst], comp_id, "%s.%s", pinname, cptr)) < 0)
                     return rv;
                 break;
             case 'f': case 'F':
                 slot->substtype[subst] = HAL_FLOAT;
-                if ((rv = hal_pin_float_newf(HAL_IN, (hal_float_t **)&pin->substs[subst], comp_id, "%s.%s", pfx, cptr)) < 0)
+                if ((rv = hal_pin_float_newf(HAL_IN, (hal_float_t **)&pin->substs[subst], comp_id, "%s.%s", pinname, cptr)) < 0)
                     return rv;
                 break;
             default:
@@ -374,10 +390,12 @@
     bool en = enable;  // Cache enable pin
     // For each slot, test the trigger and save the trigger state
     for (int i = 0; i < nslots; i++) {
-        bool trig = *(pins[i].trigger);
-        if (en && trig && !slots[i].prev) {
+        bool trig = (int)!!*(pins[i].trigger) ^ (int)!!*(pins[i].edge);
+        bool force = *(pins[i].force);
+        if (en && ((trig && !slots[i].prevtrig) || (force && !slots[i].prevforce))) {
             print_slot(i);  // Rising edge -> print message
         }
-        slots[i].prev = trig;
+        slots[i].prevtrig  = trig;
+        slots[i].prevforce = force;
     }
 }
```

src/hal/components/pushmsg.comp
grandixximo: Also did the bullets, and checked they render proper

grandixximo:
Logic test passes:

Test 1 (rising edge=0, 0→1): fires val=1 ✓
Test 2 (trigger held high): no fire ✓
Test 3 (force re-emit): fires val=3 ✓
Test 4 (1→0 with edge=0): no fire ✓
Test 5 (edge flip mid-state): no fire (XOR semantics) ✓
Test 6 (edge=1 falling, 1→0): fires val=6 ✓

grandixximo:
> Should we add a deprecation message to the documentation for the "message" component?

Deprecation message pointing at pushmsg added

BsAtHome (approved):

BsAtHome:
Question: now that there is that "force" pin, does it make sense that it is gated by the enable pin?

It would probably make more sense to change the logic whereby "force" actually means what it says and will print the message regardless:
```C
if ((en && trig && !slots[i].prevtrig) || (force && !slots[i].prevforce)) {
    print_slot(i);  // Rising edge -> print message
}
```
If so, then the docs may be changed a bit to reflect this too.

grandixximo:
Good morning Bertho, agreed force can be more useful if it is completely unconditional. Changed to PR, ready to merge

andypugh:
Good to go?

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,415 @@
...
+
+    int rv;
+    // Copy the pin name for later pin creation use.
+    // It already includes the prefix for our convenience.
+    char pinname[HAL_NAME_LEN+1] = {};  // The init ensures termination
+    strcpy(pinname, pfx);  // "pushmsg"
+    pinname[pfxlen] = '.'; // "pushmsg."
+    memcpy(&pinname[pfxlen+1], &msg[2], pinnamelen);
+    // Only checks the actual pin name, not the prefix
+    if ((rv = check_pinname(&pinname[pfxlen+2])) < 0) {
```
BsAtHome: I think this should be `check_pinname(&pinname[pfxlen+1])` because that is where the pinname was copied to in the memcpy above.

src/hal/components/pushmsg.comp
```diff
@@ -0,0 +1,415 @@
...
+`pushmsg.<pinname>.trigger` (in bit): emits on transition to `edge` value.
+`pushmsg.<pinname>.force` (in bit): a FALSE->TRUE transition emits the
+message unconditionally, ignoring both the master `pushmsg.enable` gate
+and the current `trigger` state.
+`pushmsg.<pinname>.edge` (in bit): selects edge polarity. The default value
+of FALSE makes a FALSE->TRUE trigger transition emit (rising-edge mode).
+
+Pin names use HAL-allowed characters (alphanumeric, dot, dash, underscore).
+Maximum 64 entries per loadrt, 2 substitutions per message, 128 chars per
+message.
```
BsAtHome: There needs to be a bit work on this explanation text. Will post a diff.

BsAtHome:
docu update diff:
```diff
--- hal/components/pushmsg.comp.orig	2026-05-07 14:58:12.704075606 +0200
+++ hal/components/pushmsg.comp	2026-05-07 16:08:56.013067921 +0200
@@ -1,63 +1,83 @@
-component pushmsg "Push configurable RT messages to non-RT logging on rising edges";
+component pushmsg "Push configurable RT messages to non-RT logging on pin change";
 
 description """
-Each entry in the `msgs` modparam describes a level, pin name, and message
-text using `level|pinname|message` syntax. Level is one letter:
-`e` (error), `w` (warning), `i` (info), `d` (debug). Each entry creates a
-HAL pin sub-tree `pushmsg.<pinname>.{trigger,force,edge}` and emits the
-corresponding text via `rtapi_print_msg()` at the chosen level.
-
-`pushmsg.<pinname>.trigger` (in bit): emits on transition to `edge` value.
-`pushmsg.<pinname>.force` (in bit): a FALSE->TRUE transition emits the
-message unconditionally, ignoring both the master `pushmsg.enable` gate
-and the current `trigger` state.
-`pushmsg.<pinname>.edge` (in bit): selects edge polarity. The default value
-of FALSE makes a FALSE->TRUE trigger transition emit (rising-edge mode).
-
-Pin names use HAL-allowed characters (alphanumeric, dot, dash, underscore).
-Maximum 64 entries per loadrt, 2 substitutions per message, 128 chars per
-message.
-
-Message text may include `{t:refname}` placeholders. Each one creates a
-typed input pin `pushmsg.<pinname>.<refname>` that the user wires to the
-source signal. The type letter selects the pin type:
-
-* `b` - bit
-* `s` - s32
-* `u` - u32
-* `l` - s64
-* `k` - u64
-* `f` - float
+The *pushmsg* component enables you to generate configurable messages to your
+GUI, terminal or other interface from real-time components using HAL pins. 
 
-Use `\\{` to embed a literal curly brace.
+Up to 64 messages can be configured using the *msgs* module parameter using the
+following syntax: +
+  `"level|pinname|message"`[,...]
+
+`level`:: is a letter denoting the severity level of the message with: +
+  * `e` - error
+  * `w` - warning
+  * `i` - info
+  * `d` - debug
+
+`pinname`:: is the HAL pin name base used in the creation of the trigger pins:
+  * *pushmsg*.__<pinname>__.*trigger* (bit) - The trigger input to activate
+    the message
+  * *pushmsg*.__<pinname>__.*edge* (bit) - The trigger input edge selection
+    with FALSE=rising trigger edge and TRUE=falling trigger edge
+  * *pushmsg*.__<pinname>__.*force* (bit) - An always active rising edge
+    trigger to activate the message, regardless of the setting of the global
+    enable pin (*pushmsg*.*enable*)
+ 
+`message`:: is a free form text message. +
+  The text may include `{t:refname}` placeholders for HAL pin value
+  substitutions (use `\\{` to embed a literal curly brace).
+  Up to two placeholders are supported.
+  Each placeholder creates a typed input pin *pushmsg*.__<pinname>__.__<refname>__
+  that the user wires up to the source signal.
+  The type letter in the placeholder selects the pin type:
+  * `b` - bit
+  * `s` - s32
+  * `u` - u32
+  * `l` - s64
+  * `k` - u64
+  * `f` - float
+
+Each message line is limited to 128 characters in length. The length count
+includes the `level` indicator, `pinname`, `message` text and the `|` separator
+characters.
 
-Example:
+HAL file example:
 [source,hal]
 ----
 loadrt pushmsg msgs="e|nooil|No oil pressure","w|low|Oil {f:level} L low"
 addf pushmsg servo-thread
 net no-oil classicladder.0.out-21 pushmsg.nooil.trigger
 net oil-level oil.level pushmsg.low.level
+...
 ----
 
+The INI-file may be an administrative help for message configuration.
+A longer list is much easier to read and maintain when split over multiple
+lines. The INI-file format supports continuations that improve readability and
+maintainability.
+
 Example INI expansion:
 [source,ini]
 ----
 [PUSHMSG]
-# Note: no surrounding whitespace
+# Note: no surrounding whitespace, last line has no comma
 MSGS=\\
 "e|nooil|No oil pressure",\\
 "w|low|Oil {f:level} L low"
 ----
+
+The HAL file then looks like:
 [source,hal]
 ----
 loadrt pushmsg msgs=[PUSHMSG]MSGS
+addf pushmsg servo-thread
+...
 ----
 """;
 
-pin in bit enable = TRUE "Enable message generation";
+pin in bit enable = TRUE "Enable message generation through trigger input pins";
 
-modparam dummy msgs "Comma separated list of custom messages in format t|pin|message";
+modparam dummy msgs "Comma separated list of custom messages in format level|pinname|message";
 
 option period no;
 option extra_setup yes;
```

BsAtHome:
> Good to go?

Almost... One off-by-one issue and some documentation changes for both better explanation and formatting.

grandixximo:
@andypugh ready

PR #4004, opened 2026-05-04
rcs: Remove arbitrarily sized static char buffers in print routines.

BsAtHome:
Remove the static char buffers from rcs_print() and friends and use properly sized std::vector storage instead. This is an adaptation of #1922 and replaces that PR completely.

### DIFF
--- a/src/libnml/rcs/rcs_print.cc
+++ b/src/libnml/rcs/rcs_print.cc
@@ -17,6 +17,7 @@
 #include <string.h>		/* strchr(), memmove() */
 #include <stdlib.h>		/* malloc(), free(), realloc() */
 #include <errno.h>		// errno()
+#include <vector>
 
 #include <sys/types.h>
 #include <unistd.h>		/* getpid() */
@@ -41,7 +42,6 @@ FILE *rcs_print_file_stream = NULL;
 char rcs_print_file_name[80] = "rcs_out.txt";
 
 char last_error_bufs[4][error_buf_size];
-int error_bufs_initialized = 0;
 int last_error_buf_filled = 0;
 
 void set_rcs_print_destination(RCS_PRINT_DESTINATION_TYPE _dest)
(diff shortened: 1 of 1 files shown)


rmu75:
I still find the use of std::vector instead of std::string a bit strange in this context. std::string default ctor doesn't allocate (small string optimization), reserve() would allocate and avoid the memset.

Whatever. This and other printf-based function should probably be phased out anyways and be replaced with something based on fmtlib, at least for new c++ code.

BsAtHome:
Sure, you can use string::resize, but using vector allows you to specify the reserve size at creation. Using string you must create the string and then resize (creating string with size enforces an init). And, yes, small strings (I think < 16) have no allocation. But that is not often the case in these printing functions. All of this in non-RT code anyway and when the need to print something, well, then timing is often off already.

Also, you are right, it should all be replaced using libfmt. Heck, the whole printing/message strategy should be revamped, channelized and configurable. And then there are the printf/fprintf calls that are messing things up in other parts of the code...

FWIW, I changed rcs because it was hooked into a very old PR (#1922). I wanted to get rid of that one but rebasing was out of the question. This fix was the result. Adding one PR for closing two :-)

rmu75:
On So, 2026-05-10 at 03:23 -0700, BsAtHome wrote:
> BsAtHome left a comment (LinuxCNC/linuxcnc#4004)
> Sure, you can use string::resize, but using vector allows you to
> specify the reserve size at creation. Using string you must create
> the string and then resize (creating string with size enforces an
> init).

c++20 std::string also has ctor with size and character to initialize.

c++23 will have resize_and_overwrite, that would avoid initialization,
but we can't use that yet.

BsAtHome:
Indeed.
Moving to C++23 may be quite some time away because we need to support older distros for quite some time. By the time we can make the switch we should already have rewritten this rcs code ;-)
(just wishful thinking... it is allowed)

Issue #3668, opened 2025-12-15
joint.N.amp-fault-in does not unhome joints set to VOLATILE_HOME = 1


Sigma1912:
As reported on the forum:

https://forum.linuxcnc.org/49-basic-configuration/58055-volatile-home-not-unhoming-on-fault#340101

I can confirm this testing on current master.

1. Add 'VOLATILE_HOME = 1' to  ini section [JOINT_0] of 'axis.ini'
2. Start the sim config and home the machine.
3. Open 'halshow' tool and manually set 'joint.0.amp-fault-in' to True 

Machine state is switched to OFF but joint.0 remains homed.

The joint unhomes correctly if an Estop occurs or the machine state is switched to off using motion or halui halpins.

Documentation states:

> VOLATILE_HOME = 0 - When enabled (set to 1) this joint will be unhomed if the Machine Power is off or if E-Stop is on. This is useful if your machine has home switches and does not have position feedback such as a step and direction driven machine. 
https://linuxcnc.org/docs/html/config/ini-config.html

I guess one could work around this by connecting 'joint.0.amp-fault-in' to 'halui.machine.off' but it does seem like a bug to me since amp-fault-in automatically switches machine state to off.

Sigma1912:
Possible place to fix this:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/motion/control.c#L873

Sigma1912:
> VOLATILE_HOME is intended to unhome joints if encoder tracking is lost. I am not sure that amp-fault necessarily implies tracking lost.

I always thought this was used on joints that cannot hold position whenever the drive is disabled (regardless of encoder feedback). Like on a heavy Z Axis.  

Ether way, neither does switching the machine off or even an Estop condition imply a loss of encoder position. 

Problem is that the current behavior is VERY inconsistent:
- Estop unhomes volatile joints
-  Switching the machine state to off using the gui button or halui.machine.off unhomes volatile joints
-  amp-fault switches the machine state to off (visible in the gui and reflected by halui.machine.is-on going False) but volatile joints are not unhomed. Machine motion appears to be halted without any ramping just as in an Estop event.

This does not seem to make much sense.

Maybe we should just remove the built in behavior of amp-fault and leave it up to the user to connect this in hal.

snowgoer540:
> * amp-fault switches the machine state to off (visible in the gui and reflected by halui.machine.is-on going False)

[...]

PR #4008, opened 2026-05-05
task: unhome VOLATILE_HOME joints on any motion-disable edge

### LINKED ISSUE #3954: With `VOLATILE_HOME = 1`, triggering an end-stop disables the machine but keeps the axis homed
I have a machine with MESA 7i76EU for IO, traditional stepper motors, no encoders, nothing fancy. Since the axis are relatively easy to move if the steppers are not powered, I enabled `VOLATILE_HOME = 1`. This generally works, i.e. if I home the X-axis, and then disable the machine (via [F2] in the GUI), the homed icon in the readout goes away. But, if I run an unhomed axis (let's say the Y-axis) into an endstop, the machine is disabled as well (and I have to press [F2] to re-enable it), but, all other axis maintain their homed status (despite their amps/stepper drivers also being disabled).

From my point of view this is a bug; if accurate absolute permission is (potentially...) lost the moment the steppers are disabled, than with `VOLATILE_HOME = 1`, the homed status should vanish the moment the steppers are disabled.

grandixximo:
## Summary
- Fixes #3954. With VOLATILE_HOME=1, tripping a limit switch (or amp-fault, ferror, misc-error, spindle-fault) disables the drives but leaves all joints flagged as homed. Only ESTOP and F2/machine-off paths cleared the homed flag.
- Adds a 1->0 edge detector on `emcStatus->motion.traj.enabled` in the task main loop. On the edge it calls `emcJointUnhome(-2)`, which unhomes only joints marked VOLATILE_HOME. Estop and machine-off continue to do their own unhome and remain idempotent.

## Why this scope
- VOLATILE_HOME is opt-in. A user setting it has declared the joint cannot hold position when its drive is off, so the discriminating event is "drive disabled", not "task state went OFF".
- Three reports over five years (#3954, #3668, plus a 2021 forum thread linked in #3954) all hit the same gap. Per-cause patches would miss the next path; an edge detector covers limit, amp-fault, ferror, misc-error, spindle-fault and any future cause uniformly.
- Default is VOLATILE_HOME=0, so configs with absolute feedback are untouched.
- Soft limits do not disable motion, so they do not trigger an unhome. This matches the contract: home is only lost when the drive is actually cut.

## Test plan
- [x] Sim config, VOLATILE_HOME=1: home X, jog Y into a hard limit switch. X loses homed status.
- [x] Sim config, VOLATILE_HOME=1: home X, ESTOP. X loses homed status (existing path, regression check).
- [x] Sim config, VOLATILE_HOME=1: home X, F2 off. X loses homed status (regression check).
- [x] Sim config, VOLATILE_HOME=1: home X, hit a soft limit. X stays homed (drive not cut).
- [x] Sim config, VOLATILE_HOME=0: home X, trigger a fault. X stays homed (no regression).
- [ ] Hardware (stepper + limit switch): reproduce the exact #3954 scenario.

### DIFF
--- a/src/emc/task/emctaskmain.cc
+++ b/src/emc/task/emctaskmain.cc
@@ -3295,6 +3295,7 @@ int main(int argc, char *argv[])
     }
     while (!done) {
         static int gave_soft_limit_message = 0;
+        static int prev_traj_enabled = 0;
         task_beat++;  // Task's heartbeat
 
         check_ini_hal_items(emcStatus->motion.traj.joints);
@@ -3314,6 +3315,14 @@ int main(int argc, char *argv[])
 	// update subordinate status
 
 	emcMotionUpdate(&emcStatus->motion);
+	// On a motion enabled -> disabled edge (limit switch, amp fault,
+	// following error, etc.) unhome joints marked VOLATILE_HOME. The estop
+	// and machine-off paths already do their own unhome; this catches the
+	// remaining drive-disable causes that don't flip task state.
+	if (prev_traj_enabled && !emcStatus->motion.traj.enabled) {
+	    emcJointUnhome(-2); // only those joints which are volatile_home
+	}
+	prev_traj_enabled = emcStatus->motion.traj.enabled;
 	// synchronize subordinate states
 	if (emcStatus->io.aux.estop) {
 	    if (emcStatus->motion.traj.enabled) {


BsAtHome:
There may be some tab/space indent mismatch.

So, if a joint fault does not trigger machine off or estop, then this is not triggered? How is the specific joint unhomed in that case?

grandixximo:
> There may be some tab/space indent mismatch.
> 
> So, if a joint fault does not trigger machine off or estop, then this is not triggered? How is the specific joint unhomed in that case?

The only only fault that does not machine off (that I could find) is soft limit, and if machine did not go off, joint should still be in good position, homing should still be valid.
amp-limit already turns off the machine, so no need to loose homing independently.

BsAtHome:
But that means you _must_ have any joint's fault signal(s) wired up so that the machine goes either off or in estop. Otherwise, your joint may lose tracking and linuxCNC is not aware of that.

grandixximo:
> But that means you _must_ have any joint's fault signal(s) wired up so that the machine goes either off or in estop. Otherwise, your joint may lose tracking and linuxCNC is not aware of that.

Which joint fault does not turn off the machine? limit switches machine off, amp-fault machine off.

BsAtHome:
That which comes from external (like a signal from the drive controller hardware).

grandixximo:
> That which comes from external (like a signal from the drive controller hardware).

I cannot find a fault path that disables a single joint while leaving the machine on. joint.N.amp-fault-in, hard limit, ferror, misc error, spindle fault, all route through `emcmotInternal->enabling = 0`, which disables whole motion (traj.enabled 1=>0). edge detector catches that uniformly, so every fault visible to the core unhomes volatile joints.

I cannot find a "one joint faulty, machine stays running" mode, if it exist I can hook signals to unhome only the effected joint that's faulty and volatile.

BsAtHome:
If there is no path for a single joint failure while the machine stays on, then that is just the way it is and this solution will be adequate to fix the current problem.

If anybody wants to use such a single-joint-failure-mode feature (probably only for very special robotics), then they can send a patch ;-)

PR #4009, opened 2026-05-05
Add Support for Dogbone Homing

sundtek:
I'm retrofitting my Mitsubishi Meldas M65 based CNC, the homing switches are dogbone style switches.
The axis first approaches the endstop bar at coarse speed, then switches to a fine homing speed once the switch is triggered. When the switch is released again, the controller registers the axis as homed.

I have tested this behaviour with Linuxcnc 2.9.8

This patch introduces a new option HOME_DOGBONE = [bool] for the [JOINT_n] section. Default = off


hansu:
What are dogbone style switches?

BsAtHome:
> What are dogbone style switches?

If I understand it correctly from the PR description... The homing switch is of assert/deassert type, where you "roll over the switch" and the deassertion is the exact home position.

So, move v<sub>coarse</sub> towards home until the homing switch activates and when it does, continue direction but reduce speed to v<sub>fine</sub> until the switch deactivates. However, the code does not do the high/slow speed transition afaics. Or does it?

mozmck:
It's there in homing.c line 878-885.
The logic looks good to me.  The only small issue I see is that I *think* any custom homing components will have to be updated for the extra home_dogbone variable as was done in homecomp.comp
Not a big deal I don't think and it probably won't affect many people but it does me as I have a custom component to improve dual gantry homing.

BsAtHome:
> It's there in homing.c line 878-885.

Check, thanks.

grandixximo:
There are also components for ethercat deferred homing floating around, those will need update as well if this lands?

hansu:
> > What are dogbone style switches?
> 
> If I understand it correctly from the PR description... The homing switch is of assert/deassert type, where you "roll over the switch" and the deassertion is the exact home position.
> 
> So, move vcoarse towards home until the homing switch activates and when it does, continue direction but reduce speed to vfine until the switch deactivates. 

I understand the loigic, but I wonder how the switch looks like and if it's a common name for that because I have never heard of that and also cannot find anything about that notion.

It could look like this

grafik

hansu:
Further it would be good to add documentation about that in https://linuxcnc.org/docs/html/config/ini-homing.html

grandixximo:
If I understand correctly this should be incompatible with managing homing+limits with the same sensor, correct?
`[HOME_IGNORE_LIMITS] = TRUE` will be a bad flag to have when using this style of homing??

sundtek:
I couldn’t find the documentation in the source code yesterday. Maybe I was just too tired and missed it. Certainly it should be added to ./docs/src/config/ini-homing.adoc

The switch is around 25 years old and was made by a German company:
This is used for the Z axis:
image
This is used for the X axis:
image

One switch is used for homing, the other one for overtravelling detection, in total every axis has 3 dogbones/bars installed (one for homing, and two or overtravel detection which is hardwired and able to cut the power stage to the amplifiers)

grandixximo:
I think the dogbone term doesn't really fit here.
`HOME_UNIDIRECTIONAL`
`HOME_FALLING_EDGE`
Maybe a better flags?

sundtek:
As a reference:
https://dl.mitsubishielectric.com/dl/fa/document/manual/cnc/bnp-b2237(eng)/bnp-b2237(eng)j.pdf
Page 389 (III – 3)

we can certainly use a different name for it.

grandixximo:
What you posted is different from proposal, the homing happens on next encoder zero edge, not on the sensor off edge.

sundtek:
> What you posted is different from proposal, the homing happens on next encoder zero edge, not on the sensor off edge.

of course when the sensor transitions from on to off. Of course there's some room for interpretation and how it's wired.
image

Has anyone seen this type of homing with other machines?
HOME_UNIDIRECTIONAL would be fine with me, HOME_DOGTYPE too (according to the Mitsubishi documentation).

grandixximo:
Just saying HOME_USE_INDEX seems very relevant here.
Does current implementation account for that?
In the shared picture, motion seems to anticipate the slow down to end on the next grid point, this might be crucial to what makes this accurate and repeatable on certain machines? I don't see much room for interpretation in the manual, it is pretty clear its doing encoder Z input grid calculations.

sundtek:
No the current implementation only takes care about the switch.

That's a good point I did not pay attention about the index signal (it's a digital protocol), I'll put it on my list for the next days because I need to be in-front of the machine for that.

andypugh:
I thought that this behaviour was already configurable? With a HOME_SEARCH_VELOCITY and (lower) HOME_LATCH_VELOCITY with the same sign? (I haven't tested this)

grandixximo:
> I thought that this behaviour was already configurable? With a HOME_SEARCH_VELOCITY and (lower) HOME_LATCH_VELOCITY with the same sign? (I haven't tested this)

No, currently there is always at least one direction switch, same sign adds a second direction change, this PR is unidirectional, missing behavior.

image

sundtek:
The patch as it is is okay and works as described - it just doesn't take care about Z-Index of the servo / which is another type of homing / which is described in the Mitsubishi documentation.

Anyway I will keep working on it I didn't read the documentation about it since I just needed some kind of homing work first with my machine - I only used the deasserted point as zero point.

Basically after deasserting the switch I would have to check for the Z-Index of the corresponding servo, it needs another stage. I need to check how to get the index signal since it's a digital protocol.

I will have a look at that within the next days - it's good to discuss that topic here before merging it.

grandixximo:
You should get much more repeatable homing using the Z-Index

src/emc/motion/command.c
```diff
@@ -657,7 +657,8 @@ void emcmotCommandHandler_locked(void *arg, long servo_period)
 	                            emcmotCommand->latch_vel,
 	                            emcmotCommand->flags,
 	                            emcmotCommand->home_sequence,
-	                            emcmotCommand->volatile_home
+	                            emcmotCommand->volatile_home,
+				    emcmotCommand->home_dogbone
```
andypugh, line 661: Looks like a space/tab thing that doesn't match existing code

andypugh:
> No, currently there is always at least one direction switch, same sign adds a second direction change, this PR is unidirectional, missing behavior.

Maybe all that is needed is a new parameter HOME_NO_REVERSE ?

(I don't think that DOGBONE is particularly self-explanatory)

Then the "dogbone" home sequence could (probably) re-use existing home states. 
ie here: https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/motion/homing.c#L960
set state HOME_FALL_SEARCH_START if the sign of home and latch is the same _and_ HOME_NO_REVERSE is set?

grandixximo:
I think `HOME_NO_REVERSE = TRUE` should force non reverse behavior regardless of  `HOME_LATCH` sign, `HOME_SEARCH` is the direction that you will not _reverse_ for latching, regardless of latch sign, users will otherwise  experience unexpected behavior, where the flag is on but skipped.

hansu:
>Maybe all that is needed is a new parameter HOME_NO_REVERSE ?

Or `HOME_NO_BACK_OFF` to match the the docs...

Maybe better avoid double negation with and name it `HOME_BACK_OFF` which is true  by default.

rodw-au:
Sounds like Mtsibishi named it after a dog or pawl which you might find on a hand winch. Or on a Warrego Wool Press ratchet mechanism, a 8 ft high lever excellent for physical workouts but I'm showing my age and ethnicity there!

andypugh:
Presumably is named after the classic "thing with a bulge at each end" as classically depicted with cartoon dogs. 
https://www.istockphoto.com/de/vektor/hund-mit-knochen-gm177050080-26530901
Clearly in cartoon land there exist meat animals with knees on both ends of the femur. 
Many things get the name, for some similarity to the anatomically incorrect bone:
https://www.aliexpress.com/item/1005006376883246.html

This is partly why I would prefer not to see the name used here, as the "dogbones" in question do not appear to be dogbone shaped....

grandixximo:
@BsAtHome Had a chat with the sundtek on Discord, he said he will be back at this next week, will help him clean this up as you nudged me at the meeting.

sundtek:
I will suspend to work on it for 1 1/2 weeks because I'm waiting for prototype interface PCBs. I will continue with it once the first prototype board is assembled.

sundtek:
Yesterday I deployed the first revision of my Mitsubishi M65 LinuxCNC adapter PCB, so I will continue to work on this again now.

* If the "dog" is already engaged when starting the homing procedure it better throws an error.
* there's already an index option in LinuxCNC, I will check to re-use this flag.

There should be two options:
1. dog style homing without motor index (the current patch is adressing this already)
2. dog style homing with motor index (I will work on this now)

sundtek:
I got around to implement and test this yesterday, the results are good (+/- 5 microns for that 25 year old machine after re-powering the machine and measured with a Mitutoyo dial indicator with 1 micron resolution)

After this I even dared to enable my umbrella toolchanger implementation - and it worked better than ever before (before I was only able to rotate the umbrella in one direction with M6, now it will pick the shortest path)

Fast homing to the endstop switch, afterwards switching to index searching once the endstop switch is released. 
For the index searching I'm re-using the existing infrastructure, so the effective extra code will be just 3-5 lines added to the current patch - and the rest is handled via the configuration.

The code is on the Raspberry PI now which is in the workshop, I will get it back around next week and submit it.

dirkx:
> Just saying HOME_USE_INDEX seems very relevant here. Does current implementation account for that? In the shared picture, motion seems to anticipate the slow down to end on the next grid point, this might be crucial to what makes this accurate and repeatable on certain machines? I don't see much room for interpretation in the manual, it is pretty clear its doing encoder Z input grid calculations.

So in de Deckel-MAHO DMC series (50, 70, 100, etc) this is also used; and only in one direction as well. The grid offset (Grdmask param) is normally zero - but can be useful if the dogbone itself shifts/was moved or otherwise needs to be aligned with something from outside; such as a pallet. So you do not have to rely on getting it (re)mounted by hand with a few microns accuracy.

dirkx:
> As a reference: https://dl.mitsubishielectric.com/dl/fa/document/manual/cnc/bnp-b2237(eng)/bnp-b2237(eng)j.pdf Page 389 (III – 3)
> 
> we can certainly use a different name for it.

I think that there is a subtle twist to this - as I understand/observe it --  the Mitsubishi servo controller does the fast onto dogbone, slow to come `off' - an then takes the (relative) linear encoder readout, adds any grid correction factor set -- and then uses this as the _absolute_ correction. I surmise that this works as the repeating grid on the (relative) linear encoder is in fact in a fixed position. 

So it makes a fairly accurate position more accurate. 

So essentially - it does use the end of the dogbone as a rough position into the linear encoder/DRO - and then uses that to `round' to the actual, machine fixed, reference (before applying its temperature adjustment)

sundtek:
Yes it's exactly that way, initially I thought it's only using the dog for homing - not the motor index.

I'll update this patch until Monday so it can go upstream, it worked well during all my tests.
The final  version will rename dogbone to dogtype homing; and with a few more lines it can also re-use the existing index homing infrastructure after the dogtype homing and just needs an appropriate configuration file.

dirkx:
With regard to:

> not the motor index.

I am fairly sure it uses the motor index first - and then, on the machines that have it, relies on the Z and relative to that pulses from the linear encoder.

And I speculate that in each case - it makes sure that it has traveled enough to have seen at least one Z pulse (as it seems to do a very short funny dance if you are too near the end of dogbone when powering up).

So I **think** but am not sure - that it is the SPS/CNC control that needs to ensure this - but have no way yet to test this.

grandixximo:
@sundtek still on this?

sundtek:
Hi, yes we're testing it with @dirkx setup this week.

PR #4013, opened 2026-05-09
support DB_PROGRAM to be absolute, relative or in PATH

### LINKED ISSUE #3982: unable to run binaries on PATH in DB_PROGRAM
Hello

`DB_PROGRAM` var from the `INI`  doesn't allow binaries in `PATH`

```ini
[EMCIO]
EMCIO = io
CYCLE_TIME = 0.100
DB_PROGRAM = tool_db_backend
# TOOL_TABLE = tool.tbl
```

```
!!!db_init: <tool_db_backend> not executable

 2561 IO::tooldata_db_init() FAIL


tooldata_load: Failed to load tooldata from database

emc/iotask/ioControl.cc 786: can't load tool table.

```

it requires an absolute path of the database program

Thanks

TurBoss:
Hello

This commit allows DB_PROGRAM to handle absolute, relative path or binaries in env PATH

closes #3982 

Thanks

Edit: also check if path/file exists

### DIFF
--- a/src/emc/tooldata/tooldata_db.cc
+++ b/src/emc/tooldata/tooldata_db.cc
@@ -246,11 +246,11 @@ int tooldata_db_init(char progname_plus_args[],int random_toolchanger)
     char* saveptr;
     char* token = strtok_r(progname_plus_args, " ", &saveptr);
 	int n;
-	char* fork_argv[MAX_DB_PROGRAM_ARGS] = {0};
+	char* fork_argv[PATH_MAX] = {0};
 	char* cp;
 	char* envpath;
-	char prog_name[MAX_CMD_LEN+1] = "\0";
-	char prog_path[MAX_CMD_LEN+1];
+	char prog_name[PATH_MAX] = "\0";
+	char prog_path[PATH_MAX];
 	struct stat stat_buf;
     while (token != NULL) {
         child_argv[child_argc] = token;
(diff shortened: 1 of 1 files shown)


src/emc/tooldata/tooldata_db.cc
```diff
@@ -243,6 +245,13 @@ int tooldata_db_init(char progname_plus_args[],int random_toolchanger)
     char* child_argv[MAX_DB_PROGRAM_ARGS] = {0};
     char* saveptr;
     char* token = strtok_r(progname_plus_args, " ", &saveptr);
+	int n;
+	char* fork_argv[PATH_MAX] = {0};
```
BsAtHome: Why are you allocating PATH_MAX pointers to char? I think that PATH_MAX is normally about 4096 and you are allocating that many pointers. Should use std::vector.

See also other comment about this.

src/emc/tooldata/tooldata_db.cc
```diff
@@ -255,13 +264,66 @@ int tooldata_db_init(char progname_plus_args[],int random_toolchanger)
...
     is_random_toolchanger = random_toolchanger;
-
-    if (access(child_argv[0],X_OK)) {
-        fprintf(stderr,"!!!!db_init: <%s> not executable\n",child_argv[0]);
+	if (db_childname[0] != '/') {  // prog path not absolute
+		if(strchr(db_childname, '/') != NULL){ // prog path is relative
+			strncpy(prog_path, db_childname, PATH_MAX);
+		}
+		else { // just prog name, search  in the actual directory or PATH
+			envpath = getenv( (char*)"PATH");
```
BsAtHome, line 272: ...and the rest...
I think this needs to be redesigned using std::string and std::vector\<std::string>. There is no point nowadays to use C style (more error prone) string handling in a C++ source file. It was already outdated in 2021 when written and added. Now, when you try to fix it, it should be modernized to remove the standard pitfalls.

BsAtHome (changes requested):

src/emc/tooldata/tooldata_db.cc
TurBoss: reduced that to 10, was a mistake thanks!

src/emc/tooldata/tooldata_db.cc
TurBoss: sorry I'm not much familiar with c++

BsAtHome:
Some more observations while reading this code...
The code is wrong how it splits arguments. It uses a simple ' ' (space) tokenizer and neglects both IFS and quoting. I can understand IFS ignorance, but when you split arguments from a value you got from the INI-file and want to execute the program, how then can you supply it properly with arguments if such an argument includes a space?

When you change the code to use a path search, why then not make it easier for yourself and run it via a shell -c exec (_not_ a call to system)? Then the shell can deal with all the intricacies of argument splitting and path searches?

The code will crash and burn when the maximum number of arguments are parsed because the call to execv() must terminate the arg list with a NULL. In effect, this is a off-by-one error.

The code also spits out stuff using fprintf() and not the usual rcs_print stuff. This happens a lot of places (not only in the tooldata stuff).

Other things in the code:
When the pipe creating or forking fails, it simply calls exit() killing the whole _parent_ process and not gracefully returning an error that can be handled up the call-tree.

The code gets debug variables from the environment, where it should be reading those from the INI-file as the central place for configuration.

The SIGCHLD handler waits for all processes that may be a zombi (that is good), however, it then only checks the _last_ of the zombi's status. What about the others?

Probably a lot more if I start looking deeper. Enough for now.

BsAtHome:
BTW, using a shell -c is also a very tricky suggestion requiring a lot of escaping.

andypugh:
We reviewed this in the 24/5/26 video meeting and decided this is not a desirable feature.

PR #4015, opened 2026-05-11
axis: fix Z jog after touch off on unhomed lathe (2.9 backport of #4007)

grandixximo:
## Summary

Backport of #4007 (merged to master/2.10) to the 2.9 branch.

Fixes #3994: on an unhomed lathe, touching off forces teleop mode via `set_motion_teleop(1)`. A follow-up Z jog then hit the teleop axis slot 1 (Y) instead of slot 2 (Z) and silently no-op'd, because `ja_from_rbutton` used the compact `xzabcuvw` letter map whenever the machine was unhomed, regardless of jog mode.

Fix gates the compact map on `jjogmode` so teleop axis jogs always use the full `xyzabcuvw` slot map.

Cherry-pick of 7a6dfdd onto 2.9, no conflicts.

## Context

I communicated with the original reporter of #3994. They are on stable 2.9.8 and do not want to build from source. The best I can offer is a PR against 2.9 so the fix lands in the next stable release.

## Test plan

- [x] Verify Z jog works after touch off on unhomed lathe (the original repro from #3994)
- [x] Verify joint jog mode on lathes without Y still uses compact `xzabcuvw` map
- [x] Verify non-lathe configs unaffected

PR #4017, opened 2026-05-11
ci: install adduser explicitly in package-arch test step

grandixximo:
## Summary

The package-arch *Test debian packages* step calls `eatmydata adduser ... testrunner` to set up a non-root user for runtests, but never installs the `adduser` package. It used to work because `adduser` was pulled transitively via Recommends of the installed `.deb` files; sid no longer pulls it, so builds fail with:

```
E: eatmydata: unable to find 'adduser' in PATH
##[error]Process completed with exit code 1.
```

Surfaced today on PR #3999 (all six package-arch jobs across bookworm/trixie/sid x amd64/arm64 failed). Same failure mode reproduces on rerun, so this is not transient.

## Fix

Add `adduser` to the existing `apt-get install sudo` line so the testrunner setup does not depend on a fragile Recommends chain.

## Test plan

- [ ] CI passes on this branch (this PR self-tests the fix)
- [ ] Re-run CI on PR #3999 after merge confirms package-arch jobs progress past testrunner setup

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,7 +208,7 @@ jobs:
         set -e
         set -x
         eatmydata apt-get --yes --quiet -o Acquire::Retries=5 install ../*.deb
-        eatmydata apt-get --yes --quiet -o Acquire::Retries=5 install sudo # some tests run sudo...
+        eatmydata apt-get --yes --quiet -o Acquire::Retries=5 install sudo adduser # some tests run sudo; adduser no longer pulled implicitly on sid
         eatmydata adduser --disabled-password --gecos "" testrunner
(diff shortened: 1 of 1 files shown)


BsAtHome:
This conflicts with #3984, which is the one that should go in.

You'll have to rebase when I do...

BsAtHome:
If you could be so kind, I think you need to re-apply this but to 2.9 branch. There the CI is not going to be upgraded and it fails with the same error.

grandixximo:
I was about to text you about it, just realized I need this on 2.9 for #4015

grandixximo:
@BsAtHome I did it as #4018 but it fails cause of GitHub now...

PR #4018, opened 2026-05-11
ci: install adduser explicitly in package-arch test step

grandixximo:
Backport of master commit f9b363292c onto 2.9.

The test step runs `eatmydata adduser ... testrunner` but never installs `adduser`. It was pulled transitively via Recommends on the installed `.deb` files; sid no longer pulls it, so builds fail with:

```
eatmydata: unable to find adduser in PATH
```

Add it to the existing sudo install so testrunner setup does not depend on Recommends.

2.9's `ci.yml` has the same vulnerable line (line 169 pre-patch) as master had, so the fix is identical.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,7 +166,7 @@ jobs:
         set -e
         set -x
         eatmydata apt-get --yes --quiet -o Acquire::Retries=5 install ../*.deb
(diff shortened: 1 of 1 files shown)


grandixximo:
that makes no sense, error on linuxcnc/linuxcnc ???
@BsAtHome must be transient fail, please re-run...

grandixximo:
We either hit some CI limit, or github is just 💩 today

BsAtHome:
You can now rebase you other 2.9 PRs.

PR #4024, opened 2026-05-14
Ci improvements backport 2.9

hdiethelm:
If you think this is useful, I will do spot checks on created Debian packages and then it is ready to merge.

It was quite easy to cherry-pick. Let's see if the CI passes. When merging 2.9 to master, care has to be taken that no CI changes are merged.

### DIFF
--- a/.github/scripts/add-linuxcnc-repository.sh
+++ b/.github/scripts/add-linuxcnc-repository.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+set -eu #Needed so CI fails when anything is wrong
+set -x
+
+DIST=$1
+
+echo "deb [arch=amd64,arm64 signed-by=/etc/apt/keyrings/linuxcnc.gpg] https://www.linuxcnc.org/ $DIST base" | sudo tee /etc/apt/sources.list.d/linuxcnc.list > /dev/null
+case $DIST in
+	'buster' | 'bullseye' | 'bookworm')
+                gpg --homedir "${PWD}/gnupg" --export 3CB9FD148F374FEF | sudo tee /etc/apt/keyrings/linuxcnc.gpg > /dev/null
+                ;;
+        *)
+                GPGTMP=$(mktemp -d /tmp/.gnupgXXXXXX)
+                gpg --homedir "$GPGTMP" --keyserver hkp://keyserver.ubuntu.com --recv-key e43b5a8e78cc2927
+                gpg --homedir "$GPGTMP" --export 'LinuxCNC Archive Signing Key' | sudo tee /etc/apt/keyrings/linuxcnc.gpg > /dev/null
+                ;;
+esac
+sudo apt-get --quiet update
--- a/.github/scripts/add-linuxcnc-repository.sh
+++ b/.github/scripts/add-linuxcnc-repository.sh
@@ -5,6 +5,7 @@ set -x
 
 DIST=$1
 
+mkdir -p /etc/apt/keyrings
 echo "deb [arch=amd64,arm64 signed-by=/etc/apt/keyrings/linuxcnc.gpg] https://www.linuxcnc.org/ $DIST base" | sudo tee /etc/apt/sources.list.d/linuxcnc.list > /dev/null
 case $DIST in
 	'buster' | 'bullseye' | 'bookworm')
--- a/.github/scripts/build-doc.sh
+++ b/.github/scripts/build-doc.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+set -eu #Needed so CI fails when anything is wrong
+set -x
+
+cd src
+./autogen.sh
+./configure --disable-check-runtime-deps --enable-build-documentation=html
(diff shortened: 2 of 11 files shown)


BsAtHome:
And,... crash and burn. Typical... ;-)

hdiethelm:
> And,... crash and burn. Typical... ;-)

Some gitignores are missing, failed as it should... ;-)

hdiethelm:
Ah, cpp / shellcheck don't exist in 2.9... :-D Other job on 2.9:
https://github.com/LinuxCNC/linuxcnc/actions/runs/25679989364/job/75388517378
```

Run scripts/shellcheck.sh
/home/runner/work/_temp/39747ac5-1c7a-439a-a3f8-6dda3c7f7c3b.sh: line 1: scripts/shellcheck.sh: No such file or directory
Error: Process completed with exit code 127.
```
Anyway, how long until 2.10 is released?

BsAtHome:
Shellcheck and cppcheck, yes, they were added and a _**lot**_ of problems got solved that way around.

2.10 is still some time in the future. Working hard on fixing the hal type problem. That should go in to eliminate the 32/64 bit differentiation. Then we're probably mostly done for release.

hdiethelm:
Do you think this back-port is useful? Doc is still failing. I would have to investigate...

BsAtHome:
The backport is useful. The CI on 2.9 takes 55 minutes or so, where master is down to 21..23 minutes. That would be an improvement.

hdiethelm:
So, this one should hopefully pass. The last one had a timeout but indep is now fine.
The issue was the po4a installed from linuxcnc repo which is needed for bullseye. Is bullseye even needed?

hdiethelm:
So, passed. 52min down to 30min. It seams the dock build is less performant on 2.9.

BsAtHome:
I don't think you need to do the cppcheck/shellcheck part at all. The list of errors, warnings and other problems is so long... It was fixed in master and I'm not going to backport that effort. You can just remove the CI for it IMO (never worked anyway).

BsAtHome:
> Is bullseye even needed?

It is supported for 2.9, so yes.

hdiethelm:
> I don't think you need to do the cppcheck/shellcheck part at all. The list of errors, warnings and other problems is so long... It was fixed in master and I'm not going to backport that effort. You can just remove the CI for it IMO (never worked anyway).

Ok, so I commented out the target.

BsAtHome:
Do you think this is ready or are you not satisfied yet?
Increasing the resilience and speed of CI would be a nice addition :-)

hdiethelm:
The only thing missing is:

Download a few Debian packages from the old and the new CI and compare the contents to detect if might be some configure stuff automatically deactivated some parts due to missing packages.

If you can do that, I will set it to ready. Or I can do that until end of the week.

BsAtHome:
No rush... I'm working on HAL and would need to familiarize myself with the procedure of testing.
So, if it takes a week, or two, or three, I'd be fine with that too. I was just wondering.

hdiethelm:
What tested based on possible impacts by this change:

I compared the unpacked deps for bookworm / trixie with https://github.com/LinuxCNC/linuxcnc/actions/runs/25679989364 (Base of this PR): 
No alarming differences. Differences in dates in files & PDF / generated images. I checked a few of the images and they look the same.

Additionally, I installed and run the trixie deb's with axis_mm / gmoccapy, all works fine.

I am not able to test the release stage (might be I could by tag&push my branch but that would mess up your releases... ;-) ) but there are only minor changes in this stage.

hansu:
@hdiethelm You added the line

` mkdir -p /etc/apt/keyrings` in .github/scripts/add-linuxcnc-repository.sh

in the 2.9 backport but not in master. Why that? Should I keep it like it is?

```
#!/bin/sh

set -eu #Needed so CI fails when anything is wrong
set -x

DIST=$1

<<<<<<< HEAD
=======
mkdir -p /etc/apt/keyrings
>>>>>>> upstream/2.9
echo "deb [arch=amd64,arm64 signed-by=/etc/apt/keyrings/linuxcnc.gpg] https://www.linuxcnc.org/ $DIST base" | sudo tee /etc/apt/sources.list.d/linuxcnc.list > /dev/null
case $DIST in
	'buster' | 'bullseye' | 'bookworm')
                gpg --homedir "${PWD}/gnupg" --export 3CB9FD148F374FEF | sudo tee /etc/apt/keyrings/linuxcnc.gpg > /dev/null
                ;;
        *)
                GPGTMP=$(mktemp -d /tmp/.gnupgXXXXXX)
                gpg --homedir "$GPGTMP" --keyserver hkp://keyserver.ubuntu.com --recv-key e43b5a8e78cc2927
                gpg --homedir "$GPGTMP" --export 'LinuxCNC Archive Signing Key' | sudo tee /etc/apt/keyrings/linuxcnc.gpg > /dev/null
                ;;
esac
sudo apt-get --quiet update

```

hdiethelm:
Yes, this is only needed in 2.9 due to debian bullseye does not has that folder yet. Not needed in master.

PR #4031, opened 2026-05-16
task: unblock tool-change abort by setting io.status=DONE

### LINKED ISSUE #3675: LinuxCNC 2.10 becomes unresponsive if aborted during tool change
Tested with 2.9 too, behaviour is as-expected there. 

To reproduce use a simple G-code routine with tool changes. 
eg
```
M6 T1 G43
G1 F50 X10
M6 T2 G43
G1 F50 X10
M2
```
Open LinuxCNC and sim-axis config
Load the file, and open the Halshow window. 
Add iocontrol.0.tool-change and tool-changed
unlink tool-changed (right click on the pin in halshow watch is the easiest way to go)
Open a terminal and set up python to send aborts:
```
python3
import linuxcnc
c=linuxcnc.command
c.abort()
```
Run the code and send an abort during normal motion. The program stops, all is well, the program can be re-started. 
See that manually toggling tool-changed in halshow when tool-change is set allows the tool-changes to complete. 
Now send an abort from Python while tool-change is set and without having set tool-changed, so that the system thinks it is mid-change. 

In my testing the machine can then not be caused to re-run the G-code, or do MDI, or very much at all.

grandixximo:
## Summary

Fixes #3675 (2.10 regression: aborting during a tool-change wedges the UI; MDI and further auto runs are dead until restart).

After 764655eb4d moved IO handling from the out-of-process `iocontrol` into `Task`, `EMC_TOOL_ABORT` no longer flips `emcioStatus.status` back to `RCS_DONE`. The old NML dispatcher in `ioControl.cc` set `RCS_DONE` by default for every command before the switch; the new in-process `Task::emcIoAbort` clears the `tool_change`/`tool_prepare` pins but leaves status stuck at `RCS_EXEC` from the prior `emcToolLoad`/`emcToolPrepare`.

`emctaskmain`'s MDI, auto and state-restore paths gate on `io.status == RCS_DONE`, so abort while tool-change is pending wedges everything.

One-line fix: set `emcioStatus.status = RCS_STATUS::DONE` at the end of `Task::emcIoAbort`, restoring the old default.

@rene-dev, sorry for stepping on this one. Bisect from #3675 pointed at 764655eb4d so I had a clear lead and went ahead. Happy to defer if you'd prefer to take it.

## Test plan

- [x] Reproduced original bug on master (sim/axis, sample G-code with `M6 T1 G43` / `M6 T2 G43`, `iocontrol.0.tool-changed` unlinked, `linuxcnc.command().abort()` while `tool-change=1`): UI sluggish/unresponsive, MDI dead.
- [x] With patch applied: abort recovers cleanly, MDI works, G-code can be re-run.
- [x] Reviewer to confirm on real hardware tool-changer.

### DIFF
--- a/src/emc/task/taskclass.cc
+++ b/src/emc/task/taskclass.cc
@@ -369,6 +369,9 @@ int Task::emcIoAbort(EMC_ABORT /*reason*/)//EMC_TOOL_ABORT_TYPE
     iocontrol_data.coolant_flood = 0;                /* coolant flood output pin */
     iocontrol_data.tool_change = 0;                /* abort tool change if in progress */
     iocontrol_data.tool_prepare = 0;                /* abort tool prepare if in progress */
+    // release task wait on pending tool-change/prepare (old NML iocontrol
+    // returned RCS_DONE by default; in-process call must do it explicitly)
+    emcioStatus.status = RCS_STATUS::DONE;
     return 0;
 }
 


grandixximo:
@andypugh I reproduced with your steps in #3675 and this fixes it. 2.9 not effected.

PR #4039, opened 2026-05-17
docs: Fix trailing many '+' and delimit note blocks.

### LINKED ISSUE #4038: malformed adoc file for qtdragon
I just noticed this in the logs when building master branch. Its not a biggie as it did not break the build but probably good to fix it.

```
make[2]: Entering directory '/tmp/linuxcnc-dev/src'
cd ../docs &&  po4a  --msgmerge-opt='-v' --no-translations po4a.cfg
src/gui/qtdragon.adoc:458: It seems that you are adding unindented content to an item. The standard allows this, but you may still want to change your document to use indented text to provide better visual clues to writers.
src/gui/qtdragon.adoc:460: It seems that you are adding unindented content to an item. The standard allows this, but you may still want to change your document to use indented text to provide better visual clues to writers.
src/gui/qtdragon.adoc:461: It seems that you are adding unindented content to an item. The standard allows this, but you may still want to change your document to use indented text to provide better visual clues to writers.
 (34180 entries)

```

BsAtHome:
Remove most trailing '+' to force line breaks. Hard line breaks are mostly inappropriate and it should be left to paragraph formatting.
When using hard breaks in lists, then the line after the break must be indented or a warning will be issued. This was the actual cause of the messages identified in #4038.

The change caused some note/important/caution blocks to be slightly off because they were not delimited and relied on empty line delimit. These are now properly delimited so that blocks are actually identified as blocks.

This fixes #4038.

### DIFF
--- a/docs/src/gui/qtdragon.adoc
+++ b/docs/src/gui/qtdragon.adoc
@@ -23,9 +23,11 @@ You can use LinuxCNC's external offsets capability to automatically raise the sp
 If you the VersaProbe option and remap code you can add auto tool length probing at tool changes.
 
 [NOTE]
+====
 QtDragon and QtVCP are relatively new programs added into LinuxCNC.
 Bugs and oddities are possible. Please test carefully when using a dangerous machine.
 Please forward reports to the forum or maillist.
+====
 
 === QtDragon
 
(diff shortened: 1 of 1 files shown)


c-morley:
I dont want this. I like my docs as they were. If there were warnings on a couple lines I will look at those, but please don't reformat my docs.
Thanks

BsAtHome:
Forced line breaks are inappropriate formatting and actually break proper formatting when floating formatting is used (like in a browser).

However, you want it, you do it and I'll refrain from looking at your the docs.

PR #4041, opened 2026-05-18
tests: Silence left-over (debug) message in tests

BsAtHome:
This removes the redundant (annoying) message "line 1314: accel phase over" from tests/motion/g0.

It was a left-over debug message where the message starting the acceleration already was commented out. This PR takes care of the second message and has no functional change.


BsAtHome:
Also remove the messages from checkresult in tests/module-loading/rtapi-app-main-fails. No news is good news.

PR #4051, opened 2026-05-21
resolve #4048

### LINKED ISSUE #4048: GCode interpreter now depends on emcStatus
The addition of #3995 has introduced a serious problem. The whole interpreter library is now dependent on `emcStatus` and by extension, libnml. This is obviously not a good thing.

The addition of the following line is the culprit:
```C++
   int machineUnits = (emcStatus->motion.traj.linearUnits > 0.5) ? CANON_UNITS_MM : CANON_UNITS_INCHES;
```
see: https://github.com/LinuxCNC/linuxcnc/blame/60597ee0718873d2449058c824262a275e5e4bad/src/emc/rs274ngc/interp_convert.cc#L6523

The real killer is that the value is not used in any sensible way at all. The `machineUnits` variable is used to scale the radius of a circle, but the variable `radius` is subsequently never used. It is only assigned a value (see lines 6529..6538). The `arc_radius` field value is written with a calculated value and that is the same call as originally in the assignment in line 6519.

@rodw-au: The question then arises: Why is this code in interp_convert.cc lines 6519..6538 necessary _at all_? It has no meaningful purpose because its calculations are never used. If scaling is required, then it must be done in the lower layer, but it seems odd that neither X, Y or Z is scaled in that same procedure and suggests that it is superfluous.

Error message when compiled with undefined sanitizing enabled:
```
  File ".../linuxcnc.git/bin/hal_manualtoolchange", line 87, in <module>
    import nf, rs274.options
  File ".../linuxcnc.git/lib/python/rs274/__init__.py", line 18, in <module>
    from .interpret import Translated, ArcsToSegmentsMixin
  File ".../linuxcnc.git/lib/python/rs274/interpret.py", line 17, in <module>
    import math, gcode
ImportError: .../linuxcnc.git/lib/librs274.so.0: undefined symbol: emcStatus
Found file(REL): ./cooling.hal
  File ".../linuxcnc.git-bsathome/bin/axis", line 61, in <module>
    import gcode
ImportError: .../linuxcnc.git/lib/librs274.so.0: undefined symbol: emcStatus
```

rodw-au:
Resolves an error reported in #4048 that was introduced in #3995 
I could not see that any header files had been added in the earlier changes. 
Please check me on that.

All tests pass locally.
`Runtest: 281 tests run, 281 successful, 0 failed + 0 expected, 4 skipped, 0 shmem errors
`

### DIFF
--- a/src/emc/rs274ngc/interp_convert.cc
+++ b/src/emc/rs274ngc/interp_convert.cc
@@ -6515,28 +6515,6 @@ int Interp::tag_arc(block_pointer block, double x, double y, double z, double ce
                 // Normalise 0-360
                 while (heading < 0) heading += 360.0;
                 while (heading >= 360.0) heading -= 360.0;
-
-                double radius = hypot(dx, dy);
-
-                // We need to scale the radius if the gcode units are not the same as machine units
-                // If linearUnits > 0.5, it's 1.0 (Metric). If < 0.5, it's 0.03937 (Imperial).
-                int machineUnits = (emcStatus->motion.traj.linearUnits > 0.5) ? CANON_UNITS_MM : CANON_UNITS_INCHES;
-
-                // 3. Identify the active G-code unit type
-                int gcodeUnits = _setup.length_units;
-
-                // 4. If they don't match, scale the radius to Machine Units
-                if (gcodeUnits != machineUnits) {
-                        // longhand below to avoid clang errors
-                        if (gcodeUnits == CANON_UNITS_INCHES) {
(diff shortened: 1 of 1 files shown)


BsAtHome:
> I could not see that any header files had been added in the earlier changes. Please check me on that.

Already did.

You are right, no additional headers were added and the PR should be fine. Funny I hadn't caught this particular issue in review. Oh well,... shit happens (on a very regular basis) ;-)

rodw-au:
You are right, I have no idea why I wasted a day or so researching and coding stuff that I never needed.
Shit happens here too :)

PR #4052, opened 2026-05-23
dbounce: add instant-read for probing moves

djdelorie:
Add motiontype and probedelay pins

If motiontype is 5, then the logic changes as follows:

- if state is max or min (i.e. signal stable), the first change forces an output change and sets state to the midpoint

The effective result is that an input change after the delay expires, causes an immediate output change, but any further transitions within the delay are ignored.

If probedelay is set (non-zero) it is used instead of delay during probing moves.

If motiontype is not connected, the behavior is as it was before this patch, for backward compatibility.

### DIFF
--- a/src/hal/components/dbounce.comp
+++ b/src/hal/components/dbounce.comp
@@ -1,13 +1,18 @@
-component dbounce """alternative debounce component""";
-description """
-This component is similar to the *debounce*(9) component but uses settable
-delay pins for each instance and supports *count*= or *names*= parameters
-(groups are not used)
+component dbounce """alternative debounce component\n
+This component is similar to the \\fBdebounce\\fR component
+(man \\fBdebounce\\fR) but uses settable delay pins for each instance
+and supports \\fBcount\\fR= or \\fBnames\\fR= parameters
+(groups are not used).  It also supports an instant-hit mode for probing moves
+where the output changes immediately but further changes (bounces)
+are ignored until stable.
 """;
 
 pin in  bit in;
 pin out bit out;
-pin in  u32 delay = 5;
+pin in  u32 delay = 5 "delay in cycles when not probing";
+pin in  u32 probedelay = 0 "delay in cycles when probing (if unconnected, uses delay)";
+pin in  s32 motiontype = -1 """connect to motion.motion-type -
+when equal to 5 it means probing""";
 
 variable unsigned state;
 option period no;
@@ -16,10 +21,16 @@ license "GPL";
 author "Dewey Garrett";
 ;;
 FUNCTION(_) {
+    rtapi_u32 d = (motiontype == 5 && probedelay > 0)
+      ? probedelay : delay;
 
     if (in) {
+        /* probing, trigger on first edge.  */
+        if (motiontype == 5 && state == 0) {
+	    state = d / 2;
+            out = 1;
         /* input true, is state at threshold? */
-        if (state < delay) {
+	} else if (state < d) {
             /* no, increment */
             state++;
         } else {
@@ -27,8 +38,12 @@ FUNCTION(_) {
             out = 1;
         }
     } else {
+        /* probing, trigger on first edge.  */
+        if (motiontype == 5 && state >= d) {
+	    state = d / 2;
+	    out = 0;
         /* input false, is state at zero? */
-        if (state > 0) {
+        } else if (state > 0) {
             /* no, decrement */
             state--;
         } else {


src/hal/components/dbounce.comp
```diff
@@ -1,34 +1,49 @@
...
+pin in  u32 probedelay = 0 "delay in cycles when probing (if unconnected, uses delay)";
+pin in  s32 motiontype = -1 """connect to motion.motion-type -
+when equal to 5 it means probing""";
 
-variable unsigned state;
-option period no;
-function _;
+/*variable int state;*/
+pin out u32 state;
+function _ nofp;
```
BsAtHome: There no longer is a difference between `fp` and `nofp`. All threads are now floating point because newer compilers use sophisticated optimizations using special (floating point) registers. The `nofp` must not be used anymore and you should not change this line.

src/hal/components/dbounce.comp
```diff
@@ -1,34 +1,49 @@
...
 pin in  bit in;
 pin out bit out;
-pin in  u32 delay = 5;
+pin in  u32 delay = 5 "delay in cycles when not probing";
+pin in  u32 probedelay = 0 "delay in cycles when probing (if unconnected, uses delay)";
+pin in  s32 motiontype = -1 """connect to motion.motion-type -
+when equal to 5 it means probing""";
 
-variable unsigned state;
-option period no;
```
BsAtHome, line 13: There is a reason why the `option period no` is present. The code does not use the period argument in the `_` function and it would therefore result in a warning, which are now treated as errors.

src/hal/components/dbounce.comp
```diff
@@ -1,34 +1,49 @@
...
-function _;
+/*variable int state;*/
+pin out u32 state;
```
BsAtHome: Changing this from `unsigned` to `u32` makes no sense because they are effectively the same. The comparison below will fail on a signdness mismatch.

src/hal/components/dbounce.comp
```diff
@@ -1,34 +1,49 @@
...
+      ? probedelay : delay;
 
     if (in) {
+        /* probing, trigger on first edge.  */
+        if (motiontype == 5 && state == 0) {
+	    state = d / 2;
+            out = 1;
         /* input true, is state at threshold? */
-        if (state < delay) {
+	} else if (state < d) {
```
BsAtHome, line 33: This is a comparison with different signedness.

src/hal/components/dbounce.comp
```diff
@@ -1,34 +1,49 @@
...
-pin in  u32 delay = 5;
+pin in  u32 delay = 5 "delay in cycles when not probing";
+pin in  u32 probedelay = 0 "delay in cycles when probing (if unconnected, uses delay)";
+pin in  s32 motiontype = -1 """connect to motion.motion-type -
+when equal to 5 it means probing""";
 
-variable unsigned state;
-option period no;
-function _;
+/*variable int state;*/
```
BsAtHome: Is there a reason to keep this as a comment?

src/hal/components/dbounce.comp
```diff
@@ -1,34 +1,49 @@
...
+pin out u32 state;
+function _ nofp;
 license "GPL";
 author "Dewey Garrett";
 ;;
 FUNCTION(_) {
+    int d = (motiontype == 5 && probedelay > 0)
```
BsAtHome: Using an `int` results in line 33 with a signdness mismatch.

The reason is that the variable `d` is declared an int, which should have been `rtapi_u32` to make it consistent with the underlying types used in `delay` and `probedelay`.

andypugh:
I don't really understand the point of this. Is it to address the issues already addressed by:
https://github.com/LinuxCNC/linuxcnc/pull/3537

src/hal/components/dbounce.comp
djdelorie: Fixed, sorry.  My working machine is not quite at git head and my testing didn't notice the difference.

src/hal/components/dbounce.comp
djdelorie: I planned on reverting it  after debugging

djdelorie:
> I don't really understand the point of this. Is it to address the issues already addressed by: #3537

I don't think so, I'm trying to solve a mechanical/electrical problem.  In my case, one of my probes occasionally "glitches" just right at certain slow speeds, despite using probing moves to move away.  So I occasionally get a "probe hit during non-probe move"  and the job halts.  The glitch could take up to 50 ms (it's not a real bounce, I think it's more like two contacts sliding somewhere causing sporadic noise, and sometimes happens during the retract - so the "move away" "probing" move completes, then we move normally, then trigger again...) and the existing debounce and dbounce logic delays the trigger by a non-predictable amount, which makes probing less accurate.
So what this change does is allow the probe to trigger on the first hit immediately, instead of waiting for the debounce logic to stabilize, and *then* ignores any noise for the delay time.
The probe-v-nonprobe logic is so that if the probe triggers while travelling (due to vibration), it's smoothed out as normal (along with backwards compatiblity ;)  as my machine is not that rigid, so vibrations can trigger the probes (especially the toolsetter).

src/hal/components/dbounce.comp
```diff
@@ -1,13 +1,18 @@
-component dbounce """alternative debounce component""";
-description """
-This component is similar to the *debounce*(9) component but uses settable
-delay pins for each instance and supports *count*= or *names*= parameters
-(groups are not used)
+component dbounce """alternative debounce component\n
+This component is similar to the \\fBdebounce\\fR component
+(man \\fBdebounce\\fR) but uses settable delay pins for each instance
+and supports \\fBcount\\fR= or \\fBnames\\fR= parameters
```
BsAtHome, line 4: The documentation is no longer in troff format. All components got updated to use the asciidoc format. You are changing it back to the older wrong state of affairs.

src/hal/components/dbounce.comp
```diff
@@ -16,19 +21,29 @@ license "GPL";
 author "Dewey Garrett";
 ;;
 FUNCTION(_) {
+    rtapi_u32 d = (motiontype == 5 && probedelay > 0)
+      ? probedelay : delay;
 
     if (in) {
+        /* probing, trigger on first edge.  */
+        if (motiontype == 5 && state == 0) {
```
BsAtHome, line 29: You are reading and comparing a pin multiple times. this is not very efficient.

Better is to use a local variable `bool mtype5 = motiontype == 5;` and then use `mtype5`. The compiler is then allowed to do proper optimization (which it cannot do on pins).

Also, comparing and assigning `probedelay` is inefficient for the same reason. The better way is:
```C
bool mtype5 = motiontype == 5;
rtapi_u32 pd = probedelay;
rtapi_u32 d = mtype5 && pd > 0 ? pd : delay;
```
This allows the compiler to optimize the code better because the pins are only read once.

src/hal/components/dbounce.comp
```diff
@@ -1,13 +1,18 @@
...
+are ignored until stable.
 """;
 
 pin in  bit in;
 pin out bit out;
-pin in  u32 delay = 5;
+pin in  u32 delay = 5 "delay in cycles when not probing";
+pin in  u32 probedelay = 0 "delay in cycles when probing (if unconnected, uses delay)";
+pin in  s32 motiontype = -1 """connect to motion.motion-type -
+when equal to 5 it means probing""";
```
BsAtHome, line 15: This is a _**very**_ specific change to the debouncer.

Wouldn't it be warranted to create a new component for exactly this purpose? Combining it with the old code is rather confusing. I'm not sure anybody understands why there is a motion type involved in debouncing. Even though I understand what you are trying to achieve here, I'm sure I don't understand why a "normal" debouncer would have a motion type input. Also, why only compare it to <i>one</i> motion type? Wouldn't you want a component to be more abstract and able to trigger on a user-defined motion type?

That is why it probably would be better to separate the functionality into a component for itself with the proper documentation why it exists and how it works.

andypugh:
I feel that this change does not belong in dbounce. 
The same result can be achieved by bypassing dbounce in HAL when motion-type == probing, using a mux to switch between debounced and not-debounced inputs into motion.probe-in.

(You can, of course, retain your special debounce on your own machine)

djdelorie:
I already have a mux on my probes, but I still need to properly debounce the probe during a probing cycle, because it still bounces then.  dbounce is not a debouncing module, it's a low-pass filter.  My patch changes the way it debounces, it doesn't disable it.  I need a proper debounce for probing - one that returns the original signal but filters out the bounce that follows.

PR #4056, opened 2026-05-25
ci: cancel in-flight runs on PR push

grandixximo:
Concurrency cancel-in-progress, gated on pull_request events only:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

A fresh push to a PR cancels the prior in-flight run for that ref. `master`, release tags, and `check_suite` triggers are left alone, so a tagged release or a manually-rerequested check-suite never cancels itself.

Saves wasted runner minutes on force-pushes.

The DEB_BUILD_OPTIONS commit from the earlier revision was dropped: CI logs already show `make -j4`, so `debuild` was auto-parallelizing. No measurable gain, and `.github/scripts/*` does not run for local `dpkg-buildpackage` invocations either, so the "local build" justification did not hold up.

## Test plan

- [x] After merge, push a second commit to a PR and confirm the older run gets cancelled.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,10 @@ on:
 permissions:
   contents: read #  to fetch code (actions/checkout)
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 env:
   DEBIAN_FRONTEND: noninteractive
 


grandixximo:
Follow-up on the timing claim: I compared per-job times for this PR vs a recent master run on the same 4-CPU runners, and there is no measurable gain. All deltas are within noise.

Reason: master logs already show `cd src && make -j4 -O <target>` for every step. `debuild` on CI auto-sets parallel jobs, so `dh_auto_build` was already fanning out without the explicit `DEB_BUILD_OPTIONS`. My local 32=>7 min measurement was apples-to-oranges (local invocation path that did not auto-parallelize).

Still worth keeping the change:

1. Helps local builds via `dpkg-buildpackage` or `debian/rules binary` directly, where parallel is not auto-enabled.
2. Documents intent so a future reader does not need to know `debuild`'s implicit defaults.
3. Two lines, zero risk.

I will update the commit message and PR description to drop the "~3x on CI" claim and reframe as a local-build / explicit-intent fix. The concurrency cancel commit stands on its own and is the real CI win.

BsAtHome:
> Follow-up on the timing claim: I compared per-job times for this PR vs a recent master run on the same 4-CPU runners, and there is no measurable gain. All deltas are within noise.

Then there is no real reason to keep it.

The only potential gain is aborting the old CI when you push on top of your tree. I'm not sure that is what we want when master merges occur. For branches, yes, that is mostly obvious. However, to master, it may not be nice.

> Still worth keeping the change:
>     1. Helps local builds via `dpkg-buildpackage` or `debian/rules binary` directly, where parallel is not auto-enabled.

But the changes are in `.github/*` and therefore irrelevant for local builds.

Changing it explicitly in the rules may actually break things. IIRC, these options are set on/by build-bots to do parallel builds.

grandixximo:
You are right on the DEB_BUILD_OPTIONS commit. Local builds run `dpkg-buildpackage` directly, not the `.github/scripts/*` wrappers, so the "helps local builds" framing was wrong. Dropped that commit, force-pushed.

On the master cancel concern: `cancel-in-progress` is gated on the event:

```yaml
cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

So master pushes, release tags, and `check_suite` reruns never trigger cancellation. Only PR pushes do. PR is now concurrency-only.

BsAtHome:
We're going to try and see. We can always revert.

PR #4057, opened 2026-05-26
docs: re-export Architecture C1 SVG with light theme only

grandixximo:
## Summary

The `LCNC_Architecture_C1.drawio.svg` diagram (under [docs/src/getting-started/images/](docs/src/getting-started/images/)) was previously exported by draw.io with `color-scheme: light dark` on the root `<svg>` and `light-dark(LIGHT, DARK)` wrappers on every fill/stroke/text color. On a system in dark mode the browser then swaps the canvas to dark and switches all per-element colors to their dark variants, so the white-fill boxes turn into very dark boxes on a near-black canvas and the diagram is essentially unreadable.

The classic LinuxCNC HTML docs have no dark theme, so the diagram should always render in light regardless of system preference.

This PR re-exports the SVG with draw.io's light theme forced:
- root style is now `background: #ffffff` with `color-scheme: light` (no `dark`),
- no `light-dark()` wrappers anywhere in the file,
- the source `.drawio` is unchanged.

No other files touched.

this is the original look

image

New look

image

## Test plan

- [x] Build docs with `make docs` and open `docs/html/getting-started/about-linuxcnc.html` in a browser configured for dark mode; verify the architecture diagram renders white-background with readable boxes.
- [x] Same check with the browser in light mode (no regression).


BsAtHome:
This SVG file is of very very terrible quality when you look at the XML. The original is just ad bad.

Text is defined as a foreign (html) object and not as text. And then the foreign object is aliased with an image if you can't render the foreign html object. Editing in Inkscape only gives you access to the (fixed DPI) image and not the underlying text.

One think can be that you look at the export options and see if there is a better way to export to SVG. The alternative is to use my attached version here. It is "only" 26.5 kByte (the fixed source you suggest weighs in at about 500 kByte).

Flattening the text with 5 decimal places makes the file ~300 kByte (and inserts the correct aria-label markers for the text). I guess it doesn't matter for the PDF version (text uses Noto Sans, which should get embedded). The web-version may require to be flat to prevent rendering mismatches on different systems.

Image small with text as text:
LCNC_Architecture_C1 inkscape+optimized

Image with flattened text (rendered as paths):
LCNC_Architecture_C1 inkscape+flattened

grandixximo:
if I export with every option disabled in draw.io I get an svg that I can modify in inkscape, with a few double clicks, but the svg ugliness remains. I'll swap with your svg for now, need make a doc about SVG creation, going trough draw.io and then incscape to do cleanup?

BsAtHome:
I think we should drop the drawio version after I've seen what it produced. We can do a lot better in Inkscape than anything from drawio... There were many a drawing detail that I fixed too in my version.
If you add the flattened text version, it needs a small update to add stroke:none on the enclosing group. The default seems to be stroking, which I missed when I uploaded it. However, I suggest to use the smaller version for now.

grandixximo:
Pusdhed smaller version, ready for merge

PR #4066, opened 2026-05-29
build: fail configure if libcap-dev is missing

### LINKED ISSUE #4065: runtime error on recent compiled master
I recently pulled in the latest in master.
I fist got error while compiling:

rtapi/uspace_rtapi_main.cc:49:10: fatal error: sys/capability.h: No such file or directory
   49 | #include <sys/capability.h>
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.

I then  sudo apt-get install libcap-dev to fix that.
It compiled fine but when I run anything I get an error.

error message in terminal:

Machine configuration file is 'qtdragon_inch.ini'
  File "/home/chris/emc/bin/linuxcnc_check_ini", line 25, in <module>
    import linuxcnc
ImportError: /home/chris/emc/lib/python/linuxcnc.so: undefined symbol: _ZN8linuxcnc7IniFile14mapLinearUnitsERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
linuxcnc_check_ini validation failed

I use mint 22.3 (with no problem till now)

Any idea what to do to fix this?

BsAtHome:
Fixes #4065. The rootless implementation uses libcap and the dev package needs to be installed. This patch will fail configure if the sys/capability.h header is missing.

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1544,6 +1544,9 @@ if test "x$HAVE_LIBFMT" = "xno"; then
     AC_MSG_ERROR([libfmt header not found. Install with 'sudo apt-get install libfmt-dev'])
 fi
 
+AC_CHECK_HEADERS([sys/capability.h], [], [
+    AC_MSG_ERROR([libcap header not found. Please install libcap-dev])
+])
 AC_CHECK_HEADERS([editline/readline.h histedit.h], [], [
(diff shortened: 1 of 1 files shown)


grandixximo:
My bad...

BsAtHome:
> My bad...

We all fail, regularly. We'd be almighty if we didn't ;-)

PR #4074, opened 2026-05-30
libnml: gate CMS_DISPLAY_ASCII_UPDATER banner behind debug flag

grandixximo:
The CMS_DISPLAY_ASCII_UPDATER constructor printed a multi-line WARNING
banner unconditionally to stderr. That updater is also created transiently
by NML::msg2str()/str2msg() to render an xdr buffer as an ASCII string, so
the banner fired during normal operation for a harmless use.

Route it through rcs_print_debug(PRINT_CMS_CONSTRUCTORS, ...) so it is
silent by default but still available with CMS constructor debug printing.

Complements the touchy DEBUG=0 change by fixing the noise at the source
for any config that legitimately uses the ASCII updater.

Surfaced in the ui-smoke review, PR #4054.

Test: touchy sim with DEBUG = 0x10 (instantiates the updater) under xvfb,
banner no longer printed.

### DIFF
--- a/src/libnml/cms/cms_dup.cc
+++ b/src/libnml/cms/cms_dup.cc
@@ -32,7 +32,12 @@
     CMS_DISPLAY_ASCII_UPDATER::CMS_DISPLAY_ASCII_UPDATER(CMS * _cms_parent):
 CMS_UPDATER(_cms_parent)
 {
-    fprintf(stderr,
+    // This updater is also used transiently by NML::msg2str()/str2msg() for
+    // ASCII string conversion of otherwise xdr-encoded buffers (e.g. task
+    // command logging under EMC_DEBUG_TASK_ISSUE), so the banner fired on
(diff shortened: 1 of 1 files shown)


BsAtHome:
Converting to and from strings with the ascii updaters generate values that are artificially truncated. They may produce results that are not reflecting the actual state of affairs. That is why the banner exists and the warning is made so explicit. So hiding the message is probably not what you want. Once you instantiate an ascii updater class, you can get into serious trouble without knowing it when some process or person starts using its data.

I suppose that the ascii updater is instantiated when debug is enabled and some debug messages use msg2str. In that case, just setting debug to zero should suffice. Or are there other instances where this might happen?

grandixximo:
Yes, there's another instance (any `ascii`-encoded buffer via `CMS::open`), but it's not in any shipping config, and that path is precisely where the truncation warning is _most_ warranted. But that just makes you more correct, the banner should stay loud.

Closing #4074, keeping #4073

PR #4053, opened 2026-05-24
docs: migrate documentation toolchain from asciidoc-py + dblatex to asciidoctor

grandixximo:
## Summary

Replaces the docs toolchain: asciidoc-py + dblatex + xsltproc + source-highlight + inkscape are dropped, build now goes through `asciidoctor` + `asciidoctor-pdf` + `rouge`. Motivation as discussed on #4051: asciidoc-py is EOL, dblatex unmaintained, asciidoctor is actively packaged in Debian and uses prawn-svg natively (no inkscape detour). Continues hansu's earlier asciidoctor branch and solves the cross-document anchor problem that stalled it.

## What changed (10 commits, oldest first)

1. **`docs: add Rouge syntax highlighters for HAL, NGC, INI`** — three custom Rouge lexers (rouge_hal/ngc/ini.rb) loaded by asciidoctor via `-r`. NGC handles LinuxCNC `#<_ini[...]NAME>` / `#<_hal[...]>` references.

2. **`docs: add asciidoctor extensions for image lookup and cross-doc xrefs`** — `image_resolver.rb` (treeprocessor: per-lang tree, EN fallback, `foo_en.ext` -> `foo_<lang>.ext` substitution, default `pdfwidth=75%`); `xref_resolver.rb` (preprocessor: bare `<<anchor,Title>>` -> qualified `<<relpath/file.adoc#anchor,Title>>`, anchor index cached on disk by mtime).

3. **`docs: HTML/PDF theme CSS overrides and docinfo`** — `lcnc-overrides.css` (heading `#1a3a6c` incl. `#header h1`, prefers-color-scheme dark, no JS), `pdf-theme.yml` (asciidoctor-pdf theme: A4, dblatex-blue headings, top header `doc-title | chapter | page/total`), `docinfo.html`, static-page CSS refresh.

4. **`docs: portable CJK font (CFF -> TTF) converter`** — `otf2ttf.py` builds a TrueType (glyf) copy of Noto Serif CJK SC from the system .ttc. prawn 2.4 corrupts CFF embeds. Subsets to glyphs actually used in docs + .po catalogues (~1.5s vs ~200s for full face).

5. **`build: probe optional docs deps; graceful disable, hard-fail explicit`** — `configure.ac` probes ruby-rouge, python3-fonttools, NotoSerifCJK fontconfig. Warn-and-disable when implicit; hard-fail for explicit `--enable-build-documentation=pdf`. `debian/configure` + `control.top.in` swap build-deps to asciidoctor stack.

[...]

PR #4081, opened 2026-05-30
docs(build): unified output subtree under docs/build/

grandixximo:
## Summary

Move all rendered docs artefacts (HTML, PDF, po4a-translated `.adoc`) under a single `docs/build/` tree, drop asciidoc-py-era leftovers in `docs/html/`, and stop leaking build-time artefacts into `docs/src/`.

Builds on top of #4053 (asciidoctor migration). Implements Bertho's layout proposal agreed in the #4053 discussion.

## Layout

```
docs/build/html/index.html           static redirect -> en/
docs/build/html/{shared CSS, logo}
docs/build/html/<lang>/index.html
docs/build/html/<lang>/<topic>/<page>.html
docs/build/html/<lang>/man/manN/<page>.html
docs/build/html/<lang>/pdf/LinuxCNC_*[_<lang>].pdf
docs/build/adoc/<lang>/              po4a-translated source
```

English now lives under `en/` like every other language, so the top-level redirect lands users on the right tree.

Manpage troff stays at `docs/man/` because halcompile output, `install-man`, `mandb`, and `rip-environment` read it.

## Source-tree cleanup

Three build-time artefacts no longer leak into `docs/src/`:

| Was generated under | Now generated under |
|---|---|
| `docs/src/man/man1/linuxcnc.1.adoc` (config.status output) | `docs/build/adoc/en/man/man1/linuxcnc.1.adoc` |
| `docs/src/drivers/mb2hal_HOWTO.ini` (include fixture) | `docs/build/adoc/en/drivers/mb2hal_HOWTO.ini` |
| `docs/src/gui/gmoccapy_release_notes.txt` | dropped (was dead under new layout) |

`mb2hal.adoc`'s `include::` now uses the `{fixturedir}` attribute, set per-language by the asciidoctor HTML rule. The 8 `.po` files are repointed at the new include path (mechanical msgid edit, no po4a regen).

The manpage HTML rule grows a build-tree fallback so the English manpage build finds `linuxcnc.1.adoc`; the `LinuxCNC_Manual_Pages.pdf` master generator grows the same fallback. An explicit troff rule for `linuxcnc.1` replaces the generic pattern (which expects sources under `docs/src/man/`).

Static assets in the former output dir `docs/html/` move to `docs/src/` (`gcode.html`, `index.css`, logo). asciidoc-py-era files (`asciidoc.css`, `linuxcnc.css`, `minus.png`, `plus.png`) are deleted.

`debian/*.docs` and `docs/po4a.cfg` are repointed at the new tree. Install destinations under `usr/share/doc/` are unchanged.

## Test plan

- [x] `cd src && make -j$(nproc)` clean full build
- [x] English HTML + 5 Master PDFs + Manual_Pages.pdf
- [x] All 8 languages (en, de, es, fr, nb, ru, uk, zh_CN) build HTML + 4 Master PDFs each
- [x] `mb2hal.html` contains the example INI fixture in all 8 languages
- [x] `LinuxCNC_Documentation.pdf` contains the mb2hal INI fixture
- [x] `linuxcnc(1)` troff renders to `docs/man/man1/linuxcnc.1`
- [x] `linuxcnc(1)` HTML and PDF (in `LinuxCNC_Manual_Pages.pdf`) both contain the manpage body
- [x] `make docclean` leaves no generated `.adoc` / `.ini` / `.txt` in `docs/src/`
- [x] `git clean -dxn docs/src/` reports only `hal/components_gen.adoc` (pre-existing, generated by `gen_complist.py`, not from this PR)


grandixximo:
@hansu @BsAtHome See if this is the right shape, the submake file since I'm getting blame almost everywhere, I fixed everything to tabs just to make it mostly consistent, adoc folder naming is sensible?

grandixximo:
reverting to draft till it builds on sid

hansu:
I tested your branch just for the english language.

1. I get 4 times `sed: can't read ../docs/src/drivers/{fixturedir}/drivers/mb2hal_HOWTO.ini: No such file or directory`, but the file is correctly included.
2. The PDF docs are at  `docs/build/html/<lang>/pdf/` but should be in  `docs/build/pdf/<lang>/` 
3. There are a lot of `gen_complist: Broken link: ../man/man1/axis.1.html`, but the links seem to work
  except that one: `gen_complist: Broken link: ../man/man9/homecomp.9.html`. The HTML man page of homecomp.9 isn't build.

Another thing we may could do in this rush: there is a folder `help`  in `docs`  which would better fit in `docs/src,` right?
```
docs/help
├── asciidoc-markup_es.adoc
├── rtfaults.adoc
├── rtfaults_es.adoc
└── tklinuxcnc.adoc
```

BsAtHome:
Just discovered: Running make docs the second time invokes `../scripts/make-docs-pdf-index`. So it is not completely stable.

Anyway, this script needs to be changed when the reorganisation moves stuff to other places.

grandixximo:
Pushed commits should fix most issues reported, other than what's parallel in #4091 (homecomp.9 missing: the .comp is filtered out of COMPS so no manpage rule exists; #4091 adds it back alongside the halcompile --adoc extraction rework).

I can squash everything once done.

BsAtHome:
Build is stable now.

Next up is to review it all...

BsAtHome:
I think there is some whitespace problem here. Seems to happen for each language:
```
sed: can't read ../docs/build/adoc/<lang>/drivers//...myhome.../grandixximo/docs/build/adoc/<lang>/drivers/mb2hal_HOWTO.ini: No such file or directory
sed: can't read ../docs/build/adoc/<lang>/drivers//...myhome.../grandixximo/docs/build/adoc/<lang>/drivers/mb2hal_HOWTO.ini: No such file or directory
```

BsAtHome:
There are PDF build artifacts left in the docs/src directory.

The languages at the top of the master index.html does not match the list made in the po4a.cfg. The list must be generated but the `@TRANSLATIONS@` `@ENDTRANSLATIONS@` part in the index.tmpl does not seem to be regenerated from the po4a list.

The PDF tree is outside the HTML tree. That means we cannot link the PDFs in one tree. The index.html generated needs to be rewritten. We need one index.html per language and it must be linked in the master HTML index to find the PDFs.

A lot of the .gitignore files in the docs tree have become obsolete and should be removed entirely. The master .gitignore needs to be cleaned up too. I suggest that the ignores for the docs are actually in `docs/.gitignore` and not the master ignore file.

The man page tree still contains the cat[139] parts. I think we can get rid of the (old) catalog stuff. We only really need the man-page tree. That would also clean up some stuff in the makefiles again. FWIW, I don't really understand why `mandb` is invoked. This step is usually only for packages interesting and is done by dpkg and friends.

Another issue seen is:
```
Processing manual pages under /...myhome.../grandixximo/src/../docs/man/uk...
/usr/bin/mandb: warning: /...myhome.../grandixximo/src/../docs/man/uk/man1/pi500_vfd.1: whatis parse for pi500_vfd(1) failed
/usr/bin/mandb: warning: /...myhome.../grandixximo/src/../docs/man/uk/man1/thermistor.1: whatis parse for thermistor(1) failed
/usr/bin/mandb: warning: /...myhome.../grandixximo/src/../docs/man/uk/man1/wj200_vfd.1: whatis parse for wj200_vfd(1) failed
Updating index cache for path `/...myhome.../grandixximo/src/../docs/man/uk/man9'. Wait...
/usr/bin/mandb: warning: /...myhome.../grandixximo/src/../docs/man/uk/man9/abs_s32.9: whatis parse for abs_s32(9) failed
...
```
This only seems to happen for the Ukranian (uk) translation. There must be something in there that trips it and it is a consistent error.

(and probably more...)

grandixximo:
Thanks for the thorough review. Most of it I can just fix, one item needs a decision from you first.

PDF tree location: `docs/build/pdf/` is currently a sibling of `docs/build/html/`, so the HTML tree cannot link the PDFs with relative paths and does not zip up self-contained. You leaned earlier toward putting the PDFs under the HTML tree. Before I rewrite the index generation, which layout do you want: `build/html/<lang>/pdf/`, `build/html/pdf/*_<lang>.pdf`, or keep them separate and link by served URL? This also drives the per-language `index.html` work.

On the master index language list not matching po4a.cfg: that one is handled in the topbar/lang-switcher PR, which drops the hardcoded `@TRANSLATIONS@` block and builds the list from po4a.cfg, which is coming soon...

BsAtHome:
> PDF tree location: `docs/build/pdf/` is currently a sibling of `docs/build/html/`, so the HTML tree cannot link the PDFs with relative paths and does not zip up self-contained. You leaned earlier toward putting the PDFs under the HTML tree. Before I rewrite the index generation, which layout do you want: `build/html/<lang>/pdf/`, `build/html/pdf/*_<lang>.pdf`, or keep them separate and link by served URL? This also drives the per-language `index.html` work.

For consistency, all PDFs should IMO be called `bla-name-bla_<lang>.pdf`, regardless where they are placed. This makes a browser downloaded file's name always reflect its language and multiple languages can exist side by side when downloaded.

I think the target directory `build/html/<lang>/pdf/` is most consistent so that all links are language prefixed in the same manner. When you switch language you don't address the individual PDFs but the index.html in that directory (well, just address the directory and let the server determine what to serve).

> On the master index language list not matching po4a.cfg: that one is handled in the topbar/lang-switcher PR, which drops the hardcoded `@TRANSLATIONS@` block and builds the list from po4a.cfg.

Check, great!

grandixximo:
Pushed a batch addressing the review.

**PDF layout (your call):** PDFs now build into `docs/build/html/<lang>/pdf/`, each suffixed `_<lang>` (English `_en`) for downloads. Package filenames stay unchanged: `install-doc-pdf` strips `_en` on install, so the launchers and doc-base still resolve. I kept a flat aggregate at `docs/build/html/pdf/` (hardlinks + listing) so the current `/pdf/` URL keeps working; it now lists all languages. That aggregate and `make-docs-pdf-index` are transitional and can go once the per-lang `index.html` links the PDFs and the infra repoints `/documents/`. Verified with a full 8-language build.

Also in this push: PDF artifacts out of `docs/src`, the build-time `mandb` step dropped (takes the `cat[139]`/`index.db` cruft and the `whatis` noise with it), `.gitignore` consolidated into `docs/.gitignore`, and the doubled `{fixturedir}` include path fixed.

Deferred to the topbar PR: the lang-list/`@TRANSLATIONS@` and the per-language PDF index linking.

BsAtHome:
The real problem here is that you updated the .pot/.po outside weblate. That will give serious trouble.

The way is to change the source file, then goto weblate and fix all the translations. Then, probably @smoe, should perform a pot/po update to make things consistent. It sucks this way, but otherwise we'll have a merge conflict on translations. Maybe @smoe can chime in here on the preferred sequence?

docs/.gitignore
```diff
@@ -1,2 +1,54 @@
...
+# Rendered output: HTML, PDF, and po4a-translated .adoc all land under build/.
+/build/
+
+# Intermediate .html asciidoctor writes beside each source .adoc before the
+# copy step moves it into build/html/.
+/src/*/*.html
+
+# source-highlight local cache.
+/src/source-highlight/local
+
```
BsAtHome: Source-highlight was removed and you are using rouge. All the source-highlight stuff is obsolete.

smoe:
You have not merged anything against master yet Weblate, right? Weblate is not in a good state at the moment. I suggest I bring this back to normal and independently you remove anything on the .pot/.po from your branch. My biggest concern is that we lose externally contributed translations. Everything else is fine with me.

docs/.gitignore
```diff
@@ -1,2 +1,54 @@
...
+# Generated component manpage list (gen_complist.py).
+/src/hal/components_gen.adoc
+
```
BsAtHome, line 13: Wouldn't it be better to run po4a on the `build/adoc/en` tree and not have to generate the `components_get.adoc` in the src tree?

docs/.gitignore
```diff
@@ -1,2 +1,54 @@
...
+# SVGs rendered from the .dot sources in src/code/ (CMS_buffer.svg is committed).
+/src/code/homing.svg
+/src/code/homing_es.svg
+/src/code/hss.svg
+/src/code/hss_es.svg
+/src/code/task-state-transitions.svg
+/src/code/task-state-transitions_es.svg
+
```
BsAtHome: These should be generated into the build tree.

We should keep the docs/src tree clean from anything generated for as much as possible.

docs/.gitignore
```diff
@@ -1,2 +1,54 @@
...
+/man/pt_BR/*
+/man/ro/*
+/man/ru/*
+/man/sai/*
+/man/sk/*
+/man/sv/*
+/man/tr/*
+/man/uk/*
+/man/vi/*
+/man/zh_CN/*
```
BsAtHome: This should now all be under build.

docs/.gitignore
```diff
@@ -1,2 +1,54 @@
...
+# All ignores for the documentation tree live here (not in the top-level
+# .gitignore).  Paths are relative to docs/.
+
+# Rendered output: HTML, PDF, and po4a-translated .adoc all land under build/.
+/build/
+
+# Intermediate .html asciidoctor writes beside each source .adoc before the
+# copy step moves it into build/html/.
+/src/*/*.html
+
```
BsAtHome: Why have these a leading `/`? Using relative paths is a thing ;-)

(other places too)

docs/.gitignore
```diff
@@ -1,2 +1,54 @@
...
+/src/code/homing_es.svg
+/src/code/hss.svg
+/src/code/hss_es.svg
+/src/code/task-state-transitions.svg
+/src/code/task-state-transitions_es.svg
+
+# Troff manpages generated from .adoc by the asciidoctor manpage backend.
+/man/man1/
+/man/man3/
+/man/man9/
```
BsAtHome: Aren't they be generated into the right place now?

docs/src/asciideps
```diff
@@ -5,10 +5,29 @@ set -e
...
+# Substitute asciidoctor attributes the Submakefile passes by env so include
+# paths like `include::{fixturedir}/drivers/mb2hal_HOWTO.ini[]` resolve to a
+# real file rather than a literal `{fixturedir}` string sed cannot read.
+subst_attrs() {
+	if [ -n "$FIXTUREDIR" ]; then
+		sed -e "s|{fixturedir}|$FIXTUREDIR|g"
+	else
+		cat
+	fi
+}
```
BsAtHome: I'm not sure that using the `{fixture}` way is the right fix.

docs/src/drivers/mb2hal.adoc
```diff
@@ -147,7 +147,7 @@ Click link:mb2hal_HOWTO.ini[here] to download.
 
 [source,ini]
 ----
-include::mb2hal_HOWTO.ini[]
+include::{fixturedir}/drivers/mb2hal_HOWTO.ini[]
 ----
```
BsAtHome, line 151: This document is never translated. It is a verbatim version from the sources. It is doubtful that you want it to be translated. If you want translations, then the document including this should do a better job explaining things in a more granular way instead of dropping in a big INI file. It is also in the wrong place and the docs/src/Submakefile hacks the document (and also gmoccapy release notes) to be copied somewhere else.

It is IMO an abomination to copy sources from the code-tree into the docs tree. It is either documentation or source, not both.

Placing this stuff the right place inlined will also make it easier and you don't need to "{fixture}" hacks and the pot/po problem vanishes instantly.

(Also, the release notes are not translatable)

BsAtHome:
> You have not merged anything against master yet Weblate, right?

No, and if it is up to me we shouldn't touch the pot/po in normal PRs at all. And in this case here it is actually not necessary.

> Weblate is not in a good state at the moment. I suggest I bring this back to normal and independently you remove anything on the .pot/.po from your branch. My biggest concern is that we lose externally contributed translations. Everything else is fine with me.

My PR comments already suggest to remove all these pot/po changes so we keep the status quo.

hansu:
Hmm I get this, but doesn't seem to be a problem on the CI:

```
/bin/bash: line 1: ../scripts/make-docs-pdf-index: Permission denied
make: *** [../docs/src/Submakefile:457: ../docs/build/html/pdf/index.html] Error 126
```

hansu:
>It is IMO an abomination to copy sources from the code-tree into the docs tree. It is either documentation or source, not both.
>
>Placing this stuff the right place inlined will also make it easier and you don't need to "{fixture}" hacks and the pot/po problem vanishes instantly.
>
>(Also, the release notes are not translatable)

We can move the `mb2hal_HOWTO.ini` to the docs tree, no problem. That would make life easier :)
For the gmoccapy release notes it is fine for me too (even though it's handy to have it close to the sources). Also ok for you @gmoccapy?

grandixximo:
Thanks, this is all addressed or routed.

**`mb2hal_HOWTO.ini` / `{fixturedir}`:** done your way. Moved the file from `src/hal/user_comps/mb2hal/` into `docs/src/drivers/` next to `mb2hal.adoc`, made the include plain (`include::mb2hal_HOWTO.ini[]`), and removed the `{fixturedir}` attribute and the asciideps substitution entirely (asciideps is back to its original form). Since the include string is unchanged from master, the pot/po stay at status quo, no weblate churn. The per-language copy into the build trees now sources from the docs tree, not the code tree. (gmoccapy's `release_notes.txt` left for now, pending its own move.)

**`docs/.gitignore`:** dropped the obsolete `source-highlight` entry and switched to relative paths (no leading `/`).

On "keep `docs/src`/`docs/man` clean, generate into `build/`": agreed in principle. I'd rather not grow this PR with it, so I'm routing the three pieces to focused follow-ups:
- `components_gen.adoc` generated into `build/adoc/en` instead of `src/hal/`
- the `.dot`-rendered SVGs into `build/` instead of `src/code/`
- the whole `man/` tree under `build/`, which is the big one (touches `install-man`, the html man rule, debian, #4091, and the translated-manpages branch that currently also writes into `docs/man/`), so it gets its own PR and the translated-manpages work rebases onto it.

docs/.gitignore
grandixximo: gone

docs/.gitignore
grandixximo: removed leading /

docs/.gitignore
grandixximo: Agreed it's cleaner. The catch: `components.adoc` includes `components_gen.adoc` with a plain relative include, so they must sit in the same dir. To generate it into `build/adoc/en/hal/`, English has to render from `build/adoc/en/` like the translations do, otherwise the src-tree `components.adoc` can't resolve the include without a `{fixturedir}`-style hack.

That's a deeper change (the `en` pipeline is special-cased throughout the Submakefile), so I'd do it as its own follow-up. Two ways to populate `build/adoc/en/`:

1. Add `en` to po4a as an identity passthrough (empty `en.po`, `--keep 0`). Reuses po4a's tree generation, but touches `po/`.
2. Copy the English `.adoc` + images into `build/adoc/en/` and point the render rules there. No `po/` touch.

Which do you prefer? The same move then lets the `.dot` SVGs generate into `build/` too.

docs/.gitignore
grandixximo: see above

docs/.gitignore
grandixximo: Not this PR. It's the coming, dedicated man-relocation PR.

docs/.gitignore
grandixximo: Same boat, same coming PR.

docs/src/asciideps
grandixximo: gone, will avoid this hack.

docs/src/drivers/mb2hal.adoc
grandixximo: Agreed, moved.

docs/.gitignore
grandixximo: https://github.com/grandixximo/linuxcnc/compare/docs/unified-build-output...docs/render-en-from-build

Went for no `po/` touching

Expand scope here? Wait for follow up PR?

grandixximo:
@BsAtHome this is my main blocker for follow up PRs on the docs, if this merges I'll send those tomorrow, night night for now

BsAtHome:
There are still some corners that need some smoothing, but those can wait and come in later.

hansu:
I get a lot of those, also in the log of the CI: https://github.com/LinuxCNC/linuxcnc/actions/runs/26826775053/job/79096405966#step:5:20548

```
Use of uninitialized value $_ in pattern match (m//) at /usr/bin/checklink line 1296.
Use of uninitialized value $_ in pattern match (m//) at /usr/bin/checklink line 1296.
Use of uninitialized value $_ in pattern match (m//) at /usr/bin/checklink line 1296.
Use of uninitialized value $_ in pattern match (m//) at /usr/bin/checklink line 1296.
***
*** warning: bad links found in English docs!
*** oh well, continuing anyway
***
Use of uninitialized value $_ in pattern match (m//) at /usr/bin/checklink line 1296.
Use of uninitialized value $_ in pattern match (m//) at /usr/bin/checklink line 1296.
```

hansu:
Where are the docs now located? The link to the 2.10 docs "http://linuxcnc.org/docs/devel/html/" from https://linuxcnc.org/documents/ does not work any more.

BsAtHome:
That is a good question and depends on how they are pushed, pulled or referenced.
Probably needs some update and a fix on the homepage (https://github.com/LinuxCNC/wlo/blob/master/documents/index.md).

hansu:
@andypugh do you know how the docs are copied to the Linuxcnc website?

grandixximo:
@SebKuzminsky @AndyPugh

After #4081 the generated docs tree moved from `docs/html/` to `docs/build/html/`, so `http://linuxcnc.org/docs/devel/html/` 404s, the server-side deploy still pulls the old path. This publish runs as `emcboard@www.linuxcnc.org` and is not in any repo I can find (wlo's `pull-build-rsync` / `githook.post-receive` only deploy the Jekyll `_site/`), so it needs one of you.

New build layout:
- `docs/build/html/index.html` is a master language picker, so `/docs/devel/html/` resolves.
- English under `docs/build/html/en/`, translations under `docs/build/html/<lang>/`.
- PDFs under `docs/build/html/pdf/` (flat aggregate, each suffixed `_<lang>`) plus per-language `docs/build/html/<lang>/pdf/`. No more `docs/pdf/` sibling.

Suggested deploy change:
- repoint the docs rsync source `docs/html/` -> `docs/build/html/`, target `/docs/devel/html/`.
- old `/docs/devel/pdf` URL: redirect to `/docs/devel/html/pdf/` or drop it. I have a one-line homepage patch ready that repoints the `documents/` PDF link there, will open it against wlo once the served layout is settled.
- same repoint applies to `stable` once 2.10 ships with this layout.

Issue #4029, opened 2026-05-15
Discussion: Remove rtapi_get_clocks()


hdiethelm:
Hello

I was wondering why the .time value is so big in Linuxcnc. Probably this is documented somewhere but this value is not in ns but in arbitrary clocks returned by rtapi_get_clocks().

According to the documentation, rtapi_get_time() is really slow sometimes. So I created a component to test this:
https://github.com/hdiethelm/linuxcnc-fork/tree/timetest_component

It is not that slow according to my tests and the issues having with rtapi_get_clocks() returning random time units outweigh the 16ns rtapi_get_time() takes more in my opinion.

It might have been slow long time ago, the code is like 20 years old: https://github.com/LinuxCNC/linuxcnc/commit/ed4089d02a7b1201265225a10fb3e0cce6a35f5f It is still an bit an issue on RTAI and it might still be an issue on certain systems. More tests would be good.

Results so far, Debian Trixie:
- VM, PREEMT_RT
  - Clock to time ratio: 4.3
  - rtapi_get_clocks(): 20ns
  - rtapi_get_time(): 46ns
- CNC PC, PREEMT_RT
  - Clock to time ratio: 3.2
  - rtapi_get_clocks(): 8ns
  - rtapi_get_time(): 20ns
- CNC PC, RTAI
  - Clock to time ratio: 3.2
  - rtapi_get_clocks(): 11ns
  - rtapi_get_time(): 112ns
- CNC PC, Xenomai3
  - Clock to time ratio: 3.2
  - rtapi_get_clocks(): 8ns
  - rtapi_get_time(): 24ns
- CNC PC, Xenomai4
  - Clock to time ratio: 3.2
  - rtapi_get_clocks(): 8ns
  - rtapi_get_time(): 22ns

Started using:
`scripts/halrun -I -f timetest.hal`

[timetest.txt](https://github.com/user-attachments/files/27794671/timetest.txt)

I created a prototype removing this function all together: https://github.com/hdiethelm/linuxcnc-fork/tree/rtapi_get_clocks_removal

While doing so, I even discovered issues where the time to clock conversion is wrong. For example here:
https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/drivers/hal_gpio.c#L367
This assumes one clock is 1 ns. However, on my PC, one clock is 0.24 ns

[...]

PR #4082, opened 2026-05-30
Remove rtapi_get_clocks()

hdiethelm:
As discussed in https://github.com/LinuxCNC/linuxcnc/issues/4029, I think it would make sense to remove `rtapi_get_clocks()`. 

An optimized function using rdtsc() and calculating the ration to ns would be possible, but I see no point in over complicating things due to my tests showed next to no impact.

On my setup, I don't see any change before and after this PR. Tested with:
Debian Trixie / i5-6500 / Latest RT kernel

Both latency_test and linuxcnc with my CNC show the exact same CPU load (+-0.1%) and timing behavior.

This PR also fixes a few bugs where the CPU frequency was assumed to be 1GHz, so some timings where applied wrongly.

Additionally, all thread.time values are now in ns and not an arbitrary clock count.

Open points:

- ~~Test with RTAI: My PC is unstable with RTAI, so I can not test this.~~
- ~~Docs: I updated the man page. But I do not really know how to manage all translation files. Can someone help me or show me the tools needed to do that? I can do English / German.~~
- ~~Test on different systems~~
  - Only x86/amd64, all other systems never used this functionality: https://github.com/LinuxCNC/linuxcnc/blob/master/src/rtapi/uspace_common.h#L291
 - Test hal_gpio / hal_parport: I have no parport on my PC. Can someone  do this?

### DIFF
--- a/src/emc/motion/motion.c
+++ b/src/emc/motion/motion.c
@@ -593,9 +593,6 @@ static int init_hal_io(void)
 
     // export timing related HAL pins so they can be scoped and/or connected
     CALL_CHECK(hal_pin_u32_newf(HAL_OUT, &(emcmot_hal_data->last_period), mot_comp_id, "motion.servo.last-period"));
-#ifdef HAVE_CPU_KHZ
-    CALL_CHECK(hal_pin_float_newf(HAL_OUT, &(emcmot_hal_data->last_period_ns), mot_comp_id, "motion.servo.last-period-ns"));
-#endif
 
     // export timing related HAL pins so they can be scoped
     CALL_CHECK(hal_pin_float_newf(HAL_OUT, &(emcmot_hal_data->tooloffset_x), mot_comp_id, "motion.tooloffset.x"));
--- a/docs/src/config/core-components.adoc
+++ b/docs/src/config/core-components.adoc
@@ -170,10 +170,8 @@ change or removal at any time.
 * 'motion.debug-float-3' - (float, RO) This is used for debugging purposes.
 * 'motion.debug-s32-0' - (s32, RO) This is used for debugging purposes.
 * 'motion.debug-s32-1' - (s32, RO) This is used for debugging purposes.
-* 'motion.servo.last-period' - (u32, RO) The number of CPU cycles between invocations of the servo thread.
-  Typically, this number divided by the CPU speed gives the time in seconds,
-  and can be used to determine whether the realtime motion controller is meeting its timing constraints
-* 'motion.servo.last-period-ns' - (float, RO)
+* 'motion.servo.last-period' - (u32, RO) The time in ns between invocations of the servo thread.
+  This number can be used to determine whether the realtime motion controller is meeting its timing constraints
 
 === Functions
 
--- a/docs/src/hal/basic-hal.adoc
+++ b/docs/src/hal/basic-hal.adoc
@@ -373,12 +373,13 @@ If you used the Stepper Config Wizard to generate your config you will have up t
 == HAL Parameter
 
 (((HAL Parameters)))
-Two parameters are automatically added to each HAL component when it is created.
-These parameters allow you to scope the execution time of a component.
+One pin and two parameters are automatically added to each HAL component when it is created.
+These allow you to scope the execution time of a component.
 
 [horizontal]
-`.time`(((HAL time))):: Time is the number of CPU cycles it took to execute the function.
-`.tmax`(((HAL tmax))):: Tmax is the maximum number of CPU cycles it took to execute the function.
+`.time`(((HAL time))):: Pin time shows in ns how long it took to execute the function.
+`.tmax`(((HAL tmax))):: Parameter tmax is the maximum time in ns it took to execute the function.
+`.tmax-increased`(((HAL tmax-increased))):: This parameter is set to true for one cycle if tmax increased.
 
 `tmax` is a read/write parameter so the user can set it to 0 to get rid of the first time initialization on the function's execution time.
 
--- a/docs/src/hal/components.adoc
+++ b/docs/src/hal/components.adoc
@@ -511,7 +511,6 @@ rtapi_clock_set_period.3
 rtapi_delay.3
 rtapi_delay_max.3
 rtapi_exit.3
-rtapi_get_clocks.3
 rtapi_get_msg_level.3
 rtapi_get_time.3
 rtapi_inb.3
--- a/docs/src/man/man3/rtapi_get_time.3.adoc
+++ b/docs/src/man/man3/rtapi_get_time.3.adoc
@@ -4,14 +4,13 @@
 
 == NAME
 
-rtapi_get_time, rtapi_get_clocks - get the current time
+rtapi_get_time - get the current time in ns
 
 == SYNTAX
 
 [source,c]
 ----
 long long rtapi_get_time();
-long long rtapi_get_clocks();
 ----
 
 == DESCRIPTION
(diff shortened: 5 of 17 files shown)


src/emc/motion/motion.c
```diff
@@ -593,9 +593,6 @@ static int init_hal_io(void)
 
     // export timing related HAL pins so they can be scoped and/or connected
     CALL_CHECK(hal_pin_u32_newf(HAL_OUT, &(emcmot_hal_data->last_period), mot_comp_id, "motion.servo.last-period"));
-#ifdef HAVE_CPU_KHZ
-    CALL_CHECK(hal_pin_float_newf(HAL_OUT, &(emcmot_hal_data->last_period_ns), mot_comp_id, "motion.servo.last-period-ns"));
-#endif
```
BsAtHome, line 598: Removing this could be a problem. Or, is nobody using it internally or externally?

Or, if the define HAVE_CPU_KHZ is not generally available, then it it would be a non-consistent pin and can/should be dealt with in some way.

hdiethelm: HAVE_CPU_KHZ was only defined with RTAI: https://github.com/LinuxCNC/linuxcnc/blob/master/src/emc/motion/mot_priv.h#L352

I removed it all together. I grep'ed the repo for last_period_ns, no hits except the two I removed. If someone uses it with RTAI in his .hal file, he can change to use the pin `last_period` which was in clocks and is now in ns.

The change from clocks to ns for the .time / .last_period might confuse some users but I somewhat expect that most users expected these signals to be in ns. But they where in tsc clocks before this PR.

Is there any reference in the doc about this .time pin's? I did not find anything the last time I searched it.

BsAtHome: The (small) catch is that `last-period` is integer and `last-period-ns` is floating point.

The pins are (wrongly) documented as parameters in docs/src/config/core-components.adoc in line 173-176.

There may be some history here to this pin/param confusion that may need some investigation. Commit 2b22052c8035788a98d74d4a17818977753d8b2f in 2017 changed both from params to pins.

hdiethelm: Ah of course, i grepped for `last_period` instead of  `last-period`. There are only occurrences in the doc's, nowhere in code or hal files. I fixed the doc for this pin.

However, I am completely new to adoc / po translations. Is there any way to update the .po files in an automated way?

Thanks for the link. I did not even know hal-histogram exists. This commit was mainly about converting motion.servo.last-period from a parameter to a pin. motion.servo.last-period-ns was added but it works only with RTAI. Probably it was added as float, so the division by cpu_khz is straight forward. The other time pins are in s32.

I tested `hal-histogram motion.servo.last-period` which works. However, hal-histogram is quite slow in updating and doesn't scale nicely. Just to test, I changed the pin to float, same issue.

hdiethelm: With the right config, it works fine:
grafik

BsAtHome: > However, I am completely new to adoc / po translations. Is there any way to update the .po files in an automated way?

Translations are done in <a href="https://hosted.weblate.org/projects/linuxcnc/">weblate</a> and they are synced once in a while. Just write text in English in the source adoc and they will be marked in weblate that they need attention for translation.

> I tested `hal-histogram motion.servo.last-period` which works. However, hal-histogram is quite slow in updating and doesn't scale nicely. Just to test, I changed the pin to float, same issue.

The slowness of the histogram is how data is taken out of RT into non-RT. It sets an index to read the +/- binnumber[index] and that takes at least 1 thread cycle per bin. So, for +/- 200 bins it takes 0.2 seconds (on a 1kHz thread), just to get the data out (and it is race-tainted data). Actually, the delay is set default at 10 ms and then takes 2 seconds to get the data out.

There is work in progress to fix these histogram programs (none of them work on systems with Tcl9).

hdiethelm: Ok, so it should be fine then. Just some help needed from others for testing.
@grandixximo fixed already the latency-histogram as I remember. It works perfectly now.

hdiethelm:
^^ Found other clocks -> ns instance.

While grep'ing for .time, I found this:
https://github.com/LinuxCNC/linuxcnc/blob/master/docs/src/drivers/hal_pi_gpio.txt#L77

~~I guess it should be documented globally that each hal function gets a `.time` pin and a `.tmax`, `.tmax-increased` parameter for timing analysis, not only motion. It is documented in in several docs in different ways.~~

The code is this one:
https://github.com/LinuxCNC/linuxcnc/blob/7351e0ee04c88c211a2879a979a6a91c3fba0dbe/src/hal/hal_lib.c#L2046
https://github.com/LinuxCNC/linuxcnc/blob/7351e0ee04c88c211a2879a979a6a91c3fba0dbe/src/hal/hal_lib.c#L2062
https://github.com/LinuxCNC/linuxcnc/blob/7351e0ee04c88c211a2879a979a6a91c3fba0dbe/src/hal/hal_lib.c#L2062

hdiethelm:
Disregard, I found also this part and corrected it. It is here:
https://linuxcnc.org/docs/html/hal/basic-hal.html 4. HAL Parameter
`.tmax-increased` was missing, I added it.

src/emc/motion/motion.c
grandixximo: I only fixed the UI, @BsAtHome actually did the rework on the Hal component, credit goes to him for the response improvement 😁

hdiethelm:
I tested RTAI with axis_mm example / latency-test.
With this PR:
latency-test: 
- Jitter approx. +100ns
- .time values: Increase from 70 to 100ns (Calculated using cpu_khz, the old values are in clocks, the new in ns)

axis-mm:
- Hard to say anything, lot of jitter with my PC, uspace runs way better. Approx. + 15% on thread.time.

@NTULINUX 
Do you have time to test this? Is there a way to see CPU load in RTAI?

In uspace, I see no noticeably impact. On RTAI, it is a bit slower with this PR as expected. The question is if this is acceptable or mitigations are needed.

NTULINUX:
I'm going to try and test this on RTAI later today, sorry for the delay! Hosting a network here so whenever I reboot or switch distros, the network for everyone here goes down.

hdiethelm:
Thans. No hurry.
Might be you can also test changing to rt_get_time_ns(): https://github.com/machinekit/machinekit/issues/162
I think this would make sense anyway, but I did not see any difference.

NTULINUX:
debian-01Jun2026-38
Looks good so far! Max CPU usage hits 12.5%

CPU usage looks to be about the same before and after this commit.

hdiethelm:
> CPU usage looks to be about the same before and after this commit.

Thanks!
How do you check the CPU usage with RTAI? I see just a zero in top.
`cat /proc/rtai/scheduler` gives a "TIME" number at the end but I did not found any doc about it.

hdiethelm:
Checking https://github.com/NTULINUX/RTAI/blob/master/src/sched/sched.c#L1457
Replacing rt_get_time_ns() by rt_get_cpu_time_ns() makes no sense, it's the same.

Reading the RTAI code, what rt_get_cpu_time_ns() basically does is reading the tsc register, adding a core specific offset and dividing it by the clock frequency. I don't see why this should be much slower than doing this ourself.
Except raw_smp_processor_id(), which is hidden in the base kernel somewhere, I traced all functions:

Code:
``` C
#define ipipe_read_tsc(t)				\
	__asm__ __volatile__("rdtsc" : "=A"(t))

#define rtai_cpuid()  ipipe_processor_id()
#define ipipe_processor_id()	raw_smp_processor_id()

#define rtai_rdtsc() ({ unsigned long long t; ipipe_read_tsc(t); t + rtai_tsc_ofst[rtai_cpuid()]; })

static inline long long rtai_llimd(long long ll, long mult, long div) {
	return rtai_imuldiv(ll, mult, div);
}

RTIME rt_get_cpu_time_ns(void)
{
	return rtai_llimd(rtai_rdtsc(), 1000000000, tuned.clock_freq);
}
```

Might be an issue how I am testing it or something else but I don't see right now why this call should take 100ns. But CPU's are complex and 100ns is also not that much. You still can read 250x the clock before violating the 25000ns from the base thread.

So I call this ready.

NTULINUX:
> > CPU usage looks to be about the same before and after this commit.
> 
> Thanks! How do you check the CPU usage with RTAI? I see just a zero in top. `cat /proc/rtai/scheduler` gives a "TIME" number at the end but I did not found any doc about it.

top always works for me in RTAI, I used btop and htop as well.

hdiethelm:
> > > CPU usage looks to be about the same before and after this commit.
> > 
> > 
> > Thanks! How do you check the CPU usage with RTAI? I see just a zero in top. `cat /proc/rtai/scheduler` gives a "TIME" number at the end but I did not found any doc about it.
> 
> top always works for me in RTAI, I used btop and htop as well.

Strange. I don't see anything. I disabled "Hide kernel threads"
grafik
grafik

I have the same issue with xenomai, but there I can use `evl ps -s` or `cat /proc/xenomai/sched/threads` and I see the usage.

andypugh:
The "time" parameters have only ever existed for users to monitor behaviour. It has never made any sense for them to be part of a HAL net, but were changed as part of a "parameters make little sense" policy.

PR #4086, opened 2026-05-30
build: hard-fail configure if asciidoctor is missing

### LINKED ISSUE #4084: master: make fails on missing asciidoctor
doing a rip install of current master I got this error on make:

```
Compiling rtapi/uspace_posix.cc
../src/config.status --file=../docs/src/man/man1/linuxcnc.1.adoc:../docs/src/man/man1/linuxcnc.1.adoc.in
config.status: creating ../docs/src/man/man1/linuxcnc.1.adoc
Making manpage 5axisgui.1
/bin/bash: line 1: asciidoctor: command not found
make: *** [../docs/src/Submakefile:1040: ../docs/man/man1/5axisgui.1] Error 127
make: *** Waiting for unfinished jobs....
make: Leaving directory '/home/dave/git/linuxcnc-current-master/src'
```

looks like there might need to be a check for the asciidoctor package.

grandixximo:
## Summary

Fixes #4084.

`docs/src/Submakefile` builds manpages for halcompile components (`5axisgui.1` etc.) via `asciidoctor`, regardless of `--enable-build-documentation`. The configure probe added by #4053 was nested inside `if BUILD_DOCS = yes`, so `./configure --enable-run-in-place` (no docs opt-in) silently passed and `make` later crashed with `asciidoctor: command not found`.

Changes:
- `src/configure.ac`: probe moved out of the `BUILD_DOCS` gate; missing asciidoctor is now `AC_MSG_ERROR` with apt / dnf / gem install hints. PDF / HTML / Rouge probes stay gated.
- `debian/control.top.in`: drop `&lt;!nodoc&gt;` from `asciidoctor` (per @BsAtHome; even `nodoc` builds need it).
- `debian/configure`: remove now-duplicate `asciidoctor` from the head of `DOC_DEPENDS`.

## Test plan

- [x] `./configure --enable-run-in-place` with asciidoctor present: succeeds; `make ../docs/man/man1/5axisgui.1` builds.
- [x] After `apt remove asciidoctor`: configure aborts with `no asciidoctor, manpages cannot be built` + install hint, instead of failing inside `make`.
- [x] `./configure --enable-run-in-place --enable-build-documentation=html` + `make` builds full English HTML (585 pages, 453 manpage HTMLs, link checker clean).
- [x] Indentation of untouched lines preserved (blame intact on the GS / RSVG / Rouge probe blocks).

### DIFF
--- a/debian/configure
+++ b/debian/configure
@@ -110,7 +110,7 @@ elif [ -f /etc/lsb-release ]; then
 fi
 
 if [ -n "$ENABLE_BUILD_DOCUMENTATION" ]; then
-    DOC_DEPENDS="asciidoctor,\n    asciidoctor-pdf | ruby-asciidoctor-pdf,\n    fonts-dejavu,\n    fonts-noto-cjk,\n    ghostscript,\n    graphviz,\n    librsvg2-bin,\n    python3-fonttools,\n    ruby-rouge,\n    w3c-linkchecker"
+    DOC_DEPENDS="asciidoctor-pdf | ruby-asciidoctor-pdf,\n    fonts-dejavu,\n    fonts-noto-cjk,\n    ghostscript,\n    graphviz,\n    librsvg2-bin,\n    python3-fonttools,\n    ruby-rouge,\n    w3c-linkchecker"
 else
     DOC_DEPENDS=''
 fi
--- a/debian/control.top.in
+++ b/debian/control.top.in
@@ -14,7 +14,7 @@ Build-Depends:
     @EXTRA_BUILD@,
     groff-base <!nodoc>,
     imagemagick <!nodoc>,
-    asciidoctor <!nodoc>,
+    asciidoctor,
     libunicode-linebreak-perl <!nodoc>,
     autoconf,
     automake,
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1067,16 +1067,15 @@ AC_ARG_ENABLE(build-documentation,
 # --enable-build-documentation=html).  =html itself warn-and-disables
 # only the missing PDF tools; users who want PDF best-effort can pass
 # =html and rebuild after installing the missing tools.
-if ( test "$BUILD_DOCS" = "yes" ) ; then
-    AC_PATH_PROG(ASCIIDOCTOR,asciidoctor,"none")
-    if ( test "none" = "$ASCIIDOCTOR" ) ; then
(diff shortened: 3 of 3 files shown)


BsAtHome:
In debian/control.top.in you have:
```
    asciidoctor <!nodoc>,
```
Shouldn't you also remove the nodoc tag because halcompile needs it?

grandixximo:
Thanks, pushed an amend (`0c24fb2f02`):
- dropped `<!nodoc>` from `asciidoctor` in `debian/control.top.in`
- removed the now-duplicate `asciidoctor` entry at the head of `DOC_DEPENDS` in `debian/configure`
- tightened the configure.ac comment to one line

Re-ran the configure probe with / without asciidoctor and a full `--enable-build-documentation=html` build (585 HTML pages, 453 manpage HTMLs, link checker clean). `groff-base`, `imagemagick`, `libunicode-linebreak-perl`, `po4a` kept their `<!nodoc>` tags since the asciidoctor manpage pipeline doesn't reach them.

Tangentially: `make htmldocs` / `pdfdocs` / `docs` work standalone after one full `make` (halcompile component pages come from `objects/man/*.adoc`, built by `halcompile` during the C build, so a fresh-tree docs-only build silently omits them). Worth wiring a `docs-with-components` target that depends on `halcompile manpages` first, and maybe promoting that to the default for docs-only flows? Happy to draft a follow-up if you want it.

BsAtHome:
> Tangentially: `make htmldocs` / `pdfdocs` / `docs` work standalone after one full `make` (halcompile component pages come from `objects/man/*.adoc`, built by `halcompile` during the C build, so a fresh-tree docs-only build silently omits them). Worth wiring a `docs-with-components` target that depends on `halcompile manpages` first, and maybe promoting that to the default for docs-only flows? Happy to draft a follow-up if you want it.

Yes, that needs to be fixed too. The `manpages` target must run halcompile to extract the docs. However, the current setup was only necessary because the PDFs built the manpage collection from the troff sources instead of the adoc sources. This is no longer true and the current hack is no longer necessary. Halcompile still needs to be able to run asciidoctor because it can generate troff manpages directly (as compatibility measure).

Now you can run `halcompile --adoc` to extract only the component's manpage in adoc format from the component. Then you use a second rule to build troff from adoc manpages and thereby simplify the build. Take a look at the `hal/components/Submakefile` on the rules for `COMP_MANPAGES` and `COMP_DRIVER_MANPAGES`.

PR #4087, opened 2026-05-30
Create ping component

hdiethelm:
This component does basically the same as icmp ping. However, it runs in the linuxcnc realtime environment, so it can be used to tune network latency.

I created it for network tuning due to `sudo taskset 8 chrt 90 ping -i .001 -q 192.168.1.121` is a bit cumbersome to use and has no histogram option.

Feel free to merge if you think this is useful for everyone.

TBD
- Integrate into latency-test and/or latency-historgam?
- Create example file?
- Is there a way to create userspace only components? So I could get rid of this silly `
#if defined(__KERNEL__) -> Only report an error and quit`

### DIFF
--- a/scripts/latency-ping
+++ b/scripts/latency-ping
@@ -1,0 +1,89 @@
+#!/bin/bash
+SCRIPT_LOCATION=$(dirname "$(readlink -f "$0")");
+if [ -f "$SCRIPT_LOCATION/rip-environment" ] && [ -z "$EMC2_HOME" ]; then
+    . "$SCRIPT_LOCATION/rip-environment"
+fi
+
+T=$(mktemp -d)
+trap 'cd / && [ -d "$T" ] && rm -rf "$T"' SIGINT SIGTERM EXIT
+cd "$T" || { echo "E: Cannot change directory to '$T'"; exit 1; }
+
+calc() { awk "BEGIN { print ($1); }" < /dev/null; }
+icalc() { awk "BEGIN { printf \"%.0f\n\", ($1); }" < /dev/null; }
+
+parse_time () {
+    case $1 in
+    -)   echo "0" ;;
+    *ns) icalc "${1%ns}" ;;
+    *us|*µs) icalc "1000*${1%us}" ;;
+    *ms) icalc "1000*1000*${1%ms}" ;;
+    *s)  icalc "1000*1000*1000*${1%s}" ;;
+    *)   if [ "$1" -lt 1000 ]; then icalc "1000*$1"; else icalc "$1"; fi ;;
+    esac
+}
+
+human_time () {
+    if [ "$1" -eq 0 ]; then echo "-"
+    elif [ "$1" -ge 1000000000 ]; then echo "$(calc "$1"/1000/1000/1000)s"
+    elif [ "$1" -ge 1000000 ]; then echo "$(calc "$1"/1000/1000)ms"
+    elif [ "$1" -ge 1000 ]; then echo "$(calc "$1"/1000)µs"
+    else echo "$1ns"
+    fi
+}
+
+usage () {
+    echo "Usage:"
+    echo "       latency-ping ip_address [if_name] [timeout] [period]  # for single thread"
+    echo ""
+    echo "Example:"
+    echo "       latency-ping 192.168.122.1 enp1s0 800000 1000000"
+    echo ""
+    echo "Defaults:     period=${PERIOD}nS"
+    echo "Equivalently: period=$(human_time "$PERIOD")"
+    echo ""
+    echo "Times may be specified with suffix \"s\", \"ms\", \"us\" \"µs\", or \"ns\""
+    echo "Times without a suffix and less than 1000 are taken to be in us;"
+    echo "other times without a suffix are taken to be in ns"
+    echo ""
+    echo "If two periods are specified then the shortest time period will"
+    echo "always be the base-period"
+    echo ""
+    echo "The worst-case latency seen in any run of latency-test"
+    echo "is written to the file ~/.latency"
+    exit 1
+}
+
+TIMEOUT=$(parse_time 800us)
+PERIOD=$(parse_time 1ms)
+
+case $1 in
+  -h|--help) usage;;
+esac
+
+case $# in
+1) IP="$1" ;;
+2) IP="$1" IF="$2" ;;
+3) IP="$1" IF="$2" TIMEOUT=$(parse_time "$3") ;;
+4) IP="$1" IF="$2" TIMEOUT=$(parse_time "$3") PERIOD=$(parse_time "$4") ;;
+*) usage;;
+esac
+
+if [ -z "$IF" ]; then
+cat > lat.hal <<EOF
+loadrt ping ip_address=$IP timeout=$TIMEOUT
+loadrt threads name1=thread period1=$PERIOD
+addf ping thread
+start
+loadusr -w hal-histogram --minvalue 0 --binsize 1e3 --nbins 200 ping.ping-rtt
+EOF
+else
+cat > lat.hal <<EOF
+loadrt ping ip_address=$IP if_name=$IF timeout=$TIMEOUT
+loadrt threads name1=thread period1=$PERIOD
+addf ping thread
+start
+loadusr -w hal-histogram --minvalue 0 --binsize 1e3 --nbins 200 ping.ping-rtt
+EOF
+fi
+
+halrun lat.hal
--- a/scripts/latency-ping
+++ b/scripts/latency-ping
@@ -36,7 +36,7 @@ usage () {
     echo "       latency-ping ip_address [if_name] [timeout] [period]  # for single thread"
     echo ""
     echo "Example:"
-    echo "       latency-ping 192.168.122.1 enp1s0 800000 1000000"
+    echo "       latency-ping 192.168.1.121 enp1s0 800000 1000000"
     echo ""
     echo "Defaults:     period=${PERIOD}nS"
     echo "Equivalently: period=$(human_time "$PERIOD")"
--- a/src/hal/components/ping.comp
+++ b/src/hal/components/ping.comp
@@ -5,17 +5,18 @@ Ping an IP address for latency tuning / debugging hostmot2
 or other Ethernet connected peripherals.
 
 Example:
-loadrt ping ip_address=192.168.1.121
+loadrt ping ip_address=192.168.1.121 if_name=eth0 timeout=800000
 loadrt threads name1=thread period1=1000000
 addf ping thread
 start
+loadusr -w ../scripts/rip-environment ../scripts/hal-histogram --minvalue 0 --binsize 1e3 --nbins 200 ping.ping-rtt
 
 You can use hal-histogram to show a histogram:
 hal-histogram ping.ping-rtt
 Note: You need to tweak minval / bsize / nbins to have a
 useful display. maxval = minval + bsize * nbins
-A good starting point for 30...130 us is:
-hal-histogram --minvalue 3e4 --binsize 2e3 --nbins 50 ping.ping-rtt
+A good starting point for 0...200 us is:
+hal-histogram --minvalue 0 --binsize 1e3 --nbins 200 ping.ping-rtt
 
 This component works only in userspace.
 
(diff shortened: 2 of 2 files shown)


BsAtHome:
Are you sure it is a good idea to do ping in RT? It just feels wrong.

If you want histograms and such, you should create a standalone program that runs its own thread in RT and generates that kind of data. Running it together with LCNC may not produce the result you expect. Interleaving the normal RT thread(s) with ping may not be very stable for the RT thread or other functions. You also use the standard kernel network API, which gives no guarantees afaik. There may also be unexpected interactions with the normal traffic that is being sent (like ordering).

hdiethelm:
> Are you sure it is a good idea to do ping in RT? It just feels wrong.
> 
> If you want histograms and such, you should create a standalone program that runs its own thread in RT and generates that kind of data. Running it together with LCNC may not produce the result you expect. Interleaving the normal RT thread(s) with ping may not be very stable for the RT thread or other functions. You also use the standard kernel network API, which gives no guarantees afaik. There may also be unexpected interactions with the normal traffic that is being sent (like ordering).

The idea is to use this together with an UI to create something similar to latency-histogram. Just using ping + hal-histogram -> latency-histogram-ping.

The reasoning behind is that with hm2_eth, the biggest latency source is the network.

When I initially set up my machine, it failed. After some digging in the forum, I found the chrt ping ... command which showed network issues. Future digging and I discovered that I had to install an other network card driver / set interrupt priorities / assign the interrupts to the isolated core / set some magic ethtool parameters and so on. Such a tool + good doc would have helped a lot. Also document all this and create a kernel cmdline + network parameter tool is on the bucket list but one step at a time.

Ping latency went down from > 1000us to now ~60-80us.

As much as I know, there is no reltime socket with PREEMT_RT. Also hm2_eth uses just standard sockets with firewall + peresistent ARP.

Firewall / ARP is missing here. I will add ARP but firewall would add a lot of copy-paste code and is not necessary as long as you don't do dumb stuff.

However, this component will not behave idealy with Xenomai3/4 which have dedicatet network sockets. I have to add a warning now that there is a way to query which type is running.

hdiethelm:
So, first prototype with gui works:
`latency-ping 192.168.122.1 enp1s0 0.8ms 1ms`

A nicer gui would be nice but the existing latency-histogram gui is in tcl... Yet an other language for me... ;-)

Note: This is in a VM. I have to test it tomorrow on the real machine, but the last time I tested it withouth gui, max was ~ 60-70us.
grafik

hdiethelm:
I could not let it be and did two short tests.
Well tuned:
Bildschirmfoto_2026-07-24_00-40-48
Tuning ethernet disabled:
Bildschirmfoto_2026-07-24_00-35-59
This maches well the read.tmax of 67 and 92 us I see in LinuxCNC. However, I need to run longer tests, it will get way worse with no tuning as much as I remember.

hdiethelm:
So, a detailed comparsion hm2_eth vs ping on my machine, each ~30min

`scripts/latency-ping 192.168.1.121 enp2s0f0 800000 1000000`
ping_tuned_30min

`scripts/linuxcnc ~/linuxcnc/configs/camelot_mill/7i96s.ini` +
`scripts/rip_environment scripts/hal-histogram --minvalue 0 --binsize 1e3 --nbins 200 hm2_7i96s.0.read.time`
hm2_eth_tuned_30min

`scripts/latency-ping 192.168.1.121 enp2s0f0 800000 1000000` + 
`stress-ng --cpu 2 --cpu-method all --vm 2 --vm-bytes 50% --fork 2`
ping_tuned_stress_30min

`scripts/linuxcnc ~/linuxcnc/configs/camelot_mill/7i96s.ini` +
`scripts/rip_environment scripts/hal-histogram --minvalue 0 --binsize 1e3 --nbins 200 hm2_7i96s.0.read.time` +
`stress-ng --cpu 2 --cpu-method all --vm 2 --vm-bytes 50% --fork 2`
hm2_eth_tuned_stress_30min

Looks well repeatable.

Now the question is, if such a tool is usable.

Advantages:
- You can test any real time Ethernet device that supports ping excluding possible issues generated by LinuxCNC
  - Mesa Ethernet
  - Ethercat 
  - Whatever else, some home brew LinuxCNC Ethernet device
- You can tune and test Ethernet without having to run a full LinuxCNC config. For me, LinuxCNC stopped working with really bad network card due to timeouts. While this ping tool, this would still work, you just increase the timeouts. I can even ping my router and measure a latency histogram.

Disadvantage:
- Yet an other tool that needs doc / maintenance

@grandixximo @pcw-mesa What do you think, is this useful? If yes, I will improve it and make it merge ready. If no, I will stop here and might be provide it in a separate repo in my space, no issue for me.

hdiethelm:
@zz912 Would you like to test it on your machine? I somehow expect to see a difference to hm2_eth there.

That would involve two commands while a PREEMT_RT kernel is running. 30min test should be enough for now.

First, check your hm2_read.time using a histogram:
Start your machine config +
`scripts/rip_environment scripts/hal-histogram --minvalue 0 --binsize 1e3 --nbins 200 hm2_7i96s.0.read.time`

Second, run this ping tool:
`scripts/latency-ping 10.10.10.10 eno1 800000 1000000`

BsAtHome:
> What do you think, is this useful? If yes, I will improve it and make it merge ready. If no, I will stop here and might be provide it in a separate repo in my space, no issue for me.

It would be useful, not as a separate tool, but as a part in tuning the network stack.

You need to have full documentation how actually to do the tuning with all details and testing steps. Creating (and maintaning) the docs is the real task and big works in optimizing performance, not the tool itself. (It would also be useful to have a decision tree how you do things based on results from sifferent sources.)

hdiethelm:
> > What do you think, is this useful? If yes, I will improve it and make it merge ready. If no, I will stop here and might be provide it in a separate repo in my space, no issue for me.
> 
> It would be useful, not as a separate tool, but as a part in tuning the network stack.

Is there already anything for tuning the network stack?

> 
> You need to have full documentation how actually to do the tuning with all details and testing steps. Creating (and maintaning) the docs is the real task and big works in optimizing performance, not the tool itself. (It would also be useful to have a decision tree how you do things based on results from sifferent sources.)

Yes, I know. Especially with moving kernel versions / drivers and so on. All useful doc I found so far is spread in the forum /  in blogs and in LinuxCNC unrelated general doc.

https://www.linuxcnc.org/docs/html/install/latency-test.html only describes what to tune but not how.

There is one hint on hardware-irq-coalesce-rx-usecs in https://www.linuxcnc.org/docs/html/man/man9/hm2_eth.9.html but there is way more you can do.

zz912:
to hdiethelm:
zdenek-02Aug2026-1894-w_modbus

zdenek-02Aug2026-3234-w_modbus

zdenek-02Aug2026-1894-w_modbus_EVL

zdenek-02Aug2026-3588-w_modbus

For the "hm2 + Xenomani V4" test I had to merge the branches ping_component_v2 and hm2_eth_oob_v5

to BsAtHome:

I agree that having a comprehensive guide for diagnosing and optimizing Ethernet latency would be very valuable. However, I don't think it should be a prerequisite for merging this PR.

In my opinion, the documentation for this component only needs to explain how to interpret the test results. For example, what kind of results can be considered acceptable, what should be considered a warning, and what indicates that Ethernet latency is likely the source of the problem.

From my experience, many LinuxCNC users are machine builders, electricians, or CNC technicians rather than Linux or networking experts. For many of them, LinuxCNC is their first encounter with Linux. When they experience communication problems with Mesa Ethernet or EtherCAT, they usually find information about running latency-test and checking the PC's realtime latency. If those numbers look good, they have no indication that they should start investigating Ethernet latency instead.

In practice, this often leads to trial-and-error: trying a different PC, a different network card, or changing hardware without knowing where the actual problem is.

This tool doesn't need to teach users how to optimize their systems. Simply helping them identify that Ethernet latency is the source of the problem is already a significant improvement over the current situation. A comprehensive Ethernet latency tuning guide would certainly be a welcome addition, but I see that as a separate piece of documentation rather than a requirement for merging this PR.

hdiethelm:
@zz912 Ok, there is definitively something wrong with your PC / real time config or mesa card. Both ping and hm2_7i96s.0.read.time show huge latency distribution.

PREEMT_RT:
- ping 70-500us
- read 100-500us

Xenomai4:
- ping 70-250us
- read 95-250us

In comparison, my setup, PREEMT_RT:
- ping 50-80us
- read 60-100us

Funny enough, xenomai4 works better with ping, even thought no PREEMT_RT patch is applied, so ping should not work better. I except to run worse.

Did you run any load while running this tests? Did you have network card tuning active?

Main suspects:
- Graphics
  - I assume you use the integrated one, no PCIe card?
  - I should have something similar in my PC. I will try to reproduce it but could take some time.
  - PREEMT_RT has a patch applied for this GPU while Xenomai4 has not. Could be the reason why PREEMT_RT runs way worse than Xenomai4 if this patch has a bug.
  - Do you have by chance an AMD GPU lying around? If yes, you can try this one. Was way better in my setup, except rare spikes to ~150us when watching you-tube. I'm on this since months but I did not yet find a solution.
- Network card
  - You can try an other one. I recommend an I250 / I350. I only tested an I350 but this behaves really well.

hdiethelm:
I did a few short tests, amdgpu vs. i915, all PREEMT_RT.

TLDR: i915 is bad for real-time.

Most probable reason: iGPU's use the CPU memory. So the CPU can be blocked while the GPU is accessing the memory. In this case, proper network tuning helps even more. If the Ethernet Interrupt is on the same CPU than LinuxCNC, less CPU<->CPU communication is needed.

- amd + tuning 52-65us
- amd + no tuning 52-95us (See above in the thread)
- i915 + tuning 52-146us
- i915 + no tuning 50-226us

@zz912 My PC has a different CPU:
mine: https://www.intel.de/content/www/de/de/products/sku/88184/intel-core-i56500-processor-6m-cache-up-to-3-60-ghz/specifications.html
yours: https://www.intel.de/content/www/de/de/products/sku/68316/intel-core-i53470-processor-6m-cache-up-to-3-60-ghz/specifications.html
Mine is a bit faster and also faster memory, that is probably the reason for the worse latency on yours.

If you can, I would try to buy a cheap second hand AMD GPU. I got an RX 460 for ~20$. It is also not perfect, no watching you-tube while milling, but is works way better. However, I can not guarantee that it works for you. There might be additional issues.

amdgpu + network tuning + glxgears:
ping_amdgpu_rt
i915 + network tuning + glxgears:
ping_igpu_rt
i915 + NO network tuning + glxgears:
ping_igpu_rt_no_tune

zz912:
Thanks for taking the time to run these additional tests. They are very interesting and help explain the differences I'm seeing.

Maybe I haven't explained my goal clearly enough. At the moment I'm not trying to optimize this PC for production use. I'm not building or commissioning a CNC machine right now. Instead, I'm testing your tool as a LinuxCNC community member because I want to help validate it on a wider range of hardware.

From my perspective, having a less-than-ideal system is actually useful. If the tool can reliably identify that Ethernet latency is the problem on hardware like mine, then it will also help many other users who have similar systems.

This discussion is actually a good example of why I think the tool is valuable. Before running these tests, I had no quantitative way to tell that my integrated GPU and hardware configuration were contributing to the Ethernet latency. Many LinuxCNC users are in the same situation.

Once these tools (and their documentation) become part of LinuxCNC, I'll be happy to optimize my own hardware. At that point I'll probably replace the network card and maybe also the GPU. For now, though, I think testing on an ordinary, imperfect system provides useful feedback for validating the tool.

One small usability suggestion: before you explained that the histogram shows time, I didn't fully understand what I was looking at.

I think it would help if the histogram (or the documentation) explicitly stated the units, e.g. that the X-axis is latency in nanoseconds (ns). For someone seeing the tool for the first time, that would make the results much easier to interpret.

> Did you run any load while running this tests? 

No

> Did you have network card tuning active?

I have same setup from previous communication with you. I did not have time for experiments.

zz912:
I think a more appropriate name for this script is ping-histogram.

zz912:
I created a separate issue for the ongoing discussion and tuning of my PC:
https://github.com/hdiethelm/xenomai4-linuxcnc/issues/3
Let's continue the PC tuning and test results there. I'd prefer to keep this PR focused on the ping tool itself.

PR #4088, opened 2026-05-30
qtvcp probe help: fix broken/blurry images and typos

PeterStolz:
Fixes the bundled help pages for the **VersaProbe** and **Basic Probe** dialogs in qtvcp. Three commits:

### 1. Fix broken image paths
The versaprobe (pages 4-7) and basic_probe (pages 7-8) help pages referenced their images through an absolute path baked into a developer's home directory (`/home/chris/emc/share/qtvcp/images/...`), so they rendered as broken-file icons on every install. basic_probe page 6 also omitted the `.png` extension. Pointed them at the existing assets via the relative `../images/` form the loader rewrites.

### 2. Render help diagrams crisp at display size
The help dialogs (`QTextEdit`) rescale images with no smoothing, so the diagrams looked soft or pixelated: most pages forced `width=500` onto images of widely varying native sizes, and the versaprobe diagrams pointed at tiny 300px thumbnails. Shipped the diagrams pre-scaled to their 500px display width (Lanczos) in new `probe_icons_help/` and `basic_probe_help/` dirs and reference those, so Qt draws them 1:1 with no rescaling. Generalised the loader image-path rewrite from `../images/probe_icons/` to `../images/` so any image subdir resolves. Also centred the versaprobe page-4 image to match the other pages and dropped the width override on basic_probe page 4 (480px native, was being upscaled).

### 3. Fix typos in help text
Corrected several typos in the help text (`then`->`than`, `imprial`->`imperial`, `clearence`->`clearance`, a missing word, a garbled sentence rephrased to match the wording already used on page 1, and the stray `EXTRA CLEARANCE` references renamed to the actual field name `EXTRA DEPTH`).

No functional/behaviour changes - HTML, help images, and one loader string.


BsAtHome:
Wouldn't it be better to replace the images, that are no screen-shots, to be SVG images? Or can't QtTextDocument render SVG images linked in html `` tags?

PeterStolz:
Good question — I looked into it properly. `QTextDocument` *does* render SVG via `` (Qt's qsvg imageformat plugin, which qtvcp already pulls in for icons — I verified it draws vector-crisp even when upscaled), and you're right that the diagrams are vector at heart: the versaprobe ones carry `Software: www.inkscape.org` in their PNG metadata, so they were originally drawn in Inkscape.

The snag is that the `.svg` sources were never committed — only the PNG exports are in the tree (I checked the full history too), and a couple of the help images are actually `gnome-screenshot` captures of the panel, which can't be vectorised anyway.

So instead of bundling resized PNG copies, I switched the help-dialog loader to pre-scale each raster image to its display size with a smooth filter and register it as a document resource, so Qt draws it 1:1 rather than nearest-neighbour (which was the cause of the jaggies). SVG `` are passed straight through untouched, so if the original Inkscape sources ever surface they can just be dropped in with no further code change. Pushed as the latest commit — no new assets added.

PeterStolz:
For scope, while it's fresh: because this fix lives in the `VersaProbe`/`BasicProbe` widgets themselves, it lands everywhere those are embedded: `qtaxis`, `qtdragon`, `qtdragon_hd`/`qtdragon_hd_vert`, `qtdragon_lathe`, `woodpecker`, `x1mill`, plus the standalone `panels/versaprobe` (which embeds into any GUI, including AXIS/gmoccapy), not just one screen.

One thing I noticed in there: the two `HelpDialog` classes in `versa_probe.py` and `basic_probe.py` are ~80% identical (they differ only in title, page list, preference key, page step, and one fallback string). They could be collapsed into a single shared HelpDialog base, which would also let this scaling helper live in one place instead of being duplicated. Happy to do that as a follow up if there's interest.

c-morley:
Thanks for work. More pull requests will be happily welcomed.

Issue #2292, opened 2023-01-21
Create section in documentation for supported hardware


hansu:
We don't really have an overview of the hardware LinuxCNC can be used with in the docs.

So this is about creating a page in the docs based on this wiki page: http://wiki.linuxcnc.org/cgi-bin/wiki.pl?LinuxCNC_Supported_Hardware and extend/update it.

Maybe a separate page describing the Mesa hardware would be useful.

andypugh:
There is http://linuxcnc.org/docs/stable/html/drivers/hostmot2.html though it's not great.

hansu:
> There is http://linuxcnc.org/docs/stable/html/drivers/hostmot2.html though it's not great.

Yeah and the most popular cards are not mentioned :smile: .
Here is also a bit about the Smart Serial devices: https://linuxcnc.org/docs/stable/html/man/man9/sserial.9.html

petterreinholdtsen:
[Hans Unzner]
> So this is about creating a page in the docs based on this wiki page:
> http://wiki.linuxcnc.org/cgi-bin/wiki.pl?LinuxCNC_Supported_Hardware
> and extend/update it.
>
> Maybe a separate page describing the Mesa hardware would be useful.

Instead of supported hardware, what about a list if of various hardware
and their support status and driver references?  This way one could also
find non-supported hardware and hardware with out-of-tree drivers?  We
would have one place to point people to when they wonder if their
hardware work. :)

-- 
Happy hacking
Petter Reinholdtsen

hansu:
Hardware that has been supported or is one the way being supported could be added, yes.
And hardware with external drivers as well.

PR #4089, opened 2026-05-31
Optimize docs

hansu:
Some optical fixes:

1. The default asciidoctor CSS scales the font up the window width > 768px, this looks too big --> revert this rule.
2. Align the caption of images according to the alignment of the image itself, default center align images .
3. Remove the "Chapter ###" from the PDF sidebar TOC, set inline level to so that we get a flat list of the man pages there
4. Set image width for images that doesn't fit into the default 75% scheme
5. Fix formatting of those block which renders as quote with asciidoctor
    ```
     ____
    ...
     ____ 
    ```

### DIFF
--- a/docs/src/Submakefile
+++ b/docs/src/Submakefile
@@ -766,7 +766,8 @@ objects/LinuxCNC_Manual_Pages.adoc: objects/var-PDF_MAN_ORDER $(MAN_SRCS) $(DOC_
 	    echo ":doctype: book"; \
 	    echo ":source-highlighter: rouge"; \
 	    echo ":toc:"; \
-	    echo ":numbered:"; \
+	    echo ":outlinelevels: 1"; \
+	    echo ":toclevels: 1"; \
 	    echo; \
 	    for M in $(PDF_MAN_ORDER); do \
 	        if   [ -r "$(DOC_SRCDIR)/man/$$M.adoc" ]; then \
--- a/docs/src/lcnc-overrides.css
+++ b/docs/src/lcnc-overrides.css
@@ -399,3 +399,9 @@ h1{font-size:revert}
 h2{font-size:revert}
 h3,#toctitle,.sidebarblock>.content>.title{font-size:revert}
 h4,h5{font-size:revert}}
+
+.imageblock.text-left   > .title { text-align: left; }
+.imageblock.text-center > .title { text-align: center; }
+.imageblock.text-right  > .title { text-align: right; }
+.imageblock             > .title { margin-top: 0.5em;}
+  
--- a/docs/src/lcnc-overrides.css
+++ b/docs/src/lcnc-overrides.css
@@ -392,3 +392,10 @@ body:not(.article):not(.book):not(.manpage) tr:has(> th:only-child) > * {
     color: #f0f0f0;
   }
 }
+
+@media screen and (min-width:768px){
+h1,h2,h3,#toctitle,.sidebarblock>.content>.title,h4,h5,h6{line-height:revert}
+h1{font-size:revert}
+h2{font-size:revert}
+h3,#toctitle,.sidebarblock>.content>.title{font-size:revert}
+h4,h5{font-size:revert}}
--- a/docs/src/pdf-theme.yml
+++ b/docs/src/pdf-theme.yml
@@ -128,3 +128,8 @@ table:
     border_bottom_width: 1
   border_color: CCCCCC
   cell_padding: 4
+
+image:
+  caption:
+    align: inherit
+    text-align: inherit
--- a/docs/src/Master_Developer.adoc
+++ b/docs/src/Master_Developer.adoc
@@ -7,7 +7,7 @@
 
 == Introduction
 
-image::common/images/emc2-intro.png[]
+image::common/images/emc2-intro.png[pdfwidth=25%]
 
 include::common/overleaf.adoc[]
 
--- a/docs/src/Master_Getting_Started.adoc
+++ b/docs/src/Master_Getting_Started.adoc
@@ -9,7 +9,7 @@ The LinuxCNC Team
 
 :leveloffset: 1
 
-image::common/images/emc2-intro.png[]
+image::common/images/emc2-intro.png[pdfwidth=25%]
 
 include::common/overleaf.adoc[]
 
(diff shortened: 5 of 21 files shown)


docs/src/lcnc-overrides.css
```diff
@@ -392,3 +392,10 @@ body:not(.article):not(.book):not(.manpage) tr:has(> th:only-child) > * {
     color: #f0f0f0;
   }
 }
+
+@media screen and (min-width:768px){
+h1,h2,h3,#toctitle,.sidebarblock>.content>.title,h4,h5,h6{line-height:revert}
+h1{font-size:revert}
+h2{font-size:revert}
+h3,#toctitle,.sidebarblock>.content>.title{font-size:revert}
+h4,h5{font-size:revert}}
```
BsAtHome: Please set your editor to emit a newline on the last line if the file. Having the last line without termination is inconsistent.

docs/src/pdf-theme.yml
```diff
@@ -128,3 +128,8 @@ table:
     border_bottom_width: 1
   border_color: CCCCCC
   cell_padding: 4
+
+image:
+  caption:
+    align: inherit
+    text-align: inherit
```
BsAtHome: Same last-line missing newline problem. (And the YAML file format is very picky too)

hansu: Fixed

docs/src/lcnc-overrides.css
hansu: Fixed

docs/src/Submakefile
```diff
@@ -766,7 +766,7 @@ objects/LinuxCNC_Manual_Pages.adoc: objects/var-PDF_MAN_ORDER $(MAN_SRCS) $(DOC_
 	    echo ":doctype: book"; \
 	    echo ":source-highlighter: rouge"; \
 	    echo ":toc:"; \
-	    echo ":numbered:"; \
+	    echo ":outlinelevels: 1"; \
```
grandixximo, line 769: can we add
`echo ":toclevels: 1"; \`
as well here? so the TOC is minimal in the PDF output?

hansu: Yeah sure, good catch!

hansu: Done.

hansu:
I think the chapters on the main docs page should be collapsed by default as they were before. Otherwise one have to scroll a lot. And as the javascript was removed, one have to scroll every time...

I added an corresponding commit.

BsAtHome:
> I think the chapters on the main docs page should be collapsed by default as they were before. Otherwise one have to scroll a lot. And as the javascript was removed, one have to scroll every time...

I think you should have at least the first one open. Having all collapsed does not give the user an indication that these are foldable structures. Having one (the first) open is a visual cue indicative of an action possible (together with the arrow; which should be styled a bit).

(the removal of javascript has not changed anything in the behaviour because open/closed state was not saved afaik)

FWIW, we need to rethink the front page IMO to support a more logical organisation. Regardless of the current open/closes list, it is difficult to find the information you want (because there is too much). The focus should be the user's perspective and how a user thinks.

hansu:
> > I think the chapters on the main docs page should be collapsed by default as they were before. Otherwise one have to scroll a lot. And as the javascript was removed, one have to scroll every time...
> 
> I think you should have at least the first one open. Having all collapsed does not give the user an indication that these are foldable structures. Having one (the first) open is a visual cue indicative of an action possible

Good point :+1:
Or have them underlined as classic link style (but a little bit ugly).

> (together with the arrow; which should be styled a bit).

The old plus was more clear in this relation

 
> (the removal of javascript has not changed anything in the behaviour because open/closed state was not saved afaik)

It rememberes the opened trees, you can still compare with the 2.9 docs: https://linuxcnc.org/docs/html/


> FWIW, we need to rethink the front page IMO to support a more logical organisation. Regardless of the current open/closes list, it is difficult to find the information you want (because there is too much). The focus should be the user's perspective and how a user thinks.

Getting started and general information are on top. I don't think it's that bad. What do you have in mind?
Only a chapter "supported hardware" could be of interest. I have had this idea earlier but forgot about that: https://github.com/LinuxCNC/linuxcnc/issues/2292

BsAtHome:
> > (the removal of javascript has not changed anything in the behaviour because open/closed state was not saved afaik)
> 
> It rememberes the opened trees, you can still compare with the 2.9 docs: https://linuxcnc.org/docs/html/

Well, I am not completely against using js. I just require it not necessary for any normal operation of the pages. Anything should be completely functional and visible using standard html. So, we can hook a listener and use local storage (for the poor souls who cannot live without a js kick ;-))

> > FWIW, we need to rethink the front page IMO to support a more logical organisation.
> 
> Getting started and general information are on top. I don't think it's that bad. What do you have in mind? Only a chapter "supported hardware" could be of interest. I have had this idea earlier but forgot about that: #2292

If I had the answer, I'd tell you right away. There are too many things on my mind right now to redesign the documentation content. It is just that my own use of the pages always was accompanied with frustration and that is a sure sign that it is off (when you browse through the 1400 page PDF you also know that something is amiss). I need to do a deep dive and inventory before I can make a proper proposal.

grandixximo:
> The focus should be the user's perspective and how a user thinks.

Agreed, there should be less granular linkage on the main page, content should be divided in main categories the user is visiting the docs for.

Intro/getting started
Hardware
HAL
INI
UI
Gcode
Contribute

Maybe a few more, but those are generally what I visit the docs for, there maybe some cross referencing.
Just my two cents, but I'd rather have Bertho map it out in detail 😁
Man pages should be under HAL IMO, I would not mind more clicks if things are organized logically, under HAL you could have a sub-page subdividing introduction to HAL, and all components to a further sub-page, granularity has to be discussed, I think with about 3 max indirection with sub pages, you should be able to get to the page you wanted, without implementing a fuzzy search in JS.

BsAtHome:
@hansu, you haven't fixed the "have one item list open" yet? Or did I miss it when I looked at the diff?

hansu:
> @hansu, you haven't fixed the "have one item list open" yet? Or did I miss it when I looked at the diff?

Now updated.


>Well, I am not completely against using js. I just require it not necessary for any normal operation of the pages. Anything should be completely functional and visible using standard html. So, we can hook a listener and use local storage (for the poor souls who cannot live without a js kick ;-))

Okay let's do this in another PR.

PR #4091, opened 2026-05-31
build: extract halcompile manpages via adoc, drop troff sed hacks

grandixximo:
## Summary

Per @BsAtHome's review on #4086.  The old `COMP_MANPAGES` /
`COMP_DRIVER_MANPAGES` rules ran `halcompile --document` (troff straight
out) and sed-escaped `.als` / `.URL` groff directives the old dblatex
PDF pipeline could not parse.  Asciidoctor now reads the component
adoc directly, so the sed hacks are obsolete.

New chain:
- `halcompile --adoc -o objects/man/man9/<comp>.9.adoc <comp>.comp`
- `asciidoctor --doctype=manpage --backend=manpage` -> `../docs/man/man9/<comp>.9`

The adoc extraction only needs Python + halcompile.py (yapps-generated,
not the C build), so a fresh-tree docs-only build (`make htmldocs` /
`pdfdocs` / `docs`) extracts and renders component manpages without
first compiling the rest of LinuxCNC.  Previously a docs-only build
silently omitted them.

Also adds `homecomp.9` to `COMP_MANPAGES` alongside `tpcomp.9` (both
`.comp` files are filtered out of `COMPS` because they do not compile
to `.o` modules, but their manpages are still worth shipping).
Surfaces in @hansu's review of #4081 as a
`gen_complist: Broken link: ../man/man9/homecomp.9.html` warning.

`.9.adoc` suffix matches the convention the docs HTML manpage rule
already expects under `objects/man/`.

## Test plan

- [x] Clean-tree `make docs` (English-only): 207 troff + 154 extracted
      component adocs, link checker clean.
- [x] `make ../docs/man/man9/homecomp.9` produces troff (was silently
      missing before).
- [x] Standalone `make ../docs/man/man9/abs.9` from a tree with no
      halcompile binary rebuilds halcompile and produces the manpage
      without touching C code.

### DIFF
--- a/src/hal/components/Submakefile
+++ b/src/hal/components/Submakefile
@@ -33,7 +33,7 @@ CONVERTERS :=  \
     conv_u64_u32.comp \
     conv_u64_s64.comp
 COMPS := $(filter-out hal/components/tpcomp.comp hal/components/homecomp.comp, $(sort $(wildcard hal/components/*.comp) $(addprefix hal/components/, $(CONVERTERS))))
-COMP_MANPAGES := $(patsubst hal/components/%.comp, ../docs/man/man9/%.9, $(COMPS)) ../docs/man/man9/tpcomp.9
+COMP_MANPAGES := $(patsubst hal/components/%.comp, ../docs/man/man9/%.9, $(COMPS)) ../docs/man/man9/tpcomp.9 ../docs/man/man9/homecomp.9
 ifeq ($(BUILD_SYS),uspace)
 COMP_DRIVERS += hal/drivers/serport.comp
 COMP_DRIVERS += hal/drivers/mesa_7i65.comp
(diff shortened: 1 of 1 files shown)


src/hal/components/Submakefile
```diff
@@ -54,19 +54,40 @@ endif
...
+	@mkdir -p $(dir $@)
+	$(Q)../bin/halcompile -U --adoc -o $@.tmp $<
+	$(Q)mv -f $@.tmp $@
 
-$(COMP_DRIVER_MANPAGES): ../docs/man/man9/%.9: hal/drivers/%.comp ../bin/halcompile
+$(COMP_DRIVER_MANPAGE_ADOCS): objects/man/man9/%.9.adoc: hal/drivers/%.comp ../bin/halcompile
+	$(ECHO) Extracting adoc manpage $(notdir $@)
+	@mkdir -p $(dir $@)
+	$(Q)../bin/halcompile -U --adoc -o $@.tmp $<
+	$(Q)mv -f $@.tmp $@
```
BsAtHome: Why keep the `mv $@.tmp $@`? There are no intermediate steps anymore so the halcompile output can write directly to the target.

grandixximo: You're right, there's no post-processing step left, so the direct `-o $@` works. I tested it and the output is byte-identical.

The `$@.tmp` + `mv` only buys marginal crash/kill mid-write safety. A clean halcompile error is already safe without it (halcompile builds the adoc in a `/tmp` tempdir and only moves it to the target at the very end), and a plain Ctrl+C is handled by make deleting the target of the interrupted rule. The one case the `mv` catches is halcompile dying partway through its final write (disk full, OOM, kill -9): with `-o $@` that leaves a partial target that make, with no `.DELETE_ON_ERROR`, treats as up to date, whereas the junk lands in `$@.tmp` and `$@` stays clean.

That's marginal though, and the old rule wrote `-o $@` directly for years. Happy to drop the `.tmp`/`mv`. If you'd rather have the guarantee tree-wide, I can add `.DELETE_ON_ERROR:` instead.

grandixximo: Drop the `mv`?

BsAtHome: Yes, simply drop the move.

You cannot guard against OOM, kill -9 or disk full in any meaningful and consistent way when doing a build. It is anybody's call what happens in the extreme. And, doing a build on a constraint system should be considered a bug.

grandixximo: Done

PR #4092, opened 2026-05-31
configs: remove duplicate INI keys (first occurrence wins)

PeterStolz:
## What

Several shipped `sim`/demo configs define the same INI key more than once within one section. LinuxCNC only reads the **first** occurrence, so the later lines are silently ignored — in two cases hiding the author's intent. Split into two commits:

1. **conflicting** duplicates (different values — a value was silently dead)
2. **redundant** duplicates (same value — pure no-op de-clutter)

## How I found it

I'm building a LinuxCNC VS Code extension (HAL/INI/G-code language support) — https://marketplace.visualstudio.com/items?itemName=PeterStolz.linuxcnc-lsp — and its INI linter flagged duplicate keys across a number of in-tree configs, so I went through each to confirm they're real.

## Why the second value is ignored

`IniFile::findTag(tag, section, num)` defaults to `num=1` and returns the first match in file order (`src/emc/ini/inifile.cc`); there is no concatenation for scalar keys. The same is true of the `inivar` CLI used by `scripts/linuxcnc` and of `linuxcnc.ini(...).find()` used by the GUIs. Confirmed on a 2.10 run-in-place build:

```
$ inivar -ini configs/sim/axis/histogram_demo.ini -var MAX_LINEAR_VELOCITY -sec TRAJ
1.2          # first wins; the file's later 1.2345 is ignored

>>> linuxcnc.ini(".../woodpecker_xyzab.ini").find("DISPLAY","GEOMETRY")
'XYZABCUVW'  # findall() -> ['XYZABCUVW','xyzab']; the intended 'xyzab' was dead
```

## Commit 1 — conflicting duplicates

Behaviour-changing (restore intent):

- `sim/axis/gladevcp/probe.ini` — `[RS274NGC] SUBROUTINE_PATH` was set twice, so the second value (`../../nc_files/gladevcp_lib`, the gladevcp demo's O-word subs, per its own comment) was never on the path. Merged into one colon-separated `SUBROUTINE_PATH`.
- `sim/woodpecker/woodpecker_xyzab.ini` — `[DISPLAY] GEOMETRY = xyzab` (added for this XYZAB variant) was overridden by a leftover `XYZABCUVW`; made `xyzab` effective.

No-op (the removed value was already the ignored one):

- `woodpecker_xyzab.ini` — duplicate `[RS274NGC] PARAMETER_FILE` (`metric_parameters.txt` dead; keeps `woodpecker.var`).
- `woodpecker_xyza.ini` — a pasted-twice `[DISPLAY]` spindle-override block.
- `histogram_demo.ini`, `ini_hal_demo.ini` — stray second `[TRAJ] MAX_LINEAR_VELOCITY = 1.2345`.
- `sim/axis/vismach/5axis/bridgemill/5axis.ini` — redundant `[TRAJ] MAX_ANGULAR_VELOCITY = 360` (identical to `360.0`).

## Commit 2 — redundant (same-value) duplicates

Pure de-clutter, no behaviour change:

- `[TRAJ] MAX_LINEAR_VELOCITY = 58` in `sim/axis/gantry/gantry.ini`, `gantry_jjog.ini`, `sim/qtaxis/gantry/qt_gantry.ini`, `sim/qtvcp_screens/qt_gantry.ini`, `qtdefault_gantry.ini`.
- `[RS274NGC] HAL_PIN_VARS = 1` in `sim/axis/vismach/scara/scara.ini`, `sim/qtaxis/non-trivial/scara/scara.ini`, `sim/qtvcp_screens/non-trivial/scara/scara.ini`.
- `[DISPLAY] GEOMETRY = XYZCBW` in `sim/axis/vismach/5axis/bridgemill/5axis.ini`.

## Not included

The Sherline `Sherline4Axis_{inch,mm}.ini` configs have a `# for gui only` block where the first `[TRAJ] MAX_LINEAR_VELOCITY` (`25`) wins over a more realistic second value (`.36` in/s · `8` mm/s). Fixing that changes the effective max velocity — a judgement call — so it's left for separate discussion.


configs/sim/woodpecker/woodpecker_xyzab.ini
```diff
@@ -11,7 +11,7 @@ DISPLAY = qtvcp -f woodpecker
 PREFERENCE_FILE_PATH = WORKINGFOLDER/woodpecker.pref
 POSITION_OFFSET = RELATIVE
 POSITION_FEEDBACK = ACTUAL
-GEOMETRY = XYZABCUVW
+GEOMETRY = xyzab
```
grandixximo: Almost all INI use uppercase

grandixximo (approved):

configs/sim/woodpecker/woodpecker_xyzab.ini
PeterStolz: Done

PR #4093, opened 2026-05-31
interp: add new ini setting disabling auto reset to G54 on M2,M30

Sigma1912:
As requested here:
https://forum.linuxcnc.org/38-general-linuxcnc-questions/58821-how-to-remove-automatic-g54-after-m2-m30#346816

Adds a new ini file setting to disable the automatic resetting of the current WCS to G54 on M2 and M30.

Setting defaults to '0' so this modification is opt-in and default behavior is unchanged.

### DIFF
--- a/docs/src/config/ini-config.adoc
+++ b/docs/src/config/ini-config.adoc
@@ -635,6 +635,8 @@ The maximum number of `USER_M_PATH` directories is defined at compile time (typ:
   Warn rather than error in case of errors in O-word subroutines.
 * `DISABLE_G92_PERSISTENCE = 0` (Default: 0)
   Allow to clear the G92 offset automatically when config start-up.
+* `DISABLE_AUTO_G54 = 0` (bool, Default: 0) +
+  When set M2 and M30 will no longer automatically reset the active WCS to 'G54'.
 * `DISABLE_FANUC_STYLE_SUB = 0` (Default: 0)
   If there is reason to disable Fanuc subroutines set it to 1.
 * 'G73_PECK_CLEARANCE = .020' (default: Metric machine: 1mm, imperial machine: .050 inches)
--- a/src/emc/rs274ngc/interp_internal.hh
+++ b/src/emc/rs274ngc/interp_internal.hh
@@ -825,6 +825,7 @@ struct setup
     bool loop_on_main_m99;
 
   int disable_g92_persistence;
+  bool disable_auto_g54;
 
 // add new geometric fields for our new tags
   double heading;
--- a/src/emc/usr_intf/pncconf/build_INI.py
+++ b/src/emc/usr_intf/pncconf/build_INI.py
@@ -176,6 +176,9 @@ class INI:
         print(file=file)
         print("[RS274NGC]", file=file)
         print("PARAMETER_FILE = linuxcnc.var", file=file)
+        print("# Set 'DISABLE_AUTO_G54 = 1' to disable", file=file)
+        print("# automatic resetting of the active WCS to 'G54' on M2 and M30", file=file)
+        print("DISABLE_AUTO_G54 = 0", file=file)
         # qtplasmac has extra rs274ngc variables
         if self.d.frontend == _PD._QTPLASMAC:
             code = 21 if self.d.units == _PD._METRIC else 20
--- a/docs/src/gcode/m-code.adoc
+++ b/docs/src/gcode/m-code.adoc
@@ -67,7 +67,7 @@ to stop after each line of input anyway.
 Both of these commands have the following effects:
 
 * Change from Auto mode to MDI mode.
-* Origin offsets are set to the default (like 'G54').
+* Origin offsets are set to the default (like 'G54') unless disabled by INI setting 'DISABLE_AUTO_G54 = 1'.
 * Selected plane is set to XY plane (like 'G17').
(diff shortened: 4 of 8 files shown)


src/emc/rs274ngc/interp_internal.hh
```diff
@@ -825,6 +825,7 @@ struct setup
     bool loop_on_main_m99;
 
   int disable_g92_persistence;
+  int disable_auto_g54_on_progstop;
```
BsAtHome: This is a boolean value in a C++ program. Why are you adding it as `int` and initialize it with 0 (zero) instead of false (in interp_setup.cc)?  Wouldn't it be more appropriate to use `bool`?

docs/src/config/ini-config.adoc
```diff
@@ -635,6 +635,8 @@ The maximum number of `USER_M_PATH` directories is defined at compile time (typ:
   Warn rather than error in case of errors in O-word subroutines.
 * `DISABLE_G92_PERSISTENCE = 0` (Default: 0)
   Allow to clear the G92 offset automatically when config start-up.
+* `DISABLE_AUTO_G54_ON_PROGSTOP  = 0` (Default: 0) +
+  When set, M2,M30 and M99 will no longer automatically reset the active WCS to 'G54'.
```
BsAtHome: This is a boolean value. You should note that it is a boolean value here.
(also the other boolean values should be marked as such, but one things at a time)

Also, you should also add it to the list of boolean values to check in `src/emc/ini/linuxcnc_check_ini.py` in the list of the check_bools() function.

zz912:
Can I ask you add it in to pncconf?

here: https://github.com/LinuxCNC/linuxcnc/blob/1c36c76308e267c833ab0a8e2d071ecbe5a45448/src/emc/usr_intf/pncconf/build_INI.py#L235
add:
```
        print("# When DISABLE_AUTO_G54_ON_PROGSTOP = 1'", file=file)
        print("# M2,M30 and M99 will no longer'", file=file)
        print("# automatically reset the active WCS to 'G54'", file=file)
        print("DISABLE_AUTO_G54_ON_PROGSTOP = 0", file=file)
```

zz912:
I tested main functionality of this PR in Gmoccapy. It works.

docs/src/config/ini-config.adoc
Sigma1912: - Changed to '(bool, Default: 0)'. Happy to change that to another format if preferred.

- Added to the list in check_bools()

src/emc/rs274ngc/interp_internal.hh
Sigma1912: Indeed, changed.

(I was following the existing setting 'disable_g92_persistence')

Sigma1912:
@zz912 
Added as requested (I made the comment a bit more compact)

src/emc/usr_intf/pncconf/build_INI.py
```diff
@@ -232,7 +232,9 @@ def write_inifile(self, base):
             print("PATH_PREPEND = ./python", file=file)
             print("TOPLEVEL = ./python/toplevel.py", file=file)
             print("LOG_LEVEL = 0", file=file)
-
+        print("# Set 'DISABLE_AUTO_G54_ON_PROGSTOP = 1' to disable", file=file)
+        print("# automatic resetting of the active WCS to 'G54' on M2,M30,M99", file=file)
+        print("DISABLE_AUTO_G54_ON_PROGSTOP = 0", file=file)
```
BsAtHome: Is this supposed to be added when the frontend is gmoccapy or always (like the indent now does)?

It is hard to see how @zz912 indented the lines in the comment and it is a bit confusing how they are added to understand where they belong.

Sigma1912:
@zz912 could you please clarify whether your requested addition to pncconf was for gmoccapy only or for all guis

zz912:
I want it for all GUIs. I made wrong numbers of spaces.
Thank you for check.

src/emc/usr_intf/pncconf/build_INI.py
Sigma1912: moved those lines to before the ui specific section

andypugh:
I don't like "PROGSTOP". 
Can you make it "DISABLE_AUTO_G54_ON_M2_M30 ?
(This makes it more explicit that a g-code file wrapped in "%" won't trigger the behaviour.)

zz912:
I prefer "DISABLE_AUTO_G54".

Sigma1912:
>I don't like "PROGSTOP".
Can you make it "DISABLE_AUTO_G54_ON_M2_M30 ?
(This makes it more explicit that a g-code file wrapped in "%" won't trigger the behaviour.)

Yes, 'progstop' might indeed be a bit misleading. Note though that for a g-code file wrapped in "%" the WCS is not (and has never been) automatically switched to G54 so there isn't really much cause for confusion.

In this sense it might indeed be better to just make it   "DISABLE_AUTO_G54" and leave the details to the documentation. Opinions?

Sigma1912:
Changed to  "DISABLE_AUTO_G54" for now. Happy to change again if preferred.
Also dropped the misleading reference to M99 as that does only apply to preview (ie outside of task).

PR topic and comment have also been updated.

PR #4098, opened 2026-06-01
fix(interp): reset G52/G92 offsets on prog_stop when DISABLE_G92_PERSISTENCE=1

### LINKED ISSUE #4096: bug(interp): G52,G92 is not reset on M2,M30 regardless of DISABLE_G92_PERSISTENCE setting
'Coordinate Systems' documentation states:
https://linuxcnc.org/docs/html/gcode/coordinates.html
>Programmers that wish for Fanuc behavior, where G92 offsets are cleared at machine start and after a reset or program end, may disable G92 persistence by setting DISABLE_G92_PERSISTENCE = 1 in the [RS274NGC] section of the INI file.

However, testing with [RS274NGC]  DISABLE_G92_PERSISTENCE = 1 in axis.ini shows different behavior:

1. Start axis.ini -> Dro X and 'G92 X' are '0.000'
2. MDI: `(debug, #5210)`  ->  shows '0.000'
3. MDI: `G92 x1` -> Dro X changes to '1.000', 'G92 X' changes to '-1.000' 
4. MDI: `(debug, #5210)`  -> now shows '1.000'
5. MDI: `m2`
6. MDI: `(debug, #5210)`  ->  shows '0.000',  Dro X  and G92 X SHOULD change to '-1.000' and '0.000' respectively  but remain unchanged 
7. MDI: `g0 x0` -> tool moves to the offset position and confirms that the offset is indeed still applied instead of having been cleared on `m2`

Note that the 'INI' documentation seems to have only ever  mentioned  ' config start up':
https://linuxcnc.org/docs/html/config/ini-config.html#sub:ini:sec:rs274ngc
> DISABLE_G92_PERSISTENCE = 0 Default 0 Allow to clear the G92 offset automatically when config start-up.

Meanwhile 'interp_convert.cc' comments for 'convert_stop:
https://github.com/LinuxCNC/linuxcnc/blob/bf20900daa881f476094d6b909955c0c3df342b1/src/emc/rs274ngc/interp_convert.cc#L5124
and runs this on M2,M30,M99 if 'DISABLE_G92_PERSISTENCE = 1'
https://github.com/LinuxCNC/linuxcnc/blob/bf20900daa881f476094d6b909955c0c3df342b1/src/emc/rs274ngc/interp_convert.cc#L5256-L5260
Which seems to indicate that g92/g52 offsets were indeed intended to be reset to zero.

Tested on 2.9 and 2.10pre (current master)
This is a bug first discovered while investigating issue https://github.com/LinuxCNC/linuxcnc/issues/2908 but I think it should have it's own issue since I have just run into it again.

Sigma1912:
This fixes https://github.com/LinuxCNC/linuxcnc/issues/4096

The current code only resets the parameter values to zero but does not actually reset the applied G92 offset values.

- Adds canonical command to actually reset applied G92/G52 offset on M2,M30,M99 with [RS274NGC] 'DISABLE_G92_PERSISTENCE = 1'
- Updates INI documentation on DISABLE_G92_PERSISTENCE to reflect and agree with 'GCODE' documentation
- Modify the 'g92-offset-persistence' test to account for the added 'SET_G92_OFFSET(0.0000, 0.0000, 0.0000, 0.0000, 0.0000, 0.0000)'
- Change type to 'bool' rather than 'int'


andypugh:
Can you fix the conflicts?

Sigma1912:
Something didn't go quite right there. Let me try to do this locally.

Sigma1912:
Ready now.

Issue #3556, opened 2025-09-09
Connecting joint.n.index-enable to spindle.n.index-enable breaks spindle-synchronized motion


Sigma1912:
[edit] 
See this post further down on how to reproduce this issue in a sim config: 
https://github.com/LinuxCNC/linuxcnc/issues/3556#issuecomment-3622360845
[/edit]

Background:
Some users want to setup a joint as a spindle/rotary-axis (eg using a lathe spindle as a Caxis). The usual approach is to configure a rotary axis with homing to index and then use hal logic to disconnect the encoder signal before switching to spindle mode to avoid triggering a joint following error. The spindle has to be rehomed (using the encoder index) before the encoder can be reconnected in hal. 

Issue:
While the above method works reliably, it has been reported that synchronized motion (eg G33, G33.1) shows erratic behavior while in spindle mode:   

https://forum.linuxcnc.org/38-general-linuxcnc-questions/55116-g33-1-synchronized-tapping-problem?start=20#324268

As described in the forum thread above, I have done testing which reproduced the issue. Further testing showed that the erratic behavior seems to only appear  if 'joint.n.index-enable' (where n is the joint for the caxis) is connected to the encoder index-enable and subsequently also to spindle.n.index-enable. This is used to home the caxis to index on startup.

Testing a configuration where joint 4 is setup as rotary/spindle:
1. start config and home all (joint 4 is set to home to index)
2. testing  g33 shows broken behavior (does a rapid move to target coordinate)
3. open 'hal configuration' tool and issue HAL command: 'unlinkp joint.4.index-enable'
4. testing  g33 shows expected behavior
5. issue HAL command: 'net c-index-enable joint.4.index-enable'
6. testing  g33 shows broken behavior (does a rapid move to target coordinate) 

So a workaround seems to be to disconnect  'joint.n.index-enable' (where n is the joint for the caxis) when switching to spindle mode.

I attach my test configuration including the components used for future reference.

[Drehachse.tar.gz](https://github.com/user-attachments/files/22237631/Drehachse.tar.gz)

Some notes about the test config:
- Rotary (Caxis) is on Joint 4, uses a stepgen with settings from ini [JOINT_4] section
- Spindle also uses stepgen but with setting from ini [SPINDLE_0]
- Startup mode is rotary with joint.4 homing to index (ie joint.4.index-enable is connected in HAL)
[...]

PR #4099, opened 2026-06-01
hal: Add new types and API framework for the getter/setter change.

BsAtHome:
This is the first in a series of updates to move HAL pin and param access over to getter/setter methods and eliminate direct user access to the underlying data. This PR sets the stage and puts the type, data and access infrastructure in place.

The new data interface is neither slower nor faster than direct access. All access is expanded through inline function expansion. The new interface prevents access to the <i>wrong</i> data because access is strongly typed and clearly separates the application and implementation layers.

The new interface will replace the current HAL types when we perform the API break. The new types are (underlying type in parentheses):
* `hal_real_t` - (`rtapi_real`) floating point
* `hal_sint_t` - (`rtapi_sint`) 64-bit signed integer
* `hal_uint_t` - (`rtapi_uint`) 64-bit unsigned integer
* `hal_bool_t` - (`rtapi_bool`) boolean
* `hal_port_t` - (`rtapi_sint`) [not yet exposed because a change breaks the current API]

The new HAL types _cannot_ be dereferenced and the compiler will emit an error if you try:
```C
hal_sint_t mypin;
...
*mypin = 10; // Compiler barfs
rtapi_sint bla = *mypin; // Compiler barfs
```

You must use the proper access functions:
* `rtapi_<type> hal_get_<type>(hal_<type>_t pin)`
* `rtapi_<type> hal_set_<type>(hal_<type>_t pin, rtapi_<type> value)`
* special for compatibility (setter will _always_ write the <i>larger</i> type):
  * `rtapi_s32 hal_get_si32(hal_sint_t pin)`
  * `rtapi_s32 hal_get_si32_clamped(hal_sint_t pin)`
  * `rtapi_u32 hal_get_ui32(hal_uint_t pin)`
  * `rtapi_u32 hal_get_ui32_clamped(hal_uint_t pin)`
  * `rtapi_s32 hal_set_si32(hal_sint_t pin, rtapi_s32 value)`
  * `rtapi_u32 hal_set_ui32(hal_uint_t pin, rtapi_u32 value)`

This change also makes it clear that <i>setting</i> a pin is not possible with simple assignment. The setter and getter are not valid lvalues. Both can be used as rvalues. The 32-bit compatibility works as expected and is transparent for the user API. All current code will be updated to use the `si32` and `ui32` variants where appropriate. Then, when all code is upgraded, we can start fixing the code to be 64-bit clean. The `*_clamped()` versions read the larger type and convert to the smaller type with min/max bounds.

The distinction 32/64 bit will be gone completely in the underlying data once all code uses the getter/setter access. That is the point where the user relevant code can be changed and break the API to eliminate userspace visible 32-bit values (i.e. HAL_S32, HAL_U32, HAL_S64, HAL_U64 type indicators will vanish and replaced by HAL_SINT and HAL_UINT).

A new set of pin/param creation functions have been made that use the underlying old API for compatibility until we do the full API break. New pins and params are created with:
* `int hal_pin_new_<type>(int comp_id, hal_pdir_t direction, hal_<type>_t *pinref, rtapi_<type> dflt, const char *namefmt, ...)`
* `int hal_param_new_<type>(int comp_id, hal_pdir_t direction, hal_<type>_t *paramref, rtapi_<type> dflt, const char *namefmt, ...)`

There are two additional pin/param new functions that use the `si32` and `ui32` markers. These are there for compatibility with the current system such that we can use both APIs concurrently until all code has been updated/fixed/ported. They can be eliminated once all code is 64-bit clean with respect to pin and param use and access.

The underlying HAL pin/param structures have been adapted to match each other's layout. Access to params are treated the same as pins from the user's (component) perspective and the underlying structure types will be merged in a future PR. This also allows to merge the lists and use a RB-tree for faster access from the python interface (halmodule).

The `halcompile` program, to create components, has been updated to support the new types. It is an <i>addition</i> to the status quo so we can use both systems while we migrate. Using the new types is as easy as:
```
pin real in input;
pin real out output;
;;
output_set(input);
```
Note that <i>setting</i> a pin must be done using the `<pinname>_set()` format. Reading a pin is transparant in halcompile compiled programs (just like today). An update for all components is already in the pipeline to be submitted after this PR is merged. Halcompile can drop the old style types once everything is moved over and we do the API break.


src/hal/hal.h
```diff
@@ -339,6 +373,182 @@ typedef union {
...
+// FIXME: Get rid of the 32-bit types when we have upgraded everything using
+// getter/setter access only so we have guaranteed content.
+#define __HAL_MAPPED_TYPE union __hal_mapped_type { \
+        volatile rtapi_bool _b; \
+        volatile rtapi_s32  _ss; \
+        volatile rtapi_u32  _su; \
+        volatile rtapi_sint _s; \
+        volatile rtapi_uint _u; \
+        volatile rtapi_real _r; \
+    }
```
grandixximo, line 424: Small thought on the `__hal_mapped_type` union: since it gets type-punned against` hal_data_u` (a different union type), strict aliasing technically applies. The RT side is covered because `RTFLAGS` adds `-fno-strict-aliasing`, but the userspace ULAPI build runs at `-O2` without it, so there the access leans on the `volatile` members to stay well-behaved. Would it be worth tagging the union with `__attribute__((__may_alias__))` to make it portably safe in both builds regardless of flags? Happy to defer if you think the `volatile` guarantee already covers it.

src/hal/hal_lib.c
```diff
@@ -1616,6 +1683,102 @@ int hal_param_new(const char *name, hal_type_t type, hal_param_dir_t dir, void *
...
+    return hal_param_new_anyapi(name, type, dir, data_addr, comp_id, 1);
+}
+
+// *** New interface ***
+
+int hal_param_new_bool(int compid, hal_pdir_t dir, hal_bool_t *ref, bool def, const char *fmt, ...)
+{
+    va_list ap;
+    va_start(ap, fmt);
+    int ret = hal_param_new_newapi(HAL_BIT, dir, (void**)ref, compid, fmt, ap);
```
grandixximo, line 1723: used `(void**)ref` here and `ref` in the others, unify the style?

src/hal/utils/halcompile.g
```diff
@@ -475,18 +493,28 @@ static int comp_id;
...
-            if value is not None:
+            if not isnewtype and value is not None:
                 print("    inst->%s_p[j] = %s;" % (to_c(name), value), file=f)
             print("    }", file=f)
         else:
-            print("    r = hal_param_%s_newf(%s, &(inst->%s_p), comp_id," % (
-                type_, dirmap[dir_], to_c(name)), file=f)
+            if isnewtype:
+                dflt = "%s" % value if value is not None else "0"
+                print("        r = hal_param_new_%s(comp_id, %s, &(inst->%s_p), %s," % (
```
grandixximo: these should be back to 4 spaces, they are outside the `for` block...

grandixximo (commented):
After this all out-of-tree components must be recompiled, add a note in the commit message? Not that anyone reads those...

src/hal/hal_lib.c
BsAtHome: > used `(void**)ref` here and `ref` in the others, unify the style?

Only in the boolean version the compiler sees a size difference and complains. All others are implicit.

All of this will vanish again. These call the old API now, but will eventually be the only ones standing where casting like this is no longer necessary.

src/hal/hal.h
BsAtHome: You are right, technically (the right kind of right), and understand the problems. However, the implementation has some guarantees that should mitigate the problem. Also,  the compiler is not complaining ;-)

We know that all `hal_malloc` allocated `hal_data_u` units are always guaranteed at a 8-byte boudary (always enforced because the alloc implementation now enforces minimum size of 8). This also includes the units in the pin/param/signal structures, which also are properly aligned afaik.

We also know that the the new `hal_<type>_t` are actually pointers into hal memory to the internal `hal_data_u`, so they should always be properly aligned because they are supplied by the HAL API. This is now also true for params because they are treated the same as pins.

The only (unaligned) aliasing problem would then arise from someone punning a pointer that is not supplied by the new API. That would be an error because access to that part should not be done anymore. Punning the new `hal_<type>_t` types is possible, but risky because they are opaque pointers (to structures that do not exist). If someone does type-pun to/from the new hal types, then they are already in very muddy waters and I sincerely hope they pay for it with a crash. It is exactly this coding behaviour we want to avoid.

BTW, the public `hal_data_u` type will have to be retired in the end.

src/hal/utils/halcompile.g
BsAtHome: Fixed.

andypugh:
> The new HAL types _cannot_ be dereferenced and the compiler will emit an error if you try:

Does this cause any problems for Smart Serial or any other driver that determines pin types at load-time? 

eg: https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/drivers/mesa-hostmot2/sserial.c#L350

Where p is dereferenced, and than dereferenced again in line 359. 

(Because smart-serial has variable-size arrays of HAL pins, because the hardware itself defines the pin names)

BsAtHome:
> > The new HAL types _cannot_ be dereferenced and the compiler will emit an error if you try:
> 
> Does this cause any problems for Smart Serial or any other driver that determines pin types at load-time?
> (Because smart-serial has variable-size arrays of HAL pins, because the hardware itself defines the pin names)

Should not be any problem. The construct uses a union of old hal types and is IMO wrongly implemented (also has invalid types `long long s64_written` and `long long s64_param`). It should have used the `hal_data_u` type in the first place.

Run-time determined types already require you to multiplex/demultiplex the reference/dereference at run-time. This still needs to be done with the new system. There is no actual change there, just written in a different way. BTW, the hm2_modbus component also uses run-time determined types and uses the types of multiplexing I am referring to.

The new system has a `hal_refs_u`, a union of the new hal types, that is created to support the run-time type determination.

For sserial (and most of hostmot2), it will just move over to the new access patterns. It'll be some (more) work, but no blockers have been detected.

The change for parameters will be a bit more complex, but that should not pose any real problem. I have not seen any constructs in the code that would prevent us from migrating everything to setter/getter and unifying the underlying pin/param API.

The biggest problems detected:
* the pyhal module using CDLL access. Already retired.
* `hal.hh` bypasses any usual access and is another (re)implementation of bad access patterns. Luckily, it is only used in `emc/task/taskclass.{cc,hh}` and is already retired in my tree.
* the `hal/components/xhc-hb04.cc` has a simulate mode that replaces the hal memory with local memory. That cannot work anymore. The simulate allocation style has to go. You can always simulate. You just need to have the hal shared memory segment available and that is no problem at all. So, there is no need for the complexity.

There may be other details lurking, but so far none have been of any blocking character.

BsAtHome:
Just force pushed a cppcheck update. It turns out that cppcheck "older" wrongly trips on an added `#error` directive because it fails to evaluate a preprocessor directive correctly. This would trip CI as Debian/Ubuntu are on the older version. My local version is 2.19.1 and has no such problem.

BTW, Debian trixie running on 2.17.1 has one more false positive, but that is not seen in CI. Needs to be sorted out in another PR.

grandixximo:
With the stricter CI have to be more picky, unrelated off topic about the docs, should we also force the docs to build clean with the CI? Like second build of docs should be nothing to do, then pass the CI?

BsAtHome:
> With the stricter CI have to be more picky, unrelated off topic about the docs, should we also force the docs to build clean with the CI? Like second build of docs should be nothing to do, then pass the CI?

That would be a whole new can of worms... Let us discuss that somewhere else.

rene-dev:
> * `hal.hh` bypasses any usual access and is another (re)implementation of bad access patterns. Luckily, it is only used in `emc/task/taskclass.{cc,hh}` and is already retired in my tree.

Dont do this. hal.hh is my attempt to create a cpp interface, which is then used in the pybind11 branch to create the python bindings automatically. this already works, but needs cleaning.

rene-dev:
https://github.com/rene-dev/linuxcnc/blob/pybind11/src/hal/halmodule.cc

grandixximo:
Rather than retire `hal.hh` vs keep it, could we land on re-basing it on the new getter/setter API? `hal_pin<T>` holds the new handle and calls `hal_get_*`/`hal_set_*` internally, so Rene keeps the C++ interface for pybind11 and Bertho gets the no-bypass invariant.

BsAtHome:
> > * `hal.hh` bypasses any usual access and is another (re)implementation of bad access patterns. Luckily, it is only used in `emc/task/taskclass.{cc,hh}` and is already retired in my tree.
> 
> Dont do this. hal.hh is my attempt to create a cpp interface, which is then used in the pybind11 branch to create the python bindings automatically. this already works, but needs cleaning.

The hal.hh is incompatible with HAL getter/setter because it accesses HAL private constructs it should not be accessing and is using assumptions on the underlying structure. It is also incomplete and uses the wrong types. Every HAL pin/param access needs to go through the proper API and there should not be exceptions. I'm not sure you can fix the pin/param access in a proper way using the current constructs because the types have changed to non-dereferencable opaque pointers (and you are not allowed to re-implement access to the underlying memory to prevent the same problem as we have today from happening again).

If you want to make a C++ wrapper, then it must be transparently wrapping and using the C API. It must not bypass it. You also want to use the exact same HAL types as basis for pins/params and not a newly defined C++ template construct. It may be better to model it somewhat after halmodule, although there are some (solvable) problems there as well. However, I do not see the advantage in having it abstracted too much so you have to multiplex/demultiplex the types at every pin/param access as is required in halmodule. Then it is easier to use the C API. Besides, it may not always be a good thing to hide function call by call type selection in the code.

BsAtHome:
> Rather than retire `hal.hh` vs keep it, could we land on re-basing it on the new getter/setter API? `hal_pin<T>` holds the new handle and calls `hal_get_*`/`hal_set_*` internally, so Rene keeps the C++ interface for pybind11 and Bertho gets the no-bypass invariant.

You do not want a template type for HAL pins and params. It would be rather pointless because the template has to expand the HAL type directly as its passed type (remember it is an opaque type). The template cannot use a generic accessor in HAL types because there is no generic accessor. All acessors are HAL type bound. You'd need to demutiplex on basis of a saved type indicator to call the proper access function. So, instead you are using more storage for no gain and worse performance?

If you want to wrap pins and params in a C++ class, then you will need one explicitly for each HAL type where both lvalue and rvalue overloads are present and use the proper `hal_set_xxxx()` and `hal_get_xxxx()`. This will still be not enough because we have the 32/64 bit legacy. You'd have to make exceptions in the `hal_sint_t` and `hal_uint_t` wrapper classes because there are multiple accessors defined to ensure compatibility with the current code base.

grandixximo:
Fair on the 32/64-bit point, I missed that. A single `operator T()` can't express the `sint`/`uint` accessor variants, so those wrappers need explicit methods (`get()` / `get32()` / `get32_clamped()` and the matching setters) rather than one operator. Agreed.

On the template though, I think we're picturing different things. I'm not suggesting a value-typed or type-erased template, which would need a stored type tag and runtime demultiplex like you describe. I mean a template keyed on the opaque handle type, resolved at compile time:

```cpp
template<> struct hal_traits<hal_real_t> {
    static rtapi_real get(hal_real_t h)               { return hal_get_real(h); }
    static rtapi_real set(hal_real_t h, rtapi_real v) { return hal_set_real(h, v); }
};
```

`hal_pin<hal_real_t>` then stores only the 8-byte handle, `operator rtapi_real()` calls `hal_get_real(h)`, all inline. No type indicator, no branch, no extra storage, same machine code as the bare C call. That is exactly your "one explicit class per HAL type with lvalue/rvalue overloads using the proper `hal_set/get_xxxx()`", just generated once instead of copy-pasted. The template is the DRY way to produce those classes, not an alternative to them.

The demultiplex cost is real but only for genuinely runtime-typed access (the Python/pybind11 path), which has to happen in any design since the C API has no generic accessor, halmodule included. So I'd split it: zero-cost per-type wrappers for component code, and a variant/registry layer on top for the runtime-typed Python bindings. Both go through the C API, neither bypasses it.

That keeps Rene's C++ interface alive without reintroducing the access problems you're removing. Does that reframing change the "retire it" call?

BsAtHome:
Two problems: 1) the hal_xxxx_t lives in HAL memory and 2) the C++ version wants to put pins/params into a map. The first is solved by doing a hal_malloc on the target, but the second is not solvable using templates. You need the type-information of the class/pin retained in the map for it to be dereferenced using the proper accessor (and besides, it incurs an extra indirection).

There is a genuine need for unionized hal types in, to name a few, halmodule, hm2/sserial and hm2_modbus. But these <i>always</i> use stored type information to multiplex/demultiplex. If you want a C++ wrapper, then it must store type information. There is no free ride to hal memory anymore, by design. You either know the pin/param type by using the hal types or you use a unionized version and store the type to multiplex/demultiplex.

grandixximo:
Agreed, you're right. The map is the deciding factor: a string-keyed `std::map` of heterogeneous pins forces type erasure, so it has to store the type and demultiplex on access. Templates only buy you zero-cost typing for compile-time-known named pins, which is not what hal.hh does. So my template framing doesn't apply here.

That still points at a rehab rather than a removal: rebuild hal.hh as a union + stored-type-tag wrapper over the proper `hal_get_*`/`hal_set_*` (using `hal_refs_u`), modeled on halmodule, no memory bypass. Runtime-typed by necessity, but going through the C API. That keeps Rene's C++ interface for the pybind11 bindings while satisfying the no-exceptions access rule. Does that work as the direction for hal.hh?

grandixximo:
@rene-dev the discussion has been ongoing on the mailing list, if you are not there you might have missed it.

https://sourceforge.net/p/emc/mailman/emc-developers/?viewmonth=202606

After reading, are you still blocking this?

rene-dev:
Yes, I have read it, and then got sidetracked on the absolute encoder bug. Im not blocking, I just think it needs some more thought. I have prepared some statistics.

rene-dev:
My preferred order of doing things would be:

1. get noparam branch working, and merge that. this massively simplifies a lot of things.
2. define a c++ api, which can be used by all the c++ code( mostly task and halui)
3. use this api to automatically create python bindings, which is used by all the python code, and UIs(this already works in the pybind11 branch)
4. not necessary last, but in the process think about which types we actually need, this is just too many. https://github.com/LinuxCNC/linuxcnc/blob/master/src/hal/components/Submakefile#L6-L34

I have looked at a few typical linuxcnc configurations.
386 pins are bits
359 pins are floats
4 pins are u32
66 pins are s32

the 4 unsigned pins are bitmasks and tool number, and a time.
none of those are linked.
of the 66 signed pins, if you leave out the time pins, 45 are left.
2 of those are used, for the tool number.

rene-dev:
neither here, nor on the ML anyone has provided an actual use case for:
- an unsigned integer pin
- any integer pin

BsAtHome:
> neither here, nor on the ML anyone has provided an actual use case for:
>     * an unsigned integer pin
>     * any integer pin

This is just as irrelevant as the floating point discussion. The suggested changes are not altering the user-seen API and would actually enable you to do so if you so please at some point in time. The changes are internal changes to clean up the mess where everybody is accessing data they should not be accessing. This has been clearly outlined and described in the ML post.

This cleanup needs to be performed _regardless_ of whether or not you want no parameters, change types or have a c++ interface. This cleanup is a result from the suggestion that there should not be both 32 and 64 bit types. And this seemingly "small" change cannot be done without performing a complete and thorough cleanup (see #3286).

FWIW, the case for any type of integer _at the hardware side_ are given by the fact that we interface with hardware that uses integer types. That is the place where we need them. But again, this is internal API used by drivers and that is what is being addressed with the getter/setter infrastructure.
How the underlying structure evolves or what users ultimately will see is _not_ part of these changes or discussion. Any evolution can only happen once you have a code-base that can support it without falling apart. The necessary infrastructure as outlined allows you to discuss and make these decisions <i>in the future</i>.

grandixximo:
> none of those are linked

The statistics need a wider sample. That holds for the u32 pins in your sample, but the s32 count has the same sampling gap: none of the surveyed configs had a pendant. Two integer pin classes are linked in a large fraction of real configs:

* `encoder.N.counts => joint.N.jog-counts` (s32 to s32): every MPG pendant configuration. In-tree alone: woodpecker mpg.hal, touchy, shuttle, xhc-hb04, and pico pendant.hal (rawcounts via ilowpass, s32 throughout). Motion's jog protocol is an exact per-cycle count delta (control.c:1066).
* `iocontrol.0.tool-prep-number`: axis_manualtoolchange.hal is pulled in by 50 in-tree inis, and smithy/924.hal nets it into classicladder s32 word logic on shipped commercial machines.

I will concede up front: every 32-bit value in these chains fits exactly in a double. The case for integer pins is not numeric capacity, it is semantics:

1. Unsigned: hm2_modbus (in 2.9) maps raw device registers to pins, including 64-bit unsigned compounds (`hal_pin_u64_newf`, plus a u64 `.offset` tare doing modulo-2^64 subtraction). A totalizer register above 2^53 loses bits on a float pin; the range [2^63, 2^64) reads negative on s64. The pin is the user-facing register image, so "keep it driver-internal" does not apply.
2. Bit patterns: hm2 `raw.write-data` and bitwise (u32 pins), plus lut5 (stepconf generates `setp lut5.0.function 0x10000`, a u32 param). On an integer pin, NaN and 2.5 are unrepresentable; today `setp <int-pin> nan` is rejected while `setp <float-pin> nan` succeeds, and `(int32_t)NaN` is 0x80000000 straight into an FPGA register. No driver checks isnan; the type is the guard.

greatEndian:
Hello gentlemen,

I think erasing u32 will bring us much more pain than gain.

Many natural industrial standard protocols—like EtherCAT, CANopen, and Modbus—rely natively on 32-bit unsigned integers for their data structures. Forcing these into s32 or float introduces unnecessary casting overhead and potential bugs.

When dealing with hardware registers, Status/Control words, or bitmasks, utilizing the 31st bit on an s32 turns the value negative, which makes bitwise operations dangerous and confusing. Furthermore, dealing with absolute encoder rollovers is natural with u32, but a headache with s32.

Thank you @BsAtHome for doing the architectural work to keep the u32 compatibility layers alive in the new RTAPI. Preserving this is absolutely vital for those of us interfacing directly with modern industrial hardware!

BsAtHome:
The changes will not eliminate u32/s32 as seen from a component/driver's view. The getters and setters support that view and are fully compatible.

The getter/setter system is the result of the discussion not to have both 32-bit and 64-bit integer HAL types and only have 64-bit types. The getter/setter infrastructure allows for better abstraction. It decouples the backing storage and access in the components/drivers with a real abstraction. It does not incur a performance penalty and adds strong typing to let the compiler check for faulty constructs.

The getter/setter changes to HAL are just a part of a more thorough cleanup of the code and allow HAL to be the abstraction layer for both userspace and codespace. So, yes, this is (a lot of) architectural work and a bit broader than just the HAL types.

If it is agreed to do make accept this PR...
This happens in the following set of patches (will be much more fine grained in practice and some in different order):
* Access to HAL backing storage is disallowed and has to go through the getter/setter. All components' and programs' pins and params are updated to use the getter and setter infrastructure. This change works in parallel with the current system for compatibility so we don't break the current API, yet.
* Update components and drivers to become 64-bit clean <i>internally</i>. Currently most drivers are hard-wired to the small type but could very well use a generic integer type. These instances need to be cleaned up. Places where 32-bit is required is already supported by the getters and setters but need to support the generic pin types and not perform type conversions that will result in data problems. This step should eliminate <i>the necessity</i> of pins/params created as U32/S32 pins and instead <i>allow</i> them to be created as the new generic UINT/SINT types.
* Remove U32/S32 and U64/S64 and replace with signed/unsigned (SINT/UINT) HAL type. The external view will now only see signed/unsigned. Internally, all old HAL types are removed. The getters and setters still have their 32-bit versions (introduced in the first change) and they will do the translation transparently. Also BIT will become BOOL and FLOAT becomes REAL. This breaks the current HAL API and is planned as a clean cut. The old API would have been deprecated for "some time" before this happens. The API break removes the old API.
* Cleanup support programs not to use HAL private data and code and disallow inclusion of hal_priv.h. A query API will be used to support all access to HAL's inner secrets without exposing the storage. The new API is needed to perform user-space control (halmodule, halcmd, halscope, halmeter, halrmt and maybe more). This step cleanly abstracts HAL into a library and the backing storage/inner workings become fully opaque for library users.
* <sub>Internal change: Replace the pin/param backing storage's linear lists with a red/black tree so that access becomes O(logN) instead of O(N). A normal system with 1000 pins+params will then only require 10 compares to find the item instead of 500 compares (on average) today. This helps python code that often addresses pins by name. Previous changes already require unification of pins and params, which now only differ by attribute. The backing storage has been unified in previous changes as a necessity of the getter/setter changes.</sub>

grandixximo:
@BsAtHome I am on board with the getter/setter direction. I will not make this Sunday's meeting, but FWIW my vote is yes.

I do not read @andypugh as blocking, @rene-dev has said himself he is not, and @rmu75 seems not fully on board but not blocking either. So I would like to keep this moving.

@rene-dev on noparam: I want to be upfront that I am questioning the goal here, and I know you have already put real work into that branch, so I am not raising this lightly. Happy to help with whichever direction we land on. But first I would like a good argument for why removal should be done at all. What does it actually simplify? AFAICT the getter/setter work already unifies pin and param storage. Once that lands we can allow params to be connected as a small additive patch rather than full removal, and #4147 already keeps them in one view. So we end up with connectable and well visible params without removing anything.

IMO removal can also be detrimental. Take `bldc.N.poles`: as a param it signals "you are not supposed to change this at runtime." It is motor config that scales the commutation angle every cycle, wire it to a live signal and commutation desyncs. The param type carries that intent. Erase it and everything becomes connectable, including the things that _should_ not be.

So before **more** effort is put into noparam, I would like to understand what it buys that connectable, well visible params do not. I genuinely think the additive path is better, but I am completely open to a counter-argument that shows otherwise.

rodw-au:
I still think this is heading in the wrong direction. HAL stands for HARDWARE ABSTRACTION LAYER. 

C/C++ was chosen for the interface because it can get down to very low levels to match hardware and its stood the test of time. To move away to a higher level of abstraction makes no sense to me. This isn't the place to replicate typeless variables found in other high level languages.

The Development team can look at the code base as much as it likes but it can have no confidence that it knows how HAL has been applied in the wild. I for one have code that uses a complex bit mapped union over a 32 bit variable mapped to hardware registers. Ultimately, I will submit it as a PR. To add additional abstraction between the hardware and HAL makes no sense to me. In some instances, we need every available nanosecond to process the hardware within servo thread constraints. Surely higher levels of abstraction are going to add to the processing time on the servo thread.

I'm really quite apprehensive about this proposal and fear it will be the beginning of the end for the LinuxCNC project. Please focus limited development resources on useful stuff, not unecessary fluff.

grandixximo:
> I'm really quite apprehensive about this proposal and fear it will be the beginning of the end for the LinuxCNC project. Please focus limited development resources on useful stuff, not unecessary fluff.

The getter/setter is modern approach, and is the proper cleanup, and does not effect the hot path, the "everything is a float" does have some incredibly minor effects, which I agree we should avoid.

BsAtHome:
> I still think this is heading in the wrong direction. HAL stands for HARDWARE ABSTRACTION LAYER.

Yes, it is a hardware abstraction layer seen from the .hal-file system. It allows you to load and connect modules, components and drivers using loadrt, addf, net,stp, sets, etc... That is the HAL as you know it from the user side. That is all normal users will ever see. They write .hal files, load realtime components, make nets and connect pins.

That is _not_ what is being addressed here in this PR.

This PR only addresses the side where data is pushed and pulled from the hardware into and out of the pins and params.

> C/C++ was chosen for the interface because it can get down to very low levels to match hardware and its stood the test of time.

For starters, there is no C++ at the core of the HAL library. It is a pure C library because it needs to support kernel modules. Any C++ interface is part of the <i>other</i> side and lives in user-space supporting user-land. All of the C++ stuff lives in non-real-time land (like the USB pendent interfaces). What actually talks to hardware in realtime are components and especially the drivers. This is the code that talks to hardware and is time critical.

> To move away to a higher level of abstraction makes no sense to me. This isn't the place to replicate typeless variables found in other high level languages.

You confuse "abstraction" with "slow(er)". An abstraction is nothing more than a set of rules you follow to achieve your objective. You do not need to know the underlying details if you abide by the supplied API. The underlying operations may change as systems evolve, but the API is allowed to be stable. That means: the programmer will always see a <i>consistent</i> interface when you provide the proper abstraction. It has nothing to do with speed.

The suggested changes here make the rules to access pin's and param's data very strict and the underlying data structures opaque. It is not a concern for the component/driver writer how pin data is transferred at the lowest level. The only thing you need to know is that you can do so consistently.
One reason to do this getter/setter abstraction is to make it _**very**_ clear to programmers that accessing pins/params comes with a penalty. This is penalty is also present today, but the fewest realize that. Accessing pin/param data is very slow and the compiler is not allowed to do a large set of optimizations because the data is a shared entity and marked volatile. Code that re-reads or re-writes a pin value forces the CPU to fetch/store that data from/to main memory on <i>every</i> access. Indiscriminate multiple access is a bad programming pattern seen very often in the current codebase. Pins and params are used as <i>very expensive</i> variables. Having to write that access as a function call is a stark reminder for the programmer that it is a special operation, which it is. And, please note, all of this is designed so that the compiler generates code that is exactly the same or better than what you do today.


> The Development team can look at the code base as much as it likes but it can have no confidence that it knows how HAL has been applied in the wild.

And that is also a real problem. Indiscriminate abuse of hal types for variables has been done throughout. This impacts performance considerably and the proposed changes put a stopper to that abuse.

> I for one have code that uses a complex bit mapped union over a 32 bit variable mapped to hardware registers. Ultimately, I will submit it as a PR. To add additional abstraction between the hardware and HAL makes no sense to me. In some instances, we need every available nanosecond to process the hardware within servo thread constraints. Surely higher levels of abstraction are going to add to the processing time on the servo thread.

You confuse the <i>hardware</i> side and the <i>pin data</i> interface side. Your 32-bit mapped register is on the hardware side. You can have that data exposed as one pin or as multiple pins and multiplex them in the component. How you do that translation is up to you and a choice how you interface with other components.

There is _no_ extra processing time involved with the proposed changes.


> I'm really quite apprehensive about this proposal and fear it will be the beginning of the end for the LinuxCNC project. Please focus limited development resources on useful stuff, not unecessary fluff.

I 'm note sure how to respond to that.

grandixximo:
@BsAtHome a question on where this leads, since you have the clearest picture of the getter/setter endgame.

A bug I just fixed (#3556) has me thinking this PR might be the right moment to also sharpen the HAL object kinds. In #3556, homing wrote `joint.N.index-enable` (a `HAL_IO` pin) every servo cycle, clamping it. Net that onto `spindle.N.index-enable` (a normal way to home a spindle/C-axis to index) and homing stomps the spindle's index request every cycle, so G33/G33.1 rapids instead of waiting. The link is fully legal: two `HAL_IO` pins on one signal is allowed (`bidirs` is uncapped in `hal_lib.c`), with no warning.

To me that bug exposes a semantic gap: `HAL_IO` means two unrelated things the direction does not distinguish.

1. A cooperative handshake/bus (`index-enable`, stepgen `latch-enable`): many writers share one wire, each writes only on its own edge. Safe to link IO-to-IO.
2. An "IO for convenience" pin (`scale`, tuning pins, `out-io`): really single-writer and level-driven, IO only so you can `setp` it before wiring.

Opposite contracts, identical type. HAL has no "not driving" state and no contention detection, so it cannot catch a level-driver wired onto a cooperative bus. The protection is purely social, and #3556 is what happens when a component quietly breaks it.

That is also why the suggestion to drop parameters worries me. Merging the underlying storage (as this PR does) seems clearly right; but erasing the concept would remove the one place HAL says "configuration, single owner, not a signal," which feels like the wrong direction when we are already short of that distinction.

My question: does your larger getter/setter plan have room to make the object kinds sharper at the same time? Once access goes through `hal_set_*`/`hal_get_*` and assignment is no longer an lvalue, the setter looks like the natural hook for the semantics we lack today: edge-only/release discipline plus contention detection for bus pins, single-writer enforcement for settable params, and a real config-vs-signal split that survives the type merge. Is that something you see fitting cleanly into the path you have planned, or does it cut against the simplification you are aiming for? You know the endgame better than I do, so I would value your read on whether "what are the real HAL object kinds" is a question this PR can answer, or one best kept separate.

BsAtHome:
> Once access goes through `hal_set_*`/`hal_get_*` and assignment is no longer an lvalue, the setter looks like the natural hook for the semantics we lack today: edge-only/release discipline plus contention detection for bus pins, single-writer enforcement for settable params, and a real config-vs-signal split that survives the type merge. Is that something you see fitting cleanly into the path you have planned, or does it cut against the simplification you are aiming for? You know the endgame better than I do, so I would value your read on whether "what are the real HAL object kinds" is a question this PR can answer, or one best kept separate.

There is no clear ownership and directional enforcement in any HAL pin or param construct. Only connecting signals has a check to prevent multiple (HAL_OUT) writers. This is by design and cannot be solved easily. The getters/setters could, in principle, be extended to include more enforcement of any desired kind, but that would come at a _very significant computational cost_ and <i>I do not recommend that</i>. 

Besides, what do you want to do when an erroneous construct is detected? You can't bail, you can't really warn (or you'd be warning at the servo-/base-thread rate). There is no point in test-while-running. Remember, this is real-time code.

What you can do is to export the pin/param/signal-trees and do an offline analysis. However, you still cannot catch the "no driver" or "wrong driver" situation in a HAL_IO collection because it is <i>coded in software</i> how that is driven, not the pin-label. There is no way you can know unless you look at the code.

grandixximo:
Fair, it will have to be checked in code, still some IO might be worth changing to parameters if we are keeping them, I think the driver authors reached for the wrong construct, but there really isn't a good one...

BsAtHome:
> still some IO might be worth changing to parameters if we are keeping them, I think the driver authors reached for the wrong construct, but there really isn't a good one...

Although the idea behind HAL_IO is interesting, it is a real problem in practice because you have no chance of checking validity when you create a net. I'm not even sure you need bidirectional pins or bus-like nets.

hdiethelm:
I went quickly trough your PR. Some feedback from my side:

- Getter / Setter are a good thing. `*pin_value += 5` completely hides that you are just reading, incrementing and writing a volatile shared memory variable and it might even be an IN pin.
- Params are a good thing: This way, you discriminate between what makes sense to connect between components and what should be set only once. It does not have to be different in the underlying data, it's more a concept to show what this value should be used for. However, sometimes it is used wrong. servo-thread.tmax (maxtime) is for sure no param, it changes all the time...
- 32 bit values are often useful for counters. You can copy the HW value directly into it and don't have to care about overflows (assuming the other side is aware and compensates). Dropping 32 bit would make it necessary to always do overflow checked updates of a 64 bit value based on a 32 bit hardware value. This can be done by adding a hal_set_ui32_to_ui64_unwrap(). Clamping is often wrong, especially for counters. Imagine you have a 24 bit encoder on the spindle. It will overflow every 256 revolutions.
- IO pins make no sense, either it is IN or OUT. IO will break all the time.
- Input/Output can not be discriminated with the actual functions as I understand it, right? So you can create an IN pin and write to it? Checking it at runtime would be wasteful. However, by having hal_bool_out_t / hal_bool_in_t and hal_pin_new_bool_out() / hal_set_bool(hal_bool_out_t ref, rtapi_bool val) ... the compiler can check the typing. No need for anything else. But this would involve even more change.
- I don't like sint / uint / real naming. Proper type names should always have the bit size in, otherwise it's the same mess as with int in C, you never know what size it is . One day a 128 bit PC comes along and breaks everything again... The C standard says an int is >= 16 bits what ever it is.

BsAtHome:
> * 32 bit values are often useful for counters. You can copy the HW value directly into it and don't have to care about overflows. Dropping 32 bit would make it necessary to always do overflow checked updates of a 64 bit value based on a 32 bit hardware value. This can be done by adding a hal_set_ui32_to_ui64_unwrap(). Clamping is often wrong, especially for counters. Imagine you have a 24 bit encoder on the spindle. It will overflow every 256 revolutions.

There is a difference between what you <i>use</i> and what you <i>transport</i>. The point being, the getter/setter change is a transport layer abstraction. What you put in or take out is up to you. However, the underlying structure is created such that writing a 32-bit value will create a _valid_ transport-layer value (which incidentally is 64-bit). This way you can connect any integer pin to any integer signal, regardless whether the driver want to use the extra bits or not.
The part you are addressing is the "64-bit clean" problem (and probably in the future the "128-bit clean" problem). That is a different problem, though being an important thing to resolve. I mention it that it needs to be checked/resolved in a lot of code in my point plan.

> * IO pins make no sense, either it is IN or OUT. IO will break all the time.

Indeed. They should be removed. But that could be done later. Luckily, they are not used very often. They do not change anything about this PR or the getter/setter infrastructure.

> * Input/Output can not be discriminated with the actual functions as I understand it, right? So you can create an IN pin and write to it? Checking it at runtime would be wasteful. However, by having hal_bool_out_t / hal_bool_in_t and hal_pin_new_bool_out() / hal_set_bool(hal_bool_out_t ref, rtapi_bool val) ... the compiler can check the typing. No need for anything else. But this would involve even more change.

I looked into this and rejected it... The problem is that input and output is not that clear cut as you like it to be. A strictly write-only type cannot be read, thus cannot be incremented or decremented (disregarding the atomicity discussion for a moment).

The code is also not very nice to look at when you see pins, params and signal be used as variables. That fact alone would require a far larger rewrite when compounded with true read/write isolation. So, yes, you are right that the compiler can help. And, no, it is currently a hell to enforce.

> * I don't like sint / uint / real naming. Proper type names should always have the bit size in, otherwise it's the same mess as with int in C, you never know what size it is . One day a 128 bit PC comes along and breaks everything again... The C standard says an int is >= 16 bits what ever it is.

The whole point here is that the underlying type stays opaque. We have multiple types at the moment with named sizes and they are making things very difficult already. The whole getter/setter construct is to *prevent* you to care about the underlying sizes and remove S32/S64/U32/S64. The user may assume that the transport layer will cover your arse for transporting enough bits.

[Note: the <i>float</i> naming has led to a lot of confusion]

The sizes only matter in the conversion layer. That is why there are getters and setters with the 32-bit moniker. If, in the future, we change the underlying sizes, then sint and uint will automatically follow. In that event, in the future, we could do a global rename of the current get/set on sint/uint to become get/set ui64/si64 and change the underlying sizes to 128 bits. That renaming trick can be done without problems because of the abstraction. In the current situation today, where code accesses various sizes simultaneously, that is impossible. But when the current code is fixed well, then it should not matter at all when the size <i>increases</i>.

I'm not for a renaming strategy because the problem is usually not increasing the size. Besides, portable code adapts. Having to deal with systems that have a smaller size than what your code expects and code using indiscriminate non-portable constructs is the real problem. The "64-bit clean" problem is because people often code with assumptions instead of proper generics. If they used the proper types in the first place, then it would mostly be a non-issue. However, they did not do that and now we pay the price by having to read through the code and fix it. But by fixing it, we can also reduce the problems for any future event.
You know, writing portable code is possible. It is just hard work ;-)

hdiethelm:
To not make the thread to long:
- Writing 32 bit to 64 bit shm and reading 32 bit is completely fine. Reading it as 64 bit would need sign extension based on the type.
- Yes, pin++ won't work with write only. However, reading OUT pin's is somewhat stupid to begin with, you better have a local variable to increment and only access the shm once.
- Most bit operations should be fine but you must know the minimum size to work with. This can be specified like sint / uint >= 64 bits. There is one exception: When you have counters like ns / encoder on 32 bits, you have to care about overflows. 64 bits solves this, not even the fastest CPU clocks can overflow that before the CPU and the user already died some time ago... ;-) So rethinking this, yes, drop all 32 bit stuff and use only sint / uint / real / bool in the future.

I see you invested a lot of work in it and I just spent a few hours trying to understand it... ;-)

BTW: Are you aware of how modern CPU's work? If you really care about real time, this should be taken into account.

I am a bit concerned about https://github.com/LinuxCNC/linuxcnc/issues/3286#issuecomment-4450835980 stating memory does not matter.

Basically, duet to the shm architecture between the userspace and realtime context, I expect tons of cash misses if the shm does get to big. It will be way worse with multiple tile CPU's. If a thread on tile one writes to a signal that the real time thread on tile two reads, it goes from die to die and worst case trough main memory, all introducing random latency depending what else is going on.

Most modern CPU's have ~1MB of L2 / 32 MB of L3. My old machine has 256KB L2 / 6MB L3 cache. If you can fit all your real timecode and data in there, it will be really fast. However, as soon as you leave it, bad luck.

And also as soon as a non realtime thread writes a lot of data spread over the whole shm segment, it will block the real time thread due to main memory IO.

A good video explaining it on the surface: https://www.youtube.com/watch?v=wGSSUSeaLgA TLDR: It is complex, there are many pitfalls and no person will ever fully understand it.

A good example is the following slide from the video, last time I checked, it was a 10x speed penalty for large arrays:
grafik

hdiethelm:
Hmm, am i right assuming that signals are allocated anyway already 8 bytes with `hal_data_u`? So you change will not increase the used size at all?

Just having bit a hard time understanding how the hal stores and links the data... ;-)

Edit:
How I understand it, all the hal info structs are allocated down and all the real time accessed data `hal_malloc()` / `hal_data_u` is allocated up, so it's not interleaving? Looks pretty well optimized.

Edit2:
Looks like dummysig is in the upper region, so when you write to a non assigned pin, it all breaks down.

Anyway, a bit off topic.

hdiethelm:
I could not let it be and added a print for shm offsets to "halcmd show" to see what is going on: https://github.com/hdiethelm/linuxcnc-fork/tree/halshow_addr (Quick & Dirty)

The pins are a bit complex because they have an shmptr pointing to an address relative to the component's shmem_base.
You see all big addresses are in the upper area, all small in the lower. So some violation there it seams to what was intended.
[show_latency_test.txt](https://github.com/user-attachments/files/29617367/show_latency_test.txt)

If you think such an add-on to the show command is useful, I can do it in a nice way and create a PR.

BsAtHome:
The getter/setter changes do not change the deep-down structure. It just makes them generally inaccessible by making it opaque for the casual user. That also means that `hal_data_u` will be removed from the user's view once all is converted and you are required to use the getters/setters.

On the issue of cache coherence and cache trashing,... The majority of the communication in HAL data transport is done at the signal level, which is the really active part. All these signal values are physically allocated close together and are (now enforced) allocated on an minimum 8-byte alignment and always at least multiple of 8-byte sized.
And, yes, you saw that right, the allocations for this data is separated from the data structures. That means that unconnected pins and params are a cost increase in location spread for caching. But these are generally used less than the active signals. There is some room for optimization here, but it cannot be done as long as everybody is accessing hal's internal structure.
The real kicker is that RT is supposed to run on an isolated CPU, which ensures that its local cache is less affected by changing code and data. It is allowed to keep much more in its local cache than other CPUs can. You still have the issue of cache coherency. The good part is that the most writers are running in RT and fewer writers live in non-RT. Anyhow, this memory region is tagged as shared between processes. Shared memory comes with an increased cost because the system/hardware must be aware of it. Non-shared memory in single threaded access can be handled quite a bit faster because it does not care about any multi-core problems.

hdiethelm:
Do you have somewhere a comment describing the whole concept? It seams to be quite breaking and quite an effort. So if you breake anyway everything:
- only int / real (Is there any need for uint?)
- type checked IN / OUT ( Reading out and writing in is both unneccesary / bad)
- Convert int / double in a nice way, i have allrady 4 type converters in my hal file and growing... ;-)

The cache thing can be investigated later, I see no real issue so far with my setup. This needs first good benchmarks. Or is there already something? Wouldn't be the first time things get slower after optimisation... ;-)

BsAtHome:
> Do you have somewhere a comment describing the whole concept?

Background, as mentioned earlier in this thread, #3286. There is also a thread on the developer mailing list about it.

> It seams to be quite breaking and quite an effort. So if you breake anyway everything:
> * only int / real (Is there any need for uint?)

No. This has been discussed.

> * type checked IN / OUT ( Reading out and writing in is both unneccesary / bad)

You cannot control read/write access to the transport layer the way you think. For a signal, yes, there may be a conceptual idea, but it fails for the code and you no longer can access data in a read-modify-write cycle without parallel non-volatile non-shared storage (preferably marked for single threaded access only).

You were talking memory usage and cache behaviour and want to add another variable? Sure, but then you need to remove most of the current code and rewrite it with performance in mind and double up for different architectures. Oh, and you should prevent the casual coder from changing anything. Then we can discuss squeezing the last bit of performance out of it.

> * Convert int / double in a nice way, i have allrady 4 type converters in my hal file and growing... ;-)

This is beyond the scope of what is being done here and you should read the previous discussions for the background.
That said, conversion is a limitation of how signals are handled. The transport of data in a signal path is a mere storage location without code intervention. Transparent conversions would need a very complex new type of connection beyond a signal type. It also creates a problem <i>when</i> the conversion needs to take place and <i>who</i> will be responsible (function ordering is important for signal propagation). This is a very non-trivial problem and each solution has its advantages and drawbacks without any clear winner in either direction (except for dramatic and emotional discussions).

> The cache thing can be investigated later, I see no real issue so far with my setup. This needs first good benchmarks. Or is there already something? Wouldn't be the first time things get slower after optimisation... ;-)

Already answered in the discussion(s) before. Nothing in the layout or access patterns is changed, at all. The speed is the same with the changes (or even better) than what is done today.

hdiethelm:
> > Do you have somewhere a comment describing the whole concept?
> 
> Background, as mentioned earlier in this thread, #3286. There is also a thread on the developer mailing list about it.

I  have to read trough it, it's a lot. Just skimmed until now. It will take some time...

Is it this thread? https://sourceforge.net/p/emc/mailman/emc-developers/thread/051f1d9c-e615-42e0-91c1-9776b623ccf2%40aitalmac.com/#msg59344193

> > It seams to be quite breaking and quite an effort. So if you breake anyway everything:
> > 
> > * only int / real (Is there any need for uint?)
> 
> No. This has been discussed.
> 
> > * type checked IN / OUT ( Reading out and writing in is both unneccesary / bad)
> 
> You cannot control read/write access to the transport layer the way you think. For a signal, yes, there may be a conceptual idea, but it fails for the code and you no longer can access data in a read-modify-write cycle without parallel non-volatile non-shared storage (preferably marked for single threaded access only).

Due to the transport layer is hidden and hopefully correct, it should be fine. I meant prohibiting write to IN on the user side.

Basic quick&dirty concept, based on your code. It does not show anything else than a single connected in and out pin but no connect logic. I added also read from out, so read/modify/write is possible.

Disadvantage

- 3x the amount of functions (Could be done by a code generator)

Advantage

- No write to IN, checked by compiler
- Looking at the code, it is immediately clear if a pin is IN or OUT

```c
#include <stdio.h>

struct __hal_stype_pin_out_real_t;
struct __hal_stype_pin_in_real_t;

typedef struct __hal_stype_pin_out_real_t *hal_pin_out_real_t;
typedef struct __hal_stype_pin_in_real_t *hal_pin_in_real_t;

typedef double rtapi_real;
double data = 0.0;

void hal_pin_new_out_real(hal_pin_out_real_t *pin){
    *(double**)pin = &data;
}

void hal_pin_new_in_real(hal_pin_in_real_t *pin){
    *(double**)pin = &data;
}

rtapi_real hal_set_out_real(hal_pin_out_real_t ref, rtapi_real val) {
    *(double*)ref = val;
    return val;
}
rtapi_real hal_get_out_real(const hal_pin_out_real_t ref) {
    return *(double*)ref;
}
rtapi_real hal_get_in_real(hal_pin_in_real_t ref) {
    return *(double*)ref;
}

int main(int argc, char *argv[]) {
    hal_pin_out_real_t out;
    hal_pin_in_real_t in;

    hal_pin_new_out_real(&out);
    hal_pin_new_in_real(&in);

    //Does not compile
    /*
    hal_pin_new_out_real(&in);
    hal_pin_new_in_real(&out);
    */

    hal_set_out_real(out, 3.5);
    printf("%f\n", hal_get_in_real(in));
    printf("%f\n", hal_get_out_real(out));

    //Does not compile
    /*
    hal_set_out_real(in, 3.5);
    printf("%f\n", hal_get_in_real(out));
    printf("%f\n", hal_get_out_real(in));
    */

    //Read/Modify/Write
    hal_set_out_real(out, hal_get_out_real(out) + 2.0);
    printf("%f\n", hal_get_in_real(in));

    return 0;
}
```

> You were talking memory usage and cache behaviour and want to add another variable? Sure, but then you need to remove most of the current code and rewrite it with performance in mind and double up for different architectures. Oh, and you should prevent the casual coder from changing anything. Then we can discuss squeezing the last bit of performance out of it.

Agreed. read/modify/write is a good thing as long as you are not doing it in a for loop... ;-)

> 
> > * Convert int / double in a nice way, i have allrady 4 type converters in my hal file and growing... ;-)
> 
> This is beyond the scope of what is being done here and you should read the previous discussions for the background. That said, conversion is a limitation of how signals are handled. The transport of data in a signal path is a mere storage location without code intervention. Transparent conversions would need a very complex new type of connection beyond a signal type. It also creates a problem when the conversion needs to take place and who will be responsible (function ordering is important for signal propagation). This is a very non-trivial problem and each solution has its advantages and drawbacks without any clear winner in either direction (except for dramatic and emotional discussions).
> 
> > The cache thing can be investigated later, I see no real issue so far with my setup. This needs first good benchmarks. Or is there already something? Wouldn't be the first time things get slower after optimisation... ;-)
> 
> Already answered in the discussion(s) before. Nothing in the layout or access patterns is changed, at all. The speed is the same with the changes (or even better) than what is done today.

Both points would be for later. Sure not now.

Except: Are there already hal benchmarks that check the performance? Would probably be beneficial also for this change.

BsAtHome:
> Disadvantage
> * 3x the amount of functions (Could be done by a code generator)

You just broke uniformity between pins and parameters.

And **all** of this can be done _after_ the proposed changes if someone want to take this further. Your suggestion requires a tremendous amount of effort and a large rewrite of the code-base. The current state of the code is not in a state where you can just "fix" this.


> > Nothing in the layout or access patterns is changed, at all. The speed is the same with the changes (or even better) than what is done today.
> 
> Except: Are there already hal benchmarks that check the performance? Would probably be beneficial also for this change.

You are welcome to do so yourself. I have analyzed the before/after generated code. There is no <i>extra</i> benefit for your proposal because the PR does not really change the code generation at the lowest level (besides better information for the compiler for it to make the right choices). The proposed changes fix the <i>access patterns</i> from a code writing perspective where the compiler enforces a coding pattern. It hides the underlying mechanisms but does _not_ change them. The changed perspective enforce you to use an abstracted view of the data you send onto and get from pins/params. This has been explained before and when you read the code then you can draw your own conclusion of the same.

BsAtHome:
@hdiethelm I looked a bit deeper into your idea of explicit input/output typing a bit more. I do see the charm it represents.

Changing the input/output typing requires following changes:
* Create a new set of hal types, only seen in code space, like in `hal_xxxxi_t` to mark input and `hal_xxxxo_t` to mark output on the type. The user-seen types do not change (i.e. HAL_BOOL/HAL_SINT/HAL_UINT/HAL_REAL are unchanged).
* Number of getters must be doubled to distinguish between reading input pins and reading output pins.
* Number of setters is the same but only accept `hal_xxxxo_t` types.
* Parameters are always seen as output types `hal_xxxxo_t` (HAL_RW is equivalent to unconnected HAL_IO and HAL_RO is equivalent to HAL_OUT).
* There are two sets of pin creation routines like in `hal_pin_new_xxxx[io]()` and the directional argument can be dropped because it is now implicit. The parameter creation routines stay the same, but require `hal_xxxxo_t` type.
* The entire code-base needs to be changed and checked (manually) and any discrepancy needs to be fixed. The real problem is to create the proper compatible code changes for constructs that assume access in any particular way.
* Components using halcompile do not need additional changes because access to the pins is governed by a set of defines that already today prevent input pins from being written. The type changes are transparent to the user.
* The type combining union must be split into an input and output union `hal_refs[io]_u` to distinguish between direction. Code using the union already uses demultiplexing and should normally not combine both in and out types in the same case.
* hal_set_p() will need some additional internal casting to write to the actual underlying data. However, this function is now part of hal_lib and its internals are hidden for normal users anyway.
* Pin iteration still needs to be careful to distinguish between input and output. This case, and many similar cases, cannot be covered automatically because you still need union types. Most cases are either set or get, however, the get case is now code-wise the more complex case because there are twice as many getters.
The good thing is that normal driver coding does not require pin iteration. Iteration is normally only required by special utility programs.

There will still be no changes for any normal user as seen from the user-land HAL abstraction (anybody or anything that works with .hal-files and halcmd).
Anybody using components with halcompile only see the changes as they already have been proposed. No changes there.
The only real difference is all the C and C++ code that codes directly against hal_lib.

I guess this is what you imagined?

BsAtHome:
Forgot to add..
The HAL_IO pins must either be removed or they are to be interpreted as output pins. Neither is really attractive.

Removing HAL_IO means, contrary to the current proposal, to change the user-space abstraction and having to deal with a much larger fallout. That option would involve changing the ~100 pins that are currently marked IO and finding a workaround at the code level and user-space procedural and functional levels for each of them (and dealing with user's .hal files because they will break too). It also means that you need to change the homing procedure(s) and <i>that</i> may not be a good idea at this time.

Seeing HAL_IO as output pins, although functional, is not technically correct and would make the exercise to do full compiler type checking non-functional for the distinction. It would be an exception that will probably bite you sooner than later. Creating proper type distinctions for IO means you have to create a <i>third</i> set of getters/setters and pin creation functions to make this in any way consistent. The alternative is that you need to pass the distinction of HAL_OUT/HAL_IO to the creation functions and that is not really consistent from an API perspective.

Considering the complications, it would IMO be unwise, at this time, to enforce input/output. It would probably be better to split this into a later change where you:
1. do the currently proposed changes
2. deal with all the possible fallout, this will take some time, also because of deprecation periods
3. if possible:
 . get rid of HAL_IO and signal bus functionality and create new procedures how things need to function
 . deal with the fallout for (quite) some time until the userbase has adapted to it
 . split HAL_IN and HAL_OUT into different types to enforce read/write properties on pins.

The "if possible" tag means that removing HAL_IO is probably not an easy undertaking.

grandixximo:
Although Some IO can be replaced by parameters, some IO are genuinely to be connected as IO to other IO where both sides can write and read, IMO you cannot drop IO as a concept, and neither Input nor Output fit the bill.

rodw-au:
> Although Some IO can be replaced by parameters, some IO are genuinely to be connected as IO to other IO where both sides can write and read, IMO you cannot drop IO as a concept, and neither Input nor Output fit the bill.

This is true. **joint.N.index-enable** is one example but other hardware use pins for both input and output for 2 way sync.

hdiethelm:
> @hdiethelm I looked a bit deeper into your idea of explicit input/output typing a bit more. I do see the charm it represents.

That sums it up nicely. It's mainly for C/C++ components to have the direction visible and checked. It might create a headache for components that create pin's dynamically. I would not say it's impossible but surely a lot of effort.

I fully understand if you discard this or postpone it. Sadly, none of us has infinite time available... ;-) And even if I would like to work on this, it's just impossible for me due to time constraints. If you feel comfortable discussing this concept future, we can do this. If not, I'm also fine leaving it here and might be continue some other day.

You all have way more experience in LinuxCNC than me, so you can estimate the effort / impact better.

It might be simpler to use your actual PR implementation. Then change to fully typed IN/OUT/IO after the legacy types are gone, so not so many functions are needed any more. The disadvantage is that you have to break the API twice.

I agree that this would be a big undertaking, having the pins fully typed. But having nice setter/getter that allow to do wrong things (setting an IN pin) is not nice. Still, this PR is an improvement anyway.

>  Parameters are always seen as output types `hal_xxxxo_t` (HAL_RW is equivalent to unconnected HAL_IO and HAL_RO is equivalent to HAL_OUT).

I fully ignored parameters so far. There are RO/RW, basically equal to OUT/IO. Why is there no WO parameter type? Many parameters are to configure a component. So the component has basically no business changing the value you just set.
However, parameters are often set to a default value at start. So something like: `void hal_param_new_in_real(hal_param_in_real_t *pin, rtapi_real ini_val);` would make sense. To not break param vs pin, we could go to `hal_in_real_t` and use this for pin and prams. That said, different types for pins and params help for visibility: You instantly see what it is, no need for a `//this is a parameter`. Downside is yet more functions.

> hal_set_p() will need some additional internal casting to write to the actual underlying data. However, this function is now part of hal_lib and its internals are hidden for normal users anyway.

hal_set_p(): You mean halcmd_commands.cc do_setp_cmd()? That moves anyway inside hal_lib where you can check the type/direction by the flags. My suggestion is not removing the typ/direction flags completely, they just live inside hal_lib and no one else has access to them. Inside hal_lib, you do the dangerous stuff while outside, it is enforced by types.

> Forgot to add.. The HAL_IO pins must either be removed or they are to be interpreted as output pins. Neither is really attractive.

There is no need to remove IO for this. Having IO would only mean yet another type and functions to handle this.
IO can be a single type with read/write allowed everywhere.
```
void hal_pin_new_io_real(hal_pin_io_real_t *pin);
hal_get_io_real(hal_pin_io_real_t ref);
hal_set_io_real(hal_pin_io_real_t ref, rtapi_real val);
```

BTW: benchmark: I will look into it. Would be nice any way to have more performance checks. Not that I don't trust your code, more to really check for regressions. Might take some time... The only disadvantage: You will need to run them on an RT enabled and well tuned PC. CI makes no sense except we connect a PC standing somewhere in a basement to the Github CI. As much as I know, this is possible.

BsAtHome:
> I fully ignored parameters so far. There are RO/RW, basically equal to OUT/IO. Why is there no WO parameter type?

There is a WO parameter. Planned for release in version ℯlog<sub>2</sub>(ℯ) (not rounded) and it has been modeled after https://web.archive.org/web/20120316141638/http://www.national.com/rap/files/datasheet.pdf already with great success. It has never lost a value since it was first tried.

> hal_set_p(): You mean halcmd_commands.cc do_setp_cmd()? That moves anyway inside hal_lib where you can check the type/direction by the flags.

I was thinking out loud... Directionality is retained. The do_setp_cmd() was (re-)implemented at least three times in different parts of the code base and all have been reduced to a call down to hal_lib and some added value interpretation logic.
All code that touches hal's internal layout is put into the query API abstraction to isolate all internal structures. The internal hal_lib interpretation of direction depends on the perspective (only boggles the mind on an odd afternoon and confuses every now and then when you least expect it).

andypugh:
Just to clarify. Having read back through this, this particular PR implements the new interface but leaves the old way of doing things working in parallel?

BsAtHome:
> Just to clarify. Having read back through this, this particular PR implements the new interface but leaves the old way of doing things working in parallel?

Yes, it is a completely parallel implementation that adds the new API of hal_lib while maintaining the current API for compatibility. The underlying data path between components (the signals' data link) is completely compatible with the old API. Keeping compatibility in the transition phase is important or there would be no easy migration path and a lot of complaining.

Once everything in-tree is converted we can deprecate the old API, which will warn out-of-tree components and applications. We then wait for "some" time for others to catch up and finally do the actual API break by removing the old API.
That is also the moment where all creation of pins, params and signals will become 64-bit only (as seen from the users; the external API). You can still get/set 32-bit types with the appropriate getter/setter (the internal API) and they will (inter-)act properly on the underlying data.

hdiethelm:
> > I fully ignored parameters so far. There are RO/RW, basically equal to OUT/IO. Why is there no WO parameter type?
> 
> There is a WO parameter. Planned for release in version ℯlog2(ℯ) (not rounded) and it has been modeled after https://web.archive.org/web/20120316141638/http://www.national.com/rap/files/datasheet.pdf already with great success. It has never lost a value since it was first tried.

Sounds interesting... ;-) However, having WO parameters with an initial value needs an API that checks this, ideally at compile time. Otherwise it makes no sense.

> 
> > hal_set_p(): You mean halcmd_commands.cc do_setp_cmd()? That moves anyway inside hal_lib where you can check the type/direction by the flags.
> 
> I was thinking out loud... Directionality is retained. The do_setp_cmd() was (re-)implemented at least three times in different parts of the code base and all have been reduced to a call down to hal_lib and some added value interpretation logic. All code that touches hal's internal layout is put into the query API abstraction to isolate all internal structures. The internal hal_lib interpretation of direction depends on the perspective (only boggles the mind on an odd afternoon and confuses every now and then when you least expect it).

Classical in/out problem. Not even MOSI / MISO avoids connecting the stuff wrong from time to time.

BsAtHome:
> Sounds interesting... ;-) However, having WO parameters with an initial value needs an API that checks this, ideally at compile time. Otherwise it makes no sense.

After getting to #4247 I do not think adding WO parameters would be any problem. At least that way we can fill all that unused memory with proper values. All those address locations are too lonely without some written love bits once in a while.

PR #4100, opened 2026-06-02
docs: fix cross-document xref links broken under compat-mode

grandixximo:
Follow-up to #4053. Cross-document references resolve to the current page instead of the target page. For example on `hal/tools.html` the "HAL Tutorial" link points to:

```
hal/tools.html#tutorial.adoc#cha:hal-tutorial
```

instead of:

```
hal/tutorial.html#cha:hal-tutorial
```

**Cause:** `xref_resolver.rb` rewrote bare `<<anchor,Title>>` to the inter-document form `<<file.adoc#anchor,Title>>`. That syntax is asciidoctor-native, and the docs build with `-a compat-mode` (needed for the legacy asciidoc-py source). Under compat-mode asciidoctor parses `<<file.adoc#anchor>>` as a literal in-page anchor id, so every cross-doc link lands on the current page.

**Fix:** the resolver now emits a `link:` macro, which compat-mode leaves intact, with `.adoc` rewritten to `outfilesuffix` (`.html` / `.pdf`). A `link:` macro has no automatic reference text, so the anchor index also captures each target's reftext (explicit `[[id,reftext]]` or the following heading) to label bare `<<anchor>>` uses. Cache format bumped to v2.

One file changed; no source `.adoc` edits. The alternative, rewriting the 365 cross-document xrefs in source by hand, was rejected as far more churn and not self-maintaining.

**Verified** on `hal/tools.html` (the reported case), the two no-title cross-doc refs (`sec:hal-components`, `mdiline`), and INI `[SECTION]` bracket labels. All resolve to the correct target page.

### DIFF
--- a/docs/src/extensions/xref_resolver.rb
+++ b/docs/src/extensions/xref_resolver.rb
@@ -1,8 +1,12 @@
 # docs/src/extensions/xref_resolver.rb
 #
 # Asciidoctor preprocessor that resolves bare <<anchor,Title>> cross-doc
-# references to qualified <<relpath/file.adoc#anchor,Title>> form, the way
-# asciidoc-py used to do via objects/xref_<lang>.links.
+# references to a link: macro pointing at the target's output file.
+#
+# We emit link: rather than the inter-document xref form (<<file.adoc#id>>)
+# because the docs build with -a compat-mode, which disables that syntax and
+# renders it as a literal in-page anchor.  link: carries no auto reftext, so
+# we capture the target's reftext (heading or [[id,reftext]]) for bare uses.
 #
 # Wire-up (in Submakefile):
 #   asciidoctor -r $(DOC_SRCDIR)/extensions/xref_resolver.rb \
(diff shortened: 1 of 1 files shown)


grandixximo:
@hansu see if this fixes the cross link, alternative is rewriting them all, which I don't think is wise...

hansu:
Thanks for picking this up so quickly!

This works for links for target files where file.adoc = file.html. But for files where adoc are files included in another adoc file it doesn't.

For example the link "Halshow" in section 3: https://linuxcnc.org/docs/devel/html/hal/tools.html#_halshow

Before this fix it was pointing to: `html/hal/tools.html#halshow.adoc#sec:halshow`
and with your fix to: `html/hal/halshow.html#sec:halshow`

But halshow.adoc is included tutorial.adoc so it should point to` html/hal/tutorial.html#sec:halshow`

A bit tricky, yes...

grandixximo:
Good catch. Sent another commit, I think is fixed, the resolver now builds the include graph and redirects anchors defined in `included` partials to the page they render into, so `halshow`/`halscope` point to` tutorial.html#....` Verified the include-graph result matches the `DOC_SRCS_EN` build list exactly.

hansu:
Yeah looks good now, thanks!

hansu (approved):

PR #4103, opened 2026-06-02
docs: fix malformed and dangling cross-references

grandixximo:
Five source-level cross-reference bugs found while auditing the doc links. Independent of the resolver (#4100); these are wrong in the source regardless.

- **qtvcp-widgets.adoc** (x2), **qtplasmac.adoc**: `<<id>>[Label]` is malformed. The `[Label]` leaked as literal text after the link instead of being the link text. Fixed to `<<id,Label>>`.
- **o-code.adoc**: `<<Numbering>>` had no matching anchor (the section's auto id is `_numbering`), so it rendered as literal `[Numbering]`. Added an explicit `[[ocode:numbering]]` anchor, matching the file's convention.
- **pncconf.adoc**: `<<sub:install:sec:latency-test,...>>` used a stale prefix; the real anchor is `sec:latency-test`.
- **remap.adoc**: dropped a commented-out line carrying a dangling `<<remap:extracting-words>>` ref (a stale duplicate of the line below it), so nobody copies the broken form.

Verified with a full English `make htmldocs`: checkref reports "all links are good!" and the four rendered links now resolve correctly.

### DIFF
--- a/docs/src/config/pncconf.adoc
+++ b/docs/src/config/pncconf.adoc
@@ -66,7 +66,7 @@ Just like a railroad, LinuxCNC requires everything on a very tight and consisten
 LinuxCNC requires and uses a 'real-time' operating system, which just means it has a low-latency (lateness) response time.
 When LinuxCNC requires and is performing calculations, it cannot be interrupted by lower priority requests (such as user input to screen buttons or drawing etc).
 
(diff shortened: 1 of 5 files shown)


hansu:
:+1:

PR #4105, opened 2026-06-02
docs: fix two broken cross-reference targets

grandixximo:
Two broken links surfaced once HTML link checking actually runs (companion to the checkref fix, #4104). Both are stale fragment ids: the generated ids are `_commands` (halcmd.1) and `_subcommands` (linuxcncrsh.1).

### DIFF
--- a/docs/src/gui/qtvcp-widgets.adoc
+++ b/docs/src/gui/qtvcp-widgets.adoc
@@ -1537,7 +1537,7 @@ Enter any of these _case sensitive_ commands to load the respective program or a
 *`CLEAR HISTORY`*::
   _Clears the MDI History_.
 *`net`*::
-  See link:../man/man1/halcmd.1.html#COMMANDS[`halcmd net` commands]. +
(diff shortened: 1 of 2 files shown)


BsAtHome:
> The third, in plasma-cnc-primer.adoc, had `[AXIS_<letter>]` as the xref title; the `>` broke the xref so the resolver could not qualify it, switched to the `[AXIS_l]` wording the anchor's other references already use.

Did it break the asciidoctor resolver or your resolver? If if was yours, then it needs fixing.

grandixximo:
> > The third, in plasma-cnc-primer.adoc, had `[AXIS_<letter>]` as the xref title; the `>` broke the xref so the resolver could not qualify it, switched to the `[AXIS_l]` wording the anchor's other references already use.
> 
> Did it break the asciidoctor resolver or your resolver? If if was yours, then it needs fixing.

Fair, fixed my resolver in #4100, and I reverted the workaround here, so this PR is now just the two stale-anchor fixes.

PR #4107, opened 2026-06-02
latency-test, latency-histogram: add realtime status bar

### LINKED ISSUE #4044: Latency test discrepancy after ethercat feature adding
Hello gentelmen,

@BsAtHome I am facing weird latency issue from moment of installation last RIP of linuxcnc 2.10.0 where is implemented the `initf lcec.activate servo-thread`... 

This issue created real mess in the ethercat setup mainly noticeable during running beckhoff ax5yzz SoE which is pretty sensitive DC at latency at all.

The difference we noticed with the @grandixximo during testing between custom ec master from "community branch" and "grandixximo branch"...

We can replicate it if it will be necessary.. time of duration take with spare, its mainly for visualisation

Visualization at pictures

GRUB_CMDLINE_LINUX_DEFAULT="quiet isolcpus=domain,managed_irq,2-3 irqaffinity=0-1 nohz_full=2-3 rcu_nocbs=2-3 rcu_nocb_poll rcutree.kthread_prio=5 preempt=full workqueue.cpumask=1 processor.max_cstate=1 intel_idle.max_cstate=1 intel_pstate=active cpufreq.default_governor=performance ipv6.disable=1 i915.enable_dc=0 i915.enable_fbc=0 i915.enable_psr=0 mitigations=off nmi_watchdog=0 nosoftlockup tsc=reliable clocksource=tsc transparent_hugepage=never skew_tick=1 acpi_irq_nobalance irqpoll noirqbalance pcie_aspm=off module_blacklist=evbug,intel_cstate,intel_rapl"

Image
Image
Image
Image

grandixximo:
## What

`latency-test` and `latency-histogram` now show a status bar along the bottom with three fields: the running realtime type (or `No realtime`), the CPU frequency governor, and isolcpus. A field is coloured red only when it flags a condition that makes the measured latency unrepresentative, so normal states stay in the theme colour. On a real RT machine it reads e.g. `Preempt RT` / `performance` / `isolcpus=2,3`.

## Why

In #4044 a run-in-place build was used without `sudo make setuid`. `rtapi_app` ran unprivileged, the latency numbers blew up, and it was mistaken for a code regression. A visible realtime indicator surfaces that immediately. The governor and isolcpus fields cover the other two common reasons a latency measurement is misleading.

## How

The realtime type now comes authoritatively from rtapi itself: `realtime verify` returns the type that `rtapi_app` reports for the running backend (`Preempt RT`, `RTAI`, `Xenomai`, ..., or `No realtime`), instead of each tool re-sniffing kernel signals. The governor is read from `/sys` and isolcpus from `/proc/cmdline`. `latency-test` renders the bar as a pyvcp footer; `latency-histogram` as a Tk status bar at the bottom. The histogram's earlier modal `no realtime` popup is replaced by this persistent bar (the console warning is kept for non-X runs).

Docs: `install/latency-test.adoc` gains a status-bar section explaining each field and linking the fixes (realtime kernel, setuid/setcap, isolcpus), plus a CPU frequency governor tuning note.

## rtapi_app change (needs RT/C++ review)

To make the type available without sniffing, this branch carries @hdiethelm's change to `rtapi_app` (`src/rtapi/uspace_rtapi_main.cc`): the `check_rt` socket result is extended to return a string alongside the int status, and the realtime-type enum is mapped to a human-readable name. I added a follow-up commit tidying it: a short-read check in `recv_result()`, a frame-size underflow guard in `recv_result()`/`recv_args()`, an exhaustive type-name switch, and typo fixes. The two commits are kept separate so the tidy delta reads on its own.

## Follow-ups (separate PRs)

- A realtime indicator in the main GUIs (axis, gmoccapy, qtdragon, touchy), showing just the realtime field; cross-GUI placement is #4118.
- A headless `--quiet`/`--seconds` mode for `latency-test` so scripts can capture latency (suggested by @rodw-au).

Depends on #4132 (merged). Closes #4044.

### DIFF
--- a/scripts/latency-histogram
+++ b/scripts/latency-histogram
@@ -111,6 +111,26 @@ proc which_exe {name} {
   return -code error "$name: executable not found"
 } ;# which_exe
 
+proc check_rt_privileges {} {
+  # Without 'sudo make setuid' (or 'sudo make setcap') rtapi_app runs
+  # unprivileged: realtime threads get no SCHED_FIFO and no locked memory, so
+  # the latency readings are wildly inflated.  Warn rather than mislead.
+  if {![catch {exec id -u} uid] && $uid == 0} return ;# root has every privilege
+  if {[catch {which_exe rtapi_app} rtapi]} return ;# cannot locate, stay quiet
+  if {![catch {file attributes $rtapi -permissions} mode] && ($mode & 04000)} return ;# setuid root
+  if {![catch {exec getcap $rtapi} caps] && [string match *cap_sys_nice* $caps]} return ;# file caps
+  set msg "Warning: rtapi_app is not setuid root and has no realtime capabilities."
+  append msg "\nRealtime threads get no SCHED_FIFO priority or locked memory,"
+  append msg "\nso the latency shown will be far worse than reality."
+  append msg "\nRun 'sudo make setuid' (or 'sudo make setcap') in src/."
+  puts stderr $msg
+  if {$::LH(use_x)} {
+    package require Tk
+    tk_messageBox -parent . -icon warning -type ok \
+        -title "Latency Warning" -message $msg
+  }
+} ;# check_rt_privileges
+
 proc program_check {plist} {
   foreach prog $plist {
     if [catch {
@@ -972,6 +992,7 @@ proc windowToFile { win } {
 if ![info exists ::LH(start)] {
   set_defaults
   config
+  check_rt_privileges
   progress "Loading packages"
   load_packages
   signal trap SIGINT finish
--- a/scripts/latency-histogram
+++ b/scripts/latency-histogram
@@ -115,6 +115,11 @@ proc check_rt_privileges {} {
   # Without 'sudo make setuid' (or 'sudo make setcap') rtapi_app runs
   # unprivileged: realtime threads get no SCHED_FIFO and no locked memory, so
   # the latency readings are wildly inflated.  Warn rather than mislead.
+  # Only meaningful on an RT-capable kernel; stay quiet on non-RT (e.g. dev).
+  set rt 0
+  if {![catch {exec uname -v} kv] && [string match *PREEMPT_RT* $kv]} {set rt 1}
+  if {[string first rtai [string tolower $::tcl_platform(osVersion)]] >= 0} {set rt 1}
+  if {!$rt} return
   if {![catch {exec id -u} uid] && $uid == 0} return ;# root has every privilege
   if {[catch {which_exe rtapi_app} rtapi]} return ;# cannot locate, stay quiet
   if {![catch {file attributes $rtapi -permissions} mode] && ($mode & 04000)} return ;# setuid root
--- a/scripts/latency-histogram
+++ b/scripts/latency-histogram
@@ -112,22 +112,15 @@ proc which_exe {name} {
 } ;# which_exe
 
 proc check_rt_privileges {} {
-  # Without 'sudo make setuid' (or 'sudo make setcap') rtapi_app runs
-  # unprivileged: realtime threads get no SCHED_FIFO and no locked memory, so
-  # the latency readings are wildly inflated.  Warn rather than mislead.
-  # Only meaningful on an RT-capable kernel; stay quiet on non-RT (e.g. dev).
-  set rt 0
-  if {![catch {exec uname -v} kv] && [string match *PREEMPT_RT* $kv]} {set rt 1}
-  if {[string first rtai [string tolower $::tcl_platform(osVersion)]] >= 0} {set rt 1}
-  if {!$rt} return
-  if {![catch {exec id -u} uid] && $uid == 0} return ;# root has every privilege
-  if {[catch {which_exe rtapi_app} rtapi]} return ;# cannot locate, stay quiet
-  if {![catch {file attributes $rtapi -permissions} mode] && ($mode & 04000)} return ;# setuid root
-  if {![catch {exec getcap $rtapi} caps] && [string match *cap_sys_nice* $caps]} return ;# file caps
-  set msg "Warning: rtapi_app is not setuid root and has no realtime capabilities."
-  append msg "\nRealtime threads get no SCHED_FIFO priority or locked memory,"
+  # Ask the realtime layer whether realtime is actually available, via
+  # 'realtime verify' (exits 0 when realtime is up, non-zero otherwise,
+  # usually a missing 'sudo make setuid' or 'sudo make setcap').
+  if {[catch {exec linuxcnc_var REALTIME} rt]} return ;# cannot locate, stay quiet
+  if {![catch {exec $rt verify 2>@1}]} return         ;# realtime available
+  set msg "Warning: realtime is not available."
+  append msg "\nThreads get no SCHED_FIFO priority or locked memory,"
   append msg "\nso the latency shown will be far worse than reality."
-  append msg "\nRun 'sudo make setuid' (or 'sudo make setcap') in src/."
+  append msg "\nOn a run-in-place build, run 'sudo make setuid' (or 'sudo make setcap') in src/."
   puts stderr $msg
   if {$::LH(use_x)} {
     package require Tk
@@ -649,6 +642,7 @@ proc start_collection {} {
     }
   }
   hal start
+  check_rt_privileges
   if {!$::LH(legacy)} {
     # Attach the Tcl hal_stream binding (added in halsh.c) to each
     # thread's latencybinstream FIFO. The handle stays open for the
@@ -997,7 +991,6 @@ proc windowToFile { win } {
 if ![info exists ::LH(start)] {
   set_defaults
   config
-  check_rt_privileges
   progress "Loading packages"
   load_packages
   signal trap SIGINT finish
--- a/scripts/latency-histogram
+++ b/scripts/latency-histogram
@@ -111,24 +111,6 @@ proc which_exe {name} {
   return -code error "$name: executable not found"
 } ;# which_exe
 
-proc check_rt_privileges {} {
-  # Ask the realtime layer whether realtime is actually available, via
-  # 'realtime verify' (exits 0 when realtime is up, non-zero otherwise,
-  # usually a missing 'sudo make setuid' or 'sudo make setcap').
-  if {[catch {exec linuxcnc_var REALTIME} rt]} return ;# cannot locate, stay quiet
-  if {![catch {exec $rt verify 2>@1}]} return         ;# realtime available
-  set msg "Warning: realtime is not available."
-  append msg "\nThreads get no SCHED_FIFO priority or locked memory,"
-  append msg "\nso the latency shown will be far worse than reality."
-  append msg "\nOn a run-in-place build, run 'sudo make setuid' (or 'sudo make setcap') in src/."
-  puts stderr $msg
-  if {$::LH(use_x)} {
-    package require Tk
-    tk_messageBox -parent . -icon warning -type ok \
-        -title "Latency Warning" -message $msg
-  }
-} ;# check_rt_privileges
-
 proc program_check {plist} {
   foreach prog $plist {
     if [catch {
@@ -353,6 +335,44 @@ proc load_packages {} {
   }
 } ;# load_packages
 
+proc rt_status_items {} {
+  # Returns {text color} pairs; color is "" except where it needs attention.
+  set red "#cc4444"
+  set items {}
+
+  # realtime: type from 'realtime verify' stdout, yes/no from its exit status.
+  set rt_text "no realtime"
+  set rt_ok 0
+  if {![catch {exec linuxcnc_var REALTIME} rt]} {
+    if {![catch {open [list | $rt verify 2> /dev/null] r} chan]} {
+      set out [string trim [read $chan]]
+      set rt_ok [expr {![catch {close $chan}]}]
+      if {$out ne ""} {set rt_text $out}
+    }
+  }
+  lappend items [list $rt_text [expr {$rt_ok ? "" : $red}]]
+
+  # cpu governor: red when not 'performance'
+  set gov ""
+  catch {set gov [string trim [exec cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor]]}
+  if {$gov eq ""} {set gov "gov?"}
+  if {$gov eq "performance"} {
+    lappend items [list $gov ""]
+  } else {
+    lappend items [list $gov $red]
+  }
+
+  # isolcpus
+  set cmdline ""
+  catch {set cmdline [exec cat /proc/cmdline]}
+  if {[regexp {isolcpus=(\S+)} $cmdline -> iso]} {
+    lappend items [list "isolcpus=$iso" ""]
+  } else {
+    lappend items [list "no_isolcpus" ""]
+  }
+  return $items
+} ;# rt_status_items
+
 proc make_gui { {w .} } {
   set f [frame ${w}fa]
   pack $f -side top -fill x -expand 1
@@ -454,6 +474,28 @@ proc make_gui { {w .} } {
            -command "xaxis $thd"] -side left
     }
 
+  }
+
+  # --- realtime status bar (true footer): one line of axis-style sunken
+  # cells split into equal thirds, left/centre/right aligned ---
+  set sb [frame ${w}statusbar]
+  pack $sb -side bottom -fill x
+  set sbitems [rt_status_items]
+  # Fixed-width sides (chars), elastic middle gets the rest.
+  # Left fits the longest RT label (RT-Xenomai-EVL); right fits isolcpus up to ~4 cpus.
+  set sbwidth  {14 0 18}
+  set sbweight {0  1 0}
+  set sbanch   {w center e}
+  set i 0
+  foreach item $sbitems {
+    lassign $item text color
+    label $sb.s$i -text $text -relief sunken -bd 1 -anchor [lindex $sbanch $i] -padx 6
+    set wdt [lindex $sbwidth $i]
+    if {$wdt > 0} {$sb.s$i configure -width $wdt}
+    if {$color ne ""} {$sb.s$i configure -fg $color}
+    grid $sb.s$i -row 0 -column $i -sticky we
+    grid columnconfigure $sb $i -weight [lindex $sbweight $i]
+    incr i
   }
 
   set f [frame ${w}bot]
@@ -642,7 +684,6 @@ proc start_collection {} {
     }
   }
   hal start
-  check_rt_privileges
   if {!$::LH(legacy)} {
     # Attach the Tcl hal_stream binding (added in halsh.c) to each
     # thread's latencybinstream FIFO. The handle stays open for the
--- a/scripts/latency-test
+++ b/scripts/latency-test
@@ -20,6 +20,11 @@ icalc() { awk "BEGIN { printf \"%.0f\n\", ($1); }" < /dev/null; }
 # unprivileged: realtime threads get no SCHED_FIFO and no locked memory, so the
 # latency readings are wildly inflated.  Warn rather than mislead.
 check_rt_privileges () {
+    # only meaningful on an RT-capable kernel; stay quiet on non-RT (e.g. dev)
+    case "$(uname -v) $(uname -r)" in
+        *PREEMPT_RT*|*rtai*) ;;
+        *) return 0 ;;
+    esac
     [ "$(id -u)" = 0 ] && return 0          # root has every privilege
     rtapi_app=$(command -v rtapi_app) || return 0
     [ -n "$rtapi_app" ] || return 0
--- a/scripts/latency-test
+++ b/scripts/latency-test
@@ -16,30 +16,6 @@ kversion=$(uname -v)
 calc() { awk "BEGIN { print ($1); }" < /dev/null; }
 icalc() { awk "BEGIN { printf \"%.0f\n\", ($1); }" < /dev/null; }
 
-# Without 'sudo make setuid' (or 'sudo make setcap') rtapi_app runs
-# unprivileged: realtime threads get no SCHED_FIFO and no locked memory, so the
-# latency readings are wildly inflated.  Warn rather than mislead.
-check_rt_privileges () {
-    # only meaningful on an RT-capable kernel; stay quiet on non-RT (e.g. dev)
-    case "$(uname -v) $(uname -r)" in
-        *PREEMPT_RT*|*rtai*) ;;
-        *) return 0 ;;
-    esac
-    [ "$(id -u)" = 0 ] && return 0          # root has every privilege
-    rtapi_app=$(command -v rtapi_app) || return 0
-    [ -n "$rtapi_app" ] || return 0
-    [ -u "$rtapi_app" ] && return 0         # setuid root
-    if command -v getcap >/dev/null 2>&1; then
-        case $(getcap "$rtapi_app" 2>/dev/null) in
-            *cap_sys_nice*) return 0 ;;     # file capabilities
-        esac
-    fi
-    echo "Warning: rtapi_app is not setuid root and has no realtime capabilities." >&2
-    echo "         Realtime threads get no SCHED_FIFO priority or locked memory," >&2
-    echo "         so the latency shown will be far worse than reality." >&2
-    echo "         Run 'sudo make setuid' (or 'sudo make setcap') in src/." >&2
-}
-
 parse_time () {
     case $1 in
     -)   echo "0" ;;
@@ -193,7 +169,5 @@ L=\$(($SIGNALS
     ) | sort -n | tail -1)
 echo \$L > "$HOME"/.latency
 EOF
-
-check_rt_privileges
 
 halrun lat.hal
--- a/scripts/latency-test
+++ b/scripts/latency-test
@@ -81,6 +81,23 @@ if [ "$BASE" -eq "$SERVO" ]; then BASE=0; fi
 BASE_HUMAN=$(human_time "$BASE")
 SERVO_HUMAN=$(human_time "$SERVO")
 
+# --- status-bar fields (colour red only when they need attention) ---
+# realtime: type from 'realtime verify'; non-zero exit means none.
+RED="#cc4444"
+if RT_TEXT=$(realtime verify 2>/dev/null); then
+    RT_FG=""
+else
+    RT_FG="foreground=\"$RED\""
+fi
+GOV=$(cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor 2>/dev/null)
+[ -z "$GOV" ] && GOV="gov?"
+if [ "$GOV" = performance ]; then GOV_FG=""; else GOV_FG="foreground=\"$RED\""; fi
+if grep -q 'isolcpus=' /proc/cmdline 2>/dev/null; then
+    ISOL_TEXT=$(tr ' ' '\n' < /proc/cmdline | grep '^isolcpus=' | head -1)
+else
+    ISOL_TEXT="no_isolcpus"
+fi
+
 if [ "$BASE" -eq 0 ]; then
 cat > lat.hal <<EOF
 loadrt threads name1=slow period1=$SERVO
@@ -152,6 +169,15 @@ fi
 cat >> lat.xml <<EOF
 <tablerow/>
   <button text="Reset Statistics" halpin="reset" />
+
+<tablerow/><tablesticky sticky="we"/><tablespan columns="5"/>
+<table flexible_columns="[1,2,3]" uniform_columns="aaa">
+<tablesticky sticky="we"/>
+<tablerow/>
+  <label text="$RT_TEXT" $RT_FG relief="sunken" bd="1" anchor="w" padx="6"/>
+  <label text="$GOV" $GOV_FG relief="sunken" bd="1" anchor="center" padx="6"/>
+  <label text="$ISOL_TEXT" relief="sunken" bd="1" anchor="e" padx="6"/>
+</table>
 
 </table>
 </pyvcp>
--- a/scripts/latency-test
+++ b/scripts/latency-test
@@ -16,6 +16,25 @@ kversion=$(uname -v)
 calc() { awk "BEGIN { print ($1); }" < /dev/null; }
 icalc() { awk "BEGIN { printf \"%.0f\n\", ($1); }" < /dev/null; }
 
+# Without 'sudo make setuid' (or 'sudo make setcap') rtapi_app runs
+# unprivileged: realtime threads get no SCHED_FIFO and no locked memory, so the
+# latency readings are wildly inflated.  Warn rather than mislead.
+check_rt_privileges () {
+    [ "$(id -u)" = 0 ] && return 0          # root has every privilege
+    rtapi_app=$(command -v rtapi_app) || return 0
+    [ -n "$rtapi_app" ] || return 0
+    [ -u "$rtapi_app" ] && return 0         # setuid root
+    if command -v getcap >/dev/null 2>&1; then
+        case $(getcap "$rtapi_app" 2>/dev/null) in
+            *cap_sys_nice*) return 0 ;;     # file capabilities
+        esac
+    fi
+    echo "Warning: rtapi_app is not setuid root and has no realtime capabilities." >&2
+    echo "         Realtime threads get no SCHED_FIFO priority or locked memory," >&2
+    echo "         so the latency shown will be far worse than reality." >&2
+    echo "         Run 'sudo make setuid' (or 'sudo make setcap') in src/." >&2
+}
+
 parse_time () {
     case $1 in
     -)   echo "0" ;;
@@ -169,5 +188,7 @@ L=\$(($SIGNALS
     ) | sort -n | tail -1)
 echo \$L > "$HOME"/.latency
 EOF
+
+check_rt_privileges
 
 halrun lat.hal
--- a/src/rtapi/uspace_rtapi_main.cc
+++ b/src/rtapi/uspace_rtapi_main.cc
@@ -533,7 +533,25 @@ static int do_debug_cmd(const std::string &value) {
     }
 }
 
-static int do_check_rt_cmd(void) {
+// Human-readable name of a realtime type, suitable for display by callers
+// such as 'realtime verify' and the latency tools.
+static const char *realtime_type_name(rtapi_realtime_type_t type) {
+    switch(type) {
+        case REALTIME_TYPE_UNINITIALIZED:   return "Uninitialized";
+        case REALTIME_TYPE_NONE:            return "No realtime";
+        case REALTIME_TYPE_UNKNOWN:         return "Unknown (SCHED_FIFO)";
+        case REALTIME_TYPE_PREEMPT_DYNAMIC: return "Preempt Dynamic";
+        case REALTIME_TYPE_PREEMPT_RT:      return "Preempt RT";
+        case REALTIME_TYPE_RTAI:            return "RTAI";
+        case REALTIME_TYPE_LXRT:            return "LXRT";
+        case REALTIME_TYPE_XENOMAI:         return "Xenomai";
+        case REALTIME_TYPE_XENOMAI_EVL:     return "Xenomai EVL";
+    }
+    return "BUG: unknown realtime type";
+}
+
+static int do_check_rt_cmd(std::string &out) {
+    out = realtime_type_name(rtapi_get_realtime_type());
     return rtapi_is_realtime() == 0;
 }
 
@@ -597,8 +615,8 @@ static ssize_t recv_data(int fd, void *buf, size_t n, int flags) {
  * Protocol:
  * 
  * client->master: std::vector<std::string> args
- * master processes the args and returns result
- * master->client: int result
+ * master processes the args and returns result and stdout string
+ * master->client: int result, std::string out
  *
  * Packing:
  * args are serialized as:
@@ -611,40 +629,11 @@ static ssize_t recv_data(int fd, void *buf, size_t n, int flags) {
  * }
  * 
  * result is serialized as:
- * int
+ * uint16_t size (full package size including the size field)
+ * int result
+ * uint16_t out_size
+ * char[out_size] out
  */
-
-static bool send_result(int fd, int result) {
-    ssize_t res = send_data(fd, &result, sizeof(int), 0);
-    if (res != sizeof(int)) {
-        if (res == -1) {
-            rtapi_print_msg(RTAPI_MSG_ERR, "rtapi_app: send_result failed: %s\n", strerror(errno));
-        } else {
-            rtapi_print_msg(
-                RTAPI_MSG_ERR, "rtapi_app: send_result failed, send only %li of %li bytes\n", res, sizeof(int)
-            );
-        }
-        return false;
-    } else {
-        return true;
-    }
-}
-
-static bool recv_result(int fd, int *result) {
-    ssize_t res = recv_data(fd, result, sizeof(int), 0);
-    if (res != sizeof(int)) {
-        if (res == -1) {
-            rtapi_print_msg(RTAPI_MSG_ERR, "rtapi_app: recv_result failed: %s\n", strerror(errno));
-        } else {
-            rtapi_print_msg(
-                RTAPI_MSG_ERR, "rtapi_app: recv_result failed, recv only %li of %li bytes\n", res, sizeof(int)
-            );
-        }
-        return false;
-    } else {
-        return true;
-    }
-}
 
 static void push_uint16(std::vector<char> &buf, uint16_t value) {
     buf.push_back((char)(0xff & (value >> 0)));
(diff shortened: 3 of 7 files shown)


BsAtHome:
These tests are moot when you run on a non-RT kernel (like I do in dev). I'm not sure the noise is really necessary in that case.

grandixximo:
silenced on non-rt

rodw-au:
Great. Every little bit helps. Reduces user frustration and more importantly less developer time wasted on spurious issues.

hdiethelm:
Hmm, in https://github.com/LinuxCNC/linuxcnc/issues/4044 in the images in the background you clearly see:
grafik

So the new warning will not help that much. If desired, you can add it in the place where `Note: Using POSIX non-realtime` is printed, so at least i is shown always, not only in these test tools.

hdiethelm:
Connected to this:
https://github.com/LinuxCNC/linuxcnc/issues/4118

A general way for GUI's to show "You don't have realtime" warnings that you can not overlook would help the most. When milling, I start linuxcnc with the link. So no console. If I accidentally start the wrong kernel, bad luck.

BsAtHome:
On a <i>production system</i>, you may want to warn all the time when this is amiss. Maybe something with a background color turning from gray into gray-red tinted and do something similar consistently in _all_ GUIs.

For dev builds you don't really want this because you want to see what the operator sees while you are working on stuff. I'd go for an opt-in choice by setting a value in the INI file. Maybe something like a boolean `[DISPLAY]VISUAL_WARN_NONRT`, defaulting to false. Then add the entry commented out in our configs and add a choice in pnconf and friends.

hdiethelm:
I was just checking the code. @grandixximo You already added a better warning in the C code in your nonroot patch, so this was probably not even applied in the screenshot. Now there is a double warning in the console. The note from C++ and the Warning from this PR:
grafik

With latency-histogram there is also a pop-up. With latency-test, this doesn't work. And the most important app, linuxcnc, also just shows noting if you don't start it in a console.

Anyone has a good idea to check easily and globally for real time capability? 

There is already a function rtapi_is_realtime(). This is not 100% reliable, if harden_rt() fails, it will return true, even if it runs in SCHED_OTHER. But this can be fixed.

 rtapi_is_realtime() is also linked to userspace apps but there it will not work, it checks if the userspace app has realtime... ;-)

I could add a halcmd that checks for realtime. Or a pin that is true when all is ok, false otherwise.

This could then be used in all gui's for an opt-in or opt-out warning. But I am not that deep into all these various gui's and how they communicate with the RT part.

hdiethelm:
> On a production system, you may want to warn all the time when this is amiss. Maybe something with a background color turning from gray into gray-red tinted and do something similar consistently in _all_ GUIs.
> 
> For dev builds you don't really want this because you want to see what the operator sees while you are working on stuff. I'd go for an opt-in choice by setting a value in the INI file. Maybe something like a boolean `[DISPLAY]VISUAL_WARN_NONRT`, defaulting to false. Then add the entry commented out in our configs and add a choice in pnconf and friends.

Might be an option that is default on when you deploy an default off in rip-mode? But this is annoying to test.

Otherwise, I would tend for default on, dev's will manage it better to switch it off than users will fight not knowing that they don't have real time enabled. Instead of in ini, might be an environment variable LINUXCNC_NO_RT_WARN. Dev's can set it on their dev machine in .profile if they are annoyed and it works for all test configs.

BTW, just brainstorming options.

hdiethelm:
Just a POC, if you think a halcmd getrt (or better name) would help I can create a PR. Was easy.
You can use that everywhere and it will return 1 if failed / 0 if good.
```
../bin/halcmd getrt ; echo Return value $?
<commandline>:0: exit value: 1
<commandline>:0: No realtime available
Return value 1
```
make setuid
```
../bin/halcmd getrt ; echo Return value $?
Realtime available
Return value 0
```

You find it on my fork:
https://github.com/hdiethelm/linuxcnc-fork/tree/halcmd_getrt
https://github.com/hdiethelm/linuxcnc-fork/compare/master...hdiethelm:linuxcnc-fork:halcmd_getrt

hdiethelm:
Meanwhile, I found also something that looks like it is exposed to the python code:
https://github.com/LinuxCNC/linuxcnc/blob/888cb941fbf23ee32fc8e15ca180997d99a39117/src/hal/halmodule.cc#L2382
But this is broken: https://github.com/LinuxCNC/linuxcnc/issues/4129

grandixximo:
Thanks both, this is more useful than my original per-tool heuristic.

I have pivoted the PR to use @hdiethelm's `halcmd getrt` as the single source of truth instead of hand-rolling a setuid-bit / `getcap` probe in bash and tcl. Both scripts now just run `halcmd getrt` and warn only when it reports `No realtime available`. An rtai/non-uspace build, an older `halcmd` without `getrt`, or a working realtime setup all stay silent, so the check rides on the authoritative `rtapi_is_realtime()` path rather than guessing from file permissions. This also drops the weaker logic @hdiethelm rightly flagged.

A few things I would like your input on, since they touch the broader direction in #4118:

1. Console double-warning. `rtapi` already prints `Note: Using POSIX non-realtime` at the source (uspace_posix.cc). For a console tool like latency-test that note is arguably enough, and a second line from the script is the duplication you saw. I am inclined to keep the script warning only for its actionable hint (the `make setuid` / `make setcap` pointer) and let the GUI popup be the real value-add in latency-histogram. Happy to drop the latency-test console line entirely if you would rather the C note be the only console source.

2. getrt invocation/cleanup. Since `getrt` goes through `hal_systemv` and a HAL init, calling it standalone before the test seems to bring up an rtapi instance. @hdiethelm, where do you intend callers to invoke it, and does it need a `halrun -U` afterward so it does not collide with the session the tool then starts? I did not want to bake in a cleanup that could disturb a running setup.

3. Dev opt-out policy. I wired a `LINUXCNC_NO_RT_WARN` env opt-out per @hdiethelm's suggestion, which keeps @BsAtHome's dev boxes quiet without a per-kernel heuristic. If the consensus in #4118 lands on an INI key like `[DISPLAY]VISUAL_WARN_NONRT` instead, I will switch to that. The env var is easy to set once for all test configs, which is why I started there.

This now depends on the `getrt` command landing. @hdiethelm, if you open that as its own PR I will rebase on top and reference it.

BsAtHome:
Instantiating a HAL memory segment on invocation may be problematic. It will surely confuse because you have to remember to call halrun -U afterwards. That is not a good design.

Opt-out policies are generally designed to force you to do a thing, even if you do not want to. That is why they should be avoided.

hdiethelm:
> Instantiating a HAL memory segment on invocation may be problematic. It will surely confuse because you have to remember to call halrun -U afterwards. That is not a good design.

The way `halcmd getrt` runs this in the brackground for uspace is by executing `rtapi_app getrt`. Now there are two possibility's:
- rtapi_app is already running (for example you start latency_test in an other terminal): The command is executed and the result returned.
- rtapi_app is not yet running: master starts, runs the command and exits again due to instance_count==0. So nothing stays behind. No halrun -U needed. However, this is kind of a low-likelyness race condition: If you manage to break realtime somehow between `halcmd getrt` and `halrun lat.hal`, then no error is reported.

You see that in the following test where I added a message when rtapi_app exits:
```
halcmd loadrt and2
#Note: Using POSIX realtime
halcmd getrt
#Realtime available
halcmd getrt
#Realtime available
pgrep rtapi_app
#3799
halrun -U
#exit master
```
vs
```
halcmd getrt
#Note: Using POSIX realtime
#exit master
#Realtime available
halcmd getrt
#Note: Using POSIX realtime
#exit master
#Realtime available
pgrep rtapi_app
#no process running
```

I don't see any big downside in doing it like that. But i also do not 100% like starting up rtapi_app just to exit right away. Better would be running it with or after halrun in the script. Or might be an approach using a signal / parameter.

Of course, also RTAI / doc and so on needs to be checked / updated before I will call that ready.

Better ideas are welcome. But I prefer using a check executing the same code path for realtime checks always instead the variant before where you then have most likely diverging real time checks spread in all possible apps.

grandixximo:
@hdiethelm I have restructured to avoid the standalone HAL instantiation @BsAtHome flagged:

- latency-histogram now calls `halcmd getrt` from inside its own running session (right after `hal start`), so it attaches to the realtime already up rather than spinning up a segment that needs a separate `halrun -U`.
- latency-test drops its script-side check entirely and relies on the existing `Note: Using POSIX non-realtime` from rtapi, which already lands on the same console. That also removes the double-warning you saw.
- Dropped the `LINUXCNC_NO_RT_WARN` opt-out per @BsAtHome; the dev-suppression policy can be decided in #4118 rather than baked in here.

That keeps the scripts honest, but @BsAtHome's deeper point lands on `getrt` itself: `do_getrt_cmd` goes through `hal_systemv` + a HAL init, so any standalone caller instantiates a segment. Is it worth making `getrt` probe `rtapi_is_realtime()` without a full `hal_init`, so a GUI can ask "is realtime available" cheaply before starting anything? That would let every GUI (including `linuxcnc` started from a launcher, the #4118 case) query it without session side effects. If you think that is the right shape, this PR can depend on that and I will rebase on top once your `getrt` lands as its own PR.

hdiethelm:
@grandixximo Nice. I have to test it.

I created a PR, feel free to rebase:
https://github.com/LinuxCNC/linuxcnc/pull/4132/changes

@BsAtHome 
Yes, rtapi_app has the annoying behavior that it always creates this memory segment and initializes RT. Even if you exit right away after. However, it looks like this doesn't hurt anything, you can start any app after `rtapi_app getrt` or other commands that do not increase the instance counter or do anything else than initializing this segment and exiting afterwards.

Any hint's what should be done in this case? Before my pr rtapi_app rework pr, even  `rtapi_app exit` initialized a memory segment when it was not running. ;-)

grandixximo:
Crossed posts, @hdiethelm. Good, your "run it with/after halrun in the script" is exactly what I did for latency-histogram (getrt after `hal start`, inside the session), so no stray `rtapi_app` and it also avoids the break-in-between race you noted. For latency-test I leaned on the existing rtapi note rather than a second getrt call; happy to switch it to `getrt` inside its HAL flow if you would rather every tool go through the one path. I will rebase on your `getrt` PR once it is up with the RTAI/doc bits.

hdiethelm:
Hmm, just an idea:
Something like this for the test scripts? I find this huge popup's a bit annoying.
grafik

@BsAtHome 
What do you think about this for all GUI apps? TBD how to inhibit but there will be a way.
https://github.com/LinuxCNC/linuxcnc/pull/4132/changes/4803bb1287561a1851da85083f9ccb3ec73efe5a
Somehow gmocappy doesn't show this error. I guess there is a bug that startup errors are not shown?
grafik

BsAtHome:
A (forced) popup is the equivalent of slapping someone in the face.

The error message added to the GUI is actually not an error. Not running RT on a production system <i>may</i> be considered an error.

If you look closely in AXIS' status bar you see "Kein Werkzeug". That is also the place where you want to warn the user. Add a status bar field that is <i>obvious</i> (light red background) yet not <i>invasive</i>.

hdiethelm:
You have a point. I also get annoyed of all this popup's when using the good old microslop... :-D

Now on the status bar: Good idea. How to do that? I am already somewhat deep in the C code, so I can add any needed support there but for GUI's, someone else has to take over.

@grandixximo Can you do this based on whatever from the hal? halcmd / parameter / pin is easy to add for me.

grandixximo:
@hdiethelm thanks, I will rebase this onto #4132 once it settles. @BsAtHome agreed the popup is too much; I will drop the `tk_messageBox` and warn non-invasively in the test tools instead.

On the cross-GUI status bar, that is the right shape and I am happy to do the AXIS side (a status-bar field with a light-red background, like the existing tool slot) once @hdiethelm's "realtime ok" signal from #4132 exists. The question is where it lives.

My preference is to keep #4107 narrow: rebased on getrt, popup gone, scoped to the latency tools. It can ship as soon as #4132 lands. The GUI-wide warning cannot be written until the HAL pin exists and touches AXIS, gmoccapy and qtvcp, so I would do it as a separate PR tracking #4118 rather than make this small change wait on the slowest part.

That said, if you would rather have one PR own the whole intent, I am fine rescoping #4107 to the GUI-wide warning and retitling it; it just becomes larger and slower. Either way works for me. Which do you prefer?

grandixximo:
@hdiethelm to answer your "halcmd / parameter / pin" question directly: a `bool` pin is best for the GUIs. AXIS, gmoccapy and qtvcp already monitor HAL pins, so they can reflect realtime state live in the status bar without polling a command. I would steer away from a param given those are heading for deprecation, and a halcmd is the least convenient since a GUI would have to shell out to poll it.

BsAtHome:
I think we first have to agree on the proper conceptual design of how to detect in the different scenarios and what to do with it.

grandixximo:
@BsAtHome agreed, let me put a concrete proposal on the table.

**Detection:** one source of truth. @hdiethelm's `rtapi_is_realtime()` path, exposed as a single `bool` HAL pin ("realtime ok"). Every app reads the same signal, so there is no divergent per-app logic, which was the original concern.

**What to do with it: per UI, not one mechanism.** The right surface differs by app, so each owns its own rendering rather than forcing a single widget everywhere:
- Console tools (latency-test): the existing `Note: Using POSIX non-realtime` already covers them.
- GUIs (AXIS, gmoccapy, qtvcp): an in-window, non-invasive indicator. Obvious but not a slap, e.g. a light-red status-bar field, no forced popups.

**One thing to rule out:** coloring the window title bar / decoration is not reliable. Plenty of setups have no title bar at all (fullscreen/kiosk panels, some Wayland/WM configs), so the signal has to live inside the app window, not in the chrome.

**Still open (defer):** production-vs-dev suppression policy. That can ride with #4118 once the pin and the per-UI rendering exist.

Does that match how you see the scenarios?

hdiethelm:
Sounds like a plan. I will create new PR with a signal. Then we can test how this feels and continue from there.

I can do that tomorrow, right now I have other things to do.

About the title bar: The idea was to only use this for the two test apps. If this is cumbersome, might be just modify the text that is already displayed in them.

@grandixximo Can you mark this PR as a draft until we are done?

Sorry about the for- and back. If i dont have a good solution yet, this is often my way of brainstoming. Try things and discard until it is good. Hope this is ok for you.

BsAtHome:
> **Detection:** one source of truth. @hdiethelm's `rtapi_is_realtime()` path, exposed as a single `bool` HAL pin ("realtime ok"). Every app reads the same signal, so there is no divergent per-app logic, which was the original concern.

That is only partly satisfactory because for this realtime needs to be running.

You want to know <i>in advance</i> whether your system will be capable of running RT without starting any of it. Then, when you are running, you want to know from various applications what the actual status is by using generic API call or/and HAL pin.

grandixximo:
> Hope this is ok for you.

No problem, we brain storm it and come up with something that sticks.

hdiethelm:
Hmm, like so often, more difficult than estimated. So basically two options are needed.

`halcmd getrt` is gone. halcmd always initializes the hal before anything else, so not suited for this. With uspace, it fully cleans up afterwards, so no real issue. But with rtai, it just fails if rtai is not yet started.

New concept:

Check with script in advance / during runtime:
- `realtime verify` -> Returns 0 if realtime, 1 if not realtime
  - If rtapi_app is not running, this goes trough a fast path that doesn't start the hal
  - If rtapi_app is running, this goes trough the normal command path
  - If rtai build, always realtime

A pin that shows the status:
- New persistent component `hal_status` that is created when the hal is started. 
  - `hal_status.is-realtime` 1 when realtime / 0 otherwise

Some quirks where needed: 
- Inhinbit unload of `hal_status` -> Otherwhise unload all will try to unload a .so / kernel module which doesn't exist.
- The fast path in rtapi_app

Ideas:
- `hal_status.realtime-type`: 0 for no, 1 for RTAI, 2 for posix PREEMPT_DYNAMIC, 3 for PREEMPT_RT and so on

Tested on uspace / rtai, it works.

Alternatives:
- Separate realtime check that checks rtapi_app from the outside. Big downside: I am sure this is getting inconsistent soon or is already at time of implementation and we miss an edge case.
- Separate hal_status component: I started something. But it has also a downside: It must be loaded in each and every config to work. This is why I decided against it.
- Add a pin to a normally loaded component like motion. This is a bit out of place.

Feedback?
Anyone has an alternative idea how to create a pin without having a component directly in hal_lib.c?

hdiethelm:
Hmm, the tests really don't like my persistent hal_status component. Options:
- Just create a separate component that has to be loaded in each and every sample config
- Drop the idea with the pin and only use `realtime verify`
- Fix the tests. A persistent `hal_status` component could be useful for many things
- Pass the value trough shm directly and add it to the hal api's

BsAtHome:
> Hmm, like so often, more difficult than estimated. So basically two options are needed.

Yes, that is why we need a plan first :-)

There are three levels where you want to be able to query the RT state:
1. when nothing is started to see if the system is actually capable of running RT,
2. when starting the system,
3. when running.

These levels of interaction may use different methods to give you the result.

Using HAL while the system is up may not be an entirely bad idea. However, I do not think that most programs would or should use it. Most programs already interact with libraries and the information must be available through the C HAL library and python hal module.

If you want a "persistent" component, then it needs to be HAL itself who does the work and it must be created when the HAL library initializes the shared memory segment. For the namespace, I'd suggest either `hal` or `linuxcnc`. Not sure which is the better one. The status `is-realtime` and `realtime-type` should probably be R/O parameters.

All that said, if the data is available through hal_lib and halmodule, then why bother with a component?

hdiethelm:
> > Hmm, like so often, more difficult than estimated. So basically two options are needed.
> 
> Yes, that is why we need a plan first :-)
> 
Yes, plan, test, figure out the plan was bad, replan, retry... ;-)

> There are three levels where you want to be able to query the RT state:
> 
>     1. when nothing is started to see if the system is actually capable of running RT,
Possible in my PR right now with `realtime verify` https://github.com/LinuxCNC/linuxcnc/pull/4132

> 
>     2. when starting the system,
Before rtapi_app is started, it is stopped. After rtapi_app is started, it runs. So there is no starting. rtapi_app is started at the first loadrt command. Before, you are in 1. / after in 3. / never in 2.

There is console feedback during starting.

> 
>     3. when running.

Possible in my PR right now with `realtime verify` but this is a bash scrip.

This needs something that goes trough the process boundary: 
1. rtapi_app <-> socket <->rtapi_app (the way  `realtime verify` does it when rtapi_app is running)
2. userspace_app <-> hal_lib <-> shm <-> hal_lib <-> rtapi_app 

For a performant query in any code, I would like to avoid executing rtapi_app 1. Left over is the hal 2.

There is the standard way to use pin's / param's to do this. However, no component, no pin as much as I understand it. So the idea with a persistent component I implemented in my PR.

For a hal function like hal_realtime_status() that can be exported to python:

An option would be to add the realtime status directly to hal_data_t, so no need for pin's. Due to hal_data_t lives in shm, this should work. What do you think about this variant?

Are there other ways do pass data trough the above path? I am still trying to understand how all of this works. It is somewhat complex.

> 
> 
> These levels of interaction may use different methods to give you the result.
> 
> Using HAL while the system is up may not be an entirely bad idea. However, I do not think that most programs would or should use it. Most programs already interact with libraries and the information must be available through the C HAL library and python hal module.
> 
> If you want a "persistent" component, then it needs to be HAL itself who does the work and it must be created when the HAL library initializes the shared memory segment. For the namespace, I'd suggest either `hal` or `linuxcnc`. Not sure which is the better one. The status `is-realtime` and `realtime-type` should probably be R/O parameters.

This is implemented in https://github.com/LinuxCNC/linuxcnc/pull/4132. However, it fails some tests, due to they check the pin list and now there is a pin more that is always there. We can think about changing names / pin vs. parameter when the general concept is clear.

TBD if this is a good idea. Before, the hal was clean at startup. Now you always have at minimum one component as soon you initialize realtime and you can not get rid of it. What do you think about this variant?

> 
> All that said, if the data is available through hal_lib and halmodule, then why bother with a component?

hdiethelm:
Had to try it anyway:
Now there is also `hal_realtime_status()` and for python `hal.realtime_status()`
What do you think about this variant?
https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/673d34fbdd7544b6353d3cbfb5ccd2ff09bfa0c1
I will clean all up and document when we decided the final way.

grandixximo:
Looked through #4132 and checked the consumer side locally. I'd side with @BsAtHome: realtime status is fixed at `rtapi_app` startup and never changes during a run (`rtapi_is_realtime()` caches, and `hal_data->realtime_status` is written once), so a reactive HAL pin buys little, while the always-on `hal_status` component costs the clean-HAL-at-startup invariant, the `unloadrt` special-case, and (per your own note) the pin-list tests.

The lib path you built looks sufficient: `hal_data->realtime_status` + `hal_realtime_status()` + the `hal.realtime_status()` Python binding, plus `realtime verify` for scripts. I confirmed AXIS, gmoccapy and qtvcp all `import hal` and `hal.py` re-exports `_hal`, so `hal.realtime_status()` is reachable from every Python GUI. The tcl tools (latency-histogram) have no binding, so they'd use `realtime verify`, which works both before start (your fast path, which also settles the earlier halrun -U point) and while running.

If a HAL pin is ever genuinely needed (LED, classicladder, pyvcp), I'd make it an optional `loadrt hal_status` that just mirrors `hal_realtime_status()`, loaded only where wanted, keeping the default HAL clean.

Two nits: `hal_realtime_status()` should be `(void)` in hal.h, and `realtime_status` could go at the end of `hal_data_t` to limit offset churn. The `realtime-type` enum is a good addition and fits the same shm-field pattern, no pin needed.

hdiethelm:
Ok, so I stick with the following and drop the pin idea.

1. when nothing is started to see if the system is actually capable of running RT
    - `realtime verify`
2. at startup
    - not implemented except console output.
3. when running
    - `realtime verify` (For scripts, while running)
    - hal_data->realtime_status + hal_realtime_status() + the hal.realtime_status() Python binding

Due to realtime_status vs. realtime_type is redundant, the following suggestion:

realtime_status = -1 Uninitialized (this is the case until rtapi_app is not running, starts at first call of loadrt)
realtime_status = 0 NonRT
realtime_status = 1 ... n Different variants active

So you can use:
- hal_realtime_status() < 0: Bug, you call it to early
- hal_realtime_status() == 0: Non-RT
- hal_realtime_status() > 0:  RT
- hal_realtime_status() == RT_STATUS_XENOMAI3 (and all implemented variants)
So you can add specific checks for xenomai / rtai / posix if this is needed somewhere.

The pin idea sounded nice until I started implementing it. And now I also don't like it any more.

> Two nits: hal_realtime_status() should be (void) in hal.h, and realtime_status could go at the end of hal_data_t to limit offset churn. The realtime-type enum is a good addition and fits the same shm-field pattern, no pin needed.

- void: Sure, thanks. POC phase, I don't really try to have good code quality.
- limit offset churn: Can you elaborate? There is also exact_base_period which should be moved always in this case I think.

I will start cleaning everything up.

BsAtHome:
The code in #4132 is complex. The HAL interface is both ugly and unnecessary when you simply call `rtapi_is_realtime()` because that is available in the hal module anyway as a constant property. The only _real_ fix would then be to ensure that `rtapi_is_realtime()` returns the correct value (and could make it in halmodule a live property).

Seen in that light and that we want hal to be "clean" at start and stop... we should drop the fixed created component and <i>may</i> add a separately loadable component if anybody wants or needs it. That is a much cleaner solution. And, FWIW, a components also has access to `rtapi_is_realtime()`.

hdiethelm:
Probably our comments crossed.

`rtapi_is_realtime()` only works in real time context. In user components / python run in a different process, so even having a static value in code that is the status will not work due to this is a different process. shm (HAL) or an other way is needed to pass the process boundary.

See https://github.com/LinuxCNC/linuxcnc/issues/4129

Are there any other standard ways to communicate with the real time process except HAL pin / shm?

hdiethelm:
An option would be a rewrite / fix of `rtapi_is_realtime()`
At the moment, it uses can_set_sched_fifo() / has_setuid_root() which only work when called in the rtapi_app process.

Instead, what we could do is check the rtapi_app executable from externally. More or less:

geteuid() == 0 -> stat(EMC2_BIN_DIR "/rtapi_app", &st)

cap_get_proc() -> cap_get_file(EMC2_BIN_DIR "/rtapi_app") 

rtapi_is_realtime() is already exposed in python: PyModule_AddIntConstant(m, "is_rt", rtapi_is_realtime());

grandixximo:
@hdiethelm the `stat(rtapi_app)` path was already tried and removed in commit `49523cddb7` (the `rtapi_is_realtime()` rework from #3928). [uspace_common.h:450-453](https://github.com/LinuxCNC/linuxcnc/blob/master/src/rtapi/uspace_common.h#L450) documents the three reasons: it only checked the setuid bit (not file caps), it stat-ed `EMC2_BIN_DIR/rtapi_app` rather than the running binary (breaking wrapper installs like NixOS `/run/wrappers`), and it ran before `LINUXCNC_FORCE_REALTIME` was checked. The current `can_set_sched_fifo()` probe was the deliberate replacement.

So the shm field in #4132 is the right path: it is written by rtapi_app into shared memory and readable from any process without re-probing. It also directly fixes #4129: `hal.is_rt`/`is_sim` would derive from `hal_data->realtime_status` rather than the in-process probe, making them correct for unprivileged callers. The socket path (`realtime verify`) works too but spawns a process, fine for scripts, wasteful to poll from a GUI.

hdiethelm:
Agreed.
The shm field / checking it internally is the bullet prove way.
Even thought we can check caps / LINUXCNC_FORCE_REALTIME externally, checking it externally will always have a strange edge case resulting in:
- externally, all looks good
- internally, rt fails

For example, LINUXCNC_FORCE_REALTIME is set in the context where the check is performed but not in the context where rtapi_app is started, just to name one.

BsAtHome:
Checking the shm field (function wrapped) will advise both RT and non-RT what the status is of the <i>running</i> system (as long as HAL is originally instantiated and initialized by the right process).
This is also the value that halmodule should return. Is there a path where halmodule starts before RT is running? I.e., can halmodule instantiate the HAL shm? If so, then there might be a problem.

hdiethelm:
> Checking the shm field (function wrapped) will advise both RT and non-RT what the status is of the running system (as long as HAL is originally instantiated and initialized by the right process). This is also the value that halmodule should return. Is there a path where halmodule starts before RT is running? I.e., can halmodule instantiate the HAL shm? If so, then there might be a problem.

Yes, there is. If you do:
``` python
import hal 

h = hal.component("passthrough")
print("HAL realtime_status " + str(hal.realtime_status()))
```
the result is:
`HAL realtime_status -1`

Described in https://github.com/LinuxCNC/linuxcnc/pull/4107#issuecomment-4638051717

If you do `halcmd loadrt and2`, realtime is started and you will get:
`HAL realtime_status 1`

This can be mitigated by starting rtapi_app at the first instantiation of the hal and only exiting at exit. However, this might have other side effects. So I would like to to this later.

hdiethelm:
All this rtapi_app start / exit is a bit fuzzy. It is started at the first load_rt command and exits when there is no rt module loaded any more.

This results in sometimes funny behavior: `halcmd debug` will only have an effect when it is after the first loadrt command. If it is before, rtapi_app starts, does the debug command and exits immediately. So the status is lost and the command has no effect.

Better would be one of those:
- start at start command / exit at exit command.
- rtapi_app_deamon / rtapi_app_cmd (one is the previous master / one the previous client)

But this opens a new can of worms I would like to avoid right now.

hdiethelm:
So, I continued:
https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/ed8ef46df88b8f18170600bc0996fda74870133d: Removes the persistent component again
https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/f80af108466acc090094a0de2958bc14fdfdd607: Moves the detection to uspace_rtapi_main.cc
https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/97bb37747c009a87f62b1237da8435e39f6925d6: Reworks the whole detection and realtime_status() is now a status exactly reflecting the effective status.
All still WIP and a bit rough, but the concept should be visible.

grandixximo:
I've rebased this onto `realtime verify` from #4132 (latency-histogram now asks the realtime layer instead of the removed `halcmd getrt`).

Correction to myself: latency-test is not console-only, it brings up a pyvcp panel, so a user launching it from a menu would not see the `Note: Using POSIX non-realtime` on the console either. So the question of how to surface an RT warning in a GUI applies to latency-test as much as to latency-histogram, and that ties into the still-open GUI-warning discussion (#4118) rather than something to settle here. I'll hold on changing latency-test until we agree on how in-UI warnings should look.

hdiethelm:
Back to the discussion of how to show it, options:
- Popup
- Sneek in a title bar change at the low level UI functions
```diff
diff --git a/src/hal/user_comps/pyvcp.py b/src/hal/user_comps/pyvcp.py
index f8b9bb2216..06f2178eec 100755
--- a/src/hal/user_comps/pyvcp.py
+++ b/src/hal/user_comps/pyvcp.py
@@ -40,6 +40,7 @@ sys.path.insert(0, os.path.join(BASE, "lib", "python"))
 
 import vcpparse
 import hal
+import lcnc_realtime
 
 import tkinter as Tkinter
 from tkinter import Tk
@@ -87,6 +88,10 @@ def main():
    
     vcpparse.filename=filename
     pycomp=vcpparse.create_vcp(compname=component_name, master=pyvcp0)
+
+    if not lcnc_realtime.verify():
+        pyvcp0.title(pyvcp0.title() + " NO REALTIME")
+
     pycomp.ready()
 
     try:
```
Bildschirmfoto vom 2026-06-24 16-49-58

and similar for other UI classes.

- UI specific
  - latency-test / latency histogram: Add red text where the OS information is
  - axis: Somewhere in the footer next to "No Tool"
  - ...

For me, a popup that I can disable would be fine.  
Either with an environment variable like LINUXCNC_DISABLE_RT_WARN for example or in a config like ~/.linuxcncrc. But it should be global, not per machine configuration.
This popup should show a way how to disable it. But there are other opinions.

grandixximo:
I like the footer bar better even for latency test and hist

image

image

can pack more info, and it's more visible than the title bar IMO
I stared at your picture for a couple of minutes before realizing NO REALTIME was in the title bar.
Me like what I did better, but that's called bias, thoughts?

rodw-au:
Seeing you are working on this, what would be a useful feature to add would be a command line switch that ran for a specific **--time --quiet**(ly) and print the latency results to the console so scripts could report latency. I've wanted to do this recently and had a lot of trouble finding a way to do this. cyclictest allows you to do this but it's still not easy and not within the LinuxCNC ecosystem.

hdiethelm:
Looks nice in the footer! Visible but not a slap in the face.

I did not know how to do this, mainly the reason why I used the title bar to add some ideas.

Is this also possible in axis / gmoccapy?

grandixximo:
> Seeing you are working on this, what would be a useful feature to add would be a command line switch...

Good idea, belongs as a --quiet/--seconds mode of latency-test, please open an issue, it's a different PR.

> Is this also possible in axis / gmoccapy?

Yes, but On the main GUIs I'd show just the realtime field (`RT-<type>` / `no realtime`); the governor and isolcpus stay in latency-test / latency-histogram.

Placement, reusing each GUI's existing status area:
- Axis: the existing status footer (the "No Tool" / position row).
- QtDragon: its status bar.
- Gmoccapy: there's room near the clock. (I don't really see anywhere else, and I don't know if we are allowed to make it bigger with a footer, to discuss in #4118?)
- Touchy: it already has a status bar; it goes there.

Just the one field via `lcnc_realtime.verify()` plus the type for the label.

Keep it as its own PR, separate from this one, it's the cross-GUI convention #4118 should settle. This PR lands as the latency-tools footer.

scripts/latency-histogram
```diff
@@ -335,6 +350,56 @@ proc load_packages {} {
...
+  # realtime: concise type when available, red "no realtime" when not.
+  # 'realtime verify' is authoritative for the yes/no; the type label is
+  # derived from the same kernel signals rtapi uses, for display only.
+  set rt_ok 0
+  if {![catch {exec linuxcnc_var REALTIME} rt]} {
+    if {![catch {exec $rt verify 2>@1}]} {set rt_ok 1}
+  }
+  if {$rt_ok} {
+    set t SCHED_FIFO
+    if {[string match -nocase *-rtai* [exec uname -r]]} {
```
hdiethelm: That looks like a maintenance trap. One future change in rtapi_app and this breaks.

scripts/latency-test
```diff
@@ -81,6 +81,29 @@ if [ "$BASE" -eq "$SERVO" ]; then BASE=0; fi
 BASE_HUMAN=$(human_time "$BASE")
 SERVO_HUMAN=$(human_time "$SERVO")
 
+# --- status-bar info (axis-style footer; colour only when attention needed) ---
+# 'realtime verify' is authoritative for yes/no; the type label is derived from
+# the same kernel signals rtapi uses, for display only.
+RED="#cc4444"
+if realtime verify >/dev/null 2>&1; then
+    if uname -r | grep -qi -- '-rtai'; then RT_T="RTAI"
```
hdiethelm: Same here.

hdiethelm:
It looks good with the footer.

However, I don't like the now 2x copy paste real time check code. This is guaranteed to brake soon and be not consistent. For example, for xenomai, rtapi_app still needs setuid, so it is already not consistent when I use setcap on a xenomai kernel:
grafik

Now I wold rather not show the type than doing it this way.

Right now, you can check the effectively running type by using python `type = hal.get_realtime_type()` but this is a bit cumbersome, it works only when rtapi_app is running, so you would have to do that after `halrun lat.hal` in a separate thread.

Alternatives:
- `realtime verify`: I could print the realtime type on stdout. So I can capture it in python and forward it.
- Using non-standard return values: Probably a bad idea. I would have to use 1...9 for the types and 0 for error
- The hard way: Properly separate rtapi_app in master / client and create a client library, so library calls can be used to communicate with the rtapi_app master.

hdiethelm:
I tried out the variant passing the text from rtapi_app trough the socket up to `realtime verify` and python:
https://github.com/hdiethelm/linuxcnc-fork/tree/fix/latency-setuid-warning-4044
https://github.com/hdiethelm/linuxcnc-fork/commit/4d802490da133fddcecb0064a3c4763da57f0bc4

It works but just quickly coded together, needs some doc and tidy up. Do you think this will do the job?

I changed only latency-test to use this. TCL is unreadable for me, would take me hours to do it... ;-)

andypugh:
> Happy to drop the latency-test console line entirely if you would rather the C note be the only console source.

I don't think this is worthwhile, as that line is also useful to tell you _which_ realtime system you are using.
So I think it is good to say _which_ RT you are on, and additionally warn that it is _not_ RT. 

> 3. Dev opt-out policy. ... The env var is easy to set once for all test configs, which is why I started there.

I think I like the env var. the _vast_ majority of users need to be warned. Those who habitually test-run the code on a non-RT system are unlikely to mind clicking-away a warning.

scripts/latency-histogram
grandixximo: adapted on top of your POC commit

scripts/latency-test
grandixximo: same here ;-)

grandixximo:
@hdiethelm @BsAtHome a heads-up on the `rtapi_app` part of this PR.

To give the status bar an authoritative realtime type, I folded @hdiethelm's `check_rt`-returns-the-type-string POC (`4d802490da`) into this PR. That expands the scope here beyond the GUI status bar into `rtapi_app` (`src/rtapi/uspace_rtapi_main.cc`). hdiethelm built his refinement on top of my work rather than as its own PR, so merging and expanding scope felt like the natural move to avoid a rebase-and-wait. I do not mind splitting it back out into a separate PR if that is preferred, @BsAtHome especially if you would rather review the rtapi change on its own.

What the change does: the `check_rt` socket result is extended to return a type string alongside the int status, and the realtime-type enum is mapped to a human-readable name (`Preempt RT`, `RTAI`, `Xenomai`, ..., or `No realtime`). The serialization mirrors the existing `send_args`/`recv_args` framing.

I added one tidy commit on top (`40300eedd2`), kept separate so the delta reads on its own:

- `recv_result()` now checks the second `recv_data()` for a short read, like `recv_args()` already does.
- Guard `recv_result()`/`recv_args()` against a frame size below its own length prefix, which would underflow `buff_size`.
- Dropped a dead `out.resize()` overwritten on the next line.
- Moved the type-to-name map into `realtime_type_name()` with an exhaustive switch and no `default`, so `-Wswitch` flags a new enum value.
- Typo fixes (`to big`) and a stale `send_args` comment in `send_result()`.

@hdiethelm could you confirm you are happy with the tidy and drop the POC marker from your commit if it is good to land? It is your RT domain, so I did not touch the protocol design itself.

@BsAtHome a review of the wire-format and C++ would be appreciated since that is your area. Happy to adjust to whatever framing convention you prefer.

grandixximo:
@andypugh both your points are covered by where this landed, though the shape changed since that comment.

On showing which RT: the latency-test / latency-histogram footer now shows the realtime type directly (`Preempt RT`, `RTAI`, `Xenomai`, ..., or a red `No realtime`), sourced authoritatively from `realtime verify` rather than sniffed. So it says which RT you are on and flags when it is absent, in-window, which also covers the menu-launched case where the console note is not visible. The rtapi `Note: Using POSIX ...` console line is untouched.

On the env-var opt-out: I dropped `LINUXCNC_NO_RT_WARN` after Bertho's point. The footer is non-invasive (a status-bar field, not a popup), so there is nothing to click away and no opt-out is needed here. If a more assertive GUI-wide warning ever does need suppression, that policy belongs in #4118 alongside the cross-GUI work, not baked into this small PR.

hdiethelm:
> @hdiethelm could you confirm you are happy with the tidy and drop the POC marker from your commit if it is good to land? It is your RT domain, so I did not touch the protocol design itself.

I will look into it again let's say until tomorrow. Thanks for pulling my commit in and correcting some parts. I have to check how I can allow you to push to a repo under my github account. We are working at the same thing much lately and having two branches working at the same issue just generates unnecessary friction.

grandixximo:
I could have revised the original commit and pushed under your name from my end, as the author could have stayed, but I really did not want to do that, I'd rather keep it as it happened for now, and I can squash later, or possibly separate the concerns in two PR, waiting for Bertho on the final judgement on that. The friction is more about waiting your PRs to land. Pulling in your commit and commit on top is zero friction for me. I can push PRs to your fork, I just decided to go this route instead, why does it create friction on your end? You can just pull merge and commit on top, no? I can merge your changes here when you are done, I don't see much friction.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -614,46 +632,131 @@ static ssize_t recv_data(int fd, void *buf, size_t n, int flags) {
...
+        ret |= ((int)(unsigned char)buf.at(idx + i) << i * 8);
+    }
+    return ret;
+}
+
+static bool send_result(int fd, int result, const std::string &out) {
+    //Calculate size
+    size_t buff_size = 0;
+    buff_size += sizeof(int) + 2 * sizeof(uint16_t);
+    buff_size += out.size();
```
hdiethelm: Just use:
`size_t buff_size = sizeof(uint16_t) + sizeof(int) + sizeof(uint16_t) + out.size();`
I know that's my code but now it's in your branch... ;-) This way, the sizes are also in the order that is used later.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -614,46 +632,131 @@ static ssize_t recv_data(int fd, void *buf, size_t n, int flags) {
  * int
```
hdiethelm:
``` 
* result is serialized as:
 * uint16_t size (full package size including the size field)
 * int result
 * uint16_t out_size
 * char[out_size] out
```

src/rtapi/uspace_rtapi_main.cc
hdiethelm: * master processes the args and returns result and stdout string
 * master->client: int result, std::string out

hdiethelm:
> I could have revised the original commit and pushed under your name from my end, as the author could have stayed, but I really did not want to do that, I'd rather keep it as it happened for now, and I can squash later, or possibly separate the concerns in two PR, waiting for Bertho on the final judgement on that. The friction is more about waiting your PRs to land. Pulling in your commit and commit on top is zero friction for me. I can push PRs to your fork, I just decided to go this route instead, why does it create friction on your end? You can just pull merge and commit on top, no? I can merge your changes here when you are done, I don't see much friction.

Ok, fine for me. Waiting for other PR's is for sure friction. You can also squash your and my commit together and just mark me in the comment.

So we just go this route, I push a commit to my version of this branch and then you pull it in and squash it at the end.

Most is fine, except some not updated comments. I will quickly update them and the you can pull in my commit. You can remove the POC at the squash phase.

hdiethelm:
Cleanup from my side: https://github.com/hdiethelm/linuxcnc-fork/commit/b7f4618e2761c7074a4995881b5ebcbe0c041d55
This will close my comments.

scripts/latency-histogram
```diff
@@ -111,6 +111,21 @@ proc which_exe {name} {
   return -code error "$name: executable not found"
 } ;# which_exe
 
+proc check_rt_privileges {} {
```
hdiethelm: Is this needed? If I run `latency-histogram`, it says:
```
  Process capabilities: cap_sys_nice=no cap_ipc_lock=no.
  Falling back to POSIX non-realtime.
  Fix: 'sudo make setcap' (preferred) or 'sudo make setuid' on rtapi_app.
  Override (testing only): set LINUXCNC_FORCE_REALTIME=1.
Warning: realtime is not available.
Threads get no SCHED_FIFO priority or locked memory,
so the latency shown will be far worse than reality.
On a run-in-place build, run 'sudo make setuid' (or 'sudo make setcap') in src/.
```
so it is telling me twice more or less the same thing. If you are not happy with the text from rtapi_app, I suggest you change it there, so it is the same everywhere.

hdiethelm: The main text is from here:
https://github.com/LinuxCNC/linuxcnc/blob/1db4ebe2b71ded81713c0304aca04d3aa921ece5/src/rtapi/uspace_rtapi_main.cc#L184
The `Note:` is from here:
https://github.com/LinuxCNC/linuxcnc/blob/1db4ebe2b71ded81713c0304aca04d3aa921ece5/src/rtapi/uspace_posix.cc#L251

grandixximo:
@hdiethelm done on all your points, branch force-pushed.

- Pulled in your cleanup commit (`Pass realtime string: Cleanup`), so the protocol doc-comment and the `send_result` size calc are your versions.
- Dropped the duplicate non-realtime warning in latency-histogram: `check_rt_privileges` was repeating what `rtapi_app` already prints on the console, and the footer surfaces it persistently in the GUI, so the script no longer says it a second time. That closes your `scripts/latency-histogram` comment.
- Squashed the branch into two commits, as we agreed:
  - `rtapi_app: return the realtime type name from check_rt` (your POC + cleanup + my tidy), authored by you, POC marker gone, with me as co-author for the tidy fixups.
  - `latency-test, latency-histogram: add realtime status bar` (the GUI side), mine.

The tidy fixups folded into your commit are the short-read check in `recv_result()`, the frame-size underflow guard in `recv_result()`/`recv_args()`, the exhaustive type-name switch, and the typo fixes; shout if you would rather any of those drop.

@BsAtHome the rtapi_app wire-format change is now a single self-contained commit (`c93e8e857b`) if you want to look at it on its own.

scripts/latency-histogram
grandixximo: Dropped the duplicate, see below

src/rtapi/uspace_rtapi_main.cc
grandixximo: merged your commit, and squashed, see below

hdiethelm:
> @hdiethelm done on all your points, branch force-pushed.

I went trough the PR again and tested it. Looks good!

PR #4109, opened 2026-06-03
docs(build): relocate generated manpages under docs/build/man/, build translated

grandixximo:
Follow-up to #4081. Moves the generated troff manpage tree out of the source tree into `docs/build/man/`.

Layout follows the `/usr/share/man` hierarchy (per your call, Bertho): English at `build/man/manN` for the C/default locale, translations at `build/man/<lang>/manN`. This differs on purpose from the `build/html/<lang>` convention, because `man(1)` resolves a MANPATH entry by locale and `LANG=C` only checks `<path>/manN`; the English special-casing is documented in the Submakefile.

It also builds the translated manpages into that tree: po4a emits the generated comp manpages straight to `build/man/<lang>/`, and the `.adoc`-sourced ones render there from their po4a-translated adocs (per-language HTML reuses the same rule). This folds in what was a separate translated-manpages branch, since the relocation already carries the translated plumbing and shipping it half-translated would be odd.

Touches the manpage generation/install paths throughout: `docs/src/Submakefile`, the hal component/user_comp Submakefiles, `install-man` + swish (`src/Makefile`), Doxygen `MAN_OUTPUT`, halcompile's run-in-place fallback, `gen_complist.py`, the run-in-place `MANPATH` (`rip-environment`), `po4a.cfg` man_def paths (structural only, no string changes), `.gitignore`, and the `man -M` example in `tutorial.adoc`.

It also relocates a few committed but unreferenced pre-asciidoctor groff leftovers (`an-old-fixed.tmac`, `stylesheet.9`, `man/images/`) into `docs/src/man/` rather than discarding them. Open question: should I just delete these instead?


BsAtHome:
There is one thing I just noted again. The man-page translations are made on basis of the troff output and not on basis of the adoc sources. That will have to change too, eventually. Probably best not to do it with this PR, but this is should be a reminder. Probably best done when we separate man-page pot/po out of main body.

grandixximo:
>That will have to change too, eventually. Probably best not to do it with this PR, but this is should be a reminder.

Noted

In case you missed it in the PR body, this also relocates a few committed but unreferenced pre-asciidoctor groff leftovers (an-old-fixed.tmac, stylesheet.9, man/images/) into docs/src/man/ rather than discarding them. Open question: should I just delete these instead?

BsAtHome:
> In case you missed it in the PR body, this also relocates a few committed but unreferenced pre-asciidoctor groff leftovers (an-old-fixed.tmac, stylesheet.9, man/images/) into docs/src/man/ rather than discarding them. Open question: should I just delete these instead?

I was wondering about them. They are unused afaik. Therefore, should be removed.

If we get complaints or into trouble,... we'll just revert that ;-)

hansu:
> There is one thing I just noted again. The man-page translations are made on basis of the troff output and not on basis of the adoc sources. That will have to change too, eventually. Probably best not to do it with this PR, but this is should be a reminder. Probably best done when we separate man-page pot/po out of main body.

The build output put that into your mind but it is actually building the html man pages from the adoc sources.

See this fix:  https://github.com/LinuxCNC/linuxcnc/pull/4117

grandixximo:
> > There is one thing I just noted again. The man-page translations are made on basis of the troff output and not on basis of the adoc sources. That will have to change too, eventually. Probably best not to do it with this PR, but this is should be a reminder. Probably best done when we separate man-page pot/po out of main body.
> 
> The build output put that into your mind but it is actually building the html man pages from the adoc sources.
> 
> See this fix: #4117

Thought he was talking about man_def, that 'leg' still on troff. I'm refactoring a bit this PR, building all man in english when not translated, like that we have all main page links populated, so they can be navigated to and will have tag from topbar when merged that will show ..% translated

PR #4110, opened 2026-06-03
docs(build): render English and .dot SVGs into build/, keep docs/src clean

grandixximo:
Follow-up to #4081, applying the principle that nothing generated should live in the source tree (Bertho's gitignore review).

- `components_gen.adoc` (the auto-generated HAL component list) now renders into `build/adoc/en/` and the English build reads it from there, instead of writing into `docs/src/hal/`.
- the SVGs rendered from the `.dot` sources now land in `build/` instead of `docs/src/code/`.

No output changes; only the generated files move out of the source tree.


BsAtHome:
This might accumulate some conflicts when #4109 gets merged. Just thinking out loud...

BsAtHome:
And yes, it created a conflict as expected...

grandixximo:
I will rebase, no worries

grandixximo:
This should merge clean, checked built twice, only complains about dead links, which will have to be looked at in another PR

BsAtHome:
Agreed. Fixing links are the usual janitorial work.

PR #4111, opened 2026-06-03
hm2_eth: add nftables firewall backend

grandixximo:
Follow-up to #3964. That PR added automatic interface isolation via iptables; @NTULINUX noted nftables-only systems have no iptables binary at all, so the auto-skip just leaves them unprotected.

This adds an nft backend alongside the existing one.

## What changed

* New `firewall=auto|iptables|nft|none` module parameter. `no_iptables=1` kept as a deprecated alias for `firewall=none`.
* `auto` (default) prefers iptables when usable, preserving current behavior, and falls back to nft. Pure-nftables systems get isolation out of the box.
* nft rules live in a private `table inet hm2_eth` with an `output`-hook chain. Because the table is dedicated, flush and delete never touch the user's other rules. The `inet` family handles IPv4 and IPv6 in one chain.
* Rules mirror the iptables set: board accept (saddr/daddr/sport/dport), per-interface ICMP drop, admin-prohibited reject, and IPv6 drop.
* hm2_eth(9) updated: new parameter, backend notes, and a manual nft recipe matching the existing iptables one.

## Tested

Configured `--with-realtime=uspace` and built; hm2_eth compiles and links clean. Not hardware-tested, I do not have a Mesa ethernet card. The iptables path is unchanged, so existing setups see no difference.

cc @NTULINUX, @hdiethelm

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/hm2_eth.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_eth.c
@@ -95,8 +95,8 @@ RTAPI_MP_ARRAY_STRING(config, MAX_ETH_BOARDS, "config string for the AnyIO board
 int debug = 0;
 RTAPI_MP_INT(debug, "Developer/debug use only!  Enable debug logging.");
 
-static int no_iptables = 0;
-RTAPI_MP_INT(no_iptables, "Skip automatic iptables rule installation; firewall must be configured externally.");
+static char *firewall = 0;
+RTAPI_MP_STRING(firewall, "Firewall backend for traffic isolation: auto (default), iptables, nft, or none.");
 
 static int boards_count = 0;
 
@@ -464,36 +464,34 @@ static hm2_eth_t boards[MAX_ETH_BOARDS];
 static int eth_socket_send(int sockfd, const void *buffer, int len, int flags);
 static int eth_socket_recv(int sockfd, void *buffer, int len, int flags);
 
-// hm2_eth installs iptables/ip6tables rules to isolate the dedicated Mesa
+// hm2_eth installs firewall rules to isolate the dedicated Mesa
 // interface from non-realtime traffic.  rtapi_app raises CAP_NET_ADMIN
 // into its ambient set at startup (see uspace_rtapi_main.cc), so the
-// caps survive execve() into the iptables binaries even when we run
+// caps survive execve() into the firewall binaries even when we run
 // under file caps instead of setuid root.
+//
+// Two backends are supported and selected with the `firewall` module
+// parameter (default auto): legacy iptables/ip6tables, and nftables.
+// Auto-detection prefers iptables when it is usable (preserving the
+// historical behaviour) and falls back to nft, which is the only
+// backend present on iptables-less systems.
 #define IPTABLES_BIN  "/sbin/iptables"
 #define IP6TABLES_BIN "/sbin/ip6tables"
+#define NFT_BIN       "/sbin/nft"
 #define CHAIN "hm2-eth-rules-output"
+// nft keeps its own table so a flush/delete never touches user rules.
+// inet covers IPv4 and IPv6 in a single chain.
+#define NFT_TABLE "hm2_eth"
+#define NFT_CHAIN "output"
 
-// run_iptables(): fork+exec the named iptables binary with a
-// NULL-terminated argv list (sentinel == NULL, do not omit).  Returns
-// the child exit status, or -1 on spawn/wait failure.  When quiet, the
-// child's stdout+stderr are suppressed so probe-style "is this rule
-// present?" calls do not spam the log on the (expected) failure path.
-static int run_iptables(const char *bin, int quiet, ...) {
-    char *argv[24];
-    int n = 0;
-    // argv[0] is the program name iptables expects, not the path.
-    argv[n++] = (char *)(strrchr(bin, '/') ? strrchr(bin, '/') + 1 : bin);
+typedef enum { FW_NONE = 0, FW_IPTABLES, FW_NFTABLES } fw_backend_t;
 
-    va_list ap;
-    va_start(ap, quiet);
-    while(n < (int)(sizeof(argv)/sizeof(argv[0])) - 1) {
-        char *a = va_arg(ap, char *);
-        if(!a) break;
-        argv[n++] = a;
-    }
-    argv[n] = NULL;
-    va_end(ap);
-
+// run_cmd(): fork+exec bin with a caller-built, NULL-terminated argv
+// (argv[0] is the program name).  Returns the child exit status, or -1 on
+// spawn/wait failure.  When quiet, the child's stdout+stderr are
+// suppressed so probe-style "is this present?" calls do not spam the log
+// on the (expected) failure path.
+static int run_cmd(const char *bin, int quiet, char *const argv[]) {
     posix_spawn_file_actions_t fa, *pfa = NULL;
     if(quiet && posix_spawn_file_actions_init(&fa) == 0) {
         if(posix_spawn_file_actions_addopen(&fa, STDOUT_FILENO,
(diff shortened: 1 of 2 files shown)


src/hal/drivers/mesa-hostmot2/hm2_eth.c
```diff
@@ -515,64 +531,143 @@ static int run_iptables(const char *bin, int quiet, ...) {
...
+    if(NFT(0, "add", "chain", "inet", NFT_TABLE, NFT_CHAIN,
+           "{ type filter hook output priority 0; policy accept; }") != 0) {
+        LL_PRINT("ERROR: failed to create nft chain " NFT_CHAIN "\n");
+        return false;
+    }
+    return true;
+}
+
+static fw_backend_t fw_backend = FW_NONE;
+static int fw_state = -1;
```
hdiethelm: You could use an enum for the state, so it is clear what it means.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
```diff
@@ -1663,7 +1788,7 @@ int rtapi_app_main(void) {
 error:
     for(i = 0; i<MAX_ETH_BOARDS && board_ip[i] && board_ip[i][0]; i++)
         close_board(&boards[i]);
-    if(use_iptables()) clear_iptables();
+    if(use_firewall()) clear_firewall();
```
hdiethelm: `if(use_firewall()) cleanup_firewall();` -> `cleanup_firewall();` and check inside if a cleanup is needed, same as in install_*
Same for all instances of `cleanup_firewall();`

src/hal/drivers/mesa-hostmot2/hm2_eth.c
```diff
@@ -1676,7 +1801,7 @@ void rtapi_app_exit(void) {
     for(i = 0; i<MAX_ETH_BOARDS && board_ip[i] && board_ip[i][0]; i++)
         close_board(&boards[i]);
 
-    if(use_iptables()) cleanup_iptables();
+    if(use_firewall()) cleanup_firewall();
```
hdiethelm: Same here.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
```diff
@@ -96,7 +96,10 @@ int debug = 0;
 RTAPI_MP_INT(debug, "Developer/debug use only!  Enable debug logging.");
 
 static int no_iptables = 0;
-RTAPI_MP_INT(no_iptables, "Skip automatic iptables rule installation; firewall must be configured externally.");
+RTAPI_MP_INT(no_iptables, "Deprecated alias for firewall=none; skip automatic firewall rule installation.");
```
hdiethelm: I would just remove this, it was introduced only a few weeks ago and was never in a release.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
```diff
@@ -621,49 +716,79 @@ static int install_iptables_board(int sockfd) {
     res = getpeername(sockfd, &dstaddr, &addrlen);
     if(res < 0) return -errno;
 
-    if(!use_iptables()) return 0;
+    if(!use_firewall()) return 0;
```
hdiethelm: I think you can move this up a few lines, no need for getsockname() / getpeername() if we don't use it.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
grandixximo: Done with enum fw_state

src/hal/drivers/mesa-hostmot2/hm2_eth.c
grandixximo: Done, if I understood correctly

src/hal/drivers/mesa-hostmot2/hm2_eth.c
grandixximo: removed

src/hal/drivers/mesa-hostmot2/hm2_eth.c
grandixximo: moved

src/hal/drivers/mesa-hostmot2/hm2_eth.c
hdiethelm: There is no real need to use defined values for the enum. I would simplify it to:
`typedef enum { FW_UNRESOLVED , FW_UNAVAILABLE, FW_READY } fw_state_t;`
Otherwise way nicer to read now.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
grandixximo: Done

src/hal/drivers/mesa-hostmot2/hm2_eth.c
grandixximo: simplified

src/hal/drivers/mesa-hostmot2/hm2_eth.c
```diff
@@ -1663,7 +1794,7 @@ int rtapi_app_main(void) {
 error:
     for(i = 0; i<MAX_ETH_BOARDS && board_ip[i] && board_ip[i][0]; i++)
         close_board(&boards[i]);
-    if(use_iptables()) clear_iptables();
+    clear_firewall();
```
hdiethelm: This should probably be cleanup_firewall(); due to it is an error exit where all should be cleaned up.

grandixximo: fixed, squashed and force-pushed

hdiethelm:
As always, while looking trough the code, I found some things that where not subject to your change. Thanks for fixing.

hdiethelm:
From your side, it is ready? So I will do some testing later today. It will take some time, many options to test... Three firewall variants / vs. root/nonroot so 6 combinations.

grandixximo:
Yes, ready, I cannot test, waiting on feedback, probably will pick this up tomorrow, addressing test results ;-)

hdiethelm:
So, i tested all combinations: suid / setcap vs. auto iptables installed / auto iptables not installed / nft / iptables
Looks all good and I can not ping the board when linuxcnc is running. All is cleaned up well.

Below the tables when linuxcnc is running to verify. Looks good to me but I don't know this well enought to properly verify it.

Rules iptables:
sudo iptables -S
```
-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT
-N hm2-eth-rules-output
-A OUTPUT -j hm2-eth-rules-output
-A hm2-eth-rules-output -s 192.168.1.120/32 -d 192.168.1.121/32 -p udp -m udp --sport 43361 --dport 27181 -j ACCEPT
-A hm2-eth-rules-output -o enp2s0f0 -p icmp -j DROP
-A hm2-eth-rules-output -o enp2s0f0 -j REJECT --reject-with icmp-admin-prohibited
```
sudo ip6tables -S
```
-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT
-N hm2-eth-rules-output
-A OUTPUT -j hm2-eth-rules-output
-A hm2-eth-rules-output -o enp2s0f0 -j DROP
```
Rules nft:
sudo nft list ruleset
```
table ip filter {
	chain OUTPUT {
		type filter hook output priority filter; policy accept;
	}
}
table ip6 filter {
	chain OUTPUT {
		type filter hook output priority filter; policy accept;
	}
}
table inet hm2_eth {
	chain output {
		type filter hook output priority filter; policy accept;
		ip saddr 192.168.1.120 ip daddr 192.168.1.121 udp sport 54897 udp dport 27181 accept
		oifname "enp2s0f0" ip protocol icmp drop
		oifname "enp2s0f0" reject with icmp admin-prohibited
		oifname "enp2s0f0" ip6 version 6 drop
	}
}
```

BsAtHome:
Can someone also, please, cleanup the cppcheck mess caused by the hm2_eth.c changes?
(somehow, I don't think that ignoring it would be the appropriate solution)

```
hal/drivers/mesa-hostmot2/hm2_eth.c:542:8: portability: Passing NULL after the last typed argument to a variadic function leads to undefined behaviour. [varFuncNullUB]
    if(IPT(1, "-n", "-L", "INPUT") != 0)
       ^
hal/drivers/mesa-hostmot2/hm2_eth.c:545:8: portability: Passing NULL after the last typed argument to a variadic function leads to undefined behaviour. [varFuncNullUB]
    if(IPT(1, "-n", "-L", CHAIN) != 0) {
       ^
...
```

NTULINUX:
Is this accurate?

```
#define IPT(quiet, ...)  run_cmd(IPTABLES_BIN,  (quiet), __VA_ARGS__, (char *)0)
#define IP6T(quiet, ...) run_cmd(IP6TABLES_BIN, (quiet), __VA_ARGS__, (char *)0)
#define NFT(quiet, ...)  run_cmd(NFT_BIN,       (quiet), __VA_ARGS__, (char *)0)
```

hdiethelm:
This was introduced here:
https://github.com/LinuxCNC/linuxcnc/pull/3964
https://github.com/LinuxCNC/linuxcnc/actions/runs/26539292778/job/78176412787

@grandixximo If @NTULINUX version does not work, I would just pass the arguments as a single string and split by spaces.

@BsAtHome Recurring issue. It probably would make sense to maintain a list of wontfixnow files in cppcheck.sh / shellcheck.sh and fail the CI if any other files have issues. So no new issues are introduced in any other files than in this list.

BsAtHome:
> @BsAtHome Recurring issue. It probably would make sense to maintain a list of wontfixnow files in cppcheck.sh / shellcheck.sh and fail the CI if any other files have issues. So no new issues are introduced in any other files than in this list.

No, we have to be more careful when changing stuff. Ignoring files or difficult issues sets a precedent that "it is ok to be sloppy". No, it is not and things need to get fixed (the proper way). That is also how we were able to add -Werror to the CI build. We just need some continuous stability here too and then we'll enforce it in CI.

With respect to shellcheck,... almost there. Still pondering some options. But my attention is already spread thin by doing all the review work in the necessary docs build reorg and fixes (while working on hal getter/setter and pondering other issues too).

hdiethelm:
> > @BsAtHome Recurring issue. It probably would make sense to maintain a list of wontfixnow files in cppcheck.sh / shellcheck.sh and fail the CI if any other files have issues. So no new issues are introduced in any other files than in this list.
> 
> No, we have to be more careful when changing stuff. Ignoring files or difficult issues sets a precedent that "it is ok to be sloppy". No, it is not and things need to get fixed (the proper way). That is also how we were able to add -Werror to the CI build. We just need some continuous stability here too and then we'll enforce it in CI.
> 
> With respect to shellcheck,... almost there. Still pondering some options. But my attention is already spread thin by doing all the review work in the necessary docs build reorg and fixes (while working on hal getter/setter and pondering other issues too).

Of course, this list should be as small as possible and disappear soon. I just discovered that cppcheck is now passing normally (except the new issue introduced). So we could fail the CI if cppcheck fails. I can look into this.

Right now the CI always passes on cppcheck / shellcheck issues which is even worse, so new issues get introduced when you are not checking this CI stage manually.

grandixximo:
Fixed the cppcheck varFuncNullUB findings too: the IPT/IP6T/NFT sentinel is now `(char *)NULL`. Local cppcheck is clean for hm2_eth.c.
NTULINUX's nft commands work, so no split needed. Args go as discrete argv tokens straight to posix_spawn, no shell or space-splitting.

hdiethelm:
Thanks for fixing cppckeck. Hmm, while looking at this:
Why not directly pass argv?

I don't have time to test this right now, but this compiles. It could be simplified with the matching define.
```
static int run_iptables(const char *bin, int quiet, char **argv) {
...
}
run_iptables(IPTABLES_BIN, 1, (char *[]){IPTABLES_BIN_NAME, "-n", "-L", "INPUT", NULL});
//or
char *argv[] = {IPTABLES_BIN_NAME, "-n", "-L", "INPUT", NULL};
run_iptables(IPTABLES_BIN, 1, argv);
```

grandixximo:
> Thanks for fixing cppckeck. Hmm, while looking at this: Why not directly pass argv?
> 
> I don't have time to test this right now, but this compiles. It could be simplified with the matching define.
> 
> ```
> static int run_iptables(const char *bin, int quiet, char **argv) {
> ...
> }
> run_iptables(IPTABLES_BIN, 1, (char *[]){IPTABLES_BIN_NAME, "-n", "-L", "INPUT", NULL});
> //or
> char *argv[] = {IPTABLES_BIN_NAME, "-n", "-L", "INPUT", NULL};
> run_iptables(IPTABLES_BIN, 1, argv);
> ```

Good call, done. `run_cmd` now takes char `*const argv[]` and the IPT/IP6T/NFT macros build it as a compound literal, so no varargs and the varFuncNullUB is gone at the root. Call sites unchanged.

hdiethelm:
Nice! And now you also can not anymore pass an 1 instead of an "1"  to the macos. Before, that probably would have generated a segfault / ub. Now the compiler should catch it.

hdiethelm:
Tested again, works.

PR #4112, opened 2026-06-03
hm2_eth: don't abort board load when SIOCSARP is denied

grandixximo:
Follow-up to #3964, addressing a point @jepler raised in that review.

hm2_eth pins a permanent ARP entry for the board with `SIOCSARP`, which needs `CAP_NET_ADMIN`. On a rootless install without `make setcap` the ioctl fails with EPERM, and the old code treated that as fatal: it returned `-errno`, so the board would not load at all.

That is inconsistent with the rest of the rootless work in #3964, where the firewall setup degrades gracefully (skip with a warning) when the capability is missing. This one ioctl made the whole "run without setcap" mode unusable for ethernet boards.

## What changed

* `SIOCSARP` failure is now best-effort instead of fatal. On failure the driver warns, clears `ATF_PERM`, and continues. The kernel falls back to dynamic ARP, which works; the only cost is an occasional transmit delay when the cache entry expires and is refreshed, as @jepler noted.
* The warning points the user at `sudo make setcap` (or setuid) to restore the pinned entry.
* Clearing `ATF_PERM` keeps the `SIOCDARP` teardown in `close_board()` from trying to remove an entry that was never installed.

Installs that run with setcap or setuid are unaffected: the ioctl succeeds as before and the static entry is pinned.

## Tested

Configured `--with-realtime=uspace` and built; hm2_eth compiles and links clean. Not hardware-tested, I do not have a Mesa ethernet card.

cc @jepler, @hdiethelm

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/hm2_eth.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_eth.c
@@ -759,11 +759,17 @@ static int init_board(hm2_eth_t *board, const char *board_ip) {
         return ret;
     }
 
+    // Pinning the ARP entry needs CAP_NET_ADMIN; rootless without setcap
+    // fails with EPERM.  Best-effort, not fatal: fall back to dynamic ARP
+    // so the board still loads.  Clear ATF_PERM so the SIOCDARP teardown
+    // in close_board() does not try to remove an entry we never set.
     ret = ioctl_siocsarp(board);
     if(ret < 0) {
-        LL_PRINT("ERROR: ioctl SIOCSARP failed: %s\n", strerror(errno));
+        LL_PRINT("WARNING: ioctl SIOCSARP failed: %s; continuing with "
+                 "dynamic ARP.  Install file capabilities (sudo make "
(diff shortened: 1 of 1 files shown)


hdiethelm:
Tested, works!

hdiethelm:
Note: This is an issue created long ago:
https://github.com/LinuxCNC/linuxcnc/commit/d2e4742002ffd76a1837e665ee7e979f5302c147
hm2_eth probably never worked without suid. Due to you set `CAP_NET_ADMIN`, it works also with setcap.
Anyway, it's an improvement, so you can run it also without cap / suid now.

But it might also generate issues if someone does not notice that he has no real time at all. The warning is not well visible, especially if you only start the GUI. But this is a general issue, not connected to hm2_eth. Might be a big red popup in the gui if you don't run real time.

I would merge this and think about a good way to warn a user if he runs linuxcnc without any real time kernel. This can happen by accident if you boot the wrong kernel and then wonder why you mill just crashed.

hdiethelm:
I created a new issue to address the non-realtime trap: https://github.com/LinuxCNC/linuxcnc/issues/4118

PR #4114, opened 2026-06-03
qtvcp: fix startup segfault when no notification daemon is present

grandixximo:
On any system with no desktop notification daemon (headless servers, minimal installs, CI under xvfb), qtvcp segfaults a few seconds after startup. The crash is a use-after-free in `dbus_connection_dispatch()`.

`sys_notify.init()` wires the PyQt dbus mainloop, then the Notifications setup raises `ServiceUnknown` because nothing owns `org.freedesktop.Notifications`. dbus-python has already attached a `QSocketNotifier` to the Qt event loop; the next loop tick dispatches a queued message on the torn-down connection and crashes.

A prior attempt (`d955fc208b`) called `bus.close()` on the failure path, but by then dbus-python had already posted the dispatch to the Qt event loop, so the close came too late. This version probes for the daemon on a throwaway private connection with no mainloop, and only wires the integrated bus once it is confirmed. On the failure path nothing is wired, so nothing dangles.

Backtrace (Ubuntu 24.04, headless, gdb on the core):

```
dbus_connection_dispatch     libdbus-1.so.3
dbus/mainloop/pyqt5.abi3.so
QObject::event
...
QCoreApplication::exec
```

### DIFF
--- a/lib/python/qtvcp/lib/sys_notify.py
+++ b/lib/python/qtvcp/lib/sys_notify.py
@@ -61,39 +61,29 @@ def init(app_name):
     path = "/org/freedesktop/Notifications"
     interface = "org.freedesktop.Notifications"
 
-    mainloop = None
-    bus = None
     try:
-        if DBusQtMainLoop is not None:
-            mainloop = DBusQtMainLoop(set_as_default=True)
+        # Probe on a throwaway no-mainloop connection first. Wiring the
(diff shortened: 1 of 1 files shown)


andypugh:
The error "no notification daemon" is likely to be rather opaque to any user who ever sees it. 
I feel that the message should at least hint at what problem the absence causes. 
I will merge the change as-is, but feel that error messages shouldn't need the context of the PR to uderstand them.

PR #4115, opened 2026-06-03
gmoccapy: exit on SIGTERM/SIGINT regardless of main-loop state

grandixximo:
gmoccapy did not reliably exit on SIGTERM; callers had to escalate to SIGKILL.

#4076 routed the signal through `excepthook` to `Gtk.main_quit()`, but that only quits while `Gtk.main()` is iterating. A signal arriving during startup, or while a nested dialog is up, hits `gtk_main_quit: assertion 'main_loops != NULL'` and gmoccapy keeps running.

This installs an explicit SIGTERM/SIGINT handler that quits the loop if one is active, otherwise exits the process. Surfaced by the ui-smoke quit-path tests (#4054).

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -40,6 +40,7 @@ import hal                  # base hal class to react to hal signals
 from common import hal_glib # needed to make our own hal pins
 import sys                 # handle system calls
 import os                  # needed to get the paths and directories
+import signal              # clean shutdown on SIGTERM/SIGINT
 import atexit              # needed to register child's to be closed on closing the GUI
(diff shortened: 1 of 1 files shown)


grandixximo:
@andypugh need merge this too, then the smoke test phase 2 should pass, I think 🤔

grandixximo:
@BsAtHome when you have a moment, could you take a look at this one? It is the gmoccapy half of the SIGTERM cleanup; #4076 only quits while the Gtk main loop is iterating, so a signal during startup or a nested dialog hits `gtk_main_quit: assertion 'main_loops != NULL'` and the process runs on until SIGKILL. This adds an explicit handler that exits regardless of loop state. It is the last fix blocking the ui-smoke phase-2 quit tests in #4054.

PR #4116, opened 2026-06-03
Fix errors not printed

### LINKED ISSUE #3948: Error messages are not shown if motmod is loaded and not started

hdiethelm:
This fixes https://github.com/LinuxCNC/linuxcnc/issues/3948

The commit https://github.com/LinuxCNC/linuxcnc/pull/4116/changes/c3c06c6de67eebec154679c99f94b85f5acae52e inhibits printing twice by always using old_handler() to print from motion and inhibits console print for motion->milltask messages. TBD if this is fine like this. Otherwise, I would just accept that errors are printed twice. Better than not at all.

Some options I tried and then discarded:

- Add a flag to emcmot_error_t that shows if milltask is up and running and only invoke old_handler() if milltask is not running
  - When milltask is not yet running and then starts, the messages are still printed twice
  - When milltask dies between emcmotErrorPutfv() and the polling of the error buffer, nothing is shown

- Inhinibit print to console by flag in milltask
  - This would need a flag in the dbuf struct which tells the milltask if the message was printed also with old_handler(). This dbuf looks complicated and it is bugprone if I change something here. Something like this would be needed while inhibitConsole is passed trough the dbuf:
```c
  // read the emcmot error
    if (0 != usrmotReadEmcmotError(&inhibitConsole, errorString)) {
        // no error, so ignore
    } else {
        // an error to report
        // disable stdout print due to this error is from motion
        // and already printed to stdout
        if(inhibitConsole){
            RCS_PRINT_DESTINATION_TYPE prev_dest = get_rcs_print_destination();
            set_rcs_print_destination(RCS_PRINT_TO_NULL);
        }
        emcOperatorError("%s", errorString);
        if(inhibitConsole){
            set_rcs_print_destination(prev_dest);
        }
    }
```

### DIFF
--- a/src/emc/motion/motion.c
+++ b/src/emc/motion/motion.c
@@ -196,40 +196,66 @@ void emcmot_config_change(void)
     }
 }
 
+static rtapi_msg_handler_t old_handler = NULL;
+
 void reportError(const char *fmt, ...)
 {
     va_list args;
 
     va_start(args, fmt);
+
+    //Report trough emcmotError() so they are shown
+    //in the gui in the configured language.
     emcmotErrorPutfv(emcmotError, fmt, args);
+
     va_end(args);
+
+
+    //Report trough the old_handler which is typically
+    //the rtapi handler. These messages are shown on the console.
+    if(old_handler){
+        //We must add a \n due to reportError() strings normally
+        //don't have a newline at the end but rtapi_print() strings
+        //need one.
+        char fmt_tmp[EMCMOT_ERROR_LEN];
+        size_t fmt_len = strlen(fmt);
+        //Manually check string length and truncate if it is to long
+        if(fmt_len < EMCMOT_ERROR_LEN-1){
+            memcpy(fmt_tmp, fmt, fmt_len);
+            fmt_tmp[fmt_len+0] = '\n';
+            fmt_tmp[fmt_len+1] = '\0';
+        }else{
+            memcpy(fmt_tmp, fmt, EMCMOT_ERROR_LEN-2);
+            fmt_tmp[EMCMOT_ERROR_LEN-2] = '\n';
+            fmt_tmp[EMCMOT_ERROR_LEN-1] = '\0';
+        }
+
+        va_start(args, fmt);
+        old_handler(RTAPI_MSG_ERR, fmt_tmp, args);
+        va_end(args);
+    }
 }
 
 #ifndef va_copy
 #define va_copy(dest, src) ((dest)=(src))
 #endif
 
-static rtapi_msg_handler_t old_handler = NULL;
 static void emc_message_handler(msg_level_t level, const char *fmt, va_list ap)
 {
     va_list apc;
     // False positive. Cppcheck does not seem to know the properties of va_copy()
     // cppcheck-suppress va_list_usedBeforeStarted
     va_copy(apc, ap);
-    // cppcheck-suppress va_list_usedBeforeStarted
 
     //Report errors trough emcmotError() so they are shown
-    //in the gui and also on the console in the configured language.
+    //in the gui in the configured language.
     if(level == RTAPI_MSG_ERR){
+        // cppcheck-suppress va_list_usedBeforeStarted
         emcmotErrorPutfv(emcmotError, fmt, apc);
     }
 
     //Report everything trough the old_handler which is typically
     //the rtapi handler. These messages are shown on the console.
-    //This has the nasty side effect that error messages are shown twice.
-    //However, there is no good solution. If the milltask fails
-    //or stalls due to any reason, this results in the most important
-    //errors not visible any more if old_handler() is not invoked.
     if(old_handler){
         old_handler(level, fmt, ap);
     }
--- a/src/emc/motion/motion.c
+++ b/src/emc/motion/motion.c
@@ -217,8 +217,23 @@ static void emc_message_handler(msg_level_t level, const char *fmt, va_list ap)
     // cppcheck-suppress va_list_usedBeforeStarted
     va_copy(apc, ap);
     // cppcheck-suppress va_list_usedBeforeStarted
-    if(level == RTAPI_MSG_ERR) emcmotErrorPutfv(emcmotError, fmt, apc);
-    if(old_handler) old_handler(level, fmt, ap);
+
+    //Report errors trough emcmotError() so they are shown
+    //in the gui and also on the console in the configured language.
+    if(level == RTAPI_MSG_ERR){
+        emcmotErrorPutfv(emcmotError, fmt, apc);
+    }
+
+    //Report everything trough the old_handler which is typically
+    //the rtapi handler. These messages are shown on the console.
+    //This has the nasty side effect that error messages are shown twice.
+    //However, there is no good solution. If the milltask fails
+    //or stalls due to any reason, this results in the most important
+    //errors not visible any more if old_handler() is not invoked.
+    if(old_handler){
+        old_handler(level, fmt, ap);
+    }
+
     // cppcheck-suppress va_list_usedBeforeStarted
     va_end(apc);
 }
--- a/src/emc/motion/motion.c
+++ b/src/emc/motion/motion.c
@@ -218,7 +218,7 @@ static void emc_message_handler(msg_level_t level, const char *fmt, va_list ap)
     va_copy(apc, ap);
     // cppcheck-suppress va_list_usedBeforeStarted
     if(level == RTAPI_MSG_ERR) emcmotErrorPutfv(emcmotError, fmt, apc);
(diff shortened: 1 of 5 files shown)


hdiethelm:
Meanwhile, I found a solution that doesn't print twice on the console. But the message in the console is not translated any more. See last commit https://github.com/LinuxCNC/linuxcnc/pull/4116/changes/ddf60cda208b0580c37cb47a7bcbc3a66e7dbbda

@BsAtHome Which version do you think is better?

In the below images, I first trigger an error using `G2 I1000 F2000` which is printed with reportError(). Then I trigger `loadrt pushmsg msgs="e|error|error"` to show an rtapi error.

Before last commit:
grafik

After last commit:
grafik

Edit: There was a bug visible in the image. Fixed in force push.

src/emc/motion/motion.c
```diff
@@ -196,29 +196,57 @@ void emcmot_config_change(void)
...
     // cppcheck-suppress va_list_usedBeforeStarted
-    if(level == RTAPI_MSG_ERR) emcmotErrorPutfv(emcmotError, fmt, apc);
-    else if(old_handler) old_handler(level, fmt, ap);
+
+    //Report errors trough emcmotError() so they are shown
+    //in the gui in the configured language.
+    if(level == RTAPI_MSG_ERR){
+        emcmotErrorPutfv(emcmotError, fmt, apc);
+    }
+
```
BsAtHome, line 256: This change makes the cppcheck-suppress marker fail.
```
Checking emc/motion/control.c ...
emc/motion/motion.c:241:44: error: va_list 'apc' used before va_start() was called. [va_list_usedBeforeStarted]
        emcmotErrorPutfv(emcmotError, fmt, apc);
                                           ^
```

src/emc/motion/motion.c
```diff
@@ -196,29 +196,57 @@ void emcmot_config_change(void)
...
+    //Report trough the old_handler which is typically
+    //the rtapi handler. These messages are shown on the console.
+    if(old_handler){
+        //We must add a \n due to reportError() strings normally
+        //don't have a newline at the end but rtapi_print() strings
+        //need one.
+        char fmt_tmp[EMCMOT_ERROR_LEN];
+        strncpy(fmt_tmp, fmt, EMCMOT_ERROR_LEN-1);
+        strncat(fmt_tmp, "\n", EMCMOT_ERROR_LEN-1);
+        old_handler(RTAPI_MSG_ERR, fmt_tmp, args);
```
BsAtHome, line 234: You need to use the rtapi_str*() functions here because this can be in the kernel.

src/emc/task/taskintf.cc
```diff
@@ -2072,10 +2072,15 @@ int emcMotionUpdate(EMC_MOTION_STAT * stat)
...
     } else {
-	// an error to report
-	emcOperatorError("%s", errorString);
+        // an error to report
+        // disable stdout print due to this error is from motion
+        // and already printed to stdout
+        RCS_PRINT_DESTINATION_TYPE prev_dest = get_rcs_print_destination();
+        set_rcs_print_destination(RCS_PRINT_TO_NULL);
+        emcOperatorError("%s", errorString);
+        set_rcs_print_destination(prev_dest);
```
BsAtHome, line 2084: If you print to null, why is there a need to print? The comment does not make it clear what is happening here.

BsAtHome:
> Meanwhile, I found a solution that doesn't print twice on the console. But the message in the console is not translated any more.

Where in the chain is the translation performed?

src/emc/task/taskintf.cc
hdiethelm: emcOperatorError() does not only print, it sends the message also to the UI.

src/emc/motion/motion.c
hdiethelm: Strangely enough, it compiles. But I found possible issues, reworking it right now.

hdiethelm: The linux kernel supports basic string functions:
https://docs.kernel.org/core-api/kernel-api.html

hdiethelm:
> > Meanwhile, I found a solution that doesn't print twice on the console. But the message in the console is not translated any more.
> 
> Where in the chain is the translation performed?

I have no clue how this works. ~rcs_print() -> rcs_fputs() just does `fputs(_str, stdout);` and then it is translated. Might be some linker trick that overlays an stdout filter?~

src/emc/motion/motion.c
hdiethelm: Fixed

hdiethelm:
> > > Meanwhile, I found a solution that doesn't print twice on the console. But the message in the console is not translated any more.
> > 
> > 
> > Where in the chain is the translation performed?
> 
> I have no clue how this works. rcs_print() -> rcs_fputs() just does `fputs(_str, stdout);` and then it is translated. Might be some linker trick that overlays an stdout filter?

Ah now, that one was well hidden:

Before this line it is untranslated, after it is translated:
https://github.com/LinuxCNC/linuxcnc/blob/3cd702c6906c73300d6ae421bfcc663804d06233/src/emc/motion/usrmotintf.cc#L218
The translation is here:
https://github.com/LinuxCNC/linuxcnc/blob/3cd702c6906c73300d6ae421bfcc663804d06233/src/emc/motion/stashf_wrap.h#L29
And if %s is used, it is here:
https://github.com/LinuxCNC/linuxcnc/blob/3cd702c6906c73300d6ae421bfcc663804d06233/src/emc/motion/stashf_wrap.h#L85

snprintdbuf() is a good name for translating text... ;-)

src/emc/motion/motion.c
hdiethelm: I tested it with RTAI: It works

hdiethelm:
Tests with RTAI show one change:
Motion errors are now not shown anymore in terminal but in the kernel messages like all other real time errors.
The error is still shown in the GUI.

If desired, there is probably a way to enable the terminal errors for RTAI, but they are anyway shown in the GUI.

andypugh:
Is this still draft, or good to go?

hdiethelm:
So, I tested it on the real CNC and it works. If you are OK with:
- Messages on the console are now in english instead of translated
- With RTAI, messages are not any more on the console, only in the kernel messages and the GUI
  - With no motion + milltask running, it was always like this.

Then I would call it ready.

hdiethelm:
^The last force push adds a comment where the translation is happening. So it is not so hard to find any more.

BsAtHome:
Good enough to go.

About the translations... We need to get that fixed another time. There are plenty of paths and layers to examine and streamline, later.

hdiethelm:
> Good enough to go.
> 
> About the translations... We need to get that fixed another time. There are plenty of paths and layers to examine and streamline, later.

This could be done in `uspace_rtapi_main.cc`  at the rt message queue on the reader side. But the rt message queue runs with already expanded printf fmt strings, while the emc fifo has fmt / arguments and does the translation on the fmt string and %s arguments before printing them. So it will be some effort to change the whole way how this is passed trough that fifo. And i think it is not that bad with English on the console and translated in the UI.

So i am with you, not for now.

PR #4119, opened 2026-06-03
CI: Fail on cppcheck.sh error

hdiethelm:
Fix issue in cppcheck.sh and make CI fail on cppcheck error

~Additionally, make shellckeck.sh failing except for files that need yet to be fixed.~

### DIFF
--- a/scripts/cppcheck.sh
+++ b/scripts/cppcheck.sh
@@ -16,6 +16,7 @@ EXHAUSTIVE=$(cppcheck --check-level=exhaustive --version > /dev/null 2>&1 && ech
 CPPCHKOPT=( -j "$nproc" --force "$EXHAUSTIVE" --inline-suppr )
 CPPCHKOPT+=( "--enable=warning,performance,portability" )
 CPPCHKOPT+=( "-I$(realpath "$(dirname "$0")/../include")" )
+CPPCHKOPT+=( --error-exitcode=1 )
 
 if [ -n "$CPPCHECK_OPTS" ]; then
     read -r -a OPTS <<< "$CPPCHECK_OPTS"
@@ -107,5 +108,11 @@ do
     echo "I (4/4): checking $d"
     docheck "$d" || result=1
 done < <(find rtapi/ -type d -not -name "*__pycache__" -print0)
+
+if [ $result -gt 0 ]; then
+    echo "ERROR: Issues found"
+else
+    echo "No issues found"
+fi
 
 exit $result
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,15 +128,29 @@ jobs:
       run: |
         set -x
         sudo apt-get -q update
-        sudo apt-get --yes install cppcheck shellcheck
+        sudo apt-get --yes install cppcheck
     - name: Cppcheck
-      continue-on-error: true
       run: |
         set -x
         scripts/cppcheck.sh
+
+  shellcheck:
+    runs-on: ubuntu-24.04
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v6
+      with:
+        submodules: false
+        fetch-depth: 1
+    - name: Install dependencies
+      run: |
+        set -x
+        sudo apt-get -q update
+        sudo apt-get --yes install shellcheck
     - name: Shellcheck
       continue-on-error: true
       run: |
+        set -x
         scripts/shellcheck.sh
 
   htmldocs:


hdiethelm:
Fails as expected:
https://github.com/LinuxCNC/linuxcnc/actions/runs/26900235217/job/79350299474?pr=4119
Next try.

hdiethelm:
Still failing but the CI doesn't build my code but probably the one it would get after merge. The failing line doesn't look like this in my branch:
https://github.com/hdiethelm/linuxcnc-fork/blob/cppcheck_fail/src/hal/drivers/mesa-hostmot2/hm2_eth.c#L532

@BsAtHome That is what I was talking about in the other thread. Now the CI fails when there are cppcheck issues, so you dont't have to complain any more, the CI does it for you. There was a bug in cppcheck.sh

This can only be merged after the issue in https://github.com/LinuxCNC/linuxcnc/pull/4111 got fixed and merged, otherwise the CI on master will fail.

Let's see if I can easily add a warn only list to shellcheck.sh. You can remove it as soon as all issues are fixed.

hdiethelm:
And the shellcheck warn/error variant.
I just wanted to test if it works. If you think it's useless, I can revert this commit before merge.

This now generates CI errors except for the following files:
```
./scripts/rip-environment.in \
./scripts/realtime.in \
./scripts/linuxcnc_info.in \
./docs/src/asciideps \
./debian/update-dch-from-git \
```
This will help against creping in issues in other shell scripts and as soon as these files are fixed, you can remove the code again or also just remove the files from the list.

hdiethelm:
And after some mocking around with the github ci there is even a nice annotation:
grafik

hdiethelm:
^Rebased back to master

scripts/cppcheck.sh
```diff
@@ -58,11 +58,11 @@ docheck() {
     rv=0
     mapfile -t files < <(find "$1" -maxdepth 1 -name "*.c")
     if [ "${#files[@]}" -gt 0 ]; then
-        cppcheck -I"$1" "${CPPCHKCC[@]}" "${files[@]}" || rv=1
+        cppcheck --error-exitcode=1 -I"$1" "${CPPCHKCC[@]}" "${files[@]}" || rv=1
     fi
     mapfile -t files < <(find "$1" -maxdepth 1 -name "*.cc")
     if [ "${#files[@]}" -gt 0 ]; then
-        cppcheck -I"$1" "${CPPCHKCX[@]}" "${files[@]}" || rv=1
+        cppcheck --error-exitcode=1 -I"$1" "${CPPCHKCX[@]}" "${files[@]}" || rv=1
```
BsAtHome: There is an array CPPCHKOPT (lines 16..18) which you should use and add the option only <i>once</i> instead of changing all the invocations.
```sh
CPPCHKOPT+=( --error-exitcode=1 )
```

scripts/shellcheck.sh
```diff
@@ -53,8 +54,72 @@ else
...
         mapfile -r FILES < <(find "$TOPDIR" -type f -name "*.sh")
     fi
+    #These files don't generate a CI error
+    #ToDo: Fix these and remove from list
+    FILES_WARN_FLTER=(  ./scripts/rip-environment.in \
+                        ./scripts/realtime.in \
+                        ./scripts/linuxcnc_info.in \
+                        ./docs/src/asciideps \
+                        ./debian/update-dch-from-git \
+                     )
```
BsAtHome: I really do not want to make exceptions for files.

<i>"There is nothing more permanent than a temporary fix."</i>

We shouldn't even be contemplating it, so please do not do this. Just _fix_ the files and be done with it.

scripts/shellcheck.sh
```diff
@@ -53,8 +54,72 @@ else
...
+        echo "::error title=ERROR:: shellcheck: issues found"
+        exit 1
+    elif [ $RESULT_WARN -gt 0 ]; then
+        echo "::warning title=WARNING::shellcheck: ignored issues found"
+        exit 0
+    else
+        echo "No issues found"
+        exit 0
+    fi
+fi
```
BsAtHome: You are generating files that do not have a newline at the last line. Please make sure your editor puts a newline at the last line.

scripts/cppcheck.sh
hdiethelm: Thanks for the hint. Fixed.

scripts/shellcheck.sh
hdiethelm: VS Codium, works mostly nice except some quirks.

I should set up pre-commit hooks. Fixed.

scripts/shellcheck.sh
hdiethelm: Feel free to fix them. I am not that good in bash.

Until then:

Did you look at the result when running shellcheck? The output is exactly the same. But with this change:

 \+ The CI fails if any new issues are introduced in files not in this list: This helps alot against new issues
 \+ The way it was, just ignoring the shellcheck in the CI , new issues can come in and no one notices
 \- If you fix a file and forget to remove it from the list, new issues can come in in this file but as a mitigation, I added an annotation if there are issues in ignored files.

Options:
- If you prefer, I can also just make this CI stage failing if there are any issues. That has just the danger that people get used to it failing and don't notice any more
- Leave it as it is
- Remove this commit. Downside: New issues can come in faster than the old ones are corrected.

hdiethelm: Anyway, you where pretty clear. Removed again.

BsAtHome: > Feel free to fix them. I am not that good in bash.

Was afraid you were gonna say that ;-)

Some of the problems are in the bash auto-complete code. Any fix requires some extensive testing. I was already very happy that I was able to reduce the list in the previous fix frenzy. Getting closer every time.

> Did you look at the result when running shellcheck?

Yes,...

I'd like to start from a clean slate. Everything in a good state and then we enforce it. It just take a while to get there.

However, you split (are splitting) cppcheck and shellcheck in CI, which is probably a good thing. Then we can enforce cppcheck before we start enforcing shellcheck.

hdiethelm: Yes, the last time i looked, the issue list was way longer... :-)

I would still prefer to also partially enforce shellcheck. This way, what happened with rootless, is way less likely.

Before rootless, cppcheck was clean. Now not any more, so we have to wait merging this pr until it is clean again and hope no other issue gets in between.

Enforcing with a warn only list helps against this, even if it feels so not good.

But we can also leave it as it is now and you just have to remove the `continue-on-error: true` on the last shell fix mr.

hdiethelm:
- Rebased to master
- Split commits better (no change in code)

Now cppcheck passes due to https://github.com/LinuxCNC/linuxcnc/pull/4111 got merged.

Ready to merge from my side.

BsAtHome:
@hdiethelm Slightly off-topic... You have been digging into CI a lot lately and may have come across this in some way.

The new docs build places everything in the `docs/build/html` folder and the homepage server is not updating anymore. The current htmldocs CI target is not making any artifacts and we'd like it to so we can fetch it.

Firstly, how do you create an artifact of the complete `docs/build/html` folder when the CI's make is done?
Secondly, how can we get to the latest artifact from this CI build from the outside? Is there an (external) API we can use to find the latest build artifact and fetch that? I see only build numbers in the artifact links.

hdiethelm:
Sure, I will look into it.

In gitlab, for such jobs, you have generic packages that you can upload in CI and download from everywhere with a persistent link, so you can automatically download it from the homepage server. However, github might be different, I have to look into it.

Artifacts are linked to a CI job, so you never have a persistent link.

An alternative would be to upload it from CI directly to an endpoint like the homepage server.

I will check the options and create a PR.

BsAtHome:
> Sure, I will look into it.

Fantastic, thanks!


> An alternative would be to upload it from CI directly to an endpoint like the homepage server.

We discussed it and the problem is that you cannot keep secrets secret on github when github pushes. Such API also opens the homepage server to a DDOS surface. Better to have the homepage server pull the data from github.

hdiethelm:
So, prototype is running. Looks like github only supports docker. But people figured out how to publish any data inside a docker image: https://oras.land/docs/

Here is my prototype: https://github.com/hdiethelm/github_actions_experiments/blob/main/.github/workflows/push_package.yml
Should be easy to integrate into the linuxcnc ci.
Will appear here: 
grafik

And can be downloaded with:
`oras pull ghcr.io/hdiethelm/linuxcnc/doc:main`
where main is the tag. I set it to the branch name, so in linuxcnc it would be 2.9 / master. Any container tag can be chosen and also multiple tags for the same container, for example a git tag name for a release or just the SHA for all, so old versions can be restored based on the git hash.

Only downside: You need oras installed on the machine that pulls the package.

There are also risk involved for the web server when automatically downloading a dok package built in a CI. A bad actor could for example inject PHP code that compromises the server. I will contact you by mail about details tomorrow, should not be discussed in public.

PR #4122, opened 2026-06-04
2.9 gremlin: Initialize preview interpreter with active modal gcodes

### LINKED ISSUE #4121: bug(gremlin): incorrect gcode checking due to wrong initialization
As reported on the forum:
https://forum.linuxcnc.org/38-general-linuxcnc-questions/58828-potential-bug-in-version-2-9-8#346874

This is related to this earlier bug fix:  https://github.com/LinuxCNC/linuxcnc/commit/f6b704e8bed3ba9844ad08507319d752ee827e09
which has already been trying to address issues with incorrect gcode preview when loading gcode that does not contain complete preambles.

Cause:
While the bugfix above works for the actual tool path preview it does not work for the error checking done in the interpreter run on load as the interpreter uses it's internal defaults.

Fix:
Going back to initializing the interpreter with the RS274NGC_STARTUP_CODE is not the right fix as it will reintroduce issue #2753. Instead the interpreter run on load should be initialized with the currently active modal gcodes. This should also address other potential bugs when loading gcode with incomplete preambles.

Affects:
Any Gui using 'gremlin.py' for gcode preview. (eg gmoccapy)

Sigma1912:
fixes  #4121 and improves preview for gcode with incomplete preambles.

No regression on https://github.com/LinuxCNC/linuxcnc/issues/2753

### DIFF
--- a/src/emc/usr_intf/gremlin/gremlin.py
+++ b/src/emc/usr_intf/gremlin/gremlin.py
@@ -328,6 +328,13 @@ class Gremlin(Gtk.DrawingArea,rs274.glcanon.GlCanonDraw,glnav.GlNavBase):
 
             unitcode = "G%d" % (20 + (s.linear_units == 1))
             initcode = ""
+            # To guard against incorrect preview with incomplete preambles in the gcode program
+            # we initialize with active gcodes for modal groups 2,3,4,5,6,7,8,9,10,11,12,13,14,15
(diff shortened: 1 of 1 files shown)


grandixximo:
Looks good to me.

grandixximo:
@andypugh do we want this in?

Issue #3855, opened 2026-03-10
Incorrect arc preview (in gmoccapy)


finngineering:
The circular arc preview in (at least) gmoccapy is sometimes wrong. Possibly because it doesn't take the current machine position into consideration.

This is with LinuxCNC 2.9.8 from the ISO and stock gmoccapy 3.4.9.

I start the gmoccapy simulation with "gmoccapy" configuration, home the machine (which moves it to X10, Y10, Z-10), load and run this g-code:
```
F1000
G2 I20 J0 P1
M2
```
You can see the difference between the preview (white) and actual motion (magenta) in the screenshot below. The preview circle has incorrect coordinates in all three axes.
<img width="990" height="784" alt="Image" src="https://github.com/user-attachments/assets/001f8968-effd-4a30-a1c3-e563ea5d496b" />

I made a forum post about this a while back as well:
[https://forum.linuxcnc.org/20-g-code/58285-strange-preview-for-circular-arc-g2-g3](https://forum.linuxcnc.org/20-g-code/58285-strange-preview-for-circular-arc-g2-g3)

andypugh:
That is arguably bad G-code. 
A full-circle G2 without a positioning move prior to it will result in undefined motion. 
(A part-circle, with an end point, will fail most of the time in this situation as the starting point, end point and centre will only sometimes define a valid arc)

finngineering:
Is there a reason why a full-circle G2 without positioning move would result in undefined motion? On a high level, I don't see why that would be the case. The machine is in a known position and as far as I' aware, the starting position and the arc G-code I provided should fully define the circle. How about if I program `G2 I20 J0 P1` in the MDI? Would that give undefined motion as well?

We could also turn it the other way around. If the motion is undefined in this situation, then why is a preview shown and why can I run the code (the latter generates the expected circle)?

I did also previously search the documentation for any clues as to what I might be doing wrong, but did not really find anything.

hansu:
I think the problem here is that the preview does not take the current position into account. So the first (linear) move is never displayed. Which also makes sense - the preview needs to be reloaded as soon as you jog the machine - otherwise it would display a wrong toolpath.

andypugh:
What I am saying is that it's a bad G-code program. Depending on where the tool is positioned when you run the code you get a different diameter circle. (It would be a fixed diameter if G91.1 is active, but in a random place, but the G-code isn't setting G90.1 / G91.1 either)

finngineering:
[...]

PR #4123, opened 2026-06-04
master gremlin: Initialize preview interpreter with active modal gcodes

### LINKED ISSUE #4121: bug(gremlin): incorrect gcode checking due to wrong initialization

Sigma1912:
fixes https://github.com/LinuxCNC/linuxcnc/issues/4121 and improves preview for gcode with incomplete preambles.

- Creates a helper file in the Gremlin folder that is copied to `lib/python` on make.
-  Adds a test creating the initcode according to the current modal gcodes and checking interpreter output on loading a gcode file lacking a preamble.
- `gremlin.py` and `test-ui.py' import and use the inicodes helper function
- `qt5-graphics.py` used by QtDragon can be easily modified to use the same helper function in a separate PR.

No regression on https://github.com/LinuxCNC/linuxcnc/issues/2753
No regression on https://github.com/LinuxCNC/linuxcnc/issues/3855

Affects gmoccapy and any other gui using gremlin.


src/emc/usr_intf/gremlin/gremlin.py
```diff
@@ -336,12 +337,20 @@ def load(self,filename = None):
...
+            # sorting is not necessary but helps in debugging
+            active_gcodes = s.gcodes
+            for i in (3,6,14,7,5,4,9,12,10,16,8,11,13,15):
+                if active_gcodes[i] > -1:
+                    initcode = initcode + 'G' + str(active_gcodes[i]/10) + ' '
             result, seq = self.load_preview(filename, canon, unitcode, initcode)
             if result > gcode.MIN_ERROR:
                 self.report_gcode_error(result, seq, filename)
             self.calculate_gcode_properties(canon)
+
```
grandixximo, line 333: Don't add white lines, it muddies the blame in git

grandixximo (changes requested):

src/emc/usr_intf/gremlin/gremlin.py
Sigma1912: removed

grandixximo:
`s.gcodes` is scaled by 10 (G21=210, G38.2=382). In Py3 `active_gcodes[i]/10` is float division: `170 -> 'G17.0'` (invalid), `382 -> 'G38.2'` (ok). Whole codes get a stray `.0` and won't parse.

`hal_glib.py` does it right:
```python
code = "G%d" % (i//10) if i%10==0 else "G%d.%d" % (i//10, i%10)
```
Suggest same split, and a test that `initcode` round-trips through the interpreter.

c-morley:
As qr_graphics uses very similar cide, can you add to or create a pull request for it too? I'm away working and dont want to get them out of step.

Sigma1912:
@grandixximo 
> Whole codes get a stray .0 and won't parse.

It's an easy enough change but  having a trailing `.0` parses fine for me. 
Examples: 
- `G7.0` does the same as `G7`
- `G0.0 X1` does the same as `G0 X1`

I'll have a look at adding a test. Do you have any pointers on how I might test the preview?


@c-morley 
Sure, I'll have a look as soon as this is merged.

c-morley:
Your the best thanks ;)

grandixximo:
You're right, I checked `read_g` in `interp_read.cc` ( `value_read*10`, `floor`, fraction `<0.001` accepted ), so `G17.0` rounds to 170 and parses identically to `G17`. My bad, it's purely cosmetic. Only reason to split would be cleaner debug output, fully optional.

For testing the preview headless: `gcode.parse(filename, canon, *initcodes)` is the core call ( see `GLCanon.load_preview` in `lib/python/rs274/glcanon.py` ). You can drive it without a GUI using a canon subclassed from `Translated` in `lib/python/rs274/interpret.py` that just records the emitted `straight_feed` / `arc_feed` segments. Build the same `unitcode`/`initcode` strings your patch generates, feed a g-code file with an incomplete preamble ( e.g. a `G2`/`G3` arc relying on a modal plane, or motion relying on modal `G90`/`G91` ), and assert the recorded segments match the expected path. That can drop straight into `tests/` as a non-GUI sim test so it guards against regressions on #4121, #2753 and #3855.

Sigma1912:
@grandixximo
Trying to run `gcode.parse(filename, canon, *initcodes)` using the `PrintCanon`class but I'm struggling.
 
Running my rump python code (see below) I get this error message :
```
Starting 
File exists
  File "/home/user/git/mylinuxcnc-master/tests/interp_initcode/./test-ui.py", line 50, in <module>
    result, seq = gcode.parse(filename, PrintCanon, unitcode, initcode)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: interp_error > 0 but no Python exception set
```

After some print debugging I found that apparently ìnterp_error  is `18` after this line:
https://github.com/LinuxCNC/linuxcnc/blob/106519fb3e01b3401f3592f9d78bdd9030e2e5ba/src/emc/rs274ngc/gcodemodule.cc#L894
Any idea what that might mean? 


This is the current rump code:
```
#!/usr/bin/env python3
import gcode
import sys
import os

class PrintCanon:
    def set_g5x_offset(self, *args):
        print("set_g5x_offset", args)

    def set_g92_offset(self, *args):
        print("set_g92_offset", args)

    def next_line(self, state):
        print("next_line", state.sequence_number)
        self.state = state

    def set_plane(self, plane):
        print("set plane", plane)

    def set_feed_rate(self, arg):
        print("set feed rate", arg)

    def comment(self, arg):
        print("#", arg)

    def straight_traverse(self, *args):
        print("straight_traverse %.4g %.4g %.4g  %.4g %.4g %.4g" % args)

    def straight_feed(self, *args):
        print("straight_feed %.4g %.4g %.4g  %.4g %.4g %.4g" % args)

    def dwell(self, arg):
        if arg < .1:
            print("dwell %f ms" % (1000 * arg))
        else:
            print("dwell %f seconds" % arg)

    def arc_feed(self, *args):
        print("arc_feed %.4g %.4g  %.4g %.4g %.4g  %.4g  %.4g %.4g %.4g" % args)


sys.stderr.write("Starting \n")
filename = "test.ngc"
unitcode = "G21"
initcode = "G7"
if os.path.exists(filename):
    print("File exists")
else:
    print("File does not exist")
result, seq = gcode.parse(filename, PrintCanon, unitcode, initcode)

fail = True
if fail:
    sys.exit(1)

sys.stderr.write('Success\n')
sys.exit(0)
```

grandixximo:
error 18 isn't a single g-code error, it's the `interp_error` counter in `gcodemodule.cc`. It gets bumped every time the interpreter asks the canon a question and the canon has no method to answer ( see the `interp_error++` lines in `gcodemodule.cc`: `get_block_delete`, `get_tool`, `get_axis_mask`, `get_external_angular_units`, `get_external_length_units` ). Your bare `PrintCanon` ( copied from `lib/python/rs274/interpret.py` ) implements only the motion-output callbacks, not these query callbacks, so it racks up errors during setup. Those callbacks normally live in `StatMixin`, which `PrintCanon` doesn't inherit.

For a headless test that needs no running LinuxCNC, just stub them with constants:

```python
class PreviewCanon(PrintCanon):
    def get_external_length_units(self):  return 1.0   # mm; use 1/25.4 for inch
    def get_external_angular_units(self): return 1.0
    def get_axis_mask(self):              return 0b111111111  # all 9 axes
    def get_block_delete(self):           return 0
    def change_tool(self, pocket):        pass
    def get_tool(self, pocket):
        # toolno, x,y,z,a,b,c,u,v,w, diam, frontangle, backangle, orientation
        return (pocket, 0.,0.,0.,0.,0.,0.,0.,0.,0., 0., 0., 0., 0)
    def next_line(self, state):
        self.state = state

canon = PreviewCanon()
result, seq = gcode.parse(filename, canon, "G21", "G7")
if result > gcode.MIN_ERROR:
    print(gcode.strerror(result))
```

If you'd rather mirror what gremlin actually does ( real tool table, real units ), subclass `StatMixin` with a live `linuxcnc.stat()` instead of the stubs, but for a sim test in `tests/` the constants above keep it self-contained.

Sigma1912:
Thanks for the pointers.

I've got something that seems to be working now.

Sigma1912:
@grandixximo 
Added the test. Is that what you had in mind?

tests/interp_initcode/test-ui.py
```diff
@@ -0,0 +1,154 @@
...
+    for i in (3,6,14,7,5,4,9,12,10,16,8,11,13,15):
+        if active_gcodes[i] > -1:
+            initcode = initcode + 'G' + str(active_gcodes[i]/10) + ' '
+    #print("unitcode, initcode ", unitcode, initcode)
+    return unitcode, initcode
+
+
+# route output to log file now
+log_file = open("temp_log","w")
+sys.stdout = log_file
```
grandixximo: see test.sh:8

grandixximo:
Add a .`gitignore` for `result`, `stderr`, `test.var*`, `temp_log` like `tests/startup-state/.gitignore`.

Sigma1912:
> Add a .gitignore for result, stderr, test.var*, temp_log like tests/startup-state/.gitignore.

Done

grandixximo:
The `.gitignore` went into the wrong dir: it landed in `tests/abort/feed-rate/.gitignore` and overwrote that file's `sim.var*` line. Please revert `feed-rate` to `sim.var*` and add a new `tests/interp_initcode/.gitignore`:
```
result
stderr
test.var*
temp_log
```

Two other things still open:

**1. Duplicated init-code logic.** `create_unitcode_and_initcode()` in `test-ui.py` is a verbatim copy of the loop in `gremlin.py`'s `load()`, so the test guards its own copy, not the real code. Factor it into one helper both import.

**2. `finally: sys.exit(0)` defeats the error path.** It overrides the `SystemExit` from `except`, so the script always exits 0 and the `sys.exit(1)` calls are dead. Also `print(e)` goes to `temp_log`, not stderr. Drop the `sys.exit(0)` from `finally` and send the error to stderr.

Sigma1912:
Thanks for your valuable input (and patience).

> The .gitignore went into the wrong dir

Apologies, that was a bit sloppy on my part.

> 1. Duplicated init-code logic. create_unitcode_and_initcode() in test-ui.py is a verbatim copy of the loop in gremlin.py's load(), so the test guards its own copy, not the real code. Factor it into one helper both import.

- Created a helper file in the Gremlin folder that is copied to `lib/python` on make.
- `gremlin.py` and `test-ui.py' import and use the inicodes helper function
- `qt5-graphics.py` used by QtDragon can be easily modified to use the same helper function in a separate PR.

> 2. finally: sys.exit(0) defeats the error path. It overrides the SystemExit from except, so the script always exits 0 and the sys.exit(1) calls are dead. Also print(e) goes to temp_log, not stderr. Drop the sys.exit(0) from finally and send the error to stderr.

Done

Tested:
- Gremlin preview using Gmoccapy
- New  tests/interp_initcode/

grandixximo:
All three points look good now, thanks. One tiny thing on the `Submakefile`: the added rule targets `emc/usr_intf/preview_helpers/%`, but the file lives in `emc/usr_intf/gremlin/`, so that rule never matches. The existing `../lib/python/%: emc/usr_intf/gremlin/%` rule right above it already copies `preview_helpers.py`. The new rule can be dropped ( and there's a stray trailing space after `preview_helpers.py` in the `PYTARGETS +=` line ).

Sigma1912:
> The new rule can be dropped ( and there's a stray trailing space after preview_helpers.py in the PYTARGETS += line ).

Done. Thanks again for the review!

grandixximo:
Great, all looks good now. Thanks for the quick turnaround and the thorough testing. LGTM.

PR #4125, opened 2026-06-04
gmoccapy: exit on SIGTERM received during startup

grandixximo:
Follow-up to #4115. That handler is installed just before Gtk.main(), so a SIGTERM arriving earlier, during construction, is still routed through the excepthook as a KeyboardInterrupt. With no main loop running Gtk.main_quit() asserts, the interrupt is swallowed at the GTK C boundary, and gmoccapy keeps running until the caller escalates to SIGKILL.

This makes the excepthook force the process exit when no main loop is active, so a termination signal during startup takes effect. Verified green against the ui-smoke quit tests in #4054 (289/289).


grandixximo:
I tested this with smoke UI, that should pass once this merges, I did some updates in the smoke UI for the quit path, so needed a few more fixes...

PR #4126, opened 2026-06-04
gmoccapy: add flag for smoke tests to disable startup message

### LINKED ISSUE #4072: gmoccapy: modal startup "Important change(s)" dialog blocks headless / non-interactive startup
gmoccapy's `_startup_message()` (`src/emc/usr_intf/gmoccapy/gmoccapy.py:544`)
pops a MODAL "Important change(s)" release-notes dialog during `__init__` via
`dialogs.show_user_message()` -> `Gtk.Dialog.run()`, whenever the preference
`hide_startup_messsage` is less than the number of release-note messages.

The preference file is created at runtime with `hide_startup_messsage = 0`,
so on a fresh profile the dialog always shows. Because `Gtk.Dialog.run()` spins
its own nested main loop, the gmoccapy constructor never returns until a user
clicks the dialog. Under a headless / non-interactive launch (xvfb, CI, the
ui-smoke tests in PR #4054) nobody dismisses it, so gmoccapy hangs at startup.

Reproduce:
- Remove or reset the config's `gmoccapy.pref` (so `hide_startup_messsage = 0`).
- Launch the gmoccapy sim under xvfb.
- py-spy on the gmoccapy process shows it stuck in:
    run (gi/overrides/Gtk.py)
    show_user_message (gmoccapy/dialogs.py)
    _startup_message (gmoccapy.py:560)
    __init__ (gmoccapy.py:506)

Impact:
- Any automated / headless launch of gmoccapy with a fresh profile blocks.
- The ui-smoke gmoccapy test must pre-seed or auto-dismiss this dialog.

Possible fixes (for discussion):
- Skip the startup dialog when running non-interactively (e.g. detect no
  interactive display, an env var, or a CLI flag).
- Make the dialog non-modal / auto-timeout.
- Show the notice via the in-app notification area instead of a blocking dialog.

hansu:
Add flag to disable startup message es requested in #4072 for running the smoke tests

More flags can be easily added when needed.

Resolves #4072

### DIFF
--- a/docs/src/gui/gmoccapy.adoc
+++ b/docs/src/gui/gmoccapy.adoc
@@ -58,12 +58,18 @@ You will get a similar screen to the following (the design may vary depending on
 
 image:images/gmoccapy_3_axis_mid.png[align="left",link="images/gmoccapy_3_axis.png"]
 
+[[gmoccapy:basic_configuration]]
 == Basic Configuration
 
 GMOCCAPY 3 supports the following command line options:
 
  * '-user_mode': If set, the setup button will be disabled, so normal machine operators are not able to edit the settings of the machine.
  * '-logo <path to logo file>': If given, the logo will hide the jog button tab in manual mode, this is only useful for machines with hardware buttons for jogging and increment selection.
+ * '-smoketest=nowhatsnew': Disable the "Important changes" notification.
+* '-d': Set log level to DEBUG
+* '-i': Set log level to INFO
+* '-v': Set log level to VERBOSE
+* '-q': Set log level to ERROR
 
 There is really not to much to configure just to run GMOCCAPY,
 but there are some points you should take care off if you want to use all the features of the GUI.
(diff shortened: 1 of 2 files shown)


BsAtHome:
Shouldn't the option be added to the smoke test too? Now you define it, then better use it. But that brings us to another issue.

@grandixximo you use the default config/sim/.../...ini files. But you can't add an option to any of the GUIs.

There seems to be a slight problem here now because we want to use the "standard" ini files in the smoke test but add an exception when running under smoke testing. In this case, it might be better to use an environment variable instead so that the standard ini files work.

grandixximo:
For the smoke tests as they stand in #4136, this is already covered without a switch. I mirror each sim config to a writable tmp copy (gmoccapy and qtdragon need that anyway, since they write next to the config and CI mounts read-only) and seed `hide_startup_messsage` on the mirror, so the "Important change(s)" modal stays hidden with no change to the standard ini. The quit tests use SIGTERM, not an interactive quit, so no quit dialog is raised there either.

So nothing in the current tests needs `-smoketest=` once #4136 is in. A general suppression switch only becomes relevant if we later drive interactive UI, and how far we take the UI tests is still open, so I would rather settle that scope first before deciding whether this lands as a flag, an env var, or stays pref-based. Happy to keep the PR around in the meantime.

PR #4131, opened 2026-06-05
touchy: fit the window to the monitor

grandixximo:
Touchy's window could grow past the screen edge, worse after loading a large tool table, putting controls out of reach. The root cause is structural: a GtkNotebook sizes to its largest page, so the small visible page was forced as wide as the hidden Preferences page.

What this does:

- Wraps each notebook page in a scroller, so the notebook sizes to the current page; oversized pages (Preferences, a long tool table) scroll instead of growing the window. This cuts the minimum width from 1346 to 1094 px and keeps the height in check.
- Bounds the whole window to the monitor work area, so it can never exceed the screen.
- Hides the handwheel column on the Preferences tab (you are not jogging there) so the settings page gets the full width.
- If it still does not fit a small screen, offers once to shrink the fonts: `Shrink to fit and save` / `Not now` / `Never ask again`. It never asks when it already fits and never repeats once it fits; `Not now` is remembered per screen size, and a Preferences checkbox re-enables the offer after `Never ask again`.

Tested with the ui-smoke touchy tests and by hand at 1920x1080, 1366x768, 1280x1024, 1280x720, 1024x768 and 800x600 (all tabs, plus reload tool table).

### DIFF
--- a/docs/src/gui/touchy.adoc
+++ b/docs/src/gui/touchy.adoc
@@ -96,6 +96,27 @@ cut off, reduce the DPI setting.
 
 All other font sizes can be changed on the Preferences tab.
 
+=== Fitting the screen
+
+Touchy bounds its window to the monitor and scrolls any tab whose content
+is larger than the screen, so the window never grows past the display
+edge (for example after loading a large tool table). The handwheel column
+is hidden on the Preferences tab so the settings have the full width.
+
+If the interface is still too large for a small screen, Touchy offers,
+once, to shrink the display fonts to fit:
+
+* *Shrink to fit and save* scales the fonts to the largest size that fits
+  and saves them as your preference (the font selectors update to match).
+* *Not now* leaves the fonts unchanged and does not ask again for this
+  screen size; it is offered again only if you move to a smaller screen.
+* *Never ask again* disables the offer on every screen.
+
+To turn the offer back on, tick *Offer to shrink fonts to fit the screen*
+in Preferences / Display Options, or set `fit_fonts = ask` in
+`~/.touchy_preferences` (the *Never ask again* choice stores
+`fit_fonts = never`).
+
 === Macros
 
 Touchy can invoke O-word macros using the MDI interface.  To configure
--- a/src/emc/usr_intf/touchy/listing.py
+++ b/src/emc/usr_intf/touchy/listing.py
@@ -103,7 +103,7 @@ class listing:
                     break
         self.show_line(self.start_line)
         
-    def on_select(self, b):
+    def on_select(self, b, *args):
         pass
 
     def clear_startline(self):
--- a/src/emc/usr_intf/touchy/touchy.glade
+++ b/src/emc/usr_intf/touchy/touchy.glade
@@ -3173,7 +3173,7 @@ F1 S1</property>
                                   <object class="GtkTable" id="table3">
                                     <property name="visible">True</property>
                                     <property name="can-focus">False</property>
-                                    <property name="n-rows">5</property>
+                                    <property name="n-rows">6</property>
                                     <property name="n-columns">3</property>
                                     <child>
                                       <placeholder/>
@@ -3400,6 +3400,27 @@ F1 S1</property>
                                         <property name="x-options">GTK_FILL</property>
                                         <property name="y-options"/>
                                         <property name="x-padding">10</property>
+                                      </packing>
+                                    </child>
+                                    <child>
+                                      <object class="GtkCheckButton" id="fitfontscheck">
+                                        <property name="label" translatable="yes">Offer to shrink fonts to fit the screen</property>
+                                        <property name="visible">True</property>
+                                        <property name="can-focus">False</property>
+                                        <property name="receives-default">False</property>
+                                        <property name="focus-on-click">False</property>
+                                        <property name="draw-indicator">True</property>
+                                        <signal name="toggled" handler="on_fitfontscheck_toggled" swapped="no"/>
+                                      </object>
+                                      <packing>
+                                        <property name="left-attach">0</property>
+                                        <property name="right-attach">3</property>
+                                        <property name="top-attach">5</property>
+                                        <property name="bottom-attach">6</property>
+                                        <property name="x-options">GTK_FILL</property>
+                                        <property name="y-options"/>
+                                        <property name="x-padding">10</property>
+                                        <property name="y-padding">6</property>
                                       </packing>
                                     </child>
                                   </object>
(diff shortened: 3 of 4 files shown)


grandixximo:
Fonts render too small, will push update soon

grandixximo:
I reworked this from the earlier scroll/scale version. The key change is wrapping each notebook page so the window sizes to the visible page instead of the widest hidden one (Preferences); that also fixes the window growing after `reload tool table` that Andy reported, since the Status page now scrolls. Font shrinking is opt-in via a dialog, with a Preferences checkbox to re-enable it.

I am open to changes on any of this, the dialog wording, whether to hide the handwheel on Preferences, the readability floor, etc. I have screenshots at the resolutions above if useful.

grandixximo:
rv_800x600_0_Startup
it should fit now on 800x600 and I can add a test to check that it fits to avoid regressions in the phase 3 of smoke UI

grandixximo:
overlay_1024
Changing to hovering infobar to avoid slapping anyone on the face @BsAtHome

hansu:
Looks good :+1: , only the minimize/maximize icon in the title bar disappeared so you cannot use it windowed anymore. 


With the replacement of the deprecated table in the status (see https://github.com/LinuxCNC/linuxcnc/pull/4135), the scrollbar isn't really needed, but let's keep it just in case :)

grandixximo:
Marking this a draft for now: per the plan in #4135 it should land after the GTK3 conversion (#4135), so I will rebase this onto #4135 once that merges. It applies to master and works standalone in the meantime.

On the screenshots where the window has no title bar: that is only because the tests run under Xvfb with no window manager. On a normal desktop touchy keeps its title bar with minimize, maximize and close as usual, and close still fires the existing quit handler. This PR does not change that (and it now sets no max-size hint, so the maximize button is never affected either).

hansu:
> only the minimize/maximize icon in the title bar disappeared so you cannot use it windowed anymore.

I was talking about this:

grafik

It is gone with your commit:
grafik

grandixximo:
Try with the last commit

hansu:
The icon is there but I cannot resize the window smaller than the size it got when started maximized (=screen size) even though I get the handle.

https://github.com/user-attachments/assets/51a2c4ef-e0ec-4507-909d-c54f26319dbb

grandixximo:
Fixed in the latest commit. The scroller had its min-content pinned to the work area, which forced the window minimum to screen size. Now only the max is bounded, so it resizes smaller freely. Also added drag-to-pan on each page (finger or mouse) since touchy is a touchscreen UI.

hansu:
Rescaling works now :+1: 

I can even scale it smaller than the space the widgets need. Not sure if that is good or bad :D

https://github.com/user-attachments/assets/685a6101-be46-4718-bcfb-f26af8cec818

grandixximo:
Well, following the mailing list, the window forcing its size would have the UI pop out of the monitor, and the user had to fight it, this PR solves the issue at the core, at most there is some stuff you don't see, and you have to scroll to get to it, and you can make font a bit smaller to fit, automatic resizing overlay helper also finds the fitting font size for you. No more fighting to fit, the window can be as small as the user needs it to be. I did not force a minimum size as system tray bars and window top bar can remove some unspecified amount of space system to system, we could force a minimum to view everything without scrolling, but I'm not sure that's necessarily what the user is looking for, I think flexibility solves the issue completely at the source, IMO there is no real user downside with being able to make the window extremely small.

andypugh:
I decided that I wanted this for 2.9, so applied (largely manually) the patch to my 2.9 milling machine. 

It is definitely better, and does work as required in full-screen mode.

However I have a touchscreen and have no way to exit full-screen mode, so don't tend to use it.

But, in maximised window mode, the status bar gets cut off. The UI is entirely usable, but you can't see the error messages. 

(Touchy could maybe do with a "full screen" button, and possibly even a "exit and power-down" button.)

hansu:
>But, in maximised window mode, the status bar gets cut off. The UI is entirely usable, but you can't see the error messages.

But was this different before this PR or in 2.9 before you updated your Debian?

andypugh:
> But was this different before this PR or in 2.9 before you updated your Debian?

It was hard to tell, I couldn't even see the tab bar...

It's _much_ better with my hacked version of this PR, it's just not _quite_ there. 

TBF The whole error reporting thing in Touchy is terrible, it only shows (most of) a single line of text. Multi-line errors lose all but the last line. 

(Bit for all its flaws, I like it for enforcing "physical buttons to cause machine motion"[1]

[1] Except for homing, I would like to see "press cycle start to confirm" there, ideally. But given that the machine isn't running at that point, and flying chips are not pinging off the screen, it's probably low risk)

andypugh:
@hansu @grandixximo  I ran this for the first time actually making parts and found the touch-screen response to be very slow (barely usable). 
I am wondering if this change (in 2,9) makes the issue addressed here even worse? 
https://github.com/LinuxCNC/linuxcnc/commit/34fa4e1ad795489e4e286eaa7f21883b5259eafa

grandixximo:
Is CPU at 100% constantly on your system? When you run touchy?

hansu:
Have you tried with and without https://github.com/LinuxCNC/linuxcnc/commit/34fa4e1ad795489e4e286eaa7f21883b5259eafa?

andypugh:
Yes, I just tried it now, responsiveness is much improved with 34fa4e1.

In fact I will cherry-pick it into 2.9 as I think there is likely to be a 2.9 release soon to address bugs raised by Debian.

andypugh:
There is some slightly odd behaviour, having used it more. 
When entering numbers in MDI mode (the touchscreen-friendly MDI is the main reason I stick with Touchy), and with the font and screen sizes I have, when entering long numbers the MDI command window resizes the number pad pane. 

Probably more of a functional issue is that running a program in Auto mode, when executing a long line of G0code the right-side (up / down / run from here) control column for the listing disappears, then re-appears for short lines. 

Both these things may disappear with smaller fonts, I will experiment.

grandixximo:
I'll look into it, make sure nothing stretches as you type or load gcodes in, have the fields be draggable, so when content doe not fit, you can look around

andypugh:
There are probably more important areas in which to invest effort. I think that Touchy is probably a fairly little-used interface, and if anyone should be working on it it should probably be me...)

grandixximo:
It can be a team effort, I've tried to run touchy before, I do like the touchscreen concept for some use cases, I'm waiting on some TP updates from YangYang, it's not a complete waste, touchscreens are a sell point IMO, the UI should be less frustrating to use and setup on different res

Issue #4129, opened 2026-06-04
Python hal.is_rt / hal.is_sim is broken


hdiethelm:
https://github.com/LinuxCNC/linuxcnc/blob/888cb941fbf23ee32fc8e15ca180997d99a39117/src/hal/halmodule.cc#L2382
^ this is broken: The value `is_rt` is always false due to `rtapi_is_realtime()` checks if the actual process is real time capable. But the python runtime is hopefully never setuid... ;-)

Fixing that one is a bit difficult, we would have to call rtapi_app to get the true `rtapi_is_realtime()` value of the rtapi_app process.

Alternative: Just remove is_rt / is_sim. I would have a `halcmd getrt` POC ready that does exactly ^. But it needs some finalizing / doc. As much as I know, there is only the halcmd way to execute actions inside the rtapi_app process.

Looking at the initial commit, this worked more or less:
https://github.com/LinuxCNC/linuxcnc/commit/3eac731ce722b9cca2e7ca4f0fbbc70f272a7c6c
~until a setuid check was added in detect_realtime(), which broke it for user components:~
https://github.com/LinuxCNC/linuxcnc/commit/2faa90974373bd1869342fcaa891dd7b220698cc

It was broken in:
https://github.com/LinuxCNC/linuxcnc/pull/3964
to fix the inconsistency of `rtapi_is_realtime()`: https://github.com/LinuxCNC/linuxcnc/issues/3928

hdiethelm:
I have to track back.
https://github.com/LinuxCNC/linuxcnc/commit/2faa90974373bd1869342fcaa891dd7b220698cc
Should still work.

hdiethelm:
Work in progress in https://github.com/LinuxCNC/linuxcnc/pull/4132

andypugh:
Thanks

PR #4132, opened 2026-06-05
Proper and safe evaluation of realtime capability

hdiethelm:
Intended to be used with:
https://github.com/LinuxCNC/linuxcnc/pull/4107

Will fix is_sim / is_rt which is broken: https://github.com/LinuxCNC/linuxcnc/issues/4129

Two new functionality's for two use-cases:

1. Not running to see if the system is actually capable of running RT
2. Running to query the real time capability and type

`realtime status` can be used to check if realtime is running.

The realtime script is extended with the `verify` command returning 0 if RT capable, 1 if not. 
It is intended to use when not running and running.
- RTAI: Returns always 0
- uspace: calls `rtapi_app getrt` and returns the state
  - Not running: rtapi_app performs all the checks and returns immediately
  - Running: rtapi_app calls the master for the real time capability and returns the state

There is the new function `hal_get_realtime_type()` returning the type of the actually running realtime system trough the hal.

is_sim / is_rt use `realtime verify` at init.

`rtapi_is_realtime()` is deprecated: It works only in real time context since https://github.com/LinuxCNC/linuxcnc/pull/3964 and was never 100% reliable, also according to the doc.


BsAtHome:
I do think it to be very problematic that getting the RT status is so involved. There should be a <i>simple</i> test that does not involve instantiating larger parts of infrastructure.

hdiethelm:
:-D The CI has no realtime:
ERROR MOTION: no realtime detected.

It is somewhat a chicken and egg problem. Not involving the RT infrastructure needs separate code and this can always not be in sync with RT. Involving RT runs a lot of stuff. Let's think about this. Exactly the issue here: https://github.com/LinuxCNC/linuxcnc/issues/4129

What do you think about a parameter / pin? So you can use getp / gets to ask for realtime?

Or I could add a new command path to halcmd / rtapi_app that does not start the hal if it is not yet running.

BsAtHome:
Running non-RT is not an error per se (see CI). Therefore, you cannot and must not "simply" or "blindly" force one or the other.

There are use-cases where you want to know the RT status and that does not always mean that you will or will not be running either. Finding out what the RT status is or <i>will be</i> must be lightweight and may be different from where you ask. Doing it in a component or from the cmd-line may be different, depending <i>how</i> you ask and with what intention you ask.

grandixximo:
Both CI failures are small:

- `rip-and-test-clang`: `has_setuid_root` is unused on the default uspace build (the `detect_*` callers are all behind `USPACE_RTAI`/`XENOMAI` ifdefs, so the stubs never call it), and `-Werror=unused-function` kills it. `[[maybe_unused]]` or moving it inside the ifdefs fixes it.
- `rip-rtai`: `hal.h:240 extern rtapi_realtime_status_t hal_realtime_status();` needs `(void)`. The RTAI kernel-module build uses `-Werror=strict-prototypes`, so the empty `()` fails (clang on the uspace build let it through). Same on the definition in hal_lib.c.

Two runtime things I noticed while reading:
- `rtapi_realtime_status()` returns `LXRT` for `detect_rtai()`, but `makeApp()` only handles `RTAI`, so a uspace RTAI build hits the final `else` and sets `app = nullptr`.
- If `can_set_sched_fifo()` succeeds but the kernel string matches none of the markers, it falls through to `NONE` and runs SCHED_OTHER. The old code treated SCHED_FIFO success alone as realtime, so this is a regression on plain-PREEMPT/generic kernels; the fallback there probably wants to stay RT-capable.

Minor: `REALTIME_STATUS_PREEMT_*` is missing a `P` (`PREEMPT`), and `detect_preempt_dynamic()` tests the same string on both sides of the `||`.

src/hal/halmodule.cc
```diff
@@ -2378,8 +2387,9 @@ PyMODINIT_FUNC PyInit__hal(void)
     PyModule_AddIntConstant(m, "HAL_OUT", HAL_OUT);
     PyModule_AddIntConstant(m, "HAL_IO", HAL_IO);
 
-    PyModule_AddIntConstant(m, "is_sim", !rtapi_is_realtime());
-    PyModule_AddIntConstant(m, "is_rt", rtapi_is_realtime());
+    //ToDo: Search all references and eliminate them
+    //PyModule_AddIntConstant(m, "is_sim", !rtapi_is_realtime());
+    //PyModule_AddIntConstant(m, "is_rt", rtapi_is_realtime());
```
BsAtHome: This breaks the *external* API. Anybody who uses it gets bitten.

This can become a dynamic property instead.

src/hal/halmodule.cc
```diff
@@ -1829,6 +1829,13 @@ PyObject *get_info_params(PyObject * /*self*/, PyObject * /*args*/) {
     return python_list;
 }
 
+static PyObject *pyhal_realtime_status(PyObject * /*self*/, PyObject * /*o*/) {
+    // hal_ready did not exist in EMC 2.0.x, make it a no-op
+    TEST_HAL_SHMEM_BASE(__FUNCTION__);
+    int res = hal_realtime_status();
+    return PyLong_FromLong(res);
```
BsAtHome: Exposing this as value means you must expose the (named values of the) enum in halmodule too.

hdiethelm:
Thanks. Fixed.

One quirk: RTAI in userspace is called LXRT. I should rename this consistently.
https://www.rtai.org/userfiles/documentation/magma/html/api/group__lxrt.html#details

src/rtapi/rtapi.h
```diff
@@ -985,7 +985,24 @@ int rtapi_spawnp_as_root(pid_t *pid, const char *path,
 #endif
 
 extern int rtapi_is_kernelspace(void);
+
+typedef enum{
+    REALTIME_STATUS_UNINITIALIZED = -1,
+    REALTIME_STATUS_NONE = 0,
+    REALTIME_STATUS_RTAI = 1,
```
BsAtHome: What does status "STATUS_NONE" mean? It is not really a status but an indication of <i>what</i> is running.

Maybe it should be called `REALTIME_IS_NONE`. Would probably also work for the others.

src/hal/halmodule.cc
hdiethelm: Yes, this is true. How can I do a dynamic property?
However, the behavior will be a bit different. is_rt will return false as long as rtapi_app is not running.
is_sim is also a bit outdated I think.

src/hal/halmodule.cc
hdiethelm: Yes, still to do. How can I expose an enum to python?

src/rtapi/rtapi.h
hdiethelm: Naming: Yes, it is not a status really. It is more like an attribute.
So: 
REALTIME_STATUS_ -> REALTIME_IS_
and
rtapi_realtime_status -> rtapi_realtime_type (and others) might be?

hdiethelm: https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/62a53b8f3b6be232ac9a4b4510b945f4b6690da7
It is more a TYPE . So I renamed it REALTIME_STATUS_ -> REALTIME_TYPE_
https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/fbf2b813cd5bde9451b62d6825f78966a4bd1e54
renames rtapi_realtime_type -> rtapi_get_realtime_type
rtapi seldom uses _get_ If you think this is to much, I will revert it again.

hdiethelm:
Hmm, time for a break, to many force pushes, sorry. I will continue tomorrow.

So you are OK with the general concept? Then I will polish it up, update the doc and do some more testing in different combinations.

Open:
- Python enum
- Python is_sim / is_rt: I would prefer just removing them
  - Alternative: Dynamic property, but this is also halve breaking due to unknown has to be handled. I could throw an error in this case.

Deferred:
 - rtapi_app start / stop behaivour

grandixximo:
The CI fixes and the LXRT / fallback handling look good now.

On naming, with @BsAtHome's #4099 in mind: `rtapi_get_realtime_type()` is consistent with the existing `rtapi_get_*` getters. The hal-side `hal_get_realtime_type()` is the one I'd reconsider, since #4099 is standardizing `hal_get_<datatype>(ref)` (e.g. `hal_get_si32(ref)`, `hal_get_bool(ref)`) where the suffix is a HAL data type and it takes a typed ref. A parameterless `hal_get_realtime_type()` reads off-pattern in that family; something like `hal_realtime_type()` would keep `hal_get_*` reserved for the value accessors. @BsAtHome owns that convention, so up to him.

For the two open how-tos:
- Dynamic `is_rt`/`is_sim` without breaking the API: `lib/python/hal.py` already wraps `_hal` (`from _hal import *`), so a PEP 562 module `__getattr__` there gives live values: return `get_realtime_type() > 0` for `is_rt` and `<= 0` for `is_sim`. Existing callers (pncconf, stepconf) keep working; the one change is that before rtapi_app runs the value is -1, so is_rt reads False / is_sim True.
- Exposing the enum to Python: one `PyModule_AddIntConstant(m, "REALTIME_TYPE_NONE", REALTIME_TYPE_NONE)` per value, consistent with the existing `HAL_BIT` etc. constants.

BsAtHome:
> something like `hal_realtime_type()` would keep `hal_get_*` reserved for the value accessors. @BsAtHome owns that convention, so up to him.

Well, I don't  own it. However, I agree with the argument to leave the get/set moniker to the hal pin/param data access.

hdiethelm:
The last few commits are still figuring out ways to solve all issues, not ready for code style review yet.

@grandixximo Thanks a lot for the hints. Helps a lot not to have to search everything.

- hal_get_realtime_type: There is also hal_get_param_value_by_name and so on, so also consistent.
- Python enum: Done
- `is_rt` / `is_sim` That one was annoying: First, i need to create a component if there is none yet and a lot of error handling: https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/bdd3ef900606866db18498a907a0d48963530375  And then I discovered that stepconf.py / stepconf.py need is_sim before realtime is up and running. So roll back and just call `realtime verify` at init: https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/cdbace2c96bef8cc13e8f20210073b015dce6b8f However, now this is fully backwards compatible, no side effects expected.
- I left the python `is_initialized` function from ^ in, might be this is useful. Many functions can only be used if there is already a component. So you can check if there is already one and create one if needed. `hal.component_exists()` needs also a component to succeed. I tried... ;-)

hdiethelm:
And of course, after debian package install, the realtime script is not any more in path. And there was also no define for the path where it is. https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/af73e0883f9bc78ec3da78d845f864c2bdee07aa adds one.

For scripts, `REALTIME=@REALTIME@` is used, so it is available there.

src/hal/halmodule.cc
hdiethelm: Done using "realtime verify". rtapi_is_realtime() works only when realtime is already running and the two instances in the source I found where before. So a dynamic property would have been a breaking change.

src/rtapi/rtapi.h
hdiethelm: Done

hdiethelm:
Due to error handling is annoying not knowing when rtapi_app is running or not:
https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/579f7ed16e5036371a4bd1c47ca01d99ff263c53
Let's see how much fall out this generates...

It now works exactly the same as RTAI. `realtime start` is now needed before every halcmd involving realtime.
Everything else including halrun / linuxcnc usw. works as before what I tested so far.

And it generated a new ToDo: Cleanup the realtime script. It is a mess. There are a lot of RTAI only parts executed in uspace mode. Like loading an empty list of modules and so on. So far, I just added the things I needed.

grandixximo:
Two notes after the latest push.

Naming: you're right, I'll withdraw my concern. `hal_get_realtime_type()` matches the existing `hal_get_lock()`, which is already a parameterless global-state getter, so `hal_get_*` isn't exclusively the typed-ref family. Leave it as is.

CI / auto-start: the two failures (`raster`, `hal-show`) come from dropping the uspace auto-start of `rtapi_app` on first `loadrt`. `raster` runs `halcmd -f raster.hal` (a `loadrt` with no `realtime start`) and now gets `No master found`. That breaks standalone `halcmd -f *.hal` scripts in general, so I'd suggest not hard-breaking it: keep auto-start working but emit a one-time deprecation warning pointing at `realtime start`. That keeps existing scripts working and gives a migration path. Then migrate our own scripts/configs/tests to call `realtime start` explicitly so the tree models the new idiom (and update the expected outputs, since a stderr warning will otherwise trip output-comparison tests like `hal-show`).

Separately: is dropping the auto-start actually needed for the RT-status goal, or is it a cleanup that could be its own PR? Keeping them decoupled would let the getrt/`realtime_type` work land without the broader behavior change.

hdiethelm:
If these test fails due to a missing realtime start, they most probably fail also with RTAI, i have to test it.

It is not needed in this PR. However, as soon you like to use `hal_realtime_type()` in many places, it will help a lot, duet to realtime needs to be started before.

I will move it in a separate PR as initialy planned, this one gets already big.

hdiethelm:
^Rebased to master / removed [rtapi_app: start command / do not autostart](https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/579f7ed16e5036371a4bd1c47ca01d99ff263c53)

This commit lives on a new branch to create a PR later: https://github.com/hdiethelm/linuxcnc-fork/tree/rtapi_no_autostart I will wait with this until the PR at hand is merged. There are some conflicts if I rebase rtapi_no_autostart to main.

hdiethelm:
So, it is slowly taking shape but still not final. I updated the doc and cleaned up some parts.

Some things I am considering:
- `hal_get_realtime_type()` returns an rtapi type. A `typedef rtapi_realtime_type_t hal_realtime_type_t` would clean this up.
- `rtapi_get_realtime_type()` should only be used in hal.c. User / rt components should all use hal_get_realtime_type(). This would need a new header file rtapi_hal_priv.h to enforce this.

Before merge: The history is a mess. Should I squash it to a single commit or more? The commits could be:
- python is_initialized
- rtapi_app getrt / rtapi_get_realtime_type() / deprecate rtapi_is_realtime() / python is_sim + is_rt fix
- hal_get_realtime_type() / pass trough hal

grandixximo:
Please squash 😁

3 commits is clean

hdiethelm:
So, squashed and cleanup. New:

- renamed getrt to check_rt
- hal_realtime_type_t
- rtapi_app only shows the realtime issue report once when started in master mode
- RTAPI_HAL_PRIV to make rtapi_is_realtime() / rtapi_get_realtime_type() unavailable except in hal / rtapi

ToDo:
- Review
- Test on many different combinations setsuid / setcap / rtai / uspace and so on

> and update the expected outputs, since a stderr warning will otherwise trip output-comparison tests like hal-show

I will look into this. Is there a script to do this?

grandixximo:
No dedicated rebase script, it's manual. Run `runtests -n` (the `-n` keeps the temp files instead of cleaning up), then for each test that uses an `expected` file `cp <testdir>/result <testdir>/expected` once you've confirmed the new output is correct. Tests that use a `checkresult` script instead of `expected` don't apply. For this PR it's probably moot anyway since the auto-start warning moved to the separate PR and CI is green; it'll matter there.

hdiethelm:
So, I went trough the full testing matrix. Most works as expected. There is just something to consider:
If no PREEMPT_RT or other hard real time kernel is running, but it supports SCHED_FIFO, the type is: REALTIME_TYPE_PREEMPT_DYNAMIC
And is_rt is true.

This is strictly speaking still realtime, just a bad one. On the terminal, there is a warning.

I could change to:
```
typedef enum{
    REALTIME_TYPE_UNINITIALIZED = -1, //Realtime not running, type unknown
    REALTIME_TYPE_NONE = 0,           //No realtime available
    REALTIME_TYPE_PREEMPT_DYNAMIC = 1,
    REALTIME_TYPE_RTAI = 2,
    REALTIME_TYPE_PREEMPT_RT = 3,
    REALTIME_TYPE_LXRT = 4,
    REALTIME_TYPE_XENOMAI = 5,
    REALTIME_TYPE_XENOMAI_EVL = 6,
} rtapi_realtime_type_t;
```
and then check type > REALTIME_TYPE_PREEMPT_DYNAMIC.

Or alternatively, use `REALTIME_TYPE_NONE` if the kernel is PREEMPT_DYNAMIC and only use PREEMPT_DYNAMIC if env LINUXCNC_FORCE_REALTIME is set.

hdiethelm:
After some consideration, I think auto selecting unknown or PREEMPT_DYNAMIC is probably a bad idea and can generate a lot of confusion for not to experienced users.

Rather select the worst possible option, so it is clear something is wrong, instead of selecting a halve working variant.

So with the last commit, LINUXCNC_FORCE_REALTIME=1 is needed to select these options and there is a new type REALTIME_TYPE_UNKNOWN (SCHED_FIFO available but not PREEMT_DYNAMIC)

Feedback?

grandixximo:
The "pick the worst option so it's obvious something is wrong" instinct is right for *backend selection*,we shouldn't auto-load a half-working RTAI/Xenomai backend. But this lands on the same problem @BsAtHome flagged earlier, that we "must not blindly force one or the other" and the answer may differ by where you ask. The commit gates one knob too many: it conflates *what type do I report* with *do I grant SCHED_FIFO*.

Unforced `PREEMPT_DYNAMIC` now goes to `REALTIME_TYPE_NONE`, and `makeApp()` maps `NONE` to the posix backend at `SCHED_OTHER`. On master that same machine runs posix at `SCHED_FIFO`. So this is a real latency regression for every stock-kernel box with `setcap`/`setuid`, including the default Debian dev box (my setup) and the github CI?. `is_rt` also flips 1 => 0 for those users. The only consumers are the stepconf and pncconf wizards: their `check_for_rt()` then pops "testing / tuning of hardware is unavailable" and blocks the axis-test step (stepconf `test_axis`, pncconf test pages) on boxes that today let you jog and tune hardware. That could be a headache for some devs?

Suggestion: keep using `SCHED_FIFO` whenever `can_set_sched_fifo()` succeeds (SCHED_FIFO still beats SCHED_OTHER even on a non-RT kernel) and only gate the honest *type label* plus warning. Unforced `PREEMPT_DYNAMIC` would still run posix-SCHED_FIFO and emit the "not PREEMPT_RT, latency may be unbounded" note, but report a soft/dynamic type instead of `NONE`. `LINUXCNC_FORCE_REALTIME` then only controls whether we *also* try the hard backends. That avoids demoting working setups while never silently claiming hard RT. If we do keep unforced dynamic as `NONE`/`SCHED_OTHER`, it needs a loud behavior-change callout, since it moves existing machines off SCHED_FIFO untouched.

Smaller, not blockers:

- `harden_rt()` is now `void`, so `makeApp()` can't fall back to posix when `iopl()`/hardening fails. Master downgraded to SCHED_OTHER before; now it proceeds on the RT backend regardless. Intentional?
- `hal_get_realtime_type()` returns `-EINVAL` (-22) when `hal_data` is null, but the doc says `REALTIME_TYPE_UNINITIALIZED` (-1). Python then hands back -22, not an exposed constant. Return `UNINITIALIZED` or document it.
- `do_unload_cmd("hal_lib")` at master exit looks unrelated to RT detection; a word on why it's in this PR.
- Doc/spelling: `hal_get_realtime_type.3.adoc` says `rtapi_all` twice (=> `rtapi_app`), plus "save" => "safe", "withouth" => "without". Typo cluster in new code/comments: `PREEMT_DYNAMIC`, `recomended`, `avaliable`, `chanin`, `spawm`, `otherwhise`, and commit subject `rtapi_is_realtimei`. Cleanup pass before squash.

hdiethelm:
> The "pick the worst option so it's obvious something is wrong" instinct is right for _backend selection_,we shouldn't auto-load a half-working RTAI/Xenomai backend. But this lands on the same problem @BsAtHome flagged earlier, that we "must not blindly force one or the other" and the answer may differ by where you ask. The commit gates one knob too many: it conflates _what type do I report_ with _do I grant SCHED_FIFO_.
> 
> Unforced `PREEMPT_DYNAMIC` now goes to `REALTIME_TYPE_NONE`, and `makeApp()` maps `NONE` to the posix backend at `SCHED_OTHER`. On master that same machine runs posix at `SCHED_FIFO`. So this is a real latency regression for every stock-kernel box with `setcap`/`setuid`, including the default Debian dev box (my setup) and the github CI?. `is_rt` also flips 1 => 0 for those users. The only consumers are the stepconf and pncconf wizards: their `check_for_rt()` then pops "testing / tuning of hardware is unavailable" and blocks the axis-test step (stepconf `test_axis`, pncconf test pages) on boxes that today let you jog and tune hardware. That could be a headache for some devs?

The stepconf / pncconf should be fixed. Not having a way for "proceed anyway" is not nice. This is also nothing new, behaves like this also in 2.9 with vanilla kernel + setuid active.

> 
> Suggestion: keep using `SCHED_FIFO` whenever `can_set_sched_fifo()` succeeds (SCHED_FIFO still beats SCHED_OTHER even on a non-RT kernel) and only gate the honest _type label_ plus warning. Unforced `PREEMPT_DYNAMIC` would still run posix-SCHED_FIFO and emit the "not PREEMPT_RT, latency may be unbounded" note, but report a soft/dynamic type instead of `NONE`. `LINUXCNC_FORCE_REALTIME` then only controls whether we _also_ try the hard backends. That avoids demoting working setups while never silently claiming hard RT. If we do keep unforced dynamic as `NONE`/`SCHED_OTHER`, it needs a loud behavior-change callout, since it moves existing machines off SCHED_FIFO untouched.
> 
Just checked: Using SCHED_FIFO without PREEMT_RT was introduced by https://github.com/LinuxCNC/linuxcnc/pull/3964. Before, SCHED_OTHER was used when the kernel is not PREEMT_RT or otherwise realtime capable. SCHED_FIFO was used with non-PREEMT_RT kernels only with LINUXCNC_FORCE_REALTIME=1. Also, since  https://github.com/LinuxCNC/linuxcnc/pull/3964 is_rt is broken and always returns false. 

So it tracks back this behavior how it was before essentially.

One other option: If SCHED_FIFO is available but no realtime kernel, I can still use SCHED_FIFO. In this case, I will set REALTIME_TYPE_NONE / is_rt=false.
REALTIME_TYPE_PREEMPT_DYNAMIC / is_rt=true is only set if LINUXCNC_FORCE_REALTIME=1.
But this feels not really consistent, using PREEMPT_DYNAMIC and not reporting it also.

> Smaller, not blockers:
> 
>     * `harden_rt()` is now `void`, so `makeApp()` can't fall back to posix when `iopl()`/hardening fails. Master downgraded to SCHED_OTHER before; now it proceeds on the RT backend regardless. Intentional?

It was essentially void before. The only case when it returned 1 was when rtapi_is_realtime() failed, which was checked later with `(rtapi_is_realtime()  || harden_rt() < 0)`. So making it really void removes this redundancy.

> 
>     * `hal_get_realtime_type()` returns `-EINVAL` (-22) when `hal_data` is null, but the doc says `REALTIME_TYPE_UNINITIALIZED` (-1). Python then hands back -22, not an exposed constant. Return `UNINITIALIZED` or document it.

Python hands back:
```
  File "/home/hannes/linuxcnc-src/src/../../linuxcnc-tmp/test2.py", line 8, in <module>
    print("HAL get_realtime_type " + str(hal.get_realtime_type()))
                                         ~~~~~~~~~~~~~~~~~~~~~^^
RuntimeError: pyhal_get_realtime_type: Cannot call before creating component
```
There are many functions doing this, all undocumented. Same for the hal returning -EINVAL. I can document this everywhere where missing.

> 
>     * `do_unload_cmd("hal_lib")` at master exit looks unrelated to RT detection; a word on why it's in this PR.
Ah yes, i forgot to mention this. I believe unloading hal_lib was missed at rtapi_app exit. This is needed to set back to REALTIME_TYPE_UNINITIALIZED when rtapi_app is not running. I hope there are no other side effects, however I did not see one until now.

> 
>     * Doc/spelling: `hal_get_realtime_type.3.adoc` says `rtapi_all` twice (=> `rtapi_app`), plus "save" => "safe", "withouth" => "without". Typo cluster in new code/comments: `PREEMT_DYNAMIC`, `recomended`, `avaliable`, `chanin`, `spawm`, `otherwhise`, and commit subject `rtapi_is_realtimei`. Cleanup pass before squash.

Thanks, I will fix this.

hdiethelm:
About the behaivor with a vanilla kernel, options:

1. master now / before my last commit

always:
PREEMPT_DYNAMIC
SCHED_FIFO
is_rt = true
Will show realtime popup / warning if latency is to much.

2. 2.9/master before rootless/after my last commit

Normal:
NONE
SCHED_OTHER
is_rt = false
Will NOT show realtime popup / warning

LINUXCNC_FORCE_REALTIME=1
PREEMPT_DYNAMIC
SCHED_FIFO
is_rt = true
Will show realtime popup / warning

3. alternative

Normal:
NONE
SCHED_FIFO
is_rt = false
Will NOT show realtime popup / warning

LINUXCNC_FORCE_REALTIME=1
PREEMPT_DYNAMIC
SCHED_FIFO
is_rt = true
Will show realtime popup / warning

I am for 2. Consistent and no change in behaivour compared to before nonroot. Shows issues when the wrong kernel is running.
1. Is dangerous, i don't like it. 3. a bit inconsistent but why not.

grandixximo:
Agreed, go with 2.

I came in pushing 3, but you've convinced me 2 is the right call for a machine controller. The only one who actually benefits from best-effort SCHED_FIFO on a non-RT kernel is someone running real hardware on the wrong kernel, and that's exactly the case we want to make visibly degraded rather than silently "good enough", since masked latency spikes turn into following errors and lost steps on a real machine. Dev and sim don't need SCHED_FIFO anyway, so 2 costs them nothing, and `LINUXCNC_FORCE_REALTIME=1` is a clean escape hatch for the rare legitimate best-effort setup. 3's split (SCHED_FIFO but is_rt=false, no warning) is the worst of both: it hides the problem without admitting it.

So: 2 it is. Restores pre-rootless behavior, consistent, and surfaces a wrong-kernel boot instead of papering over it.

grandixximo:
One correction to my wording, then the concrete reason for 2.

My "3 hides the problem without admitting it" was wrong: per your tables 2 and 3 report identically (`NONE`, `is_rt=false`) and differ only in the scheduler.

But the "Unexpected realtime delay" warning is gated on the scheduler policy, not on `is_rt`:

```cpp
// uspace_posix.cc, RtapiTask::wait(), in the overrun branch:
if (policy == SCHED_FIFO)
    unexpected_realtime_delay(task);
```

`makeApp()` runs the posix backend at `SCHED_OTHER` for `NONE` and at `SCHED_FIFO` for any RT type, and `unexpected_realtime_delay()` has no `is_rt` check of its own. So a `SCHED_FIFO` thread emits that warning when it overruns, whatever type we report.

That makes the option-3 normal row (`SCHED_FIFO` + "no warning") not match the code: it would warn. The other rows are consistent: option-2 normal is `SCHED_OTHER` => no warning; option-2 forced and option-1 are `SCHED_FIFO` => warning.

So option-3 normal as drawn would behave like option 1 with `is_rt=false` bolted on: still SCHED_FIFO, still warns on overrun, just reports not-realtime. To get the "SCHED_FIFO + silent" row, you'd have to suppress the overrun warning while still scheduling SCHED_FIFO, i.e. run RT scheduling but hide RT overruns. That's the variant I'd argue against.

That's why 2 is clean: scheduler, overrun warning, and `is_rt` stay aligned with no extra code. Normal mode is `SCHED_OTHER` and genuinely non-RT; forced mode opts into all three.

hdiethelm:
Option 3 would need some additional code changes to separate SCHED_FIFO from the warning and the reported RT type. Surely possible but if you agree option 2 is the best, this is already implemented.

@BsAtHome Are you also fine with option 2?

Done:
- Spelling mistakes (Not my strength, I hope I caught all of them...)
- Doc Python Exceptions
- Doc HAL status code in hal_get_realtime_type
- Some retesting after my last change
- Test on real CNC

Still open:
- Nothing from my side

src/hal/hal_lib.c
```diff
@@ -53,6 +53,7 @@
 
 */
 
+#define RTAPI_HAL_PRIV  /* Use rtapi/hal private functions */
 #include <rtapi.h>		/* RTAPI realtime OS API */
```
BsAtHome, line 56: Either you publish enums and functions or you do not.

Everything in `rtapi.h` `hal.h` is available and should not be hidden by any define construct. If you need it such construct, then something is wrong.
What is private remains private. What is public is always public.

src/hal/halmodule.cc
```diff
@@ -27,6 +27,7 @@
 #include <rtapi_mutex.h>
 #include "hal.h"
 #include "hal_priv.h"
+#include "config.h"
```
BsAtHome: Why do you need this config include?

src/hal/halmodule.cc
```diff
@@ -2378,8 +2393,25 @@ PyMODINIT_FUNC PyInit__hal(void)
...
+    PyModule_AddIntConstant(m, "REALTIME_TYPE_LXRT", REALTIME_TYPE_LXRT);
+    PyModule_AddIntConstant(m, "REALTIME_TYPE_XENOMAI", REALTIME_TYPE_XENOMAI);
+    PyModule_AddIntConstant(m, "REALTIME_TYPE_XENOMAI_EVL", REALTIME_TYPE_XENOMAI_EVL);
+
+    //Call realtime verify to gather realtime status
+    //Most probably we don't have realtime running yet
+    int ret = system(EMC2_REALTIME " verify > /dev/null");
+    int exit_stat = WEXITSTATUS(ret);
+    if(exit_stat != 0 && exit_stat != 1){
+        PyErr_Format(PyExc_RuntimeError, "realtime verify failed, system() return value %i / exit %i", ret, exit_stat);
```
BsAtHome: Calling system() in the hal module is extremely ugly and a sign that something is amiss.

src/hal/halmodule.cc
hdiethelm: For EMC2_REALTIME

src/hal/halmodule.cc
hdiethelm: The only alternative I see is to remove is_sim / is_rt and call realtime verify in the two places this was used.

src/hal/hal_lib.c
hdiethelm: I don't need it. The target is to avoid hal realtime components using this two functions. But they can, it will work.
So i will remove this, no problem.

src/hal/halmodule.cc
BsAtHome: Making the constants properties will allow you to check in a function. The library user will not see the difference.

Question: is `is_rt` and `is_sim` called before a component is created?

If there is HAL connection, then you can directly use a pin/param's value exported by the rtapp's inner process.

If there is no connection (hal_init() was not called) then you need to establish a temporary connection (component) so that you can get the values you need.

hdiethelm: To use properties that call get_realtime_type() successfully, two conditions must be fulfilled. 

- rtapi_app must be running
- A component must exist

Both are not fulfilled at the two places this is used:
https://github.com/LinuxCNC/linuxcnc/blob/c1f4b6cd7d6b373176a6a1d7e5253048b481a531/src/emc/usr_intf/stepconf/stepconf.py#L896
https://github.com/LinuxCNC/linuxcnc/blob/c1f4b6cd7d6b373176a6a1d7e5253048b481a531/src/emc/usr_intf/pncconf/pncconf.py#L655

To have is_sim / is_rt useful, it should work always, you should not have to check that you fulfill this conditions. Two options:

1. Use `realtime verify` -> In this PR
2. Make sure both conditions are met

I already implemented 2. before 1. it but discarded it again, mainly to not overgrow this PR into yet another "fix everything" PR: https://github.com/LinuxCNC/linuxcnc/pull/4132#issuecomment-4641953139

- rtapi_app must be running
  - This needs a start / stop command
  - Removing autostart / autostop is not really needed but would make sense to avoid inconsistencies
  - Works mostly, some tests fail due to `realtime start`  / `realtime stop` is missing. They probably will also fail with RTAI.
  - Commit: https://github.com/hdiethelm/linuxcnc-fork/commit/612183ae4007dac156556302e753bdf3cd325664
- A component must exist
  - We can create a temporary one and discard it again
  - Works, needs some cleanup
  - Commit: https://github.com/hdiethelm/linuxcnc-fork/commit/cdcc34420346984f8203aec87b29c7f9ab4e5c45

Both commits contain also some changes that I found useful anyway and are still in this PR.

At the moment, rtapi_app is started at first `load_rt` command and exits at `exit` or after last component has been unloaded. This makes it cumbersome to use `get_realtime_type()` due to you can do it only after first `load_rt`, otherwise you will get `REALTIME_TYPE_UNINITIALIZED`.

RTAI always needs `realtime start` to before first command, `realtime stop` at the end, so this is already there. That's why I think it is an option to change from auto to manual start/stop without to much issues.

If you think ^^ is a better solution, I can add it again. It looks in general cleaner to me but it might also break more things.

hdiethelm: BTW: Due to it is only used in two places, replacing it by `realtime verify` and remove is_sim / is_rt would be option 3. However, I don't know who uses this in his own private machine configs.

src/hal/hal_lib.c
hdiethelm: Done, I removed also the DEPRECATED from the man page, it is now back, but only for RT components.

src/hal/halmodule.cc
BsAtHome: > To use properties that call get_realtime_type() successfully, two conditions must be fulfilled.
>     * rtapi_app must be running
>     * A component must exist

Very true. This is always the case for halmodule to be useful <i>at all</i>.

> Both are not fulfilled at the two places this is used:
> src/emc/usr_intf/stepconf/stepconf.py#L896
> src/emc/usr_intf/pncconf/pncconf.py#L655

But HAL isn't useful for pncconf or stepconf because no backend is running. The realtime state for HAL is only a valid case when rtapi_app is running. Using halmodule to determine its state while the system is not initialized is essentially wrong. It is certainly not how halmodule is supposed to work.

Trying to fix this corner case in halmodule is IMO wrong and it should be fixed in pncconf and stepconf.

It also raises the question whether is_sim/is_rt is actually necessary in halmodule? It looks like its only use is in stepconf and pncconf (and is_rt is only in stepconf).

Alternatives may be appropriate here.

hdiethelm: > > To use properties that call get_realtime_type() successfully, two conditions must be fulfilled.
> > * rtapi_app must be running
> > * A component must exist
> 
> Very true. This is always the case for halmodule to be useful at all.

Only 99%. You can use the hal with only user components and have rtapi_app not running. Not sure why but you can. In this case you probably won't bother to check for RT.

> Alternatives may be appropriate here.

If you are ok with (somewhat above in the huge thread):
- Python is_sim / is_rt: I would prefer just removing them

I just have to find a way to use the realtime script inside this stepconf / pncconf. The check is anyway annoying in there, you are forced to have realtime, no way to "continue anyway". So I will get rid of is_sim / is_rt, use `realtime status` and fix the check so it is not stopping you by force. Ok?

BsAtHome: > > Very true. This is always the case for halmodule to be useful at all.
> 
> Only 99%. You can use the hal with only user components and have rtapi_app not running. Not sure why but you can. In this case you probably won't bother to check for RT.

And that is the point. When you only use it for non-RT, then why ask about RT? You aren't using RT in non-RT so it is just like <i>"Hey, look, I can run non-RT on RT systems! Am I cool or what?"</i>. Right...


> > Alternatives may be appropriate here.
> 
> If you are ok with (somewhat above in the huge thread):
>     * Python is_sim / is_rt: I would prefer just removing them

Yes, it is not used anywhere else and it is rather uninformative/complicated/etc..

We should just create a RO parameter in rtapi_app (the RT running core) that sets the appropriate value to indicate how the system is running.

> I just have to find a way to use the realtime script inside this stepconf / pncconf. The check is anyway annoying in there, you are forced to have realtime, no way to "continue anyway". So I will get rid of is_sim / is_rt, use `realtime status` and fix the check so it is not stopping you by force. Ok?

Lets try, yes.

Question: is configuration building with stepconf and pncconf different depending on whether it detects running on a RT system? What about cross-configuration (like cross-compiling)?

hdiethelm: > We should just create a RO parameter in rtapi_app (the RT running core) that sets the appropriate value to indicate how the system is running.
> 

There is no parameter withouth a component.
I have a commit somewhere creating a static component at hal_lib init, but this fails all tests due to there is a new unexpexted component and parameter. But I could add just a simple rt_info.comp you can manually load if needed.

> 
> Lets try, yes.
> 
> Question: is configuration building with stepconf and pncconf different depending on whether it detects running on a RT system? What about cross-configuration (like cross-compiling)?

It checks RT only to say you are not allowed to do axis tests withouth RT and locks you out.

BsAtHome: RT components always run in the rtapi_app context. They can use rtapi_*() calls at any time to determine their status.

Userspace components need a HAL connection and could query a HAL parameter. This parameter must be added via a RT component. Other userspace could use halcmd getp.

The question is then do we add this component implicitly or do we need to do so explicitly.

The case for explicit (put it in a .hal file) is there because if you need the info, then you already know that you need the info and add it to the tree of components.
The case for implicit is that you no longer have to think about adding the component and the information is there. If this can be done without that unloading RT becomes a problem (the rmmod problem, I believe), then this may be the easier option for users. Otherwise, explicit must be the way.

For stepconf/pncconf, well, I guess it can spawn a `realtime verify` and sample the result? Would that suffice? That said, stepconf and pncconf need to build a working .hal file and load/run RT anyway for testing axes. Why not let them load the isrt.comp and sample the information? If it runs in sim-mode, then the programs can disable their controls or tear down rtapi_app again?

hdiethelm: With thia PR, you have hal_get_realtime_type() that works in user / rt comps and python. So no real need for a parameter for most usecases I think.

Only issue: This works only after the first load_rt due to rtapi_app is only running afterward. https://github.com/hdiethelm/linuxcnc-fork/commit/612183ae4007dac156556302e753bdf3cd325664 would fix this.

> For stepconf/pncconf

realtime verify would work, except the install path of this scipt needs to be available, i did not find it exposed in python yet.
hal.get_realtime_type()  would also work, but only after the first load_rt. No need for a parameter, it will be there anyway only after the first load_rt, implicit or explicit, same issue.

Or I just add the commit above and fix the tests, then hal_get_realtime_type() will work after halrun.

hdiethelm:
@BsAtHome https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/1799fef2b23146499814b96fbbb41028c8c475b0 replaces hal.is_sim / hal.is_rt by realtime.verify().
A python wrapper makes it simple to use. I need an .py.in file to get the path. I did not find anywhere in the python part where this path is already available.
It still needs some dok but I think the idea is clear. Do you think this is better? So I will finalize it.

BsAtHome:
Running it with system (a shell) may require you to redirect input from /dev/null too so it cannot hang on waiting for input (unlikely, but you don't need stdin).
Expanding a .py.in file has the advantage that you will get the one singular target script that matches the build. Also in RIP.

Calling the module "realtime" may not be the best of choices and may lead to confusion because the script starting/stopping realtime is called `realtime`. Maybe the python module should be called something in the direction of `lcncrtquery` (because you have both a verify and status)? I'm open for other suggestions.

hdiethelm:
> Running it with system (a shell) may require you to redirect input from /dev/null too so it cannot hang on waiting for input (unlikely, but you don't need stdin). 

Stdin is not needed but better save than sorry, right.

>Expanding a .py.in file has the advantage that you will get the one singular target script that matches the build. Also in RIP.

Yes, i found out that the realtime script has a special path after a debian package install, so this is the only way to always find it.

> Calling the module "realtime" may not be the best of choices and may lead to confusion because the script starting/stopping realtime is called `realtime`. Maybe the python module should be called something in the direction of `lcncrtquery` (because you have both a verify and status)? I'm open for other suggestions.

As a wrapper of this script, start / stop / restart / ... could be added to. There was no need to do this now, but i can add all commands. So I think the name is mostly fine. Might be lcncrealtime?

Something like:
realtime.start()
halcmd...
realtime.stop()
is sometimes usefull. It is not needed for uspace (except we move to manual start/stop of rtapi_app), but RTAI needs this.
Unitl now, this is often done in bash around calling python or other binarys or in halrun around calling halcmd.

hdiethelm:
@BsAtHome Are you ok with realtime.py or might be lcncrealtime.py?

BsAtHome:
Yes, I do think that lcncrealtime.py would be better. Otherwise we risk confusion with the realtime script. If we ever are to replace the shell script with the python code, then we can simply exec in the old script.

hdiethelm:
After looking at that name to long I went for `lcnc_realtime.py`. I found it hard to read without underscore. Is this fine to?

This probably would need some doc? Where to put it? The only doc for python code I found is `src/config/python-hal-interface.adoc`. Or is it good enough as it is documented in the python code right now?

BsAtHome:
> After looking at that name to long I went for `lcnc_realtime.py`. I found it hard to read without underscore. Is this fine to?

This is a module you use in import.I think using an underscore is rather unusual, or not?

> This probably would need some doc? Where to put it? The only doc for python code I found is `src/config/python-hal-interface.adoc`. Or is it good enough as it is documented in the python code right now?

Somewhere... It needs some reorganisation either way. However, this is not a HAL interface, but an RTAPI interface if I look at this right.

hdiethelm:
> > After looking at that name to long I went for `lcnc_realtime.py`. I found it hard to read without underscore. Is this fine to?
> 
> This is a module you use in import.I think using an underscore is rather unusual, or not?

I seldom work in python. But as much as I know you normally create packages, so it would be:
from lcnc import realtime
realtime.verify()

Or:
import lcnc
lcnc.realtime.verify()

This way, you don't take the whole namespace. The way it is now, hal is taken from linuxcnc, so you can not use any other package / class named hal on your system.

Creating a a lcnc package for everything would be an other project.
I can try to start with realtime creating a lcnc package with only this one inside.

> 
> > This probably would need some doc? Where to put it? The only doc for python code I found is `src/config/python-hal-interface.adoc`. Or is it good enough as it is documented in the python code right now?
> 
> Somewhere... It needs some reorganisation either way. However, this is not a HAL interface, but an RTAPI interface if I look at this right.

So `src/config/lcnc_realtime.adoc`?

hdiethelm:
Rebased to master and created a package.

Now you can use:
```python
from lcnc import realtime

print("realtime.verify " + str(realtime.verify()))
print("realtime.status " + str(realtime.status()))
```
or
```python
import lcnc

print("realtime.verify " + str(lcnc.realtime.verify()))
print("realtime.status " + str(lcnc.realtime.status()))
```

I am seeing quite a lot of random files from linuxcnc installed just in top of dist-packages:
```
/usr/lib/python3/dist-packages/raster.py
/usr/lib/python3/dist-packages/preview_helpers.py
/usr/lib/python3/dist-packages/propertywindow.py
/usr/lib/python3/dist-packages/pyhal.py
/usr/lib/python3/dist-packages/pyngcgui.py
/usr/lib/python3/dist-packages/hal.py
/usr/lib/python3/dist-packages/hal_glib.py
/usr/lib/python3/dist-packages/hershey.py
/usr/lib/python3/dist-packages/lineardeltakins.cpython-313-x86_64-linux-gnu.so
/usr/lib/python3/dist-packages/linux_event.py
/usr/lib/python3/dist-packages/linuxcnc.cpython-313-x86_64-linux-gnu.so
/usr/lib/python3/dist-packages/linuxcnc_util.py
/usr/lib/python3/dist-packages/multifilebuilder.py
/usr/lib/python3/dist-packages/common
/usr/lib/python3/dist-packages/common/colored_formatter.py
/usr/lib/python3/dist-packages/common/hal_glib.py
/usr/lib/python3/dist-packages/common/iniinfo.py
/usr/lib/python3/dist-packages/common/logger.py
...
```
^^ looks to me like really bad practice, even if I have basically no clue of python.
It would probably make sense to move them all to a lcnc (or linuxcnc?) python package to not collide with other packages now or in the future. However, that's a whole new project fixing all the imports if you do that.

@grandixximo Do you work more with python? Any insight? Makes my package sens the way I am doing it?

At least after installing the deb, I can run the above code directly with python and stepconf also passes the realtime check, so it is can not be to far... ;-)

Otherwise, I could also just roll back the package creation and stick to lcnc_realtime.py / realtime.py instead of starting a whole new concept of creating a proper linuxcnc python package.

It is not worse than the existing code. realtime.py will never collide with a python implementation of the realtime script, due to it will live in scripts/ or bin/ in the source and under /usr/lib/linuxcnc/realtime after install. The realtime.py module lives under lib/python/realtime.py and under /usr/lib/python3/dist-packages/realtime.py after install.

grandixximo:
Your instinct on the pollution is right: `SITEPY` is the global `/usr/lib/python3/dist-packages`, and we dump ~30 generically-named flat modules (`hal.py`, `gremlin.py`, `nf.py`, `raster.py`) plus generic dirs (`common/`, `touchy/`) straight into the shared namespace. A single branded package is the correct long-term fix, and a bare `realtime.py` there would be a real collision hazard, so you're not wrong to want a package.

That said, for this PR I'd go with the flat `lcnc_realtime.py` and not the package. Two reasons:

1. The package makes a project-wide namespace decision inside a realtime-status PR. "What is LinuxCNC's official python namespace" deserves its own thread, the naming alone already took a week here.
2. We already carry three brand names: `emc2`/`emc` in the build and env layer (`EMC2_HOME`, `EMC2VERSION`, and the `EMC2_REALTIME` define this very PR adds), `linuxcnc` in the package and user layer, and the `linuxcnc` C extension `.so`. `lcnc` would be a fourth. So picking a python namespace is really a chance to consolidate, not just add one more name, and that consolidation shouldn't be decided by default in a status PR. The cleanest target is `linuxcnc`, but the `.so` squats it, and turning that into a `linuxcnc/__init__` + inner `.so` is the real fix, which is exactly the bigger project you don't want to take on here. Starting `lcnc` now risks a second migration later.

So: ship `lcnc_realtime.py` (flat, collision-safe name, matches today's convention, unblocks this and #4107), and open a separate PR/issue for a proper namespace where the dist-packages cleanup gets the attention it deserves, ideally aimed at `linuxcnc`. If you and @BsAtHome would rather land the package now, I'm fine with it too, just get an explicit "lcnc is our namespace going forward" on record first so it isn't decided by default.

One thing that matters more than the packaging: `hal.is_sim` / `hal.is_rt` were documented public attributes (in `python-hal-interface.adoc`, which this PR deletes). Hard-removing them breaks any external user HAL/python config or custom GUI reading them. I'd keep a thin deprecation shim, the names computed via the new path with a warning, rather than removing them outright.

And if the package does stay: drop the eager `from . import realtime` in `__init__.py` so `import lcnc` doesn't pull in every future submodule; use `subprocess.run([REALTIME, "verify"], stdin=DEVNULL, stdout=DEVNULL)` instead of `os.system` (no shell, covers BsAtHome's stdin point); README is missing a trailing newline.

hdiethelm:
> That said, for this PR I'd go with the flat `lcnc_realtime.py` and not the package. ...

Agreed, that's a different project. Adding lcnc_realtime.py and migrating it later to lcnc/realtime.py does adds only minimal overhead to creating a package later.

I just created an issue: https://github.com/LinuxCNC/linuxcnc/issues/4198

> So: ship `lcnc_realtime.py` (flat, collision-safe name, matches today's convention, unblocks this and #4107), and open a separate PR/issue for a proper namespace where the dist-packages cleanup gets the attention it deserves, ideally aimed at `linuxcnc`. If you and @BsAtHome would rather land the package now, I'm fine with it too, just get an explicit "lcnc is our namespace going forward" on record first so it isn't decided by default.

At least for me, I rather do it the old fashioned way (lcnc_realtime.py) and do the package right in a different PR.

> One thing that matters more than the packaging: `hal.is_sim` / `hal.is_rt` were documented public attributes (in `python-hal-interface.adoc`, which this PR deletes). Hard-removing them breaks any external user HAL/python config or custom GUI reading them. I'd keep a thin deprecation shim, the names computed via the new path with a warning, rather than removing them outright.

That would mean digging out the PEP 562 module `__getattr__` again but instead of doing it in there, throw a deprecation message and call lcnc_realtime.verify(), right? I can do that.

> And if the package does stay: drop the eager `from . import realtime` in `__init__.py` so `import lcnc` doesn't pull in every future submodule; use `subprocess.run([REALTIME, "verify"], stdin=DEVNULL, stdout=DEVNULL)` instead of `os.system` (no shell, covers BsAtHome's stdin point); README is missing a trailing newline.

Thanks, I will do that. 

Just a side question. How can I have the package work the following way without an import in `__init__.py`? Dropping the package again, I won't need it anyway.
```python
import lcnc

print("realtime.verify " + str(lcnc.realtime.verify()))
print("realtime.status " + str(lcnc.realtime.status()))
```

grandixximo:
Sounds good, and thanks for opening #4198.

On the shim: yes, exactly. PEP 562 `__getattr__` in `lib/python/hal.py` (the wrapper that does `from _hal import *`), on access to `is_rt` / `is_sim` emit a `DeprecationWarning` and return `lcnc_realtime.verify()` for `is_rt` and `not lcnc_realtime.verify()` for `is_sim`. Keeping it in `hal.py` rather than the C extension is the right spot since that's where the old names lived for importers.

On the side question: with a genuinely empty `__init__.py`, `import lcnc; lcnc.realtime.verify()` will not work. A submodule isn't bound as an attribute of the package until something imports it (same reason `import xml; xml.etree...` fails until you `import xml.etree`). So the supported forms are `from lcnc import realtime` or `import lcnc.realtime` (then `lcnc.realtime.verify()`).

If you ever did want the `import lcnc; lcnc.realtime` form without eager-importing every submodule, the clean way is a lazy PEP 562 `__getattr__` in `__init__.py`:

```python
def __getattr__(name):
    if name == "realtime":
        import importlib
        return importlib.import_module(".realtime", __name__)
    raise AttributeError(name)
```

That imports on first access only. But since you're dropping the package for now, none of that applies, `lcnc_realtime.verify()` is all you need.

BsAtHome:
Right, it seems that we've settled on the (color)name of the shed ;-)

grandixximo:
Till we decide to rename it tomorrow...

hdiethelm:
So, I reverted the package, added doc and added the deprecation of is_sim / is_rt.

I'd like to stick with `lcnc_realtime.py`. `lcncrealtime.py` is hard on my eye and there are also other modules with underscore like `hal_glib.py`.

Anything more except final testing?

src/hal/halmodule.cc
hdiethelm: Resolved with 'lcnc_realtime.py'

grandixximo:
Coming out of draft?

hdiethelm:
Yes, why not. I tested stepconf / pnyconf and retested a few parts. So I call it ready.

hdiethelm:
Of course, my own test complained, fixed.
```
Processing	file:///home/runner/work/linuxcnc/linuxcnc/docs/build/html/uk/remap/remap.html


List of broken links and other issues:
file:///home/runner/work/linuxcnc/linuxcnc/docs/build/html/uk/config/python-lcnc_realtime.html	
  Line: 157
  Code: 404 File `/home/runner/work/linuxcnc/linuxcnc/docs/build/html/uk/config/python-lcnc_realtime.html' does not exist
 To do: The link is broken. Double-check that you have not made any typo,
	or mistake in copy-pasting. If the link points to a resource that
	no longer exists, you may want to remove or fix the link.
```

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -62,6 +62,232 @@
...
+        nice_s, lock_s);
+}
+
+// Success-probe for realtime scheduling: briefly try to set SCHED_FIFO on
+// the calling thread and restore the previous policy.  Succeeds when the
+// process holds CAP_SYS_NICE (file caps or setuid root) or has a matching
+// RLIMIT_RTPRIO.  Works on any kernel, so the probe also covers the
+// PREEMPT_RT-vs-stock distinction implicitly: if we can actually get
+// SCHED_FIFO, the platform can deliver realtime, regardless of how.
+// Only report an error if this check fails in master mode to not spawm the user.
```
grandixximo: "spawm the user" typo

lib/python/hal.py
```diff
@@ -29,6 +29,23 @@
 
 import _hal
 from _hal import *
+import warnings
+import lcnc_realtime
+
+def __getattr__(name):
+    if name == 'is_rt':
+        warnings.simplefilter('always', DeprecationWarning)
```
grandixximo: These `simplefilter('always'/'default', ...)` calls permanently mutate the process-global `warnings.filters` (simplefilter doesn't restore), so reading `hal.is_rt` once changes DeprecationWarning policy for the whole interpreter, e.g. it downgrades an app/pytest `error` policy back to `default`, so a deprecation that should fail tests passes quietly.

Since the goal is just visibility (DeprecationWarning is hidden by default outside `__main__`), use `FutureWarning` (shown by default) and drop the filter calls:

```python
warnings.warn(f"{name} is deprecated, use lcnc_realtime.verify() instead",
              FutureWarning, stacklevel=2)
```

`stacklevel=2` points the warning at the caller instead of `hal.py`. Non-blocking.

grandixximo (approved):

src/rtapi/uspace_rtapi_main.cc
hdiethelm: Thanks, fixed.

lib/python/hal.py
hdiethelm: Thanks, fixed. The warning looks better now.

I copied this pattern mostly from here:
https://github.com/LinuxCNC/linuxcnc/blob/9af46f663ece039994f4ec42182b45dd780334c6/lib/python/pyhal.py#L51
I guess that should be fixed to?

BsAtHome: Yeah, that'd be my (junk) code ;-)
Please fix as appropriate.

hdiethelm: Assuming the answer will be yes, I fixed and tested it:
https://github.com/LinuxCNC/linuxcnc/pull/4132/commits/ac599f1be13719bd733a1db5a54bb9f418f1dcf9

grandixximo:
Now back to #4107 what does the "you are not running realtime" need look like in the UI?

PR #4134, opened 2026-06-05
docs(build): symlink docs/html -> build/html for the website sync

grandixximo:
The unified docs build moved rendered HTML to docs/build/html/, but the website sync still uploads docs/html/, so the dev-docs link on linuxcnc.org/documents/ returns 404.

This re-exposes docs/html as a symlink to build/html during the HTML build (git-ignored, removed by make clean), so the sync keeps publishing from the old path with no buildbot or cron change.

Caveat: this only helps if the sync dereferences symlinks (rsync -L, a trailing-slash source on the symlinked dir, or cp -rL). If it copies the link verbatim it will not resolve on the server. This is the first thing to try per the mailing-list thread.

@BsAtHome @andypugh

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -4,6 +4,9 @@
 # Rendered output: HTML, PDF, and po4a-translated .adoc all land under build/.
 build/
 
+# Legacy publish path: a symlink to build/html for the website sync (Submakefile).
+/html
+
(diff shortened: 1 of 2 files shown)


grandixximo:
Will see if this works, hopefully the PDF's as well, finger crossed....

BsAtHome:
If the data gets copied to an accessible place we can always adjust the website's links.

PR #4136, opened 2026-06-06
ui-smoke phase 3: screenshots, writable config mirrors, homing settle

grandixximo:
Builds on the merged phase 2 (#4054). This phase makes the ui-smoke runs more diagnosable and adds writable-config handling so the GUIs start cleanly headless.

What's here:
- **Screenshot on failure, confirm on pass**: each GUI smoke test captures a screenshot, attached as a CI artifact, so a failure shows the actual on-screen state instead of just a log.
- **gmoccapy against a writable config mirror**: gmoccapy writes its preference file next to the config; with the workspace mounted read-only that raises a PermissionError mid-init and pops error dialogs. The test mirrors the sim config to a tmp dir and pre-seeds `hide_startup_messsage` so the "Important change(s)" modal stays hidden (the same effect as ticking "Don't show this again"). This is the headless-start fix discussed in #4072.
- **Settle after homing before requesting AUTO**: removes a race where the AUTO-mode request arrived before the motion controller finished homing.

Scope note: the touchy window-fit smoke test is intentionally held back on a separate branch, since it depends on the touchy fitting changes in #4131 which are not merged yet. I will send it as a follow-up once #4131 lands.


grandixximo:
For now the screenshots of the UI are uploaded, and it is up to the reviewers to check, if wanted we could save in source a screenshot to verify against that nothing changed, but then it will be a bit of a bother when anything is supposed to actually change, and for UI with multiple pages, we may want to verify with more than one screenshot at probably more than one resolution? that could take a while, or we can just call it done here, verify UI works with uploaded screenshot is already pretty good,

grandixximo:
Added the known-good image comparison. On a clean run each GUI's confirm.png is compared (ImageMagick) against a committed reference.png, and both the shot and a highlighted diff.png are uploaded as CI artifacts.

It never fails a test: font/freetype versions differ across distros, so we just record the diff to gather experience first. Even same-machine the 3D preview drifts a few hundred pixels, a handy sanity check.

Regenerate locally on a built RIP tree with tests/ui-smoke/_lib/make-references.sh (sets UI_SMOKE_UPDATE_REFERENCE=1). The four baselines were captured on Debian trixie; other distros will differ, which is what we want to observe. Also rebased onto master.

@BsAtHome This should fit the shape we discussed in the meeting and by email.
@NTULINUX Fedora and Gentoo testing please :-)

NTULINUX:
Fedora:

```
=== make-references: axis ===
Running test: /home/fedora/dev/linuxcnc/tests/ui-smoke/axis
Runtest: 1 tests run, 1 successful, 0 failed + 0 expected, 0 skipped, 0 shmem errors
make-references: wrote /home/fedora/dev/linuxcnc/tests/ui-smoke/axis/reference.png
=== make-references: gmoccapy ===
Running test: /home/fedora/dev/linuxcnc/tests/ui-smoke/gmoccapy
Runtest: 1 tests run, 1 successful, 0 failed + 0 expected, 0 skipped, 0 shmem errors
make-references: wrote /home/fedora/dev/linuxcnc/tests/ui-smoke/gmoccapy/reference.png
=== make-references: touchy ===
Running test: /home/fedora/dev/linuxcnc/tests/ui-smoke/touchy
Runtest: 1 tests run, 1 successful, 0 failed + 0 expected, 0 skipped, 0 shmem errors
make-references: wrote /home/fedora/dev/linuxcnc/tests/ui-smoke/touchy/reference.png
=== make-references: qtdragon ===
Running test: /home/fedora/dev/linuxcnc/tests/ui-smoke/qtdragon
Runtest: 1 tests run, 1 successful, 0 failed + 0 expected, 0 skipped, 0 shmem errors
make-references: wrote /home/fedora/dev/linuxcnc/tests/ui-smoke/qtdragon/reference.png
```

Have yet to test on Gentoo, still battling networking issues.

BsAtHome:
I don't really understand the diff images. They have a gray tint and reveal the actual image. I expected them to be nearly or completely black. A (biased) difference is not interesting for which direction it goes. The absolute difference would be more telling. Essentially: for each pixel and color `|b - a|`.

grandixximo:
Anything that differs is bright red, I have not looked at the latest ones yet, if they are mostly greyish means no changes

grandixximo:
Tomorrow I will do some builds, and have different screenshots so we can pick a winner, at least now it seems to be working alright, the final look it's just a matter of taste 😉

tests/ui-smoke/_lib/crashdump.sh
```diff
@@ -11,9 +11,11 @@ crashdump_arm() {
...
     export CORE_DIR
     ulimit -c unlimited 2>/dev/null || true
-    # core_pattern is global; only set it if already root. Never sudo:
-    # the suite must run unattended. Non-root falls back to a cwd "core".
-    if [ "$(id -u)" = 0 ]; then
+    # core_pattern is global; only set it if already root, never sudo
+    # (the suite must run unattended), and skip under runtests -u so a
+    # user opting out of root tweaks is respected even when invoking the
+    # suite from a root shell. Non-root falls back to a cwd "core".
+    if [ "$(id -u)" = 0 ] && [ "${RUNTESTS_NOSUDO:-0}" != "1" ]; then
```
hdiethelm: Hmm, this makes no sense, checking if we are root and then if RUNTESTS_NOSUDO not doing it.
Here, you can only be root if you use `sudo runtests` right? However, `sudo runtests` will always fail.
I would just check for `"${RUNTESTS_NOSUDO:-0}" != "1"` and then do `sudo sysctl`

Did you ever test if you get a crashdump if there is a segfault?

hdiethelm:
I missed one, you have to add gdb here to:
https://github.com/grandixximo/linuxcnc/blob/c4de595efdb60e9c3c65bf315289b32a4b39d63c/.github/workflows/ci.yml#L225-L229

Suggestion:
Try to get a test failing with a crashdump locally (Like adding a random segfault somewhere) and then push the test commit to see if the CI behaves as it should. I tried adding a segfault in rtapi_app locally but the test just freezes forever.
When creating tests, it is always good to test what happens if they fail.

The same for the image tests, break a gui on purpose and see what happens. Having tests pass is good but having them fail when there is an issue is even better... ;-)

Otherwise, the concept looks really good! Sadly I don't have the time right now to do a proper review.

grandixximo:
@BsAtHome 
with the current method this is how you would see if touchy was smaller than it is supposed to be
image
this is the absdiff method
image

this is gmoccapy current method the icon and time differ
image

this is absdiff method
image

axis current method the preview differs, the fonts I fuzzed a bit cause they were always coming up as positive all red with my reference picture, this is so when running locally you don't get all red

image

this is axis with absdiff with slightly different fonts, and the plot

image

this is with added fuzz on the absdiff method, only see plot difference

image

Both methods express the same thing in a different way, now that you see them next to each other, do you have a preference?

grandixximo:
@hdiethelm 
Thanks, good catches both.

Moved gdb as suggested.

The crashdump did work initially and was tested, but I did not run it again after recent permissions changes, my bad, I did testing again results are as follows

Failure path: I injected a real segfault into a GUI and pushed it through CI to watch the actual runners. Good news: it fails fast (the test goes red in seconds, no hang), the Python faulthandler traceback in linuxcnc.err names the crash frame, the watchdog catches it with UI_SMOKE_FAIL: linuxcnc exited before the driver finished, and the failure screenshots still upload. Result line was 291 tests run, 290 successful, 1 failed.

Your freeze was most likely the rtapi_app path: when the RT loader dies, linuxcnc hangs bringing HAL up rather than the GUI process exiting, so the launcher-PID watch never fires and only the outer timeout ends it, slowly. The GUI-crash path, which is what the crash dump is for, behaves.

Testing it again also exposed and fixed two crash-dump bugs: it could pick up a stale root-owned core from /tmp (gdb then just printed Permission denied), and a non-root run produces no core at all since we will not sudo to set core_pattern. Fixed: only use a readable core from this run, otherwise say so and lean on the Python traceback. The native gdb backtrace stays a best-effort extra for when the suite runs as root.

One note from the CI run: because the runners execute the suite non-root, the native backtrace path correctly no-ops and prints no readable core dump; see the Python traceback. The Python traceback is the crash signal CI actually delivers, and it does.

BsAtHome:
Actually, I like both image diffs (without fuzz) when viewed together. Do we need to choose or can we have them both?

grandixximo:
Shouldn't be a problem to have both, can even have multiples fuzz levels, the text rendering I think goes away around 20% fuzz with no fuzz the text will always be almost all red, so it will be harder to spot if any labels were changed

grandixximo:
Instead of having them as separate diffs, I could collage everything in a single huge image, with tiles, where each tite is titled sensibly, like original and build next to each other on top, then more tiles for the "a-b diff" and the "red marks the diff" at 0 and 20% fuzz, 6 tiles, 2x3, I think that should give a proper overview, or am I overdoing it?

BsAtHome:
Combining them in one image will make the lighter parts hide the darker parts (eye accommodation problem). Therefore, having them separate is best.

The small text rendering differences are not bright red as long as they are just rendering differences. A real text change would make that much more visible. I think it be good to see small rendering changes because it says something about stability. Therefore, no fuzz would be fine.

If we are to look at more differences, then we probably should look at differences between distributions. However, this is not a ui-smoke test problem. This would more be like a "let me see what I broke yesterday" matrix comparison. You'd need to collect images from all distros/versions and then be able to do a local RIP comparison. A different project ;-)

hdiethelm:
> @hdiethelm Thanks, good catches both.
> 
> Moved gdb as suggested.

Now gdb is installed with apt-get build-dep right? This just means everybody that builds linuxcnc will have gdb installed. But gdb is anyway needed to run the test, so fine in my opinion. An alternative would be to add it to the two places in the CI. If anyone has an issue with this, I can send you a patch to install it CI only.

> 
> The crashdump did work initially and was tested, but I did not run it again after recent permissions changes, my bad, I did testing again results are as follows
> 
> Failure path: I injected a real segfault into a GUI and pushed it through CI to watch the actual runners. Good news: it fails fast (the test goes red in seconds, no hang), the Python faulthandler traceback in linuxcnc.err names the crash frame, the watchdog catches it with UI_SMOKE_FAIL: linuxcnc exited before the driver finished, and the failure screenshots still upload. Result line was 291 tests run, 290 successful, 1 failed.
> 
> Your freeze was most likely the rtapi_app path: when the RT loader dies, linuxcnc hangs bringing HAL up rather than the GUI process exiting, so the launcher-PID watch never fires and only the outer timeout ends it, slowly. The GUI-crash path, which is what the crash dump is for, behaves.
> 

Possibly I was not patient enough.

> Testing it again also exposed and fixed two crash-dump bugs: it could pick up a stale root-owned core from /tmp (gdb then just printed Permission denied), and a non-root run produces no core at all since we will not sudo to set core_pattern. Fixed: only use a readable core from this run, otherwise say so and lean on the Python traceback. The native gdb backtrace stays a best-effort extra for when the suite runs as root.
> 
> One note from the CI run: because the runners execute the suite non-root, the native backtrace path correctly no-ops and prints no readable core dump; see the Python traceback. The Python traceback is the crash signal CI actually delivers, and it does.

You can use sudo in CI, it has nopasswd set. Changing:
`if [ "$(id -u)" = 0 ] && [ "${RUNTESTS_NOSUDO:-0}" != "1" ]; then`
to
`if [ "${RUNTESTS_NOSUDO:-0}" != "1" ]; then`
should be fine for users as much as I understood the tests readme and it should work in CI.
The ` [ "$(id -u)" = 0 ]` check makes no sense, tests can not be run as root anyway because linuxcnc checks for root and refuses to start if you start it as root. Or did you manage to do `sudo runtests ...`?

grandixximo:
You're right, the `id -u = 0` check was pointless since the suite never runs as root, so `core_pattern` was never armed and the native backtrace never ran. Dropped the check and set it via sudo, still gated by `RUNTESTS_NOSUDO` for `runtests -u`.

I tested the full path on a real host with passwordless sudo: a GUI segfault now writes a core into our dir and the crashdump prints the gdb backtrace (the crash frame, ffi_call, g_main_loop_run, gtk_main) plus all-threads. So CI should now produce a native backtrace instead of the "no readable core dump" fallback. (In a docker container `/proc/sys` is read-only so the arm no-ops and it falls back to the Python traceback, which is fine.)

One open question for you: `crashdump_arm` runs on every test to be ready for a crash, so it leaves `core_pattern` pointed at our (then-deleted) temp dir after the suite. On CI that's ephemeral and a non-issue, but a local dev running the full suite without `-u` would have their `core_pattern` altered until reboot. Worth a second `sudo sysctl` to save it in arm and restore it in report, or do you think that's overkill?

hdiethelm:
> You're right, the `id -u = 0` check was pointless since the suite never runs as root, so `core_pattern` was never armed and the native backtrace never ran. Dropped the check and set it via sudo, still gated by `RUNTESTS_NOSUDO` for `runtests -u`.
> 
> I tested the full path on a real host with passwordless sudo: a GUI segfault now writes a core into our dir and the crashdump prints the gdb backtrace (the crash frame, ffi_call, g_main_loop_run, gtk_main) plus all-threads. So CI should now produce a native backtrace instead of the "no readable core dump" fallback. (In a docker container `/proc/sys` is read-only so the arm no-ops and it falls back to the Python traceback, which is fine.)
> 
> One open question for you: `crashdump_arm` runs on every test to be ready for a crash, so it leaves `core_pattern` pointed at our (then-deleted) temp dir after the suite. On CI that's ephemeral and a non-issue, but a local dev running the full suite without `-u` would have their `core_pattern` altered until reboot. Worth a second `sudo sysctl` to save it in arm and restore it in report, or do you think that's overkill?

The `crashdump_arm`  might not by nice for some people, not finding their crasdumps any more.  Probably more an edgecase but who knows. At least for me, I would never run the tests allowing sudo except in a VM.

An other option: Add a new option in [scripts/runtests.in](https://github.com/LinuxCNC/linuxcnc/pull/4136/changes#diff-b1921f6a375d476a3e45a017774141377cd6d33a3486d6fe12157e148c26909e) that enables crashdumps like `d) export ENABLE_CRASHDUMPS=1 ;;` and add `-d` in the three places in ci.yml:  `runtests -p` -> `runtests -p -d`

Looking at your crashdump script: Instead of having it exclusive for your UI tests: Why not enable it globally for all tests in `runtest.in`? It looks useful for everything. It might be possible to add it just in the `run_test()` function.

grandixximo:
Good call, done. Crash dumps are now opt-in behind `runtests -d` (`ENABLE_CRASHDUMPS=1`), off by default, and I pass `-d` in the three CI invocations. Without `-d`, `crashdump_arm`/`report` no-op: no sudo, no `core_pattern` change, so a normal local run leaves the system untouched. The Python faulthandler traceback is independent of all this and stays on, so you still get a crash signal without `-d`.

On enabling it globally for every test in `run_test()`: I agree it is useful beyond the GUI tests, but I would rather do that as a follow-up PR to keep this one focused. There is also one thing to generalize first: the native backtrace currently assumes the crashing process is python3 (`gdb $(command -v python3) $core`) and the readable Python traceback comes from `PYTHONFAULTHANDLER`, which the GUI launch env sets. Most non-GUI tests are C (`halcmd`, `rtapi_app`, comps), so a global version needs to detect the crashing binary rather than assume python3. Happy to take that on separately once this lands.

hdiethelm:
> Good call, done. Crash dumps are now opt-in behind `runtests -d` (`ENABLE_CRASHDUMPS=1`), off by default, and I pass `-d` in the three CI invocations. Without `-d`, `crashdump_arm`/`report` no-op: no sudo, no `core_pattern` change, so a normal local run leaves the system untouched. The Python faulthandler traceback is independent of all this and stays on, so you still get a crash signal without `-d`.

Looks good I would say. Just one more thing: If crashdumps are not enabled, you return immediately. If sudo is not allowed, you create a new tmp directory and just hope that the crashdumps magically land in there. Is there any way this can work?

I would remove the RUNTESTS_NOSUDO check entirely and instead generate and error in `runtests.in` if someone uses -d and -u at the same time. Requiring crasdumps while not allowing sudo is just nothing that makes sense.

> 
> On enabling it globally for every test in `run_test()`: I agree it is useful beyond the GUI tests, but I would rather do that as a follow-up PR to keep this one focused. There is also one thing to generalize first: the native backtrace currently assumes the crashing process is python3 (`gdb $(command -v python3) $core`) and the readable Python traceback comes from `PYTHONFAULTHANDLER`, which the GUI launch env sets. Most non-GUI tests are C (`halcmd`, `rtapi_app`, comps), so a global version needs to detect the crashing binary rather than assume python3. Happy to take that on separately once this lands.

Ok, agreed from my side.

grandixximo:
Fixed: `runtests` now errors if `-d` and `-u` are combined, and `crashdump_arm` always arms under `-d` (dropped the dead `RUNTESTS_NOSUDO` check). Global crash dumps in `run_test()` as a separate PR once this lands.

hdiethelm:
> Fixed: `runtests` now errors if `-d` and `-u` are combined, and `crashdump_arm` always arms under `-d` (dropped the dead `RUNTESTS_NOSUDO` check). Global crash dumps in `run_test()` as a separate PR once this lands.

Nice, looks clean now.

All this implicit things, when this condition is fulfilled, this happens automatically gives a headache for debugging and users. How it is now, when I want creashdumps -d and when i don't allow sudo at the same time, I get an error is transparent and easy to understand.

tests/ui-smoke/_lib/crashdump.sh
```diff
@@ -1,45 +1,66 @@
...
-        else
-            echo "(gdb unavailable; core left at $core)"
-        fi
+        # "bt" first: gdb auto-selects the faulting thread on a SIGSEGV
+        # core. "thread apply all bt" after gives the rest.
+        gdb -batch -nx \
+            -ex "bt" \
+            -ex "echo \n=== all threads ===\n" \
+            -ex "thread apply all bt" \
+            "$(command -v python3)" "$core" 2>&1 | head -400
```
BsAtHome, line 60: Aren't all modern systems using systemd and coredumpctl? You don't see the core dump and need to run `coredumpctl debug` to access/debug it.

tests/ui-smoke/_lib/crashdump.sh
```diff
@@ -1,45 +1,66 @@
...
-    # the suite must run unattended. Non-root falls back to a cwd "core".
-    if [ "$(id -u)" = 0 ]; then
-        sysctl -w "kernel.core_pattern=$CORE_DIR/core.%e.%p" >/dev/null 2>&1 || true
-    fi
+    # Point the global core_pattern at our writable dir so the kernel writes
+    # a core we can read. -d implies sudo is allowed (runtests rejects -d with
+    # -u), and tests never run as root, so no id check; if sudo still fails the
+    # core lands wherever the system pattern says and we fall back to the
+    # Python traceback.
+    sudo sysctl -w "kernel.core_pattern=$CORE_DIR/core.%e.%p" >/dev/null 2>&1 || true
```
BsAtHome: It is problematic to have `sudo` in here.

BsAtHome:
Two (extra) points:
* You should probably rebase this after all the recent changes
* How do the images hold up with the changed glcanon code

grandixximo:
All review points addressed:

**Rebase.** Done, the branch is on current master (which also picks up the merged GSTAT fix #4399 and the glcanon background fix #4402), force-pushed as 7756bda24b.

**How the images hold up with the glcanon rewrite.** touchy: pixel-exact. axis: where G0 and G1 lines overlap they now render white (G1), originally they rendered blue (G0). gmoccapy: the "No Program loaded" instability is fixed (a lost GSTAT 'file-loaded' signal, fixed in #4399, plus a driver settle here for a deterministic shot). qtdragon: the rainbows are the glcanon regression #4379, fixed by #4402, so the next CI run should come back clean.

**coredumpctl.** Reworked: on a systemd host report() now collects the core via coredumpctl (list entries since arm time, dump to the work dir), no root needed for our own processes. A plain-file core_pattern and ./core in the test directory remain as fallbacks for non-systemd hosts.

**sudo.** Removed from crashdump.sh entirely. The script only raises ulimit -c and collects whatever the system produced. The kernel.core_pattern change moved into ci.yml as a job-level step on the disposable runner, so the privileged part is visible in the workflow and scoped to CI.

---

One takeaway from the screenshot comparison: had these committed references existed before the glcanon 3.3 rewrite, the qtdragon rainbow regression (#4379) would have shown up in CI immediately as a large diff against the reference, instead of being found later by manual testing. The comparison never fails a test (freetype versions differ across distros, some drift is expected), but the diff artifacts make this class of rendering regression visible at review time.

It also catches the small stuff: the axis diff is just the overlapping G0/G1 lines now rendering white (G1) instead of blue (G0), exactly the kind of line-level difference that slips past the eye but belongs on record when preview rendering changes.

grandixximo:
Also added a fix for a flaky run I spotted, ready to merge

tests/ui-smoke/_lib/drive.py
```diff
@@ -32,6 +32,13 @@
 # stability check catches that.
 STATE_STABILITY_S = 0.5
 STATE_RETRY_BUDGET = 6
+# Pause after homing before requesting AUTO. gmoccapy only enables AUTO
+# once it has processed the all-homed signal in its own event loop (and
+# re-asserts MANUAL itself on that signal). Requesting AUTO before then is
+# rejected: it bounces back to MANUAL with an "It is not possible to
+# change to Auto Mode" warning. ensure_mode would retry and win, but the
+# warning lingers on screen; this settle lets the GUI catch up first.
+POST_HOME_SETTLE_S = 2.0
```
BsAtHome, line 43: Isn't this racy? To be sure, you need to know the state of the GUI. Waiting for a certain amount of time may never be enough.

The same problem occurs other places too. Time is a very volatile quantity in CI.

The only way to solve the problem is to have one or more pins or params set by the GUI that represents the state of the GUI.

grandixximo: Fair point, and I agree time is volatile in CI. How the driver is layered: correctness never rests on the sleeps, it rests on polling stat plus retry budgets. The fixed pauses only protect the screenshots: without the post-homing pause the AUTO request bounces off gmoccapy's not-yet-updated view, ensure_mode retries and wins, the test passes, but the "It is not possible to change to Auto Mode" warning lingers in the confirm shot. Same for the 0.4 s GSTAT settle: too short just means a bogus progress percentage in the picture, not a failure.

The one place your critique bit was a GUI revert landing after the 0.5 s stability re-check and before the next command. Both such holes are now closed by retry rather than longer waits: home_all retries the whole ESTOP_RESET + ON + MANUAL + home sequence (a late gmoccapy startup ESTOP caused exactly this failure on master), and run_program retries AUTO + program_open + AUTO_RUN keyed on the state signal that the program actually started. With that, no fixed sleep in the driver is load-bearing: STATE_STABILITY_S and the settles only keep transient warning dialogs out of the screenshots, and the failure mode of any too-short pause is a cosmetic artifact, never a failed test.

On pins: no GUI currently exports an "I have processed all-homed" pin or param, and gmoccapy re-asserting MANUAL based on its own stale view is arguably the GUI-side root cause. If you want, I can look at what gmoccapy would need to expose for the driver to wait on state instead of time, but that is a GUI change beyond this PR's scope.

BsAtHome: Good that (some of) the potential race is gone :+1:

For the GUI telling outside via pins what state it is in,... we'll have that on the wish-/todo-list ;-)

grandixximo:
@BsAtHome This could be useful to test #4481 are there still blockers?

BsAtHome:
Indeed, it could be useful to see the result.

The only "blocker" I see is that it can be squashed by you or in the merge. At least, 29 commits are probably not necessary ;-)

PR #4139, opened 2026-06-08
radiobutton: fix unused-parameter build error

grandixximo:
radiobutton.comp ignores `period`, but defaults to `option period` on, so the generated FUNCTION carried an unused `long period` and broke the RT build under `-Werror=unused-parameter`. Adding `option period no;` drops the parameter.

master currently fails `rip-and-test` / `rip-and-test-clang` because of this.

@bsathome @hansu @andypugh  please merge.

### DIFF
--- a/src/hal/components/radiobutton.comp
+++ b/src/hal/components/radiobutton.comp
@@ -16,6 +16,7 @@ pin in float input-##[32:personality] "Possible output values";
 pin out float value "currently-selected output value";
 
 function _;
+option period no;
 license "GPL 2+";
 author "Andy Pugh";
 


NTULINUX:
Thank you for this!

grandixximo:
No problem, has to be merged, we have a broken tree...

NTULINUX (approved):

grandixximo:
Good Morning, Bertho :-)

BsAtHome:
@hansu broke the tree when merging 2.9 into the master branch. Any merge must be compiled from a clean slate locally before committing or you risk breaking the tree, like happened here.

hansu:
Sorry I forgot about the adaptions needed for halcomp in master. But yeah a testbuild would have been better.

hansu:
Always a pleasure doing these merges...

PR #4146, opened 2026-06-10
ui-smoke: never sudo in crash capture

### LINKED ISSUE #4143: Test ui-smoke tries to sudo and fails on OK
@grandixximo When running tests locally on Fedora 43 I was stomped to find that I was met with a `sudo` prompt. That is not a functional situation because I had to send a keyboard interrupt to continue. Tests must run without intervention and cannot do sudo when not explicitly instructed or allowed to.

Output:
```
...
Running test: /...lcnc.git/tests/ui-smoke/axis
[sudo] password for <me>: ^C
Running test: /...lcnc.git/tests/ui-smoke/gmoccapy-quit
[sudo] password for <me>: ^C
Running test: /...lcnc.git/tests/ui-smoke/gmoccapy
[sudo] password for <me>: ^C
Running test: /...lcnc.git/tests/ui-smoke/qtdragon-quit
[sudo] password for <me>: ^C
*** /...lcnc.git/tests/ui-smoke/qtdragon-quit: XFAIL: test run exited with 1
Runtest: 286 tests run, 285 successful, 1 failed + 0 expected, 3 skipped, 0 shmem errors
Failed: 
/.../tests/ui-smoke/qtdragon-quit
```

The failed test (qtdragon-quit) has this line in the `result` file output:
```
=== ui-smoke.out ===
UI_SMOKE_OK
```
I guess that means everything is fine, but the test checks for `UI_SMOKE_QUIT_OK`, which is missing. So I guess something is still wrong. However, `gmoccapy-quit` also fails to have the tag `UI_SMOKE_QUIT_OK` in the result file and seems to pass?

grandixximo:
The ui-smoke crash-capture helper (`crashdump_arm`) ran `sudo sysctl`
to set `kernel.core_pattern`, assuming passwordless sudo on CI. On a
local run without passwordless sudo, sudo prompts for a password on the
TTY and stalls the entire test suite until the user sends a keyboard
interrupt, as reported in #4143.

This drops sudo entirely: core_pattern is set (and gdb installed) only
when the suite already runs as root. Non-root runs skip it and fall
back to a cwd `core`, so the tests run unattended on every host.

Fixes #4143

Note: CI is expected to fail here for an unrelated reason. Debian sid
is currently broken, not this change.

### DIFF
--- a/tests/ui-smoke/_lib/crashdump.sh
+++ b/tests/ui-smoke/_lib/crashdump.sh
@@ -11,9 +11,11 @@ crashdump_arm() {
     CORE_DIR="$(mktemp -d -t ui-smoke-cores.XXXXXX)"
     export CORE_DIR
     ulimit -c unlimited 2>/dev/null || true
-    # core_pattern is global and needs root; best-effort (CI has sudo).
-    # If it does not take, crashdump_report still finds a cwd "core".
-    sudo sysctl -w "kernel.core_pattern=$CORE_DIR/core.%e.%p" >/dev/null 2>&1 || true
+    # core_pattern is global; only set it if already root. Never sudo:
+    # the suite must run unattended. Non-root falls back to a cwd "core".
+    if [ "$(id -u)" = 0 ]; then
+        sysctl -w "kernel.core_pattern=$CORE_DIR/core.%e.%p" >/dev/null 2>&1 || true
+    fi
 }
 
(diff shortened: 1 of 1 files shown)


grandixximo:
@BsAtHome Did you test? Good to merge?

BsAtHome:
Yes, tested and functional.

PR #4147, opened 2026-06-10
halshow: show pins and parameters in one tree

grandixximo:
Pins and parameters share one namespace, but halshow presents them as two parallel trees, so users keep looking for an item in the tree where it is not. This merges them into a single "Pins & Parameters" tree:

- pins and parameters are interleaved alphabetically under the same component prefixes; parameter leaves are drawn in the same brown as in the watch tab
- leaf node ids keep the real type (`pin+name` / `param+name`), so show, watch, popup menus and saved watchlists work unchanged
- clicking a branch shows both the pins and the parameters below it
- "Add all sub-items to watch" adds both
- a "Separate parameters" checkbox in the Tree View menu restores the old layout; the choice is saved in the preferences file

Tested with a sim config and a driven xvfb session: tree build, filter (plain and full-path regex), watch tab, Show in Tree, mode toggle, preferences persistence. Docs screenshots not updated yet; I will do that if the approach is agreed on.
image

@bsathome this follows your point that nothing prevents one tree with a uniform search, since pin and parameter names cannot collide. If pins and parameters later merge internally behind the getter/setter work, the UI is already in the right shape for it.

@rene-dev I am aware this goes in the other direction than your noparam branch: it keeps parameters and shows them where users naturally look instead of removing them. Is there a discussion to be had about which way to go, before this moves forward?


hansu:
Will this not be superfluous when we get rid of the parameters?

grandixximo:
This is an alternative to dropping the parameters, as Bertho said and wrote, parameters can coexist, the getter/setter can get us there, this solves the "I can't find them" argument, which seems to be the main case for removal?

BsAtHome:
Regardless what the ultimate fate of parameters will be. The current situation is that we have them and are a far cry from getting them removed. Fixing the presentation in halshow now is a good thing (even if it will be changed again later on).

My work on getter/setter has uncovered quite a few situations where things were (ab)used in ways as very expensive storage (due to the volatile modifier). Other places have (sometimes odd) combined normal/hal memory constructs. None of these situations can be fixed automatically (or search/replace) and need some code adaptations to work as intended. Only when the interface is clean, then removing "parameters" is possible. However, you could simply see them as pins with different attributes, which is possible too. The underlying structures will eventually be merged in the cleanup. Then there is ample time to discuss whether we should drop other attributes.

grandixximo:
Parameters let a component author state that a value was not intended to be changed in real time. I still don't see why that signal has no value, or why we should throw it away.
If "I can't find them" was the main driver for removal, this PR solves that, and parameters can stay as a sibling to pins with different implied attributes. I lean toward letting the user connect them, with the documented understanding that they are not supposed to be changed while running, which is already the status quo.
To be seen what @rene-dev has to say.

BsAtHome:
There is one more issue that can be considered a "problem". Parameters cannot be used in signals.
There is a case for having read/write parameters as drivers for input pins, which will become possible when their internal structure is unified. Parameters are the <i>"static configuration"</i>, whereas pins are the <i>"dynamic configuration"</i>. That would be a more logical approach.

c-morley:
I think it would be simpler to get rid of parameters and just use pins.

BsAtHome:
Well, if getter/setter ever gets voted in, then you can eliminate the entire parameter branch in an experiment by only changing five functions in hallib.

hansu:
>Parameters let a component author state that a value was not intended to be changed in real time. I still don't see why that signal has no value, or why we should throw it away.

But they can be changed, so its more or less only a naming...



>a "Separate parameters" checkbox in the Tree View menu restores the old layout; the choice is saved in the preferences file

There is no checkbox for the setting but I don't think we need that.

Don't get me wrong, it's fine for me to have them together in the tree.
I only think its maybe not worth the work if will not be needed in some time.

Edit: ~~But then you have also to adapt the menus~~

grandixximo:
Toggle is in tree view menu, not really a checkbox, but it does get a check mark if you click on it

hansu:
> Toggle is in tree view menu, not really a checkbox, but it does get a check mark if you click on it

In the filter/search settings? I didn't expect it there..

grandixximo:
Tree view, top menu mate

Sorry not on the computer or I'd send you a picture

grandixximo:
I'm not removing parameters, so I don't think I have to adapt the menus yet, I can keep it as draft and remove parameters menus if that's what we are going for, I still think the distinction is valuable, simplicity at the storage level does not require throwing away the concept of having pins with _parameters_ attributes

c-morley:
> I think it would be simpler to get rid of parameters and just use pins.

Yes sorry, I meant in the bigger picture on the system, which was not really what your talking of hete. Ill be quiet now :)

hansu:
> Tree view, top menu mate

Ah, found it. There I would have expected it even less :)
I suggest to move it into the settings tab.

>I'm not removing parameters, so I don't think I have to adapt the menus yet, 

Of course, my bad. Forget my comments about that.

grandixximo:
> Ah, found it. There I would have expected it even less :) I suggest to move it into the settings tab.

It is very much related to the Tree View, but no complaints about moving it, we could keep it in both places, I don't think it will hurt.

hansu:
> It is very much related to the Tree View, but no complaints about moving it, we could keep it in both places, I don't think it will hurt.

It's also a persistent setting ;-). Two places would be too much for that little setting.

hansu:
Having the option in the tree menu is fine for me but this would need an extra screenshot in the docs.
However which option you choose, it would be nice if you could update the docs accordingly.

grandixximo:
@hansu should I add a New 2.10 paragraph? I don't really like the Idea of putting new in the docs, as it can become confusing, but New 2.9 is already there, thoughts?

hansu:
> @hansu should I add a New 2.10 paragraph? I don't really like the Idea of putting new in the docs, as it can become confusing, but New 2.9 is already there, thoughts?

Then don't add it if you don't like it ;-)

To be less confusing I added the version from when it's new. But the "New in 2.9" can be removed in master.

PR #4148, opened 2026-06-10
scripts: report missing INI file before parse check

grandixximo:
I typed a wrong INI path and got "E: The INI-file contains errors that need to be fixed", which confused me; the file did not even exist.

Cause: the `inivalue` parse check runs before the file-existence check, so a missing file is reported as a broken one. This moves the existence check ahead of the parse check and drops the now-redundant later copy.

Tested in a RIP build:
- missing file: `Could not find INI file '/nothinghere.ini'`
- existing broken file: still `E: The INI-file contains errors that need to be fixed.`

@bsathome FYI, this touches the pre-run check from 1b96a45175.

### DIFF
--- a/scripts/linuxcnc.in
+++ b/scripts/linuxcnc.in
@@ -406,6 +406,14 @@ function run_applications () {
     done
 }
 
+# Make sure the INI file exists before trying to parse it. Otherwise a
+# missing file would be reported as a file with errors.
+if [ ! -f "$INIFILE" ] ; then
+    echo "Could not find INI file '$INIFILE'"
+    trap '' EXIT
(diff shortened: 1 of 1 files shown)


grandixximo:
sid still broken, incredible annoying...

BsAtHome:
This is a good catch. Checking the file exists is obvious but needs to be done in the right place.

And, yes, Sid's breakage is annoying. It seems that they upgraded libadios2 to 2.12 and the dependencies have not been updated/rebuild/uploaded and are still on 2.11.

grandixximo:
@BsAtHome merge?

PR #4153, opened 2026-06-11
Remove .comp files from docs/src

### LINKED ISSUE #4145: Remove .comp files from docs/src
There are two .comp files in `docs/src/hal` which are also in `comp.adoc` as examples for halcompile.
`out8.comp` can be safely deleted but `rand.comp` is used in tests:

https://github.com/LinuxCNC/linuxcnc/blob/e90c66dac743dc0bf9f827f70fab5cdd8898c662/tests/Submakefile#L19-L21

I think it does not belong in docs/ and should be moved to tests/  or somewhere else.

grandixximo:
Both files duplicate examples already inlined in docs/src/hal/comp.adoc and are not part of the docs build.

- rand.comp is only used to generate the halcompile/userspace test input, so I moved it next to that test and updated tests/Submakefile.
- out8.comp is referenced nowhere, so I deleted it.

Verified with a full RIP build; all 9 tests under tests/halcompile/ pass.

Fixes #4145


grandixximo:
Should contact someone upstream for sid?

grandixximo:
@hansu patch ready to merge

PR #4160, opened 2026-06-13
motion: wait for spindle at-speed after M5 spindle stop

### LINKED ISSUE #4158: `spindle-at-speed` is not waited for when disabling the spindle
I do have something like this as G-Code:

```
M5
G91
G1 Z+0.001
G38.3 Z-0.5
```

regardless of what the `spindle0.at-speed` pin gets as signal, the moves after `M5` begin immediately (also regardless of whether the `G1` is there or not). I would like to avoid having a spinning tool hit my toolprobe; so I think there is legitimate reason to also consider `at-speed` when the spindle is off and or the commanded spindle speed is 0.

Actually I would expect `G38.n` to form a `at-speed` barrier by itself, without needing to resort to the hacky `G1` above to wait for the spindle to be at speed. Attached is a halscope log from the following actions:

- Enable spindle
- Start the G-Code listed above

As can be seen, the Z movement starts immediately after the `M5`, despite `at-speed` lagging almost 10 seconds behind.

[halscope.csv](https://github.com/user-attachments/files/28891942/halscope.csv)

grandixximo:
### Problem

When a program stops the spindle (M5) and then makes a feed or probe move, motion starts immediately without waiting for the spindle to actually reach its commanded state. As reported in #4158, this is a safety hazard: a `G38.n` probe can begin while the tool is still spinning down, so a still-rotating tool can contact the probe.

The at-speed barrier was only ever armed by M3/M4 (spindle start). M5 went through `STOP_SPINDLE_TURNING` and never set the `atspeed_next_feed` flag, so the next feed move was never gated on `spindle.N.at-speed`.

### Change

M5 now arms the same at-speed barrier that M3/M4 use. The next feed-type move (G1/G2/G3, and probing G38.n) waits until `spindle.N.at-speed` reflects the stopped state before it starts. Rapids (G0), jogs and tool-change traverses are unaffected, matching existing M3/M4 behavior.

`wait_for_spindle_at_speed` is plumbed through the existing path: `EMC_SPINDLE_OFF` carries the flag, `STOP_SPINDLE_TURNING` sets it, `emcSpindleOff` forwards it to motion, and `EMCMOT_SPINDLE_OFF` sets `atspeed_next_feed`.

### Safety of the default

This is on by default, and that is safe because of how the at-speed pin is initialized. `spindle.N.at-speed` defaults to 1 (motion.c): when the pin is left unwired, its value stays 1, so the barrier is satisfied instantly and motion is never blocked. Only configs that wire `spindle.N.at-speed` to reflect actual spindle state get the new wait, which is exactly the behavior the issue asks for.

### Guarding against a stuck wait

The one remaining risk is a config that wires `at-speed` so it never goes true while the spindle is stopped. There, the post-M5 barrier would hold motion and the operator would see an idle machine with no obvious cause. To make that self-explaining:

On abort, if motion was being held waiting for at-speed after a spindle stop, I emit a loud operator error naming the spindle and pointing at `spindle.N.at-speed`, and I clear the pending barrier so it cannot strand a later move. So if a misconfigured at-speed signal ever holds the machine, the abort the operator naturally reaches for is the thing that tells them why.

### Scope and limitation

The barrier only holds feed-type and probing moves. A `G0` move toward a probe after M5 is still not gated, since rapids are never inhibited by at-speed. This matches existing M3/M4 semantics and is called out here so it is not mistaken for full protection.

### Testing

Built clean (`make` in `src/`, uspace). Verified the at-speed pin default and the abort path by inspection; no functional change for configs without `at-speed` wired.

### Note

@bsathome the indentation in command.c is a nightmare to work with (mixed tabs and spaces throughout). I tried my best to match the surrounding tab style in the abort block; let me know if it looks right on your end.

### DIFF
--- a/src/emc/nml_intf/canon.hh
+++ b/src/emc/nml_intf/canon.hh
@@ -639,7 +639,7 @@ This is usually given in rpm and refers to the rate of spindle
 rotation. If the spindle is already turning and is at a different
 speed, change to the speed given with this command. */
 
-extern void STOP_SPINDLE_TURNING(int spindle);
+extern void STOP_SPINDLE_TURNING(int spindle, int wait_for_atspeed = 1);
 
 /* Stop the spindle from turning. If the spindle is already stopped, this
 command may be given, but it will have no effect. */
--- a/src/emc/motion/command.c
+++ b/src/emc/motion/command.c
@@ -1514,6 +1514,14 @@ void emcmotCommandHandler_locked(void *arg, long servo_period)
                 }
             }
 
+	    /* A probe is a feed-type move, so honor any pending at-speed
+	       barrier just like G1: wait for spindle.N.at-speed (spin-up after
+	       M3, or stop after M5) before the probe starts. */
+	    if (emcmotStatus->atspeed_next_feed) {
+		issue_atspeed = 1;
+		emcmotStatus->atspeed_next_feed = 0;
+	    }
+
 	    /* append it to the emcmotInternal->coord_tp */
 	    tpSetId(&emcmotInternal->coord_tp, emcmotCommand->id);
 	    if (-1 == tpAddLine(&emcmotInternal->coord_tp,
@@ -1524,7 +1532,7 @@ void emcmotCommandHandler_locked(void *arg, long servo_period)
 				emcmotCommand->acc,
 				emcmotCommand->ini_maxjerk,
 				emcmotStatus->enables_new,
-				0,
+				issue_atspeed,
 				-1,
 				emcmotCommand->tag)) {
 		reportError(_("can't add probe move"));
--- a/src/emc/motion/command.c
+++ b/src/emc/motion/command.c
@@ -507,6 +507,22 @@ void emcmotCommandHandler_locked(void *arg, long servo_period)
 	    if (GET_MOTION_TELEOP_FLAG()) {
                 axis_jog_abort_all(0);
 	    } else if (GET_MOTION_COORD_FLAG()) {
+		/* If motion was being held waiting for spindle at-speed, the
+		   operator most likely aborted because the machine sat idle
+		   with no obvious reason. Tell them why, loudly. A wait that
+		   never released after a spindle stop almost always means
+		   spindle.N.at-speed is not wired, or does not go true once the
+		   spindle has actually stopped. */
+		if (MOTION_ID_VALID(emcmotInternal->coord_tp.spindle.waiting_for_atspeed)) {
+			for (spindle_num = 0; spindle_num < emcmotConfig->numSpindles; spindle_num++) {
+				if (emcmotStatus->spindle_status[spindle_num].state == 0 && !emcmotStatus->spindle_status[spindle_num].at_speed) {
+					reportError(_("Aborted while waiting for spindle %d at-speed after a spindle stop. "
+						"Check that spindle.%d.at-speed is connected and goes true once the spindle has stopped."),
+						spindle_num, spindle_num);
+					break;
+				}
+			}
+		}
 		tpAbort(&emcmotInternal->coord_tp);
 	    } else {
 		for (joint_num = 0; joint_num < ALL_JOINTS; joint_num++) {
@@ -530,6 +546,8 @@ void emcmotCommandHandler_locked(void *arg, long servo_period)
 		SET_JOINT_FAULT_FLAG(joint, 0);
 	    }
 	    emcmotStatus->paused = 0;
+	    /* Drop any pending at-speed barrier so it can't strand a later move. */
+	    emcmotStatus->atspeed_next_feed = 0;
 	    // Clear pins on abort so tests see a clean state
 		if (emcmot_hal_data) {
 				*(emcmot_hal_data->interp_arc_radius) = 0.0;
@@ -1702,6 +1720,10 @@ void emcmotCommandHandler_locked(void *arg, long servo_period)
 	        *(emcmot_hal_data->spindle[n].spindle_orient) = 0;
 	        emcmotStatus->spindle_status[n].orient_state = EMCMOT_ORIENT_NONE;
         }
+        /* Optionally make the spindle stop an at-speed barrier: the next feed
+           move waits until at-speed reflects the stopped state. Safe when
+           unwired since at-speed defaults to 1 (see motion.c). */
+        emcmotStatus->atspeed_next_feed = emcmotCommand->wait_for_spindle_at_speed;
 	    break;
 
 	case EMCMOT_SPINDLE_ORIENT:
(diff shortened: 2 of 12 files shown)


hansu:
Can this be made optional?
Normally you would turn off the spindle when your part is finished and then move the spindle away from the part so that you can safely remove the part from the vice/table. 
With this change you cannot move away if the spindle is still spinning. So it's is a bit ineffective.

grandixximo:
Move away in what way exactly?

Are you saying you have a gcode that is something like

```gcode
G0 Z0
M3S1000
G1 Z-100 F100
G1 X100
M5
G1Z-80
G0Z0
```
and the G1Z-80 will be held up? You stop the spindle in between G1 moves, and you are expecting the move to happen as the spindle slows down?

c-morley:
You could fix that with a smarter comparison module. By checking motion type. If Probe then set at speed false, if not then set true.

Or make this behavior only work with Probe commands.

grandixximo:
I don't believe the kind of gcode I posted is valid, a move in G1 after M5, really? M5 truncates the TP so you have a full stop at that point anyhow, that is not the right way to write a G1 retraction, or is that on purpose? Just trying to understand if that's a real use case, you guys are leveraging the fact that it would let you move in G1 before the spindle would actually stop?

c-morley:
True, I was thinking G0 but it gnores the spindle at speed command.

grandixximo:
Correct

> Rapids (G0), jogs and tool-change traverses are unaffected, matching existing M3/M4 behavior.

This is only about "the spindle should be at the speed I set, when I do a feed move", we already have clear separation feed move vs rapid moves for inhibit, I think doing another split for probing only for spindle-at-speed only when slowing down is just heck confusing.

The behavior I'm proposing is explained in a sentence, and is IMO the expected behavior, and what it should have been from day one.

hansu:
More like this, bit if it does not affect G0 then it's fine

``` gcode
G0 X0 Y0 Z10
M3 S1000
G1 Z-10 F100
G1 X100
G1 Z0
M5
G0 Z10
G0 X-100 Y-100 Z100
```

wucke13:
@grandixximo do you plan to backport this to 2.9.8 too? The patch does unfortunately not cleanly apply to 2.9.8 AFAICT

wucke13:
I did the backport myself, at least a flaky version of it. See https://github.com/LinuxCNC/linuxcnc/pull/4161/changes for the details.


Using that (https://github.com/LinuxCNC/linuxcnc/pull/4161), I now get the following behavior:

- waits for spindle to be stopped before the `G1`
  ```gcode
  M5
  G91
  G1 Z-0.0001
  G38.2 Z-5
  G90
  ```
- does not wait for spindle to be stopped before the `G1`
  ```gcode
  M5
  G91
  G1 Z0
  G38.2 Z-5
  G90
  ```
- does not wait for spindle to be stopped before the `G38.2`
  ```gcode
  M5
  G91
  G38.2 Z-5
  G90
  ```

This is IMHO better, but still not satisfactory; I think that `G38.n` should behave exactly like G1 with regard to the waiting for `spindle.at-speed`. However, maybe also my poor backport did break `G38.n` waiting?

grandixximo:
@wucke13 good catch, and your backport wasn't the problem. The `G38.n` behavior you saw is the original code in both 2.9 and master.

Probe moves don't go through the same path as `G1`. They use `EMCMOT_PROBE`, which passed a hardcoded `0` for the at-speed flag to `tpAddLine`, so a probe never waited for `spindle.N.at-speed` regardless of the spindle state. I had assumed it would wait if you spun the spindle up and it wasn't yet at speed, but that was never actually the case; the probe path ignored at-speed entirely.

I pushed a second commit that makes `EMCMOT_PROBE` honor the barrier like `G1` does, so `G38.n` now waits for spin-up after `M3`/`M4` and for stop after `M5`. Should work now. Could you re-test on the updated PR? Same one-spot change applies to 2.9.8 if you want it there too.

grandixximo:
Was checking if a documentation change was needed here, and actually the new behavior matches the current docs specified behavior 

---

spindle.M.at-speed IN BIT

Motion will pause until this pin is TRUE, under the following conditions: before the first feed move after each spindle start or speed change; before the start of every chain of spindle-synchronized moves; and if in CSS mode, at every rapid->feed transition.

---

It's perfectly reasonable to assume M5 causes a spindle speed change, and G38 is a feed move, therefore nothing to change in the docs, these are just edge cases that were not caught during the initial implementation.

src/emc/nml_intf/canon.hh
```diff
@@ -639,7 +639,7 @@ This is usually given in rpm and refers to the rate of spindle
 rotation. If the spindle is already turning and is at a different
 speed, change to the speed given with this command. */
 
-extern void STOP_SPINDLE_TURNING(int spindle);
+extern void STOP_SPINDLE_TURNING(int spindle, int wait_for_atspeed = 1);
```
BsAtHome, line 643: The default here is `wait_for_atspeed=1`. The default in emc.hh:426 is `extern int emcSpindleOff(int spindle, int wait_for_atspeed = 0);`

How does this relate?

BsAtHome:
> the indentation in command.c is a nightmare to work with

Yes, monsters are lurking (under the bed scaring us when we're awake)... There will be a time when we want to fix it without muddying the blame view.

src/emc/nml_intf/canon.hh
grandixximo: Two layers, defaults match their use:

- `STOP_SPINDLE_TURNING` (canon, default 1) is called by the interpreter for a programmed stop (M5, program end, tool change). 1 arms the barrier: that's the feature.
- `emcSpindleOff` (task, default 0) is only called bare by `emcSpindleAbort()` and the estop/abort cleanup. Those must not arm a barrier, so 0.

The M5 path never uses either default: the flag is passed explicitly all the way. Interp's `STOP_SPINDLE_TURNING(s)` sets `EMC_SPINDLE_OFF.wait_for_spindle_at_speed = 1`, and emctaskmain forwards it into `emcSpindleOff(spindle, msg->wait_for_spindle_at_speed)`. So `1` governs programmed M5, `0` governs only abort/estop.

wucke13:
> Yes, monsters are lurking (under the bed scaring us when we're awake)... There will be a time when we want to fix it without muddying the blame view.

Might I suggest that the LinuxCNC project adopts https://github.com/numtide/treefmt? That would provide to have a single tool to be called, that then in turn would call the actual formats with exactly the same arguments/configuration on every devs machine. That is what we use in our bigger projects to make sure everything is canonically formatted.

I would consider to format all the c/cpp/h/python source code. Of course the adoc, the XML derivates, _mostly everything_ automatically. Of course the change should be made incrementally.

wucke13:
> Should work now. Could you re-test on the updated PR? Same one-spot change applies to 2.9.8 if you want it there too.

I applied https://github.com/LinuxCNC/linuxcnc/pull/4160/changes/344395efd9ea70de91be8435d2d35bf775a4c8d0 over in https://github.com/LinuxCNC/linuxcnc/pull/4161, and it does work as intended with

```gcode
M5
G38.3
```

PR #4161, opened 2026-06-13
chore: backport of PR #4160

wucke13:
This is basically a clipped down version of
203abbf46f9c7d12b737d2523ec1432f1b721561 and 344395efd9ea70de91be8435d2d35bf775a4c8d0.

@grandixximo please verify carefully, I just blindly removed things from the aforementioned commit until when compilation ran through without errors.


grandixximo:
@wucke13 the `tests/interp/compile` test (`use-rs274.cc`) still has the old 1-arg `STOP_SPINDLE_TURNING` stub, so it fails to link against the 2-arg signature. That test is disabled on master but enabled on 2.9, which is why only your backport hit it.

Fix in `tests/interp/compile/use-rs274.cc`:
```c
void STOP_SPINDLE_TURNING(int spindle, int wait_for_atspeed) {}
```

src/emc/motion/command.c
```diff
@@ -1452,6 +1470,14 @@ void emcmotCommandHandler_locked(void *arg, long servo_period)
                 }
             }
 
+	    /* A probe is a feed-type move, so honor any pending at-speed
+	       barrier just like G1: wait for spindle.N.at-speed (spin-up after
+	       M3, or stop after M5) before the probe starts. */
+	    if (emcmotStatus->atspeed_next_feed) {
+				issue_atspeed = 1;
+				emcmotStatus->atspeed_next_feed = 0;
```
grandixximo: you have one tab too many on these, code-base rules tab=8spaces

andypugh:
I think this is too close to changing existing machine behaviour (admittedly for the better) for me to be comfortable merging it this close to the next release. 
But I will probably merge it after the release so that we get a bit more experience with it.

src/emc/motion/command.c
wucke13: Thanks, adjusted accordingly. That indentation in the file is _wild_.

grandixximo:
pull 2.9 and merge it, then push again, so CI will pass, there was a small bug for debian sid, now is fixed, so everything will turn green :-)

PR #4162, opened 2026-06-14
docs: drop 'New in 2.9' framing from halshow page

grandixximo:
Removes the "New in 2.9" admonition blocks from the halshow page and keeps their text as plain prose. The docs should describe how things work now; the old-vs-new framing gets confusing as features keep changing.


grandixximo:
@hansu looks good?

Issue #4159, opened 2026-06-12
impossible to get absolute machine coordinates from `G38.n` without erasing TLO


wucke13:
I have a use-case where I want to get the touch-off from a `G38` in absolute machine coordinates. The problem that I have is the following: `G38` only stores the machine coordinates in the relative current working coordinate system. This might, or might not include the tool offset (TLO), depending i.e. on whether `G43` was called after the last `Tn M6`.

The `G38.2` example (https://linuxcnc.org/docs/html/gcode/g-code.html#gcode:g38) cleverly avoids this pitfall by issuing a `G49` before the `G38.2`, which however erases the current TLO. At first I tried to just incorporate the TLO in the formula found in that example, to this extent:

```gcode
(                   G53+n Z  n             G92 Z   G92?    TLO Z)
#<z_work_offset> = [#[5203 + #5220 * 20] + #5213 * #5210 - #5403]
```

However, this formula is only correct half the times, depending on whether the TLO Z is also applied (via a `G43`) or not. For other optional coordinate system offsets, there is a status property (i.e. `#5210` for `G92`/`G52`, https://linuxcnc.org/docs/html/gcode/overview.html#sub:numbered-parameters) that tell us whether this offset is actually applied.

Unfortunately, as far as I can see, none such parameter is available to announce whether TLO is applied.

andypugh:
It's not impossible, but it is hacky and unsatisfying. (I figured this out a long time ago as a newbie and have forgotten that it's a terrible hack)

G30.1 - stores the current absolute position into parameters 5181-5186.

wucke13:
And then you suggest comparing the current absolute parameters to `5420`-`5428` + `work_offset` to figure out whether TLO is applied?

If so, I agree, this most definitely is deserving of the attribution "unsatisfying". And it heats up fears around floating point arithmetic quirks on my end at least.

andypugh:
No, if you run a G30.1 you can then read out the current absolute machine coordinates from #5181-5186. Which I think is what you said you wanted?

The hack is that this is not what G30 is _for_.

wucke13:
I see. Do I get exactly the same coordinates from `#5181`-`#5186` after the `G38.n` as I would get from `#5061`-`#5069` (modulo the current work coordinate system offsets of course)?

[...]

PR #4163, opened 2026-06-14
docs: G38 probe to machine-coordinate conversion via #5401-#5409

grandixximo:
Documents the official way to recover the absolute machine coordinate of a probe from the result parameters `#5061`-`#5069`.

Since #3996, `#5401`-`#5409` report the tool length offset actually applied to motion (set by the G43 family, zeroed by G49). That makes the probe-to-machine conversion reliable whether or not a tool offset is active, so it no longer requires cancelling the offset with G49 first as the existing G38.2 example does.

Added under the `G38._n_` straight probe section:

```ngc
G38.2 Z-100 F100
#<zmachine> = [#5063 + #[5203 + #5220 * 20] + #5213 * #5210 + #5403]
```

Refs #4159

### DIFF
--- a/docs/src/gcode/g-code.adoc
+++ b/docs/src/gcode/g-code.adoc
@@ -1264,6 +1264,31 @@ G10 L1 P#5400 Z#<zworkoffset> (set new tool offset)
 G43
 ----
 
+[[gcode:probe-machine-coordinates]]
+Probe results in #5061 to #5069 are expressed in the current work
+coordinate system, so they include the active coordinate system origin,
+any G92/G52 offset, and the applied tool length offset.
+To recover the absolute machine coordinate of a probe, add those offsets
+back.
+The tool length offset currently applied to motion is reported by
+parameters #5401 to #5409 (set by the G43 family, zeroed by G49), so the
+conversion works whether or not a tool offset is active and you do not
+need to cancel it with G49 first.
+
+.Reading the machine coordinates of a probe
+[source,ngc]
+----
+G38.2 Z-100 F100
+#<zmachine> = [#5063 + #[5203 + #5220 * 20] + #5213 * #5210 + #5403]
+----
+
+The added terms are the current coordinate system Z origin
+(#5223 for G54, #5243 for G55, and so on, selected by #5220), the
+G92/G52 Z offset applied only when active (#5213 multiplied by the
+G92 flag #5210), and the applied tool length offset #5403.
+The same pattern applies to the other axes using their respective
+parameters.
+
 A comment of the form '(PROBEOPEN filename.txt)' will open
 'filename.txt' and store the 9-number coordinate consisting of
 XYZABCUVW of each successful straight probe in it.


docs/src/gcode/g-code.adoc
```diff
@@ -1264,6 +1264,31 @@ G10 L1 P#5400 Z#<zworkoffset> (set new tool offset)
...
+The tool length offset currently applied to motion is reported by
+parameters #5401 to #5409 (set by the G43 family, zeroed by G49), so the
+conversion works whether or not a tool offset is active and you do not
+need to cancel it with G49 first.
+
+.Reading the machine coordinates of a probe
+[source,ngc]
+----
+G38.2 Z-100 F100
+#<zmachine> = [#5063 + #[5203 + #5220 * 20] + #5213 * #5210 + #5403]
```
wucke13, line 1282: Shouldn't this be rather:

```gcode
#<zmachine> = [#5063 + [#5203 + #5220 * 20] + #5213 * #5210 + #5403]
```
?

grandixximo: No, I copied that code from yours, and it's correct 

http://linuxcnc.org/docs/stable/html/gcode/overview.html#sub:subroutine-parameters

`5203` is just to index Z not a parameter but just a number, `#5220` is current coordinate system number, `20` is multiplier because each coordinate system is 20 parameters apart.

Therefore the result inside the square brackets, is the parameter number for the current Z offset for the current coordinate system, so the `#` stays outside, if you were to write `#5203` then that would equal 0 since it is not a system parameter that is in use, then you'd be adding 20 as a value to your expression for G54, which would make no sense and be irrelevant to current offset shifts, instead of addressing `#5223` which is what you actually want.

grandixximo: Basically the hash stays outside, the result of the expression inside the square brackets resolves to the parameter number you actually need, and the hash will then point to the correct parameter, which has the value of the offset for your active coordinate system. It is correct the way you wrote it initially, and wrong in your new proposal.

grandixximo: Maybe it can be dissected in the doc, is it hard to understand?

wucke13: I get my mistake now. Yeah, a sentence or two about it would be good. It's very simple, once one thinks about it :)

c-morley:
Would if be possible to add a parameter that automatically calculates this anytime obe of the variables changes?

grandixximo:
Other controllers expose the current point in two frames and subtract...

We already have the work-frame set in `#5420`-`#5428`. We are only missing the machine-frame set.

I checked and `#5021`-`#5029` are free and can be used for machine-frame set.

Implementation is small. `#5420`-`#5428` are already filled each line from the work-frame position; the machine set fills the same block with `current + tool_offset + g5x_offset + g92_offset` per axis.

What it gives:
- absolute machine position directly
- total offset per axis as `[#5021 - #5420]` similar to other controllers
- the machine-coordinate probe point as `[#5061 + #5021 - #5420]`

I'll look into it tomorrow...

PR #4167, opened 2026-06-15
interp: expose absolute machine position as #5021-#5029

grandixximo:
Adds a machine-coordinate current-position parameter set, `#5021`-`#5029`, for X Y Z A B C U V W.

These report the current absolute machine position (G53 frame), with no coordinate system, G92/G52 or tool length offset applied, in the current program units. They are read-only and volatile, filled each line alongside `#5420`-`#5428` and not persisted to the var file.

This is the same value already exposed by the `#<_abs_*>` named parameters, which existed only for X Y Z A B C; this also adds the missing `#<_abs_u>`, `#<_abs_v>`, `#<_abs_w>` so both sets cover all nine axes. The shared computation is factored into `get_abs_position()`.

Until now the machine position had no numbered parameter; getting it required summing the individual offset registers by formula or the `G30.1` trick (which writes the absolute position into `#5181`-`#5186`). With this set:

- absolute machine position is available directly,
- the active offset per axis is `#5021 - #5420`,
- a probe point in machine coordinates is `#5061 + #5021 - #5420` (no coordinate rotation).

The G38 documentation is updated to use this instead of the offset-summing formula. Includes a new interp test verifying the machine position tracks the commanded G53 location across coordinate system, G92, rotation and tool length offset, and that the numbered set agrees with the named set.

Follow-up to the #4159 #4163 discussion. @cmorley does this match what you had in mind?
@rene-dev this extends your global parameters, looks ok?


grandixximo:
This is already in the 2.9 release, merging in 2.10

PR #4171, opened 2026-06-16
qtvcp: move return out of finally in SAVE_PROGRAM

### LINKED ISSUE #4169: tests/ui-smote/qtdragon-quit consistently fails on F43
The new test for `qtdragon-quit` fails consistently on Fedora 43. In the logs a message is found:
`...SyntaxWarning: 'return' in a 'finally' block.`

This is probably the same issue as #4075. Fedora 43 is on python 3.14.5 and may be a bit more critical what it accepts.

Applying the following diff makes the issue go away (like what was done in #4075), but I have no idea whether this is the right fix.
```diff
diff --git a/lib/python/qtvcp/qt_action.py b/lib/python/qtvcp/qt_action.py
index 3952200bda..afe02a62b8 100644
--- a/lib/python/qtvcp/qt_action.py
+++ b/lib/python/qtvcp/qt_action.py
@@ -394,7 +394,7 @@ class _Lcnc_Action(object):
                 outfile.close()
             except:
                 pass
-            return npath
+        return npath
 
     def SET_AXIS_ORIGIN(self, axis, value):
         if axis == '' or axis.upper() not in ("XYZABCUVW"):
```

@grandixximo Please have a look.

grandixximo:
`return npath` sat inside the `finally` block, which silently overrode the `except` branch's `return None`. SAVE_PROGRAM returned a valid looking path even when the write failed, swallowing the exception. Dedenting it to the method body restores the intended error semantics and clears the SyntaxWarning Python 3.14 emits for `return` in `finally`.

Fixes #4169. Same approach as #4075.

### DIFF
--- a/lib/python/qtvcp/qt_action.py
+++ b/lib/python/qtvcp/qt_action.py
@@ -394,7 +394,7 @@ class _Lcnc_Action(object):
                 outfile.close()
             except:
                 pass
-            return npath
+        return npath
 
     def SET_AXIS_ORIGIN(self, axis, value):
         if axis == '' or axis.upper() not in ("XYZABCUVW"):


grandixximo:
@c-morley same line of bug as #4075 found by @BsAtHome builds clean, waiting on your akn to merge.

BsAtHome:
This is a clear-cut bug.

PR #4175, opened 2026-06-18
docs: tolerate stray whitespace in .so manpage aliases

grandixximo:
A `.so` alias stub is a redirect manpage with no prose: its body is just a path to the page it aliases, identical in every language. asciidoctor emits these stubs as a byproduct when it renders a translated primary page whose NAME line lists several functions, deriving the path from the translated NAME line. A stray space in the translation (`name , other` instead of `name, other`) then lands in the path, producing a broken alias that aborts the `htmldocs` build via `lang_switcher_postprocess.py`.

Rather than trusting translated text for a structural path, this generates the translated `.so` stubs from the canonical English original: once the translated primaries are built, every translated stub is overwritten with the English one. This fixes the roff page too (not just the HTML symlink) and removes the only path by which a translation can break the build.

**Seen in practice:** the German build produced `de/man3/hal_del_funct_from_thread.3` containing `.so hal_add_funct_to_thread .3`, from a space before the comma in the translated NAME line (`hal_add_funct_to_thread , hal_del_funct_from_thread`). The two affected German NAME strings have since been corrected in Weblate; this build change makes any future occurrence harmless.

Verified on a full 8-language build: clean build with 0 malformed stubs and 0 dangling symlinks, and a second build is a no-op.


grandixximo:
@BsAtHome just wanted your eyes on this for a moment, is this the right fix?
Did we miss it cause the german translation has changed?

BsAtHome:
This is handling a symptom and not the cause. You may want to fix the translation instead.

grandixximo:
Agreed, fix the cause. I corrected the two German NAME strings in Weblate, so the immediate breakage is gone. But that does not stop the next translator reintroducing it.

Root problem: a `.so` alias filename is derived from a translatable string. The NAME line bundles the function names (language-invariant identifiers) with the translatable description in one po4a unit, and asciidoctor builds `.so <first-name>.<section>` from the translated name, so a stray space lands in a path.

A `.so` stub is pure structure with no prose, so it should not depend on translation at all. I will rework this PR to generate the per-language stubs from the canonical English stub. Then a translator cannot affect any alias filename, and the roff page is fixed too, not just the HTML symlink. I'll rework to that shape; let me know if you don't agree.

PR #4179, opened 2026-06-19
build: Don't use 0 as NULL/nullptr in pointer compare and assignment

BsAtHome:
There are many places in the sources where the value `0` is used to mean `NULL` or `nullptr`. The ultimate value of these may all be zero, but they are of very different types (pointer vs. integer). This leads to confusion in comparison and wrong assignments where one forgets to dereference or an instance in classicladder assigning `FALSE` to a pointer.

This PR fixes all (detected) instances that gcc and clang could find using the `-Wzero-as-null-pointer-constant` option and is selectable at configure time. There are a few instances in the C/C++ to Python interface code where this option is explicitly ignored with a pragma because the system's headers use the same confusion and cannot be easily fixed.

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1728,6 +1728,19 @@ add_to_cflags() {
 # Always add -Wextra
 add_to_cflags "-Wextra"
 
+# Add option to enable warnings when integer 0 is used to act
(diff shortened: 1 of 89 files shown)


grandixximo (approved):
Looks good to me

PR #4181, opened 2026-06-20
check_config: validate INI datatypes (numeric / enum / duplicate) before comparison

tzuohann:
## Problem

`check_config.tcl` compares `[JOINT_n]`/`[AXIS_L]` limits with Tcl `>`/`<`.
Tcl compares numerically only when *both* operands are valid numbers; otherwise
it silently does a **string** comparison. So a limit line carrying stray text —
e.g. an inline `# comment` — gets compared as a string.

With equal limits like:

    [AXIS_A]
    MAX_LIMIT = 3600000                # placeholder
    [JOINT_4]
    MAX_LIMIT = 3600000                  # placeholder

`3600000 == 3600000`, yet check_config reports
`[JOINT_4]MAX_LIMIT < [AXIS_A]MAX_LIMIT` — because the *strings* (comments +
differing whitespace) compare unequal. Pass/fail flips on trailing spaces, and
the error points at the wrong thing.

## Fix

Add `assert_ini_datatypes`, run once right after `parse_ini`, before any
comparison. For each `[JOINT_n]`/`[AXIS_L]` item:

- **duplicate keys** (multi-valued result) → error;
- **numeric** items must satisfy `string is double -strict` (ints and reals);
- **enum** items must match a fixed set, **case-insensitively**, mirroring
  LinuxCNC's caseless maps:
  - `TYPE` → `{LINEAR ANGULAR}` (`IniFile::mapJointType`)
  - boolean home flags → `{TRUE YES 1 ON FALSE NO 0 OFF}` (`IniFile::convertBool`)
- unknown items are left alone.

Errors name the `[SECTION]ITEM` and the offending value.

### Behavior change
Duplicate `[JOINT_]`/`[AXIS_]` keys are now an **error**, replacing the warn-only
`warn_for_multiple_ini_values` (removed; same scope, now fatal).

## Testing
Patched checker vs a working 5-joint INI and corrupted copies:
- **Accepted:** valid int & real, lowercase `TYPE=angular`, all eight boolean spellings.
- **Rejected (clear message each):** inline comment on a limit, `abc`, `1,000`,
  `1.2.3`, `TYPE=LINER`, `HOME_USE_INDEX=MAYBE`, duplicate key.

## Note
I'm not a software developer by trade, and this is my first day using LinuxCNC,
so I'd genuinely welcome feedback on the approach, the Tcl style, or the
numeric/enum item lists — happy to revise.

### DIFF
--- a/lib/hallib/check_config.tcl
+++ b/lib/hallib/check_config.tcl
@@ -142,6 +142,51 @@ proc validate_identity_kins_limits {} {
   return $emsg
 } ;# validate_identity_kins_limits
 
+proc assert_ini_datatypes {} {
+  # Datatype assertions, run right after parse_ini, before any comparison.
+  # numeric: must be a real number (covers integers too: string is double)
+  set numeric {
+    MIN_LIMIT MAX_LIMIT MAX_VELOCITY MAX_ACCELERATION MAX_JERK
+    FERROR MIN_FERROR BACKLASH
+    P I D FF0 FF1 FF2 BIAS DEADBAND MAX_OUTPUT MAX_ERROR
+    INPUT_SCALE OUTPUT_SCALE SCALE
+    HOME HOME_OFFSET HOME_SEARCH_VEL HOME_LATCH_VEL HOME_FINAL_VEL
+    HOME_SEQUENCE STEPGEN_MAXVEL STEPGEN_MAXACCEL
+  }
+  # enum: must match a fixed set (case-insensitive, mirroring LinuxCNC's
+  # caseless maps mapJointType / convertBool)
+  set bool {TRUE YES 1 ON FALSE NO 0 OFF}
+  array set enum [list \
+    TYPE                        {LINEAR ANGULAR} \
+    HOME_USE_INDEX              $bool \
+    HOME_IGNORE_LIMITS          $bool \
+    HOME_IS_SHARED              $bool \
+    HOME_INDEX_NO_ENCODER_RESET $bool \
+    VOLATILE_HOME               $bool \
+    LOCKING_INDEXER             $bool \
+  ]
+  set emsg ""
+  foreach g [info globals] {
+    if {![string match JOINT_* $g] && ![string match AXIS_* $g]} continue
+    foreach item [array names ::$g] {
+      set val [set ::${g}($item)]
+      if {[llength $val] > 1} {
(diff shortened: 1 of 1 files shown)


grandixximo:
Thanks for digging into this, and welcome to LinuxCNC.

The inline `# comment` case is intentional, not a parser slip. Bertho's INI parser rewrite (April-May this year) makes `#` and `;` comment markers only at the start of a line; after the `=` they are literal characters, by design. The reason is that values legitimately need them. For example the shipped qtdragon configs:

```ini
MDI_COMMAND_MACRO0 = G0 Z2;X0 Y0;Z0, Goto\nUser\nZero
```

The `;` is a G-code command separator. If the parser stripped from `#`/`;` onward, that macro silently loses everything after `Z2`. So treating trailing text as a comment would break configurations that have to keep working. It is documented in `docs/src/config/ini-config.adoc` and pinned by `tests/inifile/comments/`.

Also worth knowing: the datatype, enum, and duplicate-key checks here already exist in `src/emc/ini/linuxcnc_check_ini.py`, the pre-run checker Bertho added in the same series. It validates `TYPE`, the boolean spellings, integer ranges, and duplicate keys, using the C++ parser's exposed enum functions so the accepted values stay in sync with the engine. That is probably the durable place for clearer errors, rather than reimplementing the checks in the hallib Tcl. Worth getting Bertho's read since this is his subsystem.

Given the above, I tend toward closing this, but happy to hear if I am missing something.

BsAtHome:
The lib/hallib/check_config.tcl script has been superseded by emc/ini/linuxcnc_check_ini.py.

Any changes to the Tcl script are in vain.

grandixximo:
@BsAtHome if `check_config.tcl` is fully superseded by `linuxcnc_check_ini.py`, why is it still in the tree? Should it be deleted, or is it kept around intentionally for now?

BsAtHome:
There were some doubts when I wrote the new checker so I left it hanging around unused. There have been no complaints, so better remove the old Tcl script.

tzuohann:
Closing this. After checking, the bug I hit was on my installed stable release (2.9.8), but this PR targets master (2.10), where it's already fixed: the new INI parser strips inline comments in values and `linuxcnc_check_ini.py` (which replaced `check_config.tcl`) already does the numeric/enum/duplicate validation. So this change doesn't add anything to master. Thanks for the review.

PR #4184, opened 2026-06-20
docs(html): add a topbar extension slot for post-processing add-ons

grandixximo:
Following the discussion here, this is reduced to the minimal in-tree piece: a generic, empty extension slot in the docs topbar.

The base build stays minimalist and fetches nothing. The slot (`data-lcnc-slot="topbar-end"`) is an empty placeholder that out-of-tree post-processing can fill, for example to bolt on search, without the LinuxCNC build depending on anything extra. It is empty in the base build, so a plain docs build is unchanged.

As a concrete consumer, I built search as a separate project that downloads the docs artifact, plugs a pagefind search box into this slot, builds a per-language index, and emits searchable docs: https://github.com/grandixximo/linuxcnc-doc-search . Anyone can run their own post-processing against the same slot.

### DIFF
--- a/docs/src/docinfo-header.html.in
+++ b/docs/src/docinfo-header.html.in
@@ -13,5 +13,8 @@
     <span class="lcnc-sep" aria-hidden="true">&bull;</span>
     <a data-lcnc-link="4" href="{lcnc-cssrel}en/gcode.html">G-Code Quick Reference</a>
   </nav>
+  <!-- Extension slot for optional post-processing add-ons (for example a search
+       box injected by a separate tool). Empty in the base build. -->
+  <div class="lcnc-topbar-slot" data-lcnc-slot="topbar-end"></div>
 @LANGUAGE_SWITCHER@
 </header>


BsAtHome:
A local search is not a bad idea as such. But it is a bad idea to have this in the tree IMO. Getting code via curl is a Bad Thing<sup>TM</sup>, even if it is pinned (not happy about rtai either). If you want a search index, then I think it should be a separate project that a user can do.

grandixximo:
Thanks, that is fair. The network fetch during the build is the part I am least comfortable with too, so I would rather drop it than defend it on precedent.

Before I rework this, could you help me understand where you would draw the line, so I build the right thing the first time:

- Would a purely static front-end be acceptable in-tree, that is the search box markup, its CSS, and the `data-pagefind-ignore` markers, with the index itself generated entirely outside the tree by a separate tool run over the built HTML? In that shape the build does no download and gains no extra binary; a plain docs build produces exactly what it does today, and the search box only activates if an index happens to be present alongside it.

- Or is your preference that nothing search-related lives in-tree at all, front-end included, with the whole thing (markup, styling and index) in the separate project?

Either is fine by me. I just want to match what you would be willing to merge before spending the effort. The only goal here is to give the docs the search readers keep asking for, without a fetched binary in the build.

BsAtHome:
The base build should be a minimalist approach and not expect anything special from the host system. Anything else are bells and whistles that can be bolted on. But it is nice to add support for bells and whistles.

An infrastructure placeholder for a search or any other thing could be provided in-tree. Then, when someone wants it, they can plug into the placeholder and put in their stuff using some post-processing. The trick is to add strategic but generic comments, placeholder classes and data-placeholder attributes in the base HTML <i>where appropriate</i>.
After that you can bolt-on search as a separate project. Or, if anyone wants to, do their own post-processing thing.

grandixximo:
Thanks, that steer was helpful. I have reworked this PR down to just the generic extension slot you described, an empty `<div data-lcnc-slot="topbar-end">` in the topbar. The base build is unchanged and fetches nothing.

Search is now a separate project that post-processes the published docs: https://github.com/grandixximo/linuxcnc-doc-search . It downloads the `linuxcnc-doc` artifact that LinuxCNC CI already produces, builds a pagefind index per language, injects a search box into the slot, and emits the searchable docs. Nothing about pagefind touches the linuxcnc tree or build.

On the points you raised:
- pagefind never enters the linuxcnc source or its build.
- It runs only on the separate project's throwaway CI runner, and even there the download is version-pinned and sha256-verified before use.
- The docs server stays static-only: it serves the prebuilt output and never fetches or runs a binary. The base build expects nothing special from the host.
- Search is a progressive enhancement; with JavaScript off the docs are unchanged and fully usable.

One question on where it should live. I created an empty `LinuxCNC/doc-search` and could move the project there so the docs deploy pulls from trusted org infra rather than my personal fork. Would you be OK with that, or do you prefer it stays a user-side project for now? Happy either way.

PR #4185, opened 2026-06-20
docs(i18n): declare the correct html lang on translated pages

grandixximo:
Several translated pages render `<html lang="en">` instead of their actual language. This fixes two cases:

- **Russian and Ukrainian doc pages:** the `:lang:` attribute in `ru.po` and `uk.po` was left untranslated as `"en"`, so every ru/uk page declared English. Set to `ru` and `uk`.
- **Translated man pages (all languages):** the manpage HTML build rule never passed asciidoctor's `lang` attribute, so every language's man pages rendered `<html lang="en">`. Pass `lang=<tag>` per language so they declare their own language, matching the translated doc pages.

A correct `lang` is standard accessibility and i18n hygiene: it drives screen-reader pronunciation, hyphenation, font selection, and per-language tooling. It is also what the in-browser docs search (#4184) keys off to show results in the page's language, so without this those pages fall back to the English index.

Each is a one-line change; no content or layout changes.

### DIFF
--- a/docs/src/Submakefile
+++ b/docs/src/Submakefile
@@ -785,6 +785,7 @@ $(DOC_OUT_HTML)/$(1)/man/%.html: $(2)/% $(DOC_SRCDIR)/docinfo.html $(DOC_SRCDIR)
 			-a compat-mode \
 			-a mansource=LinuxCNC \
 			-a manmanual='LinuxCNC Documentation' \
+			-a "lang=$(1)" \
 			-a "lcnc-cssrel=../../../" \
 			-a "lcnc-lang-label=$(LANG_LABEL_$(1))" \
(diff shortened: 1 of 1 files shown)


BsAtHome:
The problem is that the .po files are coming from upstream weblate. Changing them needs to be sync'ed with weblate and that is a delicate thing to do.

@hansu can you advise?

BsAtHome:
I fixed the two issues in the uk.po and ru.po in weblate. The change here should be removed and then we can see if it is correct when the weblate version is merged.

grandixximo:
Thanks, should have come around to do that myself...

BsAtHome:
I merged the po files and you should rebase and see if that corrects the issue of the po files. Then there should only be the one-liner left.

grandixximo:
Thanks, rebased, clean one line change...

PR #4190, opened 2026-06-21
check_config: assert INI datatypes before comparing limits

tzuohann:
## Problem

A user who puts an inline comment on a limit line — e.g. `MAX_LIMIT = 100  # travel`, easy to do without carefully reading the manual — is currently served a very confusing error from the pre-run config check:

```
[JOINT_0]MAX_LIMIT < [AXIS_X]MAX_LIMIT (100 < {100   # travel})
```

It appears to claim `100 < 100` and points at the wrong problem entirely (a phantom joint/axis limit violation), when the real issue is the stray inline comment.

**Cause:** the `[JOINT_n]`/`[AXIS_L]` limit check in `check_config.tcl` compares values with Tcl `>` / `<`, which silently falls back to *string* comparison when a value isn't a valid number. The comment text makes the value a string, so the outcome depends on trailing whitespace rather than the numbers. A duplicated key behaves the same way (the multi-valued list is compared as a string).

## Reproduce

Minimal trivkins config (`bad.ini`):

```ini
[KINS]
JOINTS = 1
KINEMATICS = trivkins coordinates=X
[AXIS_X]
MAX_LIMIT = 100   # travel
[JOINT_0]
MAX_LIMIT = 100
```

Run the checker against it:

```
$ tclsh lib/hallib/check_config.tcl bad.ini
```

Before this patch it reports the phantom `[JOINT_0]MAX_LIMIT < [AXIS_X]MAX_LIMIT (100 < {100   # travel})`.

## Fix

Add `assert_ini_datatypes()`, run once right after `parse_ini`, before any comparison. For each `[JOINT_n]`/`[AXIS_L]` item it validates the value LinuxCNC actually uses — the **first occurrence** (`IniFile::Find` defaults to `num=1`; the manual, *The INI File → Comments*, likewise states the first value is used):

- numeric items (limits, vels, accels, scales, PID terms, home vels, …) must pass `string is double`;
- enum items must match a fixed set, case-insensitively, mirroring LinuxCNC's own caseless parsers (`TYPE` → LINEAR/ANGULAR; bool set for HOME_*, VOLATILE_HOME, LOCKING_INDEXER).

A non-numeric value containing `#` or `;` additionally reports that inline comments are not allowed and cites the manual section, so the user is pointed at the real mistake:

```
[AXIS_X]MAX_LIMIT must be numeric, got: <100   # travel>
    Inline comments are not allowed: '#' and ';' start a comment
    only at the start of a line, not after a value. Move the comment
    to its own line. See the LinuxCNC manual, "The INI File" ->
    "Comments": https://linuxcnc.org/docs/stable/html/config/ini-config.html#_comments
```

Duplicate (multi-valued) keys are **not** fatal — LinuxCNC uses the first occurrence — so they emit a warning and the array is collapsed to that first value (so the downstream limit comparison uses the same value LinuxCNC would). This subsumes the warn-only `warn_for_multiple_ini_values`, which is removed.

## Testing

Verified with minimal trivkins configs:

| Case | Result |
|---|---|
| inline comment on a limit | error + manual citation (exit 1) |
| duplicate key, first value valid | warning only, **passes** (exit 0) |
| duplicate key, first value has a comment | warning + error on first (exit 1) |
| clean config | passes (exit 0) |

### DIFF
--- a/lib/hallib/check_config.tcl
+++ b/lib/hallib/check_config.tcl
@@ -139,6 +139,67 @@ proc validate_identity_kins_limits {} {
   return $emsg
 } ;# validate_identity_kins_limits
 
+proc assert_ini_datatypes {} {
+  # Datatype assertions, run right after parse_ini, before any comparison.
+  # numeric: must be a real number (covers integers too: string is double)
+  set numeric {
+    MIN_LIMIT MAX_LIMIT MAX_VELOCITY MAX_ACCELERATION MAX_JERK
+    FERROR MIN_FERROR BACKLASH
+    P I D FF0 FF1 FF2 BIAS DEADBAND MAX_OUTPUT MAX_ERROR
+    INPUT_SCALE OUTPUT_SCALE SCALE
+    HOME HOME_OFFSET HOME_SEARCH_VEL HOME_LATCH_VEL HOME_FINAL_VEL
+    HOME_SEQUENCE STEPGEN_MAXVEL STEPGEN_MAXACCEL
+  }
+  # enum: must match a fixed set (case-insensitive, mirroring LinuxCNC's
+  # caseless maps mapJointType / convertBool)
+  set bool {TRUE YES 1 ON FALSE NO 0 OFF}
+  array set enum [list \
+    TYPE                        {LINEAR ANGULAR} \
+    HOME_USE_INDEX              $bool \
+    HOME_IGNORE_LIMITS          $bool \
+    HOME_IS_SHARED              $bool \
+    HOME_INDEX_NO_ENCODER_RESET $bool \
+    VOLATILE_HOME               $bool \
+    LOCKING_INDEXER             $bool \
+  ]
+  set emsg ""
+  foreach g [info globals] {
+    if {![string match JOINT_* $g] && ![string match AXIS_* $g]} continue
+    foreach item [array names ::$g] {
+      set val [set ::${g}($item)]
+      # LinuxCNC reads the first occurrence (IniFile::Find defaults to num=1),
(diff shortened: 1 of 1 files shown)


zz912:
If I understand correctly, this fix is ​​only intended for the stable version 2.9 and is no longer needed in the master version 2.10?

IMHO:
If so, it is not a fix but an improvement. Branch 2.9 is for a fix only and branch 2.10 is for improvements and a fix.

grandixximo:
> inline-comment-stripping parser rewrite exists in master/2.10.

@tzuohann there is no comment stripping in master.

---

@zz912 2.9 can take improvements. Andy is preparing a 2.9.9 release, and I see no reason this shouldn't make it in. It gives cleaner errors for a real, confusing failure. The behavior matches what 2.10 has done for a couple of months via the Python checker, so the concept is proven; this is just a scoped Tcl implementation of the same idea for the branch that still relies on the Tcl checker.

But I think @BsAtHome has been monitoring the transition closer than I have, I've noticed a few reports where we had to guide users on adjusting INI files, but I don't think anything was relevant to spotting the inline comments and giving proper errors, or did I miss something?

418dreamworks:
@grandixximo Noted. Fixed the incorrect reference to 2.10.
@zz912 yeah, I think you are right in calling this an improvement vs a fix. I just want to start contributing and this felt doable. Just want to save someone who didn't read the manual carefully from a confusing error message and to redirect them to the manual. Didn't think it would hurt.

andypugh:
I am going to class it as "fixing" the error message.

PR #4191, opened 2026-06-21
Remove check_config.tcl, superseded by linuxcnc_check_ini.py

grandixximo:
The Tcl pre-run INI checker `lib/hallib/check_config.tcl` was ported to Python in fd1ed196b1 as `src/emc/ini/linuxcnc_check_ini.py`, and the launcher (`scripts/linuxcnc.in`) was switched to call `linuxcnc_check_ini` instead. The Tcl script has not run since: nothing invokes it, no config loads it as a HALFILE, and no test or doc references it. It is installed only incidentally via the `lib/hallib/*` glob.

@BsAtHome confirmed in #4181 that it is fully superseded and can go:

> There were some doubts when I wrote the new checker so I left it hanging around unused. There have been no complaints, so better remove the old Tcl script.

This removes the dead file. No behavior change.

One open question, @BsAtHome: both `src/emc/ini/linuxcnc_check_ini.py` and the generated `bin/linuxcnc_check_ini` still carry a `# Based on check_config.tcl` header line, which now points at a deleted file. Do you want that left as historical provenance, or dropped? I left it untouched in this PR; happy to remove it if you prefer.


zz912:
As an amateur programmer, I prefer to delete everything unnecessary. That's why I wouldn't leave anything there for historical reasons.

grandixximo:
Well, I am not the one to make the call of whether comments referencing the old script are necessary or not, so I'd rather confirm with who wrote the comment in the first place.

In the future if anyone comes in and switches between 2.9 and 2.10 having a comment referencing the old script could be helpful, so it is not _completely_ unnecessary...

BsAtHome:
The history reference is for anyone who wants to track the origins. It is easier when you have a comment. For real historic content you need to go back through git's history, logs and commits. You could argue that all the many places in the code with "based on" comments should be removed, but we don't. They provide a reference and we usually keep them. Well, unless the comment becomes very misleading, but I don't see that here.

PR #4192, opened 2026-06-21
docs: build the same HTML page set for every language

grandixximo:
## Problem

`include::`d partials (`hal/halscope`, `hal/halshow`, `help/rtfaults`, `help/tklinuxcnc`) have no standalone English page: their content renders inside the chapter that includes them. The translated build, though, rendered them as standalone pages (`de/hal/halscope.html` and so on), because the per-language HTML target list is derived from `po4a.cfg`, which lists partials as translation units.

Each such translated page then carried a language switcher linking to `en/hal/halscope.html`, which does not exist. `lang_switcher_postprocess.py` greyed the entry but kept the dead `href`, so it still 404s. This is the bulk of the broken cross-language links an HTML link checker reports in the translated trees.

## Fix

Restrict the per-language HTML targets to the same curated page set English builds (`DOC_HTML_PAGE_STEMS`, from `DOC_SRCS_EN`). Partials are no longer rendered standalone in any language; their text is still translated and inlined into the translated chapters, and `xref_resolver.rb` already redirects partial anchors to the including page.

## Verification

Full 7-language doc build (`de es fr nb ru uk zh_CN`):

- Every language now builds the same 132 pages (`de` was 136 before, the 4 extras being the partials).
- The 4 partials no longer exist as standalone pages in any language.
- No remaining links to those `en/...` pages anywhere in the output.
- Partial content still inlined in the translated chapters (e.g. `halscope` in `de/hal/tutorial.html`).
- A second build is a no-op.


hdiethelm:
@grandixximo Might be rebase to master, so the links are checked?

grandixximo:
Fixes 28 links, other will probably have to be tackled separately, sid is being flaky.

hdiethelm:
Nice, only fragments left and one link:
```
Processing	file:///home/runner/work/linuxcnc/linuxcnc/docs/build/html/es/plasma/plasma-cnc-primer.html


List of broken links and other issues:
http:./qtplasmac.html	
  Line: 281
  Code: 500 No Host option provided
 To do: This is a server side problem. Check the URI.
 ```
 
 The cross links are indeed broken:
```
Processing	file:///home/runner/work/linuxcnc/linuxcnc/docs/build/html/de/getting-started/updating-linuxcnc.html


List of broken links and other issues:
file:///home/runner/work/linuxcnc/linuxcnc/docs/build/html/de/getting-started/updating-linuxcnc.html	
 Lines: 33, 75
  Code: 200 (no message)
 To do: Some of the links to this resource point to broken URI fragments
	(such as index.html#fragment).
The following fragments need to be fixed:
	start:update-no-network       	Line: 75
```

 https://linuxcnc.org/docs/devel/html/de/getting-started/updating-linuxcnc.html
 Click on [Updating without Network](https://linuxcnc.org/docs/devel/html/de/getting-started/updating-linuxcnc.html#start:update-no-network).

Line 75: `<a href="[#start:update-no-network](view-source:https://linuxcnc.org/docs/devel/html/de/getting-started/updating-linuxcnc.html#start:update-no-network)">`
Line 297: `<h2 id="getting-started:update-no-network">2. Aktualisieren ohne Netzwerk</h2>`

Notice `getting-started:update-no-network` vs. `start:update-no-network`.

 I guess this should be a separate PR. It's probably also a translation issue, in the English version, it is fine.

BsAtHome:
> sid is being flaky.

Hm, it seems that both SID instances hang on the test `Running test: ./tests/motion-logger/basic` and having several errors previously. It may be a change in a supplied library?

hdiethelm:
> > sid is being flaky.
> 
> Hm, it seems that both SID instances hang on the test `Running test: ./tests/motion-logger/basic` and having several errors previously. It may be a change in a supplied library?

I have a PR canceling to long taking jobs automatically, so no one has to do it manually:
https://github.com/LinuxCNC/linuxcnc/pull/4207

It might be a temporary issue in sid or also a real library change. This can be investigated in a docker or VM. I would just wait a few days to see if it fixes itself... ;-)

PR #4193, opened 2026-06-22
docs: pin manpage troff filename to the make target

grandixximo:
Reported by @hdiethelm in #4178, where a translated man page rendered with the content of a sibling page and a rerun changed which page was affected.

## Problem

asciidoctor's manpage backend names its output file from the page's `== NAME` section, not from the make target it was invoked for. The troff rule passed `--destination-dir` and trusted asciidoctor to pick the name. For the translated manpages that NAME line goes through po4a, so a translation that altered the command identifier redirected the troff to the wrong filename:

```
#: src/man/man3/hm2_pktuart_read.3.adoc:6
msgid  "hm2_pktuart_read - read data from a Hostmot2 UART buffer"
msgstr "hm2_uart_read - liest Daten aus einem Hostmot2 UART-Puffer"   # dropped "pkt"
```

`de/.../hm2_pktuart_read.3.adoc` then renders to a file named `hm2_uart_read.3`, colliding with the real `hm2_uart_read` page. Under `-j` the two recipes race for one output: the loser is never written, the winner is nondeterministic, so a rerun changes which page is corrupted.

## Fix

Pass `-o "$@"` so the build graph, not the page text, decides each filename. Every target is produced exactly once at its own name, regardless of translation state. A wrong translation can still make a page's visible text wrong, but it can no longer misname a file, race, or drop a page.

NAME drift is still worth seeing, but the build is the wrong place to surface it (it disappears in the build log). So `scripts/manpage-name-check.py` scans the built man tree after the doc build and reports any page whose rendered `.TH` name disagrees with its filename. It is quiet when nothing drifted. When something has, it prints each page with the exact string to search in Weblate and its current value side by side, so the list is a ready worklist. It never fails the build: in CI it writes that same list to the job summary and emits one warning pointing at it (rather than one annotation per page). The NAME lines are translatable strings managed in Weblate, so fixes go there. The check is a convenience for translation cleanup, not load-bearing, so it is easy to drop or fold into the build later if maintainers prefer.

## Verification

Full configured doc build with translations (`--enable-build-documentation --enable-build-documentation-translation`), po4a over all languages, German built through to HTML, then built a second time:

- Race gone: the previously colliding pairs (`hm2_uart_read`/`hm2_pktuart_read`, `hm2_bspi_set_read`/`set_write`, `rtapi_stdint`/`rtapi_string`) are all produced as distinct troff with the correct title, and the man HTML renders.
- Second build is a complete no-op (`Nothing to be done`), which also confirms no target is left missing and po4a is not re-run.
- `manpage-name-check.py` over the built tree flags the drifted translated pages and is silent on the English tree.

Companion to #4175, which covers the same defect on the other output: that PR fixes the `.so` alias stubs of multi-name pages; this fixes the primary page filename. They are independent and land in parallel.

### DIFF
--- a/scripts/manpage-name-check.py
+++ b/scripts/manpage-name-check.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+# Report man pages whose rendered NAME drifted from their filename.
+# asciidoctor takes the .TH name from the page's NAME section, the build pins the filename with -o, so a translated NAME just leaves the two disagreeing.
+# Run after building the docs.  Argument: the built man tree (default docs/build/man).
+# Quiet when nothing drifted.  When something has, it prints each page with the
+# string to search in Weblate and its current value, so a translator can fix it.
+# Never fails the build: under GitHub Actions it writes the same list to the job
+# summary and emits one warning pointing at it.  The NAME lines are managed in
+# Weblate, so fixes go there.
+
+import os
+import re
+import sys
+import glob
+
+# The script lives in scripts/, the docs are one level up at ../docs, so it runs from anywhere.
+HERE = os.path.dirname(os.path.realpath(__file__, strict=True))
+DOCS = os.path.normpath(os.path.join(HERE, '..', 'docs'))
+MAN = os.path.join(DOCS, 'build/man')
+PO = os.path.join(DOCS, 'po')
+SRC = os.path.join(DOCS, 'src/man')
+
+def th_name(troff):
+    """The .TH page name, lowercased with the roff backslash stripped, or None for an .so alias stub."""
+    with open(troff, encoding='utf-8', errors='replace') as f:
+        if f.readline().startswith('.so '):
+            return None
+        f.seek(0)
+        for line in f:
+            m = re.match(r'\.TH "([^"]*)"', line)
+            if m:
+                return m.group(1).replace('\\', '').lower()
+    return None
+
+def name_line(page):
+    """The first line of the NAME section of a page's AsciiDoc source, or None for a comp page rendered straight to troff."""
+    for adoc in sorted(glob.glob(f'{SRC}/man*/{page}.*.adoc')):
+        lines = open(adoc, encoding='utf-8', errors='replace').read().splitlines()
+        for i, line in enumerate(lines):
+            if line.strip() == '== NAME':
+                for rest in lines[i + 1:]:
+                    if rest.strip():
+                        return rest.strip()
+    return None
+
+_po = {}
+def po(lang):
+    """Parse docs/po/<lang>.po into {msgid: msgstr}, joining the continuation lines that wrap one string."""
+    if lang not in _po:
+        d = {}
+        mid = mstr = mode = None
+        for line in open(f'{PO}/{lang}.po', encoding='utf-8', errors='replace'):
+            if line.startswith('msgid '):
+                if mid is not None:
+                    d[mid] = mstr or ''
+                mid, mstr, mode = line[6:].strip().strip('"'), None, 'id'
+            elif line.startswith('msgstr '):
+                mstr, mode = line[7:].strip().strip('"'), 'str'
+            elif line.startswith('"'):
+                s = line.strip().strip('"')
+                if mode == 'id':
+                    mid += s
+                elif mode == 'str':
+                    mstr += s
+            elif not line.strip() and mid is not None:
+                d[mid] = mstr or ''
+                mid = mstr = mode = None
+        if mid is not None:
+            d[mid] = mstr or ''
+        _po[lang] = d
+    return _po[lang]
+
+def weblate_string(lang, page):
+    """Return (search, current): the text to search for in Weblate and its current translation."""
+    d = po(lang)
+    th = page.upper()
+    # A comp page's name is a troff .TH string, searched by the uppercase name.
+    if d.get(th, '').strip() and d[th].strip().lower() != page.lower():
+        return th, d[th]
+    # An AsciiDoc page's name lives in the NAME line, searched by its text.
+    nl = name_line(page)
+    if nl:
+        cur = d.get(nl)
+        if cur is None:                                  # the source line and the msgid can differ in trailing text
+            for mid, mstr in d.items():
+                if mid.startswith(page) and ' - ' in mid:
+                    nl, cur = mid, mstr
+                    break
+        return nl, cur if cur is not None else '(search the text)'
+    return page, '(unknown)'
+
+def drifted(man):
+    """Yield (lang, page) for every built man page whose .TH name does not match its filename."""
+    for dirpath, _dirs, files in os.walk(man):
+        for fn in files:
+            if not re.search(r'\.\d$', fn):
+                continue
+            th = th_name(os.path.join(dirpath, fn))
+            page = re.sub(r'\.\d+$', '', fn)
+            if th is None or th == page.lower():
+                continue
+            lang = os.path.relpath(dirpath, man).split(os.sep)[0]
+            yield ('en' if re.fullmatch(r'man\d+', lang) else lang), page
+
+def collect(man):
+    """All drifted rows as (lang, search, current), sorted, English dropped."""
+    rows = []
+    for lang, page in drifted(man):
+        if lang == 'en':                                 # English is the reference, it never drifts
+            continue
+        search, current = weblate_string(lang, page)
+        rows.append((lang, search, current))
+    rows.sort()
+    return rows
+
+def print_local(rows):
+    print(f'# {len(rows)} man-page NAME mismatches. Search the "search" text in the Weblate docs component, language in brackets.')
+    for lang, search, current in rows:
+        print(f'\n[{lang}]')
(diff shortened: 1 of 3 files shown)


grandixximo:
@BsAtHome are we holding merges till sid is fixed?

hdiethelm:
Hmm, due to you can detect wrong translations: It would make sense to fix (or remove) them and fail if it happens ever again.
I dont see any warnings in the doc build, just to much text... ;-) I would have to search for it and probably no one except you will ever look out for these.

BsAtHome:
> @BsAtHome are we holding merges till sid is fixed?

Currently only the independent doc package build fails. I was not at the dev-meeting yesterday, so I don't know whether the meeting took place or what was discussed.

grandixximo:
Agreed the warning is easy to miss in the build log, and detection without enforcement is weak. I would get to a hard gate in steps rather than flip to fail now:

1. Land this as warn. With `-o "$@"` the build is already safe (no race, no missing page), so a bad translation only degrades a page's text, it cannot break the build. That removes the urgency and lets this merge today.

2. Clear the current backlog in Weblate: 12 German strings plus a few in other languages. These are translatable strings, so the fix belongs in Weblate; a direct `.po` edit gets overwritten on the next sync.

3. Once the tree is clean, add enforcement as a dedicated CI check, not a build failure. A build-level fail would fail every PR that builds docs while any translation is wrong, not just the offending one, and it lives in the 18 minute htmldocs job where it is as easy to miss as the warning. A small standalone check (scan the rendered manpages for NAME != stem) is fast, shows up as its own red status, and can be made a required check.

One caveat for that to actually gate translations: Weblate currently lands both via PR and via direct branch merges (the recent `Merge remote-tracking branch 'weblate/...'`). A required check only blocks the PR path, so Weblate would need to go through PRs for it to truly stop a bad string reaching master.

So: warn now, clean Weblate, then a required standalone check. Same end state you are after, sequenced so we can merge today without wedging other PRs.

BsAtHome:
You cannot enforce CI on weblate translations. People who translate do not usually look at CI. Translations are a tricky business and they are not well organized. It is also one-big-pile-of-text that is very hard to handle. Merges are ad-hoc and breaks are regular. It is hard to enforce structural changes to the translations as long as there is no real plan as how they are supposed to work and be organized.

grandixximo:
Fair, no enforcement then. Not a gate, a warning: a separate CI step runs the name check, never fails, and emits the mismatches as GitHub annotations plus a job-summary table. It shows up in the Checks tab instead of buried in the build log, and blocks nothing. Translators keep working as they do; whoever merges a weblate batch just gets a visible list of what drifted. That gives @hdiethelm the visibility without putting a gate on translations. Acceptable?

BsAtHome:
Yes, that should be enough for now.

grandixximo:
Updated per the discussion, folded the CI warning in rather than opening another PR. The `-o "$@"` pin is the actual fix and stands alone; the build no longer warns inline (it just buried in the log). Instead `scripts/manpage-name-check.sh` reports NAME drift as a non-blocking CI warning, annotations plus a job summary, which gives @hdiethelm the visibility without a gate on translations per @BsAtHome.

I fixed the current drift in Weblate, but that batch has not merged here yet, so the check will still flag those pages until the next Weblate sync lands. Since it only warns, that is harmless, and it shows the check doing its job in the meantime.

hdiethelm:
I like your approach with CI workflow commands and the summary. But it's a bit much:
https://github.com/LinuxCNC/linuxcnc/actions/runs/27940236294?pr=4193
Might be remove the warnings and only show the summary? It looks like there are less warnings than entry's in the summary. Probably the warnings are limited by github, so a summary is probably better. Or just one warning at the end that you should check the summary.

BTW: Here is the doc how to do this if others want this to: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands

grandixximo:
https://github.com/LinuxCNC/linuxcnc/actions/runs/27993036704
How is that? also working on a script to actually get more details about what actually needs fixing, worth it?
To be run manually.

hdiethelm:
Looks nice! Not to much warnings, so you see the relevant ones and a good sumary.

hdiethelm:
I did not look into translations yet, still new to linuxcnc. So you know probably better how much info is needed to fix it. It might be already good eneught.

At least way better than before, a lot was broken and no one took notice.

To much code has always the issue that someone needs to maintain and fix it sometimes in the future.

grandixximo:
I mean it gives you the gist of what the issue might be, but for example the mb2hal
you would not know what was wrong with that, seems to match.

image

but actually

```
[de] mb2hal   (Plain text)
    search:  mb2hal - HAL non-realtime component for Modbus
    current: *mb2hal* - HAL Nicht-Echtzeit-Komponente für Modbus
```

But getting these info out is intensive, it's not worth it to have it in all the builds IMO, so a separate script, which I can have some instructions about, and a mention in the sumamry, that runs locally on demand, and stdout the needed info from a built docs tree, I think is more reasonable, after you get those info it is trivial to find and fix it in weblate, even if you don't understand the language that needs fixing.

hdiethelm:
Ok, why not. The example above would be hard to spot manually.

grandixximo:
I'm not sure I like it that much, it works....
@BsAtHome How would you like this to look?

Edit:
I'm not happy with this, Reshaping into a single script.

BsAtHome:
> I'm not sure I like it that much, it works.... @BsAtHome How would you like this to look?

Very informative list of mismatches. Most, if not all, you'd not need to know the language to fix the translation.

BTW, why is the dash vs em-dash a problem? Is there some processing step that can't handle it? It might have been introduced by auto- or assisted-translation.

grandixximo:
asciidoctor's manpage backend derives the page name by splitting the NAME line on the literal ASCII " - ". An em-dash or en-dash is not that separator, so it folds the volume into the name (`io` renders as `io.1`), which before the `-o` pin could collide with a real page. You might be right about it being automated translation swapping the hyphen for a typographic dash. I already fixed in Weblate by putting " - " back.

grandixximo:
I simplified the columns, I think the "search" col is all you really need to find it in weblate, and the "current" col is just to confirm what we see on this end, I already fixed all 65 warning in weblate, will have to wait for next weblate cycle to see if we get more, or if I missed any.

I'm happy with it now 😊

scripts/manpage-name-check.py
```diff
@@ -0,0 +1,152 @@
...
+# summary and emits one warning pointing at it.  The NAME lines are managed in
+# Weblate, so fixes go there.
+
+import os
+import re
+import sys
+import glob
+
+# Anchor to the repo root (the script lives in scripts/) so it runs from anywhere.
+ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
BsAtHome: Wouldn't you just use `os.path.dirname(os.path.realpath(__file__, strict=True))` here?
Then the others are simply at ../docs/...

grandixximo: i went with

```python
HERE = os.path.dirname(os.path.realpath(__file__, strict=True))
DOCS = os.path.normpath(os.path.join(HERE, '..', 'docs'))
MAN  = os.path.join(DOCS, 'build/man')
...
```

Resolving the script path first then `normpath` keeps a symlinked invocation working, and outputs a clean path on the "no built man pages" error.

BsAtHome:
How are we on the docs issues/PRs? Are there inter-dependencies that would make merging problematic? What is the suggested order?

grandixximo:
I don't think there is much that conflicts, no particular order, if there is anything that gets conflicts should be minor, I can get it fixed in the morning

BsAtHome:
What is the performance of this check? It is going through a lot of files and CI has been shown not to perform stellar all the time. Any comments on that?

grandixximo:
@BsAtHome performance/status update:

Performance:
- The check walks ~16k troff files (30 languages) and only parses docs/po/*.po for languages with drift.
- Clean run: ~0.5s. With drift: a few seconds, because parsing the 30 .po files (~214 MB total) dominates.
- Against the 18m htmldocs job this is negligible.

Status:
- The 65 warnings were from before the end-of-June Weblate cleanup; I rescanned now against a fresh build and found exactly one remaining mismatch:
  [ru] gladevcp_demo — uses an em-dash (—) instead of " - ", so asciidoctor derives .TH "GLADEVCP_DEMO.1" instead of "GLADEVCP_DEMO".
- I fixed that in Weblate just now; it needs one Weblate sync cycle to reach the .po.

Question:
1. Wait for that fix to land, then merge #4193 with the check as a hard failure.
2. Merge #4193 now as a non-failing warning, then flip to hard-fail after the next Weblate sync.

I have --enforce ready; let me know which you prefer.

BsAtHome:
We need to have a clean tree before we merge this.

grandixximo:
@BsAtHome #4325 brought the last Weblate fix, so I rebased on it and the check now runs as a hard failure (`--enforce`). CI all green, including htmldocs with the enforced gate. Ready to merge.

PR #4207, opened 2026-06-28
CI: Set timeouts so it doesn't run forever

hdiethelm:
Timeout set roughly to 3x the normal runtime and some spare. Mainly due to Sid runs for hours lately due to an issue.

This might generate issues in the future if the runtime increases or also just sometime takes longer due to github being github. But it is easy to fix, just increase the timeout.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
 
   rip-and-test:
     runs-on: ubuntu-24.04
+    timeout-minutes: 45
     steps:
     - name: Dump GitHub context
       env:
@@ -57,6 +58,7 @@ jobs:
 
   rip-rtai:
     runs-on: ubuntu-24.04
+    timeout-minutes: 45
     steps:
     - name: Dump GitHub context
       env:
@@ -85,6 +87,7 @@ jobs:
 
   rip-and-test-clang:
     runs-on: ubuntu-24.04
+    timeout-minutes: 45
     steps:
     - name: Dump GitHub context
       env:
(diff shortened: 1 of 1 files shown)


hdiethelm:
Let's wait for an hour if sid timeouts and then it is ready from my side.

hdiethelm:
It works, ready to merge.
grafik

grandixximo:
I think 30min is a bit too restrictive, when we do our Sunday meetings, we merge a lot of stuff, and the CI can take 45min or more in some cases, might have to go back to Sunday's records to get a better estimate of what good runs look like when the CI is busy.

hdiethelm:
> I think 30min is a bit too restrictive, when we do our Sunday meetings, we merge a lot of stuff, and the CI can take 45min or more in some cases, might have to go back to Sunday's records to get a better estimate of what good runs look like when the CI is busy.

This timeouts are per job. So if a job is not started due to now CI runners are available, this should not trigger. The numbers to consider are these:
grafik

However, I added more time all over for good measure in the case github is just slow.

hdiethelm:
This one for example took 60 min:
https://github.com/LinuxCNC/linuxcnc/actions/runs/28195580154
But when you look at the run time:
https://github.com/LinuxCNC/linuxcnc/actions/runs/28195580154/usage
It would not have time-outed. But having a bit more time won't hurt to much.

But it is for sure better than the actual default 6h timeout. Right now, we could run out of runners quite fast if many people push.

grandixximo:
Ok, merging, we can roll back or extend if necessary

BsAtHome:
The reason for Sid to stall is due to a build error in LinuxCNC. Sid upgraded Python to 3.14 and the boost_python library provides both a compatibility library `libboost_python313` and the new `libboost_python314`. The configure script erroneously selects the wrong library and milltask crashes.

hdiethelm:
It's probably still a good idea to timeout CI runs after some time to not stall others. This is also the reason I added it everywhere.

PR #4208, opened 2026-06-28
CI: Set timeouts so it doesn't run forever (2.9)

hdiethelm:
Timeout set roughly to 3x the normal runtime and some spare.

This results in 45...74 min.

https://github.com/LinuxCNC/linuxcnc/pull/4207 back ported to 2.9

Not that urgent, most PR's go to master. Feel free to merge if https://github.com/LinuxCNC/linuxcnc/pull/4207 behaves nicely.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
 
   rip-and-test:
     runs-on: ubuntu-24.04
+    timeout-minutes: 45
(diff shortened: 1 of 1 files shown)


grandixximo:
Timeouts on 2.10 have been reasonable, didn't hit any snag yet, actually I saw some cases where these could have helped, so better merge sooner than later

PR #4209, opened 2026-06-28
build: Select the libboost_python matching the python version.

BsAtHome:
Recent changes in Debian sid cause the wrong libboost_python to be selected in configure (313 instead of 314) causing a crash in milltask.
The python version is now extracted and used to test the "proper" library first before a generic test of which potential libraries are available on the filesystem.

### DIFF
--- a/src/m4/ax_boost_python.m4
+++ b/src/m4/ax_boost_python.m4
@@ -93,7 +93,20 @@ BOOST_PYTHON_MODULE(test) { throw "Boost::Python test."; }]], [])],
 ])
 if test "$ac_cv_boost_python" = "yes"; then
   AC_DEFINE(HAVE_BOOST_PYTHON,,[define if the Boost::Python library is available])
-  ax_python_lib=boost_python
+  # Debian sid after trixie installs both libboost_python313 and
+  # libboost_python314 but it ships with Python 3.14. The detection loop fails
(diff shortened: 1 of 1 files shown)


grandixximo:
Saw you flagged this as "a bit hackery". I think reusing what configure already has would help make it less so.

`AX_PYTHON_DEVEL` already extracts the version into `$ac_python_version` (e.g. `3.14`) and builds `PYTHON_LIBS=-lpython$ac_python_version` from it. So instead of shelling out to `$PYTHON_BIN` again:

```m4
ax_python_lib="boost_python`echo $ac_python_version | tr -d .` boost_python"
```

No extra subprocess, and the boost suffix is guaranteed to match the libpython the build already links. The ordering idea itself is right (CMake derives the same `python3xx` name the same way). Does that feel less hackery to you?

BsAtHome:
Well, yes... I was looking for it but overlooked it. Now it is used with the python method as fallback. Now it feels a lot more resilient.

BTW, your version <i>also</i> creates a sub-process. How else is `tr` run? ;-)

grandixximo:
Yeah `tr` to get rid of the period is less _extra_ lol
Can you backport/cherry-pick to 2.9 as well? There is a bunch of stuff hanging there as well...

PR #4213, opened 2026-06-30
tests: make RT properly start/stop in the raster test

BsAtHome:
RTAI did not pass the raster test because RT was not started properly. This should fix the situation by using a halrun wrapper as already provided by the run_tests script.

As discussed in #4206 to start/stop RT for the raster test in the correct order.


hdiethelm:
Due to you have a PR open already:
This test has two files named "test":
```
tests/hal-stream/test.hal
tests/hal-stream/test.py
```

Probably a bad idea. The only reason this works is due to runtests first checks for *.hal.

BsAtHome:
You are right, accident waiting to happen. Updated PR to fix by renaming test.py into streamtest.py.

PR #4214, opened 2026-06-30
halcmd: add -p option to generate PlantUML diagram of HAL pins/signals

petterreinholdtsen:
Emit bracket-style component boxes grouped by instance, with the component type name (loadrt/loadusr module name) in parentheses. Signals are rendered as queue entities so one writer can fan out to multiple readers via a single node.

Components with all unconnected pins are filtered out.

Also documents the new -p option in the halcmd man page.

This patch was created with help from OpenCode using local llama.cpp server with Qwen 3.6.

### DIFF
--- a/docs/man/man1/halcmd.1
+++ b/docs/man/man1/halcmd.1
@@ -1,5 +1,6 @@
 .\" Copyright (c) 2003 John Kasunich
 .\"                (jmkasunich AT users DOT sourceforge DOT net)
+.\" Copyright (c) 2026 Petter Reinholdtsen <pere@hungry.com>
 .\"
 .\" This is free documentation; you can redistribute it and/or
 .\" modify it under the terms of the GNU General Public License as
@@ -86,6 +87,9 @@ display results of each command
 .TP
 \fB\-V\fR
 display lots of debugging junk
+.TP
+\fB\-p\fR
+Generate a PlantUML diagram of HAL pins and signals. Components are rendered as bracket-style boxes grouped by instance, with the component type name in parentheses. Signals are shown as queue entities, allowing one writer to fan out to multiple readers via a single node. Only components with at least one connected pin are included. Output is written to stdout.
 .TP
 \fB\-h\fR [\fI<command>\fR]
 display a help screen and exit, displays extended help on \fIcommand\fR if specified
--- a/src/hal/utils/halcmd_commands.cc
+++ b/src/hal/utils/halcmd_commands.cc
@@ -11,6 +11,7 @@
  *                     <fenn AT users DOT sourceforge DOT net>
  *                     Stephen Wille Padnos
  *                     <swpadnos AT users DOT sourceforge DOT net>
+ *                     Petter Reinholdtsen <pere@hungry.com>
  *
  *  This program is free software; you can redistribute it and/or
  *  modify it under the terms of version 2 of the GNU General
@@ -2821,6 +2822,264 @@ int do_setexact_cmd() {
     }
     rtapi_mutex_give(&(hal_data->mutex));
     return retval;
+}
+
+#include <map>
+#include <vector>
+#include <algorithm>
+
+static void plantuml_escape_alias(const char *name, FILE *dst) {
+    const char *p = name;
+    while (*p) {
+        if (*p == '-' || *p == '.' || *p == ' ') fputc('_', dst);
+        else fputc(*p, dst);
+        p++;
+    }
+}
+
+static void plantuml_escape_string(const char *name, FILE *dst) {
+    const char *p = name;
+    while (*p) {
+        if (*p == '\\' || *p == '"') fputc('\\', dst);
+        fputc(*p, dst);
+        p++;
+    }
+}
+
+
+static std::string pin_to_instance(const char *pin_name) {
+    const char *last_dot = strrchr(pin_name, '.');
+    if (last_dot) return std::string(pin_name, last_dot - pin_name);
+    return std::string(pin_name);
+}
+
+static const char *pin_field_name(const char *pin_name) {
+    const char *last_dot = strrchr(pin_name, '.');
+    return last_dot ? last_dot + 1 : pin_name;
+}
+
+void generate_plantuml(FILE *dst) {
+    struct pin_entry_t {
+        char full_name[HAL_NAME_LEN+1];
+        bool connected;
+    };
+
+    struct inst_pins_t {
+        char display_name[HAL_NAME_LEN+1];
+        char comp_type[HAL_NAME_LEN+1];
+        std::vector<pin_entry_t> in_pins;
+        std::vector<pin_entry_t> out_pins;
+    };
+
+    // Signal entity: hexagon with connected writer/readers per pin
+    struct sig_pin_t {
+        char alias[HAL_NAME_LEN+1];   // instance alias (e.g. and2_0)
+        char field[64];               // last segment of pin name
+    };
+
+    struct sig_entry_t {
+        char name[HAL_NAME_LEN+1];
+        sig_pin_t *writer;            // single writer pin (or NULL)
+        std::vector<sig_pin_t *> readers;
+    };
+
+    std::map<std::string, inst_pins_t *> inst_map;
+    std::vector<sig_entry_t> signals;
+
+    try {
+        rtapi_mutex_get(&(hal_data->mutex));
+
+        // Collect all pins per instance
+        SHMFIELD(hal_pin_t) next = hal_data->pin_list_ptr;
+        while (next != 0) {
+            hal_pin_t *pin = SHMPTR(next);
+            std::string inst_name = pin_to_instance(pin->name);
+
+            if (inst_map.find(inst_name) == inst_map.end()) {
+                inst_pins_t *ip = new inst_pins_t();
+                snprintf(ip->display_name, sizeof(ip->display_name), "%s", inst_name.c_str());
+                ip->comp_type[0] = '\0';
+                ip->in_pins.reserve(32);
+                ip->out_pins.reserve(32);
+                inst_map[inst_name] = ip;
+            }
+
+            // Look up component type from pin owner
+            hal_comp_t *owner = SHMPTR(pin->owner_ptr);
+            if (owner && strlen(inst_map[inst_name]->comp_type) == 0) {
+                snprintf(inst_map[inst_name]->comp_type, sizeof(inst_map[inst_name]->comp_type), "%s", owner->name);
+            }
+
+            pin_entry_t pe;
+            snprintf(pe.full_name, sizeof(pe.full_name), "%s", pin->name);
+            pe.connected = false;
+
+            if (pin->dir == HAL_IN) {
+                inst_map[inst_name]->in_pins.push_back(pe);
+            } else {
+                inst_map[inst_name]->out_pins.push_back(pe);
+            }
+
+            next = pin->next_ptr;
+        }
+
+        // Walk signals to find connected pins and build signal entries
+        SHMFIELD(hal_sig_t) sig_next = hal_data->sig_list_ptr;
+        while (sig_next != 0) {
+            hal_sig_t *sig = SHMPTR(sig_next);
+            hal_pin_t *writer = NULL;
+            std::vector<hal_pin_t *> readers;
+
+            hal_pin_t *pin = halpr_find_pin_by_sig(sig, 0);
+            while (pin != 0) {
+                if (pin->dir == HAL_OUT) writer = pin;
+                else readers.push_back(pin);
+                pin = halpr_find_pin_by_sig(sig, pin);
+            }
+
+            // Emit signals that have at least one writer or reader connected
+            if (writer || !readers.empty()) {
+                sig_entry_t se;
+                snprintf(se.name, sizeof(se.name), "%s", sig->name);
+                se.writer = NULL;
+                se.readers.reserve(readers.size());
+
+                // Build writer entry and mark pin connected
+                if (writer) {
+                    std::string wi = pin_to_instance(writer->name);
+                    auto *wip = inst_map[wi];
+                    if (wip) {
+                        for (auto &pe : wip->out_pins) {
+                            if (strcmp(pe.full_name, writer->name) == 0) pe.connected = true;
+                        }
+                    }
+                    sig_pin_t *sp = new sig_pin_t();
+                    snprintf(sp->alias, sizeof(sp->alias), "%s", wi.c_str());
+                    const char *wf = pin_field_name(writer->name);
+                    size_t wflen = strlen(wf);
+                    if (wflen >= sizeof(sp->field)) wflen = sizeof(sp->field) - 1;
+                    memcpy(sp->field, wf, wflen);
+                    sp->field[wflen] = '\0';
+                    se.writer = sp;
+                }
+
+                // Build reader entries and mark pins connected
+                for (auto &reader : readers) {
+                    std::string ri = pin_to_instance(reader->name);
+                    auto *rip = inst_map[ri];
+                    if (rip) {
+                        for (auto &pe : rip->in_pins) {
+                            if (strcmp(pe.full_name, reader->name) == 0) pe.connected = true;
+                        }
+                    }
+                    sig_pin_t *sp = new sig_pin_t();
+                    snprintf(sp->alias, sizeof(sp->alias), "%s", ri.c_str());
+                    const char *rf = pin_field_name(reader->name);
+                    size_t rflen = strlen(rf);
+                    if (rflen >= sizeof(sp->field)) rflen = sizeof(sp->field) - 1;
+                    memcpy(sp->field, rf, rflen);
+                    sp->field[rflen] = '\0';
+                    se.readers.push_back(sp);
+                }
+
+                signals.push_back(se);
+            }
+
+            sig_next = sig->next_ptr;
+        }
+
+        rtapi_mutex_give(&(hal_data->mutex));
+    } catch (...) {
+        rtapi_mutex_give(&(hal_data->mutex));
+
+        // Cleanup any partial allocations before re-throwing
+        for (auto &se : signals) {
+            delete se.writer;
+            for (auto &rp : se.readers) delete rp;
+        }
+        for (auto &kv : inst_map) delete kv.second;
+        throw;
+    }
+
+    fprintf(dst, "@startuml\n");
+
+    // Emit only instances that have at least one connected pin
+    for (auto &kv : inst_map) {
+        inst_pins_t *ip = kv.second;
+        bool has_conn = false;
+        for (const auto &pe : ip->in_pins) if (pe.connected) { has_conn = true; break; }
+        for (const auto &pe : ip->out_pins) if (pe.connected) { has_conn = true; break; }
+        if (!has_conn) continue;
+
+         fputs("[", dst);
+        fputs(ip->display_name, dst);
+        if (strlen(ip->comp_type) > 0) {
+            fprintf(dst, " (%s)", ip->comp_type);
+        }
+
+        for (const auto &pe : ip->in_pins) {
+            const char *f = pin_field_name(pe.full_name);
+            fputs("\\n", dst);
+            fputs(f, dst);
+            fputs(" in", dst);
+            if (!pe.connected) fputs(" (unconnected)", dst);
+        }
+        for (const auto &pe : ip->out_pins) {
+            const char *f = pin_field_name(pe.full_name);
+            fputs("\\n", dst);
+            fputs(f, dst);
+            fputs(" out", dst);
+            if (!pe.connected) fputs(" (unconnected)", dst);
+        }
+
+        fprintf(dst, "] as ");
+        plantuml_escape_alias(ip->display_name, dst);
+        fprintf(dst, "\n");
+    }
+
+   // Emit signals as queue entities (hexagon not available in PlantUML v1.2020)
+    for (size_t i = 0; i < signals.size(); i++) {
+        char alias_buf[HAL_NAME_LEN + 1];
+        snprintf(alias_buf, sizeof(alias_buf), "sig_%zu", i);
+        fprintf(dst, "queue \"");
+        plantuml_escape_string(signals[i].name, dst);
+        fprintf(dst, "\" as %s\n", alias_buf);
+    }
+
+    // Emit edges: writer --> signal and signal --> reader
+    for (size_t i = 0; i < signals.size(); i++) {
+        char alias_buf[HAL_NAME_LEN + 1];
+        snprintf(alias_buf, sizeof(alias_buf), "sig_%zu", i);
+
+        if (signals[i].writer) {
+            plantuml_escape_alias(signals[i].writer->alias, dst);
+            fprintf(dst, " \"");
+            plantuml_escape_string(signals[i].writer->field, dst);
+            fprintf(dst, "\" --> ");
+            fputs(alias_buf, dst);
+            fprintf(dst, "\n");
+        }
+        for (auto &rp : signals[i].readers) {
+            fputs(alias_buf, dst);
+            fprintf(dst, " --> \"");
+            plantuml_escape_string(rp->field, dst);
+            fprintf(dst, "\" ");
+            plantuml_escape_alias(rp->alias, dst);
+            fprintf(dst, "\n");
+        }
+    }
+
+    // Cleanup signal allocations
+    for (auto &se : signals) {
+        delete se.writer;
+        for (auto &rp : se.readers) delete rp;
+    }
+
+    for (auto &kv : inst_map) {
+        delete kv.second;
+    }
+
+    fprintf(dst, "@enduml\n");
 }
 
 int do_help_cmd(char *command)
--- a/src/hal/utils/halcmd_commands.h
+++ b/src/hal/utils/halcmd_commands.h
@@ -11,6 +11,7 @@
  *                     <fenn AT users DOT sourceforge DOT net>
  *                     Stephen Wille Padnos
  *                     <swpadnos AT users DOT sourceforge DOT net>
+ *                     Petter Reinholdtsen <pere@hungry.com>
  *
  *  This program is free software; you can redistribute it and/or
  *  modify it under the terms of version 2 of the GNU General
@@ -86,6 +87,7 @@ extern int do_loadusr_cmd(const char *args[]);
 extern int do_waitusr_cmd(char *comp_name);
 extern int do_save_cmd(const char *type, char *filename);
 extern int do_setexact_cmd(void);
+extern void generate_plantuml(FILE *dst);
 
 pid_t hal_systemv_nowait(const char *const argv[]);
 int hal_systemv(const char *const argv[]);
--- a/src/hal/utils/halcmd_main.c
+++ b/src/hal/utils/halcmd_main.c
@@ -11,6 +11,7 @@
  *                     <fenn AT users DOT sourceforge DOT net>
  *                     Stephen Wille Padnos
  *                     <swpadnos AT users DOT sourceforge DOT net>
+ *                     Petter Reinholdtsen <pere@hungry.com>
  *
  *  This program is free software; you can redistribute it and/or
  *  modify it under the terms of version 2 of the GNU General
@@ -87,6 +88,8 @@ static char *prompt_continue    = "halcmd+: ";
 
 #define MAX_EXTEND_LINES 20
 
+static int plantuml_mode = 0;
+
 /***********************************************************************
 *                   LOCAL FUNCTION DEFINITIONS                         *
 ************************************************************************/
@@ -113,7 +116,7 @@ int main(int argc, char **argv)
     keep_going = 0;
     /* start parsing the command line, options first */
     while(1) {
-        c = getopt(argc, argv, "+RCfi:kqQsvVhe");
+        c = getopt(argc, argv, "+RCfi:kqQsvVhep");
         if(c == -1) break;
         switch(c) {
             case 'R':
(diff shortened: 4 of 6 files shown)


BsAtHome:
Why can't you just use script-mode for output (halcmd -s) and post-process with your own script?
I cannot see a reason why you need to have plantuml specific output to be part of halcmd. If script-mode shows a deficiency for post-processing , then we can simply fix that (in a compatible way).

petterreinholdtsen:
> Why can't you just use script-mode for output (halcmd -s) and post-process with your own script?

Who say I can not?  I just believe it is more convenient and available for a large audience if the program do this directly and on its own, instead of having to track down a separate tool.

This patch is actually a byproduct of an experiment adding a graph tab to halshow.  Have not yet found a way to make that graph pretty and useful, so in the mean time I have been using plantuml to get an overview of the HAL setup and changes on machines. :)

andypugh:
Do you have a sample of the plantuml output as SVG?

BsAtHome:
But still, the correct way then is to make the script that is part of LinuxCNC for that "large audience". You still need plantuml for it to function. So, you make a script that calls `halcmd -s` processes the output and wraps it all up. No need to add another mode to halcmd. Just make a script that does it all for that large audience.

The problem I'm addressing here is that you can add N+1 output interfaces to halcmd and still not be able to capture them all for any audience's liking. However, it makes halcmd unnecessarily complex and difficult to maintain. Therefore, using <i>one</i> standard script-friendly output format, which we have, should suffice and everything else is <i>external</i> or <i>encapsulated</i> post-processing.

petterreinholdtsen:
> Do you have a sample of the plantuml output as SVG?

Sure.  Attached.
simulation

petterreinholdtsen:
> The problem I'm addressing here is that you can add N+1 output interfaces to halcmd and still not be able to capture them all for any audience's liking.

While it of course is true and sound very dramatic, I suspect that in reality, there are two major formats that are relevant, the graphviz dot format and plantuml.  So N=0 will make a lot of people happy and N=1 will provide a useful format for the wast majority.  I would be happy to add a -g option for graphviz dot, if you believe it is vital to increase the audience approval.

BsAtHome:
What about PGF/TikZ? Or R? Or Vym?. Then there are numerous other packages. Just looking at the very long list of available UML tools should give you an indication how many different ways there are to tell a similar story. There is the whole category of argument/concept mapping software and associated languages. You may not be able to surface once you down that rabbit hole and forever stay in the Red Queen's prison with your head chopped off.

FWIW, we already have N=1<sub>you know the saying, use N=0, N=1 or N=∞</sub>. With N=1 we have the script-friendly output. As said, if it is insufficient or incomplete, then it should be fixed to be generic. But I'd strongly resist adding any specific target language.

petterreinholdtsen:
[BsAtHome]
> What about PGF/TikZ? Or R? Or Vym?. Then there are numerous other
> packages. Just looking at the very long list of available UML tools
> should give you an indication how many different ways there are to
> tell a similar story. There is the whole category of argument/concept
> mapping software and associated languages. You may not be able to
> surface once you down that rabbit hole and forever stay in the Red
> Queen's prison with your head chopped off.

I can probably even come up with my own special graph format, to
increase the number of packages and formats.  But doing this do not
really affect the popularity of the most used formats, and thus not my
observation on how many formats are needed to make the majority
happy. :)

A more relevant measurement might be to see how many competing LinuxCNC
extentions exist for drawing a graph from the HAL structure, and which
format they are using.  I am thinking about projects like

<URL: https://github.com/JTrantow/LCNC_HAL_to_graphviz > and the many
mentioned in
<URL: https://www.forum.linuxcnc.org/24-hal-components/37821-graphing-a-hal-configuration >

> FWIW, we already have N=1<sub>you know the saying, use N=0, N=1 or
> N=∞</sub>. With N=1 we have the script-friendly output. As said, if it
> is insufficient or incomplete, then it should be fixed to be generic.
> But I'd strongly resist adding any specific target language.

Always fun with some math, and while induction is a powerful tool, it is
unlikely to apply if the goal is to provide for the majority, not track
down every potential user and cater for every need. :)

-- 
Happy hacking
Petter Reinholdtsen

andypugh:
The current script-friendly output isn't something that any apt-installable package can parse into a diagram. 
This UML output can go straight into existing tools (as could .dot format)
There are a number of third-party tools, though some require a running HAL session to query in order to work, whereas some work from the HAL file itself. 

Ideally we would have a version of Halshow where you could drag the blocks around to untangle lines, and the values were displayed live on the nets. But at that point we would be re-inventing Simulink or LabView. 

It's worth remembering, also, that most HAL files are fit-and-forget. after initial configuration they are not looked at again for years at a time. For most users there are more valuable places to invest developer effort. 

This PR was prompted by trying to get to the bottom of the spindle control on Petter's Mazak, which after a few years was a mystery to both of us. Using[ this script ](https://forum.linuxcnc.org/media/kunena/attachments/23278/dot_2018-08-02.py)(converted to Python3) was very helpful when we were trying to work out how the spindle was controlled

hdiethelm:
I agree that `halcmd` is the wrong place to implement a specific graph format. At the end, someone wants to implement a VisualStudio drawing generator in there... ;-)

However, having ways to generate a graph from the connections is for sure nice and looking at the above repo, parsing the output of `halcmd -s` looks cumbersome.

Options:
- Use the available https://linuxcnc.org/docs/devel/html/en/config/python-hal-interface.html to gather the data. It looks like a function to show the output to signal connection is missing. Signal to input can be shown using `hal.get_info_signals()`. If you add generic functions here that can be used to generate a nice diagram in what ever form you want in a python script will probably be approved. I could help.
- Add json output to halcmd which can easily parse by most programming languages. I would say that's not needed, the python hal library is the better way.

Adding a few python graph generators into the repo in an appropriate location would probably also be fine.

Together with  python hal -> graph generator -> render tool, this can be even integrated into an UI.

The drag mode is also realizable by improving python hal so all info can be gathered and then a python tool that does this. As much as I know, there are nice library's to do that.

BsAtHome:
In principle, you need a "dump" of all modules/threads/functions/components/signal/pins/parameters with all the retained information in a structured format. Although halmodule may be interesting, it is not necessarily the right place. However, there may be a case for it (see below).
The current constructs in these tools require you to instantiate a (temporary) component, just to get access to the shared memory. This is something I am working on so it no longer is a requirement to create temporary components. It simplifies things significantly. Many functions, like getting/setting pins/params/signals do not need a component at all to work.
I already have a complete new query API in (user-space) hal_lib in my tree that lets you look at everything <i>without</i> the need for people poking around in HAL's internal structures (as part of complete removal of access to hal_priv.h). A wrapper could be implemented in halcmd to dump all content in a structured way. It can also be implemented as a query interface in halmodule when it no longer requires a (temporary) component to function. Then you can write a new program "haldump.py" if you like.

hdiethelm:
Hmm, is this an issue if you have to create a temporary halmodule? Many components do this like halshow / halcmd and so on.

Of course, having better separation would help and there is really no need for a component just to access shm.

A small python script that shows what is already available in hal python. For the get_info_pins() a DRIVER field is missing and there is also no get_info_threads(). But this could be done and then all should be there needed to create a graph in python.
```python
import hal
import os

comp_name = f"halpy{os.getpid()}"
if not hal.is_initialized():
    comp = hal.component(comp_name)


print("pins-----------")
for pin in hal.get_info_pins():
    for k, v in pin.items():
        print(k, v)
    print()

print("signals-----------")
for sig in hal.get_info_signals():
    for k, v in sig.items():
        print(k, v)
    print()

print("params-----------")
for par in hal.get_info_params():
    for k, v in par.items():
        print(k, v)
    print()
```

hdiethelm:
A small change in halmodule.cc and I think everything needed to generate a diagram from python is there. If hal.get_info_components() / hal.get_info_threads() is needed, this should also be possible. It would just expand an already existing pattern.

@BsAtHome Do you think this could be a way? Yes, it needs a component but for now, all modules accessing the HAL need one. If your changes gets in, the component creation could be dropped, done.

```diff
diff --git a/src/hal/halmodule.cc b/src/hal/halmodule.cc
index b0a132fd0c..8be61a160d 100644
--- a/src/hal/halmodule.cc
+++ b/src/hal/halmodule.cc
@@ -1554,10 +1554,12 @@ PyObject *get_info_pins(PyObject * /*self*/, PyObject * /*args*/) {
     static const char str_v[] = "VALUE";
     static const char str_t[] = "TYPE";
     static const char str_d[] = "DIRECTION";
+    static const char str_s[] = "SIGNAL";
     hal_data_u *d_ptr;
 
     hal_pin_t *pin;
     hal_sig_t *sig;
+    char* sig_name;
 
     PyObject* python_list = PyList_New(0);
     PyObject *obj;
@@ -1573,61 +1575,69 @@ PyObject *get_info_pins(PyObject * /*self*/, PyObject * /*args*/) {
         if (pin->signal != 0) {
             sig = (hal_sig_t*)SHMPTR(pin->signal);
             d_ptr = reinterpret_cast<hal_data_u *>(SHMPTR(sig->data_ptr));
+            sig_name = sig->name;
         } else {
             sig = NULL;
             d_ptr = &(pin->dummysig);
+            sig_name = (char*)"UNCONNECTED";
         }
-
         /* convert to dict of python values */
         switch(type) {
             case HAL_BIT:
-                obj = Py_BuildValue("{s:s,s:N,s:N,s:N}",
+                obj = Py_BuildValue("{s:s,s:N,s:N,s:N,s:s}",
                         str_n, pin->name,
                         str_v, PyBool_FromLong((long)d_ptr->b),
                         str_d, PyLong_FromLong(pin->dir),
-                        str_t, PyLong_FromLong(HAL_BIT));
+                        str_t, PyLong_FromLong(HAL_BIT),
+                        str_s, sig_name);
                 break;
             case HAL_U32:
-                obj = Py_BuildValue("{s:s,s:k,s:N,s:N}",
+                obj = Py_BuildValue("{s:s,s:k,s:N,s:N,s:s}",
                         str_n, pin->name,
                         str_v, (unsigned long)d_ptr->u,
                         str_d, PyLong_FromLong(pin->dir),
-                        str_t, PyLong_FromLong(HAL_U32));
+                        str_t, PyLong_FromLong(HAL_U32),
+                        str_s, sig_name);
                 break;
             case HAL_S32:
-                obj = Py_BuildValue("{s:s,s:l,s:N,s:N}",
+                obj = Py_BuildValue("{s:s,s:l,s:N,s:N,s:s}",
                         str_n, pin->name,
                         str_v, (long)d_ptr->s,
                         str_d, PyLong_FromLong(pin->dir),
-                        str_t, PyLong_FromLong(HAL_S32));
+                        str_t, PyLong_FromLong(HAL_S32),
+                        str_s, sig_name);
                 break;
             case HAL_U64:
-                obj = Py_BuildValue("{s:s,s:K,s:N,s:N}",
+                obj = Py_BuildValue("{s:s,s:K,s:N,s:N,s:s}",
                         str_n, pin->name,
                         str_v, (unsigned long long)d_ptr->lu,
                         str_d, PyLong_FromLong(pin->dir),
-                        str_t, PyLong_FromLong(HAL_S64));
+                        str_t, PyLong_FromLong(HAL_S64),
+                        str_s, sig_name);
                 break;
             case HAL_S64:
-                obj = Py_BuildValue("{s:s,s:L,s:N,s:N}",
+                obj = Py_BuildValue("{s:s,s:L,s:N,s:N,s:s}",
                         str_n, pin->name,
                         str_v, (long long)d_ptr->ls,
                         str_d, PyLong_FromLong(pin->dir),
-                        str_t, PyLong_FromLong(HAL_S64));
+                        str_t, PyLong_FromLong(HAL_S64),
+                        str_s, sig_name);
                 break;
             case HAL_FLOAT:
-                obj = Py_BuildValue("{s:s,s:d,s:N,s:N}",
+                obj = Py_BuildValue("{s:s,s:d,s:N,s:N,s:s}",
                         str_n, pin->name,
                         str_v, (double)d_ptr->f,
                         str_d, PyLong_FromLong(pin->dir),
-                        str_t, PyLong_FromLong(HAL_FLOAT));
+                        str_t, PyLong_FromLong(HAL_FLOAT),
+                        str_s, sig_name);
                 break;
             case HAL_PORT:
-                obj = Py_BuildValue("{s:s,s:l,s:N,s:N}",
+                obj = Py_BuildValue("{s:s,s:l,s:N,s:N,s:s}",
                         str_n, pin->name,
                         str_v, (long)d_ptr->p,
                         str_d, PyLong_FromLong(pin->dir),
-                        str_t, PyLong_FromLong(HAL_PORT));
+                        str_t, PyLong_FromLong(HAL_PORT),
+                        str_s, sig_name);
                 break;
             case HAL_TYPE_UNSPECIFIED: /* fallthrough */ ;
             case HAL_TYPE_UNINITIALIZED: /* fallthrough */ ;
@@ -1636,7 +1646,8 @@ PyObject *get_info_pins(PyObject * /*self*/, PyObject * /*args*/) {
                         str_n, pin->name,
                         str_v, NULL,
                         str_d, PyLong_FromLong(pin->dir),
-                        str_t, NULL);
+                        str_t, NULL,
+                        str_s, sig_name);
                  break;
         }

```
Python script:
```python
import hal
import os

comp_name = f"halpy{os.getpid()}"
if not hal.is_initialized():
    comp = hal.component(comp_name)


print("pins-----------")
for pin in hal.get_info_pins():
    for k, v in pin.items():
        print(k, v, end='; ')
    print()

print("signals-----------")
for sig in hal.get_info_signals():
    for k, v in sig.items():
        print(k, v, end='; ')
    print()

print("paramteters-----------")
for par in hal.get_info_params():
    for k, v in par.items():
        print(k, v, end='; ')
    print()
```
Output:
```
pins-----------
NAME fast.time; VALUE 40; DIRECTION 32; TYPE 3; SIGNAL UNCONNECTED; 
NAME lat.bj; VALUE 1731629; DIRECTION 16; TYPE 3; SIGNAL bj; 
NAME lat.bl; VALUE 1756629; DIRECTION 16; TYPE 3; SIGNAL bl; 
NAME lat.bt; VALUE 90; DIRECTION 16; TYPE 3; SIGNAL bt; 
NAME lat.reset; VALUE False; DIRECTION 32; TYPE 1; SIGNAL reset; 
NAME lat.sj; VALUE 1741692; DIRECTION 16; TYPE 3; SIGNAL sj; 
NAME lat.sl; VALUE 2741692; DIRECTION 16; TYPE 3; SIGNAL sl; 
NAME lat.st; VALUE 969250; DIRECTION 16; TYPE 3; SIGNAL st; 
NAME slow.time; VALUE 150; DIRECTION 32; TYPE 3; SIGNAL UNCONNECTED; 
NAME timedelta.0.avg-err; VALUE -0.002530682045127165; DIRECTION 32; TYPE 2; SIGNAL UNCONNECTED; 
NAME timedelta.0.current-error; VALUE -24910; DIRECTION 32; TYPE 3; SIGNAL UNCONNECTED; 
NAME timedelta.0.current-jitter; VALUE 24910; DIRECTION 32; TYPE 3; SIGNAL UNCONNECTED; 
NAME timedelta.0.err; VALUE -52410; DIRECTION 32; TYPE 3; SIGNAL UNCONNECTED; 
NAME timedelta.0.jitter; VALUE 1731629; DIRECTION 32; TYPE 3; SIGNAL bj; 
NAME timedelta.0.max; VALUE 1756629; DIRECTION 32; TYPE 3; SIGNAL bl; 
NAME timedelta.0.min; VALUE 79; DIRECTION 32; TYPE 3; SIGNAL UNCONNECTED; 
NAME timedelta.0.out; VALUE 90; DIRECTION 32; TYPE 3; SIGNAL bt; 
NAME timedelta.0.reset; VALUE False; DIRECTION 16; TYPE 1; SIGNAL reset; 
NAME timedelta.0.time; VALUE 40; DIRECTION 32; TYPE 3; SIGNAL UNCONNECTED; 
NAME timedelta.1.avg-err; VALUE -0.0012931562094168484; DIRECTION 32; TYPE 2; SIGNAL UNCONNECTED; 
NAME timedelta.1.current-error; VALUE -30750; DIRECTION 32; TYPE 3; SIGNAL UNCONNECTED; 
NAME timedelta.1.current-jitter; VALUE 30750; DIRECTION 32; TYPE 3; SIGNAL UNCONNECTED; 
NAME timedelta.1.err; VALUE -26781; DIRECTION 32; TYPE 3; SIGNAL UNCONNECTED; 
NAME timedelta.1.jitter; VALUE 1741692; DIRECTION 32; TYPE 3; SIGNAL sj; 
NAME timedelta.1.max; VALUE 2741692; DIRECTION 32; TYPE 3; SIGNAL sl; 
NAME timedelta.1.min; VALUE 321; DIRECTION 32; TYPE 3; SIGNAL UNCONNECTED; 
NAME timedelta.1.out; VALUE 969250; DIRECTION 32; TYPE 3; SIGNAL st; 
NAME timedelta.1.reset; VALUE False; DIRECTION 16; TYPE 1; SIGNAL reset; 
NAME timedelta.1.time; VALUE 150; DIRECTION 32; TYPE 3; SIGNAL UNCONNECTED; 
signals-----------
NAME bj; VALUE 1731629; DRIVER timedelta.0.jitter; TYPE 3; 
NAME bl; VALUE 1756629; DRIVER timedelta.0.max; TYPE 3; 
NAME bt; VALUE 90; DRIVER timedelta.0.out; TYPE 3; 
NAME reset; VALUE False; DRIVER lat.reset; TYPE 1; 
NAME sj; VALUE 1741692; DRIVER timedelta.1.jitter; TYPE 3; 
NAME sl; VALUE 2741692; DRIVER timedelta.1.max; TYPE 3; 
NAME st; VALUE 969250; DRIVER timedelta.1.out; TYPE 3; 
paramteters-----------
NAME fast.tmax; DIRECTION 192; VALUE 46206; TYPE 3; 
NAME slow.tmax; DIRECTION 192; VALUE 8265; TYPE 3; 
NAME timedelta.0.tmax; DIRECTION 192; VALUE 46206; TYPE 3; 
NAME timedelta.0.tmax-increased; DIRECTION 64; VALUE False; TYPE 1; 
NAME timedelta.1.tmax; DIRECTION 192; VALUE 8265; TYPE 3; 
NAME timedelta.1.tmax-increased; DIRECTION 64; VALUE False; TYPE 1;
```

multigcs:
here my one-script solution:

https://www.youtube.com/watch?v=isgavyVCYJc
https://github.com/multigcs/halviewer

it using graphviz for the initial layout and QGraphicsScene for display and editing

multigcs:
```
PlantUML natively supports Graphviz (DOT) code. You do not need to rewrite or convert your DOT files. Simply change @startdot to @startuml and place your DOT code directly inside your PlantUML block.
```
So if you're going to use an export format, Graphviz / dot is the better choice

BsAtHome:
Halmodule does not provide any relation to the owner component of the pins and parameters. You also need to implement component listing. And while you are at it, you want all the other info too. That is simply not implemented in halmodule and is a bit off-topic for halmodule. It would then be better to create a `halquery` module specifically to get information on the HAL shared memory content and structure.

And, about the format, there is no point in having a "well, most use XYZ" format implemented. You need a generic interface so that the question of "who can interface" is answered with "anyone, use your own parser/converter to mangle the generic output".

FWIW, I already created a `halquery` module in my tree. It can query individual parts or list entire categories. A run on a default axis sim gives [axis-idle.txt](https://github.com/user-attachments/files/29595225/axis-idle.txt) with this script:
```python
#!/usr/bin/env python3
import halquery
print("# Pins")
for k, v in halquery.pins().items():
    print(k, v)
print("# Params")
for k, v in halquery.params().items():
    print(k, v)
print("# Signals")
for k, v in halquery.signals().items():
    print(k, v)
print("# Components")
for k, v in halquery.comps().items():
    print(k, v)
print("# Functions")
for k, v in halquery.functs().items():
    print(k, v)
print("# Threads")
for k, v in halquery.threads().items():
    fc = v['functions']
    v['functions'] = '...'
    print(k, v)
    for f in fc:
        print("   ", f)
```

It reveals the complete internal structure in python dictionaries which you can use in any way you want and mangle into whatever you like.

hdiethelm:
> ```
> PlantUML natively supports Graphviz (DOT) code. You do not need to rewrite or convert your DOT files. Simply change @startdot to @startuml and place your DOT code directly inside your PlantUML block.
> ```
> 
> So if you're going to use an export format, Graphviz / dot is the better choice

That's why I think we should not export a specific format at all as @BsAtHome pointed out to. This is a discussion that can never be ended... ;-)

The better way is good support for everyone that wants to write a graph exporter or a real time graph. Based on all scripts I have seen so far, having this as a python module probably would do it.

That would involve the following steps:
1. Create a new module or improve the hal module to allow for easy creations of such viewers.
2. Create a common entry point to start, something like halviewer which is installed in the path from where all nice tools can be started.
3. Integrate a few very useful tools in there.

This would result in the end:
halviewer --export-dot -> uses lib/python/halviewer/export-dot.py
halviewer --export-plantuml -> uses lib/python/halviewer/export-plantuml.py
halviewer --live-view -> uses lib/python/halviewer/live-view.py (Based on  multigcs code for example)

This way, all can be integrated without one integration breaking an other one and without having a single line of change in halcmd.

Reasoning behind: 
- If halcmd breaks, linuxcnc is toast.
- If a single viewer breaks, who cares, it can be fixed or also just dropped it it is unfixable or the original contributor is not available. A few people won't be happy but they are free to fix it and create a PR.
- Dependency's of a viewer can be suggestions, so you can install them if you want it or also not.

That said, this is just my opinion and a short concept. Other opinions are welcome.

@multigcs I really like you viewer. The only thing I don't really like is that you have to use halcmd and decode the text to get the graph. Not your fault, I think there is just now other way to do this but we should change that.

hdiethelm:
> It reveals the complete internal structure in python dictionaries which you can use in any way you want and mangle into whatever you like.

Looks nice! I was not aware you are already working on it.

One thing: Having type / dir as the C enum value doesn't look python style to me: https://docs.python.org/3/library/enum.html @grandixximo seams you know python better than me, any insight?

I know it is also done this way in halmodule.cc but I think there are better ways. Either directly in the C Python library or also on the python side by using the C constants.

Do you have a branch to share?

multigcs:
>@multigcs I really like you viewer. The only thing I don't really like is that you have to use halcmd and decode the text to get the graph. Not your fault, I think there is just now other way to do this but we should change that.

Thanks !

I would have liked to do it entirely using the HAL module, but as you already mentioned, I was missing some information:

> For the get_info_pins() a DRIVER field is missing

But the advantage of the halcmd parser is that you can use a dump to view other people's configurations. :)

hdiethelm:
> I would have liked to do it entirely using the HAL module, but as you already mentioned, I was missing some information:

Looks like @BsAtHome is working on it.

> But the advantage of the halcmd parser is that you can use a dump to view other people's configurations. :)

The disadvantage is, it will break as soon someone decides to change the way the values are shown on the command line... ;-) The scripting mode `halcmd -s show` should probably be more stable but you use `halcmd show` as it seams, this can change any time.

You can still dump the config. Just dump the python structure as json. Then you can read it in again on your side to check. https://docs.python.org/3/library/json.html should do the trick.

grandixximo:
The values are raw C enum ints, and I think it's worth noting the module already exports matching constants (`hal.HAL_BIT`, `hal.HAL_IN`, ...), so `pin['TYPE'] == hal.HAL_BIT` already works today. IMO the only real gap here is readability: `TYPE 3` is just opaque when you print it.

I'd lean toward the `enum` route you linked, but with one distinction I think matters: type is a plain enum, but direction is really a bitmask. `HAL_IO == HAL_IN | HAL_OUT` and `HAL_RW == HAL_RO | HAL_WO` (see hal.h). So in my view the pythonic mapping is `enum.IntEnum` for type and `enum.IntFlag` for direction. I like that IntFlag composes in the repr (`HalPinDir.IN|OUT`), which reads well.

The nice part, I think, is that since both are int subclasses this stays fully backward compatible: `== hal.HAL_BIT` keeps working, and `json.dumps` still serializes them as the plain int, so dumping a config to json is unaffected. The only visible change is that printing gives you `HalType.BIT` instead of `3`.

My preference would be to define these enums once on the python side (in hal.py, or a shared module that both halmodule and your new halquery could import) and alias `hal.HAL_BIT = HalType.BIT` so there's a single source of truth, rather than building enum objects in the C extension. I think that keeps the C side simple and avoids duplicating the mapping across halmodule and halquery. But that's just my opinion, happy to hear if you'd rather keep it C-side.

BsAtHome:
Yes, the enum question is a bit harder because they are slightly disjoint between C/C++ and Python. An enum is part of a class in python. That is in itself is fine, but you need to set it from C, which is involved afaik. Secondly, you need to split the enum definition out into a separate module (like f.ex. haldefs) because both halmodule and halquery would need them.

Creating the a halquery module was very easy because I already have separated out all access to HAL in a query API. It also was a good test-case for the query API. There is no code uploaded in a branch, yet. This is my working tree that builds on #4099 (and that seems to be in limbo). There is, in my version, a very clear separation between hal_lib and everything else. No code can any longer include hal_priv.h (only halrmt still needs fixing). I'd rather not start backporting all that because it means distributing more mess to cleanup again later.

Let me see if I can fork off a branch so you can see what is living in it. Will take a some time...

@multigcs for dumping running configs we just add a command `haldump` that wraps the above example in a more structured way (encapsulated in json, yaml, plain text or whatever generic formats you deem appropriate).

multigcs:
> The disadvantage is, it will break as soon someone decides to change the way the values are shown on the command line... ;-) The scripting mode `halcmd -s show` should probably be more stable but you use `halcmd show` as it seams, this can change any time.

Oops, I didn't know about that option—I've changed it. Thanks!

BsAtHome:
> Do you have a branch to share?

This is the local tree pushed:
https://github.com/BsAtHome/linuxcnc/tree/wip_hal-types-and-isolation

Beware: this is WIP. Adjustments and fixing some ordering/naming issues are expected before the query API is finalized and fixed (have a look at the bottom of hal.h). Nearly all old hal types are gone and replaced, still working on the remaining few.

Tests pass, except... Warnings are not an error at the moment in the build. Rtai has several deprecation warnings that are exposed but I don't fully understand why they trigger. Uspace RIP builds should be warning free.
There is an awkward cppcheck issue that depends on cppcheck's version. It might be quelled by include ordering issue, but not yet sure why it happens. It doesn't happen on newer cppcheck.

BsAtHome:
> My preference would be to define these enums once on the python side (in hal.py, or a shared module that both halmodule and your new halquery could import) and alias `hal.HAL_BIT = HalType.BIT` so there's a single source of truth, rather than building enum objects in the C extension. I think that keeps the C side simple and avoids duplicating the mapping across halmodule and halquery. But that's just my opinion, happy to hear if you'd rather keep it C-side.

The enums can be defined on the implementation side of halmodule and be part of the `hal` namespace. I tested this and you then get enums like: `hal.Type.BIT` and `hal.Dir.OUT`. These are created as `enum.IntEnum`, which make them transparent in use and directly interoperable as integers.

The tricky part is that using them in `halquery` will require it to import `hal` (internally) as well. That should be no problem, but halmodule should first be "fixed" not to require a component before accessing the shared memory. Not requiring a component has already been fixed in my tree and can be upstreamed separately.

The part that still has me confused is how to instantiate the enum in C so I can pass it to `Py_BuildValue()`.

andypugh:
You might want to pause work on this, as I think that Petter has something in progress that integrates a graph directly in Halshow.

BsAtHome:
Using enums in Python are a separate issue from this PR. The discussion should be moved elsewhere.

petterreinholdtsen:
[Andy Pugh]
> You might want to pause work on this, as I think that Petter has
> something in progress that integrates a graph directly in Halshow.

I am working on a python rewrite of halshow, with graph support, yes.
The early draft is limping along, but I am not yet happy with the graph
layout.  It is worth noting that the implementation include extending
_hal.so (aka the python->shm integration) to return a bit more
information, which I suspect might have some overlap with the work you
are doing.

-- 
Happy hacking
Petter Reinholdtsen

BsAtHome:
> I am working on a python rewrite of halshow, with graph support, yes. The early draft is limping along, but I am not yet happy with the graph layout.

That would probably be an improvement.

> It is worth noting that the implementation include extending _hal.so (aka the python->shm integration) to return a bit more information, which I suspect might have some overlap with the work you are doing.

Please do _not_ extend the _hal.shm part. It is a mistake to have it in there at all. And do not extend any access to hal's internal structures. Having all these utility programs hack into private constructs is a maintenance nightmare and every internal hal change needs to be checked and duplicated across many external programs and places.
All access to HAL internals should never be done directly. Hal_lib's internals are private to hal_lib. That is the whole point of me making that query API, which gives you the tools without having to mess in hal's internals.

petterreinholdtsen:
[BsAtHome]
> That would probably be an improvement.

The current edition is now available from
<URL: https://github.com/petterreinholdtsen/linuxcnc/tree/reimplement-halshow-python >.
I believe it got all the features of the tcl edition, and a new graph
tab.  If you find some missing feature, please let me know.  The build
system update to switch out the tcl edition with the python edition is
not yet in place.

The branch contain two commits, one creating a replacement halshow
implemented in python, the other adding graph support to this new pythin
edition.

> Please do _not_ extend the _hal.shm part. It is a mistake to have it
> in there at all.

It speed up the HAL input gathering process significantly, so I believe
it is a valuable tool to have in the tool box.  When that is said, I
have made sure the new halshow also work well without any changes to
_hal.so (and fall back to halcmd if the changed edition is not
available), to stay backwards compatible with the 2.9 edition.

-- 
Happy hacking
Petter Reinholdtsne

BsAtHome:
After discussion we've decided not to go this path to use halcmd for PlantUML data generation.

You have already started work on #4231 to replace the old Tcl based halshow and modernize it using Python. We think this is a good idea and the right path. It updates and modernizes the code base. Including/adding a graphing option in the new halshow would be a welcome addition.

PR #4199, opened 2026-06-24
Mesa Hostmot2 Xenomai4 OOB networking support

hdiethelm:
This PR adds Xenomai4 EVL out of band networking support for Mesa Hostmot2.

Out of band networking is basically a fast path inside the xenomai real time kernel that enables networking without involving the normal kernel.

~Due to some users might want to use Xenomai4 with the standard kernel networking, I decided to create a new component called `hm2_eth_evl` while `hm2_eth` behaves exactly like before.~

~Common code is left in `hm2_eth.c` and network specific code is moved to `hm2_eth_net.c` and `hm2_eth_net_evl.c`.
The linker is used to link `hm2_eth_evl` and `hm2_eth` with two different network implementations and the same common code.~

To select the mode, you can use: `board_rtnet=posix` or `board_rtnet=evl`, where the default is posix, so it is a non-breaking change.

With `board_rtnet=evl`, `initf hm2_eth.realtime-init servo-thread` needs to be added to the .ini file to initialize the realtime network inside the realtime context.

A few changes in the existing code where performed:
- Pass board instead of board->fd to network functions.
- Don't use a cumbersome void pointer casting one line function for SIOCSARP
- The ifname is fetched in init_board() and stored in board, so the install_firewall_perinterface() loop is simplifyed and I can reuse the name inside the evl implementation
- The ip address is stored in board because I need it later
- Print which board is connected, nice if you have more than one board and it throws an error

Things still open:
- I have to set the Xenomai4 C and LD flags to all modules, I did not find a way to only set it for a specific module in the Makefile. It will only change anything if you have libevl installed.
- There are more global symbols. Would it make sense to rename `fetch_hwaddr` to `hm2_fetch_hwaddr` for example to avoid conflicts?
- ~Loading `hm2_eth_evl` and `hm2_eth` at the same time would probably create a runtime linker issue which generates undefined behavior.~

It's a bit hard to review due to moved code. `git diff master:src/hal/drivers/mesa-hostmot2/hm2_eth.c hm2_eth_oob_v5:src/hal/drivers/mesa-hostmot2/hm2_eth_net_posix.c` helps.

This three PR's contain fixes for bugs discovered here but are not directly connected to OOB networking:
https://github.com/LinuxCNC/linuxcnc/pull/4217
https://github.com/LinuxCNC/linuxcnc/pull/4218
https://github.com/LinuxCNC/linuxcnc/pull/4240

PR #4217, opened 2026-07-01
gtk_mesa_tests: Fix calculation in %

hdiethelm:
https://github.com/LinuxCNC/linuxcnc/pull/4082 changed the time values from cycles to ns. 

This broke the calculation of % values and the CPU clock is not needed any more.

Discovered here: https://github.com/LinuxCNC/linuxcnc/pull/4199

TBD: Remove the CPU Speed read fully? I would say yes.

### DIFF
--- a/lib/python/gladevcp/builtin-panels/gtk_mesa_tests/gtk_mesa_tests.glade
+++ b/lib/python/gladevcp/builtin-panels/gtk_mesa_tests/gtk_mesa_tests.glade
@@ -2,16 +2,6 @@
 <!-- Generated with glade 3.40.0 -->
 <interface>
   <requires lib="gtk+" version="3.24"/>
-  <object class="GtkAdjustment" id="adj_cpu_speed_nic">
-    <property name="upper">10000000</property>
-    <property name="step-increment">1</property>
-    <property name="page-increment">10</property>
-  </object>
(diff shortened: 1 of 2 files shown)


zz912:
> TBD: Remove the CPU Speed read fully? I would say yes.

I say yes too.

hdiethelm:
^Rebased to master

grandixximo (commented):
Thanks for this. I checked the calculation and it is correct: tmax and period are both in ns from HAL, so the CPU clock is indeed no longer needed.

I built and ran the panel from the branch. Python compiles and the glade file validates, so no blockers from my side.

One nitpick: the two GtkAdjustment objects `adj_cpu_speed_nic` and `adj_cpu_speed_servo` at the top of gtk_mesa_tests.glade (lines 4-13) are now orphaned; the spin buttons that referenced them were removed. Please delete them.

hdiethelm:
> One nitpick: the two GtkAdjustment objects `adj_cpu_speed_nic` and `adj_cpu_speed_servo` at the top of gtk_mesa_tests.glade (lines 4-13) are now orphaned; the spin buttons that referenced them were removed. Please delete them.

Thanks, fixed + tested.

1. force push: Fix
2. force push: Rebase master

PR #4218, opened 2026-07-01
hm2_eth: read / write buffer checks

hdiethelm:
This generated a segfault when eth_socket_send() fails due to
accumulating data in the buffer.

Issue found while working on https://github.com/LinuxCNC/linuxcnc/pull/4199 but this doesn't really fit that PR.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/hm2_eth.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_eth.c
@@ -1234,12 +1234,11 @@ static int hm2_eth_send_queued_writes(hm2_lowlevel_io_t *this) {
     send = eth_socket_send(board->sockfd, (void*) &board->write_packet, board->write_packet_ptr - board->write_packet, 0);
     if(send < 0) {
         LL_PRINT("ERROR: sending packet: %s\n", strerror(errno));
-        return 0;
     }
     t1 = rtapi_get_time();
     LL_PRINT_IF(debug, "enqueue_write(%d) : PACKET SEND [SIZE: %d | TIME: %llu]\n", board->write_cnt, send, t1 - t0);
     board->write_packet_ptr = board->write_packet;
-    return 1;
+    return send < 0 ? 0 : 1;
 }
 
 static int hm2_eth_enqueue_write(hm2_lowlevel_io_t *this, rtapi_u32 addr, const void *buffer, int size) {
--- a/src/hal/drivers/mesa-hostmot2/hm2_eth.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_eth.c
@@ -31,6 +31,7 @@
 #include <stdarg.h>
 #include <sys/types.h>
 #include <sys/wait.h>
+#include <assert.h>
 
 #include <rtapi_slab.h>
 #include <rtapi_ctype.h>
@@ -993,9 +994,30 @@ static int hm2_eth_read(hm2_lowlevel_io_t *this, rtapi_u32 addr, void *buffer, i
     return 1;  // success
 }
 
+static void hm2_eth_reset_queued_reads(hm2_eth_t *board){
+    board->read_packet_ptr = board->read_packet;
+    board->queue_reads_count = 0;
+    board->queue_buff_size = 0;
+}
+
 static int hm2_eth_send_queued_reads(hm2_lowlevel_io_t *this) {
     hm2_eth_t *board = this->private;
     int send;
+
+    //Check size we are going to write
+    size_t read_packet_size = board->read_packet_ptr - board->read_packet;
+    if (read_packet_size + 3*sizeof(lbp16_cmd_addr) + sizeof(board->read_cnt) > sizeof(board->read_packet)) {
+        LL_PRINT("ERROR: send_queued_reads: buffer full, dropping all data\n");
+        //We need to drop the data to recover
+        hm2_eth_reset_queued_reads(board);
+        return 0;
+    }
+    if (board->queue_reads_count + 2 > MAX_ETH_READS) {
+        LL_PRINT("ERROR: send_queued_reads: queue_reads full, dropping all data\n");
+        //We need to drop the data to recover
+        hm2_eth_reset_queued_reads(board);
+        return 0;
+    }
 
     // read (low 16 bits of) last write number from space 4 address 0010
     LBP16_INIT_PACKET4(*(lbp16_cmd_addr*)(board->read_packet_ptr), CMD_READ_COMM_CTRL_ADDR16(1), 0x8);
@@ -1010,8 +1032,9 @@ static int hm2_eth_send_queued_reads(hm2_lowlevel_io_t *this) {
     // write then read back space 4 scratch register at 0010 to verify we got the right receive packet
     LBP16_INIT_PACKET4(*(lbp16_cmd_addr*)(board->read_packet_ptr), CMD_WRITE_TIMER_ADDR16_INCR(2), 0x10);
     board->read_packet_ptr += sizeof(lbp16_cmd_addr);
-    *(uint32_t*)board->read_packet_ptr = board->read_cnt;
-    board->read_packet_ptr += sizeof(uint32_t);
+    static_assert(sizeof(board->read_cnt)==4, "board->read_cnt must be 4 bytes");
+    memcpy(board->read_packet_ptr, &board->read_cnt, sizeof(board->read_cnt));
+    board->read_packet_ptr += sizeof(board->read_cnt);
 
     LBP16_INIT_PACKET4(*(lbp16_cmd_addr*)(board->read_packet_ptr), CMD_READ_TIMER_ADDR16_INCR(4), 0x10);
     board->read_packet_ptr += sizeof(lbp16_cmd_addr);
@@ -1091,9 +1114,7 @@ do_recv_packet:
         i++;
     } while (recv != board->queue_buff_size && t2 < read_deadline);
     if(recv != board->queue_buff_size) {
-        board->read_packet_ptr = board->read_packet;
-        board->queue_reads_count = 0;
-        board->queue_buff_size = 0;
+        hm2_eth_reset_queued_reads(board);
         if(!record_soft_error(board)) return 0;
         return -EAGAIN;
     }
@@ -1107,9 +1128,7 @@ do_recv_packet:
     if(board->confirm_read_cnt != board->read_cnt && t2 < read_deadline)
         goto do_recv_packet;
 
-    board->read_packet_ptr = board->read_packet;
-    board->queue_reads_count = 0;
-    board->queue_buff_size = 0;
+    hm2_eth_reset_queued_reads(board);
 
     int result = 1;
     // (this means that one in 2^32 lost writes will not be diagnosed,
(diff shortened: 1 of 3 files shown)


hdiethelm:
^Rebased to master

hdiethelm:
@BsAtHome ping

BsAtHome:
Do see the problem. But it also makes me question things a bit.
Why does the data need to be dropped? Is that about timing? Dropping a write command may be much more serious than dropping a read (well, that is round-trip packet write->packet read).
Secondly, what is the maximum packet size? Does local or remote not support (8..16kB) jumbo frames or is that not supposed to happen?

BsAtHome:
BTW, if I'm barking up the wrong tree, just let me know and try to make me understand the actual story.

hdiethelm:
As much as I understood from the code, the enqueue queues commands. send_queue finalizes the package and sends it. If this finalizing fails, there is no other way than drop all.

All four size checks should never trigger except something else is wrong. In that case, better show an error and try recover than segfault and gone.

This PR only avoids segfaults. If the size check fails, this would need fixing somewhere else.

In my case, eth_socket_send() had a bug for Xenomai4, so it returned -1. In the next cycle, the data to be sent was 2x the size until segfault. Took me some time to figure out that the reason rtapi_app crashed was here.

An additional barrier would be to drop the data if send failed. But that might have side effects. So I opted to only add checks.

In my case, the packages are ~150 bytes, far away from the max. size.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
```diff
@@ -1185,8 +1201,16 @@ static int hm2_eth_send_queued_writes(hm2_lowlevel_io_t *this) {
     hm2_eth_t *board = this->private;
 
     board->write_cnt++;
-    // XXX this is missing a check for exceeding the maximum packet size!
     lbp16_cmd_addr *packet = (lbp16_cmd_addr *) board->write_packet_ptr;
+
+    if(board->write_packet_size + sizeof(*packet) + 4 > sizeof(board->write_packet)){
+        LL_PRINT("ERROR: send_queued_writes: buffer full, droping all data\n");
```
BsAtHome: Where does the "magic" number 4 come from?

I guess it has something to do with `board->write_cnt` (used further down). If so, then the value 4 should probably be a sizeof (and the type of write_cnt checked that it is actually the correct type of size 4 (rtapi_s32 or rtapi_u32, ah well, int32_t or uint32_t because this is exclusively uspace code)).

src/hal/drivers/mesa-hostmot2/hm2_eth.c
```diff
@@ -997,6 +997,16 @@ static int hm2_eth_send_queued_reads(hm2_lowlevel_io_t *this) {
     hm2_eth_t *board = this->private;
     int send;
 
+    size_t read_packet_size = board->read_packet_ptr - board->read_packet;
+    if(read_packet_size + 3*sizeof(lbp16_cmd_addr) + sizeof(uint32_t) > sizeof(board->read_packet)){
+        LL_PRINT("ERROR: send_queued_reads: buffer full, droping data\n");
```
BsAtHome: The 3 times sizeof(lbp16_cmd_addr) I can see. What is the additional uint32_t doing there? Isn't that supposed to be board->read_cnt and then the sizeof(uint32_t) further down should probably be sizeof(board->read_cnt)?

BsAtHome:
> As much as I understood from the code, the enqueue queues commands. send_queue finalizes the package and sends it. If this finalizing fails, there is no other way than drop all.

Yes, it combines the queued reads/writes for sending in one packet.

> All four size checks should never trigger except something else is wrong. In that case, better show an error and try recover than segfault and gone.
> This PR only avoids segfaults. If the size check fails, this would need fixing somewhere else.

Well, segfaulting is never the right answer... So, yes, preventing it is good.

Your "fix" does seem to be correct in that it drops the command(s). Also the read command resolution queue is cleared properly on fail.

> In my case, eth_socket_send() had a bug for Xenomai4, so it returned -1. In the next cycle, the data to be sent was 2x the size until segfault. Took me some time to figure out that the reason rtapi_app crashed was here.

There may also be a segfault in the read resolution queue. Is saves the address of the buffer(s) where the read data needs to be copied into. If you double up, then that queue is probably going to contain the same address for resolution, which is definitely wrong. So, dropping the queue is the correct action.

> An additional barrier would be to drop the data if send failed. But that might have side effects. So I opted to only add checks.

Well, you actually do drop the read queue. So, that data is dropped. The write queue, well, that is a different question. You might want to have a look at the code that coalesces the write-queue and actually sends.

> In my case, the packages are ~150 bytes, far away from the max. size.

Yeah, should never be a problem.

hdiethelm:
You are right, why not improve the code around these sections a bit.

Changes:
- function hm2_eth_reset_queued_reads() added instead of copy&paste
- also check board->queue_reads_count
- less magic numbers
- static_assert for sizeof(read_cnt) / sizeof(write_cnt)
- use memcpy for read_cnt: no guarantee of alignment / same is already in for write_cnt
- write_cnt++ after size check (was a bug from my side)
- remove write_packet_size (was redundant)

Tested again. By dropping packet 10000...10020 in eth_socket_send(), i can trigger 2 of 5 error cases:
hm2_eth: ERROR: enqueue_write: buffer full
hm2_eth: ERROR: send_queued_writes: buffer full, dropping all data

With `setp hm2_[HOSTMOT2](BOARD).0.packet-error-limit 1000` it recovers cleanly and the machine continues working.

max packet size = 340 / max queue_reads_count = 10 on my setup.
This will be different for different configs. At leas now, there is an error message instead of a crash if somebody manages to exceed 1400 bytes.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
hdiethelm: Changed to `sizeof(board->read_cnt)` + `static_assert(sizeof(board->read_cnt)==4, "board->read_cnt must be 4 bytes");`

src/hal/drivers/mesa-hostmot2/hm2_eth.c
hdiethelm: Changed to `sizeof(board->write_cnt)` all over + `static_assert(sizeof(board->write_cnt)==4, "board->write_cnt must be 4 bytes");`

hdiethelm:
> > In my case, eth_socket_send() had a bug for Xenomai4, so it returned -1. In the next cycle, the data to be sent was 2x the size until segfault. Took me some time to figure out that the reason rtapi_app crashed was here.
> 
> There may also be a segfault in the read resolution queue. Is saves the address of the buffer(s) where the read data needs to be copied into. If you double up, then that queue is probably going to contain the same address for resolution, which is definitely wrong. So, dropping the queue is the correct action.
> 
> > An additional barrier would be to drop the data if send failed. But that might have side effects. So I opted to only add checks.
> 
> Well, you actually do drop the read queue. So, that data is dropped. The write queue, well, that is a different question. You might want to have a look at the code that coalesces the write-queue and actually sends.

In hm2_eth_receive_queued_reads(), the read queue is reset when the read fails, looks good. Probably the reason I can not trigger a read overflow.

However, in hm2_eth_send_queued_writes(), the queue is not reset if eth_socket_send() fails. This is probably the initial reason for the buffer overflow. This would fix it. I have to test if it works.

```patch
diff --git a/src/hal/drivers/mesa-hostmot2/hm2_eth.c b/src/hal/drivers/mesa-hostmot2/hm2_eth.c
index 308e442100..2ea22245dc 100644
--- a/src/hal/drivers/mesa-hostmot2/hm2_eth.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_eth.c
@@ -1232,13 +1232,13 @@ static int hm2_eth_send_queued_writes(hm2_lowlevel_io_t *this) {
     
     t0 = rtapi_get_time();
     send = eth_socket_send(board->sockfd, (void*) &board->write_packet, board->write_packet_ptr - board->write_packet, 0);
+    t1 = rtapi_get_time();
+    LL_PRINT_IF(debug, "enqueue_write(%d) : PACKET SEND [SIZE: %d | TIME: %llu]\n", board->write_cnt, send, t1 - t0);
+    board->write_packet_ptr = board->write_packet;
     if(send < 0) {
         LL_PRINT("ERROR: sending packet: %s\n", strerror(errno));
         return 0;
     }
-    t1 = rtapi_get_time();
-    LL_PRINT_IF(debug, "enqueue_write(%d) : PACKET SEND [SIZE: %d | TIME: %llu]\n", board->write_cnt, send, t1 - t0);
-    board->write_packet_ptr = board->write_packet;
     return 1;
 }
```

Hmm, I found an other funny looking line:
https://github.com/LinuxCNC/linuxcnc/blame/3e8a44f30acae0cff1eff158f96ebcc6a6ad6220/src/hal/drivers/mesa-hostmot2/hostmot2.c#L1628
No big deal due to hm2_finish_read() will fail anyway when hm2_queue_read() fails...

> 
> > In my case, the packages are ~150 bytes, far away from the max. size.
> 
> Yeah, should never be a problem.

hdiethelm:
So, the last commit fixes root cause of the growing write buffer and the missing return value check.

Tested: Now there is no buffer full error any more. The size checks make still sense but they are not triggered any more.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
```diff
@@ -1184,25 +1214,31 @@ static int hm2_eth_send_queued_writes(hm2_lowlevel_io_t *this) {
...
     t0 = rtapi_get_time();
-    send = eth_socket_send(board->sockfd, (void*) &board->write_packet, board->write_packet_size, 0);
+    send = eth_socket_send(board->sockfd, (void*) &board->write_packet, board->write_packet_ptr - board->write_packet, 0);
+    t1 = rtapi_get_time();
+    LL_PRINT_IF(debug, "enqueue_write(%d) : PACKET SEND [SIZE: %d | TIME: %llu]\n", board->write_cnt, send, t1 - t0);
+    board->write_packet_ptr = board->write_packet;
     if(send < 0) {
         LL_PRINT("ERROR: sending packet: %s\n", strerror(errno));
         return 0;
     }
```
BsAtHome: You cannot access `errno` after you printed or called anything that may (re)set it. You need to save its value after eth_socket_send(), do your stuff, and then in the if-clause use the saved errno value.

You are not setting board->write_packet_size to zero anymore, never. Is that correct?

BsAtHome:
The "funny" line is a real bug :-)

src/hal/drivers/mesa-hostmot2/hm2_eth.c
hdiethelm: board->write_packet_size is gone, I use now `board->write_packet_ptr - board->write_packet`, same as for read.
errno: Fixed

src/hal/drivers/mesa-hostmot2/hm2_eth.c
```diff
@@ -1184,25 +1214,33 @@ static int hm2_eth_send_queued_writes(hm2_lowlevel_io_t *this) {
...
     t1 = rtapi_get_time();
     LL_PRINT_IF(debug, "enqueue_write(%d) : PACKET SEND [SIZE: %d | TIME: %llu]\n", board->write_cnt, send, t1 - t0);
     board->write_packet_ptr = board->write_packet;
-    board->write_packet_size = 0;
+    if(send < 0) {
+        return 0;
+    }
     return 1;
 }
```
BsAtHome, line 1243: Slightly nitpicking, when the last action is a simple return on a condition, then maybe:
```C
return send < 0 ? 0 : 1;
```
?

hdiethelm:
> The "funny" line is a real bug :-)

Just one that will never creates any issues. Fixed any way.

By:
```patch
diff --git a/src/hal/drivers/mesa-hostmot2/hm2_eth.c b/src/hal/drivers/mesa-hostmot2/hm2_eth.c
index 56c4c126de..bce9fc7dda 100644
--- a/src/hal/drivers/mesa-hostmot2/hm2_eth.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_eth.c
@@ -1114,7 +1114,7 @@ do_recv_packet:
         i++;
     } while (recv != board->queue_buff_size && t2 < read_deadline);
     if(recv != board->queue_buff_size) {
-        hm2_eth_reset_queued_reads(board);
+        //hm2_eth_reset_queued_reads(board);
         if(!record_soft_error(board)) return 0;
         return -EAGAIN;
     }
```
I was also able to trigger:
```
hm2_eth: ERROR: enqueue_read: queue_reads full
hm2/hm2_7i96s.0: TRAM read error! (addr=0x2100, size=16, iter=4287)
hm2_eth: ERROR: send_queued_reads: queue_reads full, dropping all data
```
Recovered nicely.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
hdiethelm: Yes, why not. Fixed.

PR #4223, opened 2026-07-02
hal: split 'toggle2nist.comp' for toggle and momentary inputs

Sigma1912:
Creates different components for handling toggle buttons and momentary buttons with component names that actually reflect the functionality.

- Restores the legacy functionality (as in 2.9) of 'toggle2nist' with added input debounce.
- Fixes the toggle2nist documentation so it matches actual (legacy) behavior.
- Adds a new 'momentary2nist' component that does what the old 'toggle2nist' docs described.

***Background*** 
 
Before #3193 'toggle2nist' component docs described the functionality as converting
_momentary_ inputs to nist outputs but the actual functionality was converting _toggled_
inputs to nist outputs. This required using a preceding 'toggle' component to match 
documented behavior.           

The fix in #3193 changed the behavior to match the documentation (and added debouncing
to the inputs). While this fixed the issues reported on the forum it breaks functionality
for hal files that use the preceeding 'toggle' component. Also the name 'toggle2nist' does
no longer reflect the actual function.

### DIFF
--- a/src/hal/components/momentary2nist.comp
+++ b/src/hal/components/momentary2nist.comp
@@ -1,18 +1,18 @@
-component toggle2nist "toggle button to nist logic";
+component momentary2nist "momentary button to nist logic";
 
 description
 """
-Toggle2nist can be used with a latching push button 
+Momentary2nist can be used with a momentary push button
 to control a device that has separate on and off inputs
 and an is-on output.
 A debounce delay in cycles can be set for 'in'. (default = 2)
 
 * On a rising edge on pin *in* when *is-on* is low: It sets *on* until *is-on* becomes high.
-* On a falling edge on pin *in* when *is-on* is high: It sets *off* until *is-on* becomes low.
+* On a rising edge on pin *in* when *is-on* is high: It sets *off* until *is-on* becomes low.
 
 ....
-       ┐     ┌─────────────────────────────┐
-in   : └─────┘                             └───────────────────────
+       ┐     ┌─────xxxxxxxxxxxx┐           ┌─────xxxxxxxxxxxx┐
+in   : └─────┘     xxxxxxxxxxxx└───────────┘     xxxxxxxxxxxx└─────
 
        ┐     ┌───────────┐
 on   : └─────┘           └─────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


hansu:
What about a pin to change the mode from momentary to toggle button behaviour?

And what exactly is the definition of "nist outputs" or "nist logic"? I haven't found anything about that.

Sigma1912:
> What about a pin to change the mode from momentary to toggle button behaviour?

The thing that made me lean towards a separate component is that `toggle2nist`  is a very specific name, which we cannot really change to anything more generic.   


> And what exactly is the definition of "nist outputs" or "nist logic"? I haven't found anything about that.

All I can tell you is what is in the man page: 

`a device that has separate on and off inputs and an is-on output.`

src/hal/components/momentary2nist.comp
```diff
@@ -0,0 +1,77 @@
+component momentary2nist "momentary button to nist logic";
+
+description
+"""
+Monemtary2nist can be used with a momentary push button
```
hansu: typo ;-)

hansu:
> > And what exactly is the definition of "nist outputs" or "nist logic"? I haven't found anything about that.
> 
> All I can tell you is what is in the man page:
> 
> `a device that has separate on and off inputs and an is-on output.`

So if you don't know what the name of your component means, maybe it's a chance to name it different? (Even though the name is the not so close to toggl2nist)

What about something like momentary_toggle or momentary2onoff ?

Sigma1912:
I get your point but I don't think we should start calling the same things differently. Given that the two components both provide a logic conversion that only differs in the input logic it seems reasonable to retain 'nist' since that is what the output logic has been called for years.

src/hal/components/momentary2nist.comp
Sigma1912: Thank you. Fixed.

grandixximo:
I thought I was the only one not knowing that nist-logic was a thing, but apparently it just is not a thing outside of the toggle2nist component.

Only reference I could find on the web is on the blogpost of the component release from 2008

https://www.anderswallin.net/2008/04/toggle2nist/

According to the author it means

> there is one on-switch/signal to switch a thing on, one off-switch/signal to switch it off, and one indicator-signal to tell you the state.

Had AI guess an acronym that matches the intended meaning

Negate => off
Initiate => on
Status => indicator
Toggle => mechanism

Does not really make sense, I think possibly the original author conflated the fact that EMC was a NIST (National Institute of Standards and Technology) project with the way its IO work.

Anyhow we are kinda stuck with it, might as well stick to it lol

Sigma1912:
What is the general mood about this PR?  Should I update it to getter/setter?

grandixximo:
Please do, we are falling behind on merging cause the dev meetings have been skipped, holiday season I guess...

PR #4224, opened 2026-07-02
Dropped no longer used gtk.glade import from versa_probe_screen.py.

petterreinholdtsen:


andypugh:
Why is this targeted at 2.9? Does it fix a bug?

petterreinholdtsen:
[andypugh]
> Why is this targeted at 2.9? Does it fix a bug?

I targeted it towards 2.9 to reduce the number of dependencies of the
old and obsolete gtk edition (now only classicladder) in 2.9, in case
the idea to switch to 2.10 in Debian unstable take a while, and we end
up having to update the 2.9 edition in Debian with a rewritten
classicladder using the newer gtk version.  As it is a single line
change and only remove a not used dependency, I believed it was a good
idea to get it in as early as possible in the "port forward" stack.

With this change, there is only one component of LinuxCNC 2.9 trying to
use the old gtk version, and in this component it is optional, thus also
allowing another option to drop the ladder editor, as I mentioned in
<URL: https://bugs.debian.org/1089809 >.

-- 
Happy hacking
Petter Reinholdtsen

PR #4227, opened 2026-07-03
build: keep comp manpage adoc rules out of the kbuild pass

### LINKED ISSUE #4226: docs? RTAI compile is confused about some components
Compiling RTAI seems to be confused in a Submakefile with the following messages:
```
src/hal/components/Submakefile:65: target 'hal/drivers/mesa_7i65.comp' doesn't match the target pattern
src/hal/components/Submakefile:65: target 'hal/drivers/mesa_uart.comp' doesn't match the target pattern
src/hal/components/Submakefile:65: target 'hal/drivers/pcl720.comp' doesn't match the target pattern
src/hal/components/Submakefile:65: target 'hal/drivers/pluto_servo.comp' doesn't match the target pattern
src/hal/components/Submakefile:65: target 'hal/drivers/pluto_step.comp' doesn't match the target pattern
src/hal/components/Submakefile:65: target 'hal/drivers/serport.comp' doesn't match the target pattern
```

It seems to be related to building the documentation. There is a difference what is included depending uspace or other system, but I'm not sure what is going on.

@grandixximo you have been around this stuff. Do you know or have an idea?

grandixximo:
In the KERNELRELEASE (kbuild) pass `COMPS` is rebuilt from a wildcard that mixes `hal/components` and `hal/drivers`. The manpage adoc targets were derived from `COMPS` outside the KERNELRELEASE guard, so in that pass they picked up `hal/drivers/*.comp` entries that fail the `hal/components/%.comp` static pattern rule, breaking RTAI builds.

Move the manpage adoc vars and rules inside the non-kbuild `else` branch, where manpage generation belongs. uspace builds are unaffected.

Fixes #4226.

### DIFF
--- a/src/hal/components/Submakefile
+++ b/src/hal/components/Submakefile
@@ -50,10 +50,12 @@ ifeq ($(TRIVIAL_BUILD),no)
 -include $(patsubst %.comp, objects/%.mak, $(COMPS) $(COMP_DRIVERS))
 Makefile: $(patsubst %.comp, objects/%.mak, $(COMPS) $(COMP_DRIVERS))
 endif
-endif
 
-obj-m += $(patsubst hal/drivers/%.comp, %.o, $(patsubst hal/components/%.comp, %.o, $(COMPS) $(COMP_DRIVERS)))
-
+# Manpage generation is a top-level (non-kbuild) concern.  Keep it inside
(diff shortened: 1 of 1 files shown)


grandixximo:
@BsAtHome does this fix #4226 for you?

BsAtHome:
Yes, it is fixed.

Issue #4216, opened 2026-07-01
2.9.9: changes to parameters 5401-5409 break existing probing routines


Sigma1912:
As reported on the forum:
https://forum.linuxcnc.org/38-general-linuxcnc-questions/58927-g38-2-weird-behaviour#347435

While https://github.com/LinuxCNC/linuxcnc/commit/725231f348c374816074be25cde3fd8e55ef030d makes the parameters behave as documented it has broken the behavior of probing routines in the wild.

Maybe a new set of parameters might need to be considered instead of altering the behavior of the established ones.

Sigma1912:
cc @grandixximo

grandixximo:
Sure, they can be shifted over, but I don't think they should, user knows he should load the offsets if he wants them on, he just doesn't want to do it, old docs also stated to load with G43, nowhere it ever said that it should contain the offset that _will_ be loaded when you send G43, he was using them incorrectly IMO

andypugh:
The way I see it, if users have found that their routines did not follow the docs, and worked around that, and now we have broken their routines, that's our fault, not theirs. 
I am feeling that this is a breaking change that I shouldn't have included in 2.9. 
But it was also a bug fix. 
So, going forward, we probably need a policy for bug fixes that are breaking changes. 
(and that is probably "leave the bug alone" unfortunately) 

Anyway, the question is whether I need to re-release a 2.9.9, or a 2.9.10 or whether we try to cover this in the release notes. 

Is this going to break tools, ruin parts and break machines? I suspect that it is. 

Note that I haven't made a release announcement yet, though I see that RodW did on Facebook.

grandixximo:
users already updated, so you need go for 2.9.10 otherwise they will not get the update.
I'm preparing the patch, moving over the new behavior to #5081-#5089
what about master? same thing, we are keeping the old behavior on the old parameters, and new behavior in the new parameters?
Are deprecation messages from gcode something we want to try? I thought about it, but I think better not...

grandixximo:
@wucke13 we are putting back the bug on the default parameters, will have to use #5081-#5089 for the new behavior.

grandixximo:
> The way I see it, if users have found that their routines did not follow the docs, and worked around that, and now we have broken their routines, that's our fault, not theirs.

We should not break old behavior even when buggy, that sucks but is the reality we live in...

Sigma1912:
Thanks for the quick response.

[...]

PR #4228, opened 2026-07-03
interp: split loaded vs applied tool offset params (#5401-#5409, #5081-#5089)

grandixximo:
Forward-port of #4219 (merged to 2.9) to master, fixing #4216.

#5401-#5409 again hold the loaded tool's stored offset (refreshed on M6 / G10 / init), restoring the behavior probing routines in the wild rely on. The offset actually applied to motion by G43/G43.1/G43.2 moves to the new parameters #5081-#5089, and is zeroed by G49.

Two commits:
- Cherry-pick of the runtime + docs fix from #4219.
- Revert of the test-assertion commit from #3996 (49ddfe633b), which had rewritten the tool-offset assertions for the applied semantics. The restored assertions match the loaded-offset behavior.

Full test suite passes locally (291 tests, 0 failed).


wucke13:
@grandixximo I think this did miss to update the absolute Z position formula example on the G38.n gcode?

grandixximo:
I did not look at the absolute Z position formula, because I had a vague memory of having separated the formula out from `#5403`, but reading your message gave me a bit of a scare that I might have missed something, fortunately I did not miss, it was related before my changes in 6da626e737
the new official formula is

```
G38.2 Z-100 F100
#<zmachine> = [#5063 + #5023 - #5422]
```
In both 2.9 and 2.10

The new one above is simpler than this following initial proposal we had
```
#<zmachine> = [#5063 + #[5203 + #5220 * 20] + #5213 * #5210 + #5403]
```
Which is not working well anymore, has to be changed to either
```
#<zmachine> = [#5063 + #[5203 + #5220 * 20] + #5213 * #5210 + #5083]
```
or the new one from the new docs
```
G38.2 Z-100 F100
#<zmachine> = [#5063 + #5023 - #5422]
```
So it should all be good, nothing missed, just gave me a heart attack lol

wucke13:
Ah, sorry for the scare --- I didn't read careful enough. Two additional notes:

- At first I thought `#5422` was a typo, but indeed the `#542x` parameter block starts with 0, not 1, so Z is 2, not 3.
- In the current docs, it reads:
  > 5420-5428 - Current relative position in the active coordinate system including all offsets and in the current program units for X, Y, Z, A, B, C, U, V & W, volatile. 
  
  This is indeed the **only** parameter that is described as _relative_. I wonder what the word conveys there, I would assume that this is an absolute (but work-, as opposed to machine-) coordinate system? If indeed it is relative, then, relative to _what_?

grandixximo:
I think it's in contrast to what is referred to as "absolute machine position" which is what you reference with G53. So it is relative to the active coordinate system if I interpret it correctly

PR #4231, opened 2026-07-09
Draft: Reimplement halshow in Python

petterreinholdtsen:
Replace the tcl/tk implementation of halshow with a Python/Qt implementation, and add new graph support using graphviz to show the interconnection in HAL between pins and signals.

### DIFF
--- a/scripts/halshow.py
+++ b/scripts/halshow.py
@@ -25,11 +25,11 @@ from pathlib import Path
 
 # Determine locale directory for translations (same domain as Tcl halshow)
 for _p in sys.path:
-    if '/lib/python' in _p:
-        if '/usr' in _p:
-            _LOCALEDIR = 'usr/share/locale'
+    if "/lib/python" in _p:
+        if "/usr" in _p:
+            _LOCALEDIR = "usr/share/locale"
         else:
-            _LOCALEDIR = os.path.join(_p.split("/lib")[0], 'share', 'locale')
+            _LOCALEDIR = os.path.join(_p.split("/lib")[0], "share", "locale")
         break
 else:
     _LOCALEDIR = None
@@ -39,19 +39,72 @@ _translate = _  # alias so it survives tuple-unpacking shadows on `_`
 
 
 from qtpy.QtWidgets import (
-    QApplication, QMainWindow, QWidget, QVBoxLayout, QHBoxLayout, QSplitter,
-    QTabWidget, QTreeWidget, QTreeWidgetItem, QTextBrowser, QLineEdit,
-    QPushButton, QLabel, QCheckBox, QScrollArea, QFrame, QFileDialog,
-    QMessageBox, QInputDialog, QMenu, QAction, QHeaderView, QApplication as app,
-    QSizePolicy, QGraphicsView, QGraphicsScene, QGraphicsRectItem, QGraphicsTextItem,
-    QGraphicsEllipseItem, QGraphicsPathItem, QGraphicsPolygonItem, QGraphicsItem, QDialog, QDialogButtonBox,
-    QFormLayout, QColorDialog
+    QApplication,
+    QMainWindow,
+    QWidget,
+    QVBoxLayout,
+    QHBoxLayout,
+    QSplitter,
+    QTabWidget,
+    QTreeWidget,
+    QTreeWidgetItem,
+    QTextBrowser,
+    QLineEdit,
+    QPushButton,
+    QLabel,
+    QCheckBox,
+    QScrollArea,
+    QFrame,
+    QFileDialog,
+    QMessageBox,
+    QInputDialog,
+    QMenu,
+    QAction,
+    QHeaderView,
+    QApplication as app,
+    QSizePolicy,
+    QGraphicsView,
+    QGraphicsScene,
+    QGraphicsRectItem,
+    QGraphicsTextItem,
+    QGraphicsEllipseItem,
+    QGraphicsPathItem,
+    QGraphicsPolygonItem,
+    QGraphicsItem,
+    QDialog,
+    QDialogButtonBox,
+    QFormLayout,
+    QColorDialog,
 )
-from qtpy.QtCore import Qt, QTimer, QSize, Signal, Slot, QPointF, QRectF, QThread, QObject, QEvent
-from qtpy.QtGui import QFont, QColor, QIcon, QTextCursor, QPainter, QTextOption, QPainterPath, QPen, QBrush, QFontMetrics
+from qtpy.QtCore import (
+    Qt,
+    QTimer,
+    QSize,
+    Signal,
+    Slot,
+    QPointF,
+    QRectF,
+    QThread,
+    QObject,
+    QEvent,
+)
+from qtpy.QtGui import (
+    QFont,
+    QColor,
+    QIcon,
+    QTextCursor,
+    QPainter,
+    QTextOption,
+    QPainterPath,
+    QPen,
+    QBrush,
+    QFontMetrics,
+)
 
 # Cross-backend window flag compatibility (PyQt5/PySide2 vs PyQt6/PySide6)
-_WIN_STAYS_ON_TOP = getattr(getattr(Qt, "WindowType", None), "WindowStaysOnTop", None) or getattr(Qt, "WindowStaysOnTop", 0x00080000)
+_WIN_STAYS_ON_TOP = getattr(
+    getattr(Qt, "WindowType", None), "WindowStaysOnTop", None
+) or getattr(Qt, "WindowStaysOnTop", 0x00080000)
 
 # ---------------------------------------------------------------------------
 # HAL API — direct shared memory access via _hal C extension
@@ -69,21 +122,20 @@ class HalApi:
     Supports pins, signals, params, components, functions, and threads via SHM.
     Falls back to subprocess halcmd only on older _hal versions that lack get_info_*."""
 
-
     # Mapping between _hal numeric type constants and human-readable names
     TYPE_NAME = {
-        0: "bit",     # HAL_BIT
-        1: "float",   # HAL_FLOAT
-        2: "s32",     # HAL_S32
-        3: "u32",     # HAL_U32
-        4: "s64",     # HAL_S64
-        5: "u64",     # HAL_U64
-        6: "port",    # HAL_PORT
+        0: "bit",  # HAL_BIT
+        1: "float",  # HAL_FLOAT
+        2: "s32",  # HAL_S32
+        3: "u32",  # HAL_U32
+        4: "s64",  # HAL_S64
+        5: "u64",  # HAL_U64
+        6: "port",  # HAL_PORT
     }
 
     # Mapping for direction constants
-    PIN_DIR = {16: "IN", 32: "OUT", 48: "I/O"}   # HAL_IN=16, HAL_OUT=32, HAL_IO=48
-    PARAM_DIR = {64: "RO", 192: "RW"}          # HAL_RO=64, HAL_RW=192
+    PIN_DIR = {16: "IN", 32: "OUT", 48: "I/O"}  # HAL_IN=16, HAL_OUT=32, HAL_IO=48
+    PARAM_DIR = {64: "RO", 192: "RW"}  # HAL_RO=64, HAL_RW=192
 
     _initialized = False
     _comp = None
@@ -122,10 +174,16 @@ class HalApi:
             try:
                 cls._comp.ready()
             except Exception as e:
-                print(f"[halshow] cleanup: ready() failed ({e}) — proceeding to exit anyway", file=sys.stderr)
+                print(
+                    f"[halshow] cleanup: ready() failed ({e}) — proceeding to exit anyway",
+                    file=sys.stderr,
+                )
             cls._comp.exit()
         except Exception as e:
-            print(f"[halshow] ERROR: cannot unload component {comp_name}: {e}", file=sys.stderr)
+            print(
+                f"[halshow] ERROR: cannot unload component {comp_name}: {e}",
+                file=sys.stderr,
+            )
         finally:
             # Set _initialized to False so a new init() can run if halshow is
             # restarted within the same Python process. Clearing _comp lets GC
@@ -164,7 +222,11 @@ class HalApi:
             # Fallback to halcmd subprocess
             try:
                 lines = cls._halcmd_list("pin").splitlines()
-                indexed = {line.strip(): {"NAME": line.strip()} for line in lines if line.strip()}
+                indexed = {
+                    line.strip(): {"NAME": line.strip()}
+                    for line in lines
+                    if line.strip()
+                }
                 cls._cache["pins"] = indexed
             except Exception as e:
                 print(f"[halshow] Error caching pins: {e}", file=sys.stderr)
@@ -184,7 +246,11 @@ class HalApi:
             # Fallback to halcmd subprocess
             try:
                 lines = cls._halcmd_list("param").splitlines()
-                indexed = {line.strip(): {"NAME": line.strip()} for line in lines if line.strip()}
+                indexed = {
+                    line.strip(): {"NAME": line.strip()}
+                    for line in lines
+                    if line.strip()
+                }
                 cls._cache["params"] = indexed
             except Exception as e:
                 print(f"[halshow] Error caching params: {e}", file=sys.stderr)
@@ -204,7 +270,11 @@ class HalApi:
             # Fallback to halcmd subprocess
             try:
                 lines = cls._halcmd_list("sig").splitlines()
-                indexed = {line.strip(): {"NAME": line.strip()} for line in lines if line.strip()}
+                indexed = {
+                    line.strip(): {"NAME": line.strip()}
+                    for line in lines
+                    if line.strip()
+                }
                 cls._cache["signals"] = indexed
             except Exception as e:
                 print(f"[halshow] Error caching signals: {e}", file=sys.stderr)
@@ -240,8 +310,12 @@ class HalApi:
     def _halcmd_list(cls, halcmd_type):
         """List HAL entities via halcmd subprocess."""
         try:
-            result = subprocess.run([cls._find_halcmd(), "list", halcmd_type],
-                                    capture_output=True, text=True, timeout=5)
+            result = subprocess.run(
+                [cls._find_halcmd(), "list", halcmd_type],
+                capture_output=True,
+                text=True,
+                timeout=5,
+            )
             if result.returncode != 0:
                 return ""
             names = result.stdout.strip().split()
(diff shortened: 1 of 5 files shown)


grandixximo:
You might want to enable the executable bit in git for halshow.py
I did quick testing, not looked at the graph yet, should pygraphviz become a dependency?
The grouped view of pins and parameters is the new default, but seems not working in the port?

petterreinholdtsen:
[Luca Toniolo]
> You might want to enable the executable bit in git for halshow.py

Yes, and integrate it into the build system and install rules.

> I did quick testing, not looked at the graph yet, should pygraphviz
> become a dependency?

Yes, it is a dependency for the graph drawing.

> The grouped view of pins and parameters is the new default, but seems
> not working in the port?

What is this?  I do not believe I have used or seen this myself.
Something from after 2.9?

-- 
Happy hacking
Petter Reinholdtsen

grandixximo:
> What is this?  I do not believe I have used or seen this myself.
Something from after 2.9?

Yes, just made it in 2.10 a few weeks back, see #4147

petterreinholdtsen:
[Luca Toniolo]
> Yes, just made it in 2.10 a few weeks back, see #4147

Right.  Then I understand.  It had an option in the settings tab, set to
true by default.  There was a bug in the settings tab, now fixed, making
it impossible to flip the switch.  Added the option as I was surprised
the tree did not look the way I was used to, but could drop it if the
new wanted default is the combined pin/parameter tree.

-- 
Happy hacking
Petter Reinholdtsen

scripts/halshow.py
```diff
@@ -0,0 +1,3672 @@
...
+        exec_btn.clicked.connect(lambda: self._execute_cmd(self.cmd_entry.text()))
+        cmd_layout.addWidget(exec_btn)
+
+        main_layout.addWidget(cmd_frame)
+
+        # Status bar text (below splitter, full width — matches Tcl layout)
+        self.status_text = QTextBrowser()
+        self.status_text.setMaximumHeight(24)
+        self.status_text.setStyleSheet("border-width: 1px; border-style: solid;")
+        self.status_text.setHtml(f"<i>{_("Commands may be tested here but they will NOT be saved")}</i>")
```
hansu:
```suggestion
        self.status_text.setHtml(f'<i>{_("Commands may be tested here but they will NOT be saved")}</i>')
```

I get a lot of errors like` SyntaxError: f-string: unmatched '('`,  on line 3108, 3162, 3347 ...

petterreinholdtsen: Good idea.  I applied this and several other similar cleanups, as well as ran black to fix a bunch more.

hansu: Two more need fixing ;-)

```bash

  File "/home/cnc/linuxcnc-master/scripts/halshow.py", line 3787
    self.setWindowTitle(f"{fname} - {_("Halshow")}")
                                        ^^^^^^^
SyntaxError: f-string: unmatched '('

  File "/home/cnc/linuxcnc-master/scripts/halshow.py", line 3817
    self.setWindowTitle(f"{fname} - {_("Halshow")}")
                                        ^^^^^^^
SyntaxError: f-string: unmatched '('
```

hansu:
Nice! It needs some fine tuning, but most of the features of the current halshow I can find :+1: 

Some notes:

- In the SHOW tab, it should list the pins,... when clicking on the node, e.g. clicking ob "Pins" should list all pins, clicking "axis" should list all pins axis.*
- Clicking on a slection in the SHOW tab crashes the application with   
  ```
  File "/home/cnc/linuxcnc-master/scripts/halshow.py", line 3390, in _show_context_menu
      selected_text = self.show_browser.selectedText().strip()
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  AttributeError: 'QTextBrowser' object has no attribute 'selectedText'
  ```

- I think it looks better if the bubbles for bool values are right aligned
- The expand/collapse button for the tree view is missing
- Very good idea to integrate the graphviz graph. But it's very hard to see the small boxes. Maybe you can get some inspiration from: https://github.com/multigcs/halviewer (or just integrate that one)

petterreinholdtsen:
[Hans Unzner]
> Two more need fixing ;-)

Fixing.  How do you trigger these errors?  My tests using Python 3.13.5
do not show these problems.

-- 
Happy hacking
Petter Reinholdtsen

Sigma1912:
There seems to be an issue with some of the value conversions (eg joint.N.pos-cmd):

hal_show_py

hansu:
> [Hans Unzner]
> Two more need fixing ;-)
> Fixing. How do you trigger these errors? My tests using Python 3.13.5 do not show these problems.

Python 3.11.2 on Bookworm

>Clicking on a slection in the SHOW tab crashes the application with

Sorry I was talking about right-click:


https://github.com/user-attachments/assets/cf4972ce-0fc5-4066-a2cb-371be0ac48ba

hansu:
>There seems to be an issue with some of the value conversions (eg joint.N.pos-cmd):

What value did you expect (does halshow show)?

Sigma1912:
> What value did you expect (does halshow show)?

Well, it shows a float, like 5.325

hansu:
> > What value did you expect (does halshow show)?
> 
> Well, it shows a float, like 5.325

I get a float display. Probably the value is too large to fit into the column width, so the end is cut.
And I see the setting for the values' column width is gone. @petterreinholdtsen please add.

Sigma1912:
> Probably the value is too large to fit into the column width, so the end is cut.

Yes, seems to be missing formatting.


As hansu already mentioned, the graph takes a lot of panning and zooming so having the mouse scroll wheel not supported for zooming is real pain.  To make the graph more compact  I would drop the extra nodes for the signal names and show those when selecting the line. 

IMO the biggest problem with  these hal graphs is that they very quickly become too large and unwieldy.

Sigma1912:
Just as an example, this is a config from my test setup with a relatively small hal file (my actual machines have hal setups with >600 lines). This config has spindle orientation and this is where a graphical representation would be really useful. As it is the 'Graph' tab just gives me everything and I have to zoo, pan and scroll to find the relevant components and then drag along trying to follow signals back and forth:

Screenshot from 2026-07-15 11-46-32


The problem is that I'm given too much information to actually be helpful. What would be much more useful is a graph of the components I select from the tree. After all the 'Show' and 'Watch' tabs don't just list every possible piece of information there is either and it would be very bad if they did.

grandixximo:
A decent solution I found on a similar tool I experimented with, was filtering by the relevant pin to visualize, plus a selection for connections branching, that should filter everything connected to what you are wanting to see without including everything that's not relevant.

grandixximo:
image
the brownish color on the list, does not match for parameters, like current 2.10 halshow when pins & Params are visualized together
image

petterreinholdtsen:
I've updated the python edition slightly:

    halshow.py: major improvements to SHOW, WATCH, and GRAPH tabs
    
    SHOW tab:
    - Compact one-line-per-item tables matching halcmd visual format
    - One table per section (pins/params/signals) with aligned columns
    - Clicking non-leaf tree nodes expands and shows all underlying items
    - Signal connection info works with old _hal.so lacking SIGNAL SHM field
      via halcmd subprocess fallback (batched single call for efficiency)
    - Fixed reversed signal arrow directions: OUT pins show ==>, IN pins <==
    - Context menu always shows all items (Tcl-style), tooltip at click when
      nothing selected, right-click no longer crashes
    
    WATCH tab:
    - Column headers (Value / Name) with draggable separator to resize columns
    - Value column default width 150px, values left-aligned
    - Boolean bit indicator circles left-aligned within value area
    - Param leaves render in brown color (Tcl match), sig labels in blue
    - Context menu on empty space: add from clipboard/HAL text, erase watchlist
    
    GRAPH tab:
    - Mouse wheel zoom centered on cursor position
    - Keyboard +/- keys for zoom in/out
    
    General:
    - Preferences path displayed with ~ notation instead of full home directory
    
    This patch was created with help from OpenCode using local llama.cpp
    server with Qwen 3.6.

[Hans Unzner]
> - The expand/collapse button for the tree view is missing

I do not understand what you mean?  I only see these options in the
menu in the tcl edition, and the same is the case with the python
edition.

> - Very good idea to integrate the graphviz graph. But it's very hard to see
> the small boxes. Maybe you can get some inspiration from:
> https://github.com/multigcs/halviewer (or just integrate that one)

I did not try to improve the graph yet, and will have a look at this
when I have time to look at the graph layout again.  I agree that the
graph of real machine HAL setup is confusing and hard to figure out.  It
is true also for my own machine, and thus a motivation to come up with a
better graph approach.  When that is said, I have already found the
current graph useful when looking at the setup of my own machine,
despite being slightly painful to find the relevant part by panning and
zooming in and out to try to find the relevant part of the graph. :)

[Sigma1912]
> To make the graph more compact I would drop the extra nodes for the
> signal names and show those when selecting the line.

I decided against this as signals must be represented as one to many
connections, and a single line can not represent this well.

-- 
Happy hacking
Petter Reinholdtsen

rene-dev:
@petterreinholdtsen can those changes be added to https://github.com/LinuxCNC/linuxcnc/pull/4251

petterreinholdtsen:
[Rene Hopf]
> @petterreinholdtsen can those changes be added to
> https://github.com/LinuxCNC/linuxcnc/pull/4251

I assume so.  The changes is quite simple to make more information
available quicker, no real algoritm involved.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
#### 1. 
The handle to adjust the column width
grafik

 is nice, but crashes on my machine with

```
  File "/home/cnc/linuxcnc-master/src/../scripts/halshow.py", line 879, in mousePressEvent
    self._press_x = event.globalPosition().x()
                    ^^^^^^^^^^^^^^^^^^^^
```

#### 2. 
The widgets in the settings tab should be top aligned, not expanded.

hansu:
>> [Hans Unzner]
>>- The expand/collapse button for the tree view is missing
>>
> I do not understand what you mean? I only see these options in the menu in the tcl edition, and the same is the case with the python edition.

This handy little function:

halshow_expand_collapse

petterreinholdtsen:
Thank you for testing.

[Hans Unzner]
> #### 1.
> The handle to adjust the column width is nice, but crashes on my
> machine with

I am unable to reproduce this.  The pull marker work with my left mouse
button and fail to crash all my three mouse buttons.  Any idea what is
needed to reproduce the crash?

> #### 2. 
> The widgets in the settings tab should be top aligned, not expanded.

I agree.

> This handy little function:

Ah, got it.  Had not even noticed that little button.  I'll see what I
can do.  One can get the same effect with the current interface by
dragging the tree view square to the left until it collapses, but I
agree that a quick button click would be useful.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
>> #### 1. The handle to adjust the column width is nice, but crashes on my machine with
> I am unable to reproduce this. The pull marker work with my left mouse button and fail to crash all my three mouse buttons. Any idea what is needed to reproduce the crash?

Seems to be caused by a change in the API. Here a possible fix:

```diff
diff --git a/scripts/halshow.py b/scripts/halshow.py
index 009559524b..b0ffb88229 100755
--- a/scripts/halshow.py
+++ b/scripts/halshow.py
@@ -876,17 +876,26 @@ class _ColumnSeparator(QWidget):
     def mousePressEvent(self, event):
         if event.button() == Qt.LeftButton:
             self._dragging = True
-            self._press_x = event.globalPosition().x()
+            self._press_x = (
+                event.globalPosition().x()
+                if hasattr(event, "globalPosition")
+                else event.globalPos().x()
+            )
             event.accept()
 
     def mouseMoveEvent(self, event):
         if self._dragging:
-            delta = event.globalPosition().x() - self._press_x
+            current_x = (
+                event.globalPosition().x()
+                if hasattr(event, "globalPosition")
+                else event.globalPos().x()
+            )
+            delta = current_x - self._press_x
             header = self.parent()
             if hasattr(header, "_value_width"):
                 new_width = max(40, min(header._value_width + int(delta), 500))
                 self.width_changed.emit(new_width)
-                self._press_x = event.globalPosition().x()
+                self._press_x = current_x
 
     def mouseReleaseEvent(self, event):
         if event.button() == Qt.LeftButton:

```

petterreinholdtsen:
Thank you for the patch and the feedback.  The crash bug should be
fixed, the collapse button in place and a new and improved graph drawing
implementation based on halviewer.py replaced the old one.

Is this also working for the rest of you?

-- 
Happy hacking
Petter Reinholdten

scripts/halshow.py
```diff
@@ -0,0 +1,4129 @@
...
+
+# ---------------------------------------------------------------------------
+# HAL API — direct shared memory access via _hal C extension
+# ---------------------------------------------------------------------------
+
+try:
+    import _hal
+except ImportError:
+    _hal = None
+
```
BsAtHome, line 121: You must not import `_hal`. This is the private part of the HAL interface. The public API of the HAL module is available as `hal`.

petterreinholdtsen:
[BsAtHome]
> You must not import `_hal`. This is the private part of the HAL
> interface. The public API of the HAL module is available as `hal`.

I am not quite sure what you base your 'must not' on.  This tool is
intended to be part of linuxcnc and replace the tcl/tk implementation,
and as such will be relased in lockstep with the rest of HAL, so no need
to avoid "private parts" of the HAL interface.

Besides, The current public API of the HAL module is not complete enough
to provide what halshow.py needs, like get_info_pins and
get_info_signals().

-- 
Happy hacking
Petter Reinholdtsen

grandixximo:
The CI failure comes from the merge with master: #4317 removed all `hal_priv.h` access from halmodule.cc, so the added `get_info_components`/`get_info_functions`/`get_info_threads` no longer compile (`hal_comp_t`, `SHMPTR`, `scoped_hal_mutex` are gone).

On the `_hal` import: `hal.py` does `from _hal import *`, so `get_info_pins`/`get_info_signals`/`get_info_params` are already available through the public `hal` module. No private import needed for those.

Proposal that solves both: reimplement the three new `get_info_*` functions on the public query API. `hal_list_comp`, `hal_list_funct`, and `hal_list_thread` (hal.h) use the same callback pattern as the rewritten `get_info_pins`, so the port is mechanical. That fixes the build on master, drops the `hal_priv.h` dependency, and completes the Python HAL API surface for comps/functs/threads properly.

BsAtHome:
> You must not import `_hal`. This is the private part of the HAL interface. The public API of the HAL module is available as `hal`. I am not quite sure what you base your 'must not' on.

That I base on the fact that I have spent the last few months untangling HAL from the rest of the code base.

Including the lower/implementation layers of code from top layer API users is what is called a layer violation and it makes maintenance extremely hard. Systems with layer violations require that every time you need to alter a private and normally hidden implementation detail, then you also have to go through all the API users to see what they have been doing. That should not be necessary and therefore you must not include the implementation part and have to stick to the public API.

Also, see note at top: https://linuxcnc.org/docs/devel/html/en/config/python-hal-interface.html

hansu:
Have you tried scrolling in the graph view?

It crashes for me:
```
  File "/home/cnc/linuxcnc-master/src/../scripts/halshow.py", line 2032, in eventFilter
    self._wheel_zoom(event)
  File "/home/cnc/linuxcnc-master/src/../scripts/halshow.py", line 2040, in _wheel_zoom
    mouse_pos = self.view.mapFromGlobal(
                ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: mapFromGlobal(self, a0: QPoint): argument 1 has unexpected type 'QPointF'
Aborted
```

Further I wonder about this message I get when switching to the graph view:

```
how] OWNER field missing (old _hal.so), deriving from pin names, known_comps=['__base-thread', '__servo-thread', '_halshow_122969', '_halshow_123162', 'abs', 'comp', 'ddt', 'encoder', 'gmoccapy', 'halcmd123484']...
```


Other open points (which I also can fix later):
- Alignment of settings
- The highlighting of the filtered items isn't implemented yet
- The window size isn't changes on cpllapse/expand

BsAtHome:
The new query API has now been fully integrated and there is no longer any need to modify halmodule.cc. You can now rebase your code to master's head and use the full HAL query API for all information about pins, params, signals, components, functions and threads.

In a short while, the documentation at https://linuxcnc.org/docs/devel/html/en/config/python-hal-interface.html will be automatically updated to describe the new query API. In the meantime you can also find the current documentation at https://github.com/LinuxCNC/linuxcnc/blob/master/docs/src/config/python-hal-interface.adoc#sub-module-hal-query
The listing methods are `hal.query.XX()` where XX is one of `pins`, `params`, `signals`, `comps`, `functs` or `threads` (note the trailing 's').

petterreinholdtsen:
[Hans Unzner]
> Have you tried scrolling in the graph view?

I have now when I tracked down a scroll mouse to tested.  It crashed for
me too, and this is now fixed.

> Further I wonder about this message I get when switching to the graph view:
> 
> ```
> how] OWNER field missing (old _hal.so), deriving from pin names,
> known_comps=['__base-thread', '__servo-thread', '_halshow_122969',
> '_halshow_123162', 'abs', 'comp', 'ddt', 'encoder', 'gmoccapy',
> 'halcmd123484']...
> ```

This is leftover debug output that should be removed in the final
edition.

> Other open points (which I also can fix later):
> - Alignment of settings

This has been fixed.

> - The highlighting of the filtered items isn't implemented yet

I believe this is not needed with the dyamically updating tree view when
typing in search terms, as the only entries left in the view are those
matching the search term.

> - The window size isn't changes on cpllapse/expand

I suspect this is fixed, or I do not understand what you mean.

[BsAtHome]
> The new query API has now been fully integrated and there is no longer
> any need to modify halmodule.cc. You can now rebase your code to
> master's head and use the full HAL query API for all information about
> pins, params, signals, components, functions and threads.

OK.  Rebase done on top of master and implementation adjusted to use the
new API when available.

-- 
Happy hacking
Petter Reinholdtsen

hansu:
> I have now when I tracked down a scroll mouse to tested. It crashed for me too, and this is now fixed.

Works now, thanks

Settings also look good now



> - The highlighting of the filtered items isn't implemented yet
> I believe this is not needed with the dyamically updating tree view when typing in search terms, as the only entries left in the view are those matching the search term.


I think this is still helpful to see what actually matched:

grafik

> - The window size isn't changes on collapse/expand
> I suspect this is fixed, or I do not understand what you mean.

The goal of this function is to display halshow after adding the items beside the GUI on screens with medium resolution:
halshow-expand-collapse

Further when closing halshow with hidden tree, the expand button is hidden at next start.



1. I would not translate the button names, otherwise it doesn't fit the reserved space:
     grafik


1. The watchlist is not saved automatically even when enabled in settings.

BsAtHome:
Have a look at the attached version... I removed nearly all of the halcmd proxy calling and changed all normal operation to `hal.query.*` calls. Only the graph stuff still has a halcmd proxy, which you may fix. The show pane now also works again for components, functions and threads.

[halshow.py](https://github.com/user-attachments/files/31599272/halshow.py)

Note 1: It will require master to run. Older versions do not support the `hal.query` API.
Note 2: The FP column (thread/function uses floating point) has been removed because that is always true in master and no longer an option.
Note 3: None of the UI stuff has been addressed. This version only fixes the HAL API access (except graph stuff)

PR #4237, opened 2026-07-12
linuxcncrsh: Properly return std::nullopt and not a nullptr

BsAtHome:
Null return value for `std::optional` is `std::nullopt` and not `nullptr`. This fixes a potential crash.


andypugh:
Is this bugfixy enough to belong in 2.9?

BsAtHome:
No, it is not in 2.9. This bug was introduced when I rewrote linuxcncrsh, which is in master.

PR #4240, opened 2026-07-13
Xenomai rootless and fixes

hdiethelm:
This PR has 3 commits:
- Xenomai EVL: Abort on evl_attach_self() failed: Otherwhise, there where 1000's of error messages/s if something went wrong
- Xenomai EVL: Remove uneeded debug print
- Rootless for Xenomai and Xenomai EVL: As proposed in https://github.com/LinuxCNC/linuxcnc/pull/3964

This needs libevl >= 59-1 from https://github.com/hdiethelm/xenomai4-linuxcnc where the udev rules are installed. Of course, this can also be done manually. The rules are here: https://github.com/hdiethelm/xenomai4-linuxcnc/blob/lfs/libevl-debian/udev/90-libevl.rules

Tested and works both with Xenomai3 / Xenomai4 EVL.

### DIFF
--- a/src/rtapi/uspace_rtapi_main.cc
+++ b/src/rtapi/uspace_rtapi_main.cc
@@ -38,6 +38,7 @@
 #include <time.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <grp.h>
 #ifdef HAVE_SYS_IO_H
 #include <sys/io.h>
 #endif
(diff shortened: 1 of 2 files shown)


src/rtapi/uspace_rtapi_main.cc
```diff
@@ -96,18 +97,44 @@ static inline int detect_preempt_dynamic() {
...
-// with udev rules + a 'xenomai'/'evl' group; @hdiethelm has a follow-up
-// planned.  Until then, an unprivileged user on a Xenomai kernel cannot
-// claim the Xenomai backend, and falls back to the SCHED_FIFO probe.
 [[maybe_unused]] static inline int has_setuid_root() {
     return geteuid() == 0;
 }
 
+[[maybe_unused]] int is_current_user_in_gid(gid_t target_gid) {
+    int ngroups = getgroups(0, NULL);
+    if (ngroups < 0) {
```
BsAtHome, line 109: There is no use for `[[maybe_used]]` because this is a real non-static function. Therefore, it is used or it isn't and the compiler cannot determine which is which because it is not static.
(note: marker previously used in a static inline function, which behaves quite differently)

My guess is that this function is only used when USPACE_XENOMAI or USPACE_XENOMAI_EVL are defined. If so, then you should embed the function in the appropriate `#if defined(USPACE_XENOMAI) || defined(USPACE_XENOMAI_EVL)` preprocessor conditional. Also, this function should probably be declared `static`.

hdiethelm: Thanks. Of course I forgot the static.

Changed to `if defined()` and removed the inline for has_setuid_root(). There is no need for inlining this function.

src/rtapi/uspace_rtapi_main.cc
```diff
@@ -96,18 +97,48 @@ static inline int detect_preempt_dynamic() {
...
-[[maybe_unused]] static inline int has_setuid_root() {
+#if defined(USPACE_RTAI) || defined(USPACE_XENOMAI) || defined(USPACE_XENOMAI_EVL)
+static int has_setuid_root() {
     return geteuid() == 0;
 }
+#endif
+
+#if defined(USPACE_XENOMAI) || defined(USPACE_XENOMAI_EVL)
+static int is_current_user_in_gid(gid_t target_gid) {
+    int ngroups = getgroups(0, NULL);
```
BsAtHome: Why is the return type not bool?

hdiethelm: is_current_user_in_gid() has -1 on failure, 0 on not member, 1 on member.

It's a bit a mess in rtapi_app_main.cc with the return values. Sometimes 1 is pass, sometimes 0 and sometimes it is bool.

The detect_* functions return 1 on success. For example handle_command() and all commands have -1 on failure, 0 on success.

It would probably make sense to change all detect_* and related functions to bool?

BsAtHome: But you are using it like:
```C
bool isInGid = is_current_user_in_gid(gid) > 0;
```

If this is supposed to distinguish between "user in gid", "user not in gid" or "error condition", then you should actually use the return value to distinguish the cases. However, you add this function and convert its return value to bool. You are not doing anything to distinguish error from not and treat errors like "user not in gid". That is fine by me... But then, my question stands: why is the return value not bool?

hdiethelm: https://github.com/LinuxCNC/linuxcnc/pull/4240/commits/f5316b6c293d416b892e707f3a6b8af479fe150b changes all return values to bool in the detect_* code. I think this makes more sense.

is_current_user_in_gid() should never fail so the perror() handling should be enough. According to the man page, a wrong address or wrong size is the only reason and this is verified before. So out of memory is the only remaining reason.

hdiethelm:
1. force push: rebase master
2. force push: check for gid == 0 and improve warning

hdiethelm:
@BsAtHome Anything blocking a merge or is just everybody busy? ;-)

BsAtHome:
Just busy on the getter/setter.

hdiethelm:
> Just busy on the getter/setter.

Thanks for merging. Yes, I've seen your 20k lines PR, quite a project.

PR #4242, opened 2026-07-15
shellcheck: Fix remaining warnings in scripts.

BsAtHome:
This cleans up the last of the shellcheck warnings seen in CI. The clean source tree is does no longer contain any messages in the scripts.

However, there is one left _after_ running configure in the expanded linuxcnc script. This is in the quoted value of the `LINUXCNC_CONFIG_PATH` that contains a path and the first entry starts with a tilde. However, quoted strings are not tilde expanded and shellcheck complains.

The expansion comes from configure.ac and is probably meant to be expanded to the user running LinuxCNC. But it seems to be broken the way it is done. After searching, the LINUXCNC_CONFIG_PATH environment variable is apparently not used, anywhere. So maybe it should be removed.

### DIFF
--- a/docs/src/asciideps
+++ b/docs/src/asciideps
@@ -1,14 +1,16 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
 test -z "$1" && exit 0
 test -f "$1" || exit 1
 
+includestack=( )
+
 includes () {
 	DIR=$(dirname "$1")
 
-	for f in $(sed -ne "s|^include::\(.*\)\[\]$|$DIR/\1|p" "$1"); do
+	while IFS= read -r f; do
 		# components_gen.adoc will contain only generated content
 		case "$f" in
 			*/components_gen.adoc)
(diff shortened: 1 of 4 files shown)


docs/src/asciideps
```diff
@@ -1,23 +1,32 @@
...
+	while IFS= read -r f; do
 		# components_gen.adoc will contain only generated content
 		case "$f" in
 			*/components_gen.adoc)
 			touch "$f"
 			;;
 		esac
 		echo "$f"
+		ff=$(realpath "$f")
+		if [[ " ${includestack[*]} " =~ [[:space:]]${ff}[[:space:]] ]]; then
```
grandixximo, line 22: quoted `"${ff}"` would be safer here, not strictly necessary, just being picky ;-)

grandixximo (approved):

docs/src/asciideps
BsAtHome: The bash manual says: "If any part of the pattern is quoted, the quoted portion is matched literally, as above. If the pattern is stored in a shell variable, quoting the variable expansion forces the entire pattern to be matched literally. Treat bracket expressions in regular expressions carefully, since normal quoting and pattern characters lose their meanings between brackets."

The pattern expression is enclosed in two [[:space:]] markers. It should be guarded correctly with those markers. The expression is not word-split after expansion afaik. And, lets just say that if the filepath contains spaces or quotes, then we're royally screwed anyway. ;-)

grandixximo: It was more about that unquoted `$ff` being a regex, so `.` in a path matches any char.

grandixximo: It's really picky because it would be impossible to have

`.../foo.adoc` and `.../fooXadoc`

same length only `.` and whatever `X` would be, it's an effectively impossible occurrence.

Issue #4144, opened 2026-06-09
Hostmot2 sserial state machine reassigns state unconditionally


BsAtHome:
In the `hm2_sserial_update_params()` function in `hal/drivers/mesa-hostmot/sserial.c` there is a state machine running at several levels (using pins `state2` and `state3`).

There is a curious line in there https://github.com/LinuxCNC/linuxcnc/blob/e90c66dac743dc0bf9f827f70fab5cdd8898c662/src/hal/drivers/mesa-hostmot2/sserial.c#L1569-L1576

To highlight the part in question:
```C
case 1:
    ret = hm2_sserial_wait(hm2, inst, period);
    if (ret > 0) break;

   // **** This part here
    if (ret < 0) *inst->state3 = 100; // quit and tidy up
    *inst->state3 = 2;
    //  **** Until here

    inst->timer = 20000000;
    *inst->command_reg_write = 0xF00 | (1 << r->index); // channel in setup mode
    break;
```

The `state3` variable is assigned 100 if the `hm2_sserial_wait()` fails. But `state3` is assigned the value 2 <i>unconditionally</i> in the next line. It looks like there is a missing break and the actual code should have been like this:
```C
    if (ret < 0) {
        *inst->state3 = 100; // quit and tidy up
        break;
    }
    *inst->state3 = 2;
```
There is a `case 100` with the comment "error recovery", so I guess that the "quit and  tidy up" comment refers to error recovery.

The same mistake is seen several places in the switch of the state machine and most likely all need the same fix.

grandixximo:
Good catch. I dug into the history and behavior a bit.

The pattern comes from d92f694380 (2017, "Smart Serial: Start to enable parameter setting via HAL"). All 6 occurrences (cases 1, 2, 3, 5, 7, 9) were introduced in that same commit, so `case 100` has been unreachable dead code since day one; nothing else ever assigns 100.

The missing `break` reading matches the intent: the "quit and tidy up" comment clearly refers to the `case 100` "error recovery" handler.

Worth noting what the proposed fix changes. Today an error from `hm2_sserial_wait()` is logged but the state machine marches on as if it succeeded; in case 5 it even sets `written = param` after a failed write, so the parameter change is silently lost but also never retried. With just the missing `break` added, `case 100` is reached but does not set `written = param`, so a persistently failing write (dead remote, bad cable) would be retried on every pass through the parameter loop: infinite retries plus an HM2_ERR each time.

[...]

PR #4245, opened 2026-07-16
hostmot2: Fix sserial error test to use 1 << x and not 1 < x

BsAtHome:
This is a partial fix for #4144, where it was discovered that the error check of a bit field used `1 < x` instead of `1 << x`.

The other issue will be cleaned up separately. See #4144 for details.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/sserial.c
+++ b/src/hal/drivers/mesa-hostmot2/sserial.c
@@ -75,7 +75,7 @@ int hm2_sserial_wait(hostmot2_t *hm2, hm2_sserial_instance_t *inst, long period)
                 "Timeout waiting for CMD to clear\n");
         return -1;
     }
-    if (*(inst->data_reg_read) & (1 < inst->remotes[inst->r_index].index)){
+    if (*(inst->data_reg_read) & (1 << inst->remotes[inst->r_index].index)){
         HM2_ERR("Error after doit clear\n");
         return -1;
     }


grandixximo:
Should this be tested on hardware? @hdiethelm have anything on hand?

hdiethelm:
> Should this be tested on hardware? @hdiethelm have anything on hand?

It would make sense. However, I don't have an sserial card.

BsAtHome:
Sure, check. But,... the error in the code is very clearly an error that would result in:
```
 index=0 ==> pinval & (1 < inst->remotes[inst->r_index].index) = pinval & (1 < 0) = pinval & 0
 index=1 ==> pinval & (1 < inst->remotes[inst->r_index].index) = pinval & (1 < 1) = pinval & 0
 index=2 ==> pinval & (1 < inst->remotes[inst->r_index].index) = pinval & (1 < 2) = pinval & 1
 index=3 ==> pinval & (1 < inst->remotes[inst->r_index].index) = pinval & (1 < 3) = pinval & 1
...
```

Therefore, you'd either not check at all (x & 0 resolves to false) and all the others would only test remote 0. And that remote may not be applicable or invalid for the instance the test was to check.

If there are installs where you get an error _after_ this patch, then at least you know something is wrong. Currently, that chance of having a correct error is just about zero.

grandixximo:
@pcw-mesa could you give a thumbs up?

hdiethelm:
@BsAtHome Agreed that this looks like an error and the fix looks good. However, fixing it could surface an other bug...

Now testig is not that simple, you would have to make sure this function is called and ideally also trigger an error.

grandixximo:
Yeah, it's more about making sure nothing else that's been hidden would surface once it runs on real hardware, it is indeed an error that never showed, the fix is correct, it's just about making sure nothing else was lurking in the dark waiting to bite :-)

BsAtHome:
Actually, it does not change the behaviour of the code path other than emitting an error message.

The case when this hits and returns -1 is the case that initiated the #4144 report. A negative return will set the state to 100 in the statemachine, which immediately thereafter is overwritten to set another state.

So, if the statemachine is changed no longer to skip state 100, _then_ we need a real world test that triggers the path. And this PR does not change the statemachine handling, so this just actually prints the intended message that something went wrong. And that is not a bad thing.

When the rest of #4144 is addressed, we need to assure functionality if state 100 is ever to be executed. However, the idea there was to leave it as is (and drop the error recovery state 100 completely).

pcw-mesa:
Looks OK especially if just in master for the time being.

I think there are several weaknesses in sserial error handling but fixing obvious mistakes
is a good start.

Issue #4244, opened 2026-07-16
Feature request: Have HTML manpages link


BsAtHome:
There are many manpages and they reference each other in "SEE ALSO" and other places using the `name(section)` format.

For example, `linuxcnc(1)` references `halcmd(1)`. It would be nice if we just could click on a link in the HTML version. Very helpful in all man sections when digging through man pages. This could maybe/probably be done in the an asciidoctor filter (?) while generating the HTML?

It would probably also be interesting for the normal docs. But I'm not sure if that should be done using "normal" linking or post-processing. It is harder to do in normal docs than man pages, as man pages have a stricter format.

hansu:
Do you mean with normal linking just adding links to the items in the "see also" section which points then to the corresponding man web page?

BsAtHome:
Not only in the "see also" section. All occurrences of name(section) should link when they refer to an in-tree manual page. We have really many of them.

Don't know If you ever used pinfo (cmdline man/info reader). Pinfo will scan the manual page and make all name(section) references links you can browse between (scroll with arrow keys up/down) and go to using right arrow or back with left arrow. Not that it is the best program, but the navigation with links in manual pages is superb.

[note: example links do not point to a valid target, just an example]

Something like <a href="man/man1/blabla.1.html">blabla</a>(1) will link to program blabla in section 1 and we invoke <a href="man/man1/foobar.1.html">foobar</a>(1) to add suspense.

PR #4246, opened 2026-07-17
docs: link manpage cross-references in HTML output

grandixximo:
Addresses #4244.

## What

Adds an asciidoctor postprocessor (`docs/src/extensions/man_xref.rb`) that turns `name(section)` manpage references (`halcmd(1)`, `stepgen(9)`, ...) into clickable links to the corresponding HTML page. Applied to both:

- the generated **HTML manpages** (sibling links under `man/`), and
- the **narrative manuals** (User/HAL/Integrator), which cite man pages in the same form and now link across into `../man/manN/`.

Wired into both HTML rules in `docs/src/Submakefile`. Per-page `manxref-root` (troff tree) and `manxref-linkbase` (depth-adjusted path) attributes drive it; absent => no-op.

## How it stays correct

Index-gated: a token links only when a real page `name.section` exists in that language's troff tree. Common false positives resolve to nothing and stay plain, with no skip list:

- enum values, e.g. `feed(2)`/`arc(3)` in `motion(9)`
- external pages, e.g. `open(2)`, `printf(3)`, `udev(8)`
- value literals in prose, e.g. `TRUE(1)`, `time.sleep(1)`
- typo'd or renamed names

Also: never self-links; skips `<a> <code> <pre> <script> <style> <head> <title> <h1>` (code, attributes, existing links untouched); case-insensitive match with verbatim visible text.

## Verification

- Manpages: 279 links, 89 targets, 0 broken, 0 self-links.
- Narrative (English): 6 links, all resolve; second build a clean no-op.
- Per-language: index-gating links only man pages translated in that language (e.g. a German narrative page links the German man page, into `de/man/`).
- A full `BUILD_DOCS_TRANSLATED=yes` build currently cannot complete due to a pre-existing po4a NAME-drift issue (a stray space in a German man page NAME yields a dangling `.so` alias), tracked in #4193 / #4178. Unrelated to this change; the links themselves render correctly per language.

## Also included

Source refs the linker surfaced as broken, fixed to the real pages:

- `motion(1)` => `motion(9)` in `hal_manualtoolchange(1)`
- `hal_thread_new(3)` => `hal_create_thread(3)` in `hal_add_funct_to_thread(3)`
- `stegen(9)` => `stepgen(9)` in the hostmot2 driver docs

## Notes

- Pre-existing explicit `link:` refs (e.g. `SSERIAL(9)`, `debounce(9)`) are left untouched: the postprocessor skips anything already inside an `<a>`.
- `hm2_uart_setup_chan(3)` in `hostmot2(9)` names no existing page; left as-is. @pcw-mesa stale/renamed name, or a missing page?

### DIFF
--- a/docs/src/extensions/man_xref.rb
+++ b/docs/src/extensions/man_xref.rb
@@ -13,9 +13,15 @@
 #     so code samples, tag attributes and existing links are left untouched.
 #   * Case-insensitive match ("AXIS(1)" => axis.1); visible text kept verbatim.
 #
+# Used for both the HTML manpages (sibling links under man/) and the narrative
+# manuals (User/HAL/Integrator), which reference man pages in the same
+# name(section) form and now link across into ../man/man<N>/.
+#
 # manxref-root (passed from the Submakefile) points at the troff man tree
 # docs/build/man[/<lang>], whose man<N>/ dirs enumerate every page including
 # generated component pages and .so stubs.  Absent => no-op, safe to always load.
+# manxref-linkbase is the relative path from the page to those man<N>/ dirs
+# (default "../" for a sibling manpage; narrative pages pass their own depth).
 
 require 'asciidoctor'
 require 'asciidoctor/extensions'
@@ -93,6 +99,12 @@ module LinuxCNCDocs
       self_name = (document.attr('mantitle') || '').downcase
       self_vol  = (document.attr('manvolnum') || '').to_s
 
+      # Relative path from this page to the man<N>/ dirs.  Manpages sit beside
+      # each other under man/, so the default reaches a sibling section dir;
+      # narrative pages pass their own depth-adjusted base (../man/, ../../man/).
+      base = document.attr('manxref-linkbase')
+      base = '../' if base.nil? || base.empty?
+
       self.class.each_editable_text(output) do |text|
         text.gsub(TOKEN) do
           whole = Regexp.last_match(0)
@@ -101,7 +113,7 @@ module LinuxCNCDocs
           # Never link a page to itself.
           next whole if name.downcase == self_name && sec == self_vol
           href = idx["#{name.downcase}\t#{sec}"]
-          href ? %(<a class="man-xref" href="../#{href}">#{whole}</a>) : whole
+          href ? %(<a class="man-xref" href="#{base}#{href}">#{whole}</a>) : whole
         end
       end
     end
--- a/docs/src/extensions/man_xref.rb
+++ b/docs/src/extensions/man_xref.rb
@@ -1,0 +1,113 @@
+# docs/src/extensions/man_xref.rb
+#
+# Asciidoctor postprocessor that turns manpage cross-references in the
+# conventional name(section) form (e.g. "halcmd(1)") into clickable links to
+# the sibling HTML page ../man<section>/<name>.<section>.html.
+#
+#   * Index-gated: a token is linked only when a page <name>.<section> exists
+#     in the troff tree, so false positives never resolve and stay plain text
+#     ("feed(2)"/"arc(3)" enum values in motion(9); external open(2)/udev(8);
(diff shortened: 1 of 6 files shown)


BsAtHome:
Some parts of the manual pages are a bit stale and needs cleaning up. In fact, I was doing exactly that for the hal_*(3) set of pages and adding a lot of them when this issue of linking caught my attention.

grandixximo:
Should I drop the doc fixes then? just have the script?

BsAtHome:
> Should I drop the doc fixes then? just have the script?

No, your fixes are just fine.

The "other" fixes is to read them all from start to end, fix inconsistencies, add missing, remove obsolete and reformat them to become uniform. But that is a different affair of incremental changes over a longer period.

BsAtHome:
One thing though, maybe add to lcnc-overrides.css:
```css
.man-xref {
  font-family: monospace;
  font-size: 90%;
}
```
We'll tackle the spurious bolds, italics and splits (`*name*(sect)`) is a cleanup session to make things consistent.

docs/src/extensions/man_xref.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+    # Split HTML into an alternating stream of text runs and tags, tracking a
+    # stack of PROTECTED elements; yield only text runs that are safe to edit.
+    def self.each_editable_text(html)
+      depth = 0
+      pos = 0
+      out = +''
+      html.scan(/([^<]+)|(<[^>]*>)/) do
+        text, tag = Regexp.last_match(1), Regexp.last_match(2)
```
BsAtHome: Isn't there supposed to be at least one char in between the '<' and '>' in a tag? How are you catching `<mytag attrib="<oops>">`?

The "proper" way to look at HTML is to use a HTML/XML parser and then find the text nodes matching the manpage(sec) pattern. That would be safer than regex parsing, which is known to be unstable for structured files like HTML.

Wouldn't it just be better to do this in a Tree-processor extension?

There are examples of tree-processors: https://github.com/asciidoctor/asciidoctor-extensions-lab/tree/main/lib/
You should also be able to check for the selected output format while it is running.

grandixximo: Good point on the tree-processor, I have rewritten the extension as one. It now runs on the parsed AST before conversion, so code blocks, monospace spans, passthroughs and existing links are excluded by node context instead of by splitting HTML with a regex, and it checks `document.backend` while running, as you suggested.

On the two concrete regex cases: they cannot occur in asciidoctor output. `>` inside an attribute value is escaped (`link:https://x/?a=1>b[t]` produces `href="https://x/?a=1&gt;b"`), and a stray `<` in text becomes `&lt;`, so neither `attrib="<oops>"` nor an empty `<>` reaches the output. The exception is raw `pass:[...]` passthrough, which the docs do use, so the concern was valid in principle even if no current input triggered it.

The rewrite also fixed a real bug this kind of parsing hid: list item and table cell text accessors apply substitutions, which had to be handled explicitly. Verified by rebuilding all 456 English manpages: link sets and visible text are unchanged compared to the postprocessor version.

grandixximo: One correction and one addition. The substitution issue I mentioned was introduced by my first treeprocessor draft (the #text getters on list items and table cells apply inline substitutions, so raw text must be rewritten instead), not something the postprocessor version hid. Verification since extended to the full htmldocs build: 537 English pages, no missed links, no broken targets, no self-links. Related asciidoctor gotcha found along the way: in compat-mode legacy 'single quotes' render as emphasis, not code, so manpage references written that way stay linkable.

BsAtHome:
Needs some conflict resolving care ;-)

grandixximo:
Done

grandixximo:
Any more blockers here?

BsAtHome:
> Any more blockers here?

Time and focus :-)

PR #4247, opened 2026-07-17
Draft - HAL types update - Remove 32-bit and move to 64-bit

BsAtHome:
Please note: This is a tracking branch that shows (many) cumulative changes in advance of being merged in smaller portions. Most changes are already merged (the entire getter/setter change and the query API). Additional changes will be visible here as the changes develop. Every time when changes are merged, then this tracking branch is rebased.

## Migrate all HAL pins/params to 64-bit exclusively

* Remove `HAL_[SU]32` and replace with `HAL_[SU]INT`. The old names will result is a compiler deprecation warning that they are no longer available.
* Rename `HAL_BIT` to `HAL_BOOL`, `HAL_FLOAT` to `HAL_REAL` and `HAL_[SU]64` to `HAL_[SU]INT`. The old names will result in a deprecation warning.
* Deprecate the old pin direction types `hal_pin_dir_t` and `hal_param_dir_t`. They have been replaced by the unified `hal_pdir_t`.
* Remove `hal_data_u`, `real_t` and `ireal_t`.
* Remove `hal_{bit,float,s32,u32,s64,u64}_t` old hal types.
* Retyped `hal_port_t` to match the new hal types.
* Remove the old API `hal_pin_XXX_new()`, `hal_pin_new()`, `hal_param_XXX_new()` and `hal_param_new()`.
* Remove `hal_param_set_XXX()` and `hal_param_set()` functions. These are replaced by the query API.
* Remove `hal_hal_get_pin_value_by_name()`, `hal_hal_get_param_value_by_name()`, `hal_hal_get_signal_value_by_name()`. These are replaced by the query API.
* Remove the `uses_fp` argument in `hal_export_funct()`, `hal_export_functf()` and `hal_create_thread()`. All threads are always floating point capable.
* Remove the `uses_fp` argument in `rtapi_task_new()` and the underlying `RtapiApp::task_new()` API.
* Update the converter components (conv_XXXX_YYYY.comp). Removes the old names and replaces them with new ones.
* Replace `abs_s32.comp` and `abs_s64.comp` with `abs_sint.comp`.
* A `conv_real_bool.comp` is added to convert a real value to boolean, just like an integer. The window of false is determined by an 𝜀-value set as a parameter (this is not the same as comp.comp).
* Update halcmd/halrmt to accept the new type names and map the old type names to the new type names.
* Update all sample configs to use the new type names.
* Update all Python code to use the new pin/param/signal types and change them to use the newly created IntEnums for both type and direction.
* Update documentation and remove the reference to the old types and replace them with the new types.
* Make halmodule emit a deprecation warning that the old type names `HAL_{BIT,FLOAT,S32,U32,S64,U64}` have been replaced with the `hal.Type.XXXX` IntEnum.
* Component pins using `si32` and `ui32` are adapted, when easily possible, to use plain `sint` and `uint`.
* Pin names of several components, notably `conv_*_*.comp`, `mux_generic` and `demux_generic`, are renamed to match the new type names. Several other components may follow if appropriate.
* Update `pyvcp` to recognize `sint`, `uint`, `image_bool` and `image_uint`. The old s32/u32/image_u32/image_bit are still recognized for compatibility. All in-tree xml files have been updated.
* Update various comments to use the new HAL type terminology.

## Getter/setter update (all done)
(original change set was about 30 kLOC)

This is the full tree conversion for the HAL types where #4099 would be the first step.
Changes:
* finalize (small minor) details regarding the new HAL types
* retire code that performs direct access to the underlying HAL memory for all HAL types
* covert all HAL components to the new API
* convert all kinematics to the new API
* convert all HAL drivers to the new API
* convert halmodule and other code that used the "old" pin/param access method(s)
* rewrite halrmt such that it is actually functional
* introduce a HAL query API for user-space, including hal_get_p, hal_set_p, hal_get_s and hal_set_s and more. This reduces code duplication and re-implementations that may or may not be fully identical
* make the HAL mutex recursive (for query callback reentry) and only accessible by the HAL library
* convert all code that included hal_priv.h and used direct HAL memory access to use the new query API
* revamp of the hal_*(3) man pages and full documentation of all the new API (old stuff will be removed eventually)

State of progress of these changes:
* the HAL library's internals are now fully private to the library
* the hal_data_u is never used in the code anymore (except for HAL library internal) and is ready to be retired
* the HAL API still retains full compatible with the "old" ways until the API break is performed
* all CI passes, but there may be some missed issues. Every (incremental) change obviously needs critical review before submitted
* waiting to move forward...

Missing in this tree:
* HAL_PORT/hal_port_t changes because they would be a breaking change


grandixximo:
Never cease to amaze me...
Are we having a meeting with #4099 as central discussion? We skipped a few weeks, not sure why @rmu75 
Seems about time... or is everyone on summer vacation? Probably better send to the mailing list...

rmu75:
On Fr, 2026-07-17 at 07:08 -0700, Luca Toniolo wrote:
> grandixximo left a comment (LinuxCNC/linuxcnc#4247)
> Never cease to amaze me...
> Are we having a meeting with #4099 as central discussion? We skipped
> a few weeks, not sure why @rmu75
> Seems about time... or is everyone on summer vacation? Probably
> better send to the mailing list...


I had some unexpected troubles with "remote" internet access around
21st of june and it seems I got a bit sidetracked... sorry for that. 


Last meeting I have notes for was 2026-06-07 10:00 CEST

According to our cadence, 2026-06-21 would have been late, 2026-07-05
early and the next meeting would be tomorrow "late" at 20:00 CEST.


I'm not sure I can make it this weekend but I will try.


I think luca wants to discuss the hal interface redesign, so maybe
change of plan and meeting at 10:00 would be better? please reply
whoever can make it at what time.



https://greenlight.bbb.uni-rostock.de/b/ste-c4d-brs-3k6
Access code: 869782

(hope this is still active)


best regards


-- 
Robert Schöftner ***@***.***>

grandixximo:
Should send to the mailing list, my bad for starting it here 😞

src/hal/utils/halcmd_commands.cc
```diff
@@ -1457,25 +1212,25 @@ static int unloadrt_comp(char *mod_name)
...
-        rtapi_mutex_get(&(hal_data->mutex));
-        comp = halpr_find_comp_by_name(mod_name);
-        if(comp) type = comp->type;
-        rtapi_mutex_give(&(hal_data->mutex));
-        if(type == COMPONENT_TYPE_UNKNOWN) {
-            halcmd_error("component '%s' is not loaded\n",
-                mod_name);
+        hal_query_t q = {};
+        int rv = hal_comp_by_name(mod_name, &q);
+        if(!rv) {
```
grandixximo: hal_comp_by_name returns 0 when found, therefore won't this treats success as an error? Or am i missing something?

src/hal/hal_lib_query.c
```diff
@@ -0,0 +1,1072 @@
...
+//=====================================================
+
+static int set_common(hal_type_t type, const hal_query_value_u *v, hal_refs_u u, bool setport)
+{
+    switch(type) {
+    case HAL_BOOL: hal_set_bool(u.b, v->b); break;
+    case HAL_S32:  hal_set_si32(u.s, v->s); break;
+    case HAL_U32:  hal_set_ui32(u.u, v->u); break;
+    case HAL_SINT: hal_set_sint(u.s, v->s); break;
+    case HAL_UINT: hal_set_uint(u.u, v->u); break;
```
grandixximo, line 48: If I understand correctly, these all store through the 8-byte union member in hal.h
AFAIK old-API params keep caller-sized storage hal_bit_t=1B, hal_s32_t=4B. So halcmd setp / hal_set_p on any out-of-tree legacy BIT/S32/U32 param overruns it corrupting adjacent shmem. Won't that already break the old API?

src/hal/halmodule.cc
```diff
@@ -1327,522 +1260,359 @@ PyObject *disconnect(PyObject * /*self*/, PyObject *args) {
...
-
-        // add to list
-        PyList_Append( python_list, obj);
-        next = pin->next_ptr;
+    hal_query_t q = {};
+    q.qtype = HAL_QTYPE_PIN; // Only handle pins
+    int rv = hal_list_p(&q, pinparaminfo_cb, python_list);
+    if(0 != rv) {
+        Py_DECREF(python_list);
+        PyErr_Format(PyExc_RuntimeError, "hal_list_p: returned '%s' (%d)", rv, hal_strerror(rv));
```
grandixximo: `"returned '%s' (%d)", rv, hal_strerror(rv)` passes the int to `%s`. You have a few more of these copy pasta in this file.

src/hal/halquery.c
```diff
@@ -0,0 +1,614 @@
...
+static int get_threads_cb(hal_query_t *q, void *arg)
+{
+    threadscb_t *tcb = (threadscb_t *)arg;
+
+    if(HAL_QTYPE_THREAD == q->qtype) {
+        if(tcb->functs) {
+            // Seen functions from previous thread
+            PyObject *obj = build_dict_thread(&tcb->qt, tcb->functs);
+            PyObject *str = PyUnicode_FromString(tcb->qt.name);
+            PyDict_SetItem((PyObject *)arg, str, obj);
```
grandixximo: I think the dict insert uses the wrong target, `arg` is the callback state struct, not the threads dict, so every thread except the last fails to insert and leaves a Python error pending, should be `tcb->threads` ?

src/hal/halquery.c
```diff
@@ -0,0 +1,614 @@
...
+    if(!strcmp(tlp->name, q->name)) {
+        if(q->qtype == HAL_QTYPE_THREAD) {
+            // Thread construct
+            tlp->qt = *q;
+            tlp->dict = PyDict_New();
+            if(!tlp->dict)
+                return -ENOMEM;
+        } else if(q->qtype == HAL_QTYPE_THREAD_FUNCT) {
+            // Attached function to thread
+            PyObject *obj = build_dict_threadfunct(q);
```
grandixximo: This keys each function entry by the *thread* name instead of the function name, so N functions collapse into one dict entry (last one wins)

src/hal/halquery.c
```diff
@@ -0,0 +1,614 @@
...
+    int rv = hal_list_thread(&q, get_thread_cb, &tl);
+    if(rv >= 0) {
+        // Return value can be zero if the last in the list was the one we
+        // wanted. Iteration stops by end-of-list and not by a name change.
+        if(!tl.dict) {
+            // List not created means name not found
+            Py_INCREF(Py_None);
+            return Py_None;
+        }
+        return tl.dict;
```
grandixximo: This returns the raw functs dict instead of the wrapped thread dict when the requested thread happens to be last in the list. Net effect: `threads()`/`thread()` return incomplete or wrongly-shaped results depending on list order.

src/emc/rs274ngc/interp_namedparams.cc
```diff
@@ -246,68 +246,66 @@ int Interp::fetch_ini_param( const char *nameBuf, int *status, double *value)
...
-    }
-    char *s;
-    int n = strlen(nameBuf);
-    if ((n > 6) &&
-	((s = (char *) strchr(&nameBuf[5],']')) != NULL)) {
 
+    // Make sure we can query pins/params/signals
+    retval = hal_lib_init();
+    CHKS(retval < 0, "fetch_hal_param: hal_lib_init(): error=%d", retval);
+
```
BsAtHome, line 273: FIXME: init/exit is reference counted

src/hal/hal_lib_query.c
```diff
@@ -0,0 +1,1072 @@
...
+            halpr_mutex_release();
+            return -ENOENT;
+        }
+        // We have the param, copy the data, but we still can fail
+        q->qtype     = HAL_QTYPE_PARAM;
+        q->name      = param->name;
+        q->pp.type   = t = param->type;
+        q->pp.ref    = (hal_refs_u)(hal_sint_t)SHMPTR(param->data_ptr);
+        q->pp.dir    = param->dir;
+        q->pp.alias  = param->oldname ? (const char *)SHMPTR(param->oldname) : NULL;
```
grandixximo: `SHMPTR(param->oldname)` points at the `hal_oldname_t` struct, whose first field is `next_ptr` not the name.

src/hal/hal_lib_query.c
```diff
@@ -0,0 +1,1072 @@
...
+        // FIXME: This can be targeted to param's data when the old API is retired
+        u = (hal_refs_u)(hal_sint_t)SHMPTR(param->data_ptr);
+    } else {
+        // We have a pin, copy the data, but we still can fail
+        q->qtype     = HAL_QTYPE_PIN;
+        q->name      = pin->name;
+        q->pp.type   = t = pin->type;
+        q->pp.ref    = (hal_refs_u)(hal_sint_t)&pin->dummysig;
+        q->pp.dir    = pin->dir;
+        q->pp.alias  = pin->oldname ? (const char *)SHMPTR(pin->oldname) : NULL;
```
grandixximo: same as above...

src/hal/utils/halrmt.cc
```diff
@@ -0,0 +1,3087 @@
...
+    int cl = -1;
+    if (ctx.toks.size() > 4 && !to_int(ctx.toks[4], cl)) {
+        errornl(ctx, fmt::format("Invalid argument '{}', must be integer", ctx.toks[4]));
+        return rtError;
+    }
+    if(cl < -1) {
+        errornl(ctx, fmt::format("Invalid argument '{}', must be >= -1", ctx.toks[4]));
+        return rtError;
+    }
+    int rv = hal_add_funct_to_thread(ctx.toks[2].c_str(), ctx.toks[3].c_str(), cl);
```
grandixximo, line 2205: are toks 2 and 3 always passed?

src/hal/utils/halrmt.cc
```diff
@@ -0,0 +1,3087 @@
...
+        errornl(ctx, fmt::format("Failed to add fucntion '{}' to thread '{}', error=%d", ctx.toks[2], ctx.toks[3], rv));
+        return rtError;
+    }
+    return rtOk;
+}
+
+static cmdResponseType setDelF(connectionRecType &ctx)
+{
+    // SET DELF <name> <thread>
+    int rv = hal_del_funct_from_thread(ctx.toks[2].c_str(), ctx.toks[3].c_str());
```
grandixximo, line 2216: same question as above...

src/hal/utils/halrmt.cc
```diff
@@ -0,0 +1,3087 @@
...
+            if (retval != 0) {
+                if(EINTR == err) {
+                    continue;
+                }
+                exited = true;
+                ready = false;
+                errornl(ctx, fmt::format("LoadUsr: waitpid failed '{}' ('{}') errno={}", argv[0], new_comp_name, err));
+                break;
+            }
+            if(WIFEXITED(status)) {
```
grandixximo: is `status` already initialized here?

src/hal/utils/halrmt.cc
```diff
@@ -0,0 +1,3087 @@
...
+            flushall();
+        }
+        // did it work?
+        if (!ready) {
+            if (retval < 0) {
+                errornl(ctx, fmt::format("LoadUsr: waitpid({}) failed errno={} ({})", pid, err, strerror(err)));
+            } else {
+                errornl(ctx, fmt::format("LoadUsr: {} exited without becoming ready", argv[0]));
+            }
+            return -err;
```
grandixximo: This returns whatever `errno` happens to hold, if the component never became ready but no syscall failed, that's 0, will `loadusr -W` report **success** for a component that never came up?

src/hal/utils/halrmt.cc
```diff
@@ -0,0 +1,3087 @@
...
+    bool exited = false;
+    int err;
+    if(wait_comp_flag) {
+        bool ready = false;
+        int count = 0;
+        int retval = 0;
+        while(!ready && !exited) {
+            mysleep(0.01);
+            // Program done?
+            retval = waitpid(pid, &status, WNOHANG);
```
grandixximo: What happens when the child exits normally? Doesn't the code falls into the "waitpid failed" branch with a stale `errno`?

src/hal/utils/halrmt.cc
```diff
@@ -0,0 +1,3087 @@
...
+    }
+
+    // Create the signal if it does not yet exist
+    if(-ENOENT == havesig) {
+        hal_query_t qp = {};
+        qp.name = ctx.toks[2].c_str();
+        int rv = hal_getref_p(&qp);
+        if(0 == rv) {
+            errornl(ctx, fmt::format("Signal name '{}' is already in use as a {} name",
+                            ctx.toks[2], HAL_QTYPE_PIN == qp.qtype ? "pin" : "param"));
```
grandixximo, line 2012: missed `return rtError;`

src/hal/utils/halcmd_commands.cc
BsAtHome: Indeed, the if should be `if(0 != rv)`

src/hal/hal_lib_query.c
BsAtHome: The HAL allocation has been changed never ever to allocate a chunk less than 8 bytes (and at least 8-byte aligned). So any write is always safe.

Older code may indeed write to the smaller part. That is why the _getters_ also only address the smaller part. The _setters_ will always write to the larger type (even bool). We are assured of always full writes only when the old code is disallowed (removal of hal_data_u and the old hal types). In the transition, we only have compatibility to assure.

And, old style code cannot link to HAL unless recompiled. The HAL memory version has been changed.

src/hal/halmodule.cc
BsAtHome: Check.

src/hal/halquery.c
BsAtHome: Good catch!

src/hal/hal_lib_query.c
grandixximo: The 8-byte minimum covers standalone allocations, but how does it protect this case?

```c
struct comp_data {
    hal_s32_t speed;   // hal_param_s32_new(..., &speed) -> 4-byte storage
    hal_s32_t accel;   // adjacent field, same hal_malloc'd struct
};
```

A recompiled legacy component still exports fields *inside* one larger struct allocation. When `hal_set_si32` writes the sign-extended 64-bit value at `&speed`, where do the upper 4 bytes (`0xFFFFFFFF` for -1) land? Does the compiler fixes this? Or will it not compile?

src/hal/halquery.c
BsAtHome: The function retrieval must be done by list. That was changed in the get_thread<b><i>s</i></b> version but not in the get_thread one.

src/hal/halquery.c
BsAtHome: Fixed with other issue.

src/hal/hal_lib_query.c
BsAtHome: Nice one... did it right other places.

src/hal/utils/halrmt.cc
BsAtHome: Now it is :-)
Argument are `<funct> <thread> [order]`. Was set to one arg instead of two minimum.

src/hal/utils/halrmt.cc
BsAtHome: Yes, from the waitpid() call in line 768.

src/hal/utils/halrmt.cc
BsAtHome: errno=0 now set prior to waitpid call. The err=errno copy will now be correct.

src/hal/utils/halrmt.cc
BsAtHome: Good question... not my code ;-)
Needs to be investigated a bit more. The whole loadusr code is fishy in both halcmd and halrmt.

src/hal/utils/halrmt.cc
BsAtHome: Indeed.

src/hal/hal_lib_query.c
BsAtHome: The case where new code will write an old style _param_. That is the problem. Need to think a bit more on the scenario.

BsAtHome: As long as the support programs (mostly halcmd and halmodule) are "old style", then there is no problem because they will dereference a hal_data_u construct and write the smaller type.

The new library with hal_set_p() abstracted needs to divert the new setter construct and use hal_data_u dereferencing for parameters until the API break. Then you won't have any problems.

BsAtHome:
@grandixximo the whole `loadusr` code needs to be reviewed separately. The constructs in both halcmd and halrmt feel off seem to have a bad smell. I don't think all cases of what can happen are caught correctly in either program. Luckily, it will be some time before this code will make its way. So time to put the thinking cap on and see if it can be made into something smelling fresh and likable ;-)

grandixximo:
Thanks for the quick turnaround, checked the new push. The `unload` inversion, `PyErr_Format` swaps, alias pointers, `halquery` threads (all three), SET NET, and the ADDF bounds are all correctly fixed. On the legacy-param setter: agreed with your conclusion, diverting `hal_set_p` to type-width `hal_data_u` writes for old-style params until the API break closes the hole.

Two Out-of-Bounds of the same family as ADDF remain:

1. **DELF**: table still declares `nset=1` (halrmt.cc:2215) but `setDelF` reads `ctx.toks[3]` unconditionally. Same fix as ADDF: `nset=2`.
2. **SET SAVE**: error path formats `ctx.toks[3]` (halrmt.cc:2092) with `nset=1`, so `SET SAVE badkey` without a filename reads out of bounds.

Minor: literal `%d` inside `fmt::format` strings (e.g. setAddf's error message) prints verbatim and drops `rv`, there are a few in `src/hal/utils/halrmt.cc` fix is mechanical, replace %d with {}.

Loadusr: agreed, the whole path wants a proper rewrite, here or in a follow up.

BsAtHome:
Should be all fixed now, except the param write in hal_set_p, which will take more than a mere moment to alter.

rene-dev:
I just tested this branch on a machine with smartserial slaves. the driver loads, and creates all the pins, but nothing works. IO are not getting to the hardware. I cannot see any error messages.

BsAtHome:
> I just tested this branch on a machine with smartserial slaves. the driver loads, and creates all the pins, but nothing works. IO are not getting to the hardware. I cannot see any error messages.

Thanks for testing.
At least the pin creation went well and it didn't crash. That is not entirely bad.

But, I haven't run locally on the hardware for some time. Need to check up on that. Also, the changes in hostmot2 have not yet been properly reviewed, so there can be lurking any detail in any layer of the hm2 driver stack. I'll give a peep when this can be tested again.

src/hal/drivers/mesa-hostmot2/abs_encoder.c
```diff
@@ -159,8 +159,8 @@ int hm2_absenc_setup_ssi(hostmot2_t *hm2, hm2_sserial_remote_t *chan,
     chan->data_written[0] = 0;
 
     
-    chan->params->float_param = 500;
-    chan->params->timer_num = 0;
+    hal_set_real(chan->params->param.r, 500);
+    hal_set_ui32(chan->params->timer_num, 0);
```
grandixximo: `param.r` (frequency-khz)  and `timer_num` are **handles that are only created later** in `hm2_absenc_parse_md()` (abs_encoder.c:~455-465).

If @rene-dev 's test setup has absenc channels configured, this is the prime suspect for IO not working

src/hal/drivers/mesa-hostmot2/abs_encoder.c
```diff
@@ -186,8 +186,8 @@ int hm2_absenc_setup_biss(hostmot2_t *hm2, hm2_sserial_remote_t *chan,
             + (3 * md->register_stride);
     chan->data_written[0] = 0;
     
-    chan->params->float_param = 500;
-    chan->params->timer_num = 0;
+    hal_set_real(chan->params->param.r, 500);
+    hal_set_ui32(chan->params->timer_num, 0);
```
grandixximo: same as above

src/hal/drivers/mesa-hostmot2/abs_encoder.c
```diff
@@ -219,15 +219,13 @@ int hm2_absenc_setup_fabs(hostmot2_t *hm2, hm2_sserial_remote_t *chan,
...
+    if (hal_param_new_ui32(hm2->llio->comp_id, HAL_RW, &(chan->params->param.u),
+            0, "%s.filter", chan->name)){
         HM2_ERR("error adding param fanuc param 2, aborting\n");
         return -EINVAL;
     }
-    chan->params->float_param = 1024.0;
-    chan->params->u32_param = 0xF;
-    chan->params->timer_num = 0;
+    hal_set_ui32(chan->params->param.u, 0xF);
+    hal_set_ui32(chan->params->timer_num, 0);
```
grandixximo: same as above

src/hal/hal_lib.c
```diff
@@ -1359,13 +1571,13 @@ int hal_link(const char *pin_name, const char *sig_name)
         case HAL_U32:
             *((hal_u32_t *) data_addr) = pin->dummysig.u;
             break;
-        case HAL_S64:
+        case HAL_SINT:
             *((hal_s64_t *) data_addr) = pin->dummysig.s;
```
grandixximo: should this be .ls or that is after the transition period?

src/hal/hal_lib.c
```diff
@@ -1359,13 +1571,13 @@ int hal_link(const char *pin_name, const char *sig_name)
...
             break;
-        case HAL_U64:
+        case HAL_UINT:
             *((hal_u64_t *) data_addr) = pin->dummysig.u;
```
grandixximo: same as above, should this be `.lu` ?

grandixximo:
@rene-dev 

Does your test config string enable absenc (ssi/biss/fanuc formats)?
What does `halcmd show pin hm2_*.sserial.port-0.port_state` shows?
Could you grep `dmesg` for `pin_new(...) called with already-initialized memory` or `Timeout waiting for CMD to clear`

src/hal/components/reset.comp
```diff
@@ -37,18 +37,18 @@
...
-pin io  bit     out_bit      = false  "Bit integer output value";
-pin in  bit     reset_bit    = false  "Bit reset value";
-pin in  bit     retriggerable = true  "Allow additional edges to reset";
-pin in  bit     rising        = true  "Trigger on rising edge";
-pin in  bit     falling       = false "Trigger on falling edge";
+pin in  bool trigger             "Trigger input";
+pin io  ui32 out_u32      = 0    "Unsigned 32 bool integer output value";
+pin in  ui32 reset_u32    = 0    "Unsigned 32 bool integer reset value";
+pin io  si32 out_s32      = 0    "Signed 32 bool integer output value";
+pin in  si32 reset_s32    = 0    "Signed 32 bool integer reset value";
```
grandixximo: 32 bit integers were hit by find/replace

src/hal/components/histobinstream.comp
```diff
@@ -46,32 +46,32 @@ Maintainers note: hardcoded for MAXBINNUMBER==200
...
-pin out bit   stream_error "Set if the data stream did not fit and was never sent.";
-pin out bit   input_error  "Set when input rules are violated; updates are invalid until fixed and a reset.";
-
-pin out float input_min    "Minimum input value seen (latched on stream rising edge).";
-pin out float input_max    "Maximum input value seen (latched on stream rising edge).";
-pin out u32   nsamples     "Number of samples accumulated (latched on stream rising edge).";
-pin out float variance     "Variance of the input (latched on stream rising edge).";
-pin out float mean         "Mean of the input (latched on stream rising edge).";
-pin out s64   stream_time  "The time it took for the stream data to be pushed out in nanoseconds (ns).";
+pin in  ui32 pintype      "Select the input pin (0:input 1:input-si322:input-u32 3:input-bit).";
```
grandixximo: input-s32 2 hit by find/replace

src/hal/hal_lib.c
BsAtHome: No, this was already wrong in the old version. Or, said in another way, the current code is faulty.

But, luckily, usually a pin gets written and then sets the value correctly.

src/hal/drivers/mesa-hostmot2/abs_encoder.c
BsAtHome: You may be right... The ordering of assignment and the creation of the param seems to be completely screwed..

BsAtHome:
> I just tested this branch on a machine with smartserial slaves. the driver loads, and creates all the pins, but nothing works. IO are not getting to the hardware. I cannot see any error messages.

The hostmot2 code has now been reviewed and is functional on my Rpi5 with a 7i96s and a connecting hm2_modbus device.

There were detected several smaller issues, primarily parameters that were accessed before being allocated. These should all be caught now, I think.
Some parameters were not allocated at all in certain configurations and still dereferenced, which now fails with parameters behaving as pins. I already created `hal_param_new_fake()` for just these occasions, so that storage is available for code that is hard to change. It just needed to be used several more places.

If you have any sserial fanuc style absolute encoders, then you should be aware that there was made an additional change in that code path. The configuration used an overlapping floating point value with an unsigned value. That construct has been unraveled and the bit-rate calculation in control register 1 has been changed so it no longer depends on overlapping values. If this causes problems, please let me know.

grandixximo:
@rene-dev another test attempt on smartserial if you could, whenever you have a minute...

grandixximo (commented):
Did the review below...

src/hal/halquery.cc
```diff
@@ -0,0 +1,484 @@
...
+    py::list flst;
+    size_t tag = 0;
+    for(size_t i = 1; i < qrec.size(); i++) {
+        if(HAL_QTYPE_THREAD == qrec.rec(i)->qtype) {
+            tag = i;
+        } else if(HAL_QTYPE_THREAD_FUNCT == qrec.rec(i)->qtype) {
+            flst.append(build_dict_threadfunct(qrec.rec(i)));
+            if(i == qrec.size() - 1 || (i < qrec.size() - 1 && HAL_QTYPE_THREAD == qrec.rec(i+1)->qtype)) {
+                // There is no next enrty or next entry is new thread
+                dict[qrec.rec(tag)->name] = build_dict_thread(qrec.rec(tag), flst);
```
grandixximo: A dict entry is only written inside the `HAL_QTYPE_THREAD_FUNCT` branch. What happens to a thread whose `THREAD` record is followed directly by the next `THREAD` record, or sits last in the list? As far as I can tell it never appears in the result. Should the pending thread be emitted when the next `THREAD` record shows up, independent of whether any functions were collected?

src/hal/halquery.cc
```diff
@@ -0,0 +1,484 @@
...
+    return dict;
+}
+
+static py::object build_dict_pin_param(const hal_query_t *q, bool ispin)
+{
+    py::dict dict;
+    dict["haltype"] = ispin ? "pin" : "param";
+    dict["name"]    = q->name;
+    dict["type"]    = (int)q->pp.type;
+    dict["dir"]     = (int)q->pp.dir;
```
grandixximo: `type` and `dir` go out as plain ints (same at line 49 for comps and line 77 for signals). Should these carry the `hal.type` / `hal.dir` IntEnum members instead? That is where I planned to hook the #4251 enum classes in after this lands: rebase, tag these spots with the shared classes, and query results come out as `HALType`/`HALDir` members. Note the comp dict `type` at line 49 is the component type, not a `hal_type_t`, so that one stays an int either way.

src/hal/Submakefile
```diff
@@ -26,7 +26,16 @@ PYSRCS += $(HALMODULESRCS)
 HALMODULE := ../lib/python/_hal.so
 $(HALMODULE): $(call TOOBJS, $(HALMODULESRCS)) $(HALLIB)
 	$(ECHO) Linking python module $(notdir $@)
-	$(Q)$(CXX) $(LDFLAGS) -shared -o $@ $^
+	$(Q)$(CXX) $(LDFLAGS) -shared -o $@ $^ -lfmt
+
+HALQUERYSRCS := hal/halquery.cc
```
grandixximo: `halquery.so` builds unconditionally. What does a source build on a host without python3-pybind11 do now, given configure never probes for the headers? Debian is covered via `debian/control`, but elsewhere it looks like a mid-compile missing-header failure rather than a clean configure diagnostic. Worth a configure probe, or a note in the build docs?

src/hal/halquery.cc
```diff
@@ -0,0 +1,484 @@
...
+    HalQRec *qrec;
+} threadlist_t;
+
+static int get_thread_cb(hal_query_t *q, void *arg)
+{
+    threadlist_t *tlp = reinterpret_cast<threadlist_t *>(arg);
+    if(!strcmp(tlp->name, q->name)) {
+        // The first match has q->qtype set to HAL_QTYPE_THREAD. The following
+        // matches with have it set to HAL_QTYPE_THREAD_FUNCT.
+        tlp->qrec->append(q);
```
grandixximo: The return value of `qrec->append()` is ignored here, while `get_qrec_cb()` propagates it. On -ENOMEM, is silently truncating the thread's function list the intended behavior, or should the error go back through `hal_list_thread()` like everywhere else?

src/hal/halquery.cc
```diff
@@ -0,0 +1,484 @@
...
+static py::object value_object(hal_type_t type, hal_refs_u u)
+{
+    switch(type) {
+    case HAL_BOOL: return py::int_(hal_get_bool(u.b));
+    case HAL_S32:  return py::int_(hal_get_si32(u.s));
+    case HAL_SINT: return py::int_(hal_get_sint(u.s));
+    case HAL_U32:  return py::int_(hal_get_ui32(u.u));
+    case HAL_UINT: return py::int_(hal_get_uint(u.u));
+    case HAL_REAL: return py::float_(hal_get_real(u.r));
+    case HAL_PORT: return py::int_(0);  // FIXME
```
grandixximo: Is a wrong value better than no value here? `py::none()` would at least not lie until port support lands.

src/hal/halqrec.hh
```diff
@@ -0,0 +1,95 @@
...
+        n++;
+        return 0;
+    }
+
+    size_t size() const { return n; }
+    size_t maxsize() const { return na; }
+    const hal_query_t *rec(size_t i) const {
+        if(i < n)
+            return &qr[i];
+        throw std::runtime_error(fmt::format("HalQRec: Index {} out of range [0,{}]", i, n-1));
```
grandixximo: Nit: with `n==0`, what does `n-1` print in the range message? I believe `[0, 18446744073709551615]`. Cosmetic.

grandixximo (commented):
Checked and not a problem: `value_object()` has no duplicate-case risk (`HAL_SINT`/`HAL_UINT` are unconditionally `HAL_S64`/`HAL_U64` on master now). The piecemeal `py::dict` style dodges the cppcheck portability warning as commented in pm. The `get_info_*` semantics in halmodule.cc (list of dicts, same keys) are unchanged, only the collection moved behind the HalQRec snapshot. CI green on a7a528d27f.

src/hal/halquery.cc
BsAtHome: Yes, indeed. The whole - is there a next - condition must be handled in both cases to capture the lone thread case. Fixed.

src/hal/halquery.cc
BsAtHome: Yes, as I mentioned, still need to integrate the enums.

src/hal/Submakefile
BsAtHome: Yes, this is a test build so far.
Need to have a configure test to make sure.

src/hal/halquery.cc
BsAtHome: Yes, should return the error. Fixed.

src/hal/halquery.cc
BsAtHome: `None` would be an option, it is the "error" marker.

Implemented the halmodule version. Still needs to be fixed later.

src/hal/halqrec.hh
BsAtHome: Yeah, differentiating between the n==0 and n>0 cases.

src/hal/halquery.cc
BsAtHome: Done

src/hal/Submakefile
BsAtHome: Now, the real problem I'm having is that the halquery.cc file is a normal module that has a a `PyInit_halquery` entry.

I'm currently at a loss how to create only the submodule with pybind11 that can be added in halmodule using a PyModule_AddObject() call (and does it need a PyType_Ready() call too?).

Your version uses a fake module created within the bounds of halmodule. That does not work here, I believe.

grandixximo: `PYBIND11_MODULE(halquery, m)` is only a macro: it emits `PyInit_halquery` plus a function body that takes a `py::module_ &`. Nothing else in pybind11 depends on that entry point, so you can write the function yourself and hand it the `_hal` module object.

In `halquery.cc`, drop the macro and keep the body:

```cpp
extern "C" int halquery_add_submodule(PyObject *parent)
{
    try {
        py::module_ p = py::reinterpret_borrow<py::module_>(parent);
        py::module_ m = p.def_submodule("query", halquery_module_doc);

        m.def("pin",   &get_pin,   "Get information about the named pin or None if not found.");
        // ... the rest of the defs verbatim

        halenumtype = p.attr("type");
        halenumdir  = p.attr("dir");
        halenumcomp = p.attr("comptype");
        return 0;
    } catch(const std::exception &e) {
        PyErr_SetString(PyExc_ImportError, e.what());
        return -1;
    }
}
```

`def_submodule()` already does the registration you were going to do by hand: it builds `PyModule_GetName(parent) + ".query"`, calls `PyImport_AddModule()` on it, so `_hal.query` lands in `sys.modules` and `import _hal.query` resolves, and it assigns the result as the `query` attribute of the parent. So no `PyModule_AddObject()` of your own.

No `PyType_Ready()` either. That is for static types you define with the plain C API; pybind11 finalizes its own types inside `py::class_`, and halquery has no classes, only free functions.

Then in `halmodule.cc`, in place of my placeholder block:

```cpp
if(halquery_add_submodule(m) < 0) {
    Py_DECREF(m);
    return NULL;
}
```

Two ordering points:

- Call it after `type`, `dir` and `comptype` have been added to `m`. That is also why the enums are read from `parent` above rather than with `py::module_::import("_hal")`: during `PyInit_hal` the module is not in `sys.modules` yet, so that import would re-enter the import machinery.
- The GIL is held throughout module init, which is all pybind11 needs, and `hal_lib_init()`/`hal_lib_exit()` stay exactly as they are now that everything lives in one module.

Build side: fold `hal/halquery.cc` into `HALMODULESRCS` and drop the `HALQUERY` target and its `PYTARGETS` entry. One `.so` instead of two also removes the question of two pybind11 modules sharing internals across separate objects.

Unrelated to the mechanics, but it comes along for the ride: the three file-scope `static py::object` globals are destroyed at library unload, after `Py_Finalize()`, which is a `Py_DECREF` without an interpreter. Either release them from an atexit-style hook or make them intentionally leaked (`static py::object *`).

The fake module in my tree was only ever a placeholder for exactly this. Once the real bindings live in `_hal`, the `PyModule_New()` block goes away and the only piece that stays is the `sys.modules["hal.query"]` alias in `hal.py`, because `hal` is a plain module and not a package.

BsAtHome: > `PYBIND11_MODULE(halquery, m)` is only a macro: it emits `PyInit_halquery` plus a function body that takes a `py::module_ &`. Nothing else in pybind11 depends on that entry point, so you can write the function yourself and hand it the `_hal` module object.

_**That**_ was what I was missing. Thanks!

I had the idea that pybind11 would have had a construct for this as it probably isn't an impossible situation.

> Two ordering points: [snip]

yes... all fixed now...


> Unrelated to the mechanics, but it comes along for the ride: the three file-scope `static py::object` globals are destroyed at library unload, after `Py_Finalize()`, which is a `Py_DECREF` without an interpreter. Either release them from an atexit-style hook or make them intentionally leaked (`static py::object *`).

Yes, I thought about that they would leak. If they are not decref'ed, then they hold the enums and therefore the _hal module. But, it does not prevent `Py_AtExit` to call the `hal_lib_exit()` function, which is the real issue. The only time halmodule is unloaded is at program exit (or crash). Not sure if it is essential to do a full deconstruction. Does it matter?

grandixximo: No, it does not matter, and my remark was too strong: I rebuilt the pattern outside LinuxCNC and it is harmless.

What I am less sure of is whether we want to depend on *why* it is harmless. The static destructor does run after finalization, but the object still carries 10 references, so the decref never reaches `tp_dealloc`. That count is a CPython artifact: an IntEnum class is self-referential through its members and `_value2member_map_`, and the C-held reference sits outside the cycle, so the collector cannot take it either. It holds with the parent module on multi-phase init and dropped from `sys.modules` as well; the only shape I could get to abort was a plain acyclic object whose sole owner is the static.

That shape stops aborting when the global is a raw `PyObject *`, incref'd once at init and never released: same object, same sole ownership at refcount 1, exit 0 instead of `Fatal Python error: PyThreadState_Get`. It is `py::object(...).release().ptr()` at init, and the call sites take a `py::handle(...)` wrap; tagging still returns `<HalType.BOOL: 1>` as before.

So, a question I cannot answer for you: is "the refcount is never 1" something you want in the trust base, or would you rather those three globals never run a destructor at all? The pointer version is the same leak you have already accepted, without the dependency. Is there something that argues for keeping them as `py::object`?

BsAtHome: I'd hate to use raw PyObject pointers. Afaik, using py::object should have increased the refcount, so when the destructor (tries to) kills it, the underlying object should remain. Alternatively, we can enforce it by doing:
```c++
    halenumtype = p.attr("type");
    halenumdir  = p.attr("dir");
    halenumcomp = p.attr("comptype");
    halenumtype.inc_ref();
    halenumdir.inc_ref();
    halenumcomp.inc_ref();
```
Then we don't care about the C++ destructor and purposefully leak the reference.

grandixximo: Agreed, and better than my version: no call site changes and the C++ type stays. I ran it against the shape that aborts, a plain acyclic object whose only owner is the static, and with the `inc_ref()` it goes from `Fatal Python error: PyThreadState_Get` to refcount 2 at init, 1 after the destructor, exit 0, clean under `PYTHONMALLOC=debug`.

One thing to be careful about in the reasoning, though. The reference `py::object` takes is exactly the one its destructor gives back, so on its own it nets to zero and protects nothing; the object survives today only because something else still holds it. Your `inc_ref()` works because it adds a *second* reference, not because the first one is there.

Which is why I would put a line of comment next to the three calls. They read as redundant, and the next person tidying up will delete them.

grandixximo:
is it time to do the `~pre2` bump as well?

BsAtHome:
> is it time to do the `~pre2` bump as well?

What again was the criteria to bump? Deprecation or removal of the old API?

grandixximo:
:thinking: I was jumping ahead, this does not break the API yet...
Criteria plan was/is :
`#define HAL_API_VERSION 1` compile-time anchor for out-of-tree code during the deprecation period
`1:2.10.0~pre2` bump at the API-break PR, so not here.

PR #4249, opened 2026-07-18
orient.comp: Add modes with preceding referencing to index signal

Sigma1912:
Some spindle encoder setups cannot track the encoder position when using high spindle speeds and thus cannot use the current 'orient' component due to lost encoder position.
This PR expands the orient component with three additional modes that offer referencing to an index signal prior to orienting the spindle.

Current behavior is unchanged. All modes have been tested on real hardware.

Based on work by @andypugh 

For more:
https://forum.linuxcnc.org/10-advanced-configuration/42792-m19-and-spindle-index-enable-issues?start=10

### DIFF
--- a/src/hal/components/orient.comp
+++ b/src/hal/components/orient.comp
@@ -1,16 +1,21 @@
 component orient "Provide a PID command input for orientation mode based on current spindle position, target angle and orient mode";
 
 pin in  bit   enable      "enable angular output for orientation mode";
-pin in  s32   mode        "0: rotate - shortest move; 1: always rotate clockwise; 2: always rotate counterclockwise";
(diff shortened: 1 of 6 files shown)


src/hal/components/orient.comp
```diff
@@ -63,31 +69,85 @@ license "GPL";
 
 FUNCTION(_) {
 
-    double target_angle;
-    double latched_position;
+    static double target_angle;
+    static double latched_position;
```
BsAtHome: This is not a singleton component so these should be instance variables declared as such, next to the other instance variables. Making them static will overwrite these values when another instance runs and that is not what you want.

Sigma1912: Done, thank you.

PR #4250, opened 2026-07-19
2.9 Fix: Replace deprecated 'locale.format()'

Sigma1912:
Replace deprecated 'locale.format()' with 'locale.format_string()'

Fixes this issue (but now also fixed in all other places I found) :
https://github.com/LinuxCNC/linuxcnc/issues/4254

### DIFF
--- a/lib/python/gladevcp/calculatorwidget.py
+++ b/lib/python/gladevcp/calculatorwidget.py
@@ -299,11 +299,11 @@ class Calculator( Gtk.Box ):
         self.displayOperand( "+" )
 
     def displayMmInch( self, widget ):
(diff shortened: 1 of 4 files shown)


Sigma1912:
@hansu, @andypugh 
Not exactly sure why it fails on sid but it seems unrelated to the proposed changes.
Could somebody please restart the tests?

grandixximo:
I did re run already, still not sticking, i will try again, lets see how attempt 3 goes, I think it might be something with 2.9

grandixximo:
@Sigma1912 I have a fix in #4255 once that builds ok, I'll merge, and you can rebase, so all shall be green, hopefully.

Sigma1912:
Ok, thanks for the quick fix.

grandixximo:
Merged, ready for rebase

PR #4251, opened 2026-07-19
hal: C++ API (hal.hh) and pybind11 bindings on the new HAL API

grandixximo:
@rene-dev @BsAtHome - this is the unification pass we discussed in the #4099 meeting: Rene's C++/pybind11 interface rebuilt on Bertho's getter/setter API, satisfying both constraints (no HAL internals access; simple, low-maintenance, self-documenting C++ surface).

Based on master, fully additive: `_hal`/`hal.py` are untouched and nothing is replaced. The by-name query functions and HAL_PORT creation are compiled out when the query API header is absent and activate automatically when #4247 lands.

## What this adds

- `src/hal/hal.hh` - type-safe C++ layer, header-only, in `namespace linuxcnc::hal`:
  - `hal::pin<T>` typed handles via `traits<T>` over the rtapi_ types: `rtapi_s32`/`rtapi_sint` are distinct C++ types, so 32/64 accessor selection happens at compile time with zero cost. Handles re-read the shmem slot per access, so `hal_link()` rewrites stay visible (same semantics as C pin pointers). Handles are move-only; HAL objects are unique.
  - `hal::pin_t` variant + `hal::anypin`: runtime-typed access multiplexes on the variant tag (the stored-type pattern required for name-keyed collections), with the tag supplied by the query API.
  - `hal::component`: pins + params, `newpin`/`newparam` (typed and runtime-typed), `add_pin` for the struct-member idiom, item access, prefix handling.
  - ULAPI by-name functions (`get_value`, `set_value`, `set_signal`, `component_exists`, `pin_has_writer`, signal management) implemented entirely on `hal_get_p`/`hal_set_p`/`hal_get_s`/`hal_set_s`/`hal_comp_by_name`. No `hal_priv.h`, no direct shmem access, no re-implemented parsers.
- `src/hal/halpybind.cc` - pybind11 module (`halpp.so`), adapted from @rene-dev's bindings, extended with params and error paths. Built *alongside* `_hal`/`hal.py`; module replacement is a separate (3.0) discussion.
- `src/emc/task/taskclass.cc` - converted to the new API (it was the only hal.hh consumer).
- `tests/halpp/` - Python and native C++ smoke suites (both green), plus `python3-pybind11` build dep.

## Safety properties worth noting

- Query callbacks are strictly exception-free: conversion errors are reported by flag and thrown after the library releases the HAL mutex. (Throwing through a callback wedges the session-wide mutex; found the hard way, and tested.)
- Set-side coercion uses the *item's* type with range checks; an overflow raises instead of truncating.
- Typed access compiles to the same two-load sequence as the C API and the old pointer dereference, so there is no hot-path cost.

## Relationship to the original pybind11 branch

Kept: the variant/map/PyPin structure and the binding shape.
Dropped: the private `set_common` string parser (superseded by `hal_set_p`), `hal_mutex_guard`, the `get_info_signals` stub, `waitWritable` (was non-functional).
Compatibility aliases (`hal_comp`, `hal_pin<T>`, `hal_dir`, `PyPin`) are provided so existing consumers of the old hal.hh keep working.

## Test evidence

Full tree builds; `tests/halmodule.0/.1` pass; new smoke suites cover component lifecycle, typed pins/params (incl. negative S32 and 64-bit values), signals, linking, by-name get/set with coercion, port pins, and error paths. The suites auto-skip query-API checks when built without the query API, so they run on both master and the #4247 tree.

### DIFF
--- a/lib/python/hal.py
+++ b/lib/python/hal.py
@@ -29,8 +29,15 @@ KeyboardInterrupt exception will be raised.
 
 import _hal
 from _hal import *
+from _hal import query
+import sys
 import warnings
 import lcnc_realtime
+
+# _hal.query is a submodule, not a plain attribute. Registering it in
+# sys.modules under its dotted name makes 'import hal.query' and
+# 'from hal import query' work as they would for a real package.
+sys.modules['hal.query'] = query
 
 def __getattr__(name):
     if name == 'is_rt':
--- a/lib/python/hal.py
+++ b/lib/python/hal.py
@@ -34,6 +34,10 @@ import sys
 import warnings
 import lcnc_realtime
 
+# _hal exports the IntEnum tagging classes as 'Type' and 'Dir', bound
+# here by the star import: hal.Type.REAL, hal.Dir.IN, and pickle's
+# lookup of the class names all resolve to the shared classes in _hal.
+
 # _hal.query is a submodule, not a plain attribute. Registering it in
 # sys.modules under its dotted name makes 'import hal.query' and
 # 'from hal import query' work as they would for a real package.
--- a/src/hal/hal.hh
+++ b/src/hal/hal.hh
@@ -1,166 +1,605 @@
+/*
+    hal.hh - C++ interface for HAL
+
+    A thin, type-safe C++ layer on top of the public HAL C API.
+    All pin/param access goes through the typed hal_get_X and hal_set_X
+    accessors and the user-land query API. No direct shared memory
+    access, no hal_priv.h, no re-implemented library internals.
+
+    Header-only, no runtime overhead: typed access expands to the same
+    inline accessor calls as the C API.
+
+    The by-name query/set section requires the HAL query API and is
+    compiled only when the public header defines the COMPONENT_TYPE_*
+    macros. The typed component/pin layer only needs the base
+    getter/setter API.
+
+    HAL_PORT pins are intentionally not supported yet. The port handle
+    type and creation semantics are still in flux until the API break
+    ("the port change must be done later" in hal.h); a proper port
+    wrapper follows once hal_port_t and hal_pin_new_port() are final.
+*/
 #ifndef HALXX_HH
 #define HALXX_HH
 
 #include <map>
 #include <string>
+#include <vector>
 #include <variant>
-#include "hal.h"
+#include <stdexcept>
+#include <cstdint>
+#include <cstring>
+#include <cerrno>
 
-enum class hal_dir{
-    IN = HAL_IN,
+#include <hal.h>
+
+// The HAL query API (hal_get_p, hal_set_p, hal_comp_by_name, ...) is
+// present when the public header defines the COMPONENT_TYPE_* macros.
+#if defined(COMPONENT_TYPE_REALTIME) && !defined(HALXX_WITH_QUERY_API)
+#define HALXX_WITH_QUERY_API 1
+#endif
+
+#ifndef RTAPI_SINT_MAX
+#define RTAPI_SINT_MAX RTAPI_INT64_MAX
+#define RTAPI_SINT_MIN RTAPI_INT64_MIN
+#define RTAPI_UINT_MAX RTAPI_UINT64_MAX
+#endif
+
+namespace linuxcnc {
+namespace hal {
+
+// Unified pin/param direction. Values are identical to hal_pdir_t.
+enum class dir : int {
+    IN  = HAL_IN,
     OUT = HAL_OUT,
+    IO  = HAL_IO,
+    RO  = HAL_RO,
+    RW  = HAL_RW,
 };
 
-#if 0
-// If this class is ever necessary, then it needs to be moved into a new
-// header 'hal_priv.hh' because it uses internal access methods from
-// 'hal_priv.h' that should not be available to the casual source file.
+// Runtime value of a pin, param or signal. Used whenever the HAL type
+// is not known at compile time (name-based access, script bindings).
+using value_t = std::variant<rtapi_bool, rtapi_s32, rtapi_u32,
+                             rtapi_sint, rtapi_uint, rtapi_real>;
 
-#include "hal_priv.h"
-
-class hal{
-    public:
-    static bool component_exists(const std::string& name){
-        return halpr_find_comp_by_name(name.c_str()) != NULL;
-    }
-    static bool pin_has_writer(const std::string& name){
-        hal_pin_t *pin = halpr_find_pin_by_name(name.c_str());
-        if(!pin) {//pin does not exist
-            return false;
-        }
-        if(pin->signal) {
-            hal_sig_t *signal = (hal_sig_t*)SHMPTR(pin->signal);
-            return signal->writers > 0;
-        }
-        return false;
-    }
-    static bool component_is_ready(const std::string& name){
-        // Bad form to assume comp name exists - stop crashing!
-        hal_comp_t *thecomp = halpr_find_comp_by_name(name.c_str());
-        return thecomp && (thecomp->ready != 0);
-    }
-};
+namespace detail {
+// Error text that works with and without hal_strerror() in the library.
+inline std::string errstr(int rv)
+{
+#ifdef HALXX_WITH_QUERY_API
+    return hal_strerror(rv);
+#else
+    return std::strerror(-rv);
 #endif
+}
+} // namespace detail
 
+//----------------------------------------------------------------------
+// Type traits: map an rtapi_ value type to its HAL handle, HAL type and
+// accessor/creator functions. Using an unsupported type is a compile
+// error because traits<T> is intentionally left undefined.
+//
+// The 32-bit versions (rtapi_s32, rtapi_u32) are compatibility types
+// that will be retired after the API break; only their traits entries
+// and the variant alternatives need to change then.
+//----------------------------------------------------------------------
+template<typename T> struct traits;
+
+template<> struct traits<rtapi_bool> {
+    using handle_t = hal_bool_t;
+    static handle_t *slot(hal_refs_u *u) { return &u->b; }
+    static constexpr hal_type_t type = HAL_BOOL;
+    static rtapi_bool get(handle_t h) { return hal_get_bool(h); }
+    static rtapi_bool set(handle_t h, rtapi_bool v) { return hal_set_bool(h, v); }
+    static int new_pin(int c, hal_pdir_t d, handle_t *h, rtapi_bool def, const std::string &n) {
+        return hal_pin_new_bool(c, d, h, def, "%s", n.c_str());
+    }
+    static int new_param(int c, hal_pdir_t d, handle_t *h, rtapi_bool def, const std::string &n) {
+        return hal_param_new_bool(c, d, h, def, "%s", n.c_str());
+    }
+};
+
+template<> struct traits<rtapi_s32> {
+    using handle_t = hal_sint_t;
+    static handle_t *slot(hal_refs_u *u) { return &u->s; }
+    static constexpr hal_type_t type = HAL_S32;
+    static rtapi_s32 get(handle_t h) { return hal_get_si32(h); }
+    static rtapi_s32 set(handle_t h, rtapi_s32 v) { return hal_set_si32(h, v); }
+    static int new_pin(int c, hal_pdir_t d, handle_t *h, rtapi_s32 def, const std::string &n) {
+        return hal_pin_new_si32(c, d, h, def, "%s", n.c_str());
+    }
+    static int new_param(int c, hal_pdir_t d, handle_t *h, rtapi_s32 def, const std::string &n) {
+        return hal_param_new_si32(c, d, h, def, "%s", n.c_str());
+    }
+};
+
+template<> struct traits<rtapi_u32> {
+    using handle_t = hal_uint_t;
+    static handle_t *slot(hal_refs_u *u) { return &u->u; }
+    static constexpr hal_type_t type = HAL_U32;
+    static rtapi_u32 get(handle_t h) { return hal_get_ui32(h); }
+    static rtapi_u32 set(handle_t h, rtapi_u32 v) { return hal_set_ui32(h, v); }
+    static int new_pin(int c, hal_pdir_t d, handle_t *h, rtapi_u32 def, const std::string &n) {
+        return hal_pin_new_ui32(c, d, h, def, "%s", n.c_str());
+    }
+    static int new_param(int c, hal_pdir_t d, handle_t *h, rtapi_u32 def, const std::string &n) {
+        return hal_param_new_ui32(c, d, h, def, "%s", n.c_str());
+    }
+};
+
+template<> struct traits<rtapi_sint> {
+    using handle_t = hal_sint_t;
+    static handle_t *slot(hal_refs_u *u) { return &u->s; }
+    static constexpr hal_type_t type = HAL_SINT;
+    static rtapi_sint get(handle_t h) { return hal_get_sint(h); }
+    static rtapi_sint set(handle_t h, rtapi_sint v) { return hal_set_sint(h, v); }
+    static int new_pin(int c, hal_pdir_t d, handle_t *h, rtapi_sint def, const std::string &n) {
+        return hal_pin_new_sint(c, d, h, def, "%s", n.c_str());
+    }
+    static int new_param(int c, hal_pdir_t d, handle_t *h, rtapi_sint def, const std::string &n) {
+        return hal_param_new_sint(c, d, h, def, "%s", n.c_str());
+    }
+};
+
+template<> struct traits<rtapi_uint> {
+    using handle_t = hal_uint_t;
+    static handle_t *slot(hal_refs_u *u) { return &u->u; }
+    static constexpr hal_type_t type = HAL_UINT;
+    static rtapi_uint get(handle_t h) { return hal_get_uint(h); }
+    static rtapi_uint set(handle_t h, rtapi_uint v) { return hal_set_uint(h, v); }
+    static int new_pin(int c, hal_pdir_t d, handle_t *h, rtapi_uint def, const std::string &n) {
+        return hal_pin_new_uint(c, d, h, def, "%s", n.c_str());
+    }
+    static int new_param(int c, hal_pdir_t d, handle_t *h, rtapi_uint def, const std::string &n) {
+        return hal_param_new_uint(c, d, h, def, "%s", n.c_str());
+    }
+};
+
+template<> struct traits<rtapi_real> {
+    using handle_t = hal_real_t;
+    static handle_t *slot(hal_refs_u *u) { return &u->r; }
+    static constexpr hal_type_t type = HAL_REAL;
+    static rtapi_real get(handle_t h) { return hal_get_real(h); }
+    static rtapi_real set(handle_t h, rtapi_real v) { return hal_set_real(h, v); }
+    static int new_pin(int c, hal_pdir_t d, handle_t *h, rtapi_real def, const std::string &n) {
+        return hal_pin_new_real(c, d, h, def, "%s", n.c_str());
+    }
+    static int new_param(int c, hal_pdir_t d, handle_t *h, rtapi_real def, const std::string &n) {
+        return hal_param_new_real(c, d, h, def, "%s", n.c_str());
+    }
+};
+
+//----------------------------------------------------------------------
+// pin<T> - typed pin or param handle.
+//
+// Holds a pointer to the handle slot in HAL shared memory and re-reads
+// it on every access: hal_link() may rewrite the slot when the pin is
+// linked to a signal, exactly like a pin pointer variable in the C API.
+// All access goes through the type's inline hal_get_*/hal_set_*
+// accessor.
+//
+// Pins, params and signals are unique HAL objects. Their handles are
+// not copyable (no reference counting); use references or move
+// semantics. dup() creates an explicit second handle to the same slot
+// where that is really intended.
+//----------------------------------------------------------------------
 template<typename T>
-class hal_pin{
-    public:
-    volatile T** ptr;
-    T operator=(const T& value){
-        **ptr = value;
-        return **ptr;
+class pin {
+public:
+    using value_type = T;
+    using handle_t = typename traits<T>::handle_t;
+
+    pin() = default;
+    explicit pin(handle_t *slot) : slot_(slot) {}
+    pin(const pin &) = delete;
+    pin &operator=(const pin &) = delete;
+    pin(pin &&) = default;
+    pin &operator=(pin &&) = default;
+
+    // Explicit second handle to the same HAL object.
+    pin dup() const { return pin(slot_); }
+
+    T get() const { check(); return traits<T>::get(*slot_); }
+    T set(T v) const { check(); return traits<T>::set(*slot_, v); }
+
+    operator T() const { return get(); }
+    T operator=(T v) { return set(v); }
+
+    handle_t handle() const { check(); return *slot_; }
+    bool valid() const { return nullptr != slot_ && nullptr != *slot_; }
+
+private:
+    void check() const {
+        if(!slot_)
+            throw std::logic_error("hal::pin: access to uninitialized pin handle");
     }
-    operator T(){
-        return **ptr;
-    }
+    handle_t *slot_ = nullptr;
 };
 
-using pin_t = std::variant<hal_pin<double>,hal_pin<bool>,hal_pin<int32_t>,hal_pin<uint32_t>>;
+//----------------------------------------------------------------------
+// pin_t - runtime-typed pin/param/ports. The variant index is the
+// stored type tag used for multiplexing, as required for any
+// heterogeneous (name-keyed) collection of HAL items.
+//----------------------------------------------------------------------
+using pin_t = std::variant<pin<rtapi_bool>, pin<rtapi_s32>, pin<rtapi_u32>,
+                           pin<rtapi_sint>, pin<rtapi_uint>, pin<rtapi_real>>;
 
-class hal_comp{
-    int comp_id;
-    std::string comp_name;
-    std::map<std::string,pin_t> map;
-    int add_pin_(const std::string& name, hal_dir dir, hal_pin<bool> pin){
-        return hal_pin_new(name.c_str(), HAL_BIT, static_cast<hal_pin_dir_t>(dir), (void **)(pin.ptr), comp_id);
-    }
-    int add_pin_(const std::string& name, hal_dir dir, hal_pin<int32_t> pin){
-        return hal_pin_new(name.c_str(), HAL_S32, static_cast<hal_pin_dir_t>(dir), (void **)(pin.ptr), comp_id);
-    }
-    int add_pin_(const std::string& name, hal_dir dir, hal_pin<uint32_t> pin){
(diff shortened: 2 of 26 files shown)


src/hal/hal.hh
```diff
@@ -0,0 +1,601 @@
...
+#include <map>
+#include <string>
+#include <variant>
+#include <stdexcept>
+#include <exception>
+#include <cstdint>
+#include <cstring>
+#include <cerrno>
+
+#include "hal.h"
```
BsAtHome: Should not include the local version, but the global (`#include <hal.h>`).

src/hal/hal.hh
```diff
@@ -0,0 +1,601 @@
...
+    IN  = HAL_IN,
+    OUT = HAL_OUT,
+    IO  = HAL_IO,
+    RO  = HAL_RO,
+    RW  = HAL_RW,
+};
+
+// Runtime value of a pin, param or signal. Used whenever the HAL type
+// is not known at compile time (name-based access, script bindings).
+using value_t = std::variant<bool, int32_t, uint32_t, int64_t, uint64_t, double>;
```
BsAtHome: The underlying types are called:
* rtapi_bool
* rtapi_s32
* rtapi_u32
* rtapi_sint
* rtapi_uint
* rtapi_real

Where the 32-bit versions are to be retired eventually and are only there for compatibility.

There are multiple places where the types are addressed wrongly. The reason for this abstraction is to prevent us from going through the sources again when something changes.

src/hal/hal.hh
```diff
@@ -0,0 +1,601 @@
...
+        return hal_pin_new_real(c, d, h, def, "%s", n.c_str());
+    }
+    static int new_param(int c, hal_pdir_t d, handle_t *h, double def, const std::string &n) {
+        return hal_param_new_real(c, d, h, def, "%s", n.c_str());
+    }
+};
+
+//----------------------------------------------------------------------
+// pin<T> - typed pin or param handle.
+// Copyable, near zero overhead. Holds a pointer to the handle slot in
```
BsAtHome: Pins, params and signals are not copyable or you would have to start reference counting. Copy constructor and assignment operator should be deleted.

src/hal/hal.hh
```diff
@@ -0,0 +1,601 @@
...
+    const hal_port_t *p() const { check(); return reinterpret_cast<const hal_port_t*>(*slot_); }
+    hal_sint_t *slot_ = nullptr;
+};
+
+//----------------------------------------------------------------------
+// pin_t - runtime-typed pin/param/ports. The variant index is the
+// stored type tag used for multiplexing, as required for any
+// heterogeneous (name-keyed) collection of HAL items.
+//----------------------------------------------------------------------
+using pin_t = std::variant<pin<bool>, pin<int32_t>, pin<uint32_t>,
```
BsAtHome: A port is not a normal pin. There must be one in and one out port pin and they connect to a signal of type port. You cannot assign to a port pin and cannot read from it.

That said, reading a port pin _will_ function and returns the queue reference.

src/hal/hal.hh
```diff
@@ -0,0 +1,601 @@
...
+        return {};
+    }
+    std::string peek(unsigned count) {
+        std::string buf(count, '\0');
+        if(count && hal_port_peek(p(), buf.data(), count))
+            return buf;
+        return {};
+    }
+
+    hal_sint_t handle() const { check(); return *slot_; }
```
BsAtHome: The handle of a port is hal_port_t.

However, in the transition period we cannot change the current version (hal_port_t*) into the opaque version hal_port_t. The two conflict.

grandixximo:
This seems to be working, but I am not sure if it is the right shape we want, these are the direction I took on the issue raised in the meeting, subject to revision.

| Problem | Solution |
|---|---|
| Handle is width-blind | Type lives in `traits<T>` (compile time) or the variant tag (runtime, from `q->pp.type`) |
| `hal_link` rewrites the handle slot | `pin<T>` stores a *pointer to the slot* and re-reads per access (C semantics, staleness impossible) |
| Throwing through query callbacks wedges the HAL mutex session-wide | Exceptions captured in the callback, rethrown after the library releases the mutex |
| Handle storage must be shmem | `hal_malloc`'d `hal_refs_u` slot per item (halmodule's own pattern) |


@BsAtHome A bare hal_sint_t is width-blind, and wrapping foreign raw handles across widths is unsafe in the interim. But the C++ layer never selects from the handle alone. Component-created pins carry the type at construction via traits<T> (int32_t vs int64_t are distinct C++ types, compile-time selection), and the runtime path uses q->pp.type from your query API to pick the variant alternative, the stored-type-tag multiplex you described in #4099. The only unsafe path is explicitly reinterpreting a raw handle, which can be guarded with a query lookup. So the layer works transitionally, it just gets simpler after the break.

src/hal/hal.hh
```diff
@@ -0,0 +1,601 @@
...
+    } catch(...) {
+        req->eptr = std::current_exception();
+        return -ERANGE;
+    }
+}
+inline int coerce_sig_cb(hal_query_t *q, void *arg)
+{
+    auto *req = static_cast<coerce_req *>(arg);
+    try {
+        q->sig.value = convert_value(q->sig.type, *req->v);
```
BsAtHome: This must not throw.

Exceptions in a callback while holding the HAL lock are not a good thing.

src/hal/hal.hh
```diff
@@ -0,0 +1,601 @@
...
+#include <variant>
+#include <stdexcept>
+#include <exception>
+#include <cstdint>
+#include <cstring>
+#include <cerrno>
+
+#include "hal.h"
+
+namespace hal {
```
BsAtHome, line 51: I think we should have the namespace under linuxcnc (see IniFile)

src/hal/hal.hh
```diff
@@ -0,0 +1,601 @@
...
+    }
+
+    // Create a typed pin "<prefix>.<name>" and keep it in the item map.
+    // The handle slot is allocated from HAL shared memory (hal_malloc),
+    // as required by the pin/param creation API: hal_link later updates
+    // the value through this slot, so it must live in HAL memory. Like
+    // halmodule, the slot is released with the component's HAL memory.
+    template<typename T>
+    pin<T> newpin(const std::string &name, dir d, T def = T{}) {
+        hal_refs_u *u = (hal_refs_u *)hal_malloc(sizeof(hal_refs_u));
```
BsAtHome: You should allocate: `hal_malloc(sizeof(*u))`.

rene-dev:
oh, that was quick. I think this should be separate from the other big PR, as this can go in master fairly quickly. the only reason I stopped was that buster had no pybind11 package. this should also replace the current pyhal/halmodule instead of adding a third module. and, the same cpp api should be used in xhc, task and other places which are implemented in cpp.

grandixximo:
Buster is basically not supported anymore with master, I will fix the review points tomorrow, thank you both ☺️
My bad on missing the include, getting late time for bed...

src/hal/hal.hh
grandixximo: fixed

src/hal/hal.hh
grandixximo: all renamed

src/hal/hal.hh
grandixximo: Port documented as FIFO endpoint pair, no scalar access

src/hal/hal.hh
grandixximo: Transitional hal_port_t* vs opaque hal_port_t conflict documented in the port class

src/hal/hal.hh
grandixximo: pin/port/anypin move-only, no implicit copies (dup() for explicit re-handle)

src/hal/hal.hh
grandixximo: Callback path is now strictly exception-free: convert_value returns a flag, callers throw after the library releases the mutex

src/hal/hal.hh
grandixximo: namespace linuxcnc::hal (IniFile convention), compat aliases kept

src/hal/hal.hh
grandixximo: Done

src/hal/hal.hh
BsAtHome: Conversion from text to value has been implementet in `setps_common_cb()` (see hal/util/setps_util.[ch]). It provides a consistent interface to conversion from test to value for all types (especially boolean and decimal point for floating point).

BsAtHome:
There is another issue. The new hal_lib uses a reference counted init/exit and has separated out `hal_lib_init()` and `hal_lib_exit()`. The reason to take this out is that you do not need a component for pin/param/signal read/write or query interactions. You only need to have mapped shared memory.

Calling `hal_init()` to create a component (if you want to create pins) now calls the factored out hal_lib_init() and `hal_exit()`calls the factored out hal_lib_exit().

The python class must initialize using hal_lib_init() and terminate by calling hal_lib_exit(). If the user has not called hal_exit() on any created components, you will now get a proper error message (because the user forgot to terminate correctly).

rene-dev:
can you rename the module to halmodule, so that the UIs and the tests that are already in place use the new bindings, and remove the old ones?

src/hal/hal.hh
```diff
@@ -1,166 +1,686 @@
...
+        if(count && hal_port_read(p(), buf.data(), count))
+            return buf;
+        return {};
+    }
+    std::string peek(unsigned count) {
+        std::string buf(count, '\0');
+        if(count && hal_port_peek(p(), buf.data(), count))
+            return buf;
+        return {};
+    }
```
BsAtHome: Ports do not process strings. They process binary data. The proper type would be then be std::vector<char> and not std::string IMO.

src/hal/hal.hh
```diff
@@ -1,166 +1,686 @@
...
+// scalar value.
+inline value_t pin_get(const pin_t &p)
+{
+    return std::visit([](auto &&pp) -> value_t {
+        using P = std::decay_t<decltype(pp)>;
+        if constexpr (std::is_same_v<P, port>)
+            throw std::invalid_argument("hal: port pin has no scalar value");
+        else
+            return pp.get();
+    }, p);
```
BsAtHome: Actually, it does have a scalar value...
Whether it is useful, that is a different question.

src/hal/hal.hh
```diff
@@ -1,166 +1,686 @@
...
+}
+
+inline void pin_set(pin_t &p, const value_t &v)
+{
+    std::visit([&v](auto &&pp) {
+        using P = std::decay_t<decltype(pp)>;
+        if constexpr (std::is_same_v<P, port>)
+            throw std::invalid_argument("hal: port pin has no scalar value");
+        else
```
BsAtHome: Port pins do have a scalar value, but you cannot write to them.

src/hal/hal.hh
```diff
@@ -1,166 +1,686 @@
...
+            hal_refs_u *u = (hal_refs_u *)hal_malloc(sizeof(*u));
+            if(!u)
+                throw std::runtime_error("hal::component: newpin(" + name + "): hal_malloc failed");
+            int rv = hal_pin_new_port(id_, (hal_pin_dir_t)d, &u->s, "%s", fullname(name).c_str());
+            if(rv)
+                throw std::runtime_error("hal::component: newpin(" + name + ") failed: " + detail::errstr(rv));
+            items_.emplace(name, port(&u->s));
+            return anypin(port(&u->s), fullname(name));
+        }
+#endif
```
BsAtHome: Ports are created with hal_pin_new_port().

This is not part of the query API but core hal lib. Just as you like it until the API break.

src/hal/hal.hh
```diff
@@ -1,166 +1,686 @@
...
+    int id_ = -1;
+    std::string prefix_;
+    std::map<std::string, pin_t> items_;
+    std::map<std::string, pin_t> params_;
+};
+
+//----------------------------------------------------------------------
+// Signal management, thin wrappers over the C API (user-land only).
+//----------------------------------------------------------------------
+#ifdef ULAPI
```
BsAtHome, line 640: By definition of where this code lives, user-space only, the ULAPI define must be defined. It cannot be used in RTAPI code.

src/hal/hal.hh
```diff
@@ -1,166 +1,686 @@
...
+// mutex is held, and unwinding through the library would keep the
+// mutex locked and wedge the whole HAL session. Returns false on a
+// range/type error, the caller reports it after the library call.
+inline bool convert_value(hal_type_t target, const value_t &v, hal_query_value_u *out)
+{
+    bool ok = true;
+    std::visit([&ok, out, target](auto &&x) {
+        long double xv = static_cast<long double>(x);
+        switch(target) {
+        case HAL_BIT:
```
BsAtHome: HAL_BIT is not called HAL_BOOL. HAL_FLOAT is now called HAL_REAL. These should be fixed everywhere.

src/hal/hal.hh
```diff
@@ -1,166 +1,686 @@
...
+
+// True if the pin exists, is connected to a signal, and that signal
+// has at least one writer.
+inline bool pin_has_writer(const std::string &name)
+{
+    hal_query_t q = {};
+    q.name = name.c_str();
+    q.qtype = HAL_QTYPE_PIN;
+    if(0 != hal_get_p(&q, nullptr, nullptr) || !q.pp.signal)
+        return false;
```
BsAtHome, line 768: Should use `hal_getref_p()`

src/hal/hal.hh
```diff
@@ -1,166 +1,686 @@
...
+    hal_query_t sq = {};
+    sq.name = q.pp.signal;
+    if(0 != hal_get_s(&sq, nullptr, nullptr))
+        return false;
```
BsAtHome, line 772: Should use `hal_getref_s()`

src/hal/hal.hh
```diff
@@ -1,166 +1,686 @@
...
+// std::invalid_argument if the name does not exist.
+inline value_t get_value(const std::string &name)
+{
+    hal_query_t q = {};
+    q.name = name.c_str();
+    if(0 == hal_get_p(&q, nullptr, nullptr))
+        return detail::value_from_query(q.pp.type, q.pp.value);
+    if(0 == hal_get_s(&q, nullptr, nullptr))
+        return detail::value_from_query(q.sig.type, q.sig.value);
+    throw std::invalid_argument("hal: pin, param or signal '" + name + "' not found");
```
BsAtHome: The return value of hal_get_X indicate more than just "invalid value".

grandixximo:
> can you rename the module to halmodule, so that the UIs and the tests that are already in place use the new bindings, and remove the old ones?

Replacing halmodule is the goal, agreed, but renaming now would break things for two separate reasons.

First, a hard technical one: on master the new bindings cannot cover the existing Python surface. `get_info_pins/signals/params`, `pin_has_writer` and `component_exists` are used by qtvcp (`qt_halobjects`, `hal_selectionbox`) and hal_glib, and they need the query API (`hal_list_*`, `hal_get_p`) from #4247. Without it, the options are a strict subset module (GUIs break immediately) or private shmem access (exactly what this API change abolishes).

Second, the compatibility tail: `hal.error` exception type, KeyboardInterrupt on `halcmd unload` (documented in hal.py's docstring, user comps wrap their main loop in it), the exact `get_info_*` dict shapes (tests/halmodule.0 compares verbatim), the `hal.stream` object API (tests/halmodule.1), setitem coercion quirks, pin `.dir`/`.type` attributes, and the `hal.py` wrapper shims. Matching all of that is what makes the swap a user-visible event, and it deserves its own PR with a full parity pass, not a rename here.

So the proposal stands as: land this additive now (hal.hh + halpp alongside, nothing replaced), and do the rename + replacement as a follow-up once #4247 is in. Your old branch's consumer sweep (hal_glib, qtvcp core, raster, tests) is the starting checklist for that PR. Is this acceptable? Open for discussion...

src/hal/hal.hh
grandixximo: Ports are binary: `std::vector<char>` in the API, `bytes` in the bindings

src/hal/hal.hh
grandixximo: Port pins *do* have a scalar value: reads yield the queue reference (via `hal_get_sint`, since `hal_get_port` is `#if 0`'d in **both** trees, noted in a comment, one-line flip later), writes rejected

src/hal/hal.hh
grandixximo: Port creation comment corrected (core hal lib, not query API)

src/hal/hal.hh
grandixximo: addressed in e3472f0, gate stays until the break per your note

src/hal/hal.hh
grandixximo: By-name section documented user-space-only by definition

src/hal/hal.hh
BsAtHome: Yeah, but you cannot use the hal_pin_new_port() function yet. The argument `&u->s` is wrong.

I think the whole HAL_PORT stuff needs to be exempted until we do the API break. There are a lot of small details in it that make this a very hard one.

src/hal/hal.hh
grandixximo: `HAL_BOOL`/`HAL_REAL` names used throughout the C++ layer (Python enum gained them as aliases, `HAL_BIT`/`HAL_FLOAT` kept for compat)

src/hal/hal.hh
grandixximo: Changed

src/hal/hal.hh
grandixximo: Changed

src/hal/hal.hh
grandixximo: `get_value` propagates the real library error

src/hal/hal.hh
grandixximo: Agreed, exempted in 202386a1f. The port class, variant alternative and bindings are removed; hal.hh documents that HAL_PORT returns after the break as a proper hal_port_t wrapper, once the handle type and hal_pin_new_port() are final. You were right that &u->s only matched the transitional signature mechanically, not the settled contract.

rene-dev:
I see that https://github.com/LinuxCNC/linuxcnc/pull/4231 changes halmodule.cc, we should also add those functions via c++/pybind11

BsAtHome:
> I see that #4231 changes halmodule.cc, we should also add those functions via c++/pybind11

No, all of this should be in halquery.

rene-dev:
> > I see that #4231 changes halmodule.cc, we should also add those functions via c++/pybind11
> 
> 
> 
> No, all of this should be in halquery.

What is halquery?

BsAtHome:
> > > I see that #4231 changes halmodule.cc, we should also add those functions via c++/pybind11
> > 
> > 
> > No, all of this should be in halquery.
> 
> What is halquery?

That is the new interface to all of HAL's internals nicely wrapped up from the hal query interface without someone needing to dig into HAL's internals. Hal_lib's internals are fully isolated in #4247 and nobody should be writing code that circumvents that. Inclusion of hal_priv.h is strictly forbidden in any code that is not part of hal_lib.
Halmodule deals with components and everything else was "bolted on". You do not want one big blob of code to do it all. You want modularization, which halquery provides.

rene-dev:
ah, found it. I would much prefer if it would be a c++ api, which can be used to automatically generate the python bindings.
most of userspace code is c++ anyway. I have maintained the c bindings in the past(pyhton2->3 migration), and do not ever want to deal with them again.

BsAtHome:
It is a pure C version, so it does not have the same problems as C++ interfacing C-based Python ;-)

grandixximo:
This is exactly the split #4251 (this PR) implements: the query API stays pure C in the library (Bertho's layer, no bypass possible), and the C++ API sits on top of it (hal.hh), which is where Python bindings get generated (halpybind) , no hand-written CPython code. halquery.c proves the query API is sufficient; the same functions can be exposed as pybind wrappers over `hal_get_p/hal_list_* `once #4247 lands. So the endgame can be: C query API in hal_lib, C++ API above it, Python bindings generated from the C++ layer, and the hand-written C modules retired on the same schedule. Happy to demonstrate the halquery-equivalent functions in halpybind as a follow-up once the base is in.

BsAtHome:
Fine by me to get it C++ wrapped. It does not yet have priority one.

There is one thing that should be fixed in both halmodule and halquery: use IntEnum types for both type and direction. It may be necessary to have a `haltype` module with the enums so you it can be properly shared between the other modules. Not sure how this can be handled properly in using consistent namespace names and hierarchy.

grandixximo:
A shared pure-Python `haltype.py` with `HALType(IntEnum)` and `HALDir(IntEnum)` is the cleanest shape. IntEnum members are ints, so the C boundary stays plain int and everything interops both ways; `hal`, `halquery` and the pybind module import and re-export from it, keeping `hal.HAL_FLOAT` working unchanged. I would keep the namespace flat (sibling of hal.py) rather than a linuxcnc.* package, which breaks every existing `import hal`. One pybind11 note: `py::enum_` does not emit real IntEnum objects, that needs `py::native_enum` (pybind11 >= 3.0, trixie ships 2.13), so for now the bindings take constants from `haltype.py` and we flip to `native_enum` when the distro catches up.

BsAtHome:
> A shared pure-Python `haltype.py` with `HALType(IntEnum)` and `HALDir(IntEnum)` is the cleanest shape. IntEnum members are ints, so the C boundary stays plain int and everything interops both ways; `hal`, `halquery` and the pybind module import and re-export from it, keeping `hal.HAL_FLOAT` working unchanged.

That would be the simple way. If it works, then it should be nice. The real benefit is in halquery to have the iteration result show the right name instead of a number. There it is big value. Probably also need to add the same for other constants.

> I would keep the namespace flat (sibling of hal.py) rather than a linuxcnc.* package, which breaks every existing `import hal`.

Global namespace naming and fixing hierarchy is a problem we need to fix in a later version. This would be a version 3.0 thing, I guess ;-)

> One pybind11 note: `py::enum_` does not emit real IntEnum objects, that needs `py::native_enum` (pybind11 >= 3.0, trixie ships 2.13), so for now the bindings take constants from haltype.py and we flip to native_enum when the distro catches up.

Ha, and so much for "C++ is easy to maintain" ;-)
The version and support problem <i>always</i> pops up _regardless_ of language and library. You always end in hell once you need to support over long periods of time and different distros and distro versions. At least it is warm in hell.

grandixximo:
Going back to draft, waiting for master new API to materialize ;-)

grandixximo:
Branch rewritten and pushed (33012dbd78 => 63c8a05c33), four commits on current master:

- hal.hh replaces the deprecated upstream header with the new C++ API (typed pin/param handles, runtime-typed anypin, component, streams), built strictly on the public C API and the query API
- haltype.py: stdlib-only HALType/HALDir IntEnums plus tests/haltype.0 as drift guard against the _hal constants
- streams: linuxcnc::hal::stream and halpp.stream with _hal stream semantics; tests/halpp now runs under runtests
- hal.query registration prototype: empty _hal.query submodule, hal.py publishes it in sys.modules, tests/halquery.0 checks the four import paths

Build notes: the pybind11 bindings are skipped with a note when the headers are absent, not failed, so the tree still builds without python3-pybind11; tests/halpp self-skips in that case. hal.hh is exported to include/ alongside hal.h. tests/halpp/test.sh compiles cpp_test.cc against the tree and runs it in a live HAL session, so the native C++ side has CI coverage too.

Still draft on purpose: #4247 moves the same ground (hal.hh, the type system, HAL isolation), so this waits for that to land before the query bindings follow as hal.query.

grandixximo:
Branch reworked after feedback Bertho sent me in a private mail (63c8a05c33 => 7beaaf3c9b). The haltype.py commit is gone, replaced by:

- src/hal/halenum.hh builds the two IntEnum classes from the hal.h constants through the plain Python C API: one native source of truth, nothing to drift, and any extension module can instantiate the same classes, pybind11 or not.
- _hal registers the shared classes as _hal.type/_hal.dir. hal.py serves hal.type.REAL, hal.dir.IN etc. (HAL_* spellings are aliases), plus the HALType/HALDir class names so pickle round-trips.
- halpp drops its private py::enum_ copies and casts hal_type_t/hal::dir arguments and results through the shared classes, so tags print with their names and compare equal to the plain constants, which stay exported for compatibility.
- Canonical members are the platform-stable spellings (BOOL, REAL, S32, U32, PORT, S64, U64 and IN..RW); SINT/UINT follow the platform width and are aliases.

pybind11 note: real enum.IntEnum via py::native_enum needs pybind11 >= 3.0 and trixie ships 2.13, so construction goes through the enum module's functional API. The classes are genuine IntEnums either way, so moving to py::native_enum later (TODO in the header) is invisible to user code.

Query results stay plain dicts as discussed; the future hal.query bindings tag type/dir with these classes, no per-function wrappers. tests/haltype.0 is now a sanity suite for the classes, tests/halpp checks halpp shares them. All HAL suites pass.

src/hal/halenum.hh
```diff
@@ -0,0 +1,153 @@
...
+static const halenum_member halenum_type_members[] = {
+    {"BOOL", HAL_BOOL},
+    {"REAL", HAL_REAL},
+    {"S32",  HAL_S32},
+    {"U32",  HAL_U32},
+    {"PORT", HAL_PORT},
+    {"S64",  HAL_S64},
+    {"U64",  HAL_U64},
+    {"SINT", HAL_SINT},
+    {"UINT", HAL_UINT},
```
BsAtHome: These must be ordered differently. I found out in interactive tests that the auto-complete uses the order when duplicates are found (braindead, yes). Order: bool, real, sint, uint, port, s32, u32, s64, u64 (also for the prefixed).

grandixximo: Reordered and the HAL_* spellings following as aliases. SINT/UINT as canonical is safe: on current master HAL_SINT/HAL_UINT are unconditionally HAL_S64/HAL_U64, so there is no platform-width ambiguity left.

lib/python/hal.py
```diff
@@ -29,10 +29,28 @@
...
+globals().pop('dir', None)
+
+# _hal.query is a submodule, not a plain attribute. Registering it in
+# sys.modules under its dotted name makes 'import hal.query' and
+# 'from hal import query' work as they would for a real package.
+sys.modules['hal.query'] = query
+
 def __getattr__(name):
+    if name in ('type', 'dir', 'HALType', 'HALDir'):
+        return getattr(_hal, {'HALType': 'type', 'HALDir': 'dir'}.get(name, name))
```
BsAtHome: I really dislike the disparity we have here with `hal.type.X` being used whereas the printing routine will show you `hal.HALType.X` (same for direction).

This really bugged me when I was writing the documentation.

Why can't we just use `hal.Type.X` and `hal.Dir.X` consistently? The one-letter capital may be slightly sensitive to the occasional error, but having the same enum under two names is even worse from a user's perspective.

Also applicable to `hal.CompType.X` and `hal.RTType.X`.

grandixximo: Agreed, done: the class name and the attribute name are now the same spelling everywhere, `hal.Type.X` / `hal.Dir.X`. repr shows `<Type.REAL: 2>`, pickle resolves `hal.Type`, and the builtin shadowing is gone, so hal.py no longer needs the star-import unbind dance. In 9de1f112bc. `hal.CompType` / `hal.RTType` live in your tree, same rename applies there.

BsAtHome: Check.

PR #4252, opened 2026-07-19
Add missing and reverse default mutex for getter/setter

BsAtHome:
This is on top of #4099 that did not make it in there:
* a missing type and limits
* reverse default mutex

The HAL version is also upped because it crossed in #4132 and both need to change the version.


grandixximo:
Merged as essential #4099 companion HAL_VER bump is better sooner than later.

PR #4256, opened 2026-07-20
hal: Add halcompupdate to migrate .comp files to the new HAL API

grandixximo:
## Problem

With the new HAL types and getter/setter API (#4099, in-tree conversion in #4247, old API deprecated in #4472), out-of-tree `.comp` components that use the legacy declaration types (`float`, `bit`, `s32`, `u32`, `s64`, `u64`, `signed`, `unsigned`) and direct pin/param assignment will stop building when the API break is performed. The HAL shared memory version has already changed, so out-of-tree components must be recompiled regardless. This is the right moment to offer a painless migration path.

## Solution

A new tool, **`halcompupdate`**, rewrites legacy `.comp` files to the new API:

| Area | Conversion |
|---|---|
| Declaration types | `float`->`real`, `bit`->`bool`, `s32`->`si32`, `u32`->`ui32`, `s64`->`sint`, `u64`->`uint`, `signed`->`si32`, `unsigned`->`ui32` (`port` is left alone, no new-style replacement yet) |
| Writes to out/io pins and params | `x = e` -> `x_set(e)`, `x += e` -> `x_set(x + (e))`, `x++` -> `x_set(x + 1)`, arrays `x(i) = e` -> `x_set(i, e)`, chained `a = b = e` -> `a_set(b_set(e))`, `*x_ptr = e` -> `x_set(e)`, `*x_ptr` reads -> `(x)` |
| `#define` macros | bodies are rewritten too, with macro parameters correctly shadowing same-named pins |
| Legacy C types | `double`/`real_t` -> `rtapi_real`; `hal_bit_t` -> `volatile rtapi_bool` etc., the `volatile` qualifier is kept so the conversion is semantics-identical, and each conversion is reported with its line number for review (`--no-c-types` to skip) |
| Reads | unchanged, the bare name (or `name(idx)`) works in both APIs |

**Pin and parameter names are never renamed**, because they are part of the HAL configuration interface. Existing `.hal` files keep working.

Usage:

    halcompupdate mycomponent.comp        # show a diff of the required changes
    halcompupdate -i mycomponent.comp     # rewrite in place (keeps a .bak)
    halcompupdate --check *.comp          # CI mode: exit 1 if migration is needed

## Where it hooks in

`halcompile` emits a deprecation warning once per legacy declaration type used, pointing at `halcompupdate(1)`. It sits exactly at the spot `halcompile.g` reserved for it ("when we start warning ... this is where the warning goes"). Users hit the warning at compile time and get the remedy in the same message.

## Safety

- In-place rewriting is atomic (temp file + `os.replace()`, original mode preserved). Backups are created with `O_EXCL|O_NOFOLLOW` and a unique-name fallback, so a pre-existing `.bak` (or a planted symlink) is never clobbered.
- Constructs that cannot be converted safely are **left unchanged with a warning** instead of producing broken or silently different code: taking the address of a pin/param (`&x`), direct use of the legacy `hal_pin_*_new`/`hal_param_*_new` creation API, postfix `++`/`--` whose value is used (the setter yields the *new* value, postfix must yield the old), array indices with side effects, pointers to legacy HAL types, and `hal_*_t` types in `variable` declarations (the grammar allows only a single-word type there, so the volatile qualifier cannot be preserved).
- Writes to pins/params in `EXTRA_SETUP` are converted but warned about: the setter references are initialized only after `extra_setup()` runs.
- Every file ends with a summary line: how many mechanical changes were applied and how many constructs were left for manual review, with a reminder to review the diff and test before use.

## Validation

- Converted all 124 in-tree components and compiled them from a plain directory (the out-of-tree scenario): 119 compile cleanly. The other 5 (`spindle`, `raster`, `mesa_pktgyro_test`, `homecomp`, `tpcomp`) reference in-tree-only headers/skeletons and fail identically for the already-converted versions when built out-of-tree.
- Output matches the hand conversions in #4247 functionally. Remaining diffs are column alignment, deliberate pin renames done there, and two cases where the manual conversion mangled doc text (e.g. `"32 bool integer"` in reset.comp) which the tool preserves correctly.
- Regression test `tests/halcompile/update-api` covers declarations, writes, unsafe constructs, EXTRA_SETUP writes, volatile preservation, and the summary output.

## Docs

Migration section in `docs/src/hal/comp.adoc` (new-type table and `_set()` accessor documentation), `halcompupdate(1)` manpage, and a SEE ALSO link in `halcompile(1)`.

### DIFF
--- a/src/hal/utils/halcompupdate.py
+++ b/src/hal/utils/halcompupdate.py
@@ -1,0 +1,1060 @@
+#!/usr/bin/env python3
+#
+#    halcompupdate - migrate HAL .comp components to the new HAL pin/param API
+#    Copyright 2026 Luca Toniolo
+#
+#    This program is free software; you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation; either version 2 of the License, or
+#    (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program; if not, write to the Free Software
+#    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# The new HAL API replaces direct memory access to pins and params with
+# strongly typed getter/setter access:
+#
+#   * declaration types change:
+#       float -> real, bit -> bool, s32 -> si32, u32 -> ui32,
+#       s64 -> sint, u64 -> uint, signed -> si32, unsigned -> ui32
+#     ('port' is not converted yet)
+#   * writing an out/io pin or a param is done with the generated
+#     <name>_set(value) function instead of plain assignment
+#   * reading stays transparent (the bare name keeps working)
+#   * <name>_ptr is no longer dereferenceable; *<name>_ptr reads become
+#     (<name>) and *<name>_ptr writes become <name>_set(...)
+#
+# This tool rewrites a .comp file accordingly.  By default it prints a
+# unified diff; use --in-place to rewrite the file.
+
+import argparse
+import difflib
+import os
+import re
+import sys
+import tempfile
+
+OLD2NEW = {
+    'float': 'real',
+    'bit': 'bool',
+    's32': 'si32',
+    'u32': 'ui32',
+    's64': 'sint',
+    'u64': 'uint',
+    'signed': 'si32',
+    'unsigned': 'ui32',
+}
+
+# C types used with the old HAL API and their modern replacements for
+# variable use.  The volatile qualifier of the legacy hal_*_t types exists
+# only because they were used for direct access to HAL memory; as variable
+# types they were always wrong, so the plain rtapi_* forms are used here.
+# Plain C 'float' is deliberately not mapped: it is not part of the HAL
+# API and stays valid C (only 'variable' declarations are converted).
+# Pointers to the legacy types are a different matter (they referenced
+# HAL memory and become opaque hal_*_t references); those are warned
+# about and left unchanged.
+CTYPE_MAP = {
+    'double': 'rtapi_real',
+    'real_t': 'rtapi_real',
+    'hal_float_t': 'rtapi_real',
+    'hal_bit_t': 'rtapi_bool',
+    'hal_s32_t': 'rtapi_s32',
+    'hal_u32_t': 'rtapi_u32',
+    'hal_s64_t': 'rtapi_s64',
+    'hal_u64_t': 'rtapi_u64',
+}
+
+# Mapping for 'variable' declarations.  The comp grammar allows only a
+# single-word type there, and instance variables are component-private
+# memory, so the plain rtapi_* forms are used (no volatile qualifier).
+# Old code frequently used HAL type names here ('float' for HAL floats);
+# C 'float' variables holding HAL values must become rtapi_real.
+# 'signed'/'unsigned' are *not* mapped: they are legitimate plain C types.
+VAR_TYPE_MAP = {
+    'double': 'rtapi_real',
+    'real_t': 'rtapi_real',
+    'float': 'rtapi_real',
+    'hal_float_t': 'rtapi_real',
+    'hal_bit_t': 'rtapi_bool',
+    'hal_s32_t': 'rtapi_s32',
+    'hal_u32_t': 'rtapi_u32',
+    'hal_s64_t': 'rtapi_s64',
+    'hal_u64_t': 'rtapi_u64',
+}
+
+# HAL C types/macros that will disappear at the API break; used in bodies
+LEGACY_C_TYPES = [
+    'hal_bit_t', 'hal_float_t', 'hal_s32_t', 'hal_u32_t',
+    'hal_s64_t', 'hal_u64_t', 'hal_data_u', 'hal_pin_dir_t',
+    'hal_param_dir_t',
+]
+LEGACY_TYPE_MACROS = [
+    'HAL_BIT', 'HAL_FLOAT', 'HAL_S32', 'HAL_U32', 'HAL_S64', 'HAL_U64',
+]
+
+PIN_WRITABLE_DIRS = {'out', 'io'}
+PARAM_WRITABLE_DIRS = {'rw', 'r'}
+
+
+def to_c(name):
+    """Replicates halcompile's to_c(): HAL name -> C identifier."""
+    name = re.sub(r"[-._]*#+", "", name)
+    name = name.replace("#", "").replace(".", "_").replace("-", "_")
+    return re.sub(r"_+", "_", name)
+
+
+class Reporter:
+    def __init__(self, filename, quiet=False):
+        self.filename = filename
+        self.quiet = quiet
+
+    def warn(self, msg, lineno=0):
+        if not self.quiet:
+            print("%s:%d: Warning: %s" % (self.filename, lineno or 0, msg),
+                  file=sys.stderr)
+
+    def error(self, msg):
+        raise SystemExit("%s:0: Error: %s" % (self.filename, msg))
+
+
+def lineno_of(text, pos):
+    return text.count("\n", 0, pos) + 1
+
+
+# ---------------------------------------------------------------------------
+# Header parsing
+# ---------------------------------------------------------------------------
+
+def split_statements(header):
+    """Split the declaration part (before ';;') into statements terminated
+    by ';', ignoring semicolons inside strings and triple-quoted docs.
+    Returns a list of (start, end) spans, end includes the ';'."""
+    spans = []
+    i, start, n = 0, 0, len(header)
+    while i < n:
+        c = header[i]
+        if header.startswith('"""', i) or header.startswith('r"""', i):
+            if header[i] == 'r':
+                i += 1
+            i += 3
+            while i < n:
+                if header[i] == '\\':
+                    i += 2
+                    continue
+                if header.startswith('"""', i):
+                    i += 3
+                    break
+                i += 1
+            continue
+        if c == '"':
+            i += 1
+            while i < n and header[i] != '"':
+                if header[i] == '\\':
+                    i += 1
+                i += 1
+            i += 1
+            continue
+        if c == "'":
+            i += 1
+            while i < n and header[i] != "'":
+                if header[i] == '\\':
+                    i += 1
+                i += 1
+            i += 1
+            continue
+        if header.startswith('//', i):
+            j = header.find('\n', i)
+            i = n if j < 0 else j + 1
+            continue
+        if header.startswith('/*', i):
+            j = header.find('*/', i + 2)
+            i = n if j < 0 else j + 2
+            continue
+        if c == ';':
+            spans.append((start, i + 1))
+            start = i + 1
+        i += 1
+    if start < n:
+        spans.append((start, n))
+    return spans
+
+
+DECL_RE = re.compile(
+    r'^(\s*(?:(?://[^\n]*|/\*.*?\*/)\s*)*)'      # leading whitespace/comments
+    r'(pin|param)\s+(in|out|io|rw|r)\s+'
+    r'([A-Za-z][A-Za-z0-9]*)\s+'                  # HAL type
+    r'([#A-Za-z_][-#A-Za-z0-9_.]*)'               # HAL name
+    r'(\s*\[)?',                                  # array marker
+    re.S)
+
+VAR_RE = re.compile(
+    r'^(\s*(?:(?://[^\n]*|/\*.*?\*/)\s*)*)'      # leading whitespace/comments
+    r'variable\s+'
+    r'([A-Za-z_][A-Za-z0-9_]*)\s+'                # C type
+    r'\*?\s*[#A-Za-z_]',                          # (possibly starred) name
+    re.S)
+
+
+class Decl:
+    def __init__(self, kind, direction, type_, name, is_array):
+        self.kind = kind            # 'pin' or 'param'
+        self.direction = direction
+        self.type = type_
+        self.name = name            # HAL name
+        self.cname = to_c(name)
+        self.is_array = is_array
+
+    @property
+    def writable(self):
+        if self.kind == 'pin':
+            return self.direction in PIN_WRITABLE_DIRS
+        return self.direction in PARAM_WRITABLE_DIRS
+
+
+def parse_declarations(header):
+    """Find pin/param declarations, return (list of Decl, set of options)."""
+    decls = []
+    options = set()
+    for start, end in split_statements(header):
+        stmt = header[start:end]
+        m = DECL_RE.match(stmt)
+        if m:
+            _, kind, direction, type_, name, array = m.groups()
+            decls.append(Decl(kind, direction, type_, name, bool(array)))
+            continue
+        mo = re.match(r'\s*option\s+([A-Za-z_][A-Za-z0-9_]*)', stmt)
+        if mo:
+            options.add(mo.group(1))
+    return decls, options
+
+
+def convert_header(header, rep, c_types=True, legacy_api=False):
+    """Replace old HAL types by new ones in pin/param declarations and
+    modernize C types in 'variable' declarations.  Only the type token
+    itself is replaced, all other text (spacing, docs) is preserved.
+    Returns (new_header, changed, decls, options)."""
+    decls, options = parse_declarations(header)
+    if 'no_convenience_defines' in options:
+        rep.error("component uses 'option no_convenience_defines'; "
+                  "automatic body conversion is not possible")
+    var_map = VAR_TYPE_MAP
+    if legacy_api:
+        # keep hal_*_t: their addresses are passed to legacy API calls
+        var_map = {k: v for k, v in VAR_TYPE_MAP.items()
+                   if not k.startswith('hal_')}
+    edits = []          # (start, end, replacement) in header coordinates
+    for start, end in split_statements(header):
+        stmt = header[start:end]
+        m = DECL_RE.match(stmt)
+        if m:
+            type_ = m.group(4)
+            if type_ in OLD2NEW:
+                edits.append((start + m.start(4), start + m.end(4),
+                              OLD2NEW[type_]))
+            continue
+        if c_types:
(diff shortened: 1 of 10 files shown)


grandixximo:
@BsAtHome If this is too ugly I'll drop it, this does most of your careful hand conversion, to alleviate the porting churn for comps out of tree, let me know how much you hate it ;-)
Keeping as draft for now....

BsAtHome:
The old hal types should never be used anywhere anymore. All places I've see they were used wrongly as variable types and that fails the smell-test miserably because they are volatile.
For example, `hal_bit_t` must be `rtapi_bool` when used in as a variable. `hal_bit_t *` becomes `hal_bool_t` when dealing with pins. And `hal_bit_t` becomes `hal_bool_t` when dealing with parameters, _BUT_, when you deal with parameters, you need to adapt the code because parameters were used as variables.

For components (.comp files), that only declare their pins and params through halcompile means, will not see a difference (except for the _set() suffix for writing). However, those constructs that add their own pins/params need more care in the conversion.

grandixximo:
Thanks for the review. Aligned the C-type conversion with that:

* hal_bit_t (and the other hal_*_t) used as variable types now convert to plain rtapi_bool/rtapi_*. The volatile is dropped, it was only there for direct HAL memory access.
* hal_bit_t * and friends are left unchanged with a warning. Pointers into HAL memory become hal_bool_t refs and need the manual hal_get/hal_set treatment, the tool does not pretend to do that part.
* Comps that only declare pins/params through halcompile get exactly the type rename plus _set() treatment automatically. Comps using hal_pin_*_new/hal_param_*_new directly are detected, their hal_*_t types are kept intact, and a warning points at the manual part.

BsAtHome:
FWIW, I don't think we should even attempt to fix the component files automatically.

The only risk-free change is to fixup pins/params to use the new types. Everything else is intrinsically a huge risk in an auto conversion that needs to consider the context it is used in. That is not something you can do automatically. You could do considerable damage.

IMO, the risks outweigh the benefits.

grandixximo:
> FWIW, I don't think we should even attempt to fix the component files automatically.

Partially I agree, but I don't like what the lack of an attempt says...

> IMO, the risks outweigh the benefits.

The tool is strictly opt-in. It does not pop up and ask you to convert like the INI converter does, and doesn't convert automatically on build, it sits there until someone runs it, and its default mode only prints a diff for review.

> The only risk-free change is to fixup pins/params to use the new types

Types and writes are inseparable. Converting `pin out float x` -> `real` while leaving `x = 1;` in the body breaks the comp immediately (assignment to a getter macro), while doing nothing keeps it compiling until the API break. So the only coherent options are mechanical conversion of both, or nothing but warnings, and IMO only the first one helps out-of-tree comps mentioned by @rene-dev at the meeting.

The tool already stays inside the risk-free zone you describe, it converts only the mechanical subset (type renames and direct writes.

The writable namespace is exactly the declared pins/params, no aliasing, no pointers beyond `_ptr`) and defers everything that needs context with a warning, unchanged: `&pin`, raw `hal_pin_*_new`, postfix `++`/`--` whose value is used, side-effect indices. Validated by converting all 124 in-tree comps and building each from a plain directory, which is the out-of-tree scenario: 119 compile cleanly. The other 5 (spindle, raster, mesa_pktgyro_test, homecomp, tpcomp) depend on in-tree-only headers or the TOPDIR skeleton and fail the same way before and after conversion, so those failures are unrelated to the tool.

grandixximo:
@BsAtHome This needs a decision now that #4472 is merged.

I've rebased and added a commit pointing the hal_*_t deprecation warnings at halcompupdate(1), so merging or closing this PR settles the question in one action.

My position is unchanged: the tool is strictly opt-in, diff-only by default, converts only the mechanical subset, and is verified against all 124 in-tree comps. Not perfect, but for comps that only declare pins/params through halcompile it beats hand-editing every write. And "only fix the types" is not a middle ground, renaming types without converting the writes breaks the comp immediately.

If you still think the risks outweigh the benefits even in this form, I'll close it. But I'd like to hear why, and what would have to change for it to be acceptable.

src/hal/hal.h
```diff
@@ -425,7 +425,7 @@ typedef union {
     hal_s64_t ls;
     hal_u64_t lu;
 #pragma GCC diagnostic pop
-} hal_data_u __HAL_DEPRECATED("Use getter/setter API");
+} hal_data_u __HAL_DEPRECATED("Use getter/setter API; halcompupdate(1) migrates .comp files");
```
BsAtHome: Components (.comp files) should never have to use the old hal types. If they did, then that would have been an error in and of itself.

I have been removing all those instances from our library. If people are (still) using them, then they should fix their code. Telling them to use halcompupdate does not really help.

src/hal/utils/halcompupdate.py
```diff
@@ -0,0 +1,1089 @@
...
+# about and left unchanged.
+CTYPE_MAP = {
+    'double': 'rtapi_real',
+    'real_t': 'rtapi_real',
+    'hal_float_t': 'rtapi_real',
+    'hal_bit_t': 'rtapi_bool',
+    'hal_s32_t': 'rtapi_s32',
+    'hal_u32_t': 'rtapi_u32',
+    'hal_s64_t': 'rtapi_s64',
+    'hal_u64_t': 'rtapi_u64',
```
BsAtHome, line 88: If anybody is using the old hal types, then that is an error. The file cannot be converted because you don't know what they are doing.

src/hal/utils/halcompupdate.py
```diff
@@ -0,0 +1,1089 @@
...
+    'double': 'rtapi_real',
+    'real_t': 'rtapi_real',
+    'float': 'rtapi_real',
+    'hal_float_t': 'rtapi_real',
+    'hal_bit_t': 'rtapi_bool',
+    'hal_s32_t': 'rtapi_s32',
+    'hal_u32_t': 'rtapi_u32',
+    'hal_s64_t': 'rtapi_s64',
+    'hal_u64_t': 'rtapi_u64',
+}
```
BsAtHome, line 100: Same problem here. The code using a `variable` could very well be requiring a volatile decl. You don't know and can therefore not convert.

src/hal/utils/halcompupdate.py
```diff
@@ -0,0 +1,1089 @@
...
+
+# HAL C types/macros that will disappear at the API break; used in bodies
+LEGACY_C_TYPES = [
+    'hal_bit_t', 'hal_float_t', 'hal_s32_t', 'hal_u32_t',
+    'hal_s64_t', 'hal_u64_t', 'hal_data_u', 'hal_pin_dir_t',
```
BsAtHome: Same problem here. How do you know when the volatility is not a requirement without looking at the code?

BsAtHome:
I don't think it is about the risk, at least not for us. That would be a user problem ;-)

The real question is whether it is worth the bother. Is the added value greater than the user-effort required to do a manual conversion?
It may also give you a false sense of security when conversion succeeds. I think we can assume that the quality of in-tree components is generally better than out-of-tree components. Then we should ask ourselves, how do these get converted?

 I'm still quite ambivalent. Yes, it does help the user. No, it may cause hard problems. Do the advantages outweigh the (potential) problems?

src/hal/hal.h
grandixximo: Point taken, that commit is dropped. The hal.h deprecation messages no longer mention the tool. The pointer now lives only in halcompile's own deprecation warning for legacy declaration types, where the audience is exactly .comp authors.

src/hal/utils/halcompupdate.py
grandixximo: Fixed: body conversions now keep the qualifier, `hal_bit_t x` becomes `volatile rtapi_bool x`, semantics-identical whether or not the volatile matters. The plain rtapi_* form only happens under an explicit `--drop-volatile`.

src/hal/utils/halcompupdate.py
grandixximo: Agreed, and the grammar settles it: `variable` takes a single-word type, so volatile cannot even be expressed there. hal_*_t variable declarations are now left unchanged with a warning pointing at `--drop-volatile`.

src/hal/utils/halcompupdate.py
grandixximo: Same change as above: no silent volatile drop anywhere. Bodies preserve it, `variable` declarations refuse, `--drop-volatile` is the explicit opt-in.

grandixximo:
On "worth the bother": a manual conversion is not just the type renames, it is every write to every pin and param, each one a chance to typo a setter or miss an array-index form. For a comp with a handful of writes that is trivial. For the long tail of comps with dozens of writes it is exactly the kind of mechanical, error-prone editing a script does better than a human. The in-tree comps got expert hand conversion; out-of-tree authors do not get that attention.

On false sense of security: the tool now ends every file with a summary, "N mechanical changes, M constructs left for manual review; review the diff and test before use". Default mode prints only a diff, nothing is written unless asked, --in-place keeps a .bak. Everything the tool cannot prove safe (pointers, volatile qualifiers, raw creation API, side-effect indices, EXTRA_SETUP writes) is left unchanged with a specific warning. I do not claim it makes conversion safe; it makes the safe part cheap and the unsafe part loud.

And the alternative to the tool is not careful manual conversion of out-of-tree comps, it is no conversion at all: they simply fail to build after the API break. For those authors an imperfect opt-in helper beats nothing.

src/hal/utils/halcompupdate.py
BsAtHome: Adding `volatile` to the conversion is _nearly_ always wrong. That "nearly" is why I said that you cannot convert. The user needs to inspect these instances and fix them. Adding an option `--drop-volatile` just invites users to ignore the problem and that is bad.

Auto conversion should never try to fix user errors. It will always be borderline and create a false sense of security.

BsAtHome:
Okay, fine then. Let me take another look and see if I can warm to the idea of auto conversion ;-)

src/hal/utils/halcompupdate.py
grandixximo: The flag is gone, and every kept-volatile conversion now gets its own warning with the file and line: "'hal_bit_t' converted to 'volatile rtapi_bool'; the qualifier is kept because the tool cannot know if it is needed, drop it by hand if not". Nothing is decided silently, the spots are named and greppable.

On "never fix user errors": the conversion itself is not optional, hal_bit_t disappears at the API break and an untouched body does not compile. What the tool must not decide is the volatile question, and it does not: the qualifier is carried over verbatim.

`variable` declarations stay untouched with a warning naming the line; the grammar allows only a single-word type there.

BsAtHome:
There is one more thing that I noticed when doing conversion. There are several pins/params that have `s32` or `u32` in their name. This is mostly not appropriate when we kill off the 32-bit version. At least all the names that will be moving to `sint` and `uint` should not have that name.
Currently, my WIP tree is 64-bit only and (mostly) functional. For some, mux_generic.c and demux_generic.c, I've added a pin name alias. The .comp components will get a new line in halcompile `alias pinname aliasname;`. Aliases are already supported by HAL. It should function properly when building nets as hal_link searches for names and aliases. Need to checkup on all paths, also in HAL query.

grandixximo:
One thing I found while reading `mux_generic.c` on your WIP branch: both alias format strings use the `demux-gen.` prefix, lines 226 and 278, while every real pin in that file uses `mux-gen.`. Looks like it came across from `demux_generic.c`.

On the `alias` line in halcompile, some questions before I form an opinion:

Would the directive cover params as well as pins, or pins only? `hal_param_alias()` exists, so the asymmetry would be a choice rather than a limitation.

Which name becomes canonical? In `mux_generic.c` the pin is created with the old type name and the new one is the alias, so `show pin` keeps printing `in-s32-00` and `in-sint-00` is secondary. Is that the intended direction, given the goal is to get the type out of the name?

And what happens to `halcmd save all`? It calls `save_aliases()`, so aliases created by a component appear in a user's saved config as explicit `alias pin ...` lines they never wrote, plus a permanent entry under `show alias`. Is that acceptable, or does a comp-declared alias need to be distinguishable from a user-declared one?

I ask because there is no precedent for a component shipping aliases; so far they have been purely a user-side tool. For halcompupdate none of this changes anything: it does not rename and I do not intend it to, an alias makes a rename non-breaking but not mechanical. Would you want it to at least point the names out, a gentle warning naming pins and params whose name mentions a type that no longer exists, leaving the decision to the author?

BsAtHome:
Yeah, there is some copy/paste in there... :-)
All unreviewed hacks so far.

Both pins and params should be aliased when using halcompile. Working on it in my tree (pins done).

The current format:
```
# Previously:
pin in sint name-s32;

# Now:
pin in sint name-sint;
alias name-sint name-s32;
```
That generates the same "visible" pin list and moves the alias name into oldname. The `hal_pin_alias()` does something like:
* unlink pin from list
* oldname = pinname
* pinname = aliasname
* relink pin into list

It is not always easy to get rid of the type name. See for example `hal/components/reset.comp`.

The question is whether it should be:
1. `prefix.name-s32` + `prefix.name-sint`
2. `prefix.name-sint` + `prefix.name-s32`

The first keeps the same names visible as everybody was used to. The second makes the new name visible while still allowing the old name. Either way works. Not sure which is the best (probably the second).

What happens to halcmd... Well, I renamed the types, all of them. So there are already a lot of changes coming. Saving the list will already change the line.

The user alias question is a good one. No opinion as of yet. The real question is how hard we are going to break every config out there. Probably best to break it completely. There also is this issue of the `conv-*-*.comp` converters. They have the type in the filename and therefore the component name. That sucks too. So that would argue for making a complete break and drop the aliases in the stock components.

And, yes, halcompupdate should probably warm if it detects type names in the pin/param name.

BsAtHome:
You are right.

Removed the in-tree component aliases. Still implemented it in halcompile for integrators if they wish that much.

The conv-X components rename will break some substantial portion of user's configs. Therefore, the aliases are a patch that would hurt more than you get out of it. Once you need to redo the config, then you should redo it thoroughly.

That makes me wonder whether a similar tool as this will be helping to rewrite .hal files for the user. Could even get the file list from an INI file... Want to hack a bit more? ;-)

grandixximo:
Yes, happy to hack on that. Here is the shape I would propose before I write any of it, because it deliberately does not work like the INI updater.

**The trigger is the error, not the tool.** No dialog, no prompt on startup, nothing runs automatically. A user starts the config that worked yesterday, hits the failure at the exact line that uses a name we removed, and the error message itself names the replacement and mentions the converter. Users who never touch a renamed name never hear about it.

This is safe here in a way the INI updater is not: the old names are gone, so every affected config fails deterministically. There is no silent-wrong-behaviour class to worry about, and the mapping is not a guess: `conv_s32_float` was removed, `conv_sint_real` exists, one entry, no context needed.

**Pieces:**

A generated `oldname => newname` table (component names, pin names, param names) in `hal_lib`, produced from the same source as the type conversion so it cannot drift.

One lookup on three existing error paths, only after the call has already failed, so nothing costs anything when the config is fine:

* `halcmd_commands.cc:985`, `insmod for %s failed`, the `loadrt conv_s32_float` case, which is the first line most configs will trip on
* `halcmd_commands.cc:408`, `Pin '%s' does not exist`, for `net` and `linkps`
* `halcmd_commands.cc:673`, the generic `setp`/`setparam` path

Output would read roughly:

```
foo.hal:42: Pin 'conv-s32-float.in' does not exist
  This name was removed in 2.10~pre2; it is now 'conv-sint-real.in'.
  'halnameupdate foo.hal' converts the names in that file.
```

And the converter itself, shaped exactly like `halcompupdate`: diff by default, `-i` rewrites in place keeping a `.bak`, `--check` for CI. It rewrites only names in the table and leaves everything else alone. The primary input is a HAL file, because that is what the error message just handed the user; taking an INI is a convenience on top, for converting a whole config in one go:

```
halnameupdate myfile.hal        # diff
halnameupdate -i myfile.hal     # rewrite in place, keeps a .bak
halnameupdate --check *.hal     # CI mode
halnameupdate mymachine.ini     # every HAL file the INI lists
```

The name is not precious to me, so if you would rather it were called something else, suggest away.

**Two limitations I want to state up front rather than discover later.** Not every failure goes through halcmd: the Python `hal` module, custom GUIs, halui bindings in the INI, and pyvcp/qtvcp widget pin names in XML fail elsewhere. That is an argument for the table living in `hal_lib` where any consumer can query it, and for the INI mode also scanning the GUI files the INI references, but the coverage will not be total. And this does put a legacy table in HAL, so I would write the expiry into the commit when it lands, say dropped two releases after the break, rather than letting it become permanent.

Does this shape look right to you?

BsAtHome:
Embedding lists with things that changed in an executable may not be the best of ways to "fix" a situation. Reminds me of eternal backward compatibility code. You also say, you cannot catch al instances anyway, so the argument for "safe conversion" is not really applicable because you cannot fix everything automatically. Having a table of names in hal_lib is even worse and reminds me of the "wonderful" world of (shattered) windows.

We decided we were going to break people's configs. We can help them overcome the version change. We should not try to fix all problems. Making the conversion easier is all that we can do and should not try to burden our own code with more legacy.

Going through the .hal files is probably the best estimate to fix most of the user's pain. If people are using pins from other code, then that cannot be fixed automatically without full semantic analysis. Just like halcompupdate, you will not be doing this automatically. However, there will be more configurations that need a .hal file update than there will be local out-of-tree components.

You really don't know the pin names for sure without going through all the logic. Components can be loaded with a `names=` argument (several configs have this). Then the pin names will be different too. This is good for conversion because components with names= keep the pin names constant over the change of component name. However, there is also the possibility that you now have (a tiny chance for) false positives. The only really safe automatic conversion is to track the loadrt and fix pin names that match the loadrt statement (also with `count=`).
BTW, you cannot rely on reading one .hal file only. In some stock configs it was seen that the loadrt and the pin usage are split over files (had a count=3). That means only an ini-file search and convert .hal files in sequence with remembering what is there will be required if you want to play it safe.

I'm not sure we need a CI check. We do all conversion beforehand and all sims should be tested. There are some that even today need updating. However, a dry-run mode that just tells you what would be done is nice. That would be nice for a single .hal file as well as ini-file batch mode. I think you called it "diff" ;-)
```
halfileupdate myfile.hal        # dry-run diff single file
halfileupdate mymachine.ini     # dry-run diff every HAL file the INI lists
halfileupdate -i myfile.hal     # rewrite in place, keeps a .bak
halfileupdate -i mymachie.ini   # rewrite in place every HAL file the INI lists, keeps a .bak
```

My color is `halfileupdate` or `updatehalfile`, depending light or dark mode :-)

We may be somewhat fortunate in that there will be an INI-file version update too (if we get that code working). We can attach the `halfileupdate` to the ini update.

PR #4259, opened 2026-07-21
hal: Make HAL mutex recursive and use acquire/release functions

BsAtHome:
Change the HAL mutex from a simple mutex to a recursive mutex. This is necessary to support coming changes in the HAL API.

Many changes in the PR are simply to replace the old mutex get/give with a halpr_mutex_acquire/halpr_mutex_release call. This is actually a hal_lib private API, but removing access to hal_priv.h would require a huge change-set. This way it will be done gradually.
The forced mutex release function (used in halcmd) is now part of hal_lib so that the caller no longer needs to access library internals. Anyway, when you need to force-release the mutex, then all bets are off and you should restart anyway.

<strike>The rtapi_app is now cleared of accessing the HAL private API. The newinst code (if anybody is using it) now uses a call into a specially published hal_lib function to call the constructor instead of extracting a halpr_xxx function.</strike>
The has been postponed because it changes a prototype that may give a problem. However, it is unlikely that anyone is using constructable components since this is (already since 2.2) an experimental feature.

### DIFF
--- a/src/hal/hal.h
+++ b/src/hal/hal.h
@@ -1320,6 +1320,36 @@ static inline rtapi_s64 hal_extend_counter(rtapi_s64 old, rtapi_s64 newlow, int 
     return (rtapi_u64)old + (diff_shifted >> nshift); // unsigned to avoid signed overflow
 }
 
+// Only enable the query API when we are compiling the user-land HAL library
+#ifdef ULAPI
+
+//
+// *** Special functions for rtapi_app and halcmd ***
+//
+// Invoke the constructor for a new instance
+// Uspace/rtapi_app only. Not implemented in halcmd/halrmt.
+int hal_comp_invoke_make(const char *compname, const char *newname, const char *arg);
+
+// Add the insmod arguments to a named (and just loaded) RT component
+// The 'args' argument must be in HAL memory.
+int hal_comp_insmod_args(const char *compname, const char *args);
+
+// -----------------------------------------------------
+// Release the HAL mutex with brute force
+// WARNING:
+// *   Do not use this function in normal code. You will
+// *   probably kill your running instance when you do.
+// *   It is only to recover from an error and you need
+// *   to be able to shut down your instance.
+int hal_mutex_force_release(void);
+// -----------------------------------------------------
+
+// HAL will pretend that the exact base period requested is possible.
+// This mode is not suitable for running real hardware.
+// Returns zero (0) on success or a negative -EACCES error if already set.
+int hal_enforce_exact_base_period(void);
+
+#endif // ULAPI
 
 RTAPI_END_DECLS
 
--- a/src/hal/hal_lib.c
+++ b/src/hal/hal_lib.c
@@ -53,6 +53,20 @@
 
 */
 
+// Special case to get gettid() to work properly for recursive mutex.
+// Doing this in a rtapi_*.h header is a pain because hal_priv.h is still
+// included everywhere and it would fail because of include ordering.
+#ifdef __KERNEL__
+#include <asm/current.h>
+#define halpr_gettid() ((int)(get_current()->pid))
+#else
+#ifndef _GNU_SOURCE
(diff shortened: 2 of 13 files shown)


src/hal/utils/halcmd_commands.cc
grandixximo, line 544: I know you didn't touch this, but is it time to remove it? Later cleanup?

src/hal/hal_lib.c
```diff
@@ -3413,17 +3507,17 @@ static int init_hal_data(void)
     The first time through, the hal_data memory buffer is fresh from
     rtapi_shmem_new(), which means it's initialized to all zero bytes.
     This means hal_data->mutex is valid and unlocked. */
```
grandixximo: comment needs a refresh...

src/hal/hal.h
```diff
@@ -1068,7 +1068,7 @@ extern int hal_stop_threads(void);
     instance of its component.  Return value is >=0 for success,
     <0 for error.
 */
-typedef int(*constructor)(char *prefix, char *arg);
+typedef int(*constructor)(const char *prefix, const char *arg);
```
grandixximo: Out-of-tree C++ will need a change, worth a note in the forum or website? Or put it on a list of things for a later announcement?

grandixximo (commented):
Also, at **`src/hal/utils/halcompile.g:549`**:

The typedef gained const, but halcompile still emits `static int export_1(char *prefix, char *argstr)`. Out-of-tree constructable comps will warn (C) or fail (C++) against the new header. Suggest emitting `const char *` here and at the registration site (line 679 is fine as-is).

Should it be fixed here, or keeping that for later?

src/hal/utils/halcmd_commands.cc
BsAtHome: This is later cleanup. Removal of a lot of stuff is about the last thing to do.

src/hal/hal.h
BsAtHome: I don't think that any components exist that use a constructor so they may be instantiated multiple times.

But, fr the moment, it is not imperative to have it. Reverting this (including rtapi_app call).

src/hal/hal_lib.c
BsAtHome: Yeah, the last line is the culprit. Fixed.

BsAtHome:
> The typedef gained const, but halcompile still emits `static int export_1(char *prefix, char *argstr)`. Out-of-tree constructable comps will warn (C) or fail (C++) against the new header. Suggest emitting `const char *` here and at the registration site (line 679 is fine as-is).
> 
> Should it be fixed here, or keeping that for later?

This is a such a damned if you do, damned if you don't case... The whole constructable component stuff has been experimental for ages and I have never seen an instance where it is used. No in-tree component uses it and wonder if out-of-tree knows it exists. Changing it risks breaking stuff that is a fringe use from the past. Leaving it as is means some things can not be properly isolated.

The `constructable` option is not documented in halcompile and the `newinst` command in halcmd cannot process it (it is embedded i a `#if 0`). Only rtapi_app knows about it.

Probably need to ask on the dev-list if this is used by anybody.

grandixximo:
Ok, are you sending an email up to the ML?
On hold till we get a reply? Or merge as is since I think you are not touching it now if I'm not wrong...

BsAtHome:
> Ok, are you sending an email up to the ML?

Yes, just writing it.

> On hold till we get a reply? Or merge as is since I think you are not touching it now if I'm not wrong...

I took it out so this can get merged without further delay. The constructable/newinst stuff is not paramount to the progression. Just added a FIXME so it gets cleaned up later.

grandixximo:
Only found one _real_ user on github gwassem/linuxcnc_sim , but his code is broken, I think based on machinekit (where newinst actually works and is used) posted an issue there about it.
After having done a quick search of newinst and since I found it used only once, I doubt we'll get anything back from the mailing list.

PR #4260, opened 2026-07-22
hal: Fix implementation types and small adjustments

BsAtHome:
In this PR:
* Fix some types (still all compatible). The function signatures of the new functions were correct, but the implementations lagged.
* A long standing bug in hal_link would not properly copy the 64-bit value from the dummy to the signal, but would only set the lower 32-bit value. This has now been corrected.
* Final preparations are put in place for introducing the HAL query API.
* Update the comment in hal_init_data regarding the mutex.

This PR introduces hal_strerror() for textual representation of the negative error value from many of the HAL functions. The return value of some HAL functions will be adapted to become more fine-grained later. Currently, they mostly return -EINVAL, which often does not describe the actual error encountered. Changing the return value should not pose any problem because the code that actually checks return values looks for equality with zero or less than zero. Returning a more differentiating (negative) value does not functionally alter that code but allows better messages to be generated moving forward.

### DIFF
--- a/src/hal/hal.h
+++ b/src/hal/hal.h
@@ -139,6 +139,7 @@ RTAPI_BEGIN_DECLS
 #include "rtapi_errno.h"
 
 #define HAL_NAME_LEN     55	/* length for pin, signal, etc, names */
+#define HAL_PSEUDO_COMP_PREFIX "__" /* prefix to identify a pseudo component */
 
 /** These locking codes define the state of HAL locking, are used by most functions */
 /** The functions locked will return a -EPERM error message **/
@@ -233,6 +234,10 @@ extern int hal_set_unready(int comp_id);
 */
 extern int hal_unready(int comp_id);
 
+// hal_strerror() returns a brief textual description string about the error
+// identified. The argument should be the negative errno value, as returned by
+// most HAL functions.
+const char *hal_strerror(int err);
 
 /** hal_comp_name() returns the name of the given component, or NULL
     if comp_id is not a loaded component
@@ -556,6 +561,7 @@ int hal_param_new_ui32(int compid, hal_pdir_t dir, hal_uint_t *ref, rtapi_u32  d
 int hal_param_new_sint(int compid, hal_pdir_t dir, hal_sint_t *ref, rtapi_sint def, const char *fmt, ...) __HAL_PFMT(5,6);
 int hal_param_new_uint(int compid, hal_pdir_t dir, hal_uint_t *ref, rtapi_uint def, const char *fmt, ...) __HAL_PFMT(5,6);
 int hal_param_new_real(int compid, hal_pdir_t dir, hal_real_t *ref, rtapi_real def, const char *fmt, ...) __HAL_PFMT(5,6);
+int hal_param_new_fake(int compid, hal_refs_u *refs);
 #undef __HAL_PFMT
 
 /***********************************************************************
@@ -1173,12 +1179,12 @@ extern void hal_port_wait_writable(hal_port_t** port, unsigned count, sig_atomic
  */
 
 typedef union hal_stream_data {
-    real_t f;
-    bool b;
+    rtapi_real f;
+    rtapi_bool b;
     rtapi_s32 s;
     rtapi_u32 u;
-    rtapi_s64 l;
-    rtapi_u64 k;
+    rtapi_sint l;
+    rtapi_uint k;
 } hal_stream_data_u;
 typedef hal_stream_data_u *hal_stream_data_ptr_u;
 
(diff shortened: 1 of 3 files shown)


src/hal/hal_lib.c
```diff
@@ -462,6 +462,25 @@ int hal_exit(int comp_id)
     return 0;
 }
 
+int hal_is_init(void)
```
grandixximo, line 465: Doesn't this need the matching declaration of public prototype
`extern int hal_is_init(void);`
in `hal.h` ?

src/hal/hal_lib.c
```diff
@@ -662,9 +708,12 @@ int hal_set_lock(unsigned char lock_type) {
...
     if (hal_data == NULL) {
-	rtapi_print_msg(RTAPI_MSG_ERR,
-	    "HAL: ERROR: get_lock called before init\n");
-	return -EINVAL;
+        rtapi_print_msg(RTAPI_MSG_ERR, "HAL: ERROR: get_lock called before init\n");
+	// There is no point in returning a negative error value. The return
+	// type wont allow it. We just pretend that everything is locked tight
+	// and nothing can change, which would adequately describe the
+	// unavailability of shared memory.
+        return HAL_LOCK_ALL;
```
grandixximo, line 716: Are we OK with propagating the _(IMO horrible)_ spaces-tab-spaces indent style now?

src/hal/hal_lib.c
```diff
@@ -620,6 +639,33 @@ int hal_unready(int comp_id) {
...
+    case -ERANGE:  return "value out of range";
+    case -ENOMEM:  return "out of memory";
+    case -EISCONN: return "queue already allocated";
+    case -EEXIST:  return "name exists";  // also when pin/param/sig matched with different type
+    case -EFAULT:  return "shared memory not mapped";
+    case -EMFILE:  return "too many modules";
+    case -EPROTO:  return "hal data version mismatch";
+    case -EPERM:   return "hal is locked";
+    case -ENOEXEC: return "missing constructor";
+    default:       return "unknown error";
```
grandixximo, line 665: Would 
`default: return strerror(-err)`
be better here?

grandixximo:
If merge this first my #4251 will fail to build, can you provide an explicit marker like

`#define HAL_HAS_QUERY_API 1`

And I re-gate #4251 on that?

src/hal/hal_lib.c
BsAtHome: Yes, but that is postponed to the rest of the query API. It got copied in and later realized that the prototype was missing. Should be no problem because it is not yet used.

src/hal/hal_lib.c
BsAtHome: I did think about that. The problem is that these error messages are not "standard" errno messages. They are geared specifically towards HAL.
Another problem is that strerror() has a `char *` return and uses a static buffer internally. That I'd like to avoid very much.

grandixximo: Fair

src/hal/hal_lib.c
BsAtHome: Missed that one. Happens when I modify existing code and don't double check. Good you are looking.

BsAtHome:
> If merge this first my #4251 will fail to build, can you provide an explicit marker like

Why and where does that fail?

src/hal/hal.h
```diff
@@ -556,6 +561,7 @@ int hal_param_new_ui32(int compid, hal_pdir_t dir, hal_uint_t *ref, rtapi_u32  d
 int hal_param_new_sint(int compid, hal_pdir_t dir, hal_sint_t *ref, rtapi_sint def, const char *fmt, ...) __HAL_PFMT(5,6);
 int hal_param_new_uint(int compid, hal_pdir_t dir, hal_uint_t *ref, rtapi_uint def, const char *fmt, ...) __HAL_PFMT(5,6);
 int hal_param_new_real(int compid, hal_pdir_t dir, hal_real_t *ref, rtapi_real def, const char *fmt, ...) __HAL_PFMT(5,6);
+int hal_param_new_fake(int compid, hal_refs_u *refs);
```
grandixximo, line 564: Would like to see a comment for API consumer like
`// opaque-by-design, callers must not assume the size.`
or a guarantee the size irrelevant by contract
```
/* hal_param_new_fake() allocates backing storage for one HAL value of any
 * type without registering a named parameter. The returned storage is
 * opaque; access it only through the typed ref members. */
```

BsAtHome: There is a man-page coming for this.

BsAtHome:
BTW, the next PR will provide the HAL query API. That should make most changes to these parts complete.

grandixximo:
> > If merge this first my #4251 will fail to build, can you provide an explicit marker like
> 
> Why and where does that fail?

I will double check the claim in a few hours once I'm on a linux machine

arm sid is broken, are we closing an eye on that?

BsAtHome:
And now CI/Sid is acting up again. Sid has been very unstable lately.

BsAtHome:
> I will double check the claim in a few hours once I'm on a linux machine

Haven't used anything else for 28 years :-)

> arm sid is broken, are we closing an eye on that?

Hm, our messages crossed. Email notify has not yet arrived. Sometimes...

grandixximo:
I don't sid being obviously broken should stall us...

BsAtHome:
> I don't sid being obviously broken should stall us...

I don't think it is a problem either.

grandixximo:
> > If merge this first my #4251 will fail to build, can you provide an explicit marker like
> 
> Why and where does that fail?

It fails because of how I detect the query API in #4251. My by-name code (`hal_getref_p` etc.) only exists once the query API PR lands, so I gated it on

    #if defined(COMPONENT_TYPE_REALTIME)

using the enum as a proxy for "query API is present", because your query PR was the one that introduced that macro in the public header.

This PR moves COMPONENT_TYPE_* into public hal.h on its own, ahead of the query API. From that point the macro is defined but `hal_getref_p` and friends do not exist yet, so my gate opens too early and `hal.hh` fails to compile:

    src/hal/hal.hh:448:64: error: 'hal_query_value_u' has not been declared

Verified on a test branch: this PR + my #4251 commit cherry-picked on top.

BsAtHome:
> It fails because of how I detect the query API in #4251. My by-name code (hal_getref_p etc.) only exists once the query API PR lands, so I gated it on
> #if defined(COMPONENT_TYPE_REALTIME)

That is a wrong assumption. You should not assume any of these defines being there. Your code trying to figure out whether the query code is present needs to be removed anyway.

FWIW, the component type defines are moved to hal.h because they are needed for the query API. This PR is prep work. The next PR will make sense of that again. By that time, your "hack" has become redundant. So, just wait it out for a moment. It is already hard enough not to break master. Trying to keep your not-presently-merged-and-needs-to-change-anyway code running at the same time makes life very difficult. ;-)

grandixximo:
Fair, don't want to overcomplicate things, we'll hold on #4251

PR #4262, opened 2026-07-23
hal: Introduce HAL query API and add/update man-pages

BsAtHome:
This PR introduces the HAL query API as part of hal_lib. It is only available in user-land because it is non-RT code. It should be used by all user-land applications and interfaces that need to query HAL for pins, params, signals, components, functions and threads. It provides a single API for, for example, setp/sets and getp/gets, including much more, without having to duplicate complex parsing code. Current programs will be updated in following PRs.

The library initialization and finalization has been taken out of hal_init and hal_exit so that you may now use the HAL _user-space_ library API without having to create a (fake/temporary) component. Both hal_init and hal_exit call to hal_lib_init and hal_lib_exit at the appropriate time so that existing code will not see any difference.

The patch-set includes man(3) pages for the new functions and updates a lot of the old as well. The format of all has been unified to follow the same structure. There may still be required some adjustment to the docs, but that can and will be handled later.

The query API also adds a common string-to-value parser (setps_common_cb()) that does consistent conversion of values according to the type requested. Booleans accept true/false/1/0/yes/no/on/off and floating point parsing always sets the C locale for parsing with decimal point.

### DIFF
--- a/src/hal/hal_lib_query.c
+++ b/src/hal/hal_lib_query.c
@@ -1,0 +1,1111 @@
+//
+// HAL non-RT query API
+//
+// Copyright 2026  B.Stultiens
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of version 2 of the GNU General
+// Public License as published by the Free Software Foundation.
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+#include "hal.h"
+#include "hal_priv.h"
+
+//=====================================================
+//
+// HAL query API implementation of gets/sets and getp/setp
+//
+// Additionally adds retrieving the pin/param/signal
+// reference for reading and writing using hal_getref_[ps].
+//
+// Replaces all (re-)implementations in other programs as
+// one standard and unified API.
+//
+// NOTE: This API is not exported in the kernel.
+//       It is a user-land thing and should only
+//       be called from non-RT applications.
+//
+// Iteration of HAL's internal data structures is done using
+// the hal_list_* functions. These iterate over any and all
+// HAL structure with a user provided callback and may be
+// used to extract useful data.
+//
+//=====================================================
+
+static int set_common(hal_type_t type, const hal_query_value_u *v, hal_refs_u u, bool setport, bool isparam)
+{
+    // This is a special case because compatibility requires us to write to the
+    // smaller field when addressing parameters. An old-style param has a
+    // user-defined storage that may not be large enough for our new types and
+    // would overwrite other memory.
+    // FIXME: This code must be retired when we do the API break.
+    if(isparam) {
+        switch(type) {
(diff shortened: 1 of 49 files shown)


grandixximo:
Sid is imploding 😩 

This is a sizable one, I'll have to take my time tomorrow...

BsAtHome:
> Sid is imploding 😩

Yeah, once more.

> This is a sizable one, I'll have to take my time tomorrow...

Most of the files are man pages :-)
Actual code is considerably less. But, take your time.

The rest of HAL changes are const-ify, add deprecation and, a more deeper change, thread and function conversion to getter/setter. But all of these are small and easy. Those changes are patient.

src/hal/utils/setps_util.c
```diff
@@ -0,0 +1,115 @@
...
+#include <locale.h>
+
+#include "setps_util.h"
+
+//
+// Common callback parsing the string argument as a value. The actual prototype
+// is:
+//     int setps_common_cb(hal_query_t *q, const char *value)
+//
+// It expects the q->qname to be set to allow error signalling of the HAL_PORT
```
grandixximo: Typo in the comment, there is no qname field
`q->name`

src/hal/hal_lib_query.c
```diff
@@ -0,0 +1,1100 @@
...
+// Return the status of a named component
+//
+int hal_comp_by_name(const char *name, hal_query_t *q)
+{
+    if(NULL == hal_data) {
+        rtapi_print_msg(RTAPI_MSG_ERR, "hal_comp_by_name: HAL shared memory not mapped\n");
+        return -EFAULT;
+    }
+    if(!name) {
+        return -EINVAL;
```
grandixximo, line 1095: Any reason why this should silent?

src/hal/hal_lib_query.c
```diff
@@ -0,0 +1,1100 @@
...
+            // Neither pin nor param name found
+            halpr_mutex_release();
+            return -ENOENT;
+        }
+        // We have the param, copy the data
+        q->qtype     = HAL_QTYPE_PARAM;
+        q->name      = param->name;
+        q->pp.type   = t = param->type;
+        // FIXME: This can be targeted to param's data when the old API is retired
+        q->pp.ref    = u = (hal_refs_u)(hal_sint_t)SHMPTR(param->data_ptr);
```
grandixximo: `hal_set_p()` and `hal_getref_p()` both guard the param path with `!param->data_ptr` => `-EIO`, this one dereferences `SHMPTR(param->data_ptr)` unchecked. Any reason to _not_ keep them consistent?

src/hal/utils/setps_util.c
BsAtHome: Fixed.

src/hal/hal_lib_query.c
BsAtHome: Not really. Fixed. Same issue fixed in hal_statistics().

src/hal/hal_lib_query.c
BsAtHome: Made it consistent. Also, the rest of the query data is now copied before the error is returned.

However, there is no dereference. The SHMPTR just adds the offset to the shared memory address and that is the pointer. If data_ptr is zero (0), then the pointer will point into the very start of the shared memory, which still can be dereferenced. It will just give you some bad value (which is either the HAL memory layout version or that combined with the mutex).

Note that the check is not performed in the iteration (hal_list_p/hal_list_s). There you have the choice to stop iterating on error and return, effectively hiding everything that follows, forever, or, you can ignore the entry. However, since the dereference is not catastrophic, I chose not to do the test in the iteration. Lets just say, when you get into the -EIO situation, then you have probably already lost...

PR #4263, opened 2026-07-23
Pncconf: Add P value message when servo period is changed

Sigma1912:
The P value in the motor PID settings is recommended to be set according to the servo period value. If a config is changed using pncconf tool and the servo period is changed the P values should usually be adjusted as well. This PR adds a popup so the user can choose to have the P values adjusted accordingly.

Adresses https://forum.linuxcnc.org/39-pncconf/59018-if-servo-period-is-changed-the-coresponding-p-values-for-pid-do-not-change#348002

Screenshot from 2026-07-23 18-04-38

cc: @c-morley

### DIFF
--- a/src/emc/usr_intf/pncconf/pages.py
+++ b/src/emc/usr_intf/pncconf/pages.py
@@ -301,6 +301,12 @@ class Pages:
         elif self.d.axes == 2: self.d.available_axes = ['x','z','s']
         self.d.include_spindle = self.w.include_spindle_checkbutton.get_active()
         self.d.units = self.w.units.get_active()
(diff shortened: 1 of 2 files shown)


c-morley:
This works great for steppers but the message is misleading for servos, as the P value is not changed. Maybe we should add to the message 'if using PID steppers' ?

grandixximo:
FYI we are ignoring the sid failures, the issue is upstream

Sigma1912:
@c-morley 
Thank you for the comment. I have expanded the message text with a reference to 'step/dir':
newmsg

c-morley:
Perfect

PR #4266, opened 2026-07-25
hal: Update halmeter and halscope removing access to hal_priv.h.

BsAtHome:
This PR removes direct HAL access from halmeter and halscope. It now uses the query API and getter/setter infrastructure to operate. There are no functional changes.

Both halmeter and halscope are now prepared to handle 64-bit integers for when we remove the public 32-bit versions. Also, some functions had their strings arguments const'ified where appropriate.

### DIFF
--- a/src/hal/utils/scope_rt.c
+++ b/src/hal/utils/scope_rt.c
@@ -35,7 +35,6 @@
 #include <rtapi.h>		/* RTAPI realtime OS API */
 #include <rtapi_app.h>		/* RTAPI realtime module decls */
 #include <hal.h>		/* HAL public API decls */
-#include "../hal_priv.h"	/* HAL private API decls */
 #include "scope_rt.h"		/* scope related declarations */
 #include <rtapi_string.h>
 
@@ -103,7 +102,7 @@ int rtapi_app_main(void)
 	num_samples = SCOPE_NUM_SAMPLES_MAX;
     }
     /* connect to scope shared memory block */
-    skip = (sizeof(scope_shm_control_t) + 3) & ~3;
+    skip = (sizeof(scope_shm_control_t) + 7) & ~7;
     shm_size = skip + num_samples * sizeof(scope_data_t);
     shm_id = rtapi_shmem_new(SCOPE_SHM_KEY, comp_id, shm_size);
     if (shm_id < 0) {
@@ -193,7 +192,7 @@ static void sample(void *arg, long period)
 	ctrl_rt->auto_timer = 0;
 	/* get info about channels */
 	for (n = 0; n < 16; n++) {
-	    ctrl_rt->data_addr[n] = SHMPTR(ctrl_shm->data_offset[n]);
+	    ctrl_rt->data_addr[n] = (hal_refs_u)(hal_uint_t)hal_reference_map(ctrl_shm->data_offset[n]);
 	    ctrl_rt->data_type[n] = ctrl_shm->data_type[n];
 	    ctrl_rt->data_len[n] = ctrl_shm->data_len[n];
 	}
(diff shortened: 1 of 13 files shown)


src/hal/utils/scope_rt.c
```diff
@@ -323,66 +307,33 @@ static int check_trigger(void)
...
     case HAL_U32:
-	compare_result = (value->d_u32 > level->d_u32);
-	break;
+        compare_result = hal_get_ui32(ctrl_rt->data_addr[ctrl_shm->trig_chan - 1].u) > level->u;
+        break;
+    case HAL_UINT:
+        compare_result = hal_get_uint(ctrl_rt->data_addr[ctrl_shm->trig_chan - 1].u) > level->u;
+        break;
+    case HAL_REAL:
+        compare_result = hal_get_real(ctrl_rt->data_addr[ctrl_shm->trig_chan - 1].r) > level->r;
```
grandixximo, line 332: This reintroduces FPU operations in the RT thread. The old code compared REAL values as integers here precisely to avoid FPU use in this function (see the removed IEEE-754 comment above). On RTAI kernel-module builds a double comparison either fails to link (soft-fp libcall) or corrupts kernel FPU state. Please keep the integer-based comparison for REAL triggers, or state explicitly that halscope drops RTAI support.

src/hal/utils/scope_rt.c
```diff
@@ -263,30 +262,17 @@ static void capture_sample(void)
...
-	}
+        if(0 == ctrl_rt->data_len[n]) // Unused channel
+            continue;
+        switch(ctrl_rt->data_type[n]) {
+        case HAL_BOOL: dest->b = hal_get_bool(ctrl_rt->data_addr[n].b); dest++; break;
+        case HAL_S32:  dest->s = hal_get_si32(ctrl_rt->data_addr[n].s); dest++; break;
+        case HAL_U32:  dest->u = hal_get_ui32(ctrl_rt->data_addr[n].u); dest++; break;
+        case HAL_SINT: dest->s = hal_get_sint(ctrl_rt->data_addr[n].s); dest++; break;
+        case HAL_UINT: dest->u = hal_get_uint(ctrl_rt->data_addr[n].u); dest++; break;
+        case HAL_REAL: dest->r = hal_get_real(ctrl_rt->data_addr[n].r); dest++; break;
```
grandixximo, line 273: The old code read 8-byte values in a loop until two reads matched, protecting against torn reads on 32-bit platforms. hal_get_real/hal_get_sint are plain volatile reads (hal.h:556), so on armhf a REAL/SINT/UINT sample can now be torn between the two 32-bit loads. Either restore a local read-until-stable helper here, or add a tear-free getter variant at library level.

grandixximo (dismissed):
Two RT-safety regressions in the scope_rt.c rewrite, see line comments. Everything else in the PR looks good.

src/hal/utils/scope_rt.c
BsAtHome: But, RT threads are no longer the non-FP variant. You enabled FP yourself in #3900. So how does that suddenly fail?

src/hal/utils/scope_rt.c
BsAtHome: That is why the alignment was changed from 32-bit to 64-bit.
(scope.c:514, scope_rt.c:415and scope_rt.c:436)

BsAtHome: BTW, I'm not entirely sure how you reached your conclusion.
How can continuous 8-byte reads protect against a 32-bit torn read/write "until two reads match"? For that you would need to assure that the starting point is aligned.

HAL ensures that all data read/written by getters/setters from pins, params and signals is always 8-byte aligned. The data destination is _always_ 8-byte large and 8-byte aligned. That has been built into HAL (and I ensured it by disallowing less-than-8-byte-aligned allocations).

The destination `dest` is in shared memory and is a union of the value types. Having that misaligned would be wasteful at start and end, but it should not be catastrophic.

src/hal/utils/scope_rt.c
grandixximo: You're right, I forgot #3900 already landed and every RT thread now has FPU state saved. The constraint behind the integer hack is gone. Withdrawing this one.

src/hal/utils/scope_rt.c
grandixximo: Fair points. On 64-bit the aligned 8-byte loads are atomic and the loop was dead code. The only residual case is 32-bit armhf where LDRD is not single-copy atomic, but that is now a property of every getter user, not scope, so fixing it here would be the wrong layer anyway. Withdrawing.

BsAtHome: > The only residual case is 32-bit armhf where LDRD is not single-copy atomic, but that is now a property of every getter user, not scope, so fixing it here would be the wrong layer anyway.

But even that reasoning fails. The getter will get a value from an address that has been assigned by the HAL code, which ensures it is 8-byte aligned.

Currently, while transitioning, the only exception are old-style parameters that are not allocated by HAL, but by the user's code. This will be gone soon enough and the problem is also gone.

Or did I miscount the pin/param structures' offsets for the target?

grandixximo: You did not miscount the offsets; alignment is correctly enforced everywhere except old-style user-storage params, as you said. But alignment is not the mechanism that made the old loop useful. On ARMv7 an aligned 64-bit integer load is LDRD, which the architecture does not guarantee single-copy atomic; it can execute as two 32-bit accesses and a writer on another core can slip in between. VLDR.64 doubles are likewise not architecturally atomic. So on multicore armhf a getter can still tear despite the alignment guarantee. On 64-bit platforms you are completely right and the loop was dead code. I agree this belongs to the library, not halscope; if 32-bit ARM still matters, a tear-free getter variant would be the place, otherwise we accept the caveat and move on.

BsAtHome: Yes, but the atomicity problem has been a problem forever.

That said, there is a guarantee that a RT thread is not interrupted by lower-priority threads. Scope can only be member of one thread afaik. So, the only scenario is when the faster thread (base-thread) interrupts the slower thread (servo-thread). If the slower thread was in the middle of a read/write, then only the base-thread would be effected. Non-RT does not see the data until it is marked ready and for that the slower thread need to finish the copy loop first.

On other occasions ([gs]et[ps] from non-RT), you are right that the atomicity problem may prove problematic. However, when the getter/setter infrastructure is completely done, then we could actually guarantee that too very easily by making the getters and setters use atomic fetch and store. Not sure how much performance impact that will have on the ARM arch. For x86 it wont matter.

grandixximo: Agreed, that closes it for me. One footnote: the priority-preemption guarantee holds per core; on SMP uspace two RT threads can run in parallel on different cores, so the cross-thread tear is theoretically still there on A9-class silicon. Given that hardware is legacy and single-core RT pinning is standard practice, I am fine with leaving it. Atomic fetch/store in the getters is the right end state and the right layer; deferring it until the migration is done makes sense.

PR #4267, opened 2026-07-25
hal: Update the smaller components to getter/setter

BsAtHome:
This is the first set of component updates that converts them from old-style pins/params to the new getter/setter style.

The conversion is quite straight forward. The types are changed so halcompile automatically uses the getter/setter infrastructure (float->real, bit->bool, {s32,signed}->si32, {u32,unsigned}->ui32, s64->sint, u64->uint). All reading from pins and parameters stays the same. Writing requires the _set suffix and the value as argument.

Some pin/param names were changed by appending an '_' for min, max, clamp, is_positive and is_negative. These would otherwise collide with kernel used identifiers. The user visible pin/param name does not change by appending the underscore (hal names have all underscores replaced with dash and are right-stripped of dashes and periods, in that order, by halcompile).

A few components have been slightly altered to remove multiple reads or writes of the same pin/param. Multiple reads/writes are very expensive. The changes were only made if it was a rather trivial change.The functionality was not altered, even if there could have been some extra thought put in (see f.ex. integ.comp where the min/max/assign sequence could have been better if there was a guarantee that min ≤ max; in that case, the ifs could have been separated with an else).

Bad use of old hal types have been replaced by the correct rtapi types that match the underlying type that was meant. Most components will need another pass when we do the 64-bit clean test and upgrade. But that is for later after the user visible types become 64-bit only.

### DIFF
--- a/src/hal/components/abs.comp
+++ b/src/hal/components/abs.comp
@@ -15,11 +15,11 @@
 //   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 component abs "Compute the absolute value and sign of a float input signal";
 
-pin in float in "Analog input value" ;
-pin out float out "Analog output value, always positive";
-pin out bit sign "Sign of input, false for positive, true for negative" ;
-pin out bit is_positive "TRUE if input is positive, FALSE if input is 0 or negative";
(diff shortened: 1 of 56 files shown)


grandixximo:
lgtm

PR #4269, opened 2026-07-26
hal: Update halcmd to use the HAL query API and remove linkpp

### LINKED ISSUE #4265: RFC: Remove linkpp from halcmd
The `linkpp` command has been deprecated for over 19 years. The commit: https://github.com/LinuxCNC/linuxcnc/commit/38d6a090b9bd192cd09136af0ce6ddb8074e012b

It is still lurking in the code. May I please remove it?

BsAtHome:
This PR updates halcmd to use the HAL query API instead of hooking directly into HAL's private data structures. This update also touches the Tcl side (while building tcl/hal.so), which uses the same source files as halcmd.

Several functions in halcmd had their `char *` arguments changed to `const char *`. Also, some static buffers used with snprintf have been removed and replaced with std::string and fmt::format.

Interactive halcmd is supposed to use readline. However, this has been non-functional ever since the library got changed from libreadline into libedit. The changed define was not taken into account. This has been fixed and both interactive history and tab-completion work again. There are still problems in interactive mode, like exiting if a command fails instead of returning to the interactive prompt. That must be sorted out later.

There is no functional change, except a very small one in the halcmd save command. There was a trailing space when saving components. This space has been removed and the test (tests/save.0) has been updated.

This PR also removes the linkpp command, as discussed and fixes #4265.


src/hal/utils/halcmd_commands.cc
```diff
@@ -1690,527 +1398,444 @@ int do_loadusr_cmd(const char *args[])
...
+        int rvt = hal_getref_p(&qt);
+        int rvm = hal_getref_p(&qm);
+        if(!rvt && !rvm) {
+            // note that the scriptmode format string has no \n
+            halcmd_output((!scriptmode ? "%11ld  %-3s  %20s ( %8ld, %8ld )\n" : "%ld %s %s %8ld %ld"),
+                q->thread.period,
+                "YES",  // Always uses FP
+                q->name,
+                (long)hal_get_sint(qt.pp.ref.s),
+                (long)hal_get_sint(qm.pp.ref.s));
```
grandixximo: should this be should be `hal_get_si32` ?

src/hal/utils/halcmd_commands.cc
```diff
@@ -1690,527 +1398,444 @@ int do_loadusr_cmd(const char *args[])
...
+        if(q->comp.type == HAL_COMP_TYPE_USER) {
+            halcmd_output(" %5d %s", q->comp.pid, q->comp.ready ?  "ready" : "initializing");
+        } else {
+            halcmd_output(" %5s %s", "", q->comp.ready ?  "ready" : "initializing");
+        }
+        break;
+    case HAL_COMP_TYPE_OTHER: {
+        hal_query_t qc = {};
+        // Need to reacquire the component name
+        int rv = hal_comp_by_id(q->comp.comp_id, &qc);
```
grandixximo, line 1471: If the plan is for instances to return, this line will misprint.

show comp prints `INST <parent-driver> <instance-name>`

So this will print` INST child child` instead of` INST parent child`

Consider
`int rv = hal_comp_by_id(q->comp.comp_id & 0xffff, &qc);`

src/hal/utils/halcmd_commands.cc
```diff
@@ -2626,300 +2076,310 @@ int do_save_cmd(const char *type, char *filename)
...
-	}
-	next_thread = tptr->next_ptr;
+    hal_list_thread(&q, save_threads_cb, (void *)dst);
+}
+
+static int save_unconnected_input_pin_values_cb(hal_query_t *q, void *arg)
+{
+    FILE *fp = (FILE *)arg;
+    if(!q->pp.signal && (HAL_IN == q->pp.dir || HAL_IO == q->pp.dir)) {
+        fprintf(fp, "setp %s %s\n", q->name, querydata_refstr(q->pp.type, q->pp.ref).c_str());
```
grandixximo: Should this match `save_params` with the new `querydata_refstr_20` ?

src/hal/utils/halcmd_commands.cc
```diff
@@ -1227,174 +985,128 @@ int do_loadrt_cmd(char *mod_name, char *args[])
...
-	/* desired component not found */
-	halcmd_error("component '%s' is not loaded\n", mod_name);
-	return -1;
+    return 0;
+}
+
+int do_unloadrt_cmd(const char *mod_name)
+{
+    if(strcmp(mod_name, "all")) {
+        int rv = hal_comp_by_name(mod_name, NULL);
```
grandixximo: Could we use the type info the new code now has, to give a message that points the user to the actual problem, when the name is found but it is of the wrong type?
something like.
```
if(q.comp.type != HAL_COMP_TYPE_REALTIME) {
    halcmd_error("component '%s' is not a realtime component (try unloadusr instead)\n", mod_name);
    return -EINVAL;
}
```
The old "x is not loaded" and the current "unloadrt failed on x" are both kind of misleading.

BsAtHome: Sure. Just need to add a query arg to the hal_comp_by_name() and a check further down.

src/hal/utils/halcmd_commands.cc
BsAtHome: Indeed.

src/hal/utils/halcmd_commands.cc
BsAtHome: Yes.

src/hal/utils/halcmd_commands.cc
BsAtHome: The component type HAL_COMP_TYPE_OTHER (previously COMPONENT_TYPE_OTHER) is never assigned to a component's type field. So getting here is very hard.

The whole construct is messy. Using a `& 0xffff` on the component ID is unwarranted because it is set from the underlying rtapi shmem counter (at least in uspace). This counter is not a 16-bit value.

The only reason to use the construct is in the 'newinst' case, where a flag is added to the comp_id field. This code is not active and the way it is done is wrong. The code belongs in hal_lib and it should be solved differently. There should be a field in the hal_comp_t structure to point to the next component instance of this module (a comp_id linked list) and not a singular flag.

I've added some FIXME comments to both places.

BsAtHome:
@grandixximo Pushed the update. Have a look.

grandixximo:
lgtm

PR #4271, opened 2026-07-27
hal: Convert classicladder to getter/setter.

BsAtHome:
Move the classicladder HAL interface code over to getter/setter and the HAL query interface.


grandixximo:
lgtm

Issue #4276, opened 2026-07-28
RFC: Retire meson.build files due to bit-rot


BsAtHome:
The meson.build files were introduced in #619 afaics. They have been allowed to accumulate missing bits, bytes, characters and more over the past seven years. You cannot build LinuxCNC using a meson build. No documentation is covered and no HAL is covered. IMO, it is redundant and should be removed.

Any comments?

andypugh:
Are the associated unit tests still present? I am not clear if one was dependent on the other.

BsAtHome:
The unit tests are in [gitroot]/unit_tests/ and yes, they depend on meson. However, from the looks of it, all the tests have probably (quite) some bit-rot too.
The whole meson stuff will not build because we have been reorganising a lot (like the includes). Most tests are about the interpreter. Currently, the interpreter cannot be build out-of-tree because it hooks into a lot of core stuff (I remember disabling the test until it can be looked at and separated out in a proper way). The other test is about blendmath, but that seems not to be doing much interesting. I don't think anyone has run any of the unit tests since they were added 7 years ago.

One change happened 2 years ago, but that was fixing a typo in a comment...

grandixximo:
I support the removal. The only consumer of the meson build is `unit_tests/`, so I audited it and salvaged the worthwhile content as regular runtests in #4281 as a first cleanup step: the blendmath test, plus a new G20/G21 unit-conversion runtest which was the only genuine coverage gap.

The rest of `unit_tests/interp` duplicates existing runtests under `tests/interp/` (the unit tests were originally derived from them) or enshrines behavior the original author flagged as wrong with FIXME comments, so nothing of value is lost when the meson files go.

robEllenberg:
FWIW I added the meson build system a while ago as an experiment to see if
it was easier / more robust than Cmake or autotools (in some ways it is,
but only if we went to the trouble to totally replace the original build).
The unit tests here don't cover that much in part because this was a
prototype for more extensive unit tests that could save time in CI vs
runtests, and potentially do a better job of exercising specific internal
functions rather than external / functional behavior. There's nothing
special about this implementation, so if it's easier to maintain in the new
format, it makes sense to remove the meson build files / old structure.

BsAtHome:
Making a new build system is in itself not the problem. The problem is that this parallel system does not work because it was abandoned and left to bit-rot. It is academic now, the meson.build files have been removed.

[...]

PR #4281, opened 2026-07-29
tests: salvage unit tests from meson-only unit_tests

grandixximo:
First step toward #4276 (retiring the meson.build files).

The only live content depending on the meson build is `unit_tests/` (806 lines). This PR salvages what is worth keeping as regular runtests:

- `tests/blendmath`: the tp blendmath unit test, ported from the greatest framework to plain C with identical assertions. It compiles `blendmath.c` at test time; the unreferenced rest of the trajectory planner is garbage-collected at link time.
- `tests/interp/unit-conversion`: a new runtest covering G20/G21 unit switching, the one genuine coverage gap I found. It verifies that the current position converts and that the offset parameters (#5211, #5221) stay constant, since they are stored in user (INI) units.

The remainder of `unit_tests/interp` is not carried over: the G10/G92/G52 cases duplicate existing runtests under `tests/interp/` (they were originally derived from them), two sections enshrine known-wrong rotation behavior flagged with FIXME by the original author, and the block-parse and toString tests only exercise parser internals.

Verified: `scripts/runtests` passes for both new tests and the neighboring interp suites (9/9).


tests/blendmath/test_blendmath.c
```diff
@@ -0,0 +1,93 @@
...
+ * (originally by Robert W. Ellenberg) and converted from the greatest
+ * framework to plain C so it can build and run as a runtest.
+ */
+
+#include <math.h>
+#include <stdio.h>
+#include <stdarg.h>
+
+#include <rtapi.h>
+#include "posemath.h"
```
BsAtHome: Posemath comes from the include dir. Use `<>`.

tests/blendmath/test.sh
```diff
@@ -0,0 +1,17 @@
...
+# RIP layout: $HEADERS is $TOPDIR/include
+TOPDIR=$(dirname "$HEADERS")
+
+# blendmath.c references the rest of the TP (tc.c, spherical_arc.c,
+# sp_scurve.c, motion globals). Only the tested functions are used here, so
+# build with function sections and let the linker garbage-collect the rest
+# instead of linking the whole trajectory planner.
+gcc -O2 -Wall -ffunction-sections -fdata-sections -DULAPI -DUNIT_TEST \
+    -I"$HEADERS" -I"$TOPDIR/src" -I"$TOPDIR/src/emc/tp" -I"$TOPDIR/src/emc/motion" \
+    -o test_blendmath test_blendmath.c "$TOPDIR/src/emc/tp/blendmath.c" \
```
BsAtHome, line 13: We decided that we need explicit directories in includes. Only add 'emc'. and fix the include statements in the code.

BsAtHome:
Interesting,... Now the halui/mdi test fails again.

grandixximo:
Yes, seems like its failing quite often today, almost every other build, since I merged your 4 PR's this morning

BsAtHome:
> Yes, seems like its failing quite often today, almost every other build, since I merged your 4 PR's this morning

That doesn't make sense. The PRs don't touch halui (that is still in my tree).

The problem seems to be syncing halui instructions to emc via nml.

BsAtHome:
BTW, the amount of CI failures has been rather high lately (disregarding sid problems) and things have been stuck many times. Not sure how much of that is our fault... But, we probably have some race issue at hand.

tests/blendmath/test_blendmath.c
grandixximo: Fixed

tests/blendmath/test.sh
grandixximo: Done

BsAtHome:
Now all meson.build files and the unit_tests dir can be removed.

grandixximo:
> Now all meson.build files and the unit_tests dir can be removed.

Do you have that ready in the pipe? Or leaving the pleasure to me?

BsAtHome:
Nothing in my pipeline. You are welcome to take the pleasure, if you want it.

Issue #4283, opened 2026-07-29
Investigate intermittent CI fail on homing


BsAtHome:
> (from #4270) the mdi test fails sometime, have you also noticed this? I had it happen on #4280 merge build and here, not sure what's it about, was it happening before too, or is something new?

@grandixximo You may be referring to the non-homed state failure that happens. There still seems to be a RT-to-non-RT or non-RT-to-RT timing/race issue in the homing code. That I've seen happen several times.

The linuxcncrsh test fails on it once in a while. One of the joints fails to home and then the checked for output is off. Probably when the CI machines are busy and the race or timing gets hit. There are other tests that use linuxcncrsh and all would be prone to the same intermittent problem if they use homing.

grandixximo:
I looked at the two recent halui/mdi failures (the master build and the three21 PR) and I think they may be a different issue from the homing race.

The failure lines were:

- `timeout waiting for halui to report mode is-auto`
- `timeout waiting for task mode to get to 3 (it's 2)`

That test does not home: its halui.ini has `NO_FORCE_HOMING=1` and no homing moves configured, so the homing race probably does not apply there. The "it's 2" case looks like halui ran an MDI command and task never got switched back to AUTO afterwards, so maybe the mode-restore or an NML command/status round trip stalls when the CI machines are busy. It could be the same timing umbrella, but the mechanism seems different from homing.

Perhaps it is worth tracking the mdi mode-switch timeouts as a separate item. There is some history here too: d69d170db3 already bumped these timeouts once.

BsAtHome:
But that is the whole point. The pure timeout strategy is problematic because you can't be sure it holds the promise you are looking for. Normally this type of problem exists because of RT/non-RT asynchronicity. We need to look at the sync-points and how we check (that is why the heartbeat was fixed).

BsAtHome:
The problem in halui/mdi seems to be:
```
11.588: timeout waiting for task mode to get to  2 (it's 1)
```
The message pops up when `EMC_TASK_STAT::mode` does not change from `EMC_TASK_MODE::MANUAL` to `EMC_TASK_MODE::AUTO`.

The problem may be in https://github.com/LinuxCNC/linuxcnc/blob/d5c36874b541de75270b0f5d0ae75f2fe4c8e33c/src/emc/task/emctask.cc#L392-L402 because it prevents the mode from moving forward if some conditions are not met.

[...]

PR #4284, opened 2026-07-29
motion: defer mode transitions instead of dropping them when not in position

grandixximo:
Motion's set_operating_mode() honored a requested mode change (teleop, coord, free) only if motion was in position in the single servo cycle that processed the request; otherwise the request was silently erased. Task acknowledges EMC_TASK_SET_MODE as soon as it is read and flips mdiOrAuto immediately, so a dropped transition left task mode and traj mode permanently diverged: determineMode() reports MANUAL whenever traj is FREE/TELEOP, nothing re-issues the request, and no error is reported anywhere. The mode machine stays wedged until some unrelated mode command happens by.

This is the likely cause of the intermittent halui/mdi CI failures tracked in #4283 ("timeout waiting for task mode" / "halui to report mode"). Those are wedges, not slowness, so no timeout bump can fix them.

Fix: keep the request pending when not in position; the transition completes as soon as motion comes to rest, matching the documented EMCMOT_COORD contract ("can be done at any time") and the existing deferred-enable behavior.

Verification:
- deterministic repro: requesting MANUAL mid-MDI-move wedges task mode at AUTO on master; with this change the mode converges to MANUAL when the move ends
- halui (jogging, mdi), linuxcncrsh, mdi-queue and abort runtests pass
- 60/60 halui/mdi iterations under CPU and IO load

### DIFF
--- a/src/emc/motion/control.c
+++ b/src/emc/motion/control.c
@@ -956,8 +956,8 @@ static void set_operating_mode(void)
             // entering teleop (INPOS), remove ext offsets
             axis_sync_teleop_tp_to_carte_pos(-1, pcmd_p);
 	} else {
-	    /* not in position-- don't honor mode change */
-	    emcmotInternal->teleoperating = 0;
+	    /* not in position-- defer the mode change; the request stays
+	       pending and is honored when motion comes to rest */
 	}
     } else {
 	if (GET_MOTION_INPOS_FLAG()) {
@@ -996,8 +996,8 @@ static void set_operating_mode(void)
 		SET_MOTION_TELEOP_FLAG(0);
 		SET_MOTION_ERROR_FLAG(0);
 	    } else {
-		/* not in position-- don't honor mode change */
-		emcmotInternal->coordinating = 0;
+		/* not in position-- defer the mode change; the request stays
+		   pending and is honored when motion comes to rest */
 	    }
 	}
 
@@ -1016,8 +1016,8 @@ static void set_operating_mode(void)
 		SET_MOTION_TELEOP_FLAG(0);
 		SET_MOTION_ERROR_FLAG(0);
 	    } else {
-		/* not in position-- don't honor mode change */
-		emcmotInternal->coordinating = 1;
+		/* not in position-- defer the mode change; the request stays
+		   pending and is honored when motion comes to rest */
 	    }
 	}
     }


BsAtHome:
So:
1. `motion-command-handler` sets the teleoperating/coordinating flags when the command arrives
2. cases:
    * `motion-controller` resets teleoperating flag when in motion but the mode change did not happen
    * `motion-controller` sets/resets coordinating flag (depending on previous state) when in motion but the mode change did not happen
3. command from task is lost and expected state change has evaporated

I guess that multiple rapid sent commands about mode changes would not matter as long as motion is in progress. Both flags would be overwritten each time when `motion-command-handler` runs.

There is a side-effect when going to teleop mode. The `switch_to_teleop_mode()` function (motion.c:169) is called from command.c:624 and will alter the joint's free_tp.enable setting unconditionally. That means changing from one to the other would alter the state before the actual change is in effect. How does this affect the system?

grandixximo:
Good catch, I looked at that path in detail and ran some tests on it.

`switch_to_teleop_mode()` does three things at command time: refuses if non-identity kins and not all homed, clears `free_tp.enable` on all joints, and sets the `teleoperating`/`coordinating` request flags. Effects while the actual transition is still pending:

1. In COORD mode (coordinated move running): `free_tp` is not in the output path at all, `pos_cmd` comes from `coord_tp` via the cubics in `get_pos_cmds()`, so clearing `free_tp.enable` is a no-op and the running move is not disturbed. The request flags are only read by `set_operating_mode()`; `motion_state`, which selects the active planner, only flips at INPOS.

2. Via task this window barely overlaps real motion anyway: switching from MDI/AUTO to MANUAL aborts motion first (emctaskmain.cc, `EMC_TASK_SET_MODE_TYPE`), and jogs block mode changes entirely (`jogging_is_active` in `emcTaskSetMode`). So in practice the pending window overlaps at most the abort decel.

3. If a joint jog were somehow active anyway (only a direct EMCMOT_TELEOP writer could produce that): `simple_tp` with `enable=0` ramps velocity to zero at the accel limit and holds position (`pos_cmd = curr_pos` while disabled), so a controlled stop, no position step. The joint comes to rest, INPOS asserts, the transition completes.

4. Homing: for non-identity kins the allhomed check refuses. For identity kins a TELEOP request mid-home would clear `free_tp.enable` and the home state machine aborts with "end of move" (homing.c). Note this behavior is identical before and after this PR: the command-time `free_tp.enable = 0` and the flag writes are unchanged, the PR only changes whether the pending request survives until INPOS instead of being erased.

Tested on a sim config (trivkins):
- mid-MDI-move MANUAL request: move aborts (by design), clean stop, zero following error, mode converges to MANUAL at rest
- jog after the deferred transition works
- TELEOP => COORD re-entry and further MDI moves work
- rapid MANUAL-then-MDI during a move: newest request wins, MDI functional afterwards

So I think the early `free_tp.enable = 0` is harmless in every path task can produce, and the one case where it bites (identity-kins homing interrupted by a mode request) predates this change.

grandixximo:
run 30819381081 failed on 
```
tests/halui/mdi:
timeout waiting for halui to report mode is-auto
```
This is better in tree sooner than later if we want to avoid these random racing failures, merging.

PR #4285, opened 2026-07-29
hal: Update inihal, halui and motion-logger to getter/setter

BsAtHome:
The conversion of inihal and halui are quite straight forward albeit many pins. The challenge is in the dual use of the structure construct where it both handled creation of the version in HAL memory with the pins and a copy structure using the underlying types.
There is a caveat where the code needs to change when we do the 64-bit cleanup. The expansion used with the macros is based on concatenation and both 32-bit and 64-bit integers are using the same hal type (hal_[su]int_t), but the getter/setter is called hal_[gs]et_[us]i32(). The sint/uint vs ui32/si32 mismatch is solved by a temporary typedef, but that excludes using both 64- and 32-bit versions simultaneously because the compiler would see the overlap and complain. There are at present no 64-bit pins in these components so the problem is not serious and should solve itself when we do the 64-bit cleanup.

The motion-logger update is rather trivial.

### DIFF
--- a/src/emc/motion-logger/motion-logger.c
+++ b/src/emc/motion-logger/motion-logger.c
@@ -38,7 +38,7 @@
 #include "motion/axis.h"
 
 static struct motion_logger_data_t {
-    hal_bit_t *reopen;
+    hal_bool_t reopen;
 } *motion_logger_data;
 
 FILE *logfile = NULL;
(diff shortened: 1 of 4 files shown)


src/emc/motion-logger/motion-logger.c
```diff
@@ -296,13 +296,12 @@ int main(int argc, char* argv[]) {
         exit(EXIT_FAILURE);
     }
     int r;
-    if((r = hal_pin_bit_new("motion-logger.reopen-log", HAL_IO, &motion_logger_data->reopen, mot_comp_id)) < 0) {
+    if((r = hal_pin_new_bool(mot_comp_id, HAL_IO, &motion_logger_data->reopen, 0, "motion-logger.reopen-log")) < 0) {
         hal_exit(mot_comp_id);
         errno = -r;
         perror("hal_pin_bit_new");
```
grandixximo: revise the error too?

BsAtHome: Yes, indeed.

FWIW, handling errors and giving messages is not the code's strong suite (everywhere) ;-)

PR #4286, opened 2026-07-29
hal: Update motion to getter/setter.

BsAtHome:
Update the motion controller code. It is mostly a trivial conversion, but there are many pins.

Some pin creation was changed to be consistently wrapped in the `CALL_CHECK()` macro. Also, some ternaries or trivial if/else constructs in setting boolean values were reduced to one-liners.

### DIFF
--- a/src/emc/motion/axis.c
+++ b/src/emc/motion/axis.c
@@ -28,26 +28,26 @@ typedef struct {
 } emcmot_axis_t;
 
 typedef struct {
-    hal_float_t *pos_cmd;           /* RPI: commanded position */
-    hal_float_t *teleop_vel_cmd;    /* RPI: commanded velocity */
-    hal_float_t *teleop_pos_cmd;    /* RPI: teleop traj planner pos cmd */
-    hal_float_t *teleop_vel_lim;    /* RPI: teleop traj planner vel limit */
-    hal_bit_t   *teleop_tp_enable;  /* RPI: teleop traj planner is running */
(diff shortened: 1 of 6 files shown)


src/emc/motion/axis.c
```diff
@@ -380,14 +380,14 @@ void axis_handle_jogwheels(bool motion_teleop_flag, bool motion_enable_flag, boo
...
 
         // disallow accel bogus fractions
-        if (   (*(axis_data->ajog_accel_fraction) > 1)
-            || (*(axis_data->ajog_accel_fraction) < 0) ) {
+        if (   (hal_get_real(axis_data->ajog_accel_fraction) > 1)
+            || (hal_get_real(axis_data->ajog_accel_fraction) < 0) ) {
             aaccel_limit = axis->acc_limit;
         } else {
-            aaccel_limit = *(axis_data->ajog_accel_fraction) * axis->acc_limit;
+            aaccel_limit = hal_get_real(axis_data->ajog_accel_fraction) * axis->acc_limit;
```
grandixximo: Can we have the version with the single getter on top here as well?

BsAtHome: Sure, fixed.

But you have to remember though, there are many places in the code where this type of correction is needed. Many are harder to solve because they are in loops that are rather complex. All the current changes do not make it worse. Improving more can be a separate chore.

grandixximo: Got it, I'll try to be less picky, thank you for fixing it ;-)

PR #4291, opened 2026-07-30
SimConfigs: Remove obsolete config 'rolfmill'

Sigma1912:
The 'rolfmill' config has been outdated for a long time now ( The '[TRAJ]AXES' entry was removed in version 2.8).
The alternative '3axis-tutorial' config is an updated copy and should be used instead.


grandixximo:
squeaky clean

PR #4295, opened 2026-07-30
build: Retire the meson build and the meson-only unit_tests

### LINKED ISSUE #4276: RFC: Retire meson.build files due to bit-rot

grandixximo:
Closes #4276.

The meson build files were added in 2019 (#619) and have bit-rotted ever since: they cannot build LinuxCNC and cover neither the docs nor HAL. The only live consumer was `unit_tests/`, whose worthwhile content I already salvaged as regular runtests in #4281 (`tests/blendmath` and `tests/interp/unit-conversion`). The rest duplicated existing runtests under `tests/interp/` or enshrined behavior the original author flagged as wrong with FIXMEs, so nothing of value is lost here.

This removes all 13 `meson.build` files and the `unit_tests/` directory (24 files, ~17.5k lines), plus the now-orphaned `debian/copyright` stanzas for the vendored catch.hpp and greatest.h test frameworks. Nothing in the make build, CI, or docs referenced meson; I verified no references remain outside two comments in the salvaged tests.

With @BsAtHome's blessing from #4281.


BsAtHome:
It isn't often you get the pleasure to retire 17k lines of code :-)

PR #4297, opened 2026-07-30
hal: Update more kinematics to getter/setter.

BsAtHome:
These should be the last kinematics files to convert to getter/setter.

The set includes the examples userkins.comp and millturn.comp. These have useless pins/params that are part of the example and they are still there.

Apparently also examples, the xyzacb_trsrn.comp and xyzbca_trsrn.comp kinematics were cleaned up a bit. Redundant code was removed, like exporting a function and unused pins/params. The single required pin (for halcompile) is not called dummy. This should have no bearing on the functionality.

The matrixkins.comp is used in a test, so any problem would probably be rather obvious.

### DIFF
--- a/src/hal/components/millturn.comp
+++ b/src/hal/components/millturn.comp
@@ -28,7 +28,7 @@ chapter (docs/src/motion/switchkins.txt)
 """;
 // The fpin pin is not accessible in kinematics functions.
 // Use the *_setup() function for pins and params used by kinematics.
-pin out s32 fpin=0"pin to demonstrate use of a conventional (non-kinematics) function fdemo";
+pin out si32 fpin=0"pin to demonstrate use of a conventional (non-kinematics) function fdemo";
 option period no;
 function fdemo;
 license "GPL";
@@ -40,15 +40,15 @@ author "David Mueller";
 
 static struct haldata {
   // Example pin pointers:
-  hal_u32_t *in;
-  hal_u32_t *out;
+  hal_uint_t in;
+  hal_uint_t out;
   // Example parameters:
-  hal_float_t param_rw;
-  hal_float_t param_ro;
+  //hal_real_t param_rw;
+  //hal_real_t param_ro;
 
   //Declare hal pin pointers used for switchable kinematics
-  hal_bit_t   *kinstype_is_0;
-  hal_bit_t   *kinstype_is_1;
+  hal_bool_t kinstype_is_0;
+  hal_bool_t kinstype_is_1;
 } *haldata;
 
 FUNCTION(fdemo) {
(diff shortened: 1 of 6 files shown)


src/hal/components/millturn.comp
```diff
@@ -69,23 +69,21 @@ static int millturn_setup(void) {
...
 
     // hal pin examples:
-    res += hal_pin_u32_newf(HAL_IN ,&(haldata->in) ,comp_id,"%s.in" ,HAL_PREFIX);
-    res += hal_pin_u32_newf(HAL_OUT,&(haldata->out),comp_id,"%s.out",HAL_PREFIX);
+    res += hal_pin_new_ui32(comp_id, HAL_IN,  &haldata->in,  0, "%s.in",  HAL_PREFIX);
+    res += hal_pin_new_ui32(comp_id, HAL_OUT, &haldata->out, 0, "%s.out", HAL_PREFIX);
     // hal parameter examples:
-    res += hal_param_float_newf(HAL_RW, &haldata->param_rw,comp_id,"%s.param-rw",HAL_PREFIX);
-    res += hal_param_float_newf(HAL_RO, &haldata->param_ro,comp_id,"%s.param-ro",HAL_PREFIX);
+    //res += hal_param_new_real(comp_id, HAL_RW, &haldata->param_rw, 0.0, "%s.param-rw", HAL_PREFIX);
```
grandixximo, line 79: Question: userkins.comp converts these same two example params with hal_param_new_real(), here they are commented out instead. This drops millturn.param-rw and millturn.param-ro from the HAL interface. Intentional, or should they be converted like in userkins?

src/hal/components/xyzacb_trsrn.comp
```diff
@@ -5,11 +5,8 @@ description
 FIXME
 
 """;
-// The fpin pin is not accessible in kinematics functions.
-// Use the *_setup() function for pins and params used by kinematics.
-pin out s32 fpin=0"pin to demonstrate use of a conventional (non-kinematics) function fdemo";
+pin out si32 dummy=0 "dummy pin to satisfy halcompile";
```
grandixximo, line 8: Just to confirm intent: this drops the fdemo function and the demo in/out pins (same in xyzbca_trsrn.comp), while userkins.comp keeps its converted fdemo. I found no references to them in docs or configs, so no breakage that I can see. Fine with me if deliberate.

src/hal/components/millturn.comp
BsAtHome: Yes, this is intentional. The params are not used in the example nor are they addressed in any userspace example. They are there to show what is possible.

I think it is a bad example pattern to have these pins/params, but that is IMO. They have been uncritically copy-pasted and committed. But the example does use the pins and I don't feel like taking it all out.

src/hal/components/xyzacb_trsrn.comp
BsAtHome: Yes, it is intentional. It is IMO a bad pattern to have these examples proliferate with junk code in them.

PR #4299, opened 2026-07-30
halcompile: reject declarations that export the same HAL name (2.9 backport)

tzuohann:
2.9 backport of #4298.

`pin in float my_input` is exported as `component.N.my-input` — underscores become dashes (comp.adoc, HALNAME).

A component that declares both `x_y` and `x_y_` exports both as `x-y`. halcompile accepts it; the module then fails at loadrt:

```
HAL: ERROR: duplicate variable 'collide.0.x-y'
collide: rtapi_app_main: Invalid argument (-22)
```

`check_name_ok()` only compares declared names. `check_hal_name()` rejects the collision at the offending line and points at the HALNAME documentation. Functions are tracked separately from pins and params, which share one namespace in `hal_lib.c`.

Docs: NAMES section in the halcompile man page, note under the HALNAME table in comp.adoc. Tests: `tests/halcompile/halname`. All 119 in-tree `.comp` files preprocess with no new output.

Man page there is the tracked troff `docs/man/man1/halcompile.1`, not the adoc.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/man/man1/halcompile.1
+++ b/docs/man/man1/halcompile.1
@@ -76,7 +76,13 @@ Install \fB.c\fR and \fB.py\fR files into the proper directory for HAL non-realt
 Extract documentation from \fB.comp\fR files into \fB.9\fR manpage files in the proper system directory (the \fB\-\-install\fR flag), which may require \fIsudo\fR to write to system directories.
 .IP \(bu 4
 Preprocess \fB.comp\fR files into \fB.c\fR files (the \fB\-\-preprocess\fR flag)
-.SH "SEE ALSO"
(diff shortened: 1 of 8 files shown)


grandixximo:
Had a weird CI failure, re-running, @BsAtHome shmem issue in 2.9?

BsAtHome:
> Had a weird CI failure, re-running, @BsAtHome shmem issue in 2.9?

No idea...

However, I do not think this should be backported. At least we should have an agreement how this must be done before we can think about backporting.

PR #4308, opened 2026-07-31
homing: Remove the unconditional assignment to prevent race

### LINKED ISSUE #4283: Investigate intermittent CI fail on homing

### LINKED ISSUE #4289: homing: dead race-prevention guard in base_write_homing_out_pins (double write)
Found while reviewing the getter/setter conversion of the motion module (#4286), which faithfully reproduces this master code.

`src/emc/motion/homing.c`, `base_write_homing_out_pins()` (around line 547):

```c
if(!(*(addr->homing) && !H[jno].homing && homing_active)) {
    // Prevent race condition.
    // ... Do not update the homing status when going from homing --> not homing
    // and the state machine is still active. The homing status deassertion
    // must be delayed until the state machine is done.
    *(addr->homing) = H[jno].homing;
}
*(addr->homing) = H[jno].homing;       // OUT
```

The `if` guard exists to skip the pin update in one specific case (homing -> not homing while the state machine is still active), so the `homing` pin deassertion is delayed and a new homing command cannot fail on a stale state.

But the very next line writes the same value to the same pin unconditionally. Whenever the guard skips its own write, the following line does it anyway. The guard has no effect; the race it describes is not actually prevented.

Two possible fixes, depending on which behavior is intended:

1. The delayed deassertion is needed: delete the unconditional write (keep only the guarded one).
2. The eager update has proven fine in practice: delete the dead guard and its comment.

I do not know which is correct; the comment suggests (1), but the code has effectively been doing (2) for a long time without obvious fallout.

BsAtHome:
The race condition identified and fixed in #3800 did not correctly remove the unconditional assignment as noted in #4289. Therefore, the fix was incomplete.
The race issue was present in both controller.c and homing.c. The controller one was correctly addressed and addressed internal consistency whereas the homing.c version addresses a pin and external synchronization.

It is unknown if this fixes #4283 because the linuxcncrsh race does not involve a pin afaics. We'll have to look into that some more.

Fixes #4289.

### DIFF
--- a/src/emc/motion/homing.c
+++ b/src/emc/motion/homing.c
@@ -557,7 +557,6 @@ static void base_write_homing_out_pins(int njoints)
             // must be delayed until the state machine is done.
             hal_set_bool(addr->homing, H[jno].homing);
         }
-        hal_set_bool(addr->homing,     H[jno].homing);     // OUT
         hal_set_bool(addr->homed,      H[jno].homed);      // OUT
         hal_set_si32(addr->home_state, H[jno].home_state); // OUT
         // index_enable is a HAL_IO pin: the encoder driver also writes it


grandixximo:
This is the missing half of #3800. The conditional write you added in `homing.c` was immediately overwritten by the unconditional `hal_set_bool(addr->homing, H[jno].homing)` on the next line, so the HAL pin still dropped before `homing_active` cleared. I reproduced it with a 2-joint sim where joint 0 finishes while joint 1 is stalled waiting for an index pulse; on base the `joint.0.homing` pin goes `(1,0) → (0,0) → (0,1)`, i.e. it deasserts while the state machine is still active. With #4308 the same run gives `(1,0) → (1,1)` — the pin stays high until the sequence is truly done.

This closes the HAL-pin path from #4289. It does not fix #4283; that failure is in the NML/motion mode-switch path (`set_operating_mode` discarding the request when `INPOS` is false), which is what #4284 addresses.

PR #4310, opened 2026-07-31
hal: Update first batch of C components to use getter/setter

BsAtHome:
These are the first batch of components converted to getter/setter that are partly or in whole written C. There is nothing really wild in the conversion.

Several places have had old hal types replaced with the proper rtapi types. The hal types were being abused as normal variable types, making things unnecessarily slow.

There are several places that could have been optimized or streamlined. Some places it was improved, other places it was not. A future sweep across all components should review them anyway and then can be improved.

The debounce.c component has a possible define to export the state variable as a parameter. It was opted to use the preprocessor to select a hal type or a normal type and add access-functions to wrap read/write to the state. Normally, the EXPORT_STATE is undefined and using a normal variable is faster. Otherwise it would have been unnecessarily slower going though the volatility of hal types.

### DIFF
--- a/src/hal/components/counter.c
+++ b/src/hal/components/counter.c
@@ -65,19 +65,19 @@ typedef struct {
     unsigned char oldA;		/* previous value of phase A */
     unsigned char reset_on_index;
     unsigned char pad;		/* padding for alignment */
-    hal_s32_t *raw_count;	/* pin: raw binary count value */
-    hal_bit_t *phaseA;		/* quadrature input */
-    hal_bit_t *phaseZ;		/* index pulse input */
-    hal_bit_t *index_ena;	/* index enable input */
-    hal_bit_t *reset;		/* counter reset input */
-    hal_s32_t *count;		/* captured binary count value */
-    hal_float_t *pos;		/* scaled position (floating point) */
-    hal_float_t *vel;		/* scaled velocity (floating point) */
-    hal_float_t *pos_scale;	/* pin: scaling factor for pos */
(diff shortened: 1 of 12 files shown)


src/hal/components/counter.c
```diff
@@ -138,10 +138,10 @@ int rtapi_app_main(void)
...
 	counter_array[n].oldA = 0;
 	counter_array[n].reset_on_index = 0;
-	*(counter_array[n].raw_count) = 0;
-	*(counter_array[n].count) = 0;
-	*(counter_array[n].pos) = 0.0;
-	*(counter_array[n].pos_scale) = 1.0;
+	hal_set_si32(counter_array[n].raw_count, 0);
+	hal_set_si32(counter_array[n].count, 0);
+	hal_set_real(counter_array[n].pos, 0.0);
+	hal_set_real(counter_array[n].pos_scale, 1.0);
```
grandixximo, line 144: Default from `hal_pin_new_real`  below is 0.0, but then here it is set to `1.0`. Why not use the default field for 1.0 and drop the redundant init? Leaving it for later cleanup?

BsAtHome: When the pin/param creation and init assignments are separated, then I did not fix that. The problem is that you need to do a full call-tree analysis to ensure the values, which I did not do.
In this case, you are right, the inits here could be merged with the creation. However, the cost is low as this is done in the rtapi_app_main() startup. Leaving it for cleanup later would be fine, I think.

PR #4313, opened 2026-08-01
halcompile: support names= for count_function components

### LINKED ISSUE #4303: carousel: loadrt with names= fails (rtapi_app exits without becoming ready)
Loading carousel with an explicit instance name fails, while the default count-based load works.

Reproduced on current master (uspace sim, RIP build):

    loadrt carousel names=car

gives:

    waitpid failed rtapi_app carousel
    rtapi_app exited without becoming ready
    insmod for carousel failed, returned -1

Plain `loadrt carousel` loads fine.

carousel uses `option count_function` with its own get_count, so names-based instantiation may not be wired up for it. Not sure whether names= is supposed to work here; if it is not supported, an explicit error at load would be nicer than rtapi_app dying.

grandixximo:
Fixes #4303.

Components that use `option count_function` get their instance count from `get_count()` instead of a `count=` module parameter. The code generator also omitted the `names=` module parameter and parsing for these components, so `loadrt ... names=...` failed with:

    Unknown parameter `names=...'

Affected components: carousel, multiswitch, bldc, serport, pcl720, mesa_7i65.

Change `src/hal/utils/halcompile.g` so `count_function` components still declare and parse `names=`. If names are supplied, use them as instance prefixes; otherwise fall back to `get_count()` and the usual numeric names.

Tested:

- `loadrt carousel names=car`
- `loadrt carousel names=ca,cb pockets=8,12`
- `loadrt multiswitch names=ms cfg=4`
- `loadrt bldc names=bl cfg=h`
- `loadrt serport names=sp io=0x3f8`
- default `loadrt carousel`, `loadrt multiswitch cfg=4`, `loadrt bldc cfg=h`

### DIFF
--- a/src/hal/utils/halcompile.g
+++ b/src/hal/utils/halcompile.g
@@ -555,15 +555,22 @@ static int comp_id;
             print("    int arg = simple_strtol(argstr, NULL, 0);", file=f)
             print("    return export(prefix, arg);", file=f)
             print("}"   , file=f)
-        if not options.get("singleton") and not options.get("count_function") :
-            print("static int default_count=%s, count=0;" \
-                % options.get("default_count", 1), file=f)
-            if options.get("userspace"):
-                print("char *names[%d] = {NULL,};"%(MAX_USERSPACE_NAMES), file=f)
+        if not options.get("singleton"):
+            if not options.get("count_function"):
+                print("static int default_count=%s, count=0;" \
+                    % options.get("default_count", 1), file=f)
+                if options.get("userspace"):
+                    print("char *names[%d] = {NULL,};"%(MAX_USERSPACE_NAMES), file=f)
+                else:
+                    print("RTAPI_MP_INT(count, \"number of %s\");" % comp_name, file=f)
+                    print("char *names = \"\"; // comma separated names", file=f)
+                    print("RTAPI_MP_STRING(names, \"names of %s\");" % comp_name, file=f)
             else:
-                print("RTAPI_MP_INT(count, \"number of %s\");" % comp_name, file=f)
-                print("char *names = \"\"; // comma separated names", file=f)
-                print("RTAPI_MP_STRING(names, \"names of %s\");" % comp_name, file=f)
+                if options.get("userspace"):
+                    print("char *names[%d] = {NULL,};"%(MAX_USERSPACE_NAMES), file=f)
+                else:
+                    print("char *names = \"\"; // comma separated names", file=f)
+                    print("RTAPI_MP_STRING(names, \"names of %s\");" % comp_name, file=f)
         if has_personality:
             init1 = str(int(options.get('default_personality', 0)))
             init = ",".join([init1] * MAX_PERSONALITIES)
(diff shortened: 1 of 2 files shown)


BsAtHome:
But isn't the problem that with count_function you have a user-supplied count function? You actually do not know what you are supposed to have in names because you are not counting names. The user is counting and you have no control over that.

grandixximo:
@BsAtHome You are right. With `count_function` the count is determined by `get_count()`, so `names=` should only supply the instance names for that count, not replace the count.

I updated the patch to do exactly that:
- call `get_count()` first;
- parse the `names=` list and count the entries;
- if the length matches `get_count()`, export the instances with the supplied names;
- otherwise fail with a clear error like `names= count (2) does not match carousel count (1)`;
- if no `names=` is given, keep the existing numeric naming.

Tested on carousel, multiswitch, bldc, and serport with matching names, mismatching names, and default loads. The PR branch is force-pushed with the revised commit.

src/hal/utils/halcompile.g
```diff
@@ -608,15 +615,88 @@ static int comp_id;
...
+                print("        for(i=0; (i < max_names) && names[i] && r == 0; i++) {", file=f)
+                if has_personality:
+                    print("            r = export(names[i], i, p_value(\"%s\", names[i], i) );"%comp_name, file=f)
+                else:
+                    print("            r = export(names[i], i);", file=f)
+                print("            if(r != 0) break;", file=f)
+                print("        }", file=f)
+                print("    } else {", file=f)
+            else:
+                print("        char name_list[64][HAL_NAME_LEN+1];", file=f)
```
BsAtHome: Why is it necessary to use the stack for all names?

Would stream-processing be bad? Yes, you'd create stuff before you error out, but that is the case most of the times.

src/hal/utils/halcompile.g
```diff
@@ -608,15 +615,88 @@ static int comp_id;
                 print("    r = export(\"%s\", 0);" % \
                         to_hal(removeprefix(comp_name, "hal_")), file=f)
         elif options.get("count_function"):
-            print("    for(i=0; i<count; i++) {", file=f)
-            print("        char buf[HAL_NAME_LEN + 1];", file=f)
-            print("        rtapi_snprintf(buf, sizeof(buf), " \
+            print("    if(names[0]) {", file=f)
+            if options.get("userspace"):
```
BsAtHome: Are you sure we need to have this as a userspace component?

src/hal/utils/halcompile.g
grandixximo: I used `name_list[64][HAL_NAME_LEN+1]` so I could count the names and compare with `get_count()` before exporting any instance. Stream-processing in one pass would avoid the buffer but would export instances before discovering a count mismatch, then fail. Is that acceptable here?

src/hal/utils/halcompile.g
grandixximo: I added the userspace path for symmetry; no current component sets both `count_function` and `userspace`. Should I drop it and make that combination a generator error instead?

src/hal/utils/halcompile.g
BsAtHome: Yeah, this is kernel and you don't want to stress the stack.
Besides, this is not time critical, so you could just count the names before separating them out. You can even test their validity in that pass.

src/hal/utils/halcompile.g
BsAtHome: Userspace components generally are singletons. There is not much point in making multiple instances of a userspace component.
Why would you want multiple userspace instances? What is the use case?

grandixximo: @BsAtHome Hypothetical use case: a userspace driver that enumerates attached devices via `get_count()` and you want semantic names instead of `component.0`, `component.1`. For example a USB VFD driver that finds two spindles and you want `vfd.spindle_main` and `vfd.spindle_aux`, or an axis helper that exports `axis.x`, `axis.y`, `axis.z`.

Not that it is needed today; no current component uses both. But if the path is implemented and works, is there a strong reason to forbid it? Maintenance cost seems low since the generator already has to parse `names=` for other userspace components.

If we keep it, should `comp.adoc` get a note that `userspace` + `count_function` + `names=` is supported, or is it better left undocumented as an edge case?

andypugh:
I can definitely see an application for multiple instances of "thermistor" (and possibly hal_input but that isn't a .comp).
I am away from the LinuxCNC machines at the moment so don't actually know how they currently work for multiple instances.

BsAtHome:
But thermistor does a `FOR_ALL_INSTS() {}` loop. So it should already support multiple counts/names.

grandixximo:
@andypugh Before this PR, userspace components without `count_function` support `names=` but one load gives one instance. Multiple instances mean multiple processes (`loadusr -W thermistor names=bed`, `loadusr -W thermistor names=hotend`, ...). `hal_input` is a singleton C component. For modbus you would run one process per port.

This PR only affects userspace components that also set `option count_function`; `thermistor` and `hal_input` do not, so they are unchanged. The hypothetical use case is a single userspace process that derives instance count from hardware/config via `get_count()` and exports named instances (`vfd.spindle_main`, `vfd.spindle_aux`, or `modbus.port0`, `modbus.port1`) without spawning multiple processes. Whether that is worth supporting is the open question Bertho raised.

grandixximo:
@BsAtHome @andypugh You are right, thermistor already supports multiple instances via `count=` or `names=` because it does not use `count_function`. My "one load gives one instance" was wrong.

Corrected framing:

- Userspace components without `count_function` already support multiple instances via `count=` or `names=`.
- This PR only affects userspace components that use `count_function`; for those, `names=` previously failed with "Unknown parameter".
- The open question is whether `userspace` + `count_function` + `names=` is a combination worth keeping, or whether it should be a generator error.

grandixximo:
@BsAtHome @andypugh Rethinking the userspace + count_function path: it does not cleanly mirror either existing pattern.

Userspace without `count_function` parses `names=` into an array and exports one instance per supplied name. RT with `count_function` parses a comma string and requires the count to match `get_count()`. The hybrid I added keeps the array parse but adds a `get_count()` check, which fits neither model.

Options are:

1. Restructure userspace `names=` parsing to match RT (risky for existing userspace components).
2. Let `names=` override `get_count()` in userspace (Bertho already flagged this as wrong).
3. Make `userspace` + `count_function` a generator error and keep the RT-only fix.

I lean toward option 3. The actual issue #4303 is about RT components like `carousel`; the userspace path was speculative. Erroring avoids inconsistency and prevents users from writing ambiguous configs.

src/hal/utils/halcompile.g
grandixximo: Done

src/hal/utils/halcompile.g
grandixximo: Documented as not supported

grandixximo:
Removed the userspace `names=` support I added. `count_function` derives the instance count at load time and it can change across restarts, while HAL `names=` are fixed.When `count_function` is set, `names=` should error out, because it is dangerous to support setting it; users can already load multiple userspace components with `names=` when `count_function` is not used.

@BsAtHome  was correct all along, I just didn't see it...

Issue #4302, opened 2026-07-31
gantry: loadrt without personality crashes rtapi_app when function runs


grandixximo:
Loading gantry without a personality modparam and adding its function to a thread segfaults rtapi_app.

Reproduced on current master (uspace sim, RIP build):

    loadrt threads name1=t period1=1000000 fp1=0
    loadrt gantry
    addf gantry.0.read t
    start

rtapi_app dies as soon as the thread runs.

What I think happens: personality defaults to 0, so no joint pins are created, but FUNCTION(read) unconditionally accesses joint_home(0) (and again limit=joint_home(0)) before the `while (i < personality)` loop. With zero pins that reference is invalid.

Documented usage always passes personality=N, which works fine, so this only bites on a bare loadrt. Might be worth either erroring out at load when personality == 0, or guarding the first-joint access.

BsAtHome:
Yes, this may even be an error that is present in other components too. It wasn't introduced with the recent changes. This code has always been vulnerable.

The fix of all the components that require `personality > 0` at runtime should have a check somewhere, somehow. Not sure how this can be solved in a generic fashion.

grandixximo:
A generic fix rather than one by one sounds more elegant, I'll check if that's possible

grandixximo:
@BsAtHome, I took a look. I think a general fix is worth doing but narrower than I first expected. I would like your read on the shape before I draft a PR.

After my gantry fix, none of the other 13 `personality`-using comps crash on `personality=0`. They either loop safely on `personality` (silent no-op), clamp/default in `EXTRA_SETUP` (`lincurve` min 2, `limit_axis` min 2, `moveoff` =>3, `multiswitch` from `cfg`, `carousel` from `pockets`), or treat p=0 as a real working config (`bldc` VFD mode at `bldc.comp:418`). The blanket "reject p=0" rule is out because of `bldc`.

The real defect is in `halcompile`: array-sized pins `[MAX:count]` get only an upper-bound check (`if((cnt) > (lim)) ... -ENOSPC`), never a lower bound, so `count=0` exports nothing but the FUNCTION body happily dereferences NULL. The hand-written C components all do `if (howmany <= 0) return -EINVAL;` themselves, so the gap is just the generator.

Two things I would propose:

[...]

PR #4314, opened 2026-08-01
max31855: reject personality above 15 at load time

grandixximo:
Follow-up to the discussion on #4302.

max31855 sizes its sensor pins by `(personality & 0xf)` but indexes them with the raw value capped at 15. `personality=16` exports zero sensor pins while the function accesses 15 of them, dereferencing NULL pin pointers and killing rtapi_app as soon as the function runs in a thread.

Repro (uspace sim, RIP build):

    loadrt threads name1=t period1=100000 fp1=0
    loadrt max31855 personality=16
    addf max31855.0.bitbang-spi t
    start

Before: rtapi_app dies within one thread period of `start`. Silent in POSIX fallback: halrun still exits 0 and halcmd keeps answering from shm; only `pgrep -x rtapi_app` shows it.
After: `loadrt` fails with `max31855: personality must be at most 15 (use personality=N to set the number of sensors)`.

Also adds the agreed comp.adoc note: components with personality-sized pins should validate the value in EXTRA_SETUP and return -EINVAL, with an example.

### DIFF
--- a/docs/src/hal/comp.adoc
+++ b/docs/src/hal/comp.adoc
@@ -515,6 +515,34 @@ If the requested number of instances exceeds the number of allowed personalities
 personalities are assigned by indexing modulo the number of allowed personalities.
 A message is printed denoting such assignments.
 
+[NOTE]
+====
+If a component uses personality, then it should generally check its value.
+The value of personality is zero if the 'personality=N' argument is not provided to loadrt.
+Pins and params whose personality constraint evaluates to zero are not created, and their memory remains NULL.
+Unconditionally accessing such a pin or param dereferences a NULL pointer and crashes the realtime process.
+
+You can use a test in EXTRA_SETUP() to test the acceptable values of personality for your component.
+You should return -EINVAL if your conditions are not met.
+
+Example testing personality:
+
+[source,c]
+----
+EXTRA_SETUP(){
+    // Silence warnings for the unused arguments.
+    (void)prefix;
+    (void)extra_arg;
+    if (personality < 1) {
+        rtapi_print_msg(RTAPI_MSG_ERR,
+            "mycomp: personality must be >= 1 (use personality=N to set the number of channels)\n");
+        return -EINVAL;
+    }
+    return 0;
+}
+----
+====
+
 == Examples
 
 === constant
--- a/src/hal/components/max31855.comp
+++ b/src/hal/components/max31855.comp
@@ -43,6 +43,7 @@ variable unsigned data_frame [15];
 variable unsigned state = 1;
 
 option period no;
+option extra_setup yes;
 function bitbang_spi;
 license "GPL";
 author "Joseph Calderon";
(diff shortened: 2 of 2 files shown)


docs/src/hal/comp.adoc
```diff
@@ -515,6 +515,26 @@ If the requested number of instances exceeds the number of allowed personalities
 personalities are assigned by indexing modulo the number of allowed personalities.
 A message is printed denoting such assignments.
 
+[NOTE]
+Pins and parameters whose "if condition" or "[maxsize : condsize]" evaluates to false or zero are not exported,
+and the instance memory for them is left zeroed, so accessing such a pin from a function body dereferences a NULL pointer and crashes the realtime process.
+A component that uses 'personality' to size its pins or parameters should validate the value in EXTRA_SETUP()
+and return -EINVAL with a clear message for configurations that would export no pins:
```
BsAtHome: These are very long sentences. Maybe split it up in smaller sentences.
It also ends with a colon?

docs/src/hal/comp.adoc
```diff
@@ -515,6 +515,26 @@ If the requested number of instances exceeds the number of allowed personalities
...
+
+[source,c]
+----
+EXTRA_SETUP(){
+    (void)prefix;
+    (void)extra_arg;
```
BsAtHome, line 535: These may need a comment in the example why they are present.

docs/src/hal/comp.adoc
grandixximo: Split in shorter sentences.
The sentence was preamble to the code, the colon was the glue, made it more specific.

docs/src/hal/comp.adoc
grandixximo: commented

docs/src/hal/comp.adoc
grandixximo: See if it reads better now, or did I overdo it?

BsAtHome:
What about:
```adoc
[NOTE]
----
If a component uses personality, then it should generally check its value.
The value of personality is zero if the 'personality=N' argument is not provided to loadrt.
Any unconditional access to pins or params created by personality constraints, when zero, causes a crash.
Pins and params are not created when personality is zero and its memory remains NULL.
Accessing a not created pin or param accesses a NULL pointer, which inevitably leads to a crash.

You can use a test in EXTRA_SETUP() to test the acceptable values of personality for your component.
You should return -EINVAL if your conditions are not met.

Example testing personality:
[source,c]
...

----
```

BsAtHome:
A real bikeshed moment :-)

grandixximo:
Continuing the bikeshed painting: I kept your first two sentences and the remedy paragraph verbatim, and merged the three middle ones into a cause-then-effect pair, since they said "crash" twice and "a not created pin" read clunky to me. I feel it reads better like this, but your version is acceptable to me too if you prefer it.

```
[NOTE]
====
If a component uses personality, then it should generally check its value.
The value of personality is zero if the 'personality=N' argument is not provided to loadrt.
Pins and params whose personality constraint evaluates to zero are not created, and their memory remains NULL.
Unconditionally accessing such a pin or param dereferences a NULL pointer and crashes the realtime process.

You can use a test in EXTRA_SETUP() to test the acceptable values of personality for your component.
You should return -EINVAL if your conditions are not met.

Example testing personality:

[source,c]
----
```

BsAtHome:
I like your color too.

grandixximo:
Ok, it's already in, ready to merge

PR #4315, opened 2026-08-01
CI: Retry rtai downloads if they fail (often connection resets)

BsAtHome:
There are regular "connection reset" errors when downloading the RTAI kernel debs from github's own servers. This interrupts the build and fails CI.

This PR adds retry options to cURL so that transient errors are ignored. It will retry up to 5 times with 2 second delay in between. Errors that are more persistent will still make CI fail and that should be alright.

### DIFF
--- a/.github/scripts/install-rtai.sh
+++ b/.github/scripts/install-rtai.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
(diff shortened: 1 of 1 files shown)


grandixximo:
I never actually notices the errors, it just seems frozen sometimes, let's try and see if it improves...

PR #4317, opened 2026-08-01
hal: Update halmodule to getter/setter and the query API

BsAtHome:
This moves the current halmodule to getter/setter and uses only the HAL query API.

We cannot move to a new C++ halmodule implementation before everything is in place. We need to retain the infrastructure so we can move forward.

The new port pin creation is enabled (hal_pin_new_port()) using a slightly different signature than what it will become. This is done for compatibility sake because the hal_port_t is already in use and can only be fixed when we break the API. So that will have to wait. In the meantime, we use an intermediate.

Halmodule no longer accesses hal_priv.h. Quite some lines could be removed by using the query API. Also, the new API is much simpler for the halitem and removes the old pin/param/type/direction union stuff.

It is no longer required to create a component before accessing the generic pin/signal access methods. Halmodule will call `hal_lib_init()` when imported and register a `Py_AtExit(hal_lib_exit)` to remove the share memory. This also makes the check in hal_lib effective when components are left dangling. At exit it will specify which components were never exited.

You can now do `hal.get_value('some.pin.name')` immediately after importing the module. The same goes for `set_p`/`set_s`. Also, functions `get_p`/`get_s` are added. The get_value would test pins/params/signals, whereas other utilities have a split. This is now possible. An added check tests this and also uses the threadbeat for cycle advancement testing (also has a timeout, but that is to guard infinity).


src/hal/halmodule.cc
```diff
@@ -1245,63 +1172,72 @@ static PyObject * pyhal_pin_new(halitem * pin, const char * name) {
...
+    hal_query_t q = {};
+    if(!PyArg_ParseTuple(args, "s", &q.name)) return NULL;
+
+    q.qtype = HAL_QTYPE_PIN; // Only query a pin
+    int rv = hal_get_p(&q, NULL, NULL);
+    if(0 == rv) {
+        // Success reading. See if it has a driver.
+        if(!q.pp.signal) {
+            // No signal, it may be its own writer?
+            return PyBool_FromLong(q.pp.dir == HAL_OUT);
```
grandixximo: Should this be False? Master returned False for any unconnected pin; this returns True for an unconnected HAL_OUT pin, so axis.py forget() keeps spindle widgets visible and qtvcp isDriven() flips for unconnected pins. C++ hal.hh pin_has_writer still has the old semantics.

BsAtHome: You are right. Fixed to match previous behaviour.

PR #4319, opened 2026-08-01
hal: Update the remaining user component VFDs to getter/setter

BsAtHome:
These are the remaining VFD user components. They are not especially complex, but they do have many lines changed.

Unfortunately, you need to have the hardware and modbus setup to test these VFD drivers. Not something many of us have at hand. So, we need to have some eyes on the code to see if anything got missed.

Note, a lot of the content is badly indented. A few places were corrected (like parts of the hal structs), but most of it needs to be done another time. Altering too much will just confuse the blame and make tracking any potential problems hard. So, mostly left as is.

### DIFF
--- a/src/hal/user_comps/vfdb_vfd/vfdb_vfd.c
+++ b/src/hal/user_comps/vfdb_vfd/vfdb_vfd.c
@@ -112,37 +112,37 @@
 
 /* HAL data struct */
 typedef struct {
-    hal_s32_t   *error_code;
-    hal_s32_t 	*status;
-    hal_float_t	*freq_cmd;	// frequency command
-    hal_float_t	*freq_out;	// actual output frequency
-    hal_float_t	*output_volt;	// output voltage
-    hal_float_t	*RPM;
-    hal_float_t *RPS;
-    hal_float_t	*torque_ratio;
-    hal_float_t	*output_current;
-    hal_float_t *max_rpm;	// calculated based on VFD max frequency setup parameter
-    hal_bit_t	*at_speed;	// when drive freq_cmd == freq_out and running
-    hal_bit_t	*is_stopped;	// when drive freq out is 0
-    hal_bit_t	*is_e_stopped;	// true if emergency stop status set in 0xFD00
-    hal_bit_t	*modbus_ok;	// the last MODBUS_OK transactions returned successfully
-    hal_float_t	*speed_command;	// speed command input
+    hal_sint_t error_code;
+    hal_sint_t status;
+    hal_real_t freq_cmd;        // frequency command
+    hal_real_t freq_out;        // actual output frequency
+    hal_real_t output_volt;     // output voltage
+    hal_real_t RPM;
+    hal_real_t RPS;
+    hal_real_t torque_ratio;
+    hal_real_t output_current;
+    hal_real_t max_rpm;         // calculated based on VFD max frequency setup parameter
+    hal_bool_t at_speed;        // when drive freq_cmd == freq_out and running
+    hal_bool_t is_stopped;      // when drive freq out is 0
+    hal_bool_t is_e_stopped;	// true if emergency stop status set in 0xFD00
+    hal_bool_t modbus_ok;	    // the last MODBUS_OK transactions returned successfully
+    hal_real_t speed_command;   // speed command input
 
-    hal_bit_t	*spindle_on;	// spindle 1=on, 0=off
-    // hal_bit_t	*err_reset;	// reset errors when 1  - set fault reset bit in 0xFA00
-    hal_bit_t	*jog_mode;	// termed 'jog-run' in manual - might be useful for spindle positioning
-    hal_s32_t	*errorcount;    // number of failed Modbus transactions - hints at logical errors
+    hal_bool_t spindle_on;      // spindle 1=on, 0=off
+    // hal_bool_t	err_reset;	// reset errors when 1  - set fault reset bit in 0xFA00
+    hal_bool_t jog_mode;        // termed 'jog-run' in manual - might be useful for spindle positioning
+    hal_sint_t errorcount;      // number of failed Modbus transactions - hints at logical errors
 
-    hal_float_t	looptime;
-    hal_float_t	speed_tolerance; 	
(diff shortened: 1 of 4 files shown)


src/hal/user_comps/vfdb_vfd/vfdb_vfd.c
```diff
@@ -768,16 +755,13 @@ int main(int argc, char **argv)
...
-            if (p->haldata->looptime < 0.001) p->haldata->looptime = 0.001;
-            if (p->haldata->looptime > 2.0) p->haldata->looptime = 2.0;
-            loop_timespec.tv_sec = (time_t)(p->haldata->looptime);
-            loop_timespec.tv_nsec = (long)((p->haldata->looptime - loop_timespec.tv_sec) * 1000000000l);
+            rtapi_real looptime = hal_get_real(p->haldata->looptime);
+            if (looptime < 0.001) looptime = hal_set_real(p->haldata->looptime, 0.001);
+            if (looptime > 2.0)   looptime = hal_set_real(p->haldata->looptime, 2.0);
+            loop_timespec.tv_sec = (time_t)looptime;
+            loop_timespec.tv_nsec = (long)((looptime - loop_timespec.tv_sec) * 1000000000l);
             if (!p->haldata->max_speed)
```
grandixximo: `p->haldata->max_speed` is a HAL handle (opaque pointer), not the pin value. It is always non-NULL after pin creation, so this test is always false and the loop always sleeps; max-speed never takes effect. Use `if (!hal_get_bool(p->haldata->max_speed))`.

src/hal/user_comps/vfs11_vfd/vfs11_vfd.c
```diff
@@ -1047,16 +1043,13 @@ int main(int argc, char **argv)
...
-	    if (p->haldata->looptime < 0.001) p->haldata->looptime = 0.001;
-	    if (p->haldata->looptime > 2.0) p->haldata->looptime = 2.0;
-	    loop_timespec.tv_sec = (time_t)(p->haldata->looptime);
-	    loop_timespec.tv_nsec = (long)((p->haldata->looptime - loop_timespec.tv_sec) * 1000000000l);
+	    rtapi_real looptime = hal_get_real(p->haldata->looptime);
+	    if (looptime < 0.001) looptime = hal_set_real(p->haldata->looptime, 0.001);
+	    if (looptime > 2.0)   looptime = hal_set_real(p->haldata->looptime, 2.0);
+	    loop_timespec.tv_sec = (time_t)looptime;
+	    loop_timespec.tv_nsec = (long)((looptime - loop_timespec.tv_sec) * 1000000000l);
 	    if (!p->haldata->max_speed)
```
grandixximo: Same handle-vs-value issue as in vfdb_vfd. Use `if (!hal_get_bool(p->haldata->max_speed))`.

grandixximo (commented):
I went through all four components and built the branch clean. I also ran the binaries with --help and exercised a minimal repro for the handle-vs-value cases below.

There are three places where a HAL handle is tested directly instead of the pin value. `hal_bool_t` is an opaque pointer into shared memory; after a successful `hal_pin_new_bool` it is always non-NULL, so `if (!handle)` is always false and `if (handle)` is always true, regardless of the pin state. These need `hal_get_bool()`.

The rest looks like the expected mechanical migration. I agree with leaving the existing indentation alone; reindenting now would only pollute the blame.

src/hal/user_comps/vfdb_vfd/vfdb_vfd.c
```diff
@@ -578,34 +569,34 @@ int read_data(modbus_t *ctx, haldata_t *haldata, param_pointer p)
...
+    PIN(hal_pin_new_real(id, HAL_OUT, &(h->RPM), 0.0, "%s.motor-RPM", name));
+    PIN(hal_pin_new_real(id, HAL_OUT, &(h->RPS), 0.0, "%s.motor-RPS", name));
+    PIN(hal_pin_new_real(id, HAL_OUT, &(h->output_volt), 0.0, "%s.output-voltage", name));
+    PIN(hal_pin_new_real(id, HAL_IN, &(h->speed_command), 0.0, "%s.speed-command", name));
+    PIN(hal_pin_new_bool(id, HAL_IN, &(h->spindle_on), 0, "%s.spindle-on", name));
+    PIN(hal_pin_new_si32(id, HAL_OUT, &(h->error_code), 0, "%s.error-code", name));
+    PIN(hal_pin_new_si32(id, HAL_OUT, &(h->status), 0, "%s.status", name));
+    PIN(hal_pin_new_bool(id, HAL_IN, &(h->max_speed), 0, "%s.max-speed", name));
+    PIN(hal_pin_new_si32(id, HAL_OUT, &(h->errorcount), 0, "%s.error-count", name));
+    PIN(hal_pin_new_real(id,HAL_OUT, &(h->upper_limit_hz), 0.0, "%s.frequency-limit", name));
```
grandixximo: missing space after the comma: `hal_pin_new_real(id,HAL_OUT, ...`.

src/hal/user_comps/vfs11_vfd/vfs11_vfd.c
grandixximo: Missed one. Use `if (hal_get_bool(haldata->acc_dec_pattern))`.

src/hal/user_comps/vfdb_vfd/vfdb_vfd.c
BsAtHome: fixed.

src/hal/user_comps/vfs11_vfd/vfs11_vfd.c
BsAtHome: fixed.

src/hal/user_comps/vfdb_vfd/vfdb_vfd.c
BsAtHome: yeah...

src/hal/user_comps/vfs11_vfd/vfs11_vfd.c
BsAtHome: The original is bad here too. It should have read `*haldata->acc_dec_pattern`.
Fixed anyway because it was obviously wrong.

PR #4320, opened 2026-08-01
hal: Update the last uspace drivers to getter/setter

BsAtHome:
These are the last drivers (except hostmot2) that are compiled in uspace. The rest is rtai/kernel only afaict,which will come later.

Both hal_bb_gpio.c and hal_evoreg.c drivers are straight forward and not too big a change.
Apart from that, I think that the hard-coded number in the hal_evoreg.c pin-array sizes are off. The comments don't match the numbers and the implementation does neither. Not going to touch this can of worms. Also, there are several kernel-types still in the code. No sure anyone has tried this driver for some time.

The gm.h file has the hardware memory layout structure. All its members are volatile because this is a PCI MMIO. The old hal types cannot be (ab)used anymore, so the normal equivalent is used.
The hal_gm.c driver is a bit special because it abuses hal memory and pin target memory. It originally pointed to local parameters, but these do not work like that anymore. However, an overlay union was created to support taking the address of the target and pretend it to be a new hal type (it is all in HAL memory anyway). This is very ugly and non-portable, but it is the quickest way to fix the problem at hand. Otherwise the driver would need considerable work to be fixed. This hack is used in cardMgr_t and stepgen_t. Furthermore, it is a big driver with a some inefficient code. I really hope we can find anyone who can test this. The driver is "only" 14 years old with the last update 10 years ago...

The hal_parport.c driver uses parameters but does not always allocate them. It depends on the mode of the port. It has been verified that the allocations match the run-time conditions and there should be no crash.

The hal_ppmc.c driver is big and inefficient at times. Fixing that was harder than to let it be. At least it didn't get worse.

### DIFF
--- a/src/hal/drivers/hal_evoreg.c
+++ b/src/hal/drivers/hal_evoreg.c
@@ -86,14 +86,14 @@ RTAPI_MP_STRING(cfg, "config string"); */
 */
 
 typedef struct {
-	void *io_base;
-        hal_float_t *dac_out[3];  /* ptrs for dac output */
-	hal_float_t *position[3];	   /* ptrs for encoder input */
-	hal_bit_t *digital_in[47];    /* ptrs for digital input pins 0 - 45 */
-        hal_bit_t *digital_out[25];    /* ptrs for digital output pins 0 - 20 */
-        __u16 raw_counts_old[3];
-        __s32 counts[3];
-        hal_float_t pos_scale;         /*! \todo scale for position command FIXME should be one per axis */
+    void *io_base;
+    hal_real_t dac_out[3];      /* ptrs for dac output */
+    hal_real_t position[3];     /* ptrs for encoder input */
+    hal_bool_t digital_in[47];  /* ptrs for digital input pins 0 - 45 */
+    hal_bool_t digital_out[25]; /* ptrs for digital output pins 0 - 20 */
+    rtapi_u16 raw_counts_old[3];
+    rtapi_s32 counts[3];
+    hal_real_t pos_scale;       /*! \todo scale for position command FIXME should be one per axis */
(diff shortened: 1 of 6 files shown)


src/hal/drivers/hal_evoreg.c
```diff
@@ -327,7 +328,7 @@ static void update_port(void *arg, long period)
      tmp = 0x0;
      mask = 0x01;
      for (pin=0; pin < 16; pin++) {
-        if (port->digital_out[pin]) {
+        if (hal_get_bool(port->digital_out[pin])) {
         tmp |= mask;
         mask <<= 1;
         }
```
grandixximo, line 333: `mask <<= 1;` must be outside the `if`, otherwise FALSE pins don't advance the mask and the output bits collapse (pins {1,3} set writes 0x0003 instead of 0x000a). Same for the loop below.

grandixximo (commented):
Builds clean; hal_parport loads and runs with the new API, the other drivers fail gracefully without hardware. The hal_gm fake-target hack is safe: disable stays 0 forever and the handles are read-only.

One finding inline in hal_evoreg.c: `mask <<= 1` sits inside the `if` braces, so FALSE pins don't advance the mask and output bits collapse (pins {1,3} write 0x0003 instead of 0x000a). In master this was hidden by the always-true pointer test. Easy fix while these lines are being touched anyway.

src/hal/drivers/hal_evoreg.c
BsAtHome: Interesting to find bugs while converting. Fixed anyway.

grandixximo: Thanks

PR #4321, opened 2026-08-02
Fix homing bug when trying to home while homing is in progress (2.9 backport)

### LINKED ISSUE #4316: Gmoccapy: Homing second axis in parallel leads to machine crash
Linuxcnc 2.9.8
Combined switches for limit and reference.

Steps:

1. Press a button to home one axis
2. Press the home button for another axis while the homing of the first axis isn't finished yet.

Result: The first started axis is moving beyond the ref/limit switch which drives machine into its mechanical limits. The second axis is homed correctly.

grandixximo:
Backport of 9d5570ca84 (#3664) to 2.9.

Fixes #4316, which is the same bug as #3314: starting to home a second axis while a homing sequence is already in progress breaks the running homing, and the first joint keeps moving past its ref/limit switch into the mechanical limits.

The command.c change rejects EMCMOT_JOINT_HOME while homing is active, which is the machine-safety part. The axis.py and gmoccapy.py changes guard the GUIs so wait_complete() does not freeze the UI on the rejected command.

Cherry-picked cleanly onto 2.9, no conflicts.

### DIFF
--- a/src/emc/motion/command.c
+++ b/src/emc/motion/command.c
@@ -1381,7 +1381,10 @@ void emcmotCommandHandler_locked(void *arg, long servo_period)
 	                   joint_num);
                 return;
 	    }
-
+	    if ( get_homing_is_active() ) {
+	        reportError("Homing not possible until current homing process is finished.\n");
+	        return;
+	    }
 	    if (!GET_MOTION_ENABLE_FLAG()) {
 		break;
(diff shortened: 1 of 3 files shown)


hansu:
Nice, let me test on the machine...

hansu:
Thanks!

PR #4323, opened 2026-08-02
hm2_eth: Fix invalid soft error

### LINKED ISSUE #4322: hm2_modbus increments hm2_7i96s.0.packet-error-total during LinuxCNC startup
As requested by @hdiethelm , who asked me in [#4236](https://github.com/LinuxCNC/linuxcnc/issues/4236#issuecomment-5048444277):

> "Can you create an issue for this or is there already one?"

I am creating this dedicated issue for the problem.

## Environment

- LinuxCNC: current `master` branch
- Mesa: 7i96S with PktUARTTX
- Reproducibility: **100%** (every startup)

## Important

This issue is **not related to the realtime kernel**.
I have reproduced it with both:
- PREEMPT_RT
- Xenomai 4

@pcw-mesa  has confirmed the same behavior.
https://github.com/LinuxCNC/linuxcnc/issues/4236#issuecomment-4969921132

## Steps to reproduce

In the HAL file, load `hm2_modbus`, for example:

```
loadrt hm2_modbus ports="hm2_7i96s.0.pktuart.0" mbccbs="/home/zdenek/linuxcnc/configs/DedaCNC-012/modbus/h100.mbccb" debug=3
```

Start LinuxCNC.

## Expected behavior

After startup:

hm2_7i96s.0.packet-error-total = 0

## Actual behavior

After startup:

hm2_7i96s.0.packet-error-total = 1

## Additional observations

If the loadrt hm2_modbus ... line is commented out, then after LinuxCNC starts:

hm2_7i96s.0.packet-error-total = 0

Simply loading hm2_modbus is enough to trigger the packet error.

I also disconnected the RS485 wiring completely and the behavior did not change, so the issue does not appear to depend on the physical RS485 connection.

BsAtHome:
The hm2_eth driver registers a soft error when using hm2_modbus. This was tracked down to a problem in the hm2_eth driver when non-queued writes precede the first queued read without having had a queued write.

Full details of the problem and analysis in #4322.

The solution uses a flag to ensure that a queued write has set the proper data in the board before its read back value is used in the check.

Fixes #4322.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/hm2_eth.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_eth.c
@@ -1023,10 +1023,11 @@ static int hm2_eth_send_queued_reads(hm2_lowlevel_io_t *this) {
     LBP16_INIT_PACKET4(*(lbp16_cmd_addr*)(board->read_packet_ptr), CMD_READ_COMM_CTRL_ADDR16(1), 0x8);
     board->read_packet_ptr += sizeof(lbp16_cmd_addr);
     board->queue_reads[board->queue_reads_count].buffer = &board->rxudpcount;
-    board->queue_reads[board->queue_reads_count].size = 2;
+    static_assert(sizeof(board->rxudpcount)==2, "board->rxudpcount must be one 16-bit word");
+    board->queue_reads[board->queue_reads_count].size = sizeof(board->rxudpcount);
     board->queue_reads[board->queue_reads_count].from = board->queue_buff_size;
     board->queue_reads_count++;
-    board->queue_buff_size += 2;
+    board->queue_buff_size += sizeof(board->rxudpcount);
     
     board->read_cnt++;
     // write then read back space 4 scratch register at 0010 to verify we got the right receive packet
@@ -1038,11 +1039,12 @@ static int hm2_eth_send_queued_reads(hm2_lowlevel_io_t *this) {
 
     LBP16_INIT_PACKET4(*(lbp16_cmd_addr*)(board->read_packet_ptr), CMD_READ_TIMER_ADDR16_INCR(4), 0x10);
     board->read_packet_ptr += sizeof(lbp16_cmd_addr);
-    board->queue_reads[board->queue_reads_count].buffer = &board->confirm_read_cnt;
-    board->queue_reads[board->queue_reads_count].size = 8;
+    board->queue_reads[board->queue_reads_count].buffer = &board->confirm_rw_cnt;
+    static_assert(sizeof(board->confirm_rw_cnt)==8, "board->confirm_rw_cnt must be two 32-bit words");
+    board->queue_reads[board->queue_reads_count].size = sizeof(board->confirm_rw_cnt);
     board->queue_reads[board->queue_reads_count].from = board->queue_buff_size;
     board->queue_reads_count++;
-    board->queue_buff_size += 8;
+    board->queue_buff_size += sizeof(board->confirm_rw_cnt);
 
     send = eth_socket_send(board->sockfd, (void*) &board->read_packet, board->read_packet_ptr - board->read_packet, 0);
     if(send < 0) {
@@ -1125,15 +1127,16 @@ do_recv_packet:
         memcpy(board->queue_reads[i].buffer, &tmp_buffer[board->queue_reads[i].from], board->queue_reads[i].size);
     }
 
-    if(board->confirm_read_cnt != board->read_cnt && t2 < read_deadline)
+    if(board->confirm_rw_cnt.read_cnt != board->read_cnt && t2 < read_deadline)
         goto do_recv_packet;
 
     hm2_eth_reset_queued_reads(board);
 
     int result = 1;
-    // (this means that one in 2^32 lost writes will not be diagnosed,
-    // each time board->write_cnt overflows)
-    if(board->write_cnt && board->write_cnt != board->confirm_write_cnt) {
+    // The has_written_cnt is necessary because the thread cycle uses first
+    // read, then write. If we didn't check, then we'd be using uninitialized
+    // data read back from the board.
+    if(board->has_written_cnt && board->write_cnt != board->confirm_rw_cnt.write_cnt) {
         result = record_soft_error(board);
     } else {
         decrement_soft_error(board);
@@ -1229,6 +1232,7 @@ static int hm2_eth_send_queued_writes(hm2_lowlevel_io_t *this) {
     static_assert(sizeof(board->write_cnt)==4, "board->write_cnt must be 4 bytes");
     memcpy(board->write_packet_ptr, &board->write_cnt, sizeof(board->write_cnt));
     board->write_packet_ptr += sizeof(board->write_cnt);
+    board->has_written_cnt = 1;
     
     t0 = rtapi_get_time();
     send = eth_socket_send(board->sockfd, (void*) &board->write_packet, board->write_packet_ptr - board->write_packet, 0);
@@ -1815,6 +1819,7 @@ int rtapi_app_main(void) {
             continue;
         } 
         boards[i].read_cnt = boards[i].write_cnt = 0;
+        boards[i].has_written_cnt = 0;
         int *added = kvlist_lookup(&ifnames, ifptr);
         if(!added)
             goto error;
--- a/src/hal/drivers/mesa-hostmot2/hm2_eth.h
+++ b/src/hal/drivers/mesa-hostmot2/hm2_eth.h
@@ -51,9 +51,17 @@ typedef struct {
     rtapi_u8 write_packet[1400];
     rtapi_u8 *write_packet_ptr;
     uint32_t read_cnt, write_cnt;
-    // these two fields must be kept together, they're read by a single
-    // read-request
-    uint32_t confirm_read_cnt, confirm_write_cnt;
+    struct {
+        // These two fields must be kept together. They are read by a single
+        // queued read-request and retrieve the above read_cnt and write_cnt
+        // added by hm2_eth_send_queued_reads() and hm2_eth_send_queued_writes().
+        // The values of the two should match or we know that the send/recv
+        // packets are out of sync.
+        uint32_t read_cnt;
+        uint32_t write_cnt;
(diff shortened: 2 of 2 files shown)


src/hal/drivers/mesa-hostmot2/hm2_eth.c
```diff
@@ -1038,8 +1038,8 @@ static int hm2_eth_send_queued_reads(hm2_lowlevel_io_t *this) {
 
     LBP16_INIT_PACKET4(*(lbp16_cmd_addr*)(board->read_packet_ptr), CMD_READ_TIMER_ADDR16_INCR(4), 0x10);
     board->read_packet_ptr += sizeof(lbp16_cmd_addr);
-    board->queue_reads[board->queue_reads_count].buffer = &board->confirm_read_cnt;
-    board->queue_reads[board->queue_reads_count].size = 8;
+    board->queue_reads[board->queue_reads_count].buffer = &board->confirm_rw_cnt;
+    board->queue_reads[board->queue_reads_count].size = sizeof(board->confirm_rw_cnt);
     board->queue_reads[board->queue_reads_count].from = board->queue_buff_size;
     board->queue_reads_count++;
     board->queue_buff_size += 8;
```
hdiethelm: Probably also sizeof(board->confirm_rw_cnt)?

src/hal/drivers/mesa-hostmot2/hm2_eth.c
hdiethelm: If you are on it: sizeof()

hdiethelm: sizeof()

src/hal/drivers/mesa-hostmot2/hm2_eth.h
```diff
@@ -53,7 +53,12 @@ typedef struct {
     uint32_t read_cnt, write_cnt;
     // these two fields must be kept together, they're read by a single
     // read-request
```
hdiethelm: This comment is not needed any more, now that it is a proper struct.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
BsAtHome: Yes, indeed.

Also, the size must be 8 bytes (two 32-bit words) or the LBP16_INIT_PACKET4() line would be wrong. Added a static assert like the one above.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
BsAtHome: Same fix.

BsAtHome: same fix.

src/hal/drivers/mesa-hostmot2/hm2_eth.h
BsAtHome: I think it is still relevant to explain why this is a struct in a struct.

hdiethelm: In this case, I would move it inside the struct.

hdiethelm:
If it helps, i can test it tomorrow but without modbus.
Otherwhise it looks good, thanks for fixing!

BsAtHome:
> If it helps, i can test it tomorrow but without modbus. Otherwhise it looks good, thanks for fixing!

Thanks, already tested :-)

I have a RPi5 + 7i96s on my desk with a modbus device attached. I wouldn't have been able to reproduce the problem, let alone track the error if I hadn't.

<sub>BTW, hm2_modbus driver author talking here ;-)</sub>

src/hal/drivers/mesa-hostmot2/hm2_eth.h
BsAtHome: Sure. Also expanded the comment a bit.

src/hal/drivers/mesa-hostmot2/hm2_eth.h
```diff
@@ -51,9 +51,17 @@ typedef struct {
...
     rtapi_u8 *write_packet_ptr;
     uint32_t read_cnt, write_cnt;
-    // these two fields must be kept together, they're read by a single
-    // read-request
-    uint32_t confirm_read_cnt, confirm_write_cnt;
+    struct {
+        // These two fields must be kept together. They are read by a single
+        // queued read-request and retrieve the above read_cnt and write_cnt
+        // added by hm2_eth_send_queued_reads() and hm2_eth_send_queued_writes().
+        // The values of the two should match or we know that the send/revc
```
grandixximo: typo 
`"send/revc"` => `"send/recv"`.

src/hal/drivers/mesa-hostmot2/hm2_eth.h
```diff
@@ -51,9 +51,17 @@ typedef struct {
...
+        // packets are out of sync.
+        uint32_t read_cnt;
+        uint32_t write_cnt;
+    } confirm_rw_cnt;
+    // Set when a queued write has set write_cnt in the board
+    int has_written_cnt;
```
grandixximo, line 64: Could be `bool` .
And just to make it a `true` bike-shed comment (pun intended):
`confirm_write_cnt_valid `or `queued_write_sent` read better to me ;-)

grandixximo (commented):
Something else I also noticed, AFAIU the driver counts every write it sends to the board (`write_cnt`). But only one kind of write (the queued per-cycle one) also stores that count on the board, where it can be read back and compared.

So the check "does the board's copy of the count match mine?" breaks if any other kind of write happens in between: the driver's count goes up by one, the board's copy doesn't, mismatch, false error.

That other kind of write can in theory happen mid-run, e.g. a userspace tool poking the board while threads are active. Then you'd get one bogus soft error even with this PR applied.

A possible fix (if it's even worth fixing) is to keep two counters, count only queued writes in `write_cnt`, since that's the only count the board ever sees, and if necessary a second `direct_write_cnt` that counts the non-queued writes for debug/logging.

If agreed this is a real issue, probably better as a follow up, but leaving the decision to you.

hdiethelm:
> > If it helps, i can test it tomorrow but without modbus. Otherwhise it looks good, thanks for fixing!
> 
> Thanks, already tested :-)
> 
> I have a RPi5 + 7i96s on my desk with a modbus device attached. I wouldn't have been able to reproduce the problem, let alone track the error if I hadn't.
> 
> BTW, hm2_modbus driver author talking here ;-)

Ah yes, now I see it in the header. Do RPi5 work well with real time? I have one lying around. I could try to get the Xenomai4 kernel working on it... ;-) The only thing missing here is a second mesa card. The only one I have is mounted deep inside the machine. Let's see, would be anyway nice for developing and as a spare.

> That other kind of write can in theory happen mid-run, e.g. a userspace tool poking the board while threads are active. Then you'd get one bogus soft error even with this PR applied.

It would be probably be a bad idea to throw some normal reads/writes between the real time cyclic queued read/writes.

grandixximo:
> It would be probably be a bad idea to throw some normal reads/writes between the real time cyclic queued read/writes.

The current bogus errors certainly deterr the idea, silencing them completely would seem like a free pass, and I'm not advocating for that, we could leave them in consciously. I only see the current wording as possibly confusing during debug, a separate count with a clear deterr warning seems more appropriate, but let's see what Bertho thinks...

src/hal/drivers/mesa-hostmot2/hm2_eth.h
BsAtHome: Darn,... fixed.

src/hal/drivers/mesa-hostmot2/hm2_eth.h
BsAtHome: Yes , with the emphasis on could.
Making it bool would not change the read/write behavior or alignment.

WRT naming, my preference is blue. ;-)

src/hal/drivers/mesa-hostmot2/hm2_eth.h
grandixximo: That's one way to say good morning 🤭

src/hal/drivers/mesa-hostmot2/hm2_eth.h
grandixximo: I know, just enjoying the friendly banter...

BsAtHome:
> Something else I also noticed, AFAIU the driver counts every write it sends to the board (`write_cnt`). But only one kind of write (the queued per-cycle one) also stores that count on the board, where it can be read back and compared.

Yes, it does. Same for the synchronous read, but that you don't detect because it is rewritten in the queued read, which is checked in the second part of the read.

> So the check "does the board's copy of the count match mine?" breaks if any other kind of write happens in between: the driver's count goes up by one, the board's copy doesn't, mismatch, false error.

> It would be probably be a bad idea to throw some normal reads/writes between the real time cyclic queued read/writes.

If you intersperse queued operation with synchronous operation, then you cannot properly run realtime anymore. Doing so would risk killing your timing (even hm2_modbus uses queued recovery).

The only way you can do so is if a registered hm2 driver calls synchronous functions, and there are (luckily) not many. Normal addf read/write for hm2_eth are the queue wrappers.

If you have code that actually does synchronous calls while in the normal thread cycle, then I think that marking packet errors is not the most obvious thing to do. OTOH, I would consider using a synchronous write while the the thread cycle an error. So, removing it from unqueued write would hide the scenario and keeping it results in a weird error.

BsAtHome:
> Do RPi5 work well with real time? I have one lying around. I could try to get the Xenomai4 kernel working on it... ;-) 

These RPi5 boards are really fast. Also the RPi4 (and RPi3) have been successfully used to run RT.
For systems that do not require huge other computing power or other interfacing methods, like 95% of the time, I'd actually suggest a RPi5. It surely beats a normal PC on power and price. The RPi5/7i96s combi + modbus digital I/O box in my modbus test setup uses less than 10W running (without a screen and not counting mill/lath/... stuff).

The only downside I have seen, with all the RPi boards, is that you need to have a good SD card or it will get destroyed (setting noatime is a must). You can simply make a backup copy, but <i>Real Men don't make backups, Real Men cry</i>. ;-)

PR #4324, opened 2026-08-02
hal: Update mainly C components to getter/setter.

BsAtHome:
Next set of components upgraded to getter/setter. Quite straight forward.

Both demux_generic.c and mux_generic.c have removed the distinction between FP/non-FP thread function export. All threads are FP and we can make it easier. The implementation is no a all-to-all version (except is still does not have 64-bit pins). The magic numbers have been replaced with a macro.

The max31855.comp has had the `float` types replaced with `rtapi_real` (effectively double), which means that a numeric difference can occur at 5..6 decimals. However, I don't believe that this stuff can measure temperatures at those levels, so it should be academic.

In pwmgen.c several pins are now cached because they were read quite often. This should all work as expected. There are several more optimizations possible, but that is for another time.

### DIFF
--- a/src/hal/components/demux_generic.c
+++ b/src/hal/components/demux_generic.c
@@ -19,6 +19,7 @@
 
 #include <rtapi.h>
 #include <rtapi_app.h>
+#include <rtapi_math.h>
 #include <hal.h>
 
 
@@ -30,18 +31,16 @@ MODULE_LICENSE("GPL");
 #define MAX_CHAN 100
 #define MAX_SIZE 1024
 #define EPS 2e-7
-#define MAX_S32 0x7FFFFFFF
-#define MAX_U32 0xFFFFFFFF
 
 typedef struct {
-    hal_data_u *input;
-    hal_data_u **outputs;
-    hal_u32_t *sel_int;
-    hal_bit_t **sel_bit;
-    unsigned int selection;
-    hal_u32_t *debounce;
-    unsigned int timer;
-    hal_bit_t *suppress;
+    hal_refs_u input;
+    hal_refs_u *outputs;
+    hal_uint_t sel_int;
+    hal_bool_t *sel_bit;
+    hal_uint_t selection; // (param)
+    hal_uint_t debounce;
+    hal_uint_t timer;     // (param)
+    hal_bool_t suppress;
     int in_type;
     int out_type;
     int size;
@@ -56,7 +55,6 @@ typedef struct {
 static int comp_id;
 static demux_t *demux;
 static void write_fp(void *arg, long period);
-static void write_nofp(void *arg, long period);
 
 char *config[MAX_CHAN];
 RTAPI_MP_ARRAY_STRING(config, MAX_CHAN, "demux specifiers inNUMout");
@@ -79,7 +77,7 @@ int rtapi_app_main(void){
     }
 
     // allocate shared memory for the base struct
-    demux = hal_malloc(sizeof(demux_t));
+    demux = hal_malloc(sizeof(*demux));
     if (demux == 0) {
         rtapi_print_msg(RTAPI_MSG_ERR,
                 "demux_generic component: Out of Memory\n");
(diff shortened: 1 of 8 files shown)


src/hal/components/demux_generic.c
```diff
@@ -214,44 +203,60 @@ int rtapi_app_main(void){
         if (retval >= HAL_NAME_LEN) {
             goto fail0;
         }
-        retval = hal_pin_new(hal_name, inst->in_type, HAL_IN,
-                (void**)&(inst->input), comp_id);
+        switch(inst->in_type) {
+        case HAL_BOOL: retval = hal_pin_new_bool(comp_id, HAL_OUT, &inst->input.b, 0, "%s", hal_name); break;
```
grandixximo: These should be `HAL_IN`, and the ones below `switch(inst->out_type` should be `HAL_OUT` .
Would love to know how that happened ;-)

src/hal/components/matrix_kb.c
```diff
@@ -201,7 +201,7 @@ int rtapi_app_main(void){
         inst->ncols = 0;
         inst->scan = 0;
         inst->keystroke = 0;
-        inst->param.invert = 1;
+        hal_set_bool(inst->param.invert, 1);
```
grandixximo: This runs before `hal_param_new_bool()` same thing for `rollover` on line 244, and we know why that's bad ;-)

src/hal/components/demux_generic.c
BsAtHome: I must have been braindead when I did that.
But I am guessing I fell for the copy/paste trap, and that is, well, that is braindead, so it fits my earlier assessment.

src/hal/components/matrix_kb.c
BsAtHome: These are a tricky little bastard, every time.

Moved into the init value when creating the param.

src/hal/components/matrix_kb.c
```diff
@@ -237,11 +236,11 @@ int rtapi_app_main(void){
...
         
-        inst->hal.key = (hal_bit_t **)hal_malloc(inst->nrows * inst->ncols * sizeof(hal_bit_t*));
-        inst->now = hal_malloc(inst->nrows * sizeof(hal_u32_t));
-        inst->then = hal_malloc(inst->nrows * sizeof(hal_u32_t));
+        inst->hal.key = hal_malloc(inst->nrows * inst->ncols * sizeof(*inst->hal.key));
+        inst->now = hal_malloc(inst->nrows * sizeof(*inst->now));
+        inst->then = hal_malloc(inst->nrows * sizeof(*inst->then));
         inst->row = 0;
-        inst->param.rollover = 2;
+        hal_set_ui32(inst->param.rollover, 2);
```
grandixximo: this is same as invert, you missed it...

BsAtHome: Argh! Gremlins these are.  #¤%&

grandixximo: I'll go back and check if we missed any of the suckers, and probably set a self reminder for another final pass when we are done with the initial transition, I think we are getting there, at this pace another couple of weeks?

BsAtHome: Yes, please check.

From the original ~350 files we have, excluding these 8, only 71 files left in my converted branch. Over half of that is hostmot2. You can follow that in my tracking branch/PR #4247. At the current pace, we should have nearly everything else done by the end of the week. The remaining components are large ones (pid, stepgen, plasmas, raster,...). Most of these big ones will be PR'ed separately or in pairs, I think, so that bisection will work easier in case of a hidden problem. The hostmot2 code needs some people to test different setups (especially the sserial stuff). That is why it is saved to almost last.

The single biggest file change yet to come is halrmt (more than 3k lines). There is one thing I need to change in the code. Running loadusr can block forever and the code is now single-threaded. I need to add a coroutine setup for loadusr, similar to what I did in linuxcncrsh. Working on it...

PR #4329, opened 2026-08-03
hal: Migrate homecomp is_module pin declaration to bool

grandixximo:
One leftover from the getter/setter migration: homecomp.comp still declares its dummy `is_module` pin with the legacy `bit` type, while the rest of the component was migrated in f107cd307a (#4296).

Found with a tree-wide `halcompupdate --check` sweep (from #4256) against current master; this was the only genuine declaration leftover among all in-tree .comp files not already covered by open migration PRs.

Change: `pin out bit is_module=1` => `pin out bool is_module=1`. Builds clean.

### DIFF
--- a/src/hal/components/homecomp.comp
+++ b/src/hal/components/homecomp.comp
@@ -51,7 +51,7 @@ HOMEMOD=user_homecomp
 https://github.com/LinuxCNC/linuxcnc/blob/BRANCHNAME/src/hal/components/homecomp.comp
 """;
 
-pin out bit is_module=1; //one pin is required to use halcompile)
+pin out bool is_module=1; //one pin is required to use halcompile)
 
 license "GPL";
 author "Dewey Garrett";


BsAtHome:
The build is filtered out in the Submakefile together with tpcomp.comp. These are supposed to be examples of plug-able homing and TP. Neither works in practice or has never been tested to be actually functional afaik. Both modules need to hook into a lot of internals and that is quite contrary the modularization idea.

But sure, we still need to update it, even if it doesn't fix things ;-)

grandixximo:
This is all that came up, while re-reviewing looking for sets before create, that's a good sign...

BsAtHome:
> This is all that came up, while re-reviewing looking for sets before create, that's a good sign...

Yes it is!

And, have you noticed, no one complained about breakage :-)
Maybe because it is holiday season.

Sigma1912:
>  These are supposed to be examples of plug-able homing and TP. Neither works in practice or has never been tested to be actually functional afaik. 

Note that 'homecomp.comp' has definitely been used and has been proven to work. 
As an example:
https://forum.linuxcnc.org/ethercat/52968-help-needed-cracking-the-code-on-linuxcnc-servo-homing-setup?start=30#303183

'tpcomp.comp' on the other hand has apparently proven less useful.

BsAtHome:
Interesting...
However, I'm sure it can no longer compile today on master. The homecomp.comp tries to include a lot of internal headers that are no longer exposed. All of that needs to be untangled before it can work again. It starts with including motion.h and homing.h, but these include more and more of non-public stuff.

WRT tpcomp.comp, this should be removed IMO. The first refactor step is to get TP moved out of RT and make a formal TP input/output specification that can be exposed. Then we can have proper plug-able TPs. However, moving it out of RT is a big undertaking and a lot of work.

Sigma1912:
> However, I'm sure it can no longer compile today on master.

If it is to be removed then we'll want to also remove it from the documentation:
https://github.com/LinuxCNC/linuxcnc/blob/ddd8aef64b833e7c2913a2e6bc0f612b8c76f4d4/docs/src/config/ini-config.adoc?plain=1#L687-L690

BsAtHome:
We could also fix it :-)

But that will take some time. Besides, we'd need a <i>real</i> example to add, not a mostly empty shell.

Sigma1912:
> We could also fix it :-)

That would be very much my preference. :)

PR #4330, opened 2026-08-03
hal: Update kernel-only components (RTAI) to getter/setter

BsAtHome:
These are the components that only build as a kernel module. Some are ancient and are for ISA boards. You'd need a - too expensive for my pay grade - industrial PC to get that working today. I'd wonder what linux kernel version you need too.

The pluto_servo.comp and pluto_step.comp still have their troff formatted documentation. They never got converted to adoc and nobody complained.

Anyhoo,... the conversions should be rather straight forward. We can probably never test this. We really should do a survey on what hardware is used out there. There comes a time when you need to prune the ballast.

Please check for any parameter gremlins ;-)

The old hal types used in hal_motenc.comp/motenc.h (hardware MMIO map) was fixed using the rtapi types marked volatile. That should be fully compatible.

### DIFF
--- a/src/hal/drivers/pci_8255.c
+++ b/src/hal/drivers/pci_8255.c
@@ -28,24 +28,24 @@ RTAPI_MP_ARRAY_INT(dir, MAX, "I/O direction of 8255s");
 
 static int comp_id;
 
-union inv { hal_bit_t *not_; hal_bit_t invert; };
+union inv { hal_bool_t not_; hal_bool_t invert; };
 
 struct port {
-    hal_bit_t *a[8];
-    hal_bit_t *b[8];
-    hal_bit_t *c[8];
+    hal_bool_t a[8];
+    hal_bool_t b[8];
+    hal_bool_t c[8];
     union inv ai[8];
     union inv bi[8];
     union inv ci[8];
-    hal_u32_t dir_;
-    hal_u32_t ioaddr;
+    hal_uint_t dir_;
+    hal_uint_t ioaddr;
 };
 
 struct state {
     struct port ports[3];
-    hal_bit_t *relay;
-    hal_bit_t relay_invert;
-    hal_u32_t ioaddr;
+    hal_bool_t relay;
+    hal_bool_t relay_invert;
+    hal_uint_t ioaddr;
 };
 
 static void read(struct port *inst, long period);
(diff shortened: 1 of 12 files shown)


src/hal/drivers/pci_8255.c
```diff
@@ -77,58 +77,57 @@ static int export(char *prefix, struct port *inst, int ioaddr, int dir) {
     hal_pin_dir_t direction;
     int sz = sizeof(struct port);
     memset(inst, 0, sz);
-    inst->dir_ = dir;
-    inst->ioaddr = ioaddr;
+    hal_set_ui32(inst->ioaddr, ioaddr);
```
grandixximo, line 82: Where does `inst->ioaddr` point at this moment? `memset` on the line above just zeroed the struct, and I can't find `struct port.ioaddr` ever being passed to a `hal_pin_new_*`/`hal_param_new_*`. Only `struct state.ioaddr` is created (line 198). Doesn't this write go through a NULL reference on every port of every board?

src/hal/drivers/pci_8255.c
```diff
@@ -332,53 +334,54 @@ static void write(struct port *inst, long period) {
...
+        pci_8255_outb(byte, xioaddr, 0);
 	if(first) rtapi_print_msg(RTAPI_MSG_DBG, "write: 0 %02x\n", byte);
     }
     first = 0;
 }
 
 static void read(struct port *inst, long period) {
-    int p = dir_;
+    int p = hal_get_ui32(dir_);
+    rtapi_u32 xioaddr = hal_get_ui32(ioaddr);
```
grandixximo, line 347: Same reference as in `export()`: `struct port.ioaddr` is never created, so this reads through an uninitialized (NULL) reference. As far as I can tell `loadrt pci_8255` dereferences NULL before any I/O happens, either at the write in `export()` or at first read/write here. Am I missing a creation site somewhere?

src/hal/drivers/pci_8255.c
```diff
@@ -196,19 +195,18 @@ int rtapi_app_main(void) {
 	// relay, CS# as outputs
 	pci_8255_outb(0x11, io[i]+2, 0);
 
+	r = hal_param_new_ui32(comp_id, HAL_RO, &(inst->ioaddr), io[i],
```
grandixximo: While here: this is inside the board loop but always takes `&(inst->ioaddr)`, i.e. board 0's slot, not `inst[i]`. Pre-existing (the old code had the same), but it now also means `inst[i].ioaddr` for i>0 stays an uninitialized reference that `write_relay` will pass to `hal_get_ui32`.

src/hal/drivers/pci_8255.c
BsAtHome: Indeed, there are two distinct references for `ioaddr`.

The one here in export() references the port structure, which has not yet had its address allocated. The exposed iobase parameter is created just before calling export. Then, the export function has it as a modified address for the port.

Need to allocate a fake in the port structure.

src/hal/drivers/pci_8255.c
BsAtHome: No, you are not. The fix in export, creating the storage there, should be good.

src/hal/drivers/pci_8255.c
BsAtHome: Yes, a pre-existing bug. I'm sure they never tested it with multiple boards...

Fixing ayway, `&(inst->ioaddr)` --> `&(inst[i].ioaddr)`.

PR #4337, opened 2026-08-04
hal: restore working homecomp example

grandixximo:
Follow-up to #4329. homecomp.comp was filtered out of the build because it no longer compiled on master; the discussion on #4329 made it clear that should be fixed. This restores homecomp as a real, buildable plug-in homing module that demonstrates something useful, rather than an empty "always-homed" shell.

The default skeleton is a working homing module:

- Creates the same per-joint hal pins the real homemod exposes: joint.N.home-sw-in, joint.N.homing, joint.N.homed, joint.N.home-state, joint.N.index-enable, for every joint seen at homing_init time. The output pins are driven once per servo period, so joint.N.homed tracks what motion actually believes.
- Adds a module-level pin motion.allow-sim-homed (default true). With it set, all joints report homed, which is useful for bring-up on hardware without working home switches. With it cleared, each joint reports homed only while its joint.N.home-sw-in pin is true, showing how HAL inputs wire into the homing API. The flag can be flipped from a postgui or M-code without rebuilding.
- Sizes its HAL buffer from n_joints at runtime (single hal_malloc), so the template does not depend on EMCMOT_MAX_JOINTS.
- do_homing() returns 1 on the not-allhomed => allhomed transition, matching base_do_homing() semantics so motmod switches FREE => teleop when homing completes.

The skeleton's function signatures use opaque pointer and integer arguments, so the default path does not pull in motion.h's heavy transitive includes. The exported symbols still resolve to the names motmod calls.

The advanced HOMING_BASE path is preserved: with HOMING_BASE pointing at a checkout's src/emc/motion/homing.c, the template #includes that file with CUSTOM_HOMEMODULE defined and exposes per-joint request-custom-homing / is-custom-homing pins for users who want to override the base_* helpers. This is the original pattern from the 2.9-era homecomp, kept for completeness. Note that motion.h/homing.h are no longer exported to the halcompile include path since 82f05bccb3, so the custom block relies on homing.c's own quoted includes and does not include them directly.

Verified: both build modes compile clean with halcompile (zero warnings), and the skeleton was runtime-tested headless as HOMEMOD=homecomp (sim on => all homed; sim off => unhomed; sim off with all home-sw-in closed => homed again).

### DIFF
--- a/src/hal/components/homecomp.comp
+++ b/src/hal/components/homecomp.comp
@@ -5,6 +5,10 @@ Example of a plug-in homing module buildable with halcompile.
 Demonstrates the required code structure for #includes, function
 definitions, HAL pins, etc.
 
+The complete homing module interface (every function a homing module
+must export, when motmod calls it, and what it should do) is
+documented in `src/emc/motion/homing.h`.
+
 Build modes:
 
 * *Skeleton* (default).  If HOMING_BASE is _not_ #defined, the module
--- a/src/hal/components/homecomp.comp
+++ b/src/hal/components/homecomp.comp
@@ -1,42 +1,46 @@
-component homecomp"homing module template";
+component homecomp "homing module template";
 
 description """
-Example of a homing module buildable with halcompile.
-Demonstrates required code for #includes, function definitions, etc.
+Example of a plug-in homing module buildable with halcompile.
+Demonstrates the required code structure for #includes, function
+definitions, HAL pins, etc.
 
-If *HOMING_BASE* is #defined and points to a valid `homing.c` file,
-an example of a customized homing module is built.  This module
-creates input hal pins joint.n.request-custom-homing that enable an
-alternate joint homing state machine for requested joints.  A hal output
-pin joint.N.is_custom-homing verifies selection"
+Build modes:
 
-The customized homing module utilizes many of the base homing api
-routines from homing.c without modification but augments other base
-functions to add support for custom hal pins and custom joint homing
-state machines.  A user-built module will likely replace additional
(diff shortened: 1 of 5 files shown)


src/hal/components/homecomp.comp
```diff
@@ -46,421 +50,342 @@ HOMEMOD=user_homecomp
...
+        rv += hal_pin_new_bool(id, HAL_OUT, &p->homed,        0,
+                               "joint.%d.homed",         jno);
+        rv += hal_pin_new_si32(id, HAL_OUT, &p->home_state,   0,
+                               "joint.%d.home-state",    jno);
+        rv += hal_pin_new_bool(id, HAL_IO,  &p->index_enable, 0,
+                               "joint.%d.index-enable",  jno);
     }
-    return retval;
-} // custom_makepins()
+    return rv;
```
BsAtHome: We should not repeat this return value summing. It is a bad habit and the return value does no longer represent the actual error. Create a wrapping macro that tests should be used.
```C
#define CHK(x) \
    do { \
        int _rv = (x); \
        if(0 != _rv) return _rv; \
    } while(0)

    CHK(hal_pin_new_bool(...));
```

grandixximo: Easy, that's all? The overall shape is acceptable?

BsAtHome: Well, I haven't tested it. That I need to do too :-)

I do think there is a lack of documentation about all those functions defined and what they are supposed to do/provide.

grandixximo: Added docs in `homing.h`

PR #4338, opened 2026-08-04
hal: Update iocontrol to use getter/setter

BsAtHome:
This removes the use of `hal.hh` that accesses HAL's internal structure, which no longer is allowed. The PR does not (yet) remove the hal.hh file, but adds a preprocessor warning when included.
Taskclass/iocontrol was the only place where hal.hh was used and it is easily converted using standard methods and getter/setter<sup>(*)</sup>.

<sup>(*)</sup> <sub>The C++ wrapper hid the type of the pin/param. It is not clear why hiding the HAL types in get/set operations would be an advantage. It only increases complexity in finding out what you are actually using and opens the door to a class of errors of mistaken identity. Or you'd need to start naming your variables to include the type, which then would question why you used the type hiding in the first place.</sub>

### DIFF
--- a/src/emc/task/taskclass.cc
+++ b/src/emc/task/taskclass.cc
@@ -38,27 +38,40 @@ using namespace linuxcnc;
 *
 * Called By: main
 ********************************************************************/
+#define IOC0 "iocontrol.0"
+
+#define CHK(x) do { \
+        int _rv = (x); \
+        if(_rv < 0) { \
+            rtapi_print_msg(RTAPI_MSG_ERR, IOC0": cannot create HAL pin\n"); \
+            return -1; \
+        } \
+    } while(0)
+
 int Task::iocontrol_hal_init(void)
 {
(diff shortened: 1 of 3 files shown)


grandixximo:
This gives way to #4251 if I'm not mistaken?

BsAtHome:
> This gives way to #4251 if I'm not mistaken?

Only in part.

Remember, the goal of the getter/setter change is to eliminate the 32-bit types and make everything 64-bit clean. Introducing #4251 now would again make that goal a lot more difficult because the code is currently not clean. Having the C++ wrapper hiding the types is making evaluating _where_ we need to fix 64-bit compatibility a whole lot more difficult (see my comment in the PR description). Remember, we can't just replace 32-bit types with 64-bit types. Every use must be evaluated in context. We need to go through the cleanup steps first.

grandixximo:
Fair, I will just rebase #4251 cleanup and push, I'll wait on your sign to come out of draft.

PR #4340, opened 2026-08-05
docs: add derived-id section reference lint

grandixximo:
Companion to #4339 (independent, either merge order).

Sections linked via auto-generated AsciiDoc ids (`<<_normal_download>>` xrefs, `link:...html#_ini_settings[...]` URLs) break in translated docs when the section title is translated, because asciidoctor derives the id from the title text. #4339 pins the five targets that are broken in translations today. This PR makes sure the pattern does not regrow.

`scripts/docs-anchor-check.py` does two passes over docs/src: collect every explicit anchor definition (`[[id]]`, `[#id]`, `id=`, `anchor:id[]`), then report each `<<_derived>>` xref or `#_derived` link URL whose target is not explicitly defined. Namespaced anchors all carry a colon by convention, so a `_`-prefixed target with no colon is a derived id.

It is wired warn-only into the htmldocs CI job (same pattern as `htmlcheck.sh -w`), printing offenders with file:line and writing the list to the GHA job summary. Flip to `--enforce` once the tree is clean and the check has had a burn-in.

Also included: explicit anchors for the remaining lint hits, all same-page links in untranslated man pages (`_subcommands` in linuxcncrsh.1, `_mbccs_file_format`, `_modbus_functions`, `_modbus_types`, `_hal_types` in mesambccc.1). The explicit ids match the previously auto-generated ones, so rendered output is unchanged; I verified the ids in the built HTML.

Current lint output on this branch is exactly the five references pinned by #4339, so the check turns clean as soon as that PR lands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,6 +187,8 @@ jobs:
         .github/scripts/build-doc.sh
     - name: Check manpage names match filenames
       run: scripts/manpage-name-check.py --enforce docs/build/man
+    - name: Check for derived-id section references
+      run: scripts/docs-anchor-check.py --enforce
     - name: Verify no untracked or modified files after build
       run: |
         #*.po and documentation.pot are modifyed by build. Ignore them for now.
--- a/docs/src/man/man1/linuxcncrsh.1.adoc
+++ b/docs/src/man/man1/linuxcncrsh.1.adoc
@@ -178,6 +178,7 @@ The supported commands are as follows:
   Help will respond regardless of whether a 'HELLO' has been
   successfully negotiated.
 
+[[_subcommands]]
 == SUBCOMMANDS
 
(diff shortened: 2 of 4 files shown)


grandixximo:
this looks like this now:

image

Acceptable? or keep the warning only in the annotations, and remove the summary from here??

grandixximo:
Tested as working without `--enforce` before merging of #4339, see the picture in my previous comment.
I pushed the enforced version now that #4339 is merged, tree is clean, and merging this will prevent the entire fault variant to enter in tree and trickle down to the translations.

BTW I also cleaned up the rest of the html broken links warnings coming from the translations in weblate, not sure when those will gurgle up, after that we shall have a squeaky clean tree (insert here humorous comment).

BsAtHome:
> (insert here humorous comment).

So clean the birds will refuse to sit on the branches of the tree and silence was heard.

PR #4342, opened 2026-08-05
Update spindle orient docs

Sigma1912:
This corrects misinformation in

- the m-code docs: M19 does not use encoder index.
- the 'orient.comp' man page: Function description does not match the current code.

See https://forum.linuxcnc.org/38-general-linuxcnc-questions/59064-m19-orientation-drift-with-external-encoder-and-orient-component-linuxcnc-2-9-4#348461


grandixximo:
Should back-port to 2.9 ?

Sigma1912:
> Should back-port to 2.9 ?

Please do, thank you!

PR #4343, opened 2026-08-05
Orient.comp: Add modes with preceding homing to index

Sigma1912:
Some spindle encoder setups cannot track the encoder position when using high spindle speeds and thus cannot use the current 'orient' component due to lost encoder position.
This PR expands the orient component with three additional modes that offer referencing to an index signal prior to orienting the spindle.

Current behavior is unchanged. All modes have been tested on real hardware.

Based on work by @andypugh

For more:
https://forum.linuxcnc.org/10-advanced-configuration/42792-m19-and-spindle-index-enable-issues?start=10


src/hal/components/orient.comp
```diff
@@ -59,32 +69,83 @@ license "GPL";
...
     is_oriented_set(0); // spindle.is-oriented inhibits spindle.orient
-    if (enable) {
-       if (enable ^ last_enable) {     // positive edge on enable
-           is_oriented_set(0);
+
+    switch (state){
+        case 0: // waiting
+            if (enable) {
+            index_enable_set(0);
+            if (enable ^ last_enable) {     // positive edge on enable
```
BsAtHome: Reading and writing pins and params is expensive. You should cache the value of `enable` once at the start of the function<sup>(*)</sup>. Also, using xor is not the positive edge. It hits on both positive edge (en && !last_enable) and negative edge (!en && last_enable).
```
rtapi_bool en = enable;
...
if (en && !lastenable) { // This is the positive edge of enable.
}

// should also fix other use is 'enable' and replace with cached version

...
last_enable = en;
```

(*) I know, `enable` is not always read multiple times due to the switch. But using a local variable at the top will let the compiler decide on the best optimization.

src/hal/components/orient.comp
```diff
@@ -59,32 +69,83 @@ license "GPL";
...
+                is_oriented_set(0);
```
BsAtHome: Wasn't this set unconditionally at the top of the function?

src/hal/components/orient.comp
```diff
@@ -59,32 +69,83 @@ license "GPL";
...
+                command_set(floor(latched_position-target_angle) + target_angle);
                 break;
+            default:
+                rtapi_print_msg(RTAPI_MSG_ERR, "unhandled mode %i\n", mode % 3);
+        }
+        poserr_set((position - command) * 360.0);
+        debounce += (fabs(poserr) < tolerance && debounce <=100);
+        is_oriented_set(debounce > 100);
+        if (is_oriented) state_set(0);
+        break;
```
BsAtHome: There seems to be some indenting trouble here.

src/hal/components/orient.comp
```diff
@@ -59,32 +69,83 @@ license "GPL";
...
-                command_set(floor(latched_position-target_angle) + target_angle );
+                command_set(floor(latched_position-target_angle) + target_angle);
                 break;
+            default:
+                rtapi_print_msg(RTAPI_MSG_ERR, "unhandled mode %i\n", mode % 3);
+        }
+        poserr_set((position - command) * 360.0);
+        debounce += (fabs(poserr) < tolerance && debounce <=100);
+        is_oriented_set(debounce > 100);
+        if (is_oriented) state_set(0);
```
BsAtHome: You set the `is_oriented` pin and the immediately read it in the if. Better would be to write:
```
if (is_oriented_set(debounce > 100)) {
    state_set(0);
}
```
That would prevent the read after write.

src/hal/components/orient.comp
```diff
@@ -59,32 +69,83 @@ license "GPL";
...
+        break;
+
+        case 3: // index search - init
+            if (! enable) state_set(0);
+            latched_position = position; // sample now
+            state_set(4);
```
BsAtHome: The `state_set(0)` is never honored. It is unconditionally overwritten by `state_set(4)`.

The same thing happens other places. Are you forgetting a break?

src/hal/components/orient.comp
```diff
@@ -59,32 +69,83 @@ license "GPL";
...
+                break;
+            }
+            switch (mode) {
+            case 3: // shortest move to index
+                if (0 > latched_position) {
+                command_set(position + 0.5);
+                } else {
+                command_set(position - 0.5);
+                }
+                break;
```
BsAtHome: Some indenting problem here.

src/hal/components/orient.comp
Sigma1912: Done

src/hal/components/orient.comp
Sigma1912: Removed

src/hal/components/orient.comp
Sigma1912: Fixed

src/hal/components/orient.comp
Sigma1912: Factored out to one single gate before entering the case structure.

src/hal/components/orient.comp
Sigma1912: Fixed

src/hal/components/orient.comp
```diff
@@ -60,31 +70,81 @@ license "GPL";
...
+                if (en && !last_enable) {     // positive edge on enable
+                    if (mode > 2){
+                        index_enable_set(1);
+                        state_set(3);
+                    } else {
+                        state_set(1);
+                }
+            }
+            }
+            break;
```
BsAtHome, line 95: Using both `if(en)` followed by `if(en &&...)` is double up. The second if can never be true when the first if is not true. Therefore, the first `if(en)` is not needed.
(should also fix indenting)

Sigma1912: Done, thanks.

Sigma1912:
Thank you for reviewing and your pointers! Much appreciated.

BsAtHome:
You are welcome. I'm just looking at this on the structural level. Not looking at what it is actually doing ;-)

There may be one more problem. You have a message printing when the mode value is not handled. If I'm not mistaking, this message would print at every thread-cycle flooding the terminal and GUI.

grandixximo (commented):
Thanks for the patch. Tested-on-hardware confirmation is much appreciated, and the state machine structure is a clean way to add the index modes. I have a few questions and concerns inline, mostly around behavior changes in the existing modes.

src/hal/components/orient.comp
```diff
@@ -62,31 +70,79 @@ license "GPL";
 
 FUNCTION(_) {
 
-    rtapi_real target_angle;
-    rtapi_real latched_position;
+    rtapi_bool en = enable;
+
+    if (!en) state_set(0);
     is_oriented_set(0); // spindle.is-oriented inhibits spindle.orient
```
grandixximo, line 80: What keeps is-oriented high after the orient completes? Once debounce crosses 100, the pin goes high and state returns to 0, so the next cycle this line forces it back to 0 and nothing re-asserts it. That makes is-oriented a single-cycle pulse, while the old code held it high for as long as enable stayed high. Can motion miss a one-cycle assertion of spindle.is-oriented?

Related: poserr also freezes at its last value in state 0, where the old code kept updating it while enable was high.

src/hal/components/orient.comp
```diff
@@ -62,31 +70,79 @@ license "GPL";
 
 FUNCTION(_) {
 
-    rtapi_real target_angle;
-    rtapi_real latched_position;
+    rtapi_bool en = enable;
+
+    if (!en) state_set(0);
```
grandixximo: If enable drops while in state 3 or 4, what clears index-enable? As far as I can tell it stays asserted and the encoder keeps resetting position to 0 at every index from then on.

src/hal/components/orient.comp
```diff
@@ -62,31 +70,79 @@ license "GPL";
...
     is_oriented_set(0); // spindle.is-oriented inhibits spindle.orient
-    if (enable) {
-       if (enable ^ last_enable) {     // positive edge on enable
-           is_oriented_set(0);
+
+    switch (state){
+        case 0: // waiting
+            if (en && !last_enable) {     // positive edge on enable
+                if (mode > 2){
```
grandixximo: Should mode be range-checked here at the enable edge? A value > 5 passes this test and lands in the state 4 default branch, and a negative value reaches the mode % 3 default in state 2 (C remainder keeps the sign). If the mode check happened here once, the default branches below become unreachable and the print flooding Bertho mentioned goes away with them.

src/hal/components/orient.comp
```diff
@@ -62,31 +70,79 @@ license "GPL";
...
+                    }
+                    break;
+                case 4: // always cw to index
+                        command_set(position + 0.5);
+                    break;
+                case 5: // always ccw to index
+                        command_set(position - 0.5);
+                    break;
+                default:
+                    rtapi_print_msg(RTAPI_MSG_ERR, "unhandled mode %i\n", mode);
```
grandixximo: How often does this fire while waiting for index? If I read it right it prints once per servo cycle, and if the index signal never comes it prints forever. If this branch stays as a safety net after validating mode at entry, should it at least fall back to state 0 after printing once?

src/hal/components/orient.comp
```diff
@@ -62,31 +70,79 @@ license "GPL";
...
+                state_set(1);
                 break;
             }
-       }
-       poserr_set((position - command) * 360.0);
-       debounce += (fabs(poserr) < tolerance && debounce <=100);
-       is_oriented_set((debounce > 100));
+            switch (mode) {
+                case 3: // shortest move to index
+                    if (0 > latched_position) {
```
grandixximo: Does the sign of the accumulated position say anything about the distance to index? For a spindle that has turned net-positive since power-up this always picks position - 0.5. I would expect "shortest move to index" to look at the fractional revolution, e.g. latched_position - floor(latched_position), and pick direction from whether that is above or below 0.5. Am I missing something about how position is guaranteed to be scaled here?

src/hal/components/orient.comp
```diff
@@ -62,31 +70,79 @@ license "GPL";
...
+            }
+            break;
+
+        case 3: // index search - init
+            latched_position = position; // sample now
+            state_set(4);
+            break;
+
+        case 4: // waiting for spindle index
+            if (! index_enable){ // index has reset)
```
grandixximo: What happens if the index signal never arrives, broken wire or disconnected encoder? It looks like the spindle keeps rotating at the ±0.5 rev offset indefinitely with no error reported. Worth a timeout param, or at least a note in the description?

src/hal/components/orient.comp
```diff
@@ -62,31 +70,79 @@ license "GPL";
...
+                state_set(1);
```
grandixximo, line 129: When the index hits, state 1 relatches position (~0) and command jumps from the moving ±0.5 rev offset to the new target. Did you see any PID kick from that step on hardware, or is it worth smoothing?

src/emc/rs274ngc/interp_check.cc
```diff
@@ -338,8 +338,8 @@ int Interp::check_other_codes(block_pointer block)       //!< pointer to a block
       CHKS(((motion == G_2 || motion == G_3 || (block->m_modes[7] == 19)) &&
 	    fabs(p_value - block->p_number) > 0.001),
 	   _("P value not an integer with M19 G2 or G3"));
-      CHKS((block->m_modes[7] == 19) && ((p_value > 2) || p_value < 0),
-	   _("P value must be 0,1,or 2 with M19"));
+      CHKS((block->m_modes[7] == 19) && ((p_value > 5) || p_value < 0),
+	   _("P value must be 0,1,2,3,4 or 5 with M19"));
```
grandixximo, line 342: The interpreter now accepts P0..P5; should the M19 documentation be updated in this PR to match? It still describes only P0..P2.

src/hal/components/orient.comp
Sigma1912: Fixed. 
As long as 'enable' is high case 2 is executed. Thus 'poserr' is updated and 'is_oriented' is reasserted (as long 'poser < tolerance').

src/hal/components/orient.comp
Sigma1912: Good catch. Fixed.
The (!en) gate now checks for 'state==4' and deasserts 'index_enable' .

src/emc/rs274ngc/interp_check.cc
Sigma1912: Done

src/hal/components/orient.comp
Sigma1912: That is correct but can be easily handled by setting the 'maxoutput' value of the pid.

src/hal/components/orient.comp
Sigma1912: Orient timeouts are handled in motion as 'M19' expects a 'Q' word.
 If spindle.N.is-oriented does not become true within the time set by 'Q' error occurs.

src/hal/components/orient.comp
Sigma1912: Good point. I only tested with 'm3'.
Adopted suggested logic and testing now passes both 'm3' and 'm4' spindle rotation before 'M19'

src/hal/components/orient.comp
Sigma1912: Code removed.

src/hal/components/orient.comp
Sigma1912: Done.
Changed to caching the 'mode' value on 'enable' positive edge  instead of reading it multiple times and checking range right after.

Sigma1912:
Thanks again for investing the time to review!

Sigma1912:
Changed 'M19' doc:
Screenshot from 2026-08-06 13-59-19

andypugh:
The way I had anticipated solving this was to add spindle homing pins to the spindle control logic. 

It is almost certainly simpler to add it to the orient comp, but I am not sure which way is better.

grandixximo:
> The way I had anticipated solving this was to add spindle homing pins to the spindle control logic.
> 
> It is almost certainly simpler to add it to the orient comp, but I am not sure which way is better.

Something against having both?

PR #4346, opened 2026-08-06
docs -add writeup about a dual PID axis example

c-morley:
An example of dual PID axes in HAL

### DIFF
--- a/docs/src/docs.xml
+++ b/docs/src/docs.xml
@@ -88,6 +88,7 @@
 <doc name="motion_kinematics" title="Kinematics"/>
 <doc name="motion_tweaking_steppers" title="Tweaking Steppers"/>
 <doc name="motion_pid_theory" title="PID theory"/>
+<doc name="motion_dual_pid_example" title="Dual PID Example"/>
 <doc name="ladder_ladder_intro" title="Ladder Introduction"/>
 <doc name="examples_pci-parallel-port" title="PCI Parallel Port"/>
 <doc name="examples_spindle" title="Spindle Example"/>
--- a/docs/src/motion/dual-pid-example.adoc
+++ b/docs/src/motion/dual-pid-example.adoc
@@ -1,0 +1,137 @@
+= Dual Feedback PID
+
+
+== Introduction
+A dual feedback machine axis typically consists of a rotary encoder on the motor 
+and a linear encoder on the axis. The motor encoder is used for the bulk movement 
+using P, D and FF1 parameters and the linear encoder on the axis is used to 
+remove the last small steady state error using the I parameter. The two velocity 
+commands are added (summed) together before being sent to the servo amplifier.
+
+== Flow Chart Diagram
+
+.Dual PID Flow Chart
+image::pid-images/dual_pid_example.svg["Dual PID Flow Chart"]
+
+== Example HAL Code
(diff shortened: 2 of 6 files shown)


grandixximo:
svg fails to render on github, is it hand written?

c-morley:
no, exported from Dia program

grandixximo:
It does not look like an svg, is it encrypted or something? I opened with text editor and it does not look like any svg I have ever seen before...

grandixximo:
it is a PNG actually, I change the format, and now it opens up

image

the SVG file has the same sha256 as the PNG file, so they are byte identical, they only differ in the file extension.

c-morley:
Ok thanks I think this fixes it.

grandixximo:
I don't understand how summing the PID results in the position you want to sent the motor at.
You are running this on a machine?

c-morley:
I am not, other are.
https://forum.linuxcnc.org/10-advanced-configuration/37353-dual-pid-loops-and-appropriate-pins-for-feedback-to-the-trajectory-planner-et-al

grandixximo:
My bad, I missed the fact that it is controlling the servo in velocity mode, makes more sense now, forum feedback looks good...

docs/src/docs.xml
```diff
@@ -88,6 +88,7 @@
 <doc name="motion_kinematics" title="Kinematics"/>
 <doc name="motion_tweaking_steppers" title="Tweaking Steppers"/>
 <doc name="motion_pid_theory" title="PID theory"/>
+<doc name="dual_pid_example" title="Dual PID Example"/>
```
grandixximo: Sibling motion docs use `motion_` prefix.
Should be `motion_dual_pid_example`

docs/src/motion/dual-pid-example.adoc
```diff
@@ -0,0 +1,137 @@
...
+net z-pos-cmd    =>  pid.z2.command
+
+# Enable both loops together
+net z-enable     <=  joint.2.amp-enable-out
+net z-enable     =>  pid.z.enable
+net z-enable     =>  pid.z2.enable
+
+# connect index feedback to compensate during index
+# motor encoder
+net z2-index-enable  <=> pid.z.index-enable
```
grandixximo, line 71: Why connect the motor encoder index signal at all?
pos-fb comes from the linear encoder, why would you leave it to chance to have the motor index decide where homing happens on the linear feedback?
The whole reason there is a linear encoder is because the motor has some backlash which needs tweaking, why would you leave it to chance to end up on a less accurate index?
Am I missing something?
I see the note at the end, but IMO it would be better not to have this double connection in the base doc config, a better shape IMO is to comment out the motor index, and leave a note to enable only if the linear encoder index is not present because it is less accurate.

c-morley: I tend to agree in principle. I don't know if the encoder uses the index enable to reset something internally regardless if the motion component uses it.
In other words - since I can't test it, I'll leave it as is because it did work.
If we get more info later the docs can be updated.

BsAtHome:
Why is there both the SVG and PNG image attached?

We should only need the SVG. Or?

c-morley:
Yes I could delete the png.

PR #4348, opened 2026-08-06
configs: align sim axis [TRAJ] max velocity with [DISPLAY] and joints

grandixximo:
configs/sim/axis/axis.ini is the only shipped GUI sim config where [DISPLAY]MAX_LINEAR_VELOCITY (5) disagrees with [TRAJ]MAX_LINEAR_VELOCITY (4): gmoccapy sets both equal, touchy and axis_mm leave DISPLAY unset and rely on the TRAJ fallback, and the joints are 5 as well.

The mismatch is not just cosmetic. axis initializes its Max Velocity slider to the [DISPLAY] maximum, and the Tk scale -command sends EMC_TRAJ_SET_MAX_VELOCITY(5) at startup. emcTrajSetMaxVelocity() writes that straight to motion as EMCMOT_SET_VEL_LIMIT with no clamp against the ini value (src/emc/task/taskintf.cc), after initraj already seeded the limit with the [TRAJ] value (4). If the GUI's NML write lands before linuxcncsvr has formatted the command buffer it is lost and the machine keeps 4; otherwise it runs at 5. So the actual coordinated-move velocity cap for this config depends on startup timing, and the axis Max Velocity readout settles at 240 or 300 in/min accordingly. I observed this flip-flopping across CI runs of the ui-smoke screenshot tests, where both end states are stable and a settle-on-unchanged-frames grab cannot catch the loser.

Set [TRAJ]MAX_LINEAR_VELOCITY to 5 so the power-on default matches the [DISPLAY] slider maximum and the joint limits, making the settled cap deterministic. Practical behavior barely changes: axis already tries to set 5 at every startup, it just sometimes fails.


hansu:
Yeah, it makes no sense to have a smaller value in TRAJ than in DISPLAY

PR #4349, opened 2026-08-06
docs: correct statements that no longer match the source (Code Notes, G-code, HAL)

Zia-research:
**Disclaimer.** This work is AI-generated — directed, reviewed and published by me. Every
claim carries a `file:line` citation, and a script re-reads each cited line against the
source. @grandixximo reviewed these three patches on #3718 and checked the claims at the
cited locations; the one defect he found — the PAUSE paragraph's jerk wording — is fixed
here.

Three documentation corrections, each for a statement that no longer matches the source.

**`docs/src/hal/basic-hal.adoc`** — the `initf` example names a funct the lcec driver does
not export: `lcec.0.activate`. The real name is `lcec.activate`, and it is global rather
than per-master. Fixed in the prose and in the example block.

**`docs/src/code/code-notes.adoc`** — seven corrections. Two remove bug reports for bugs
fixed years ago (`OVERRIDE_LIMITS`, the `initraj.cc` example) — the failure mode where a
document records defects but never their repair. The PAUSE rewrite adds a fact documented
nowhere else: pause is bypassed while a segment is position-synchronized to the spindle,
i.e. during threading and rigid tapping, because `tpGetFeedScale()` returns `1.0` there.

**`docs/src/gcode/g-code.adoc`** — the G33/G33.1, G64 and G96 error lists aligned with the
interpreter: two phantom errors removed, several real and undocumented ones added.

The commit messages carry the evidence, `file:line` for each change. The full audit,
including its own corrections, is at https://github.com/Zia-research/linuxcnc-audit

**Base.** Rebased on current master. The three files are unchanged upstream since the
audit was made, so these are the diffs that were reviewed.

**One caveat.** The wording describes `master`. If 2.9-applicable text is wanted, the
command-count note (73/70 → 76/73) is the only figure that differs — `cmd_code_t` would
need checking on the 2.9 branch first.

Independent of #3718, and it does not touch the block diagram.

### DIFF
--- a/docs/src/code/code-notes.adoc
+++ b/docs/src/code/code-notes.adoc
@@ -366,8 +366,10 @@ The ENABLE command enables the motion controller.
 
 ==== Requirements
 
-None. The command can be issued at any time, and will always be
-accepted.
+The command is rejected with an error message ("can't enable motion,
+enable input is false") if the motion.enable HAL pin is low. The
+hardware enable chain must be satisfied before the controller can be
+enabled.
 
 ==== Results
 
@@ -464,10 +466,26 @@ modify the joint's amp enable pin.
 
 === PAUSE
 
-The PAUSE command stops the trajectory planner. It has no effect in
-free or teleop mode. At this point I don't know if it pauses all motion
-immediately, or if it completes the current move and then pauses before
-pulling another move from the queue.
+The PAUSE command pauses the trajectory planner. It has no effect in
+free or teleop mode.
+
+The machine neither stops instantly nor completes the current move: the
+planner treats pause as a request to bring the feed to zero, so the
+machine decelerates to a stop within the current segment, at that
+segment's acceleration limit, and halts wherever the deceleration ramp
+ends. The run-down is jerk-limited only on machines configured for the
+S-curve planner ([TRAJ]PLANNER_TYPE = 1 and MAX_JERK at 1.0 or above; a
+smaller jerk forces the planner back to trapezoidal). Both default to off,
+so a stock machine decelerates on a trapezoidal profile.
+
+Pause is ignored while the active segment is position-synchronized with
+the spindle (spindle-synchronized motion such as G33 threading and
+rigid tapping): for those segments the planner forces full feed and
+bypasses both pause and feed override, because tool position is slaved
+to spindle angle and pausing mid-thread would destroy the work.
+The pause takes effect when the synchronized segment ends.
+Velocity-synchronized segments (G95 feed per revolution) can be paused
+normally.
 
 ==== Requirements
 
@@ -476,7 +494,10 @@ accepted.
 
 ==== Results
 
-The trajectory planner pauses.
+The planner's pausing flag is set. Motion ramps down to zero velocity
+within the current segment (except during position-synchronized
+segments, see above) and the planner stops pulling new segments from
+the queue until RESUME or STEP.
 
 === RESUME
 
@@ -495,15 +516,15 @@ The trajectory planner resumes.
 === STEP
 
 The STEP command restarts the trajectory planner if it is paused, and
-tells the planner to stop again when it reaches a specific point. It
-has no effect in free or teleop mode. At this point I don't know
-exactly how this works. I'll add more documentation here when I dig
-deeper into the trajectory planner.
+tells the planner to pause again when the executing motion id changes
+(i.e. after the current line of the program completes). It has no
+effect in free or teleop mode.
 
 ==== Requirements
 
-None. The command can be issued at any time, and will always be
-accepted.
+The planner must already be paused. If motion is executing, the
+command is rejected with an error message ("can't STEP while already
+executing").
 
 ==== Results
 
@@ -541,10 +562,12 @@ accepted. (I think it should only work in free mode.)
 
 ==== Results
 
-Limits on all joints are over-ridden until the end of the next JOG
-command. (This is currently broken... once an OVERRIDE_LIMITS command
-is received, limits are ignored until another OVERRIDE_LIMITS command
-re-enables them.)
+Limits that are currently tripped are over-ridden until the end of the
+next jog command, at which point they are automatically re-enabled.
+Only the tripped limits are over-ridden: the command handler builds a
+mask from each joint's negative/positive hard-limit flags, so joints
+that are not on a limit keep their protection. Issuing the command
+with a negative joint number cancels an override explicitly.
 
 === HOME
 
@@ -718,10 +741,14 @@ iocontrol. These are relatively low speed events, high speed coordinated I/O is 
 
 emctaskmain.cc sends I/O commands via taskclass.cc.
 
-iocontrol main loop process:
+Historically this was a separate process named iocontrol, talking to
+task over its own NML channels; the HAL component is still named
+iocontrol.0 (and the INI section [EMCIO]) for compatibility, but the
+code now runs inside the task process. On each pass through the task
+loop it:
 
-- checks to see it HAL inputs have changed
-- checks if read_tool_inputs() indicates the tool change is finished and set emcioStatus.status
+- checks whether the iocontrol HAL input pins have changed
+- checks if read_tool_inputs() indicates the tool change is finished and sets emcioStatus.status
 
 == User Interfaces
 
@@ -844,7 +871,10 @@ The original NIST format of the buffer line is:
 * 'B name type host size neut RPC# buffer# max_procs key [type specific configs]'
 * 'B' - identifies this line as a Buffer configuration.
 * 'name' - is the identifier of the buffer.
-* 'type' - describes the buffer type - SHMEM, LOCMEM, FILEMEM, PHANTOM, or GLOBMEM.
+* 'type' - describes the buffer type - SHMEM, LOCMEM, or PHANTOM.
+  (Older RCS documents also list FILEMEM, GLOBMEM and RTLMEM; none of these is
+  implemented in src/libnml/ - cms_config() rejects RTLMEM explicitly and does
+  not recognise the other two at all.)
 * 'host' - is either an IP address or host name for the NML server
 * 'size' - is the size of the buffer
 * 'neut' - a boolean to indicate if the data in the buffer is encoded in a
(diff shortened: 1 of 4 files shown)


docs/src/code/code-notes.adoc
```diff
@@ -240,10 +240,12 @@ There are approximately 44 commands - this list is still under
 construction.
 
 [NOTE]
-The cmd_code_t enumeration, in motion.h, contains 73 commands, but the switch
-statement in command.c contemplates only 70 commands (as of 6/5/2020).
-ENABLE_WATCHDOG / DISABLE_WATCHDOG commands are in motion-logger.c. Maybe they are obsolete.
-The SET_TELEOP_VECTOR command only appears in motion-logger.c, with no effect other than its own log.
+The cmd_code_t enumeration, in motion.h, contains 76 commands, but the switch
```
grandixximo: 76/73 and "as of 8/2026" will go stale exactly like 73/70 and 6/5/2020 did; naming the three unhandled commands without the arithmetic stays true indefinitely. Nothing in the tree sends them, see #4359.

docs/src/code/code-notes.adoc
```diff
@@ -370,8 +372,10 @@ The ENABLE command enables the motion controller.
 
 ==== Requirements
 
-None. The command can be issued at any time, and will always be
-accepted.
+The command is rejected with an error message ("can't enable motion,
+enable input is false") if the motion.enable HAL pin is low. The
```
grandixximo: `motion.enable` is created with default value 1 (motion.c:528), so unconnected reads TRUE and this is never hit on a normal machine; "the hardware enable chain must be satisfied" overstates it.

docs/src/code/code-notes.adoc
```diff
@@ -500,10 +504,24 @@ future.
...
 === PAUSE
 
-The PAUSE command stops the trajectory planner. It has no effect in
-free or teleop mode. At this point I don't know if it pauses all motion
-immediately, or if it completes the current move and then pauses before
-pulling another move from the queue.
+The PAUSE command pauses the trajectory planner. It has no effect in
+free or teleop mode.
+
+The machine neither stops instantly nor completes the current move: the
```
grandixximo, line 472: This is planner behaviour rather than PAUSE-command behaviour; two sentences plus the synchronized-motion exception would cover what a Code Notes reader needs.

docs/src/code/code-notes.adoc
```diff
@@ -500,10 +504,24 @@ future.
...
+planner treats pause as a request to bring the feed to zero, so the
+machine decelerates to a stop within the current segment, at that
+segment's acceleration limit, and halts wherever the deceleration ramp
+ends. The run-down is jerk-limited only on machines configured for the
+S-curve planner ([TRAJ]PLANNER_TYPE = 1 with a non-zero MAX_JERK); both
```
grandixximo: Duplicates ini-config.adoc:947, and the threshold is `MAX_JERK < 1.0` (inihal.cc:302 and :320), not non-zero.

docs/src/code/code-notes.adoc
```diff
@@ -500,10 +504,24 @@ future.
...
+default to off, so a stock machine decelerates on a trapezoidal profile.
+
+Pause is ignored while the active segment is position-synchronized with
+the spindle (spindle-synchronized motion such as G33 threading and
+rigid tapping): for those segments the planner forces full feed and
+bypasses both pause and feed override, because tool position is slaved
+to spindle angle and pausing mid-thread would destroy the work.
+The pause takes effect when the synchronized segment ends.
+Velocity-synchronized segments (G96 style) can be paused normally.
```
grandixximo: This is G95, not G96: `TC_SYNC_VELOCITY` comes from `velocity_mode = 1` at emccanon.cc:521 and :530.

docs/src/code/code-notes.adoc
```diff
@@ -887,7 +914,10 @@ The original NIST format of the buffer line is:
 * 'B name type host size neut RPC# buffer# max_procs key [type specific configs]'
 * 'B' - identifies this line as a Buffer configuration.
 * 'name' - is the identifier of the buffer.
-* 'type' - describes the buffer type - SHMEM, LOCMEM, FILEMEM, PHANTOM, or GLOBMEM.
+* 'type' - describes the buffer type - SHMEM, LOCMEM, or PHANTOM.
+  (Older RCS documents also list FILEMEM, GLOBMEM and RTLMEM; none of these is
```
grandixximo: Trimming the list is right, but the parenthetical puts the document back to tracking its own history; also "recognise" is the only British spelling in the file.

docs/src/code/code-notes.adoc
```diff
@@ -1387,9 +1415,10 @@ FIXME: `axis_mask` and `axes` overspecify the number of axes
   An array of `EMCMOT_MAX_JOINTS` joint structures.
   `joint[0]` through `joint[joints-1]` are valid, the others do not exist on this machine and must be ignored.
 
-Things are not this way currently in the joints-axes branch, but deviations from this design are considered bugs.
-For an example of such a bug, see the treatment of axes in src/emc/ini/initraj.cc:loadTraj().
-There are undoubtedly more, and I need your help to find them and fix them.
+Deviations from this design are considered bugs; if you find one,
+please report it. (An earlier revision of this document pointed at the
```
grandixximo: Same as above, the parenthetical keeps the removed bug report alive; the sentence before it is enough on its own.

docs/src/gcode/g-code.adoc
```diff
@@ -1109,9 +1109,10 @@ M2 (end program)
 It is an error if:
 
 * All axis words are omitted.
-* The spindle is not turning when this command is executed.
-* The requested linear motion exceeds machine velocity limits
-  due to the spindle speed.
+* No K word is given.
+* An F word is given (the feed follows from K and the spindle speed).
```
grandixximo, line 1113: Additions check out (interp_check.cc:376 and :377). Could the deleted velocity-limit line stay as a NOTE though? It is real behaviour when pitch times RPM exceeds the axis limit, and it was the only mention in the manual.

docs/src/code/code-notes.adoc
Zia-research: Agreed — the arithmetic is the part that rots. Reworded to name the three commands, with no counts and no date. I had not read #4359; if that lands the note can go away entirely, so I have written this so it does not need revisiting either way.

docs/src/code/code-notes.adoc
Zia-research: You are right, it overstates. `motion.c:527` says so in a comment of its own — default TRUE so simple machines can leave it disconnected. Reworded.

docs/src/code/code-notes.adoc
Zia-research: Fair. Cut to two sentences plus the synchronized-motion exception.

docs/src/code/code-notes.adoc
Zia-research: The sentence is gone rather than corrected: it went with the trimming you asked for on the PAUSE entry, so the duplicate resolves itself. That left the threshold stated only in `ini-config.adoc`, and stated wrongly, so I fixed it there instead — it now reads 1.0 or above, and says that a smaller value drops back to trapezoidal with no message, which `initraj.cc` admits in a FIXME of its own.

One correction that lands on both of us: the threshold guards `[TRAJ]MAX_LINEAR_JERK`. `MAX_JERK` is a different key, per axis and per joint. My patch had the right number and the wrong name, and the review inherited it. This makes the patch touch a third file — say the word if you would rather have that as a separate PR.

docs/src/code/code-notes.adoc
Zia-research: Straight error on my side, corrected to G95. I followed it through afterwards: `G_95` → `enqueue_SET_FEED_MODE` → `canon.feed_mode` → `START_SPEED_FEED_SYNCH` → `velocity_mode` → `TC_SYNC_VELOCITY`. G96 goes through `convert_spindle_mode`, which is unrelated. Thank you.

docs/src/code/code-notes.adoc
Zia-research: Both taken, and they turn out to be the same change: `recognise` was inside the parenthetical, so dropping it took the spelling with it. No British spelling left in the file.

docs/src/code/code-notes.adoc
Zia-research: Same, dropped.

docs/src/gcode/g-code.adoc
Zia-research: Yes, and it is worse than one line: the same sentence was at 1113 and at 1179, and the patch removed both, so it left the manual altogether rather than one of the two places. Restored as a NOTE in both.

One caveat, because I have not verified the mechanism: there is no interpreter check and no error message for that case, so if it is real it is a silent clamp in the planner. The NOTE therefore says only that the program is not rejected, rather than asserting what happens instead.

Zia-research:
Branch force-pushed — the eight comments above are answered inline.

One thing that needs saying across all of them: the patch now touches a third file. Checking your jerk comment showed `ini-config.adoc` carried the same wrong threshold, so dropping mine as a duplicate would have left the wrong one as the only statement in the manual. I fixed it there instead. If you would rather that travelled as its own PR, say so and I will split it out.

Thank you for the review, and for the answers on #3718.

grandixximo:
We discussed this at the Sunday meeting, I tasked myself to do code changes where seen as a better alternative to documenting bad behavior, #4390 and #4392 are merged, your doc changes are now stale, do a rebase and fix conflicts, the note about the dead commands should be gone since the commands are gone, and the Planner silent switching, is no more silent. The G33 note can stay in, and we will have to see if a warning is worth the effort there, or leave it as a documented unchecked behavior.

Zia-research:
Thanks for the code changes — a better outcome than documenting the behaviour.

Rebased on master, force-pushed.

The dead-commands note is gone: #4392 removed the section, so there was nothing left to correct. The "no message is issued" clause is gone too — #4390 added one, so I reversed the sentence rather than dropping it. G33 note kept.

Four text corrections left, all in docs.

grandixximo (approved):

PR #4350, opened 2026-08-06
Split toggle2nist component

Sigma1912:
Creates separate components for handling toggle buttons and momentary buttons

- Restores the legacy functionality (as in 2.9) of 'toggle2nist' with added input debounce and a maximum output pulse length to avoid jamming up the component if 'is-on' fails to follow the 'on' / 'off' output signal.
- Fixes the toggle2nist man page description so it matches actual (legacy) behavior.
- Adds a new 'momentary2nist' component that does what the old 'toggle2nist' docs described plus the debounce and max puls length described above.

### **Background**

Before #3193 'toggle2nist' component docs described the functionality as converting
momentary inputs to nist outputs but the actual functionality was converting toggled
inputs to nist outputs. This required using a preceding 'toggle' component to match
documented behavior.

The fix in #3193 changed the behavior to match the documentation (and added debouncing
to the inputs). While this fixed the issues reported on the forum it breaks functionality
for hal files that use the preceeding 'toggle' component as a workaround.
Also the name 'toggle2nist' does no longer reflect the actual function.

### **What is 'nist'**
The term 'nist' seems to have been invented by the original author of the 'toggle2nist' component.
From the description:

> [...] device that has separate on and off inputs
and an is-on output

### Alternative
The 'toggle' and 'momentary' functionality could be combined in a single component.
However, due to the very descriptive component name, this new component should probably not be called 'toggle2nist'. Changing the name would then of course require an update to existing hal files.

**See previous discussion: https://github.com/LinuxCNC/linuxcnc/pull/4223** 

### Tested

Both components have been tested on real hard ware using mechanical switches.

### DIFF
--- a/src/hal/components/momentary2nist.comp
+++ b/src/hal/components/momentary2nist.comp
@@ -1,0 +1,85 @@
+component momentary2nist "momentary button to nist logic";
+
+description
+"""
+Momentary2nist can be used with a momentary push button
+to control a device that has separate on and off inputs
+and an is-on output.
+A debounce delay in cycles can be set for 'in'. (default = 2)
+A maximum output pulse length in cycles can be set for 'on' and 'off'. (default = 100)
+
+* On a rising edge on pin *in* when *is-on* is low: It sets *on* until *is-on* becomes high.
+* On a rising edge on pin *in* when *is-on* is high: It sets *off* until *is-on* becomes low.
+* Pins *on* and *off* remain high until *is-on* confirms the respective state or until *max-pulse-length* is reached.
+....
+       ┐     ┌─────xxxxxxxxxxxx┐           ┌─────xxxxxxxxxxxx┐
+in   : └─────┘     xxxxxxxxxxxx└───────────┘     xxxxxxxxxxxx└─────
+
+       ┐     ┌───────────┐
+on   : └─────┘           └─────────────────────────────────────────
+
+       ┐                                   ┌───────────┐
+off  : └───────────────────────────────────┘           └───────────
+
+       ┐                 ┌─────────────────────────────┐
+is-on: └─────────────────┘                             └───────────
+....
+
+""";
+
+pin in bool in  "momentary button in";
+pin in bool is_on "current state of device";
+pin in ui32 debounce = 2 "debounce delay for 'in'-pin in cycles";
+pin in ui32 max_pulse_length = 100 "max output pulse length";
+pin out bool on "turn device on";
+pin out bool off "turn device off";
+variable unsigned debounce_cntr;
+variable unsigned pulse_length;
+variable bool state;
+
+option period no;
+function _;
+license "GPL";
+author "David Mueller";
+;;
+FUNCTION(_)  {
+
+    rtapi_bool in_val = in;
+    rtapi_bool ison_val = is_on;
+    rtapi_uint debounce_val = debounce;
+
+    if (( debounce_val < 1 ) || ( debounce_val > 10000 )) {
+        debounce_val = 2; // set a sane value
+    }
+
+    if (in_val && state == 0 )  { // input has changed from debounced 0 -> 1
+        debounce_cntr++;
+        if ( debounce_cntr >= debounce_val ) {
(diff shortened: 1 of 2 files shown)


src/hal/components/momentary2nist.comp
```diff
@@ -0,0 +1,84 @@
...
+license "GPL";
+author "David Mueller";
+;;
+FUNCTION(_)  {
+
+    if (( debounce < 1 ) || ( debounce > 10000 )) {
+        debounce_val = 2; // set a sane value
+    } else {
+        debounce_val = debounce;
+    }
```
BsAtHome: May I suggest (reads `debounce` only once):
```c
rtapi_u32 debounce_val = debounce;
if(debounce_val < 1 || debounce_val > 10000) {
    debounce_val = 2; // set a sane value
}
```

You may also want to cache `is_on` and `in`:
```c
rtapi_bool inval = in;
rtapi_bool isonval = is_on;
```
and then use the cached values.

src/hal/components/momentary2nist.comp
```diff
@@ -0,0 +1,84 @@
...
+pin in bool is_on "current state of device";
+pin in ui32 debounce = 2 "debounce delay for 'in'-pin in cycles";
+pin in ui32 max_pulse_length = 100 "max output pulse length";
+pin out bool on "turn device on";
+pin out bool off "turn device off";
+variable int debounce_cntr;
+variable unsigned debounce_val;
+variable unsigned pulse_length;
+variable int state;
+
```
BsAtHome: `state`follows `in`, which is a bool. Why isn't `state` a bool? Is there a specific reason for `state` to be an integer?

src/hal/components/momentary2nist.comp
```diff
@@ -0,0 +1,84 @@
...
+pin in bool in  "momentary button in";
+pin in bool is_on "current state of device";
+pin in ui32 debounce = 2 "debounce delay for 'in'-pin in cycles";
+pin in ui32 max_pulse_length = 100 "max output pulse length";
+pin out bool on "turn device on";
+pin out bool off "turn device off";
+variable int debounce_cntr;
+variable unsigned debounce_val;
+variable unsigned pulse_length;
```
BsAtHome: `debounce_val` is unconditionally overwritten in the function. Having it as as variable here is expensive. Better use a local variable (see also other comment).

Sigma1912:
@BsAtHome 
Thank you for the input! 

I'm a bit fuzzy on the variable types.
Maybe this needs an update: https://github.com/LinuxCNC/linuxcnc/blob/master/docs/src/hal/comp.adoc

Question:
You suggest `rtapi_u32 debounce_val = debounce;` 
I wonder if I should  use `int debounce_val = debounce;` instead, so I don't need to cast to 'int' here `if ( debounce_cntr >= (int)debounce_val ) {`

Also, should I change these things in 'toggle2nist.comp' as well?

BsAtHome:
The pin type is `ui32` (previously called u32), which is an <i>unsigned 32-bit value</i>. The RTAPI type to hold that value is `rtapi_u32`. Using an integer (int) would mean a difference going from unsigned to signed. You can also use `rtapi_uint`, which is the larger unsigned type.

Going from signed to unsigned or unsigned to signed is most of the time an error. There are legitimate constructs that would require it, but usually, an unsigned construct remains unsigned and a signed construct remains signed.
Mixing the two without careful design risks a lot of potential errors, where the "integer overflow" is probably the best known.

Please note that the old documentation is not yet up-to-date with all the changes that have been going on in HAL. Working on it...

src/hal/components/momentary2nist.comp
Sigma1912: Changed as suggested

src/hal/components/momentary2nist.comp
Sigma1912: Done

src/hal/components/momentary2nist.comp
Sigma1912: Fixed. Leftover from original component.

PR #4355, opened 2026-08-07
docs: add ini-file API documentation for linuxcnc.ini module

BsAtHome:
The linuxcnc python module documentation lacked the API documentation of the new INI-file parser and implementation. This PR adds it.

A small addition to the main index adds a local link from the "Developer information" to the "User Interface programming" and appends the `(API doc)` text to the python module lines. This makes it easier to find the python API references (just a quick-fix).

### DIFF
--- a/docs/src/config/python-interface.adoc
+++ b/docs/src/config/python-interface.adoc
@@ -848,38 +848,190 @@ if error:
 ----
 
 [[python:reading-ini-values]]
-== Reading INI file values
+== Reading INI file values using `linuxcnc.ini`
 
-Here's an example for reading values from an INI file through the `linuxcnc.ini` object:
+The `linuxcnc.ini` module includes all functions to read values and information from INI-files.
+
+Reading an INI-file is done by creating an instance of the module class:
+
+[source,python]
+----
+import linuxcnc
+
+try:
+    ini = linuxcnc.ini("/path/to/ini-file.ini")
+except linuxcnc.error:
+    print("Ini-file not found or invalid")
+----
+
+The `ini` instance can then be used to extract values and other information from the INI-file.
+
+[IMPORTANT]
+====
+You should not roll your own type conversion functions to read boolean, integer or floating point values.
+The INI-file parser and Python wrapper already include these functions.
+
+Using the built-in functions guarantees consistent interpretation of values, also across locales.
+There are also helpers for enumerated types to read and convert them correctly and consistently.
+====
+
+=== `linuxcnc.ini` Methods
+
+All methods will throw an error on the following conditions:
+
+* `TypeError`, `ValueError`, `UnicodeError`, `OverflowError`: an argument was of the wrong type or value,
+* `linuxcnc.error`: the _num_ argument, if given, was less or equal zero (0).
+
+[NOTE]
+====
+All methods that retrieve a `[SECTION]VARIABLE` value can provide an empty string ("") for the _section_.
+In that case, the first matching _variable_ found in any section of the INI-file is used or returned.
+
+This behavior is for compatibility only.
+You should not rely on this property.
+====
+
+For all methods, the _fallback_= value, if given, may be of any type.
+It is returned verbatim if the method otherwise would have failed or returned `None`.
+
+Conversion to both integer and floating point will ignore trailing content if it is separated by whitespace.
+
+.Example
+[source,ini]
+----
+[BLUEBERRY]
+PIE = 3.14159265 This is _/*very*/_ bad style but accepted
+FAIL= 0.12345#Warning will be issued
+----
+
+The above example will correctly convert when using `ini.getreal("BLUEBERRY","PIE")` and the trailing content is ignored.
+A warning message (`inifile:line: warning: Trailing character(s)...`) will be reported on `ini.getreal("BLUEBERRY","FAIL")`.
+
+==== Method details
+
+ini = linuxcnc.ini(_filename_:string)::
+  Creates an INI-file instance reading INI-file _filename_.
+  Throws a `linuxcnc.error` if the file does not exist or cannot be parsed.
+
+bool = ini.hassection(_section_:string)::
+  Returns a boolean indicating whether or not the given _section_ was found in the ini-file.
+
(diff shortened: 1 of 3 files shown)


docs/src/config/python-interface.adoc
```diff
@@ -848,37 +848,192 @@ if error:
...
 # run as:
 # python3 ini-example.py ~/emc2-dev/configs/sim/axis/axis_mm.ini
 
 import sys
 import linuxcnc
 
-inifile = linuxcnc.ini(sys.argv[1])
+try:
+    inifile = linuxcnc.ini(sys.argv[1])
+except linuxcnc.error, detail:
```
grandixximo: Should this be `except linuxcnc.error as detail:`? Under python3 this line is a SyntaxError, so the example won't even parse as written.

docs/src/config/python-interface.adoc
```diff
@@ -848,37 +848,192 @@ if error:
...
-
 extensions = inifile.findall("FILTER", "PROGRAM_EXTENSION")
-print("extensions: ", extensions)
+print("extensions  :", extensions)
 
 # override default NML file by INI parameter if given
-nmlfile = inifile.getstring("EMC", "NML_FILE", fallback="")
-if nmlfile:
+nmlfile = inifile.getstring("EMC", "NML_FILE")
+if None != nmlfile:
```
grandixximo: Any reason not to write this as `if nmlfile is not None:`? That's the idiomatic python form for None checks; do you think it fits better here?

docs/src/config/python-interface.adoc
```diff
@@ -848,37 +848,192 @@ if error:
...
+
+string|None = ini.find(_section_:string, _variable_:string [, _num_:int] [, _fallback_=])::
+  Alias of `ini.getstring()`.
+
+list = ini.findall(_section_:string [,_variable_:string])::
+  Find value(s) from named _section_ as a list matching the optional _variable_ name.
+
+tuple(filename,lineno) = ini.lineof(_section_:string, _variable_:string [, _num_:int])::
+  Returns a tuple with the filename and line number of _num_'th _variable_ in the _section_.
+  The first matching _section_ _variable_ is returned if _num_ if not provided.
```
grandixximo: Typo: "if _num_ if not provided", the second "if" should be "is".

docs/src/config/python-interface.adoc
```diff
@@ -848,37 +848,192 @@ if error:
...
+For all methods, the _fallback_= value, if given, may be of any type.
+It is returned verbatim if the method otherwise would have failed or returned `None`.
+
+Conversion to both integer and floating point will ignore trailing content if it is separated by whitespace.
+
+.Example
+[source,ini]
+----
+[BLUEBERRY]
+PIE = 3.14159265 This is _/*very*/_ bad style but accepted
```
grandixximo, line 903: Listing blocks don't process inline markup, so `_/*very*/_` renders literally in the HTML. Intentional, or leftover markup to strip?

docs/src/config/python-interface.adoc
```diff
@@ -848,37 +848,192 @@ if error:
...
+==== Method details
+
+ini = linuxcnc.ini(_filename_:string)::
+  Creates an INI-file instance reading INI-file _filename_.
+  Throws a `linuxcnc.error` if the file does not exist or cannot be parsed.
+
+bool = ini.hassection(_section_:string)::
+  Returns a boolean indicating whether or not the given _section_ was found in the ini-file.
+
+bool = ini.hasvariable(_section_:string, _variable_:string)::
```
grandixximo: The binding accepts an optional third arg (`s|si` in emcmodule.cc), i.e. `hasvariable(section, variable [, num])`. Deliberately omitted, or worth documenting?

docs/src/config/python-interface.adoc
```diff
@@ -848,37 +848,192 @@ if error:
...
 extensions = inifile.findall("FILTER", "PROGRAM_EXTENSION")
-print("extensions: ", extensions)
+print("extensions  :", extensions)
 
 # override default NML file by INI parameter if given
-nmlfile = inifile.getstring("EMC", "NML_FILE", fallback="")
-if nmlfile:
+nmlfile = inifile.getstring("EMC", "NML_FILE")
+if None != nmlfile:
     linuxcnc.nmlfile = os.path.join(os.path.dirname(sys.argv[1]), nmlfile)
```
grandixximo: `os` is used here but never imported in this example. Add `import os` above? (Same gap exists in the old text, so this predates your change.)

docs/src/config/python-interface.adoc
BsAtHome: I know, it was intentional.

docs/src/config/python-interface.adoc
BsAtHome: Yes, indeed. However, the NML test has been scrapped (see also other comment).

The same issue is later down on the intval. Fixed.

docs/src/config/python-interface.adoc
BsAtHome: That was a copy/paste from the old version. Updated to be functional and tested it too.

docs/src/config/python-interface.adoc
BsAtHome: Check.

docs/src/config/python-interface.adoc
BsAtHome: That is an omission. Actually, it is worse because the arg parse should be `ss|i`. Both section and variable must be present as argument. The num is optional.

And then there was a second bug in emcmodule's implementation initializing the wrong var. All fixed. Also added the fix in emcmodule.cc in this PR.

docs/src/config/python-interface.adoc
BsAtHome: Actually, removed the nml stuff from the example. We shouldn't be altering NML anyway.

BsAtHome:
Pushed the update, including the fix in emcmodule.cc.

docs/src/config/python-interface.adoc
grandixximo: Agreed

PR #4357, opened 2026-08-08
gmoccapy: don't show_all() containers holding embedded tabs

### LINKED ISSUE #4353: Gmoccapy: EMBED_TAB create double "Y button" in Preview
RIP instalation
master branch

load LCNC with sim configuration:
gmoccapy_with_user_tabs.ini

Image

After comment "EMBED_TAB_LOCATION = ntb_preview" second Y dissapier.

grandixximo:
Fixes #4353

_init_dynamic_tabs() called show_all() on the container receiving an
embedded tab. show_all() is recursive, so with
EMBED_TAB_LOCATION = ntb_preview it undid the hides gmoccapy applied at
startup, e.g. the lathe-only back-tool view button rbt_view_y2, which
then showed up as a bogus second "Y" view button on mill configs.

Show only the container itself plus the added socket and its tab label
instead.

Tested with gmoccapy_with_user_tabs.ini and works for me: single "Y"
button on the preview row, both embedded tabs still populate. Back-tool
lathe configs keep their two Y views.

@zz912 could you test this with your config?

### DIFF
--- a/src/emc/usr_intf/gmoccapy/gmoccapy.py
+++ b/src/emc/usr_intf/gmoccapy/gmoccapy.py
@@ -1828,7 +1828,8 @@ class gmoccapy(object):
                 cmd = c.replace('{XID}', str(xid))
                 child = subprocess.Popen(cmd.split())
                 self._dynamic_childs[xid] = child
-                nb.show_all()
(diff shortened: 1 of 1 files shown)


zz912:
Thak you for fix. It works.

PR #4358, opened 2026-08-08
docs(build): stage English sources with per-file rules so one make rebuilds

### LINKED ISSUE #4354: Doc html re-build unstable
When you have build the HTML docs, then rebuilding them is not stable.

```sh
$ configure --enable-build-documentation=html
$ make -O -j8
$ touch ../docs/src/config/python-interface.adoc
$ make -O -j8
# nothing happens, only deps are checked and the setuid/setcap message
$ make -O -j8
# now the rebuild happens
```

So, the second invocation of make does start the rebuild.

@grandixximo any guess?

grandixximo:
Fixes #4354.

@BsAtHome I reproduced your recipe exactly: after a full HTML build, `touch ../docs/src/config/python-interface.adoc` then `make` re-stages the file but renders nothing, and only the run after that rebuilds the page.

The cause is mine, from c774b793f9. English sources are staged into `docs/build/adoc/en` by a bulk `find|tar` behind `.stage-en-stamp`, with the staged copies attached to it order-only:

```make
$(DOC_OUT_ADOC)/en/%.adoc: | $(DOC_BUILD)/.stage-en-stamp ;
```

make stats a prerequisite once per run and re-stats it only if it decided to remake it. An order-only edge never triggers a remake, so that empty-recipe rule was never "remade" and the staged `.adoc` kept its pre-staging mtime for the whole run. The stamp recipe rewrote that same file mid-run, but the renderer had already compared the `.html` against the stale mtime and concluded it was up to date. The next `make` saw the fresh mtime and rendered.

Replacing the stamp with one copy rule per staged file makes the change visible to make, so the render fires in the same run. `cp -p` also keeps the source mtime to the nanosecond, where `tar` rounded it down to the whole second; that could leave a staged copy a fraction older than a `.html` rendered from it in the same second and lose the edit outright.

Per-file rules alone are not enough: they stage only what some rule asks for, which is a smaller set than the bulk copy produced. asciidoctor resolves `include::` and `image:` against the staged tree at render time and most of those are not make prerequisites, so a partial like `hal/halshow.adoc` would never land and `hal/tutorial.html` would render an "Unresolved directive" placeholder instead of its content. A `stage-en` aggregate keeps the old guarantee.

Verified with HTML, PDF and the seven translations enabled: a clean build renders 5210 pages and 32 Master PDFs with no unresolved directives, the two builds after it do no work, and a touched source re-renders in a single `make` followed by a no-op.

Could you confirm this clears it on your machine?

One note for later: while tracking this down I found that `docs/src/asciideps` still strips a `.txt` suffix from when the docs were written in that format, so since the move to `.adoc` every rule it generates names a target nothing builds and the whole dependency extractor has been inert. Editing a chapter has not rebuilt the Master PDF that includes it. I have that fixed and verified on a follow-up branch, which I will open once this merges, since the two touch adjacent rules in the same file.

### DIFF
--- a/docs/src/Submakefile
+++ b/docs/src/Submakefile
@@ -976,7 +976,7 @@ $(DOC_FONT_DIR):
 # to rebuild on every run.  Order-only (after the |) instead; the SVGs
 # the PDF embeds are tracked via .adoc-images-stamp.
 define ASCIIDOCTOR_PDF_RULE
-$(4)/%.pdf: $(1)/%.adoc .adoc-images-stamp $$(DOC_FONTS) $$(DOC_EXTENSIONS) | svgs_made_from_dots
+$(4)/%.pdf: $(1)/%.adoc .adoc-images-stamp $$(DOC_FONTS) $$(DOC_EXTENSIONS) | svgs_made_from_dots stage-en
 	$$(ECHO) Building $$@
 	@mkdir -p $$(dir $$@)
 	@rm -f $$@ $$@.raw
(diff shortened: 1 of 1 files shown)


BsAtHome:
Works as expected. :+1:

PR #4361, opened 2026-08-08
docs(build): name the real target in the generated dependencies

grandixximo:
The follow-up I mentioned in #4358, now that it has merged. Thanks for the quick turnaround, @BsAtHome.

`docs/src/asciideps` derives the target it emits by stripping a `.txt` suffix, which is what the docs were written in when the script landed in 2011. Since the sources became `.adoc` nothing is stripped, so it generates rules for targets like

```
docs/build/adoc/en/hal/tutorial.adoc.html: ...
```

that no rule builds. Every `include::` and `image:` dependency it has extracted since then has been attached to a target nobody asks for, so the whole dependency extractor has been inert.

What that costs: editing `hal/halshow.adoc` never re-rendered `hal/tutorial.html`, which `include::`s it, and editing a chapter never rebuilt the Master PDF pulling it in. `Master_Documentation.adoc` alone `include::`s 119 files, so all four English PDFs stayed stale until `docclean`. Editing an image never rebuilt the page showing it either.

The target is not derivable from the source path, so it is passed in: a page is consumed by its `.html`, a Master document is never rendered as HTML and its English PDF lands in `objects/`. The unused `.dep` line went with the rewrite.

Reconnecting the dependencies exposed three latent problems, fixed in the same commit:

- Commented-out `image:` macros counted as dependencies. The macro is matched anywhere on the line so inline refs are caught, which also matched a ref behind a `//`. One in `gui/gmoccapy.adoc` points at an image deleted since it was commented out, so the build stopped on it.
- Images referenced inline were never staged into the translated adoc tree, because `.adoc-images-stamp` anchored its scan at the start of the line. The translated PDFs embed images from that tree, so they have been missing all 39 of them. Both extractors now key on the macro shape rather than a bare `image:`, so prose mentioning one is not taken for a path (`5-axis-kinematics.adoc` opens with "use image:: for equation png files" inside a `////` block).
- Staged translated images had no rule of their own. Named as prerequisites, a clean `-j` build reached one before the bulk stamp had written it and fell through to the "Required image file" error rule; it only succeeded on a second run. Each now stages on demand. The bulk pass stays, since asciideps records only the last macro on a line and cannot be the guarantee that asciidoctor-pdf finds every image.

Verified with HTML, PDF and the seven translations enabled: a clean build renders 5210 pages and 32 Master PDFs with no unresolved directives, the two builds after it do no work, and touching a page, an `include::`d partial, a Master chapter or an image rebuilds exactly what depends on it in a single `make`, the chapter reaching `objects/Master_Documentation.pdf`.

### DIFF
--- a/docs/src/Submakefile
+++ b/docs/src/Submakefile
@@ -1088,21 +1088,29 @@ $(DOC_OUT_HTML)/en/pdf/LinuxCNC_Manual_Pages_en.pdf: objects/LinuxCNC_Manual_Pag
 	    || (X=$$?; rm -f $@ $@.raw; exit $$X)
 	@test -f $@
 
+# Target a page's include:: and image: prerequisites hang off.  A page is
+# consumed by its .html; a Master document has no HTML, only the PDF that
+# pulls its chapters in, and the English ones render into objects/.  Not
+# derivable from the source path, so asciideps is told.
+EN_DEP_TARGET = $(if $(filter Master_%,$*),objects/$*.pdf,$(DOC_OUT_ADOC)/en/$*.html)
+
 depends/%.d: $(DOC_OUT_ADOC)/en/%.adoc $(DOC_SRCDIR)/asciideps .include-stamp | stage-en
 	$(ECHO) Depending $<
 	@mkdir -p $(dir $@)
-	$(Q)$(DOC_SRCDIR)/asciideps $< > $@.tmp
+	$(Q)$(DOC_SRCDIR)/asciideps $< $(EN_DEP_TARGET) > $@.tmp
 	@mv $@.tmp $@
 
 # Translated .adoc lives in $(DOC_OUT_ADOC)/<lang>/; same recipe, just a
 # different source dir so the per-language depends/<lang>/X.d targets
 # resolve.  The pattern only matches paths starting with a known language
 # tag, so the English depends/%.d above wins for non-translated entries.
+# Translated Master PDFs render beside their source, so unlike English the
+# target is the sibling .pdf.
 define TRANSLATED_DEP_RULE
 depends/$(1)/%.d: $(DOC_OUT_ADOC)/$(1)/%.adoc $(DOC_SRCDIR)/asciideps .include-stamp | stage-en
 	$$(ECHO) Depending $$<
 	@mkdir -p $$(dir $$@)
-	$$(Q)$$(DOC_SRCDIR)/asciideps $$< > $$@.tmp
+	$$(Q)$$(DOC_SRCDIR)/asciideps $$< $$(if $$(filter Master_%,$$*),$(DOC_OUT_ADOC)/$(1)/$$*.pdf,$(DOC_OUT_ADOC)/$(1)/$$*.html) > $$@.tmp
 	@mv $$@.tmp $$@
 endef
 $(foreach L,$(LANGUAGES),$(eval $(call TRANSLATED_DEP_RULE,$(L))))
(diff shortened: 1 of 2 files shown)


docs/src/asciideps
```diff
@@ -2,8 +2,20 @@
...
+# Emit a make rule attaching SOURCE's include:: and image: prerequisites to
+# TARGET.  The caller passes the targets because they are not derivable from
+# the source path: a page renders to <stem>.html next to its source, but the
+# Master documents render only as PDF and the English ones land in objects/.
+# With no TARGET the source stem plus .html is assumed, which is what a
+# standalone run of the script wants.
+SRC=$1
+test -z "$SRC" && exit 0
+test -f "$SRC" || exit 1
+shift
```
BsAtHome, line 14: If this is an error, wouldn't it be better to warn, like in:
```bash
[ -z "$SRC" ] && exit 0
[ -f "$SRC" ] || { echo "'asciideps: $SRC' not a file" >&2; exit 1; }
```

(and update the test to built-in bash)

docs/src/asciideps
```diff
@@ -31,15 +43,18 @@ includes () {
 
 images() {
 	DIR=$(dirname "$1")
-	sed -ne "s|^.*image:\{1,2\}\([^[]*\)\[.*\].*$|$DIR/\1|p" "$1" | tr '\n' ' '
+	# Drop AsciiDoc line comments first.  The macro is matched anywhere on the
+	# line so inline image: refs are caught, which without this also matches a
+	# commented-out one.  That is not a dependency, and naming it as a
+	# prerequisite fails the build once the file it points at is removed.
+	grep -v '^[[:space:]]*//' "$1" \
+		| sed -ne "s|^.*image:\{1,2\}\([^[]*\)\[.*\].*$|$DIR/\1|p" | tr '\n' ' '
```
BsAtHome: Mode: Push
Mode: Pedantic: ON

If you put one pipe on a new line, wouldn't it be better to put the next one on a new line too?

Mode: Pop

docs/src/Submakefile
```diff
@@ -1088,21 +1088,25 @@ $(DOC_OUT_HTML)/en/pdf/LinuxCNC_Manual_Pages_en.pdf: objects/LinuxCNC_Manual_Pag
 	    || (X=$$?; rm -f $@ $@.raw; exit $$X)
 	@test -f $@
 
+# Which target a page's include:: and image: prerequisites belong to.  A page is consumed by its .html; a Master document is never rendered as HTML, only as the PDF that pulls its chapters in, and the English Master PDFs render into objects/ rather than beside their source.  asciideps is told the target explicitly because none of that is derivable from the source path.
+EN_DEP_TARGET = $(if $(filter Master_%,$*),objects/$*.pdf,$(DOC_OUT_ADOC)/en/$*.html)
```
BsAtHome: This comment is a very loooooooooooooooong line...

docs/src/Submakefile
```diff
@@ -1191,14 +1195,17 @@ $(foreach L,$(LANGUAGES),$(eval $(call HTML_COPY_RULE,$(L))))
 # has no work to do (filter-out yields empty) so we skip the dep too.  HTML
 # image placement is no longer done here: the image_resolver rewrites HTML src
 # to the shared pool and .html-images-stamp materialises it.
+#
+# The macro is matched anywhere on the line so this sees the same set asciideps records as prerequisites.  Anchoring at the start of the line missed the 39 inline refs (`... e.g. image:images/p-term.png[height=25]`), which left those images out of the translated adoc tree and so out of the translated PDFs that embed them from there.
```
BsAtHome: Same here, very long comment line.
(and some other places too, no need for me to mark them all)

docs/src/asciideps
BsAtHome: (and fix my single quotes...)

grandixximo:
All four done: warn on a non-file with `[ ]` instead of `test`, quotes fixed, the second pipe on its own line, and the long comments cut down to the width of the rest of the file. Trimmed the ones I left in #4358 too.

Comment-only apart from the `asciideps` guard; output is byte-identical on a page, a Master and a translated source.

PR #4364, opened 2026-08-08
fix headless support

sleepybishop:
Hello! This PR cleans up and refactors the headless compilation support in LinuxCNC.

There is a need to run LinuxCNC on constrained or strictly headless hardware where X11, Tk, and GTK dependencies are unnecessary overhead.

To bridge this gap, this PR modifies the build system to allow LinuxCNC to be built and run completely without GUI dependencies (using `--disable-gui` and `--disable-gtk`).

My intent for this work is to allow orchestrating LinuxCNC entirely via `linuxcncrsh` and other remote control methods I am currently developing but this first pass just allows for a lighter weight build on devices that don't need the graphical tools, docs or man pages.

Please let me know if you have any feedback or suggestions on this approach or the build system modifications.


BsAtHome:
May be a good idea to have on the side. But it fails CI quite hard on the hostmot2 driver.

sleepybishop:
Thanks for looking, not quite ready for primetime. I'm working through the issues. Ill undraft the pr once I have it sorted.

BsAtHome:
I'm not sure why you suddenly need to check for `ld` and `objdump`. These are bog standard from binutils. Or is there something I'm missing here?

BsAtHome:
Something else... You can only control part of the machine through linuxcncrsh. A large part is controlled/monitored via HAL. For full headless/remote you probably also need halrmt. You are lucky, because that was just "fixed" (bugs notwithstanding).

Still, I'm not sure that everything is in place for real headless operation. Might need quite a bit of extra work to fill any gaps. Anyway, as I said, it still is a nice option to have available. So keep up the work.

sleepybishop:
Thanks for the feedback. To provide a bit more context, I have a fleet of machines with linuxcnc that I orchestrate remotely. I have a custom branch with a service that provides the equivalent of `halrmt` via a rest api, they also run a custom `linuxcncrsh` and a gcode streaming sink.

I'm in the process of extracting useful bits from my branch into upstream. I thought headless support would be a good next step but I accidentally pulled in other buildroot cross-compilation changes hence the `ld`/`objdump` stuff. The `hostmot` driver breakage was due to the inclusion of an old module I wrote that's not useful for upstream.

I've reverted all the changes that are not directly related to a light weight headless build.

src/configure.ac
```diff
@@ -810,6 +814,15 @@ else
 fi
 
 
+AC_ARG_ENABLE(gui,
+    AS_HELP_STRING([--disable-gui], [Disable building graphical user interfaces (Tk, Tcl, X11, Xinerama, GTK)]),
+    [BUILD_GUI=$enableval],
+    [BUILD_GUI=yes])
+if test "$BUILD_GUI" = "no"; then
+    BUILD_GTK=no
```
grandixximo: This forces `BUILD_GTK=no`, but the `AC_ARG_ENABLE(gtk)` a few lines below has `[BUILD_GTK=yes]` as its action-if-not-given (line 839), which runs whenever `--disable-gtk` is not passed explicitly and sets it straight back to `yes`. So `--disable-gui` alone still hits the `GTK3 missing` `AC_MSG_ERROR` on a GTK-less system, and only the combination of both flags (as in your PR description) actually works. Is requiring both flags the intent, or should the gtk default only apply when `BUILD_GUI=yes`?

src/emc/usr_intf/axis/Submakefile
```diff
@@ -27,7 +27,10 @@ $(TKDARMODULE): $(call TOOBJS, $(TKDARMODULESRCS))
 	$(ECHO) Linking python module $(notdir $@)
 	$(Q)$(CC) $(LDFLAGS) -shared -o $@ $(TCL_CFLAGS) $^ -lX11 $(TCL_LIBS)
 
-PYTARGETS += $(EMCMODULE) $(TOGLMODULE) $(TKDARMODULE)
+PYTARGETS += $(EMCMODULE)
+ifeq ($(BUILD_GUI),yes)
```
grandixximo, line 31: With `BUILD_GUI=no` the whole `emc/usr_intf/axis` directory drops out of `SUBDIRS`, so this Submakefile is never included at all. That makes this `ifeq` dead code, but more importantly it means `../lib/python/linuxcnc.so` (`EMCMODULE`) is no longer built in a headless build. `bin/mdi` and `bin/linuxcnctop` are still installed unconditionally and both do `import linuxcnc`, so they would fail at runtime, and any Python-based remote tooling loses the status channel too. Is dropping the Python module intentional for the headless use case, or should `EMCMODULE` move to a Submakefile that is always included?

src/configure.ac
```diff
@@ -696,7 +696,11 @@ AC_ARG_WITH([tcl-dir],
     [EMC2_TCL_DIR="$withval"],
     [TCL_GET_DIR=$(echo [['set d [lindex [lsearch -all -inline -glob $tcl_pkgPath /usr/share/*] 0]; if {$d eq ""} {set d [lindex [lsearch -all -inline -glob $tcl_pkgPath /usr/lib*] 0]}; puts $d']] | tclsh 2>/dev/null)
      AS_IF([test -z "$TCL_GET_DIR"],
-           [AC_MSG_ERROR([could not determine Tcl package dir from tcl_pkgPath])])
+           [
+            if test "$BUILD_GUI" = "yes"; then
+                AC_MSG_ERROR([could not determine Tcl package dir from tcl_pkgPath])
```
grandixximo, line 701: When tclsh is not installed at all (plausible on a strictly headless box), `TCL_GET_DIR` comes out empty and, with this error now suppressed, the next line sets `EMC2_TCL_DIR="$TCL_GET_DIR/linuxcnc"`, i.e. `/linuxcnc`. Since `tcldir = @EMC2_TCL_DIR@` and the `$(TCL)` / `objects/*.msg` installs in `src/Makefile` are unconditional, `make install` would then happily create `$(DESTDIR)/linuxcnc` and drop the tcl payload there. Should this get a sane fallback (or should those install lines move under the `BUILD_GUI` guard)?

src/Makefile
```diff
@@ -727,12 +735,15 @@ install-kernel-indep: install-dirs
 	$(FILE) ../tcl/TkLinuxCNC $(DESTDIR)/etc/X11/app-defaults
 	$(FILE) Makefile.modinc $(DESTDIR)$(datadir)/linuxcnc
 	$(EXE) $(TCL) $(DESTDIR)$(tcldir)
```
grandixximo: With `BUILD_GUI=no`, `../tcl/hal.so` and `../tcl/linuxcnc.so` are no longer built or installed, but the surrounding tcl payload (the `$(TCL)` scripts here, `$(TCL_BIN)`, `pkgIndex.tcl`, `objects/*.msg`, the scripts/*.tcl files) still is, and `pkgIndex.tcl` references the now-missing shared libraries. What consumes these files in a headless install? Would it be cleaner to move the whole tcl block under the same guard?

src/Makefile
```diff
@@ -799,41 +816,46 @@ install-python: install-dirs
...
 	$(EXE) ../bin/hal_bridge $(DESTDIR)$(bindir)
 	$(EXE) ../bin/mitsub_vfd $(DESTDIR)$(bindir)
 	$(EXE) ../bin/mqtt-publisher $(DESTDIR)$(bindir)
 	$(EXE) ../bin/z_level_compensation $(DESTDIR)$(bindir)
 	$(EXE) ../bin/pmx485 $(DESTDIR)$(bindir)
 	$(EXE) ../bin/sim-torch $(DESTDIR)$(bindir)
-	$(EXE) ../bin/pyvcp $(DESTDIR)$(bindir)
-	$(EXE) ../bin/gladevcp $(DESTDIR)$(bindir)
-	$(EXE) ../bin/axis $(DESTDIR)$(bindir)
 	$(EXE) ../bin/axis-remote $(DESTDIR)$(bindir)
```
grandixximo: `axis-remote` imports tkinter and talks to a running AXIS over X properties, and `hal_manualtoolchange` a few lines below pops a tkinter dialog. Both are still installed unconditionally. What are they for in a `--disable-gui` install? Should they move inside the `BUILD_GUI` block together with `axis` and `touchy`?

src/configure.ac
```diff
@@ -1075,12 +1088,22 @@ AC_ARG_ENABLE(build-documentation,
 # --enable-build-documentation=html).  =html itself warn-and-disables
 # only the missing PDF tools; users who want PDF best-effort can pass
 # =html and rebuild after installing the missing tools.
+AC_ARG_ENABLE(manpages,
+    [  --disable-manpages      Disable building man pages],
```
grandixximo: Nit: the neighboring options use `AS_HELP_STRING([--disable-foo], [description])`, which keeps `--help` output aligned; a literal string here will print as-is and drift out of alignment. Any reason not to use `AS_HELP_STRING` like the `--disable-gui` entry above?

grandixximo:
Looks ready, @BsAtHome do we want a CI job to test the `--disable-gui` path? I could do that follow up if we want it..

grandixximo (approved):

BsAtHome:
> Looks ready, @BsAtHome do we want a CI job to test the `--disable-gui` path? I could do that follow up if we want it..

The idea to build without GUI is a good one. But we might want to think this through and want to define where this needs to go (or just a one-off option?).

There are a few problems. You cannot currently build a package that is created with `--disable-gui`. There should probably be a mechanism to create a linuxcnc-nogui-\*.deb. And when we go that path, it might be a thought to think this through a bit further and instead create linuxcnc-base-\*.deb and then several linuxcnc-gui-XYZ\*.deb?

And, when you build without man-pages, then you cannot build a package because they are a requirement when building a package (Debian ant man-pages for each program). I'm not sure I see the reason for building without man-pages. These are probably the least of anybody's problem.

grandixximo:
I was just thinking rip build in CI to test the flag does not break in rip builds at least...
But yeah I see the point, bigger discussion for later....

sleepybishop:
Thanks for thinking through this folks. My intent was to implement this for source builds like my buildroot setup. My intent with the man pages was that they are simply not needed on a headless machine and omitting them results in a slimmer build.

For packaging, building everything and creating subpackages would be more appropriate as you suggested, (A core + ancillary packages), but that was out of scope from my intended changes.

Please let me know if source builds are discouraged, then I'll skip bringing over my buildroot cross-compilation changes and look for more meaningful changes to contribute.

grandixximo:
I don't think source builds changes or extensions are discouraged, building with cmake has been discussed multiple times, but I do think if we take this in as a supported flag, we should test that it keeps working. If most people show up at the meeting, we will probably merge I imagine, and maybe talk a bit about the packaging thing, I still don't see the strong points for it very clearly.

PR #4369, opened 2026-08-09
hal: Update the hostmot2 drivers to getter/setter

### LINKED ISSUE #4130: Use, abuse or erroneous implementation of overlapping hal_float_t/hal_u32_t
While doing the HAL getter/setter changes...

@pcw-mesa In `hal/drivers/mesa-hostmot2/abs_encoder.c` lines 617-624 we find this little gem:
```C
617:    case HM2_GTAG_FABS:
618:        if (chan->params->timer_num > 4) chan->params->timer_num = 4;
619:        if (chan->params->u32_param > 15) chan->params->u32_param = 15;
620:        buff3 = chan->num_read_bits << 24
621:                | (rtapi_u32)(8.0e-6 * hm2->absenc.clock_frequency) << 14;
622:        buff2 = chan->params->u32_param << 28
623:                | ((rtapi_u32)(0x100000 * (chan->params->float_param * 1000
624:                / hm2->absenc.clock_frequency)));
...
```
The addressed variables are part of a `union` in `hm2_sserial_params_t` defined in sserial.h:
* `chan->params->u32_param` - a parameter typed `hal_u32_t`
* `chan->params->float_param` - a parameter typed `hal_float_t`

Please note, these are addressing a HAL parameter, so they are visible from userland and can be changed and inspected by, f.ex, halcmd.

Both variables share the same underlying memory and the 32-bit unsigned is addressing some of the mantissa bits of the 64-bit floating point.

The code first uses the unsigned value in line 619 and rewrites it to be 15 when larger than 15, isolating 4 lower bits of the 32-bit value. Then, in line 622 it reads the value and takes the lower 4 bits to the top (the ones limited in line 619). However, the next line, 623, reads <i>the same memory</i> as a floating point value. This assumes that there is valid data in the higher 32 bits of the 64-bit memory read for the floating point value, which is doubtful, and uses the read value as is.

This is obviously either a bad hack or an error. What is the purpose of this code? How is this in any way predictable what will happen? If the parameter is created as a HAL_U32, then the upper 32-bit are undefined. If the parameter is created as HAL_FLOAT, then the lower 4 bits are for all practical purposes and intends not controllable.

Was this code written before the HAL_FLOAT became a double? If so, then there would have been a full variable size overlap. But then this code was never run or tested again after the change. Even so, I still don't get it because using a float is a potentially unstable (especially after division).

The question is now how to fix it. Insights what is actually required and suggestions are welcome.

### LINKED ISSUE #4144: Hostmot2 sserial state machine reassigns state unconditionally

BsAtHome:
This is the last in the series to update the code base to the HAL getter/setter infrastructure.

The hostmot2 drivers are a very large interconnected set of drivers in several layers that are, unfortunately, difficult to migrate step by step. This PR needs to be reviewed and tested on real hardware before it can be merged.

All changes have been tested on a RPi5 with an 7i96s board+modbus. It is therefore known that at least the basic outline of the driver is functional and behaves as it has done before.

The main issue in porting the hostmot2 drivers (and also others) has been the (ab)use of parameters. Other drivers had similar problems, but they could be handled one-by-one. Parameters are now similar to pins, which means all use of parameters (as expensive local variables) had to be detected and adapted.

An important test is with attached smart serial board(s). The sserial driver is known to have some quirks and has code that needed special attention while porting. Note that the sserial/abs_encoder part has been internally changed slightly. The Fanuc style absolute encoder had an overlapping value, which now has been untangled. It should not change behaviour or appearance for the user, but it is nice if it can be tested, just like the other absolute encoders.

It is also necessary to test this with a variety of boards, both networked and local PCI bus boards (and others). Also, the more configurations that can be tested, the better we can guarantee that no hidden defects are present.

@pcw-mesa you probably have most hardware, somewhere. Have a go, please, with different setups if possible.
@rene-dev you have tested earlier with a smart serial board, please try again.

Anybody else, please test and report.

Edit to add:
* The step_type parameter in `...stepgen.NN.step_type` is now capped at 16 to prevent read/write to unknown memory. There may be more failure modes when you set step_type to larger than 16, but that has not been considered.
* A bug in pwmgen that failed to consider all run-time config changes has been fixed (closing '}' misplaced). This should not alter operation or function.
* This does not fix #4144 and its behaviour is not changed to prevent mixing old and new failure modes.
* This PR does fix #4130 because the getter/setter update needed to untangle the access. The overlapped data access while also exporting that data as parameter does no longer function because parameters work similar as pins now.
* A bug in pwmgen when the firmware has no pwmgens was exposed with HAL's internal changes to add more checks. This has also been fixed. (see: https://github.com/LinuxCNC/linuxcnc/pull/4369#issuecomment-5236357655 below)

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/hostmot2.c
+++ b/src/hal/drivers/mesa-hostmot2/hostmot2.c
@@ -79,13 +79,13 @@ static void hm2_read_request(void *void_hm2, long period) {
     hm2->llio->period = period;
 
     // if there are comm problems, wait for the user to fix it
-    if ((*hm2->llio->io_error) != 0) return;
+    if (hal_get_bool(*hm2->llio->io_error)) return;
 
     hm2_tram_read(hm2);
-    if ((*hm2->llio->io_error) != 0) return;
+    if (hal_get_bool(*hm2->llio->io_error)) return;
     hm2_raw_queue_read(hm2);
     hm2_tp_pwmgen_queue_read(hm2);
-    if ((*hm2->llio->io_error) != 0) return;
+    if (hal_get_bool(*hm2->llio->io_error)) return;
     hm2_queue_read(hm2);
     hm2->llio->read_requested = true;
     hm2->llio->read_time = rtapi_get_time();
@@ -98,10 +98,10 @@ static void hm2_read(void *void_hm2, long period) {
     hm2->llio->read_requested = false;
 
     // if there are comm problems, wait for the user to fix it
-    if ((*hm2->llio->io_error) != 0) return;
+    if (hal_get_bool(*hm2->llio->io_error)) return;
     // if there's a temporary read failure, don't sweat it
     if(hm2_finish_read(hm2) == -EAGAIN) return;
-    if ((*hm2->llio->io_error) != 0) return;
+    if (hal_get_bool(*hm2->llio->io_error)) return;
 
     hm2_watchdog_process_tram_read(hm2);
     hm2_ioport_gpio_process_tram_read(hm2);
@@ -127,7 +127,7 @@ static void hm2_write(void *void_hm2, long period) {
     hostmot2_t *hm2 = void_hm2;
 
     // if there are comm problems, wait for the user to fix it
-    if ((*hm2->llio->io_error) != 0) return;
+    if (hal_get_bool(*hm2->llio->io_error)) return;
 
     if (!hm2->ddr_initialized) {
         hm2_ioport_initialize_ddr(hm2);
@@ -183,7 +183,7 @@ static void hm2_read_gpio(void *void_hm2, long period) {
     hostmot2_t *hm2 = void_hm2;
 
     // if there are comm problems, wait for the user to fix it
-    if ((*hm2->llio->io_error) != 0) return;
+    if (hal_get_bool(*hm2->llio->io_error)) return;
 
     hm2_ioport_gpio_read(hm2);
 }
@@ -193,7 +193,7 @@ static void hm2_write_gpio(void *void_hm2, long period) {
     hostmot2_t *hm2 = void_hm2;
 
     // if there are comm problems, wait for the user to fix it
-    if ((*hm2->llio->io_error) != 0) return;
+    if (hal_get_bool(*hm2->llio->io_error)) return;
 
     hm2_ioport_gpio_write(hm2);
     hm2_watchdog_write(hm2, period);
@@ -957,7 +957,7 @@ static int hm2_parse_module_descriptors(hostmot2_t *hm2) {
 
         md_accepted = hm2_ioport_parse_md(hm2, md_index);
 
-        if ((*hm2->llio->io_error) != 0) {
+        if (hal_get_bool(*hm2->llio->io_error)) {
             HM2_ERR("IO error while parsing Module Descriptor %d\n", md_index);
             return -EIO;
         }
@@ -1097,7 +1097,7 @@ static int hm2_parse_module_descriptors(hostmot2_t *hm2) {
 
         }
 
-        if ((*hm2->llio->io_error) != 0) {
+        if (hal_get_bool(*hm2->llio->io_error)) {
             HM2_ERR("IO error while parsing Module Descriptor %d\n", md_index);
             return -EIO;
         }
@@ -1450,22 +1450,16 @@ int hm2_register(hm2_lowlevel_io_t *llio, char *config_string) {
     //
 
     {
-        int r;
-        char name[HAL_NAME_LEN + 1];
-
-        llio->io_error = (hal_bit_t *)hal_malloc(sizeof(hal_bit_t));
+        llio->io_error = hal_malloc(sizeof(*llio->io_error));
         if (llio->io_error == NULL) {
             HM2_ERR("out of memory!\n");
             r = -ENOMEM;
             goto fail0;
         }
 
-        (*llio->io_error) = 0;
-
-        rtapi_snprintf(name, sizeof(name), "%s.io_error", llio->name);
-        r = hal_param_bit_new(name, HAL_RW, llio->io_error, llio->comp_id);
+        r = hal_param_new_bool(llio->comp_id, HAL_RW, llio->io_error, 0, "%s.io_error", llio->name);
         if (r < 0) {
-            HM2_ERR("error adding param '%s', aborting\n", name);
+            HM2_ERR("error adding param '%s.io_error', aborting\n", llio->name);
             r = -EINVAL;
             goto fail0;
         }
@@ -1676,7 +1670,7 @@ int hm2_register(hm2_lowlevel_io_t *llio, char *config_string) {
     // final check for comm errors
     //
 
-    if ((*hm2->llio->io_error) != 0) {
+    if (hal_get_bool(*hm2->llio->io_error)) {
         HM2_ERR("comm errors while initializing firmware!\n");
         goto fail1;
     }
@@ -1776,7 +1770,7 @@ void hm2_unregister(hm2_lowlevel_io_t *llio) {
         // if there's a watchdog, set it to safe the board right away
         if (hm2->watchdog.num_instances > 0) {
             hm2->watchdog.instance[0].enable = 1;
-            hm2->watchdog.instance[0].hal.param.timeout_ns = 1;
+            hal_set_ui32(hm2->watchdog.instance[0].hal.param.timeout_ns, 1);
             hm2_watchdog_force_write(hm2);
         }
 
--- a/src/hal/drivers/mesa-hostmot2/pwmgen.c
+++ b/src/hal/drivers/mesa-hostmot2/pwmgen.c
@@ -34,9 +34,9 @@ void hm2_pwmgen_handle_pwm_frequency(hostmot2_t *hm2) {
     rtapi_u32 dds;
 
 
-    if (hm2->pwmgen.hal->param.pwm_frequency < 1) {
-        HM2_ERR("pwmgen.pwm_frequency %d is too low, setting to 1\n", hm2->pwmgen.hal->param.pwm_frequency);
-        hm2->pwmgen.hal->param.pwm_frequency = 1;
+    if (hal_get_ui32(hm2->pwmgen.hal->param.pwm_frequency) < 1) {
+        HM2_ERR("pwmgen.pwm_frequency %d is too low, setting to 1\n", hal_get_ui32(hm2->pwmgen.hal->param.pwm_frequency));
+        hal_set_ui32(hm2->pwmgen.hal->param.pwm_frequency, 1);
     }
 
 
@@ -65,7 +65,7 @@ void hm2_pwmgen_handle_pwm_frequency(hostmot2_t *hm2) {
     //
 
     // can we do it with 12 bits?
-    dds = ((double)hm2->pwmgen.hal->param.pwm_frequency * 65536.0 * 4096.0) / (double)hm2->pwmgen.clock_frequency;
+    dds = ((double)hal_get_ui32(hm2->pwmgen.hal->param.pwm_frequency) * 65536.0 * 4096.0) / (double)hm2->pwmgen.clock_frequency;
     if (dds < 65536) {
         hm2->pwmgen.pwmgen_master_rate_dds_reg = dds;
         hm2->pwmgen.pwm_bits = 12;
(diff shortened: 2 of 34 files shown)


grandixximo:
Was wondering why it was so quiet here...
I see something was cooking...
Just a small patch hey...

Might be able to test on a 5i25 as well.

src/hal/drivers/mesa-hostmot2/hostmot2.c
```diff
@@ -1453,17 +1453,14 @@ int hm2_register(hm2_lowlevel_io_t *llio, char *config_string) {
...
             goto fail0;
         }
 
-        (*llio->io_error) = 0;
-
-        rtapi_snprintf(name, sizeof(name), "%s.io_error", llio->name);
-        r = hal_param_bit_new(name, HAL_RW, llio->io_error, llio->comp_id);
+        r = hal_param_new_bool(llio->comp_id, HAL_RW, llio->io_error, 0, "%s.io_error", llio->name);
         if (r < 0) {
             HM2_ERR("error adding param '%s', aborting\n", name);
```
grandixximo: `name` is never written now that the `rtapi_snprintf()` above it is gone, so this `%s` prints an uninitialised stack buffer.

    HM2_ERR("error %d adding param '%s.io_error', aborting\n", r, llio->name);

and the `char name[HAL_NAME_LEN + 1];` declaration can go with it.

src/hal/drivers/mesa-hostmot2/sserial.c
```diff
@@ -437,38 +437,43 @@ int hm2_sserial_create_params(hostmot2_t *hm2, hm2_sserial_remote_t *chan){
...
                                        global.NameString);
                 if (r < 0) {HM2_ERR("Out of memory\n") ; return -ENOMEM;}
+                break;
             case LBP_STREAM: // Don't anticipate seeing these as params
             case LBP_BOOLEAN:
             case LBP_ENCODER:
             case LBP_ENCODER_H:
             case LBP_ENCODER_L:
+            default:
+                // Still make sure we have memory allocated for a fake parameter
```
grandixximo, line 473: Every type ends up with real or fake storage here except `LBP_BITS`, which still falls out at `case LBP_BITS: break;` with `params[i].param` left NULL. Nothing dereferences it today (`get_param_value` and `register_tram` both hit an earlier break/default), but if the point of the fake param is that a reader can no longer tell the difference, is one hole in that invariant worth keeping?

src/hal/drivers/mesa-hostmot2/stepgen.c
```diff
@@ -396,73 +410,78 @@ static void hm2_stepgen_update_mode(hostmot2_t *hm2, int i) {
...
+    if (step_type > 16){
         HM2_ERR("the firmware only supports tables up to a depth of 16, you"
                 "requested %i. Reverting to step type 0\n",
-                inst->hal.param.step_type);
+                step_type);
         hm2->stepgen.mode_reg[i] = 0;
+        // Assure that we do not index outside the param.table[] array in the
+        // loop below. But the code may be off anyway and the result is
+        // anybody's guess.
+        step_type = 16;
```
grandixximo, line 471: Good catch: before this, `step_type > 16` indexed `param.table[j / 4]` past the array. Worth a line in the PR description though, since for a bogus `step_type` the driver now writes different data to the FPGA than before, which is exactly what a tester with an odd config would report as a regression.

src/hal/drivers/mesa-hostmot2/pwmgen.c
```diff
@@ -314,30 +314,30 @@ void hm2_pwmgen_write(hostmot2_t *hm2) {
...
     } 
     
     // update dither?
     for (i = 0; i < hm2->pwmgen.num_instances; i ++) {
-        if (hm2->pwmgen.instance[i].hal.param.dither != hm2->pwmgen.instance[i].written_dither) {
+        if (hal_get_bool(hm2->pwmgen.instance[i].hal.param.dither) != hm2->pwmgen.instance[i].written_dither) {
             goto force_write;
         }
 
     // check pwm & pdm frequency
```
grandixximo, line 335: Pre-existing, but this PR rewrote every line around it: the `for` at line 328 never got a closing brace, so these frequency checks and the enable loop below sit inside it. The inner loop leaves `i == num_instances`, so the outer loop runs exactly once and dither is only ever compared for instance 0. Fix it while it is open?

src/hal/drivers/mesa-hostmot2/sserial.c
```diff
@@ -1554,69 +1427,71 @@ int hm2_sserial_register_tram(hostmot2_t *hm2, hm2_sserial_remote_t *chan){
...
+                    hal_set_ui32(inst->state3, 1);
                     inst->timer = 20000000;
                     *inst->command_reg_write = 0x800; // stop all
                     break;
                  case 1:
                     ret = hm2_sserial_wait(hm2, inst, period);
                     if (ret > 0) break;
-                    if (ret < 0) *inst->state3 = 100; // quit and tidy up
-                    *inst->state3 = 2;
+                    // FIXME: Assigns 100 to state3 and then unconditionally assigns 2
```
grandixximo, line 1443: Six places do this. If `state3` is unconditionally overwritten immediately after, is state 100 reachable at all, i.e. is sserial silently ignoring every parameter-write failure today? If the FIXMEs mean "documented, fix later", would a follow-up issue keep them from reading as intentional?

src/hal/drivers/mesa-hostmot2/sserial.c
```diff
@@ -1554,69 +1427,71 @@ int hm2_sserial_register_tram(hostmot2_t *hm2, hm2_sserial_remote_t *chan){
                                 case 64:
                                     shift = 0;
                                 }
-                            if (xlabs((p->s64_param - p->s64_written) >> shift) > 2) break;
-                            *inst->state2 = 2; // increment indices
-                            return *inst->state2;
+                            // FIXME: This overlayed s64 read is very wrong!
```
grandixximo, line 1429: This reproduces the old `s64_param`/`float_param` union overlay exactly, so no regression. It survives only because `hal_get_sint()` and `hal_set_real()` happen to cover the same 8 bytes. Would comparing `p->param.r` against `float_written` with a tolerance derived from `shift` say what the code actually means?

src/hal/drivers/mesa-hostmot2/sserial.c
BsAtHome: This is addressed in #4144, but I didn't dare to change this before we have a functional driver.

src/hal/drivers/mesa-hostmot2/sserial.c
BsAtHome: This is exactly as it was. The overlaid access is an abomination, hence the FIXME comment. Howver, it needs to be fixed after we have this integrated.

BsAtHome:
For people who use stepgen and set the `...stepgen.NN.step_type` parameter to anything larger than 16 will now have it capped to 16. You'd be reading/writing data to undefined memory and can crash.
So, any change in step behaviour should be reported only when you have verified that you did not set step_type larger than 16 in your config. Otherwise you first need to fix the config and re-test.

src/hal/drivers/mesa-hostmot2/stepgen.c
BsAtHome: Well, it could have crashed... I'm not sure it would be correct previously and be a "regression". Note added to the PR.

src/hal/drivers/mesa-hostmot2/hostmot2.c
BsAtHome: Good catch... missed one.

src/hal/drivers/mesa-hostmot2/sserial.c
BsAtHome: Well seen. Fixed.

The point is that the sserial driver does so many odd things that it is not worth to take the risk with parameters. Better be safe than sorry is the motto here.

src/hal/drivers/mesa-hostmot2/pwmgen.c
BsAtHome: Indeed a pre-existing bug. Fixing it anyway because it is rather obvious.

src/hal/drivers/mesa-hostmot2/stepgen.c
grandixximo: Yeah, bad wording on my part...

src/hal/drivers/mesa-hostmot2/sserial.c
grandixximo: Fair

src/hal/drivers/mesa-hostmot2/sserial.c
grandixximo: Fair

pcw-mesa:
I can check some of this next week.

grandixximo:
Built and tested on the 5i25
One bug found: zero-instance module descriptors (PWMGen x0 on some firmware) abort board registration via the new hal_malloc() size check.
Workaround for me was num_pwmgens=0 , but it should be fixed, default is `-1` but it crashes, so limit to 0?

Also noticed a small detail while testing mode 20, we may want to fix here:
image
You requested is all in one word, the is a space missing...

Other than those minor things, it works well (on a bench test with only the 5i25 plugged in), tested a few firmware, stepgen pwmgen, encoder, all good.

I did what I could with sserial I don't actually have an sserial devices but the firmware loads ok, at least no regression there, Rene's symptoms from #4247 are untested without 7i76 or friends which I don't have.

Sigma1912:
Testing with 7i95T with a 7i84 connected on smart serial looks fine to me.
 
Screenshot from 2026-08-10 08-36-20

Connected input and output signals work correctly.


Branch tested:
```
user@debian:~/git/bs_hostmot/src$ git status
On branch halgs_hostmot2
Your branch is up to date with 'origin/halgs_hostmot2'.
```

BsAtHome:
> Built and tested on the 5i25 One bug found: zero-instance module descriptors (PWMGen x0 on some firmware) abort board registration via the new hal_malloc() size check. Workaround for me was num_pwmgens=0 , but it should be fixed, default is `-1` but it crashes, so limit to 0?

A genuine preexisting bug. Fixed. The first check only handled the config (when you set num_pwmgens=0). There is now a second check for zero pwmgens before the hal_malloc that checks the number actually present.

> Also noticed a small detail while testing mode 20, we may want to fix here:

A preexisting text error, also fixed.

grandixximo:
> A genuine preexisting bug.

Yeah, a bug that was not biting till your changes though...🐞

BsAtHome:
> I can check some of this next week.

@pcw-mesa have you been able to look into testing some setups with the updated driver in this PR?

pcw-mesa:
I got pretty busy last week, but I will take a look this week. I can also check some of the less common
interfaces like absolute encoders.

BsAtHome:
Yes, please do check if you have the possibility. The absolute encoder (abs_encoder.c) got changed to fix the parameter and calculation issue.

pcw-mesa:
How do I build this? I don't see a halgs_hostmot2 branch.

grandixximo:
https://github.com/BsAtHome/linuxcnc/tree/halgs_hostmot2

It's on Bertho's repo/fork, you can copy and build that? Do you see it on GitHub from that link?

pcw-mesa:
Got it, building now...

pcw-mesa:
Ok checked halgs_hostmot2

Sserial Ethernet  (7I96+7I84U) Checked all 7I84U I/O = OK
Sserial PCI (5I25T+7I77U) Checked all I/O = OK

Absolute encoder (7I96 + Beckhoff BISS absolute encoder) OK

grandixximo:
Thank you

PR #4370, opened 2026-08-10
Vismach: Merge functions from twp-vismach.py

Sigma1912:
This merges the added functionality in 'twp_vismach.py' to the standard 'vismach.py'
Thus the two vismach models that used 'twp_vismach' can now use the standard 'vismach.py' and 'twp_vismach.py' can be removed.

The indenting in the existing vismach code is messy. I have kept most of it to make the changes more easy to follow.

Note that on some systems there is an error message that is related to the text overlay ('hud') object. This is preexisting, prints once and only if a text overlay has been added to the scene:
```
Exception in Tkinter callback
  File "/usr/lib/python3.11/tkinter/__init__.py", line 1948, in __call__
    return self.func(*args)
           ^^^^^^^^^^^^^^^^
  File "/home/user/git/linuxcnc-master/lib/python/rs274/OpenGLTk.py", line 374, in tkExpose
    self.tkRedraw()
  File "/home/user/git/linuxcnc-master/lib/python/rs274/OpenGLTk.py", line 351, in tkRedraw
    self.redraw()
  File "/home/user/git/linuxcnc-master/lib/python/vismach.py", line 869, in redraw
    self.hud.draw()
  File "/home/user/git/linuxcnc-master/lib/python/vismach.py", line 718, in draw
    glOrtho(0.0, self.app.winfo_width(), 0.0, ypos, -1.0, 1.0)
  File "/usr/lib/python3/dist-packages/OpenGL/platform/baseplatform.py", line 415, in __call__
    return self( *args, **named )
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/OpenGL/error.py", line 230, in glCheckError
    raise self._errorClass(
OpenGL.error.GLError: GLError(
	err = 1285,
	description = b'out of memory',
	baseOperation = glOrtho,
	cArguments = (0.0, 400, 0.0, 400.0, -1.0, 1.0)
)
```

The only in-tree vismach application that uses the current Hud feature seems to be `/src/emc/usr_intf/axis/scripts/tracking-test.py`. The functionality of this is unchanged other than the color of the hud display:

Current:
Screenshot from 2026-08-10 16-32-45

New:
Screenshot from 2026-08-10 16-33-10


c-morley:
Is this something re should add to qt vismach too?

Sigma1912:
Not sure. Maybe we should find a way for qtvismach to use the functions in vismach.py.

lib/python/vismach.py
```diff
@@ -675,52 +1120,121 @@ def invert(src):
...
-                self.lines = len(drawtext)
-                #draw head-up-display
-                #see axis.py for more font/color configurability
-                if ((self.showme == 0) or (self.lines == 0)):
+                messages = []
+                show_list = [None]
+                # check if hud should be hidden
+                for a in self.hide_alls:
+                    if hal.get_value(a[0]) == a[1]:
+                        self.hide_hud = True
```
grandixximo: What happens once the pin has matched a single time and then changes back? As far as I can tell nothing ever resets `self.hide_hud`, so the first match hides the overlay for the rest of the session. The original twp_vismach.py used a local `hide_hud = 0` at the top of `draw()` and re-evaluated the condition every frame; was there a reason to move away from that?

lib/python/vismach.py
```diff
@@ -675,52 +1120,121 @@ def invert(src):
...
+
+        # legacy function, no longer supported, use add_txt() with a tag instead
         def clear(self):
-                self.messages = []
-                
+                print("vismach.py, clear() deprecated, use add_txt() with a tag instead ")
+
+        # legacy vismach models used this
+        def show(self, string):
+            self.add_txt(string)
```
grandixximo, line 1172: A legacy script that calls `hide()` and later `show()` to reveal the overlay again: what makes it visible again with this implementation? `show()` appends a line but never clears `hide_hud`. Should it? Two related compat questions: the old `show(string="xyzzy")` could be called with no argument just to reveal the hud, and the old constructor accepted `Hud(showme=1)`; both now raise `TypeError`. In-tree only tracking-test.py uses this API and is unaffected, but do we care about out-of-tree vismach models here?

lib/python/vismach.py
```diff
@@ -675,52 +1120,121 @@ def invert(src):
...
+                        self.hide_hud = True
+                if not self.hide_hud:
+                    # create list of all line tags to be shown
+                    for b in self.show_tags:
+                        if b[1] == None: # _show_tags_in_pin
+                            tag = int(hal.get_value(b[0]) + b[2])
+                        else: # _show_tag_if_same
+                            if  hal.get_value(b[1]) == b[2]:
+                                tag = b[0]
+                        if not isinstance(tag, list):
```
grandixximo: If the first entry in `show_tags` is a `show_tag_if_same` whose pin does not match, where does `tag` get its value on this line? And on later iterations, wouldn't a non-matching entry reuse the previous iteration's `tag` and leak it into `show_list`? Would initializing `tag` at the top of each loop iteration be safer? (Carried over from twp_vismach.py, but it is landing in the shared library now.)

configs/sim/axis/vismach/5axis/table-rotary_spindle-rotary-nutating/vismach/xyzacb-trsrn-gui.py
```diff
@@ -564,6 +563,7 @@ def get_values(self):
 myhud.add_pin("Zz: {:8.3f}","xyzacb-trsrn-gui.twp_zz",[1,2,3])
 myhud.add_txt("",[1,2,3])
 myhud.show_tags_in_pin("xyzacb-trsrn-gui.twp_status")
+myhud.clear()
```
grandixximo: What does this call accomplish? The new `clear()` only prints a deprecation message, so this config now prints "vismach.py, clear() deprecated..." on every startup with no other effect. Should the line just go away? (xyzbca-trsrn-gui.py did not get this call, so the two configs now differ for no apparent reason.)

lib/python/vismach.py
```diff
@@ -186,6 +256,89 @@ def apply(self):
...
+
+class HalVectorRotate(Collection):
+    def __init__(self, parts, comp, var, th, x, y, z):
+        self.parts = parts
+        self.where = th, x, y, z
+        self.comp = comp
+        self.var = var
+
+    def apply(self):
+        th, x, y, z = self.get_values()
```
grandixximo, line 349: Where does `get_values()` come from when this class is used directly? I only see it defined on the config-side subclass `HalRotateNutation`. Should the base class implement it from `self.where` (which `__init__` stores but nothing ever reads), or is this meant to be an abstract hook?

lib/python/vismach.py
```diff
@@ -203,6 +356,298 @@ def _coord(self, v):
...
+            # line 1 in y direction
+            glVertex3f( i,-s, 0)
+            glVertex3f( i, s, 0)
+        glEnd()
+
+    def draw(self):
+        s = self.s
+        self.grid(s)
+
+    def unapply(self):
```
grandixximo: `draw()` in this class never calls `glPushMatrix()`, so what is this pop balancing? If a model ever places a `HalGrid` inside a Collection, wouldn't each frame pop the caller's transform off the stack?

lib/python/vismach.py
```diff
@@ -128,6 +129,75 @@ def unapply(self):
         glPopMatrix()
 
 
+class HalRotateEuler(Collection):
+    def __init__(self, parts, comp, th1, th2, th3, order=123):
```
grandixximo, line 132: What happens when a caller relies on the default `order=123`? `apply()` looks up `self.comp[self.order]`, and `comp[123]` is not a pin name. Should an integer `order` skip the component lookup, or should the default be dropped so the pin name is mandatory?

lib/python/vismach.py
```diff
@@ -203,6 +356,298 @@ def _coord(self, v):
...
+        zx = 0 if self.zx == 0 else self.comp[self.zx]
+        zy = 0 if self.zy == 0 else self.comp[self.zy]
+        zz = 0 if self.zz == 0 else self.comp[self.zz]
+        r = self.r
+        vo = [ox, oy, oz]
+        vx = [xx, xy, xz]
+        vz = [zx, zy, zz]
+        # calculate the missing y vector
+        vy = [yx, yy, yz] = np.cross(vz,vx)
+        m=np.matrix([[ xx, yx, zx, ox],
```
grandixximo: `np.matrix` is deprecated upstream (also used again at line 559). Is there a reason to prefer it here over plain lists with the `cross()` helpers already in this file? Dropping it would also remove the need for the new `import numpy`.

grandixximo (changes requested):

lib/python/vismach.py
Sigma1912: Indeed. Fixed

configs/sim/axis/vismach/5axis/table-rotary_spindle-rotary-nutating/vismach/xyzacb-trsrn-gui.py
Sigma1912: Removed. Left over from testing.

lib/python/vismach.py
Sigma1912: Fixed, added missing `get_values' method.

lib/python/vismach.py
Sigma1912: Indeed. 
Method 'unapply()'  removed.

lib/python/vismach.py
Sigma1912: Fixed. The method now accepts pins and nummerical values.

lib/python/vismach.py
Sigma1912: Changed to local function and `import numpy` removed

lib/python/vismach.py
Sigma1912: Fixed
- constructor accepts the `showme` argument and actually uses it instead of just ignoring it.
- `show()` will clear  `hide_hud`

lib/python/vismach.py
Sigma1912: Indeed, `tag` is now initialized.

Sigma1912:
@grandixximo 
Thanks so much for spending the time to review this!

grandixximo (approved):

BsAtHome:
How is this influenced by the glcanon updates? I don't know if that is imported here (don't think so), but it did show the necessity to update some old code.

How does this code relate to the upgrade done in glcanon? Are there similar updates necessary/required/useful for this code or are there no things in this code that touch the same things as glcanon?

Sigma1912:
vismach is not touched by the glcanon updates.

Updating the vismach code to modern gl would certainly be welcome but that is well beyond my reach.

I have however ported vismach to VTK: https://github.com/Sigma1912/LinuxCNC_Demo_Configs/blob/main/serial-manipulators/vtk_vismach.py 

vtk

PR #4344, opened 2026-08-06
docs: spindle orient fix

grandixximo:
Back-port cherry-pick of #4342

PR #4371, opened 2026-08-11
Merge branch '2.9'

grandixximo:
Periodical merge of 2.9 into master, keeping the branches in sync as done previously (e.g. 3a05670232).

This one turns out to be content-neutral: every commit currently on 2.9 that was missing from master is already represented on master in equivalent form:

- The homing-while-homing guard (#4321) is a backport of 9d5570ca84, already on master
- The CI apt-mirror commits (#4345) match what master already has from #4336
- The M19/orient doc fixes (#4344) are already on master in adoc form

Two trivial conflicts, both resolved keeping master's version:

- `src/hal/components/orient.comp`: 2.9 edited the old roff markup; master has the same wording in adoc
- `gmoccapy.py`: 2.9 uses the literal `(13, ...)`; master uses `linuxcnc.OPERATOR_DISPLAY` (same value)

The merged tree is byte-identical to current master (`git diff master MERGE` is empty), so this is pure ancestry bookkeeping: it records that 2.9 is fully merged up, and gives the next 2.9 => master merge a clean base.

### DIFF
--- a/.github/scripts/install-deps.sh
+++ b/.github/scripts/install-deps.sh
@@ -3,6 +3,8 @@
 set -eu #Needed so CI fails when anything is wrong
 set -x
 
+.github/scripts/use-main-ubuntu-mirror.sh
+
 sudo apt-get --quiet update
 sudo apt-get install --yes --no-install-recommends devscripts equivs build-essential lintian clang
 debian/configure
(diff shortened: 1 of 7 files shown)


grandixximo:
@hansu does this look right?

PR #4372, opened 2026-08-11
switchkins: allow more than three kinematics types

grandixximo:
`switchkins.c` dispatched on `switchkins_type` with a three way switch statement and created a fixed set of `kinstype.is-0/1/2` pins, so a kinematics module could never offer more than three kinematics. This changes the dispatch to arrays indexed by type, and adds a way for a module to provide more.

Nothing changes for any existing module or configuration.

## What a module sees

`switchkinsSetup()` keeps its exact prototype and still provides types 0, 1 and 2. None of the eight in-tree kinematics modules is touched, and out of tree modules keep compiling unchanged.

A module that wants more calls the new export from inside its own `switchkinsSetup()`, once per additional type:

```c
int switchkinsRegister(int ktype, KS kset, KF kfwd, KI kinv);
```

`ktype` runs from 3 to `SWITCHKINS_MAX_TYPES - 1`, and every type below the highest one registered has to be provided. `SWITCHKINS_MAX_TYPES` goes from 3 to 9. It is now only a ceiling on the array sizes rather than the number of types in use, so a module still provides exactly three unless it registers more, and the extra headroom costs a few unused array entries.

## HAL pins

The `kinstype.is-N` pins are created in a loop, one per provided type. For a module providing three types that produces `kinstype.is-0`, `kinstype.is-1` and `kinstype.is-2`, the same names as before, so nothing that nets those pins breaks. A module providing five gets `kinstype.is-0` through `kinstype.is-4`.

A setup routine now runs once per type it is registered for, so a routine used for two types must not create the same pin twice. Noted in the docs.

## The first commit is a separate bug fix

`kinematicsSwitch()` stored the requested type in `switchkins_type` and only then ran the switch statement that validates it. The switch returned -1 and motion raised its error flag, but `switchkins_type` kept the bad value, so the module was left pointing at a kinematics that does not exist. Every `kinematicsForward()` and `kinematicsInverse()` call after that failed as well, printing

```
switchkins: Forward BAD switchkins_type </7>
```

once per servo cycle for as long as the machine stayed up. On the scara sim, `M68 E3 Q7` reproduces it on master.

The fix validates the request first and returns without touching `switchkins_type`, which leaves the running kinematics in place. It stands on its own and does not depend on the rest, which is why it is a separate commit.

## Testing

All eight switchable modules (`scarakins`, `genhexkins`, `genserkins`, `pumakins`, `three21kins`, `5axiskins`, `xyzac-trt-kins`, `xyzbc-trt-kins`) load under `halrun` and create exactly the same three pins, with the same names, as they do on master.

On `configs/sim/axis/vismach/scara/scara.ini`:

- `M68 E3 Q1` selects kinematics 1, `M68 E3 Q0` selects 0.
- `M68 E3 Q7` is refused, kinematics 1 keeps running, and there is no per cycle log spam.

With `scarakins` temporarily registering two extra types (a local test edit, not part of this PR), the same config gives `kinstype.is-0` through `kinstype.is-4` and switches to type 4 and back.

## Why

I have a machine that needs more than three kinematics for the same config. There is no in-tree module that needs a fourth type today, so the second commit is enabling work rather than something with a visible user in this repository. I would rather add the capability in a shape you are happy with than carry a fork of `switchkins.c`.

I looked at doing this with a second array based `switchkinsSetup()` prototype selected by `#if`, but that duplicates every dispatch site and every module's setup function for something that is decided at compile time anyway. A registration call seemed like much less to maintain.

### DIFF
--- a/src/emc/kinematics/switchkins.c
+++ b/src/emc/kinematics/switchkins.c
@@ -38,19 +38,17 @@
 // kinematic functions (default=0 for err detection):
 static kparms kp; // kinematics parms (common all types)
 
-static KF kfwd0 = NULL; // 0==switchkins_type kinematics forward
-static KF kfwd1 = NULL; // 1
-static KF kfwd2 = NULL; // 2
+// indexed by switchkins_type (NULL==not provided, for err detection):
+static KS ksetups[SWITCHKINS_MAX_TYPES] = {NULL};
+static KF kfwds[SWITCHKINS_MAX_TYPES]   = {NULL};
+static KI kinvs[SWITCHKINS_MAX_TYPES]   = {NULL};
 
-static KI kinv0 = NULL; // 0==switchkins_type kinematics inverse
-static KI kinv1 = NULL; // 1
-static KI kinv2 = NULL; // 2
+// types provided: 3 from switchkinsSetup(), more from switchkinsRegister()
+static int kins_count = 3;
 
 static int switchkins_type;
 static struct swdata {
-    hal_bool_t kinstype_is_0;
-    hal_bool_t kinstype_is_1;
-    hal_bool_t kinstype_is_2;
+    hal_bool_t kinstype_is[SWITCHKINS_MAX_TYPES];
 
     hal_real_t gui_x;
     hal_real_t gui_y;
@@ -104,15 +102,16 @@ static int gui_forward_kins(const double *joints)
     int res;
     KINEMATICS_FORWARD_FLAGS  fflags = 0;
     KINEMATICS_INVERSE_FLAGS  iflags;
-    switch (kp.gui_kinstype) {
-        case 0: res = kfwd0(joints, &lastpose[0], &fflags, &iflags);break;
-        case 1: res = kfwd1(joints, &lastpose[1], &fflags, &iflags);break;
-        case 2: res = kfwd2(joints, &lastpose[2], &fflags, &iflags);break;
-       default: rtapi_print_msg(RTAPI_MSG_ERR,
-                  "gui_forward_kins BAD gui_kinstype <%d>\n",
-                  kp.gui_kinstype);
-                  return -1;
-     }
+    if (   kp.gui_kinstype < 0
+        || kp.gui_kinstype >= kins_count
+        || !kfwds[kp.gui_kinstype]) {
+        rtapi_print_msg(RTAPI_MSG_ERR,
+                        "gui_forward_kins BAD gui_kinstype <%d>\n",
+                        kp.gui_kinstype);
+        return -1;
+    }
+    res = kfwds[kp.gui_kinstype](joints, &lastpose[kp.gui_kinstype],
+                                 &fflags, &iflags);
     hal_set_real(swdata->gui_x, lastpose[kp.gui_kinstype].tran.x);
     hal_set_real(swdata->gui_y, lastpose[kp.gui_kinstype].tran.y);
     hal_set_real(swdata->gui_z, lastpose[kp.gui_kinstype].tran.z);
@@ -130,7 +129,7 @@ int kinematicsSwitch(int new_switchkins_type)
     int k;
 
     // reject first, so a bad request leaves the running kinematics alone
-    if (new_switchkins_type < 0 || new_switchkins_type >= SWITCHKINS_MAX_TYPES) {
+    if (new_switchkins_type < 0 || new_switchkins_type >= kins_count) {
         rtapi_print_msg(RTAPI_MSG_ERR,
                         "kinematicsSwitch:BAD VALUE <%d>\n",
                         new_switchkins_type);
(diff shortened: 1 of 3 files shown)


src/emc/kinematics/switchkins.c
```diff
@@ -128,36 +127,25 @@ int kinematicsSwitchable() {return 1;}
...
-                switchkins_type);
-                hal_set_bool(swdata->kinstype_is_1, 0);
-                hal_set_bool(swdata->kinstype_is_0, 0);
-                hal_set_bool(swdata->kinstype_is_2, 0);
-                return -1; // FAIL
+
+    rtapi_print_msg(RTAPI_MSG_INFO,
+                    "kinematicsSwitch:TYPE%d\n", switchkins_type);
+    for (k=0; k < kins_count; k++) {
+        hal_set_bool(swdata->kinstype_is[k], k == switchkins_type);
```
BsAtHome, line 147: Shouldn't you be asserting an error of some kind if the `switchkins_type` value is bad?
Or is the old default switch case no longer possible?

src/emc/kinematics/switchkins.c
```diff
@@ -228,6 +216,22 @@ KINEMATICS_TYPE kinematicsType()
...
 
+int switchkinsRegister(int ktype, KS kset, KF kfwd, KI kinv)
+{
+    if (ktype < 3 || ktype >= SWITCHKINS_MAX_TYPES) {
+        rtapi_print_msg(RTAPI_MSG_ERR,
+                        "switchkinsRegister: BAD switchkins_type <%d>"
+                        " (must be 3..%d)\n",
+                        ktype, SWITCHKINS_MAX_TYPES - 1);
+        return -1;
+    }
```
BsAtHome, line 236: Wouldn't it be more consistent to allow switchkinsRegister to supplant the old switchkinsSetup?

The real issue is to deal with is empty/skipping slots. That problem will be present also for any entry larger than 3.

src/emc/kinematics/switchkins.c
grandixximo: No longer possible. The first commit moved the range check to the top of `kinematicsSwitch()`, so a bad value returns before `switchkins_type` is assigned and before any pin is touched, which makes the `default:` arm dead.

It is context rather than an addition in this diff, so the removal looks unexplained here. I can add a comment at the loop noting the range is already checked.

src/emc/kinematics/switchkins.c
grandixximo: **Supplanting `switchkinsSetup()`**: agreed the split is inconsistent. It cannot be replaced without making it optional, since every module defines it and `rtapi_app_main()` calls it unconditionally. Either convert the eight in-tree modules and drop it, which breaks out of tree modules at compile time, or declare it weak and call it only if present, which I can only test on uspace. Cheap step either way is dropping the `ktype < 3` restriction so registration covers 0 to 2 as well. Which end state do you want?

**Gaps**: already refused at module load, not at switch time. `scarakins` registering type 5 without 3 and 4 gives

```
Switchkins FAIL scarakins:<Missing setup function>
```

The message is the problem, it names neither the type nor the gap. Same area: a double registration silently overwrites, and nobody checks the return of `switchkinsRegister()`. I will name the missing type, reject double registration, and fail the load on a rejected registration.

src/emc/kinematics/switchkins.c
BsAtHome: No, fine by me. The logic in switchkins is one that needs reading in context... The diff does not always provide enough context.

src/emc/kinematics/switchkins.c
BsAtHome: The path of least resistance would be to drop the ktype < 3 condition.

In the long run, we should make it generic. I'm also confused a bit because there are examples in hal/components that do their own switch. Maybe this is the reason why the current construct feels messy and inconsistent.

grandixximo: Done, in a separate commit.

`ktype` runs from 0 now, and the count comes from the highest type filled by either route, so a module can register all of them and leave the `switchkinsSetup()` arguments alone. A type has to come from one route or the other. Registering one that `switchkinsSetup()` already filled, leaving a gap, and an out of range type are all refused at load time, naming the type at fault.

On the components with their own switch: `millturn`, `xyzab_tdr_kins`, `xyzacb_trsrn` and `xyzbca_trsrn` in `src/objects/hal/components` implement `kinematicsSwitch()`, forward, inverse and their own `kinstype.is-N` pins without linking `switchkins.o`, so there are two implementations of the same thing in the tree. They also all have the assign before validate bug the first commit fixes. Folding them onto `switchkins.c` looks doable now the type count is not fixed. Do you want that in this PR or a separate one?

For context on where I am going with this: I have G12.1 and G13.1 working locally to select the kinematics from G-code, with the interpreter tracking which one is active and exposing it as `#<_kins_type>`. Selecting the kinematics through a HAL pin and then forcing a queue buster with `M66 E0 L0` is the wrong shape: it is a state change the interpreter has to know about, and today it does not, so `#<_kins_type>` and anything built on it can disagree with what motion is running. The intention is to make G-code the proper way to switch and deprecate the HAL route, which is also what makes more than three types worth having.

BsAtHome:
I was wondering whether it would be helpful to add an example (stripped down outline) to the switchkins.adoc file. At least it should point to some examples in the sources. It would help for using the switchkinsRegister() API.

grandixximo:
Added: an outline in Code Notes showing one switchkinsSetup() filling kinstypes 0 to 2 through the pointer arguments and registering a fourth, with the kinematics left out, and a pointer to the in-tree switchkinsSetup() routines for the surrounding shape.

Worth saying plainly in the doc, and it now does: no in-tree module registers a fourth kinstype yet, so that call has no example to copy. Documented as it stands for now, and I will revise as things evolve.

PR #4374, opened 2026-08-12
Multiaxis kinematics and limits

grandixximo:
Draft. Where the multiaxis kinematics work is going, open early so it can be argued with. As pieces are agreed I split them into small PRs, rebase this, and carry on, so nothing stalls behind it.

Plan, in order:

1. Switchkins: one implementation, used by `.comp`, `.c`, in tree and out of tree. The registration half merged as #4372; the rest is in this branch.
2. Small fixes that stand on their own: `5axiskins` gets a `.tool-offset` pin like every other module, `maxkins` gets one or a doc note saying it has none, and `G53` stops ignoring the rotary tool offset on non-wrapped axes.
3. The kinematics interface learns to answer two more questions: where the tool points, and how joint motion maps to world motion. One entry point each, both weak, both defaulting to today's behaviour, and a tool frame convention written down so every module answers the same way rather than each author choosing one.
4. Those answers, and forward and inverse alongside them, become pure functions of what the caller passes in: joint values, a parameter block carrying the tool and the kinematics type, and no state kept inside the module. Today the maths reads its geometry from HAL pins and keeps scratch in `haldata`, so the only way to call it from anywhere else is to write a second copy of it. One implementation, parameters supplied at the call site.
5. Two stages the offset chain was missing. Both came out of the discussion below, both wanted 3 and 4 first, and both are in this branch now.

   A frame composed on top of the active coordinate system, with its own cancel, so the blocks in between are programmed in the tilted plane while G54 itself is untouched. Fanuc spells this `G68.2`, and that is the spelling here: `G68.2` in its four forms, `G68.3` reading the plane off the tool, `G68.4` composing onto the active plane, `G69` cancelling. In tree it is a stage in `rotate_and_offset_pos()` plus modal group 9. Nothing is rewritten, so nothing needs restoring: any abort cancels it, and `on_abort_with_twp_reset.ngc` is gone with the rest of the remap.

   A move whose target is defined but whose path is left to the joints: run the inverse once at the endpoint, then interpolate the joints to reach it, the slowest setting the pace. That is how the commercial controls cross a singularity or flip a head without changing coordinate system. In tree it is one segment type in the planner, `JOINT_TRAVERSE` and `JOINT_FEED` in canon, and three non-modal codes on `G0` and `G1`, like `G53`, each stripping one more layer of interpretation from the destination: `G53.4` to a program point, offsets and plane applied, the form a post writes; `G53.5` to the machine frame, the carriage or a robot's flange in machine coordinates with the tool left out, in program units and with no offset, the form for parking and tool change; `G53.7` to joint values, `J2=-5` sending joint 2 to -5 in the joint's own units with nothing converted, the form that works on every machine. Which kinematics type is the machine frame the module declares with a flag, `KINSTYPE_MACHINE`, next to `KINSTYPE_IDENTITY` and `KINSTYPE_PRIMARY`: the carriage on a mill, the flange in the base frame on a robot (ABB `tool0`, KUKA `$NULLFRAME`), where `pumakins` and `genserkins` flag their arm type. `G13.1` and `G49` cancel to it, so a robot stays on its arm kinematics and only `G12.1 P1` selects raw joints. A module that declares nothing has its identity type stand in, and there a letter carries a unit class and a joint does not, so `G53.5` is refused where the letters name joints of the other kind, which `[JOINT_n] TYPE` declares. With `G1` the move takes the time the straight move would at F. `G53.1`, `G53.3` and `G53.6` orient the tool to the plane, and `G53.2` solves the same orientation without moving, publishing the pose on the read-only parameters `#<_orient_x>` to `#<_orient_c>` and `#<_orient_valid>`, mirrored on `#5071` to `#5080`, so a program can read the answer and use it in a move of its own. `P` picks the solution nearest first and `Q` says whether the joints that carry the work take part; the module's tool frame inverse answers, so no config carries the formula, and the two `trsrn` configs run their demos with no remap left.

   Between them these are the reason not to add a set of offsets per kinematics type. Controls do let you add coordinate systems, Fanuc `G54.1 P1` to `P48`, Siemens `G505` to `G599`, and none of them ties a set to the active kinematics or transformation: selection is explicit in the program, or it belongs to a channel. More coordinate systems is a fair request on its own; binding them to the kinstype is the workaround for these two absences.
6. Limits taken from the joints instead of from static per-axis INI values. Joint limits are the physical fact. The region they imply in axis space depends on the tool and on where the rotaries are, and is not a box, which is why a box in the INI cannot describe it and config authors have to pick a conservative one and lose travel. Built now, in three parts, each with a test that was red before it.

   Travel is the `[JOINT_n]` limits, every move, every kinematics type. The `[AXIS_L]` box is the machine frame envelope: the carriage, or the flange, in machine coordinates, the frame that does not move when the head tilts or a tool is loaded, which is how Siemens and Fanuc read axis limits, machine-axis space after the transformation. Every move is checked against the box where its joints put the machine frame, `G0` and `G53.5` alike, so a tool centre point move whose tip lies past the box with the carriage inside it is accepted and a `G53.5` to a carriage position past the box is refused. No box on tip coordinates: the tip is not the carriage, and a tip fence would be a feature of its own. One box, one meaning, so nothing needs resetting at a switch, which retires the `setp ini.L.min_limit` the docs prescribed and the puma's M129 workaround. A module that declares no machine frame keeps the box on the world in force, as today. The one behaviour change for an existing switchkins config that declares only its identity: under TCP the box moves from the tip to the carriage.

   Velocity and acceleration come off the Jacobian. Canon samples every segment through the module's Jacobian ahead of motion, on the kinematics type and the tool offset the program is under, and lowers its velocity and acceleration to what the slowest joint can give within its own limits: the rate of each joint along the tangent, the share the changing rate takes at that speed (what a carriage on a circle feels as centripetal), and the bulge between samples so a peak between them is covered. Lines and arcs, traverse and feed, probe and rigid tap. A turn of a tilted head under the tool centre point, which swings the carriage through a circle the head's own limit never mentions, runs at the carriage's pace instead of tripping the carriage's following error; a full circle of the tip with C turning a revolution under it rides both. World jogs get the same in motion, every servo cycle, from the Jacobian at the joints the machine stands in, the planners capped so no joint is asked for more than its limits and the jog keeping its direction. `[AXIS_L] MAX_VELOCITY` and `MAX_ACCELERATION` keep meaning the world of the type in force; the joint cap sits on top. A module that runs only in realtime, one without the parameter block form, caps nothing, as before. A pole, where a joint would have to move at any speed for the path to advance, gives an operator message and the move crawls rather than being dropped, since a dropped segment sends the next one along another path; refusing it ahead of time is item 7.

   The same query finishes a repair the tool orientation codes already needed. Solutions are now unwrapped onto the nearest turn of the current pose before ranking, so a C at 180 asked for a pose at C -90 goes to C 270, 90 degrees away, instead of turning 270 degrees the long way. What ranking cannot do yet is prefer the closest pose that stays inside the joint's limits: near a limit the nearest turn may be one the joint cannot reach, and ranking closest-within-limits has to know that joint's range, which this item now supplies and the ranking does not read yet.

7. The joint limit check moves earlier and samples inside segments. It exists already, in `inRange()`, but only at move endpoints and only once the block has reached motion, so a program that will overtravel says so partway into the cut rather than before it starts. With 4 in place the same check runs at load time, against the tool and the kinematics type each block will actually use. A verify button in a GUI is then a presentation of that, not a subsystem of its own.

   Two things this needs pinned before it is written. The interpreter has no seam for the query: all 66 `GET_EXTERNAL_*` accessors take at most an index, an axis, a spindle, a pocket, and never geometry, so one taking a segment is a new shape for that interface. And a load-time answer can go stale, since the interpreter runs ahead and `motion.switchkins-type` is a HAL_IN pin, so until 9 lands the type a block is checked against is not necessarily the type it runs under.

   Where a module cannot answer, cheaply or at all, the check skips rather than refuses. Refusing a reachable program leaves the user no recourse; missing an unreachable one is still caught by `inRange()` at runtime. That asymmetry decides every ambiguous case in 6 and 7.

   Checking a program that switches type raises a question the tree currently leaves implicit: whether a type's coordinates are geometric. Most are, and for those a scan is one stream of world points with only the inverse changing at the switch. `identity` is not, since world becomes the joints, so a datum in millimetres lands on a rotary joint in degrees; nothing notices, which is why switchkins.adoc asks the operator to clear the offsets before switching.

   The joint-space syntax that question wanted is `G53.7` above: joint values by joint number, no offsets, the joint limits directly. Whether a type should also declare its space for the scan stays open.
8. Per-mode restrictions. G28, G30, G53, G92, G52 and M6 do not mean the same thing in every kinematics type, and today each one quietly does whatever the module happens to leave it doing. Refusing outright beats guessing, and with 4 in place the control knows enough to decide rather than to be told. The first of them is in: `G53`, `G28` and `G30` keep meaning the world of the kinematics in force, which on a table-side kinematics turns with the table, and `G53.5`, `G28.5` and `G30.5` are the machine frame forms; a configuration that wants the machine codes to mean that frame and nothing else sets `MACHINE_MOVES_NEED_MACHINE_FRAME` and has the world forms refused off the machine frame type.
9. Retire `motion.switchkins-type`, once the grace period has run. Which release that is has still to be decided.

G12.1 and G13.1 are #4414, split out of this branch and standing on their own.

On tool length specifically: the offset is a nine-axis pose from the interpreter, canon adds it into the coordinate, and each rotary kinematics takes the Z component back through a HAL pin and re-applies it along the real tool direction. That round trip is exact when the config wires the pin, and silently wrong otherwise: a missing `net` costs the full tool length, and an X or Y offset is applied in the rotating work frame while the tool is bolted to the machine. The mechanism works; it has no owner.

Why 4 is worth the trouble: every other way of answering "will this program overtravel" builds a second model of the machine, in the verification package or in the post, and the two descriptions drift apart. If the kinematics module can be called with parameters the caller chooses, LinuxCNC can check a program against the same code that will run it, with no second model to keep in sync. That is not available to anyone whose kinematics is locked inside the control.

## In this branch now

Items 1, 3, 4, 5 and 6, the first of 8, in that order:

- Switchkins is one implementation: the dispatch splits from `rtapi_app_main()`, the four comps call it from `EXTRA_SETUP()`, `switchkinscomp.comp` is the out-of-tree template, and the implementation installs as source. #4414 (G12.1 and G13.1) sits at the bottom.
- Item 3: every module reports its work frame and its tool frame, `5axiskins` included, which reported none until now and so refused every orientation code, the tool frame inverse and the Jacobian are shared code with closed forms registered where a module has one, and the conventions chapter fixes frames, signs, tool X and the zero pose. #4455 sits in the middle.
- Item 4: a module is a parameter block and a table of pure functions; `kinematics_userspace/` loads a module outside RT and binds to the RT instance's live pins; `tests/kins-params` checks that each module answers the same outside RT.
- Item 5: the frame, the point-to-point moves, the orientation codes and `G68.3`, as above, with the interpreter evaluating the module ahead of motion. The inverse ahead of motion runs to a fixed point, because a module may read the joints it is handed and one pass from a stale seed answers for the wrong rotary angles. In motion a point-to-point segment keeps the last forward solution and clears a status flag when the forward fails inside it rather than parking the reported position at zero, and an external offset waits while a point-to-point segment is queued, since its joints were solved without it. Three repairs the work turned up: `G68.3` no longer leaks its axis words into a stray traverse, the python preview applies `SET_G68_FRAME` in program units like every other length canon call, and orientation answers from the module are unwrapped onto the nearest turn of the current pose before nearest-first ranking. The bridgemill sim runs its sphere demo five ways, one per method: `G53.2` with a fused orient-and-travel move, `G53.6` orient-then-move, real quill drilling through `G53.5` and `G53.7`, a `G12.1` kinematics switch, and the historical W-as-tool-length program kept with the quill drawing locked, which reveals that the old trick only ever worked because the quill was not drawn.
- After item 5: a tool offset change under a tilt keeps the joints where they are and reads the programmed point back from them, so `G43` with the head tilted moves the point, not the machine, and the interpreter predicts the same through the loader. `G43.5` gives the tool axis as a vector, `I J K` on `G0` and `G1`. The machine frame is a declared kinematics type, `G53.5`, `G28.5` and `G30.5` move in it, `G13.1` and `G49` cancel to it, and the robots declare their arm type as such, so on a puma `G53.5 Z50` lifts the flange along base Z holding the rest.
- Item 6: the `[AXIS_L]` box read on the machine frame for every move, the segment cap in canon, the jog cap in motion, as above. `kinslimits` prints the cap canon reads for a move next to the Jacobian.

## User visible

Fourteen G-codes, all of them non-modal except the plane itself, which has its own cancel:

| code | what it does |
| --- | --- |
| `G68.2` | define a tilted work plane on top of the active coordinate system, by Euler angles, by angles about fixed axes, by three points or by two vectors |
| `G68.3` | define one from where the tool points now |
| `G68.4` | compose one onto the plane already active |
| `G69` | cancel the plane |
| `G53.1` | turn the rotaries until the tool is normal to the plane, the linear joints staying where they are |
| `G53.2` | the same, solved without moving; the pose is published on parameters for the program to use |
| `G53.3` | the same, and take the tool to an X Y Z given in the plane |
| `G53.6` | the same, holding the tool centre point where it is |
| `G53.4` | point-to-point to a destination in program coordinates, offsets and plane applied |
| `G53.5` | point-to-point to the machine frame, the carriage or the flange in machine coordinates with the tool left out, in program units and with no offset |
| `G53.7` | point-to-point to joint values, `J2=-5` naming joint 2 in its own units |
| `G28.5`, `G30.5` | the machine frame forms of `G28` and `G30`, through the same stored positions |
| `G43.5` | the tool axis as a vector, `I J K` on `G0` and `G1`, the rotaries solved by the module |

`P` on the orientation codes picks among the poses that reach the plane, nearest first, and `Q1` frees the joints that carry the work to take part. The point-to-point codes are modifiers of `G0` and `G1`, and with `G1` the move takes the time the straight move would at the programmed feed. `J<n>=<value>` is a new word form, read only where `G53.7` uses it.

Alongside them:

- `stat.g68_offset`, `stat.g68_rotation` and `stat.g68_active` in the python interface; halui's relative pins and the AXIS DRO apply the plane.
- `#<_orient_x>` to `#<_orient_c>` and `#<_orient_valid>`, mirrored on `#5071` to `#5080`, hold the pose `G53.2` last solved; read-only, and reading one before the first `G53.2` is an error.
- The four components gain the `coordinates=` module parameter; millturn loses its `in` and `out` example pins.
- The two `trsrn_twp` configs run on the native codes: no remap, no python, no `twp-status` pin. Their python maths is the oracle in `tests/kins-twp`.
- One new INI option, `[RS274NGC] MACHINE_MOVES_NEED_MACHINE_FRAME`, off by default: `G53`, `G28`, `G30`, `G28.1` and `G30.1` refused whenever the kinematics in force is not the machine frame type.
- `[AXIS_L] MIN_LIMIT` and `MAX_LIMIT` are the machine frame box on a module that declares its machine frame; the velocity and acceleration limits keep their meaning and the joint cap sits on top. A switch no longer needs the limits reset to keep the joints safe; it may still want the velocity limits reset where a letter changes unit at the switch, which the puma's M128 and M129 do and nothing else.
- A module declares which of its types is the identity, the primary and the machine frame, `.identity`, `.primary` and `.machine` in the ops table or `switchkinsDeclare()` the old way; `motion.kins-type` and the `switchkins-type N declared:` line at load say what it declared.
- The preview draws the tilted work planes a program defines, the one in effect highlighted. Suggested by Sigma1912.
- No existing pin or G-code changes name. Configs using the `.tool-offset` net keep working; the tool comes from motion now and the net is only read until motion has spoken.

## Left for later

- Item 7, the joint limit check ahead of motion and inside segments, and with it the refusal of a pole that today only crawls with a message.
- The orientation codes ranking the closest pose that stays inside the joint limits.
- The quick reference and the qtvcp MDI help do not list the new codes yet.

## Testing

`tests/kins-params`, `tests/kins-jacobian`, `tests/kins-frames`, `tests/tool-frame`, `tests/kins-tool-offset`, `tests/kins-twp`, `tests/twp-native`, `tests/ptp-iterative`, `tests/ptp-machine-frame` and `tests/kins-limits` cover the new code; `tests/kins-limits` holds the three tests that were red before item 6, the swing of a 400 mm pivot under `G0 C180` and under an arc, the box read on the carriage instead of the tip, and the same swing as a world jog, each joint following its command through `limit3` at 20 percent over its own limits the way a drive does, so a command over the limits falls behind and trips the following error; the last runs the xyzacb nutating head through the plane, the orientation codes and the point-to-point moves against the python maths, and checks that every point-to-point move keeps its joints on one line in joint space, all together, so a wrong interpolation fails three of its checks. `tests/ptp-robot` puts pumakins through the same codes, where `G53.5` is refused and `G53.7` is the form that works. The four comp sim configs give identical positions to master through the same MDI sequence, in every kinematics type.

## Testers wanted

Everything above runs in the sims, and the sims follow their commands exactly. What I cannot see from here is a real five axis machine or a real robot, where the joint cap either keeps a drive inside its following error window or it does not. If you run a `5axiskins`, `xyzac-trt`, `xyzbc-trt`, `maxkins` or `trsrn` config, or a `pumakins` or `genserkins` arm, a session on this branch would tell me a lot:

- A `G0` or `G1` that turns a rotary under TCP with the head tilted, at full rapid, and whether it slows to the carriage's pace or trips a following error. Same with a world jog of the rotary in teleop mode.
- Whether the `[AXIS_L]` box now reads right on your machine: a tip past the box with the carriage inside accepted, `G53.5` past the box refused, and any move you expected to be accepted that is refused, with the line and the message.
- On a robot, `G53.5` along a base axis, and whether `G13.1` and `G49` leave you where you expect.
- A config on an out-of-tree kinematics module: whether it still loads and whether `motion.kins-type` says what it declared.
- Anything slower than it was: a long program with short segments on an iterative kinematics is where the per-segment sampling would show.

`EMCCANON_KINS_DEBUG=1` in the environment makes canon print the cap it read for every segment on stderr, and `kinslimits` prints it for one move from the command line.


src/emc/kinematics/switchkins.h
```diff
@@ -6,8 +6,8 @@
 
 #include <kinematics.h>
```
BsAtHome: When you export this file into the global pool, then then you need to change the exclusion name and prefix it properly like the others.

That also means that the inclusion here needs to change from <> to "" because switchkins.h needs to include the local version of kinematics.h (from the directory where switchkins.h was found).

It also means that all users of switchkins.h, excluding switchkins.c, must use <> in the include.

src/emc/motion/motion.h
```diff
@@ -180,6 +180,8 @@ extern "C" {
         EMCMOT_SET_AXIS_LOCKING_JOINT,  /* set the axis locking joint */
 	    EMCMOT_SET_AXIS_JERK_LIMIT,         /* set the max axis jerk */
 
+	EMCMOT_ADJUST_KINS_OFFSET_DATA,  /* set the offset in kins (G12.1) */
+
         EMCMOT_SET_SPINDLE_PARAMS, /* One command to set all spindle params */
```
BsAtHome, line 182: Indent?

src/emc/nml_intf/emc_nml.hh
```diff
@@ -1158,6 +1172,10 @@ class EMC_MOTION_STAT:public EMC_MOTION_STAT_MSG {
     int numExtraJoints;
     bool jogging_active;
     uint64_t heartbeat;  // motion controller's heartbeat counter
+
+    char trajKinsType;
+    bool trajKinsTypeModified;
+    double adjustKinsVar0;
```
BsAtHome: When these are prefixed with `traj...`, don't the belong in the `EMC_TRAJ_STAT` instead?

The `...Type` as a char is strange. Types are usually enums (int) or simple counting integers. Are you trying to save space?

The `adjustKinsVar0` is an odd name. It in quite unclear why there is a `0` in that name. Is there also a `adjustKinsVar1` and `adjustKinsVar2`? And, what does it adjust?

src/emc/motion/motion.h
grandixximo: indent is a tab which equal 8 spaces per code conduct, the wrong one is the one above mine, can I take the blame and fix indent all to tab, at least in this section of the file?

BsAtHome: Ah well, never mind then. We need to plan this reformat thing sooner than later without breaking blame.

src/emc/nml_intf/emc_nml.hh
grandixximo: > When these are prefixed with traj..., don't the belong in the EMC_TRAJ_STAT instead?

Moved to EMC_TRAJ_STAT

> The ...Type as a char is strange. Types are usually enums (int) or simple counting integers. Are you trying to save space?

Left overs from original patches I was working off of, changed to int

> The adjustKinsVar0 is an odd name. It in quite unclear why there is a 0 in that name. Is there also a adjustKinsVar1 and adjustKinsVar2? And, what does it adjust?

Renamed adjustKinsVar0 to switchkins_type no number needed.
Renamed adjust elsewhere to select, improved the names in general, they were inherited from previous functional but not nice work ;-)

src/emc/motion/motion.h
grandixximo: that line above was actually mine anyway, so fixed that at least, kept tabs

src/emc/kinematics/switchkins.h
grandixximo: Missed, again 🤕 
Fixed

docs/src/man/man9/motion.9.adoc
```diff
@@ -253,9 +253,20 @@ Note: feed-inhibit applies to G-code commands -- not jogs.
...
+  select the machine kinematics functions. Extra G-code commands are
   required to synchronize task and motion before and after changes to
   the pin value.
+  *Deprecated*: the interpreter does not see this pin, so limits and
+  look ahead go on using the kinematics it last knew about. Use the
+  G-code words *G12.1 P-* and *G13.1*, which ask motion directly and
+  synchronize task and motion themselves. Motion reports the
+  deprecation once, the first time the pin is used to change the
+  kinematics. The pin is in a grace period: it keeps working for now,
+  and is meant to go.
```
BsAtHome: But you mention the pin in the g-code doc?

src/emc/kinematics/switchkins.c
BsAtHome: If switchkins.h includes kinematics, then why do you include it here?

BsAtHome:
What is the migration path for people having their own out-of-tree kinematics switch component?

docs/src/man/man9/motion.9.adoc
grandixximo: Yes, revising there too...

src/emc/kinematics/switchkins.c
grandixximo: copy pasta, will check other kins too, anyway I'm gonna have to refactor those too, better now than later when I forget ;-)

grandixximo:
>What is the migration path for people having their own out-of-tree kinematics switch component?

Their module needs no source change. G12.1 and G13.1 reach a module through the same `kinematicsSwitch()` the pin does, so a component that is switchable today is driven by G12.1 today without knowing G12.1 exists.

I checked rather than assumed. I built an out-of-tree component that deliberately does not use switchkins.h or switchkins.c: its own switchkins type, its own `kinematicsSwitch()`, its own kinstype pins, the shape somebody would have written before there was a supported way to reach the in-tree code. Against this branch:

    G12.1 P1   kins-type=1, its type 1 kinematics active
    G13.1      kins-type=0
    G12.1 P7   its own BAD VALUE, motion stays on 0

with no source change, and no deprecation warning, since the pin is never written.

The config keeps working as well. The pin is deprecated, not gone, so an existing M68/M66 config runs as before with one warning per session. Moving off it is a config edit, not code:

    M68 E3 Q1 / M66 E0 L0    ->   G12.1 P1
    M68 E3 Q0 / M66 E0 L0    ->   G13.1

and the `[HAL]` net from the analog output goes with them. The sim configs in this PR are worked examples, halui buttons and remapped M-codes included.

One group does have something to do, and it is not the one you asked about: anybody keeping a copy of switchkins.c in their own tree. Their current build is unaffected, but `rtapi_app_main()` has moved out to switchkins_main.c, so next time they re-sync that copy they either take that file too or call `switchkinsInit()` themselves. They also no longer need the copy, since switchkins.h is installed now and switchkinscomp.comp shows both routes.

Announcing it is worth raising at a weekly meeting. Nothing is removed in this PR.

docs/src/gcode/g-code.adoc
```diff
@@ -934,6 +935,69 @@ It is an error if:
...
+
+'G12.1' selects one of the kinematics provided by a switchable kinematics
+module, and 'G13.1' cancels back to kinematics 0. The 'P' word is the
+kinematics number, so 'G13.1' and `G12.1 P0` do the same thing.
+
+These are the way to select a kinematics. The `motion.switchkins-type`
+HAL pin does the same thing and takes the same numbering, but it is
+deprecated: the interpreter never sees it, so a program is read, its
+limits checked and its path looked ahead in whatever kinematics the
+interpreter last knew about, which need not be the one that runs it.
```
BsAtHome: The opening sentence is putting the reader on the wrong foot. I think you should start with something like "Deprecation notice: Using the HAL pin... fails to inform the interpreter of any change. -this-is-bad-blabla...". Maybe as an admonition.

src/emc/kinematics/5axiskins.c
```diff
@@ -59,7 +59,6 @@
 #include <rtapi_ctype.h>
 #include <hal.h>
 #include <emcmotcfg.h>
-#include <kinematics.h>
 
 #include "switchkins.h"
```
BsAtHome: This is a subordinate user and should use <> include.
(other places as well)

Only the code that is part of the _source_ must use double quotes. The implemented switching kinematics are not switchkins sources, but users. They should also be able to build out-of-tree, hence the <> include.

I know this is tricky. You have to think about it whether the include is something that is local or whether it is foreign.

src/hal/components/switchkinscomp.comp
```diff
@@ -0,0 +1,167 @@
...
+#define USE_TOPDIR(b) XSTR(TOPDIR/b)
+
+// switchkins.c provides kinematicsForward(), kinematicsInverse(),
+// kinematicsSwitch() and the rest of the kinematics interface, and
+// dispatches each call to the currently selected switchkins-type.
+// kins_util.c provides the identity kinematics and the coordinates
+// letters-to-joints mapping they use.
+#include USE_TOPDIR(src/emc/kinematics/switchkins.c)
+#include USE_TOPDIR(src/emc/kinematics/kins_util.c)
+
```
BsAtHome, line 68: Shouldn't this be provided as a library (`libswitchkins.a` archive lib) and then you add `--extra-link-args=-lswitchkins` to halcompile.
That way you can do it out-of-tree. Otherwise, what is the point of having switchkins.h pulled into the exported includes?

docs/src/gcode/g-code.adoc
grandixximo: Fair, will reword

src/hal/components/switchkinscomp.comp
grandixximo: I went looking for that path and could not find one, so let me ask it the way I asked it myself.

Which build path carries `-lswitchkins` to an RT module? `--extra-link-args` appears once in halcompile, in `build_usr()`, and the help says "only for userspace, only for .c files". A kinematics module goes through `build_rt()`: `obj-m += x.o`, then Makefile.modinc.

And what would kbuild do with the archive once it got there? Uspace links `$(CC) -shared ... $(EXTRA_LDFLAGS)`, so a PIC archive fits. A kernel module takes a `<module>-objs` list of objects. Is a route that builds on uspace but not on RTAI worth putting beside the TOPDIR `#include` that tpcomp.comp and homecomp.comp already use?

On the header, exporting it is what copies it into `include/`, which is how the four converted components compile `<switchkins.h>`, and how an out-of-tree one picks up `kparms`, `switchkinsRegister()` and the identity
kinematics instead of copying the declarations. Would you rather it were not exported?

BsAtHome: Yes,... Maybe the path that https://github.com/LinuxCNC/linuxcnc/tree/master/src/hal/drivers/mesa-hostmot2/modbus took is more workable? That has a template and installs a script to make the RT module.

src/emc/kinematics/5axiskins.c
grandixximo: I'll build a tool to catch these in the CI with a warning, and pre-commit locally
Fixed

src/hal/components/switchkinscomp.comp
grandixximo: Done, and it was where this started: switchkins.c and kins_util.c now install into share/linuxcnc beside mesa_modbus.c.tmpl, that directory is on the realtime include path, and the template reads `#include <switchkins.c>`. No TOPDIR, no extra script, halcompile builds it as it stands.

Tested by renaming the template, building it with halcompile and running it as a sim config's kinematics: G12.1 P1 switches to its own kinematics, G13.1 comes back, an out of range P is refused.

Sigma1912:
Just a thought:
Work offsets in one kinematics are usually not applicable in another one. So when using switchable kinematics I have often found it would be nice to have different sets of G54..G59 offsets.
Maybe a parameter offset set by the  'motion.kins-type' value?

grandixximo:
That matches what switchkins.adoc already warns about: coordinate system offsets are described as generally applicable only to the type 0 startup kinematics, with advice to reset them or update them before switching. So the problem is real, and today it rests on the operator remembering.

Can I ask what you switch between, and whether you cut in both types? I am trying to tell two situations apart. If both types leave the axis letters describing the same physical point, then your G54 values are still correct after the switch, and what is missing is having a different one become active. If one of the types is identity, the stored numbers stop describing anything, since X becomes joint 0 and may not even carry the same units, and then a different set does not help either; the values want clearing.

Which of these is it for you: does the zero point of the part sit somewhere different once you switch, or is the part zero the same and you simply want a different one of G54 to G59.3 selected?

On the mechanism, do you mean an association living in the INI, a coordinate system named per kinematics type? There is nothing like that today. Or do you mean it read at runtime? kins-type is a status pin, so it reports rather than commands, but you could net it into a motion analog input and branch on it with M66 in a sub, which would work now. I am unsure whether that belongs in the core or stays the sort of thing an integrator wires up, and your answer would tell me which problem I am actually looking at.

Sigma1912:
> Which of these is it for you: does the zero point of the part sit somewhere different once you switch, or is the part zero the same and you simply want a different one of G54 to G59.3 selected?

Here are two examples, the tool may cut in all kinematic modes.

Example1: Serial manipulator arm
https://github.com/LinuxCNC/linuxcnc/tree/master/configs/sim/axis/vismach/melfa-sim
To overcome singularities in industrial robots we need to be able to switch between joint and cartesian mode inside a gcode program. Here the values as well as the units change from one kinematic mode the other.

Example2: Tilted work plane
https://github.com/LinuxCNC/linuxcnc/tree/master/configs/sim/axis/vismach/5axis/table-rotary_spindle-rotary-nutating
The switch to TOOL/TWP mode requires a transformation of the offset values from one coordinate frame to another (rotated AND translated). The new work offset (that describes the same point on the work piece) has thus different values that must be activated after the switch. The original offsets must be restored after switching back.



Currently I have two ways of dealing with this:
1. Keep the offsets in separate WCS systems (eg use G54 for one and G59 for the other mode). Switching from one to the other is easily handled in the custom M command used to switch the kinematics

2. Stash the offset values in parameters (so they can be restored when switching back) and then overwrite the old ones.
This can also be handled in the custom M command but requires much thought to ensure that the correct offset values are restored in all circumstances. Generally I VERY much dislike to alter values in the offset table in the background. There is a lot that can go wrong (eg in case of an abort) 


> On the mechanism, do you mean an association living in the INI, a coordinate system named per kinematics type? There is nothing like that today. Or do you mean it read at runtime? kins-type is a status pin, so it reports rather than commands, but you could net it into a motion analog input and branch on it with M66 in a sub, which would work now.

Maybe a different set of G54...G59.3 values. One set for each kins-type. 
I have also thought about having G54...G58 for kins-type 0, G54.1 ... G58.1 for kins-type 1,  G54.2 ... G58.2 for kins-type 2 and so on (for backwards compatibility G59..g59.3 maybe the same for all)



There are also some other things to think about when working in different kinematic modes ( eg G28 , G30, G53, G92/G52, M6) it might be useful to have way of blocking certain Gcodes / Axiswords in a given mode. 
I have made a patch that allows blocking certain axis letters at runtime using a magic-comment. (Just as an illustration, not suggesting to use it)
https://github.com/LinuxCNC/linuxcnc/commit/ca9c334972c6dd2ec457df6979453d2c1d0a26bd

grandixximo:
Both examples are clear and they moved my view: I had been assuming identity was mostly a jogging mode, and you cut in all modes. I looked at how the commercial controls handle both cases, and in each of them the closest thing we could do in LinuxCNC turns out not to be an offset set.

For the singularity case, the target stays in Cartesian coordinates and only the interpolation changes: a move that runs the inverse once at the endpoint and then interpolates the joints linearly to reach it. Coordinates and units never change, so the offsets keep applying and there is nothing to switch. In tree that is a new move type from canon into the trajectory planner, and the existing KINEMATICS_INVERSE_FLAGS already carry enough to pin which solution branch it lands on. G-code would need a way to ask for it.

For the tilted plane, an extra stage in the offset chain rather than new values in it: a frame composed on top of the active G54, with its own cancel, so the blocks in between are programmed in the tilted plane while G54 itself is untouched. That is a stage in rotate_and_offset_pos() plus the modal state to go with it. Nothing is rewritten, so nothing needs restoring and on_abort_with_twp_reset.ngc has nothing left to do. Orienting the tool normal to the plane, your G53.1, does need the kinematics to say where the tool points and to solve for the rotaries, which is item 3 above.

On having more of them: that part has plenty of precedent and I would be happy to look at it separately. Fanuc adds G54.1 P1 to P48, or to P300 on the 30i, and Siemens adds G505 to G599. What I cannot find anywhere is a control that ties a set to the active kinematics or transformation. Selection is always explicit in the program, or it belongs to a channel or path. So more coordinate systems yes, but I would rather not bind them to the kinstype.

Which leaves me reading one set per kinstype as paying for the two absences above rather than as the thing you actually want. You have run this in production and I have not, so I would rather hear where that is wrong. Would a composing frame cover your case 2, or would it miss something?

Your other point stands either way. Per-mode restrictions on G28, G30, G53, G92 and M6 look right to me, and refusing outright beats guessing.

Sigma1912:
Nice that you have an overview of what other controllers are doing. Makes more sense adopting something that is established rather than coming up with something quirky that falls on our feet somewhere down the road. 
My big deficit (or at least one of them) is that I often have zero experience how things work  
 on commercial machines so I end up reinventing the wheel and often do it in a slightly odd manner because I don't really know what I'm doing. :) 

>Which leaves me reading one set per kinstype as paying for the two absences above rather than as the thing you actually want. You have run this in production and I have not, so I would rather hear where that is wrong. Would a composing frame cover your case 2, or would it miss something?

That is how I started out, an offset built into the kinematics. The reason I moved away from that was that in it's current form LinuxCNC is totally unaware of what is going on in a custom kinematics. 
So if we can handle things like G53 etc. internally then I'm _probably_ happy with not having different sets of offsets.

Basically what I'd like is to NOT have to tamper directly with the values in the offset table.

grandixximo:
That is the same conclusion from the other side, and it names the dependency. A composing frame is only worth having if motion knows the frame is there; otherwise G53 and G28 keep meaning whatever each module happens to do with them, and we are back to the operator remembering. That is items 3 and 4 above, which is why they come before anything in the offset chain.

What "aware" has to mean concretely: the module answers where the tool points and how joint motion maps to world motion, both as plain functions of the joint values and a parameter block, with no state hidden in HAL pins. Once the control can call the kinematics rather than only drive it, G53 resolves in machine axis space, the per-mode restrictions you listed get checked rather than guessed, and a frame composes on top of G54 with nothing rewriting the table.

So no per-kinstype offset sets, nothing tampering with the table in the background, and the tilted plane becomes a frame that cancels. Your two examples are the cases I will hold it to.

Sigma1912:
Sounds great!

grandixximo:
One more thing the tool round trip does not cover, found while giving `5axiskins` and `maxkins` the `.tool-offset` pin every other rotary module has.

Both modules add the W coordinate to the pivot to tip distance, so W is already a tool-length route: `G43.1 W5` followed by a block that commands `W0` puts the kinematics on the right radius and lands on the same joint solution as a Z-column offset does. It is easy to miss, because `G43.1 W5` on its own changes nothing. `convert_tool_length_offset()` pushes `w_current` down by the offset, canon adds it back, and a block with no W word hands the module zero. Command W once and it is correct from there on.

So on these two modules the tool length has three spellings: the W column with W commanded, the `sum2` into `pivot-length` that the bridge mill sim uses, and now a `.tool-offset` pin. They all compute the same radius and they all add, so using two double counts. None of them is written down anywhere, which is the actual problem.

This is the same shape as the rest of items 3 and 4. Each module infers the tool from whatever it happens to be handed, every config picks a different one of the three, and nobody owns which is meant. Once the tool arrives as a parameter at the call site there is one answer instead of three.

Sigma1912:
I have a set of kinematics that probably cover most 5 axis machine (tool side rotation, work side rotation  and mixed) including nutating joints and tilted work plane. I have long wonder if I should add those to the collection:
https://github.com/Sigma1912/LinuxCNC_Demo_Configs/tree/main/5axis-twp

grandixximo:
Yes, and I would take the kinematics first.

The nutating ones are what items 3 and 4 need most. A nutating head is where "where does the tool point" stops being obvious, so a weak default every module answers is only worth having if something in tree can break it. Today the only nutating configs are the two `trsrn` ones, which are yours already.

On the TWP remap I would rather you decide with item 5 in front of you. I am proposing to move that mechanism into `rotate_and_offset_pos()` as a frame that composes on top of G54 and cancels, which would leave the Python doing less or nothing. That does not make the configs useless, the opposite: they are the executable spec for what the frame has to do, and the baseline for checking whether a C version matches. But adding them as tree configs while I am proposing to replace the mechanism should be a decision rather than a side effect. I would rather have them in than not, and you know the ground better than I do.

One request if the kins come in: could they take the tool length through a `.tool-offset` pin the way the trt modules do? I have just been through this on `5axiskins` and `maxkins` and found three spellings of the same quantity with none of them written down. Three more modules each picking their own makes that worse, and settling it now is cheap.

Sigma1912:
I'm not in hurry it's been sitting here for a while now.
I have no problem making adjustments to make it fit better.

grandixximo:
I had a go at the split on the two trsrn configs in tree, to see how it lands: https://github.com/grandixximo/linuxcnc/tree/twp-remap-split

Your generic `remap.py` plus a `remap_funcs_twp.py` beside each config. Your two snrtr files carry exactly the two branches the tree had inline, `xyzacb-snrtr` is the (C, A) maths and `xyzbca-snrtr` is the (C, B) maths, so it was mostly substitution. Adapted for the tree: ini read through `linuxcnc.ini`, kinematics switched with `G12.1`, and `kins_set_values` converts the two joint angles to degrees because the modules in tree read those pins in degrees. Pin names unchanged, so neither `.comp` is touched.

I ran both configs through eight orientations under `G53.1 P0/P1/P2`, `G53.3`, `G53.6` and `G68.3` before and after, and it looks right: nothing that worked stopped working, and `P1`, `P2`, `G53.6` and `G68.3` now do something where they used to fail or silently no-op.

Three things I hit in your files:

`xyzbca-snrtr` has the wrong default tool-x. Both files aim for the same thing in the comment, the oriented x-vector parallel to the machine xy-plane. Tool x is the first column of `Rp*Rs*Rz(tc)`, so that is entry `[2][0] == 0`, and since `Rp` is `Rz` only the bottom row of `Rs` enters:

```
bottom row [-Sv*Ss, t, s]  ->  tc = atan2(Sv*Ss, t)
bottom row [t, Sv*Ss, s]   ->  tc = atan2(-t, Sv*Ss)
```

`xyzacb-snrtr` has the second row and the second formula and is right. `xyzbca-snrtr` has the first row and the second formula, so it orients tool y horizontal instead. Looks copied from the sibling. Invisible in use, which is presumably why it survived: the tool still points where it was told to, only x and y in the tilted plane swap, so `G68.3 R` starts a quarter turn off. I used the tree's form in the port.

`xyzacb_snrtr_kins.comp` declares `hal_float_t *virtual_rot;` twice, lines 34 and 38, so it does not compile.

`xyzacb-snrtr/remap_funcs_twp.py`, `kins_calc_primary`, vertical tool branch sets `theta_1 = 0` and then returns the empty list, so any `G53.n` to vertical fails. `xyzbca-snrtr` says `return [0]` there. I took `return [0]`.

The commit is authored by me at the moment and the code is yours, so I would set you as author before it goes anywhere. Happy either way: take the branch and send it yourself, or tell me it looks right and I will.

Sigma1912:
Thanks for looking through this. I'll need to have a good look and report back.

Sigma1912:
I have found a problem with G13.1/G12.1 when either is in a 'on_abort' call. 

Testing with `configs/sim/axis/vismach/5axis/table-rotary_spindle-rotary-nutating/xyzbca-trsrn_twp`

This config has [RS274NGC]`ON_ABORT_COMMAND = o <on_abort_with_twp_reset> call`

I have noticed strange behavior with the remapped gcodes after issuing MDI commands that trigger this routine.  For example, it looks like the ` g682` function gets stuck on `    yield INTERP_EXECUTE_FINISH`
  
You can get an indication of the issue if  you uncomment the first line in the on_abort sub:

``` 
o<on_abort_with_twp_reset> sub
(msg, on_abort START)    ; <<<< UNCOMMENT this line so the msg shows on start
M68 E2 Q0       ; reset twp-state to 'undefined' (0)
G13.1           ; back to identity kins
G64 P0.01       ; reset the toolpath tolerance as this sometimes gets set to zero on estop events
G54             ; switch to G54
(msg, on_abort END)
o<on_abort_with_twp_reset> endsub
M2
``` 

Now start the config, clear estop and switch the machine on which calls the ` on_abort` routine. Note though that it does apparently NOT reach the end since we don't get the message on the last line: 

on abort


Further testing has shown that having either a `  G13.1 ` or ` G12.1 P0` in the ` on_abort` routine triggers the problem while the former `M68 E3 Q0` does not.

grandixximo:
Good catch, and it is my bug. Reproduced and fixed.

`G12.1`/`G13.1` set a flag that makes the interpreter wait for motion to drain, so nothing is planned across a change of kinematics. `Interp::on_abort` runs the routine with a single `execute()` call that cannot service `INTERP_EXECUTE_FINISH`, and `emcAbortCleanup` treats that return as normal because it only checks `status > INTERP_MIN_ERROR`. So the sub is truncated at the first queue-busting code with no error anywhere. `M68 E3 Q0` sets no flag, which is why it was fine.

The fix is to only ask for the drain when there is something to drain:

```c
if (!GET_EXTERNAL_QUEUE_EMPTY()) {
  settings->kinsSwitch_flag = true;
}
```

By the time the switch is emitted, everything before it has already gone to canon, so an empty queue means there is nothing to wait for. During an abort the queue has just been flushed, so that case works by construction rather than by special-casing.

To check it without reading messages I used the sub's last line: set `G55`, abort, read `g5x_index`. It stays at 2 before the fix and returns to 1 after.

It fixes more than the abort routine. Running the config through eight orientations under `G53.1 P0/P1/P2`, `G53.3`, `G53.6` and `G68.3`, two commands that errored now succeed and none went the other way. Same cause: the remap issues `G12.1 P2` and hit the same truncation, which I think is the `yield INTERP_EXECUTE_FINISH` you were seeing.

Pushed to the branch. Worth noting the general trap is older than `G12.1`: any queue-busting code in an abort routine is dropped the same way, which is why `on_abort` pre-clears the toolchange, probe and input flags, and why that .ngc says only custom M-codes are safe there. I put `G13.1` in it without heeding that.

grandixximo:
Chased the one command still failing in my earlier run, `G53.1 P1`. Your split fixes it.

`kins_calc_primary` in the in-tree `remap.py` has three problems stacked: the append sits outside the loop over candidate secondary angles, so only the last one ever yields a primary; `theta` is radians while the limits are degrees; and the primary angle is checked against the secondary joint's limits. With one surviving candidate and `P1` asking for a positive rotation, nothing qualifies.

Your version fixes the first two. On `xyzbca-trsrn`, eight orientations under `G53.1 P0/P1/P2`, `G53.3`, `G53.6` and `G68.3`: 35 done and 1 error before, 36 and 0 after, with `G53.1 P1` landing at C +32.75, A -43.07.

The third is still in both snrtr files and I pushed a one line change for it. The function declares `primary_min_limit` and `primary_max_limit` and then compares against the secondary pair. Invisible on the two configs in tree, where the primary C is the wider joint, runs byte-identical before and after on both machines. It bites the other way round, on a machine whose primary is tighter than its secondary. Shout if you read it differently.

Sigma1912:
Nice to have someone else looking it over! It's been a while since I was in there. 

A small thing, but there is this printing in the terminal (not sure if you still need that):
`gcodemodule: SELECT_KINS_TYPE`

Another thing is that I have found some issues with rounding errors. 
Example using ` configs/sim/axis/vismach/5axis/table-rotary_spindle-rotary-nutating/xyzbca-trsrn_twp` :
- Change the `nutation-angle` slider to 90° 
- MDI: g68.2 q121 i0 j5
- MDI: g53.1
Results in a math error because the division results to ` -1.0000000000000016` which is then outside the range for `asin()`

https://github.com/grandixximo/linuxcnc/blob/24979ac96abacec8428fa1fe502e684a7140ac00/configs/sim/axis/vismach/5axis/table-rotary_spindle-rotary-nutating/xyzacb-trsrn_twp/remap_funcs_twp.py#L141

So to be on the save side I suppose we would need to harden all those arcus functions by rounding to, say, 10  digits . 
Question: Since this implementation is probably all only temporary is it worth going through the python code to fix this?

BsAtHome:
If you want to be on the safe side, then you need to round to the statistical bound, which is 8 decimal digits or half the binary digits (sqrt(log10(2^53)) = 53/2*log10(2) ≃ 8). More digits are only available if you can actively control the error propagation throughout the calculation(s).
If you know the upper/lower bounds of the source to fit in about ±2^36, then you could do `((int64_t)(sourceval * 0x1p26)) / 0x1p26` to round to 26 binary digits or 7.8 decimal digits. You can also use 0x1p27 for 8.1 decimal digits. You may want to have an eye out for affinity to ensure that both positive and negative bounds truncate correctly. You should use this kind of trickery sparsely.

A (very much so) mean trick would be (works only on IEEE754):
```C
union {
    double d;
    int64_t i;
} x;

x.d = sourceval;
x.i &= 0xfffffffffc000000L; // Set the 0-mask to the bits to truncate
printf("%.17lf %.17lf ", sourceval, x.d);
```
This can have slightly different results than multiply/divide if not careful, but is faster because it is integer trickery. When using it on `0.1`, then it will reduce to `0.09999999...` because `0.1` cannot be represented correctly using binary floating point.

Sigma1912:
> If you want to be on the safe side, then you need to round to the statistical bound, which is 8 decimal digits or half the binary digits

For a casual programmer like me, is that a general rule of thumb for any language (eg python)?

grandixximo:
Yes, I need this, because they are the ideal compare test cases, I can keep iterating and testing, to make sure the kinematics are stable throughout the conversion.

BsAtHome:
> > If you want to be on the safe side, then you need to round to the statistical bound, which is 8 decimal digits or half the binary digits
> 
> For a casual programmer like me, is that a general rule of thumb for any language (eg python)?

Just for starters, normally a casual user/programmer does not care. Only when it starts to hurt, then you need to care and act.

It is in principle applicable to all systems that use a non-exact representation. The problem is always repeating fractions and irrational numbers. Note that repeating fractions depend on the number base used (number base 2 represents 1/10 as a repeating fraction, but number base 10 is exact in that case).

The "8 digits" come from the statistics of error generation and propagation. When you try to converge a calculation, then the accuracy bound is typical sqrt(floating_point_accuracy). And yes, it is a rule-of-thumb, but i can be backed up. It can be improved when you know exactly how errors are generated and propagated. For example (last time I looked), the Fast Fourier Transform library fftw uses some smarts that makes it have about 12 decimal digits of accuracy.

And, yes, Python has the double precision floating point as underlying type, so it suffers from the same error propagation problems as most other programming languages. Python has the `fractions` and `decimal` modules for specialized calculation when normal floating point is insufficient.

Mostly, you can get away with some EPSILON calculation. You often see that in direct or indirect comparison with zero (0.0) because matching zero exactly is very hard:
```C
#define EPSILON 1e-8

if(fabs(valuea - valueb) < EPSILON) {
    // Both valuea and valueb are closer together than EPSILON
}
```

Similar tricks you can also apply when working with asin/acos and friends.

grandixximo:
Both fixed and pushed.

The print was mine. `gcodemodule.cc` is empty now like the canon stubs around it. `saicanon.cc` had the same line, and there it should report, but through `ECHO_WITH_ARGS` like the rest of that file rather than a raw `printf`.

The arc problem is bigger than the case you found. Putting every reachable orientation back through `kins_calc_possible_joint_angles`, primary and secondary swept in five degree steps, failures out of 5184 are 64, 44, 47, 25, 107 and 20 at nutation 15, 30, 45, 60, 75 and 90. So it is not about 90 degrees: it fails at the nutation the configs ship with, and 90 is one of the milder ones. Every failure is in `kins_calc_primary` at primary 0 or 180, so every failing tool vector has a component at zero, which is what `G68.2 ... I0` asks for.

Worth fixing this much, then, even though the python is on its way out: the replacement has no caller yet, so these configs are what ships meanwhile, and a clamp and an early return leave nothing to unpick. Going through the whole file for error propagation I would leave.

@BsAtHome thank you, that is the right diagnosis. I clamped rather than rounded. The argument is exactly plus or minus one in exact arithmetic, so pulling it back when it is already within a rounding error is exact where it applies, and anything further out is an orientation the machine cannot reach and should still raise; rounding to the statistical bound would swallow those too. 1e-9 for the width, against an observed overshoot of about 2e-16.

One more thing turned up while testing: `kins_calc_possible_joint_angles` logs the failure and then falls through to `return (theta_1_calcd, theta_2_calcd)` with that variable never assigned, so the domain error arrives as an `UnboundLocalError` on top of the real one. It returns no solution now, which is what the caller already handles.

With both changes all 5184 orientations solve at every nutation angle tried, on both machines.

Sigma1912:
> The commit is authored by me at the moment and the code is yours, so I would set you as author before it goes anywhere. Happy either way: take the branch and send it yourself, or tell me it looks right and I will.

Apologies, forgot, just go ahead. Glad if it finds a use somehow.

grandixximo:
@BsAtHome a shape question before I build this into a PR, since you decided half of it once already in #3807 and I would rather not relitigate it by accident.

Item 4 needs a kinematics module to be callable outside RT, so a limit check or a preview can evaluate it at poses the machine has not reached. Two ways to give it its geometry, both built and measured.

Yours from #3807: the module publishes a fixed-layout struct through `hal_struct_newf()` and exports `nonrt_attach()`, and userspace reads the snapshot. Joining costs a module five things: parameters in a shared header, the struct registered, a refresh inside forward and inverse, an `if (!haldata)` branch, and its maths factored to take geometry as an argument.

The other binds a non-RT copy of the module's `haldata` to the value cells the running RT instance already owns, so its own forward and inverse work unmodified on live values. One `nonrt_attach()` calling `hal_get_pin_value_by_name()` per pin, and nothing else.

Same two modules, loader and Jacobian excluded since they are identical: +715/-67 with 252 HAL lines and a `HAL_VER` bump, against +158/-10 with one HAL line. Identical limits from both, to the digit.

That one line is `EXPORT_SYMBOL(hal_get_pin_value_by_name)`. The function exists and returns the right pointer, it was just never exported. That is what I want your read on, since handing out an interior pointer into HAL's memory points the opposite way from #4458 and from the untangling you described.

So: pin binding with that export, or the struct interface on the grounds that HAL should hand out a declared blob rather than a pointer into itself? Either is fine by me. Note `hal_struct_*` is not in master, it lives in #3807, so choosing it means landing that first.

BsAtHome:
> @BsAtHome a shape question before I build this into a PR, since you decided half of it once already in #3807 and I would rather not relitigate it by accident.

If you want to "(re-)litigate", please, let me call a defense lawyer ;-)

> [snip]
> That one line is `EXPORT_SYMBOL(hal_get_pin_value_by_name)`. The function exists and returns the right pointer, it was just never exported. That is what I want your read on, since handing out an interior pointer into HAL's memory points the opposite way from #4458 and from the untangling you described.

The function `hal_get_pin_value_by_name()` will be retired as it is replaced by the query API.

The trick is that you can get the reference of the pin and then use that reference. There is no need to look it up by name all the time, unless there is a chance that there will be of signal connect/disconnect events while running. You call `hal_getref_p()` and the pin reference will be there. If you want the value, you can call `hal_get_p()` with the callback set to NULL and it will simply get you the value. When there is no chance of signal connect/disconnect events, then you can use the reference returned by `hal_getref_p()` and use the appropriate getter.

If the pins are not supposed to be connected to, then you can make sure of that. You should then not make them pins but params. You simply get the param's value references with `hal_getref_p()` and use the getter. The param's value references will stay valid throughout the lifetime of the component.


> So: pin binding with that export, or the struct interface on the grounds that HAL should hand out a declared blob rather than a pointer into itself? Either is fine by me. Note `hal_struct_*` is not in master, it lives in #3807, so choosing it means landing that first.

Essentially, if we can use existing infrastructure, then that is preferable.

grandixximo:
@BsAtHome thank you, that settles it, and it agrees with what is already on the branch.

The `EXPORT_SYMBOL` I asked about was against a revision I had already replaced and did not check before writing. The resolver calls `hal_getref_p()` once per pin and keeps the reference, and the module reads it through the typed getter. `git diff switchkins-unify..kins-nonrt-bind` touches nothing under `src/hal`. So: pin binding on existing infrastructure, no HAL change, no dependency on #3807. I will build it that way.

Params are not an option here: `5axiskins.pivot-length` is netted from the tool offset in the bridgemill sim, which is the case the binding has to follow.

BsAtHome:
Just to be clear here... When you require pins and your code is reading, then your code should create a component and a set of input pins which must then be connected to the data output pins you want have the data from. Not doing so would create a situation where you must track pin connection on every read, which means that you have to do a hal_get_p every time for the pin (and that is slow). Using your own component with pins means that you are the owner of the reference and it will always point correctly, regardless whether it is connected or not.

Sigma1912:
Nice progress!

**Question:**

> The orientation codes need the TCP kinematics (G12.1 P1). P picks the solution, nearest to the present rotary position first. Q0 holds the table and lets the head do it, Q1 lets the table take part as well.

Looks like you have changed the usage of the 'P' word in the 'G53.n' commands. What would be the use case for this particular implementation? It seems to me that in order to choose a particular solution 'P<n>' the operator would have to know the possible solutions at the time of programming.
Note that I cannot argue this point from any hands on experience as I don't have a machine with motorized tool orientation capability.



**Bug Report:**

Testing with 'configs/sim/axis/vismach/5axis/table-rotary_spindle-rotary-nutating/xyzacb-trsrn_twp' shows a bug that seems to block 'G69' from cancelling the work plane:

1. start the config, home the machine
2. open the LinuxCNC Status window
3. start the example gcode and let it run a few seconds.
4. stop program execution (STOP button in the gui)

Note that 'g68_active' remains '1' after the program stop and MDI 'G69' cannot deassert it.

 
Screenshot from 2026-09-07 15-51-35


Thanks for putting in all the work!

grandixximo:
@Sigma1912 thank you for both, and for taking the time to drive it. Feedback like this is what keeps the work from feeling like effort poured into a void.

The stop is fixed on the branch, with a regression in `tests/twp-native` that runs a program with a plane, stops it mid-move and checks that the plane clears and that a `G69` afterwards is accepted. Please give it a go on your machine.

On the `P` word you talked me out of what I built. It was an index over the poses that reach the plane, ordered nearest first from where the machine happened to be standing, so the same `P2` could select a different pose on a different run. It now names the pose instead of counting: no `P` still takes the nearest, `P1` and `P2` name the two poses by the sign of the secondary rotary, the one whose axis the other carries, which is the joint that leans the head or the table each way. One plane on the nutating sim, from two different starting positions:

| | secondary rotary |
| --- | --- |
| `P1`, from either start | 51.127 |
| `P2`, from either start | -51.127 |
| no `P` | 51.127 from one start, -51.127 from the other |

Which rotary that is comes out of the module's own maths rather than a config entry: the interpreter evaluates the tool frame either side of each joint and the one whose axis is carried by the other is the secondary. A machine without exactly two such rotaries, a robot wrist, gets no labels and says so rather than guessing, and `tests/kins-params` checks the pair every module in the tree reports.

Naming the branch rather than counting is what the commercial controls do, so a program moved across reads the same way, and it is close enough to what your `P1` and `P2` asked for that a program written for the remap usually keeps its intent.

One thing still worth your view, since it is where the two differ. The remap's modes constrained the direction of travel and mine constrains the destination, so they part company when a rotary is walking towards its limit. Do you use that in practice, or is taking the short way round enough? The short way itself I am fixing separately: `G53.1` ranks by raw joint difference, so a C at 180 asked for a pose at C -90 turns 270 degrees where 90 would do, and unwrapping that needs the joint limits, which is item 6.

Sigma1912:
> The stop is fixed on the branch,

Yes, that works now, thanks.

> Do you use that in practice, or is taking the short way round enough? 

That is fine for me as my application has a manually operated swivel head. The reason I put that in was because I found it in a Mazak manual. As long as we are following some kind of precedent we should be on the right path.

One more thing, when testing 'configs/sim/axis/vismach/5axis/table-rotary_spindle-rotary-nutating/xyzbca-trsrn_twp'. On startup there is an error message from the interpreter running on load however the program actually runs just fine:

Screenshot from 2026-09-08 06-44-34

grandixximo:
Yeah, I noticed that as well, I thought cause I had G54 way out somewhere, I will double check, thanks...

grandixximo:
@Sigma1912 I traced the error, and it is because axis loads the gcode, before loading the postgui configuration, one way to fix, is to have the pivots set with the HALCMD in the .ini
but that is ugly and duplicates where the pivots live, and is confusing for new users, I would rather change axis, but that would be a separate PR, I will prepare that, and will see if it lands, other UI have similar issues.

Sigma1912:
Nice, thanks for figuring that out.

Sigma1912:
Testing 'G53.7' on 'configs/sim/axis/vismach/melfa-sim/melfa-sim-three21' shows a bug:

1. Start config and home machine
2. MDI: g53.7 g0 j4=0  ; OK
3. MDI: g53.7 g0 j4=90  ; OK
4. MDI: g53.7 g0 j4= -10  ; OK
5. MDI: g53.7 g0 j4=90  ; ERROR joint interpolated move on line 0 would exceed joint 3's limit

There seems to be an issue with the sign change. Looks like it wants to maybe twist the forearm (J3) 180° when it should really just move joint 4.
Screenshot from 2026-09-08 10-33-12

grandixximo:
@Sigma1912 fixed and pushed, with a regression in `tests/ptp-robot`. Please give it a go once CI is green.

Sigma1912:
Now 'G53.5' and 'G53.7' don't work at all. Fail with:

``` 
emc/task/emctask.cc 71: interp_error: the kinematics cannot invert the current position
the kinematics cannot invert the current position
ERR kI - compute_jinv (joints: 0.000000 0.000000 0.000000 0.000000 0.000000 0.000000), (iterations=0)
```

grandixximo:
@Sigma1912 fixed and pushed as `5ce95b4cda`.

The interpreter works out the joints behind a point by inverting it, and had nothing but zeros to start the first inverse from. genserkins takes the Jacobian at the seed, and a serial arm with every joint at zero stands in a singular pose, so it refused before it began. Canon now reports the joint positions and the interpreter falls back on them when its own seed no longer explains where the machine is. That also names the arm the machine is standing in, which inverting a point cannot: the same point is reached again with the elbow the other way up.

On `melfa-sim-genser` here, homed at `0 -90 0 0 90 0`:

```
G0 G53.7 J4=80          joints  0.0000  -90.0010   0.0010  0.0010  80.0000   0.0010
G0 G53.7 J0=10          joints 10.0000  -90.0010   0.0010  0.0010  80.0000   0.0010
G0 G53.4 X300 Y0 Z400   joints -0.0009 -133.6644  56.0467  0.0012  67.6177  -0.0002
G0 G53.7 J5=45          joints -0.0009 -133.6644  56.0467  0.0012  67.6177  45.0000
```

One question, since I have not been able to test it: did you hit this in MDI only, or does loading a program that contains G53.7 fail the same way? The preview has no machine to ask for the joints, so it starts from world zero, which a serial arm cannot reach, and I expect it to refuse there too. If it does, that is a second fix and it needs a decision about what pose the preview starts from, so I would rather know before shaping it.

Two other things you may run into on that config, both separate from this: G68.3 answers "the kinematics reports no tool frame", because genserkins does not supply the work and tool frames yet, and G53.5 refuses the axis letters, because the INI declares the joints angular while X Y Z imply length, so G53.7 J0= is the way in.

Sigma1912:
>  did you hit this in MDI only, or does loading a program that contains G53.7 fail the same way? 

Fails the same way.

> Two other things you may run into on that config,

Noted, thank you.

grandixximo:
@Sigma1912 the load is fixed too, pushed as `146ef1c41e`.

Two things were in the way. The preview had no machine to ask where the joints stand, so it started every inverse from zeros like the interpreter did; it now asks the canon, which already reads the status buffer for the tool table and the offsets. And a kinematics module loaded in a GUI could not reach the HAL library it calls, so on genserkins the failed inverse took AXIS down with it rather than reporting. Both AXIS and the gremlin and Qt screens pick the change up, since they share the one canon mixin.

Worth a look on `melfa-sim-three21` as well: there the preview never complained, it just previewed the moves from a pose the machine was not standing in.

Sigma1912:
Thanks for the update.  Works nicely with the analytical `three21` kinematics but has problems with the iterative `genserkins`. (I've attached the test program below.)

**Details**

Testing with `configs/sim/axis/vismach/melfa-sim/melfa-sim-three21` works nicely:

Screenshot from 2026-09-09 09-36-33



Testing with `configs/sim/axis/vismach/melfa-sim/melfa-sim-genser` fails with the following error (Line17: `G0 Y150` but this changes to other lines if that one is removed ):


```
USRMOT: ERROR: invalid command
Linear move on line 17 would exceed joint 2's positive limit max:[76.000000]
Linear move on line 17 would exceed joint 4's negative limit min:[-120.000000]
invalid params in linear command

```




```
(Set the start pose to make sure the kinematics can handle it)
G90
G53.7 G0 J0=0 J1=-90 J2=0 J3=0 J4=90
(Go to start coordinates)
G53.4 G0 X450 Y200 Z350 B10 C45
(Rectangle)
G91
G0 Y150
X150
Y-150
X-150
(Go different pose)
G90
G53.7 G0 J0=90 J1=-90 J2=0 J3=0 J4=90
(Rectangle)
G91
G0 Y150
X150
Y-150
X-150
(Go to different coordinates)
G90
G53.4 G0 X450 Y-200 Z150 B0 C0
(Rectangle)
G91
G0 Y150
X150
Y-150
X-150
(Go to different coordinates)
G90
G53.4 G0 X50 Y-600 Z150 B10 C-80 a90
(Rectangle)
G91
G0 Y150
X150
Y-150
X-150
(Go to start pose)
G90
G53.7 G0 J0=0 J1=-90 J2=0 J3=0 J4=90
(Go different pose)
G90
G53.7 G0 J0=90 J1=-40 J2=-40 J3=0 J4=90
M2
```

grandixximo:
I'll look into it, I found another issue with G12.1-G13, I had to re-stack all my commits to keep git history somewhat clean, will push the re-stacked branch soon, then will look at a fix for genser...

grandixximo:
Fixed and pushed. Two separate bugs were behind that error, both on the branch now.

@Sigma1912 Could you retest your program on `melfa-sim-genser`?

Sigma1912:
Works! Amazing stuff, would love to run this on my real robot but I'll have to wait for the SSCNET driver to be updated to 2.10 which is likely to be a while yet.

greatEndian:
@grandixximo — you're authorized to port my G43.4 / G43.5 vector-G-code work onto this branch and rebase/fix it as needed while this PR evolves.

The two commits (currently at the base of my `rtcp` branch):

- `766b9a451f` — G43.4: modal group 8 registration + command-driven switchkins switch from G-code
- `4ef0cbd8ba` — G43.5: I/J/K tool-axis vector on G0/G1 solved to A/B rotary words (xyzab_tdr topology)

Everything stacked above those on `rtcp` is separate RTCP topology work and isn't part of this. The ABC/UVW extension is droppable — I won't be following up on it, and the tool-frame interface here supersedes the per-topology hand solve.

Sigma1912:
Not that it really matters but the `W` axis in `5axiskins` wasn't really meant to drive a physical joint (ie a 'quill'). It's motion was meant to be executed by the linear axes (xyz) depending on the orientation of the tool. (The functionality was a bit quirky if I remember correctly)

```
  5) W axis motion is incorporated into the motion of the
*     joints used for X,Y,Z positioning and no motor or
*     hal pin connections are required for the joint specified
*     as JW.  However, a joint must be configured for W to
*     support display of the W axis letter value for
*     complicated reasons. (motion/control.c computes joint
*     positions only for the number of configured kinematic
*     joints (NO_OF_KINS_JOINTS) and the joint positions
*     are needed to display axis letters via inverse
*     kinematics.
``

grandixximo:
@Sigma1912 you were right on the history. W enters in 119d7cf98d ("plunging/drilling with W", November 2007, three days after the module itself) already folded into the XYZ joints; no motor was ever meant for its joint. My quill reading was wrong.

Cleanup done on the branch: the quill is out of the bridgemill vismach model, the m64/m65 drawing lock is gone with it, and the two demos that drilled by moving the virtual W joint are removed. The historical demo is back as 5axisgui.ngc and drills the honest way, the XYZ slides doing the stroke. Only the preview stays blind to it, same as 2007.

On a separate note: did you get the email from Brad Collette?

Sigma1912:
I didn't mean to say that you had to remove it. If it serves a purpose to have the quill then by all means change it as I'm not sure how much use the blind 'W' axis implementation has anymore.  

> On a separate note: did you get the email from Brad Collette?

Yes, we're in contact but I'm afraid I wont be joining you guys on discord as long a they insist on having my phone number. Certainly happy to communicate through other channels though.

grandixximo:
The quill was there mostly to explain the trick, I kept W for compatibility, the new 5axis setups are definitely moving to your design with the remaps at the moment, there are probably few people running the 'W' axis trick, I am one of them, but I will have to transition, just trying to show the path for transition in the configuration. For Discord email should be enough, phone is for 2FA which I don't think is enforced, maybe the channel of freeCad has security, but we could move the conversation in a normal chat without the fuss...

Sigma1912:
Well, I just tried again to verify by email and it seems that I'm simply not receiving any of their mails (checked spam and everything). Searching for 'not receiving emails from discord' makes me think that I'm not the only one with this problem. Might try again tomorrow...

grandixximo:
Verification is not even necessary to login, I logged in and joined the chat without having to verify, verification email came through within few minutes, but I was already in at that point...
https://discord.gg/DjMAsZWv
Try to join here

Sigma1912:
Here is what I get:
complete
popup

When I click on 'Verify' the button changes to 'Email sent!' and then to 'Resend Email' but I receive no emails  no matter how many times I click this button.
Exactly this situation:
https://support.discord.com/hc/en-us/community/posts/22618616249495-I-don-t-receive-a-password-reset-email

Sigma1912:
Note in  'configs/sim/axis/vismach/5axis/bridgemill' running the example program errors with: 
`G53.2 needs a kinematics type that describes the machine; select it with G12.1 first`

Also, we might want to think about showing the workplane in the preview.

grandixximo:
Thanks, reproduced. The demo programs (and the M428 remap) said `g13.1` meaning "back to type 0", from before `G13.1` resolved to the kinematics the module declares as identity, which on 5axiskins is type 1; so they landed on the identity kinematics and `G53.2` refused. They now say `g12.1 p0`, fixed on the branch. The M428 remap has the same problem on master since #4414, that one goes in as a separate PR.

Showing the work plane in the preview: agreed, noted for later.

grandixximo:
A note from a Discord conversation with David (Sigma1912) that changed the shape of G53.5 on this branch.

David asked whether G53.5 would hold up on his `multiaxis_millturn` config, a mill-turn with a composite Y made by two slides that do not line up with the machine frame. It would not have: G53.5 read the axis letters as slides, so `Y10` there would have meant joint 1, not machine Y, and his type 0 is not the joints at all but a real transform he uses as G53 without the orientation, `M428` then `G53 G0 X500 Y300 Z900` in his tool change.

G53.5 is now a point in the machine frame with the tool left out, XYZ the pivot in machine coordinates and the rotary letters as the machine frame kinematics reports them, through whichever type the module flags as its machine frame: a third flag next to identity and primary, `.machine = 1` on the ops table entry or `switchkinsDeclare(0, KINSTYPE_MACHINE)` on the old style. A module that declares none has its identity type stand in, so every in-tree machine tool module behaves as before; identity stays a separate flag because motion and the planner skip the maths on it. There, once type 0 is flagged, `G53.5 G0 Y10` puts machine Y at 10 whatever mix of joints 0 and 1 makes it, and a letter left out holds its machine coordinate rather than a joint. G28.5 and G30.5 read their stored position the same way, and G13.1 and G49 cancel to that type.

I took it one step further to see whether the same reading holds on a robot. It does: pumakins and genserkins flag their arm kinematics as the machine frame as well as the working transform, since the world they report is the flange in the base frame with no tool in the maths, ABB's `tool0` and KUKA's `$NULLFRAME`. G53.5 Z on a robot moves the flange along base Z holding everything else, interpolated in joint space, and G13.1 and G49 leave a robot on its arm kinematics. The three roles line up with Siemens' MCS, BCS after TRAANG, and TRAORI.

`tests/ptp-machine-frame` builds a slanted-slide module out of tree and checks it, `tests/ptp-robot` does the robot on pumakins.

PR #4375, opened 2026-08-12
include: use angle brackets for exported headers, and check it

grandixximo:
This has bitten me many times in review, and it is a rule a script can check, so here is the script plus the tree made consistent with it.

The build copies every SRCHEADERS entry into `include/`, so an exported header exists twice: the source under `src/` and the copy modules compile against. A quoted include searches the includer's own directory first, an angled include does not, so the two forms can reach different copies of the same header. Both compile, always, which is why it is easy to get wrong and invisible until somebody reads the diff.

The implementation keeps quotes, wanting the source beside it rather than a stale export. Everything else is a user, builds out of tree where only the exported copy exists, and takes angle brackets.

20 includes change, no code.

`scripts/include-style-check.py` reads the SRCHEADERS list, so it follows what the build exports rather than a list of its own. Findings are warnings by default and errors with `--error`, which is how the cppcheck job runs it. Named files can be passed, which suits a pre-commit hook. It annotates the offending lines when it runs in CI, so the next case shows up on the diff before a reviewer has to write it out.

One judgment call sits in the script rather than in the rule: a header whose own directory also holds code that only uses it needs its implementation named. `inifile.hh` and `hal.h` are those cases in tree today.

Full build clean. The check reports nothing on the result, and reports each of the 20 again when they are put back one at a time.

### DIFF
--- a/scripts/include-style-check.py
+++ b/scripts/include-style-check.py
@@ -1,0 +1,177 @@
+#!/usr/bin/env python3
+#
+# Check the include style of exported headers
+# Copyright (C) 2026 L. Toniolo
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU General Public License version 2 or later.
+#
+# The build copies every SRCHEADERS entry into include/, so an exported header
+# exists twice: the source under src/ and the copy every module compiles
+# against. A quoted include searches the includer's own directory first, an
+# angled include does not, so the two forms can reach different copies and both
+# compile.
+#
+# An exported header travels to include/ and must take its siblings with it, so
+# it includes them with quotes and finds the copies beside it wherever it ends
+# up. So does anything else in its directory, wanting the source next to it
+# rather than a stale export. Everything else is a user, builds out of tree
+# where only the exported copies exist, and uses angle brackets.
+#
+import sys
+import os
+import re
+import getopt
+import subprocess
+
+error_on_warning = False
+
+# The script lives in scripts/ and the sources are one level up, so it runs
+# from anywhere.
+topdir = os.path.normpath(os.path.join(os.path.dirname(os.path.realpath(__file__)), ".."))
+
+SUFFIXES = (".c", ".cc", ".cpp", ".h", ".hh", ".comp", ".icomp")
+
+RE_QUOTED = re.compile(r'^[ \t]*#[ \t]*include[ \t]*"([^"]+)"', re.M)
+# A .comp only becomes C below its ;; line.  Above it halcompile takes its own
+# include directive, 'include "bla";', which is not a preprocessor line.
+RE_COMP_QUOTED = re.compile(r'^[ \t]*include[ \t]*"([^"]+)"[ \t]*;', re.M)
+RE_SRCHEADERS = re.compile(r"^SRCHEADERS\s*:=\s*\\\n((?:.*\\\n)*.*)$", re.M)
+
+
+def usage():
+    print("""Check the include style of exported headers.
+Usage:
+  include-style-check.py [-e] [-h] [file...]
+
+Checks every source file under src/ when given no file, which is how CI runs
+it. Named files are useful from a pre-commit hook.
+
+Options:
+  -e|--error    Treat findings as errors (--enforce is accepted as well)
+  -h|--help     This message
+""")
+    sys.exit(2)
+
+
+messages = []
+
+#
+# Collect messages
+#
+def pfind(path, lineno, msg):
+    global messages
+    messages.append((path, lineno, msg))
+
+def flush_messages():
+    kind = "error" if error_on_warning else "warning"
+    for path, lineno, msg in messages:
+        print("{}:{}: {}: {}".format(path, lineno, kind, msg))
+    # Annotate the offending lines when running under CI
+    if os.environ.get("GITHUB_ACTIONS"):
+        for path, lineno, msg in messages:
+            print("::{} file={},line={},title=Include style::{}".format(kind, path, lineno, msg))
+    if not messages:
+        return 0
+    return 1 if error_on_warning else 0
+
+
+def exported_headers():
+    """Map each exported header's basename onto its path under src/, taken
+    from the SRCHEADERS list the build installs into include/."""
+    with open(os.path.join(topdir, "src", "Makefile"), encoding="utf-8", errors="replace") as f:
+        m = RE_SRCHEADERS.search(f.read())
+    if not m:
+        print("No SRCHEADERS list found in src/Makefile", file=sys.stderr)
+        sys.exit(2)
+    headers = {}
+    for line in m.group(1).split("\n"):
+        entry = line.strip().rstrip("\\").strip()
+        if entry:
+            headers[os.path.basename(entry)] = "src/" + entry
+    return headers
+
+
+def tracked_files():
+    """All tracked files under src/, and of those the ones worth reading."""
+    try:
+        out = subprocess.run(["git", "-C", topdir, "ls-files", "src"],
+                             capture_output=True, text=True, check=True).stdout
+    except (OSError, subprocess.CalledProcessError) as err:
+        print(err, file=sys.stderr)
+        sys.exit(2)
+    # A tracked path can be absent from the working tree, deleted but not yet
+    # committed or mid-rebase. The compiler would not see it either, so drop it
+    # rather than fail on the open() below.
+    tracked = {f for f in out.split("\n")
+               if f and os.path.isfile(os.path.join(topdir, f))}
+    return tracked, sorted(f for f in tracked if f.endswith(SUFFIXES))
+
+
+def check_quoted_includes(tracked, files, headers):
+    exported = set(headers.values())
+    for path in files:
+        if path in exported:
+            # An exported header is copied to include/ and has to keep finding
+            # its siblings there, so it includes them with quotes.
+            continue
+        directory = os.path.dirname(path)
+        with open(os.path.join(topdir, path), encoding="utf-8", errors="replace") as f:
+            text = f.read()
+        found = list(RE_QUOTED.finditer(text))
+        if path.endswith((".comp", ".icomp")):
+            head, sep, _ = text.partition("\n;;\n")
+            found += RE_COMP_QUOTED.finditer(head if sep else text)
+            found.sort(key=lambda m: m.start())
+        for m in found:
+            name = m.group(1)
+            source = headers.get(os.path.basename(name))
+            if source is None:
+                continue    # not an exported header, nothing to say about it
+            # A file of that name beside the includer is a different header
+            # that happens to share the basename, not this one.
+            local = os.path.normpath(os.path.join(directory, name))
+            if local != source and local in tracked:
+                continue
+            if directory == os.path.dirname(source):
+                continue        # beside the header, taking the source copy
+            why = "exported by {}".format(os.path.dirname(source))
+            lineno = text[:m.start()].count("\n") + 1
+            pfind(path, lineno,
+                  'include "{}" is {}, use <{}>'.format(name, why, os.path.basename(name)))
+
+
+def main():
+    try:
+        opts, args = getopt.getopt(sys.argv[1:], "eh", ["error", "enforce", "help"])
+    except getopt.GetoptError as err:
+        print(err, file=sys.stderr)
+        usage()
+
+    global error_on_warning
+    for o, unused_a in opts:
+        if o in ("-e", "--error", "--enforce"):
+            error_on_warning = True
+        elif o in ("-h", "--help"):
+            usage()
+
+    if os.environ.get("INCLUDE_STYLE_CHECK_ENFORCE"):
+        error_on_warning = True
+
+    # From here on we collect findings with pfind(). They get flushed when we
+    # are done. The program's return value depends on whether findings are
+    # treated as errors or not.
+    headers = exported_headers()
+    tracked, files = tracked_files()
+    if args:
+        # Named files, as a pre-commit hook would pass them. Anything outside
+        # the set the whole-tree run covers has nothing to say about it.
+        named = {os.path.relpath(os.path.abspath(a), topdir) for a in args}
+        files = [f for f in files if f in named]
+    check_quoted_includes(tracked, files, headers)
+
+    return
+
+if __name__ == "__main__":
+    main()
+    sys.exit(flush_messages())  # Exit value depends on findings being errors
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -190,6 +190,8 @@ jobs:
       run: |
         set -x
         scripts/cppcheck.sh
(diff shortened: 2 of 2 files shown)


grandixximo:
Since I cleaned the tree we can enforce it and error the CI when mistakes are done, objections? Anyway Bertho won't let them pass ;-)

src/emc/nml_intf/emcpos.h
```diff
@@ -14,7 +14,7 @@
 #ifndef __LINUXCNC_EMCPOS_H
 #define __LINUXCNC_EMCPOS_H
 
-#include "posemath.h"		/* PmCartesian */
+#include <posemath.h>		/* PmCartesian */
```
BsAtHome: This one is special. The posemath problem is more intricate. (also see other coment)

This header is actually including a header that is also exported. When emcpos.h ins included, it must first look look locally (where this file came from). Therefore, using "" is actually correct.

There are multiple files to describe posemath includes, which should be reduced to one. However, that is probably a breaking change and therefore not (yet) done.

src/emc/kinematics/kinematics.h
```diff
@@ -15,8 +15,8 @@
 #ifndef __LINUXCNC_KINEMATICS_H
 #define __LINUXCNC_KINEMATICS_H
 
-#include "emcpos.h" /* EmcPose */
-#include "rtapi_bool.h"
+#include <emcpos.h> /* EmcPose */
+#include <rtapi_bool.h>
```
BsAtHome, line 20: This is probably better to have the "" style. When you include `<kinematics.h>`, then that file must look locally for both emcpos.h and rtapi_bool.h.

Having them use <> makes it just search the include path and never look locally. Using "" always looks first where the file that does the including came from.

src/hal/drivers/mesa-hostmot2/hostmot2-serial.h
```diff
@@ -20,7 +20,7 @@
 #ifndef __LINUXCNC_HOSTMOT2_UART_H
 #define __LINUXCNC_HOSTMOT2_UART_H
 
-#include "rtapi.h"
+#include <rtapi.h>
```
BsAtHome: Same problem here.

The _real_ question is why this needs to be exported.

src/hal/hal.h
```diff
@@ -123,7 +123,7 @@
 
 */
 
-#include "rtapi.h"
+#include <rtapi.h>
 RTAPI_BEGIN_DECLS
```
BsAtHome: Same comment as before

src/emc/kinematics/kinematics.h
grandixximo: Understood, and that is a better rule than the one I wrote. An exported header travels to include/ and has to keep finding its siblings there, so it holds the quoted form regardless of directory. Only users take angle brackets.

Reverted in every exported header: kinematics.h, hal.h, emcpos.h, gomath.h, hostmot2-serial.h. What stays is the users: the ini files, matrixkins.comp, motion-logger.c, mathtest.c and hal.hh. The check now skips any includer that is itself exported.

On the posemath headers, and on why hostmot2-serial.h is exported at all: happy to take either on, separately. For posemath, do you mean folding EmcPose and PmCartesian into one header, or the C and C++ split inside posemath itself?

BsAtHome: > On the posemath headers, and on why hostmot2-serial.h is exported at all: happy to take either on, separately. For posemath, do you mean folding EmcPose and PmCartesian into one header, or the C and C++ split inside posemath itself?

Posemath has been a mess for some time. This code has been "misplaced" in the libnml directory. It would be nice to have a proper suggestion for this. Have not looked at what is "the best way" yet. Just looked enough to become appalled.

The hostmot2-serial header is there because of the hal/driver/mesa-hostmot/modbus/ code (and a test component mesa-pktgyro-test, which can be altered easily). It needs to be available as long as that old modbus code is not retired. The new hm2_modbus driver was written to replace the old driver. I think we need to ask on the dev and user mailing lists about deprecation and removal of the old driver.

grandixximo: Looked at posemath. Proposed shape, five commits, each standing alone, none touching `libposemath.so.0` symbols or any struct layout:

- **A.** Drop the dead knobs in `posemath.h`: `PM_LOOSE_NAMESPACE`, the `USE_CONST`/`USE_CCONST`/`USE_REF` indirection, `INCLUDE_POSEMATH_COPY_CONSTRUCTORS`, the `#if 0` blocks. About 200 lines, no behaviour change, makes the rest reviewable.

- **B.** `git mv src/libnml/posemath src/libposemath`. posemath has zero includes and uses nothing from NML; the only edge is `cms_pm.cc` going the other way. Pure move plus one Submakefile path.

- **C.** Split by language behind the existing umbrella:

  | header | contents |
  | --- | --- |
  | `posemath_types.h` | POD types and constants |
  | `posemath_c.h` | C API, includes `_types` |
  | `posemath.hh` | C++ classes and operators, includes `_c.h` |
  | `posemath.h` | umbrella, unchanged: `_c.h` always, `.hh` under `__cplusplus` |

  `posemath.h` preprocesses identically to today, so nothing out of tree breaks. In-tree includers migrate afterwards, one at a time. RT and tp stop parsing the class zoo.

- **D.** Fold `emcpos.h` and `emcpose.h` into one header next to posemath, with `emcpos.h` left as a shim. `EmcPose` is a motion type, not an NML message.

- **E.** Stop exporting `gomath.h`, `gotypes.h` and `sincos.h`. gomath has three users, all kins; `sincos.h` is a one-function shim. With C that is the installed surface down to one umbrella plus two opt-in headers in one directory.

Porting genserkins and scorbot off gomath and deleting it needs the N-DOF Jacobian and matrix code moved into posemath first, so I would track that separately. Cheaper after E anyway, since the symbols are private by then.

Does the direction look right? I would start with A and B.

On `hostmot2-serial.h`: I wrote to the dev and user lists about deprecating the old modbus driver.

BsAtHome: That sounds like a plan :-)

BsAtHome:
BTW, I don't think we should enforce this in CI. At least not yet. I'd like to see more cleanup before we start enforcing. The cleanup may expose situations that require a second thought.

grandixximo:
Fair, I'll keep them as warning for now then.

scripts/include-style-check.py
```diff
@@ -0,0 +1,184 @@
...
+    "src/emc/ini/inifile.hh": (
+        "src/emc/ini/inifile.cc",
+    ),
+    "src/hal/hal.h": (
+        "src/hal/hal_lib.c",
+        "src/hal/hal_lib_extra.c",
+        "src/hal/hal_lib_query.c",
+    ),
+}
+
```
BsAtHome: The cases here are a requirement that they use the local version. Building the library code is local.

The problem with this is that there are many exceptions and marginal calls with the current organization.

Try emc or motion. The attempt to extract some of the constants was only partly done. All of that needs a cleanup too.

BsAtHome:
> Fair, I'll keep them as warning for now then.

That I need to rephrase. I don't think it should be checked in CI yet, at all. (see my added review comment)

The code organization is still messy. That is something that needs fixing first. We need to have (all) the libraries in their own isolated directory and analyze the cross-dependencies.

grandixximo:
Ok, I see, I'll get on the cleaning first then, but I'm taking a break tomorrow, traveling with family, I'll be around a bit less...

grandixximo:
@BsAtHome The cross-dependency analysis you asked for. `scripts/include-dep-report.py` does it now, and there were three cycles.

`emc/nml_intf` and `emc/tooldata` was real and is cut in #4445. `EMC_TOOL_STAT::operator=` has no caller, so the store lookup and the include both go.

The five-way one inside libnml was not a dependency. Those six subdirectories have empty `Submakefile`s, every source is listed in `libnml/Submakefile` and links into one `libnml.so`, so libnml already is one library in its own directory and the report was drawing edges across a layout. #4446 makes it bucket by what the build links.

`emc/motion` and `emc/tp` is real, and I do not think an include rule should be what cuts it. `motion.h` embeds `TP_STRUCT coord_tp` by value, so a replacement tpmod cannot change that layout, and tp reads motmod's status struct going the other way. I deleted each include and rebuilt to see which carried weight: `tc.c` and `sp_scurve.c` need `emcmot_status_t` for `planner_type`, `jerk` and `scurve_peak_scale`, `tp.c` needs `emcmot_hal_data_t`, `blendmath.c` needs `GET_TRAJ_PLANNER_TYPE()` and the `emcmotStatus` global out of motion's private header. Only `tp.c`'s `axis.h` was dead, and #4438 already removes it. Giving tp an interface is a design change, not include hygiene.

So one cycle left and it is one I would leave alone. Does that settle the gate, or is there something else you want first? This is rebased on master.

BsAtHome:
The fact that we treat some files differently is still a problem. That reminds me, have you checked `rtapi/`'s code? That is also building in a rather interesting way. The style check first becomes an issue when we have a tree that can support it without exceptions. Exceptions are a maintenance nightmare.

I still think we should split this PR. One for the actual include type changes that are warranted and then the style check we need to work on a bit more (IMO).

grandixximo:
@BsAtHome Split done. #4448 carries the include changes, this one is the check alone. CI here fails until that merges, since the eight lines it fixes are exactly what the check flags.

On rtapi. It is 25 exported headers, 4 private ones and 13 sources in one directory, and which sources build depends on `BUILD_SYS`, RTAI and xenomai, so "the implementation of `rtapi.h`" is a configuration-dependent set and there is no table entry to write for it. Its own files already include their exported headers both ways, `"rtapi.h"` fifteen times and `<rtapi.h>` four.

So rather than enumerate exceptions I would rather remove the need for them: give exported headers a directory that holds nothing else, so "sits beside an exported header" and "is an exported header" become the same statement. That is your isolated-directory point applied to the header side, and the check then needs no exceptions at all. rtapi is both the worst case and the best proof.

Two things I ran into while looking. The export set has two sources of truth: `SRCHEADERS` at `src/Makefile:401` lists 37 headers but only builds the `HEADERS` target list, while the `cp` recipes live in five per-directory `*INCS` lists plus three one-off rules. And `headersclean` removes `$(HEADERS)`, derived from `SRCHEADERS`, so a header dropped from `SRCHEADERS` is never removed from `include/`; `sincos.h` is still sitting in mine after #4443. A stale export copy is exactly what an angled include resolves to.

Worth pursuing in that order?

BsAtHome:
The multiple copy recipes are the actual problem, I think. They should not copy locally, but add to one variable and one rule does the copying.
The complicating issue is that the copying of the headers is one of the very first thing that needs to be done so all dependent rules will rebuild properly (the headers rule).

When we remove a header from being exported, then that is always a problem. You need a `make clean` and in that process remove everything from the include/ directory.

The only alternative is to move everything that is exportable to the include directory permanently (no more copying). But I fear that it also would make things difficult at this state. We need to have all the proper interfaces in the includes and nothing else. That will take some more cleanup before we get to that point.

grandixximo:
Gets back to scrubbing...

grandixximo:
@BsAtHome Rebased, and the exception table is gone. With #4448 and #4449 in, the check passes on master with `INTERFACES` empty, so I removed it rather than carry two entries that no longer fire. The rule is now one sentence: an exported header and anything in its directory keep the quoted form, everyone else takes the angled one.

One cost, stated plainly. Without the table a file beside an exported header may quote it, so the five `emc/ini` users #4448 just fixed are no longer held to the angled form; only far-directory users are. That gap closes on its own once exported headers live in a directory of their own, which is where you were heading anyway.

Also fixed a crash: the file list came from `git ls-files` and was then opened blind, so a tracked path missing from the working tree, deleted but uncommitted or mid-rebase, took the script down instead of being skipped. CI never sees that, a pre-commit hook does.

#4452 has the one-rule header export you asked for.

grandixximo:
If we don't merge, need keep checking these by hand, the exclusions are gone, any more blockers?

scripts/include-style-check.py
```diff
@@ -0,0 +1,169 @@
...
+    for path in files:
+        if path in exported:
+            # An exported header is copied to include/ and has to keep finding
+            # its siblings there, so it includes them with quotes.
+            continue
+        directory = os.path.dirname(path)
+        with open(os.path.join(topdir, path), encoding="utf-8", errors="replace") as f:
+            text = f.read()
+        for m in RE_QUOTED.finditer(text):
+            name = m.group(1)
```
BsAtHome: How does this track includes in a component that is not using `#include` but uses `include "bla";`?

scripts/include-style-check.py
```diff
@@ -0,0 +1,169 @@
...
+    return tracked, sorted(f for f in tracked if f.endswith(SUFFIXES))
+
+
+def check_quoted_includes(tracked, files, headers):
+    exported = set(headers.values())
+    for path in files:
+        if path in exported:
+            # An exported header is copied to include/ and has to keep finding
+            # its siblings there, so it includes them with quotes.
+            continue
```
BsAtHome, line 117: This is an assumption that may not be true. At least, it is not enforced.

scripts/include-style-check.py
```diff
@@ -0,0 +1,169 @@
...
+            source = headers.get(os.path.basename(name))
+            if source is None:
+                continue    # not an exported header, nothing to say about it
+            # A file of that name beside the includer is a different header
+            # that happens to share the basename, not this one.
+            local = os.path.normpath(os.path.join(directory, name))
+            if local != source and local in tracked:
+                continue
+            if directory == os.path.dirname(source):
+                continue        # beside the header, taking the source copy
```
BsAtHome, line 137: This assumes that the directory is determining whether a quoted include is fine. That is also an assumption.

BsAtHome:
> If we don't merge, need keep checking these by hand, the exclusions are gone, any more blockers?

It is not about will/wont merge. It is about me having reservations that the code base is ready for this (besides my review comments).

scripts/include-style-check.py
grandixximo: It did not. Fixed: the check now reads `include "bla";` above the `;;` line as well, where halcompile takes its own include directive rather than the preprocessor's, and `.icomp` is in the suffix list too. Both forms in a component are flagged now, and the tree is still clean.

scripts/include-style-check.py
grandixximo: Correct, nothing enforces it. It is an assumption about layout rather than about includes, and #4461 removes the layout: the 37 exported headers move into `src/include`, which holds nothing else. A sibling include then has to name a file in that same directory, so the check can verify it instead of assuming it.

scripts/include-style-check.py
grandixximo: Also correct. That branch exists only because sources sit beside exported headers. After #4461 none do, so it has no instances left and goes, and what is left is one rule with no exceptions.

PR #4376, opened 2026-08-13
hm2_modbus: Don't complain about relative paths, it is up to the user

BsAtHome:
This PR removes the absolute path check in the hm2_modbus driver.

The default is to cd into the directory where the INI-file lives. And, if your mbccb file(s) live in there, then there is no need to complain. Therefore, the path is now fully up to the user to make it work.


grandixximo:
Beat me to it I guess, should have asked, your code, sorry....

PR #4380, opened 2026-08-14
backport: hm2_modbus: Fix doc about absolute path

BsAtHome:
This is a partial backport of #4376.
The absolute path test was already disabled in the 2.9 branch but was never propagated to the master branch (now it has, with #4376). The docs in the 2.9 branch were never updated to include the change. This PR updates the 2.9 branch to fix the doc issue and removes the absolute path test that was commented out. There are no functional or behavioural changes.


grandixximo:
Docs should match behavior, weird how 2.10 and 2.9 differed, much better now, give the user the power, that's the Linux way 🐧

PR #4382, opened 2026-08-15
hal: Add the HAL query API and add IntEnum types

BsAtHome:
This PR adds the complete HAL query API to halmodule as a sub-module `hal.query`. It also adds IntEnum classes under the `hal` class so that pin/param/signal types now can be `hal.Type.BOOL`, `hal.Type.REAL`,... The same for direction `hal.Dir.IN`, `hal.Dir.RW`, etc. The hal.query methods use these in the returned dictionaries too.
The `HAL_` prefixed names are also in the enums, but the old `[HAL_]FLOAT`, `[HAL_]BIT`, `[HAL_]S64` and `[HAL_]S64` names were removed from the enumerated type. In the interim, there still are S32/U32 names registered for compatibility. Using the HAL_ prefix is just extra baggage that is not necessary. You should use the plain names without prefix. Note that all `hal.HAL_*` constants are not changed and still available for compatibility reasons.

The hal.query is a sub-module written using pybind11. The same will also happen to halmodule when we get to that point (@rene-dev will be happy). The pybind11 library is now a build dependency and added as such. A configure check has been added.

The `hal.get_realtime_type()` method has been updated to return the new `hal.RTType` IntEnum. The old constants are still available, but it is better to move to enums because they are much more informative than a bare number and IntEnum is fully compatible with numeric values.

The documentation has been updated and the new APIs added to the adoc. Some formatting has also been done.

There are now, in the docs, notices about deprecation of the old-style constants and the IntEnums should be preferred. Also, `get_info_pins`, `get_info_params` and `get_info_signals` are marked as deprecated in the doc. These are replaced by the query API.

@petterreinholdtsen When this is merged, then #4231 can be updated and use the generic query API.

### DIFF
--- a/debian/control.top.in
+++ b/debian/control.top.in
@@ -40,6 +40,7 @@ Build-Depends:
     psmisc,
     python3,
     python3-dev,
+    python3-pybind11,
     python3-tk,
     python3-xlib,
     tcl,
--- a/docs/src/config/python-hal-interface.adoc
+++ b/docs/src/config/python-hal-interface.adoc
@@ -10,8 +10,10 @@ a Python API for creating and accessing HAL pins and signals.
 [IMPORTANT]
 ====
 The classes inside the Python module `hal` are layered on top of the module implementation called `_hal`.
-You should only access the `hal` API as described in this document.
-The inherited methods, members and properties from `_hal` may change without notice.
+Many features described in this document are __**only**__ available in the `hal` Python module.
+
+You should only import and access the `hal` module API as described in this document.
+The base-class methods, members, properties and values may change without notice.
 ====
 
 == Basic usage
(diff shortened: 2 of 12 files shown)


andypugh:
You may have already answered this, but I have not been keeping track. Do we have a way to tell if a pin is connected now from inside the component that created it?

KentChenY:
placeholder

BsAtHome:
> You may have already answered this, but I have not been keeping track. Do we have a way to tell if a pin is connected now from inside the component that created it?

Yes, but not in RT while running the RT function(s).

In userspace you can simply call `hal_getref_p()` on your pin and check the pp.signal field (C/C++). If it is non-null, then it is connected to the named signal.
```C
hal_query_t q = {};
q.name = "my-pin-name";
bool connected = 0 == hal_getref_p(&q) && NULL != q.pp.signal;
```
In halmodule you can do `hal.query.pin("my-pin-name")` and check the `'signal'` entry in the dictionary, which is a string when connected or None when not.
```python
p = hal.query.pin("my-pin-name")
connected = None is not p and p['signal'] is not None
```

In RT you should never ever call the query API. You shouldn't before and you shouldn't now because it uses the HAL mutex. RT components should not care whether a pin is connected or not. But, there is a (very much discouraged and hack) method to find out in RT that does not involve the query API. However, you cannot get the name of the signal in RT without going through the query API.

PR #4384, opened 2026-08-16
  sai: give rs274 its own tool mmap instead of truncating $HOME/.tool.mmap

tzuohann:
at driver.cc:570, before `getopt()` at :578; `-t` is not read until :583.

  That file is not scratch space. `tool_mmap_fname()` builds it from
  `secure_getenv("HOME")` with a fixed name, and io, milltask, halui and the
  Python bindings all map the same inode `MAP_SHARED`. So an offline parse run
  beside a live session replaces that session's tool table with the compiled-in
  sample table. `O_TRUNC` preserves the inode, so nothing re-maps and nothing
  errors — the running session simply observes its tools change.

  Observed on 2.9.8 with `[EMCIO]DB_PROGRAM`: 15 tools became the 4 sample
  entries mid-session, `G43` applied 0.0000 for a tool no longer in the table,
  and the tool-number/drawbar guard inhibited jog and feed. `DB_PROGRAM` neither
  prevents nor repairs it — ioControl.cc creates the mmap before the `DB_ACTIVE`
  branch, and io does not re-read afterwards. The code is unchanged at v2.9.10.

  This gives sai a private, pid-qualified file and unlinks it on exit.
  `tool_mmap_close()` already unlinks, and `tool_mmap_fname()` already honours a
  preset filename — the patch only adds the setter needed to reach it. io and
  milltask are untouched and remain the only creators of the shared file.

  Reproduce before the change:

      rm -rf /tmp/rsx && mkdir -p /tmp/rsx
      HOME=/tmp/rsx rs274 -g /dev/null
      # /tmp/rsx/.tool.mmap now exists, last_index=4, holding
      # T1 z0.511 d0.125 / T2 z0.100 d0.0625 / T3 z1.273 d0.201 / T99999 P123

  After: rs274 writes `/tmp/rs274.<pid>.tool.mmap` and removes it on exit;
  `$HOME/.tool.mmap` is never opened.

  Branched from 2.9; happy to forward-port to master.

### DIFF
--- a/.github/scripts/install-deps.sh
+++ b/.github/scripts/install-deps.sh
@@ -3,6 +3,8 @@
 set -eu #Needed so CI fails when anything is wrong
 set -x
 
+.github/scripts/use-main-ubuntu-mirror.sh
+
 sudo apt-get --quiet update
 sudo apt-get install --yes --no-install-recommends devscripts equivs build-essential lintian clang
 debian/configure
--- a/.github/scripts/use-main-ubuntu-mirror.sh
+++ b/.github/scripts/use-main-ubuntu-mirror.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+# The azure.archive.ubuntu.com apt mirror on GitHub-hosted runners
+# regularly degrades to trickle speeds for extended periods, which stalls
+# package downloads until the job hits its timeout. apt retries do not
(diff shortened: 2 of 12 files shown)


grandixximo:
You are pointing this at master already...

PR #4385, opened 2026-08-16
sai: give rs274 its own tool mmap instead of truncating $HOME/.tool.mmap

tzuohann:
Running `rs274` while a LinuxCNC session is up replaces that session's tool
table with the sample one (shipped with lcnc), until restart.

A lucky user will find that conflicting MDI or gcode stops running. A less
lucky user will find code running with offsets from sample tools - potentially
disastrous.

`tool_mmap_creator()` opens `$HOME/.tool.mmap` `O_RDWR|O_CREAT|O_TRUNC`
(tooldata_mmap.cc:33, :135) and sai calls it at driver.cc:570 — before
`getopt()` at :578, so it happens on every invocation, `--help` included, and
`-t` (read at :583) cannot prevent it.

io, milltask, halui and the Python bindings all map that same inode
`MAP_SHARED`. `O_TRUNC` keeps the inode, so nothing re-maps and nothing
errors — the running session simply starts using the sample table.

Reproduce:

    rm -rf /tmp/rsx && mkdir -p /tmp/rsx
    HOME=/tmp/rsx rs274 -g /dev/null
    # /tmp/rsx/.tool.mmap, last_index=4, sample table

Patch:

    Gives sai its own pid-qualified file under /tmp, unlinked on exit, so an
    offline check can never touch a running session's table.
    tool_mmap_close() already unlinks and tool_mmap_fname() already honours a
    preset name; this only adds the setter to reach it. io and milltask are
    unchanged.

Seen on 2.9.8, identical at 2.9.10.

### DIFF
--- a/src/emc/sai/driver.cc
+++ b/src/emc/sai/driver.cc
@@ -567,7 +567,31 @@ int main (int argc, char ** argv)
   tool_nml_register((CANON_TOOL_TABLE*)& _sai._tools);
 #else //}{
   const int random_toolchanger = 0;
+  // sai gets its OWN mmap. tool_mmap_creator() opens the file O_TRUNC, and it
+  // runs before getopt() below, so every rs274 invocation -- including --help,
+  // and including one given -t -- emptied $HOME/.tool.mmap. That file is the
+  // live tool table, shared MAP_SHARED with io/milltask/halui, so an offline
+  // parse silently replaced the tool table of a running machine.
+  //
+  // mkstemp(), not a name built from the pid: TMPDIR is world-writable and a
+  // pid is guessable, so a predictable name can be pre-created as a symlink
+  // and the victim's rs274 then truncates the attacker's chosen file.
+  // mkstemp() creates it atomically with O_EXCL and mode 0600.
+  char sai_mmap_fname[LINELEN];
+  const char *tmpdir = getenv("TMPDIR");
+  if (!tmpdir || !*tmpdir) tmpdir = "/tmp";
+  snprintf(sai_mmap_fname,sizeof(sai_mmap_fname),
+           "%s/rs274.tool.mmap.XXXXXX",tmpdir);
+  int sai_fd = mkstemp(sai_mmap_fname);
+  if (sai_fd < 0) {
+      perror("rs274: mkstemp for the tool mmap failed");
+      exit(EXIT_FAILURE);
+  }
+  close(sai_fd);   // tool_mmap_creator() opens it by name; O_NOFOLLOW guards
+                   // the gap, and the file already exists and is ours
+  tool_mmap_set_fname(sai_mmap_fname);
   tool_mmap_creator((EMC_TOOL_STAT*)NULL,random_toolchanger);
+  atexit(tool_mmap_close);  // tool_mmap_close() unlinks the file
   /* Notes:
   **   1) sai does not use toolInSpindle,pocketPrepped
   **   2) sai does not distinguish changer type
--- a/src/emc/tooldata/tooldata_mmap.cc
+++ b/src/emc/tooldata/tooldata_mmap.cc
@@ -197,7 +197,7 @@ int tool_mmap_user()
         ** So print message and return fail indicator.
         */
         fprintf(stderr,"tool_mmap_user(): tool mmap not available\n");
-        tool_mmap_base = (char*)0;
+        tool_mmap_base = nullptr;
         return(-1);
     }
     tool_mmap_base = (char*)mmap(0, TOOL_MMAP_SIZE, PROT_READ|PROT_WRITE,
@@ -238,7 +238,7 @@ void tool_mmap_close()
     }
     // Idempotent from here: a second call, or an atexit after an explicit
     // close, returns at the tool_mmap_base guard above.
-    tool_mmap_base = (char*)0;
+    tool_mmap_base = nullptr;
     mmap_is_creator = false;
 } //tool_mmap_close()
 
--- a/src/emc/tooldata/tooldata_mmap.cc
+++ b/src/emc/tooldata/tooldata_mmap.cc
@@ -30,10 +30,19 @@
 
 #define TOOL_MMAP_FILENAME ".tool.mmap"
 #define TOOL_MMAP_MODE     0600
-#define TOOL_MMAP_CREATOR_OPEN_FLAGS  O_RDWR | O_CREAT | O_TRUNC
+// O_NOFOLLOW: refuse to follow a symlink at the mmap path. Without it a
+// pre-placed symlink is followed and the target truncated.
+#define TOOL_MMAP_CREATOR_OPEN_FLAGS  O_RDWR | O_CREAT | O_TRUNC | O_NOFOLLOW
 #define TOOL_MMAP_USER_OPEN_FLAGS     O_RDWR
 
 static int           creator_fd;
+// The descriptor actually backing the mapping, and whether THIS process
+// created the file. tool_mmap_user() previously dropped its fd on the
+// floor: it was neither closed nor recorded, so it leaked for the life of
+// the process and tool_mmap_close() had nothing to close but creator_fd --
+// which a user process never sets, so it is 0, i.e. stdin.
+static int           mmap_fd = -1;
+static bool          mmap_is_creator = false;
 static char          filename[LINELEN] = {};
 static char*         tool_mmap_base = 0;
 static EMC_TOOL_STAT const *toolstat;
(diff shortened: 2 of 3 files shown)


src/emc/sai/driver.cc
```diff
@@ -567,7 +567,17 @@ int main (int argc, char ** argv)
...
+  // and including one given -t -- emptied $HOME/.tool.mmap. That file is the
+  // live tool table, shared MAP_SHARED with io/milltask/halui, so an offline
+  // parse silently replaced the tool table of a running machine.
+  char sai_mmap_fname[LINELEN];
+  snprintf(sai_mmap_fname,sizeof(sai_mmap_fname),
+           "/tmp/rs274.%d.tool.mmap",(int)getpid());
+  tool_mmap_set_fname(sai_mmap_fname);
   tool_mmap_creator((EMC_TOOL_STAT*)NULL,random_toolchanger);
+  atexit(tool_mmap_close);  // tool_mmap_close() unlinks the file
   /* Notes:
```
BsAtHome, line 595: This potentially calls `exit()` while running the atexit functions. This is undefined behaviour. Changing this to _exit would terminate the atexit handling, so that is not an option. You need to make the tool_mmap_close function safe for use in atexit.

src/emc/sai/driver.cc
```diff
@@ -567,7 +567,17 @@ int main (int argc, char ** argv)
   tool_nml_register((CANON_TOOL_TABLE*)& _sai._tools);
 #else //}{
   const int random_toolchanger = 0;
+  // sai gets its OWN mmap. tool_mmap_creator() opens the file O_TRUNC, and it
+  // runs before getopt() below, so every rs274 invocation -- including --help,
+  // and including one given -t -- emptied $HOME/.tool.mmap. That file is the
+  // live tool table, shared MAP_SHARED with io/milltask/halui, so an offline
+  // parse silently replaced the tool table of a running machine.
+  char sai_mmap_fname[LINELEN];
```
grandixximo, line 580: One concern with the new filename: `/tmp` is world-writable and this name is predictable (pid only). `tool_mmap_creator()` opens it `O_RDWR|O_CREAT|O_TRUNC` without `O_EXCL` or `O_NOFOLLOW` (tooldata_mmap.cc:135), so another local user can pre-create `/tmp/rs274.<pid>.tool.mmap` as a symlink and have the victim's rs274 truncate an attacker-chosen file. PIDs are guessable, and an attacker can pre-create symlinks covering a whole pid range.

Suggest using `mkstemp()` (or adding `O_NOFOLLOW`, and/or including the uid in the name). Not a blocker for the fix itself, but since this PR touches exactly this path it would be good to close the hole at the same time.

src/emc/sai/driver.cc
tzuohann: fixed.

src/emc/sai/driver.cc
tzuohann: fixed

src/emc/tooldata/tooldata_mmap.cc
```diff
@@ -201,9 +210,12 @@ void tool_mmap_close()
...
+        // and calling exit() from within one is undefined behaviour.
+        // Report and continue: the remaining cleanup still runs.
+        perror("tool_mmap_close(): munmap fail");
         close(creator_fd);
-        perror("tool_mmap_close(): munmapfail");
-        exit(EXIT_FAILURE);
+        return;
     }
     if( unlink(tool_mmap_fname() )) {
         perror("tool_mmap_close(): unlink fail");
```
BsAtHome, line 233: Failing the munmap call drops calling the next.
The following unlink is important because you'd otherwise leave files in the filesystem. This was also wrong previously, but there is no reason to let the problem persist when you need to make changes anyway.
You want to drop the return.

Second problem: Where is `close(creator_fd)` handled when munmap succeeds? Are there code paths that leak a descriptor?

tzuohann: fixed

src/emc/tooldata/tooldata_mmap.cc
```diff
@@ -201,14 +220,26 @@ void tool_mmap_close()
...
-    close(creator_fd);
+    if (mmap_fd >= 0) {
+        close(mmap_fd);
+        mmap_fd = -1;
+    }
+    // Idempotent from here: a second call, or an atexit after an explicit
+    // close, returns at the tool_mmap_base guard above.
+    tool_mmap_base = (char*)0;
+    mmap_is_creator = false;
 } //tool_mmap_close()
```
BsAtHome, line 243: I should have checked this better when I made my other comment. But, now you also removed the close at the bottom of the function. That means you now will leak the descriptor.

The original code would leak a file on the filesystem, now it is the file descriptor. It can be fixed much easier by making it look like:
```C
void tool_mmap_close()
{
    if (!tool_mmap_base) { return; }
    // flush mmapped file to filesystem
    if (msync(tool_mmap_base, TOOL_MMAP_SIZE, MS_SYNC) == -1) {
        perror("tool_mmap_close(): msync fail");
    }
    if (munmap(tool_mmap_base, TOOL_MMAP_SIZE) < 0) {
        perror("tool_mmap_close(): munmapfail");
    }
    if( unlink(tool_mmap_fname() )) {
        perror("tool_mmap_close(): unlink fail");
    }
    close(creator_fd);
} //tool_mmap_close()

```

tzuohann: The close is still there, just not on creator_fd:

  if (mmap_fd >= 0) {
      close(mmap_fd);
      mmap_fd = -1;
  }

mmap_fd is set by both the creator and user paths after a successful mmap; both close exactly one descriptor. creator_fd is set by  creator; user process leaves it 0, so close(creator_fd) there closes stdin. Same reason unlink is guarded by mmap_is_creator.

Full disclosure if not obvious, I have no training with software architecture. Have a background in scientific programming, but that's completely different stuff. So, if I don't understand basic things, just take it as a report instead of a pull request. That may make things go faster. I use claude for all this stuff, but I don't have enough time to properly learn software, not at 42. Just relaying what claude said after quizzing it repeatedly about making sure your comments are addressed.

BsAtHome: No problem, most are not programmers by profession here. We're glad to have people participate and contribute!
Let me have a wider look at this code. Maybe I am barking up the wrong tree. However, there are many oddities with the tool database code, so a closer look may be warranted.

tzuohann: cool lmk if this works

src/emc/tooldata/tooldata_mmap.cc
```diff
@@ -201,14 +220,26 @@ void tool_mmap_close()
...
         perror("tool_mmap_close(): unlink fail");
     }
-    close(creator_fd);
+    if (mmap_fd >= 0) {
+        close(mmap_fd);
+        mmap_fd = -1;
+    }
+    // Idempotent from here: a second call, or an atexit after an explicit
+    // close, returns at the tool_mmap_base guard above.
+    tool_mmap_base = (char*)0;
```
BsAtHome: Please use NULL or nullptr instead of manual casts of the value zero (0).

src/emc/tooldata/tooldata_mmap.cc
BsAtHome: You are right. Your changes work.

BsAtHome:
BTW, I see this is targeted at 2.9 branch. The same problem in master, I guess?

tzuohann:
> BTW, I see this is targeted at 2.9 branch. The same problem in master, I guess?

Yes. Its in master. I wasn't sure if you want to retarget this, or do a new PR. Just LMK, or DIY.

src/emc/tooldata/tooldata_mmap.cc
tzuohann: Done with nullptr at 241 and also at 200 (another one).

BsAtHome:
Yes, then you need to retarget this PR to master.

grandixximo:
The bug is identical in master, so the fix is wanted in both. The question is just direction.

I asked tzuohann to point this at 2.9 since it was reported there (2.9.8, still in 2.9.10). Our usual flow is master-first, then backport. Would the reverse work here: keep this PR on 2.9 and merge 2.9 into master to carry it up? I checked the patch against both branches; it applies cleanly to 2.9, and the master differences in these files are only cosmetic, so the merge-up should be near-trivial.

If you would rather have master-first, no problem, tzuohann can retarget and we backport.

BsAtHome:
No, by all means, go for it. If this needs to be applied in both places, then it does not really matter in which order.

tzuohann:
i take it there is nothing for me to do as far as master is concerned.

grandixximo:
Yeah, it maybe time for merging 2.9 to master, I'll see to it soon

PR #4386, opened 2026-08-16
mesa_modbus: deprecate in favor of hm2_modbus

grandixximo:
Deprecate mesa_modbus in favor of hm2_modbus

Following the discussions on the mailing list about having proper deprecation warnings in place, this marks the mesa_modbus framework (modcompile + mesa_modbus.c.tmpl) as deprecated. The replacement is the hm2_modbus driver with mesambccc-compiled MBCCB files, which already ships in master and 2.9.

Warnings added:

- mesa_modbus.c.tmpl: RTAPI_MSG_WARN at module load, so every modcompile-built driver announces itself at loadrt time
- modcompile.py: stderr warning at build time
- README.adoc in the modbus source directory
- docs: WARNING block in the mesa_modbus driver page, (DEPRECATED) in the modcompile(1) man page, notes in the component list and the docs index

No functional changes; existing configurations keep working unchanged.

Question for reviewers: is a 2.9 back-port wanted? hm2_modbus and mesambccc exist in 2.9, so users there can migrate.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/modbus/README.adoc
+++ b/src/hal/drivers/mesa-hostmot2/modbus/README.adoc
@@ -1,3 +1,7 @@
+WARNING: The mesa_modbus framework (this directory and modcompile) is
+deprecated and will be removed in a future release. Use the hm2_modbus
+driver with mesambccc-compiled .mbccb files instead.
+
 Do not edit mesa_modbus.c (Unless you are sure that you want to, of
 course).
 Create a new xxxxx.mod file that describes your particular modbus device
--- a/src/hal/drivers/mesa-hostmot2/modbus/modcompile.py
+++ b/src/hal/drivers/mesa-hostmot2/modbus/modcompile.py
@@ -88,6 +88,10 @@ def main():
     if len(args) < 1:
         raise SystemExit("No modules found (*.mod files) to compile.");
 
+    print("WARNING: modcompile and the mesa_modbus framework are deprecated and will be"
+          " removed in a future release. Use the hm2_modbus driver with mesambccc-compiled"
+          " .mbccb files instead.", file=sys.stderr)
+
     # Create a temporary directory and copy the C template into it
     tempdir = tempfile.mkdtemp(prefix="modcompile")
     BASE = os.path.abspath(os.path.join(os.path.dirname(sys.argv[0]), ".."))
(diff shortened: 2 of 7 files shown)


src/hal/drivers/mesa-hostmot2/modbus/modcompile.py
```diff
@@ -88,6 +88,10 @@ def main():
     if len(args) < 1:
         raise SystemExit("No modules found (*.mod files) to compile.");
 
+    print("WARNING: modcompile and the mesa_modbus framework are deprecated and will be"
+          " removed in a future release. Use the hm2_modbus driver with mesambccc-compiled"
+          " .mbccs files instead.", file=sys.stderr)
+
```
BsAtHome, line 94: Here you name it the `.mbccs` file (source), while other places you use the compiled binary (mbccb).

Any particular reason?

src/hal/drivers/mesa-hostmot2/modbus/README.adoc
```diff
@@ -1,3 +1,7 @@
+WARNING: The mesa_modbus framework (this directory and modcompile) is
+deprecated and will be removed in a future release. Use the hm2_modbus
+driver with mesambccc-compiled .mbccs files instead.
+
```
BsAtHome, line 4: Same comment

src/hal/drivers/mesa-hostmot2/modbus/modcompile.py
grandixximo: No reason, that was a mixup. Fixed: modcompile and the README now name the compiled .mbccb files, consistent with the other warnings.

src/hal/drivers/mesa-hostmot2/modbus/README.adoc
grandixximo: Fixed as well, README now says .mbccb.

zz912:
Great warning. Thank you. I recommend backporting to 2.9, but I would like to reassure integrators that mesa-modbus will still be available in 2.9 indefinitely. Backporting is important for integrators who want to start with Modbus on Mesa to start with hm2_modbus. However, we do not want to stress integrators who are already using mesa_modbus.

BsAtHome:
Agreed, backport is a good idea. The message in the 2.9 backport can be phrased more "friendly" to ask nicely to migrate.

PR #4387, opened 2026-08-16
posemath: drop the dead knobs and give it its own directory

grandixximo:
Following on from the review discussion in #4375, where the posemath headers came up and the answer was that the code has been misplaced in libnml for some time. This is the first two steps of the five I sketched there, and the two that stand on their own.

**A. Drop the knobs that are never turned.**

`USE_CONST`, `USE_CCONST` and `USE_REF` are defined unconditionally at the top of the header and select `PM_CONST`, `PM_CCONST` and `PM_REF`. Every declaration in the file is written through that indirection, so reading a signature means resolving three macros to find out that they mean `const`, `const` and `&`. The macros are substituted out.

`INCLUDE_POSEMATH_COPY_CONSTRUCTORS` is guarded by `#if __cplusplus < 201103L`, and the tree builds as C++20, so its 13 blocks in each file have not been compiled in a long time. `PM_LOOSE_NAMESPACE` is defined nowhere. Four `#if 0` blocks go with them.

What stays is the one deliberate poison macro for `pmCartNorm()`, which exists to make a use fail to compile rather than to be compiled out.

No behaviour change: the preprocessor already resolved every one of these the same way on every build in the tree. 338 lines out, 50 in.

**B. Move it out of libnml.**

`git mv src/libnml/posemath src/libposemath`. posemath includes nothing from NML and uses nothing from it; the only edge between them runs the other way, in `cms_pm.cc`. The move is the sources plus the paths that named the old location: the Submakefile prefixes, `SRCHEADERS` and the per-module object lists in `src/Makefile`, two `USE_TOPDIR` includes in `tpcomp.comp`, and the `debian/copyright` stanza. `libposemath` gets its own `SUBDIRS` entry rather than one buried in the libnml group. All eight source files are detected as pure renames.

**Testing.**

Configured `--with-realtime=uspace` and built clean, no errors and no warnings.

`libposemath.so.0` exports the same 390 symbols before and after, byte identical lists from `nm -D --defined-only`. That is the evidence that A changes nothing: it is a large diff that the compiler cannot tell apart from the original.

`include/` holds the same 38 headers with the same names, so the installed surface is untouched by the move.

`scripts/runtests` over blendmath, realtime-math, interp, ccomp and matrixkins: 90 run, 90 successful, 0 failed, 1 skipped, 0 shmem errors.

All nine kinematics modules that link a posemath object load under `halrun`: genser, scara, puma, genhex, penta, 5axis, three21, rotarydelta, triv.

**What comes next, separately.**

C splits the header by language behind the existing umbrella, D folds `emcpos.h` and `emcpose.h` into one header beside posemath, and E takes `gomath.h`, `gotypes.h` and `sincos.h` off the exported list. Those are written and tested but want their own review, since each changes what the tree installs.


src/hal/components/tpcomp.comp
```diff
@@ -70,8 +70,8 @@ option  extra_setup;
 #include USE_TOPDIR(src/emc/tp/spherical_arc.c)
 #include USE_TOPDIR(src/emc/tp/blendmath.c)
 #include USE_TOPDIR(src/emc/nml_intf/emcpose.c)
-#include USE_TOPDIR(src/libnml/posemath/_posemath.c)
-#include USE_TOPDIR(src/libnml/posemath/sincos.c)
+#include USE_TOPDIR(src/libposemath/_posemath.c)
+#include USE_TOPDIR(src/libposemath/sincos.c)
```
BsAtHome, line 75: This construct is an abomination...

src/libposemath/posemath.h
```diff
@@ -659,40 +574,13 @@ extern "C" {
...
-*/
-
 /* some nice constants */
 
 #define PM_PI      3.14159265358979323846
 #define PM_PI_2    1.57079632679489661923
 #define PM_PI_4    0.78539816339744830962
 #define PM_2_PI    6.28318530717958647692
 
-#ifdef PM_LOOSE_NAMESPACE
```
BsAtHome, line 675: We may actually test if these should be redefined as:
```C
#define PM_PI M_PI
...
```

src/libposemath/Submakefile
```diff
@@ -10,10 +10,10 @@ TARGETS += ../lib/libposemath.so ../lib/libposemath.so.0
...
-    ./libnml/posemath/posemath.h \
-    ./libnml/posemath/sincos.h
+    ./libposemath/gomath.h \
+    ./libposemath/gotypes.h \
+    ./libposemath/posemath.h \
+    ./libposemath/sincos.h
 
-$(patsubst ./libnml/posemath/%,../include/%,$(POSEMATHINCS)): ../include/%.h: ./libnml/posemath/%.h
+$(patsubst ./libposemath/%,../include/%,$(POSEMATHINCS)): ../include/%.h: ./libposemath/%.h
 	cp $^ $@
```
BsAtHome, line 19: So, we're not cleaning up the code, yet. For a follow up?

src/hal/components/tpcomp.comp
grandixximo: It is, and I only retargeted the two paths inside it. Reworking how the skeleton gets its sources is a halcompile question rather than a posemath one, so I would rather not fold it in here.

Worth noting that tpcomp.comp is filtered out of COMPS in hal/components/Submakefile, so nothing in tree compiles it and a stale USE_TOPDIR path is invisible to the build and to CI. Setting TOPDIR and running halcompile on a copy, as the man page describes, does catch it. If that is worth having permanently it is a small test, and it would keep the paths honest whatever happens to the construct.

src/libposemath/posemath.h
grandixximo: Tested:

```
  PM_PI    == M_PI     bit identical
  PM_PI_2  == M_PI_2   bit identical
  PM_PI_4  == M_PI_4   bit identical
  PM_2_PI  != M_2_PI   6.2831853071795862 vs 0.63661977236758138
  PM_2_PI  == 2*M_PI   bit identical
```

Three map straight across. The fourth is a trap: M_2_PI in libm is 2/pi, not 2*pi, so the obvious substitution is wrong by a factor of about 9.87 everywhere PM_2_PI is used; it wants 2*M_PI written out.

Since the values are identical today nothing is waiting on this. I would do it with the rest of the constants once the split moves them into their own header.

src/libposemath/Submakefile
grandixximo: For a follow up, and that one is written. C splits posemath.h into posemath_types.h, posemath_c.h and posemath.hh with posemath.h still the umbrella, D folds emcpos.h and emcpose.h into one header beside posemath and moves emcpose.c with it, E takes these three off the exported list. Their five in-tree users, all kinematics, include them by path instead.

Kept off this PR because each changes what the tree installs. One PR or three, whichever you would rather review.

src/hal/components/tpcomp.comp
BsAtHome: Agreed. It just stood out as "oh boy, this is awful" ;-)

src/libposemath/Submakefile
BsAtHome: Check, proceed :-)

src/libposemath/posemath.h
BsAtHome: The "correct" values are in the system/libc defines. Using our own values is an ancient compatibility thing.

However, the RTAI build uses a kernel supplied header and that may not have these defines. In rtapi/rtapi_math.h we can also see the value of pi being defined as the long double value (digits get truncated). Maybe we need to check the other values in rtapi_math.h too and define if not present.

The PM_2_PI vs. M_2_PI is interesting. At least that value has a real reason to be here.

grandixximo: The kernel branch includes no math.h and fills in only M_PI, M_PIl and M_PI_2l, so M_PI_2, M_PI_4 and the other ten are absent there, and the truncated digits turn out to be cosmetic since every fallback is already glibc's value.

Sent as #4395, which defines the missing ones under the same guard.

grandixximo:
Need me to rebase here? I have the follow up ready, do we need to run this trough weekly meeting?

src/libnml/posemath/posemath.h
```diff
@@ -659,40 +574,13 @@ extern "C" {
...
-    typedef PmRpy RPY;
-    typedef PmPose POSE;
-    typedef PmHomogeneous HX;
-    typedef PmCircle CIRCLE;
-    typedef PmLine LINE;
-
-#define PI                PM_PI
-#define PI_2              PM_PI_2
-#define PI_4              PM_PI_4
-#define TWO_PI            PM_2_PI	/* 2_PI invalid macro name */
```
BsAtHome, line 693: How does removing these influence code?

Is there a chance that is is used in out-of-tree code? If not, then the issue is moot.

BsAtHome:
Rebase would be nice to see if there is any change that altered behaviour.

src/libnml/posemath/posemath.h
grandixximo: In tree it changes nothing. PM_LOOSE_NAMESPACE appears only in posemath.h, in that comment and that #ifdef, and `git log -S` over all history shows it has never been defined anywhere since the 2006 import, so the block has never been compiled. TWO_PI and PI_4 appear nowhere else. PI_2 does appear in genserkins.h, but that is its own `#define PI_2 GO_PI_2` from gomath.

Out of tree someone would have to define the macro themselves to reach it. If anyone does, removal breaks them at compile time on an unknown type rather than silently, and the fix on their side is to declare the typedefs they want.

Rebased on current master, which now has #4395.

PR #4390, opened 2026-08-16
ini: warn when PLANNER_TYPE=1 falls back to trapezoidal due to low jerk

grandixximo:
`PLANNER_TYPE = 1` with `MAX_LINEAR_JERK < 1.0` silently forced the trapezoidal planner, both at startup (`loadTraj()`, which carried a FIXME asking for this warning) and at runtime via the inihal HAL pins. Users who requested S-curve planning got no indication their setting was ignored.

This prints a warning in both paths: at startup, and once per session for HAL-pin-triggered fallbacks.

```
[TRAJ]PLANNER_TYPE = 1 (S-curve) requires [TRAJ]MAX_LINEAR_JERK >= 1.0 (got 0.5); using trapezoidal planner
```

Note: #4349 documents the old silent behavior in `ini-config.adoc`; that sentence becomes stale if this merges first.

### DIFF
--- a/src/emc/ini/inihal.cc
+++ b/src/emc/ini/inihal.cc
@@ -243,6 +243,18 @@ static void copy_hal_data(const ptr_inihal_data &i, value_inihal_data &j)
 #undef ARRAY
 } // copy_hal_data()
 
+// S-curve planner requires max jerk >= 1.0; when a HAL pin change forces
+// the planner back to trapezoidal, warn once instead of on every change.
+static void warn_planner_fallback_once(void)
+{
+    static bool warned = false;
+    if (!warned) {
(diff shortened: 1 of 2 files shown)


grandixximo:
This is one of those points that are better addressed in code rather than in the docs, discussed at the meeting

BsAtHome (approved):

BsAtHome:
> This is one of those points that are better addressed in code rather than in the docs, discussed at the meeting

This wasn't in the docs, wasn't it? Or would the docs also need an update in the PR?

grandixximo:
This is not in the doc yet, #4349 is proposing to put in the docs, instead I'm fixing the source, merge this, and then I will direct #4349 for changes

BsAtHome:
Check

PR #4392, opened 2026-08-16
motion: remove dead commands ENABLE_WATCHDOG, DISABLE_WATCHDOG, SET_TELEOP_VECTOR

### LINKED ISSUE #4359: motion: three cmd_code_t commands have no handler and no sender (ENABLE_WATCHDOG, DISABLE_WATCHDOG, SET_TELEOP_VECTOR)
`cmd_code_t` in `src/emc/motion/motion.h` has 76 entries. The switch in `src/emc/motion/command.c` handles 73 of them. The three without a handler are:

- `EMCMOT_ENABLE_WATCHDOG`
- `EMCMOT_DISABLE_WATCHDOG`
- `EMCMOT_SET_TELEOP_VECTOR`

Outside the enum itself, the only references anywhere in the tree are three `case` labels in `src/emc/motion-logger/motion-logger.c` (lines 356, 360 and 494), which just log the command name. Nothing sends any of them, so those cases can never be reached either.

The gap has been noted in `docs/src/code/code-notes.adoc` since 2020, where the note says the watchdog commands are "maybe obsolete". Six years on, nothing has started using them.

Removing the three enum values and the three logger cases would delete the discrepancy rather than keep documenting it, and would let the note in Code Notes go away too. It is a handful of lines and touches no behaviour.

Before I put a patch together: does anyone mind seeing them go? In particular, is there an out-of-tree component or a private branch that sends `ENABLE_WATCHDOG` / `DISABLE_WATCHDOG`, or anything that relies on `SET_TELEOP_VECTOR` still having a stable enum value? If someone would rather keep them as reserved slots, that is fine too; I would then just fix the Code Notes wording so it stops carrying a count that goes stale every time a command is added.

grandixximo:
Fixes #4359.

The three commands have had no handler since 1adc33f11c (the old sound-card watchdog cleanup); SET_TELEOP_VECTOR never had one. Nothing in the tree sends any of them, so the motion-logger cases removed here could never be reached. Code Notes has carried a "73 vs 70 commands" note about this since 2020; with the commands gone there is nothing left to count, so the note and the three command sections go too.

One caveat: removing enum entries shifts the numeric values of later `cmd_code_t` commands. The enum is only exchanged between in-tree components built from the same sources (task, motion, motion-logger), so this is safe in-tree. An out-of-tree sender of these commands would already be a no-op today.

This conflicts with the code-notes.adoc hunk in #4349, which rewords the count note instead of deleting it. Whichever merges second needs the hunk dropped.

### DIFF
--- a/docs/src/code/code-notes.adoc
+++ b/docs/src/code/code-notes.adoc
@@ -241,12 +241,6 @@ More on that function later.
 There are approximately 44 commands - this list is still under
 construction.
 
-[NOTE]
-The cmd_code_t enumeration, in motion.h, contains 73 commands, but the switch
-statement in command.c contemplates only 70 commands (as of 6/5/2020).
-ENABLE_WATCHDOG / DISABLE_WATCHDOG commands are in motion-logger.c. Maybe they are obsolete.
-The SET_TELEOP_VECTOR command only appears in motion-logger.c, with no effect other than its own log.
-
 === ABORT
 
 The ABORT command simply stops all motion. It can be issued at any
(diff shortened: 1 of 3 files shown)


grandixximo:
This is the other, that would be better as a cleanup in the code rather than a note in the docs from #4349
Let's merge these first, then have #4349 cleaned up...

BsAtHome:
> This is the other, that would be better as a cleanup in the code rather than a note in the docs from #4349 Let's merge these first, then have #4349 cleaned up...

Yes, this looks like janitorial work. I guess there is nothing out-of-tree that would access any of this?

grandixximo:
GitHub and web search for all three symbols: every hit is a fork carrying its own copy of the same code (machinekit even keeps a live `SET_TELEOP_VECTOR` path, but against its own motion.h), no independent consumer exists, and nothing in our tree can send these commands since there is no NML message or taskintf entry point for them. Also `motion.h` is not shipped in the dev package, so out-of-tree code never sees `cmd_code_t` in the first place.

BsAtHome (approved):

PR #4394, opened 2026-08-16
CI: test headless (--disable-gui) build

grandixximo:
## Summary

#4364 added `--disable-gui` headless build support. If we support it, we should test it does not break. This PR adds a `rip-headless` CI job and fixes one gap the new test immediately caught.

**CI job (`rip-headless`)**
- Installs a minimal dependency set with no X11, Tk, GTK or documentation tooling, so a regression that reintroduces a GUI dependency into the headless build fails at configure or build time instead of being masked by full build-deps.
- Builds RIP with `--disable-gui --disable-manpages --disable-build-documentation`.
- Asserts GUI artifacts are absent (`axis`, `touchy`, `halshow`, `pyvcp`, `gladevcp`, `tcl/linuxcnc.so`, ...) and core headless artifacts exist (`linuxcncrsh`, `halcmd`, `halui`, ...).
- Runs the GUI-free test suites: `tests/linuxcncrsh`, `tests/linuxcncrsh-tcp`, `tests/halui`, `tests/motion`.

**Build fix**
- #4364 guarded the *install* rules for `pyvcp`/`gladevcp` behind `BUILD_GUI` but not the RIP *build* rules, so headless builds still produced `bin/pyvcp` and `bin/gladevcp`. Guarded `USER_COMP_PY` in `src/hal/user_comps/Submakefile` to match.

## Test plan

- Replicated the job end-to-end locally in an `ubuntu:24.04` container with only the minimal deps installed: configure, build, artifact assertions, and all 9 headless tests pass.
- The artifact assertion failed before the `USER_COMP_PY` fix (`bin/pyvcp` was still built) and passes after, confirming the check actually tests something.


.github/workflows/ci.yml
```diff
@@ -56,6 +56,58 @@ jobs:
...
+      run: |
+        .github/scripts/verify-clean-repo.sh
+    - name: Test
+      run: |
+        set -x
+        scripts/rip-environment runtests tests/linuxcncrsh tests/linuxcncrsh-tcp tests/halui tests/motion
+    - name: Verify no untracked or modified files after test
+      run: |
+        .github/scripts/verify-clean-repo.sh
+
```
BsAtHome, line 110: This is done twice?

.github/workflows/ci.yml
```diff
@@ -56,6 +56,58 @@ jobs:
...
+                exit 1
+            fi
+        done
+    - name: Verify no untracked or modified files after build
+      run: |
+        .github/scripts/verify-clean-repo.sh
+    - name: Test
+      run: |
+        set -x
+        scripts/rip-environment runtests tests/linuxcncrsh tests/linuxcncrsh-tcp tests/halui tests/motion
```
BsAtHome: There must be many more tests that should be run. Or are 280+ tests all requiring GUI access?

.github/workflows/ci.yml
grandixximo: Both checks follow the pattern of the existing jobs (rip-and-test, rip-and-test-clang, package-arch all verify after build and again after test). The after-build check catches build artifacts dirtying the tree, the after-test check catches test leftovers. Kept for consistency.

.github/workflows/ci.yml
grandixximo: Fair point, fixed. The job now runs the full suite (`runtests -p ./tests`). Locally in an ubuntu:24.04 container with only the minimal deps: 285 pass, 14 skipped. The skips are tests that need things a headless build intentionally does not provide (pyvcp, haltcl/tcl extensions, xvfb for ui-smoke); they got `skip` scripts like the ui-smoke tests already use. Running everything also exposed two real gaps, fixed here: preview_helpers.py was only built in GUI builds (needed by GUI-free tests/interp_initcode, now moved to lib/python), and glcanon needed python3-numpy.

BsAtHome:
You need to resolve the conflict. Also, maybe squash the commits.

grandixximo:
Done and done

Issue #4379, opened 2026-08-14
glcanon regression


c-morley:
<img width="1280" height="800" alt="Image" src="https://github.com/user-attachments/assets/6dd5b7ac-cc27-4282-b20e-480356074200" />

Weird rainbow effect on new shader work.
Pretesting a while back did not show this.

BsAtHome:
@alex-pres you are most qualified to look into this, I guess.

andypugh:
It's not right, obviously, but it is rather atmospheric.

josephklittle1-spec:
This looks like a shader precision or color-space issue. I’d check recent changes to normals, UVs, and RGB calculations, especially anything different from the earlier pretest setup.

alex-pres:
> [@alex-pres](https://github.com/alex-pres) you are most qualified to look into this, I guess.


I will check this. Looks really amazing )

Most probably that is invalid coordinate system setup before drawing.

PR #4402, opened 2026-08-18
qt5_graphics bg fix

alex-pres:
Bugfix for #4379

Invalid vertex shape used for background drawing

Also removed now unused import of glColor4f


BsAtHome:
@c-morley can you please confirm on your system?

c-morley:
Thank you!

PR #4403, opened 2026-08-18
linuxcncrsh: Allow IPv4 as fallback when IPv6 is disabled

### LINKED ISSUE #4400: linuxcncrsh permanently refuses all SET commands after a client disconnects mid-command
cc @BsAtHome @grandixximo — flagging you both: @BsAtHome by commit-history weight on this file, @grandixximo because 0d3ddc10f2 (2026-08-07) fixed a *different* race in this same SET-handling path, so the code is fresh for you.

Found 2026-07-06 while driving a sim through `linuxcncrsh` from short-lived `nc` connections. All line references below verified against master b3a596f904 (2026-08-17).

### Summary

`linuxcncrsh` serializes remote SET commands through one **global** latch,
`activeSetter` (src/emc/usr_intf/emcrsh.cc:196). The latch is set when any
remote SET is dispatched (`setterSet`, defined at emcrsh.cc:211, called at
emcrsh.cc:3114) and is cleared in exactly one place: the per-client wait loop
that detects the command's completion **for the client that issued it**
(`setterClear`, emcrsh.cc:3637).

The two loops at the end of `sockMain()` are the whole bug:

- **emcrsh.cc:3616-3622** — "Remove fully closed clients" erases any client
  with `sock == -1` (`clients.erase(i)`, emcrsh.cc:3618). This branch does
  **not** look at `cmdtimeout` or `setterActive()`, so a client that owned an
  in-flight SET is dropped silently.
- **emcrsh.cc:3625-3638** — the loop that *would* clear the latch only ever
  reaches `setterClear(*i)` for a client satisfying
  `setterActive() && i->cmdtimeout > 0.0` (emcrsh.cc:3627). The owning client
  was erased by the previous loop, so no surviving client ever satisfies it.

The result: `activeSetter` stays set forever. From then on, **every SET from
every connection is refused forever** with:

    error: Cannot issue SET <X> command while SET <Y> command is still active

GET commands still work, so the server looks healthy while being
permanently read-only. Only restarting linuxcncrsh (usually: the whole
LinuxCNC session, since linuxcncrsh is commonly the DISPLAY) recovers.

### Steps to reproduce

Any sim config with `DISPLAY = linuxcncrsh`. Machine on, a program loaded,
then:

    # issue 'set run' and disconnect while the program is still running
    printf 'hello EMC x 1.0\r\nset enable EMCTOO\r\nset run\r\n' | nc -q 2 localhost 5007

    # program completes fine; now try ANY set on a fresh connection:
    printf 'hello EMC x 1.0\r\nset enable EMCTOO\r\nset abort\r\n' | nc -q 2 localhost 5007
    # -> error: Cannot issue SET ABORT command while SET RUN command is still active
    # ... and it never recovers.

The window is easy to hit from scripts (`nc -q`), scanners/monitoring
probes, or any client that crashes mid-command. Note that even `set abort`
is refused — there is no protocol-level escape.

Not every disconnect wedges it: if the SET completes before the server
notices the disconnect (fast commands like `set machine on` often finish
first), the DONE path still runs and clears the latch. The reliable trigger
is a SET whose completion outlives the connection — `set run` of a program
of a few seconds is enough.

### Suggested direction (for discussion)

When a client with an outstanding setter (`cmdtimeout > 0`) is erased from
the client list, clear the latch (or transfer ownership): e.g. in the erase
branch at emcrsh.cc:3618, `if (i->cmdtimeout > 0.0) setterClear(*i);` before
`clients.erase(i)`. Alternatively, track the owning client in
`activeSetter`'s state and clear whenever that client goes away. The
existing timeout path (emcrsh.cc:3389-3391, "command 'SET {}' timed out")
already knows how to clear on expiry; the disconnect path just never
reaches it.

Note this is *not* the race fixed by 0d3ddc10f2 (SET JOINT_HOME vs
SET JOINT_WAIT_HOMED) - that one is about status propagation timing for a
connected client, whereas this is the latch outliving the client entirely.
Verified still present after that commit.

Happy to leave the implementation to you, or to attempt a fix if that's
useful - just say the word.

### Environment

- Observed on a branch based on current master (mechanism verified present
  in master's emcrsh.cc by line inspection at b3a596f904, 2026-08-17;
  latch code unchanged).
- uspace sim, Debian 12.
- Workaround we use meanwhile: one persistent connection per channel and
  never disconnecting while a SET is in flight; pace SETs by polling GETs.

### LINKED ISSUE #4401: linuxcncrsh: IPv6-only listen socket fails to start when IPv6 is disabled
Verified against master b3a596f904 (2026-08-17).

Summary

linuxcncrsh creates its listening socket as AF_INET6 only and relies on
IPv4-mapped-IPv6 addresses to serve IPv4 clients. On a machine where IPv6 is
disabled at the kernel level (ipv6.disable=1 on the kernel command line —
common on stripped-down / RT-tuned installs), socket(AF_INET6, …) fails with
EAFNOSUPPORT, so the listener is never created and linuxcncrsh exits at
startup.

Impact

- linuxcncrsh cannot be used at all on IPv6-disabled systems.
- Anything that drives LinuxCNC through linuxcncrsh fails as a result. I first
hit this as tests/mdi-queue failures, and it reproduces standalone by just
starting linuxcncrsh on such a box.

Steps to reproduce

1. Boot a machine with IPv6 disabled (ipv6.disable=1), or otherwise make
AF_INET6 unavailable.
2. Start a configuration (or run linuxcncrsh directly).
3. linuxcncrsh exits immediately; the socket() call for the listener fails
with EAFNOSUPPORT.

Where

src/emc/usr_intf/emcrsh.cc — the listener setup in initSocket():
socket(AF_INET6, SOCK_STREAM, 0) at emcrs
address.sin6_family = AF_INET6 at emcrsh.
b3a596f904). The peer-address logging tha
inet_ntop(AF_INET6, &csa.sin6_addr, …) ca
emcrsh.cc:3541.

Suggested direction (for discussion)

When the AF_INET6 socket cannot be create
listener. A couple of details that matter

- on accept(), use sockaddr_storage so the peer-address logging works for
both address families;
- set the fallback listen socket non-blocking to match the existing IPv6 path
(the poll loop assumes a non-blocking lis

The IPv6 path should stay exactly as-is systems —
this is only a fallback for the IPv6-unavthe
implementation to you; I just wanted the approach
documented.

I've actually implemented and tested the -clean,
verified on an ipv6.disable=1 box that liead
of exiting). Happy to open a PR from it ie word.

Environment

LinuxCNC 2.10-pre (master), uspace realtih
ipv6.disable=1.

Unrelated to #4400, but same file and foung both together for context.

BsAtHome:
The new linuxcncrsh server used v4-mapped-on-v6 to have one socket accept both IPv6 and IPv4 connections. However, if IPv6 is disabled at the kernel command line (or not compiled in), then it would fail to create a listening socket. Now there is a IPv4-only fallback in case IPv6 fails. This PR also fixes a lockup in the processing of SET commands when a client disconnects while a SET command is in flight.

Fixes #4400
Fixes #4401

### DIFF
--- a/src/emc/usr_intf/emcrsh.cc
+++ b/src/emc/usr_intf/emcrsh.cc
@@ -185,6 +185,7 @@ typedef struct {
 	int commProt;					// Deprecated
 } connectionRecType;
 
+static bool isipv4 = false;	// We try AF_INET6, but can fallback to AF_INET
 static int port = 5007;
 static std::string helloPwd = "EMC";
 static std::string enablePWD = "EMCTOO";
@@ -262,6 +263,7 @@ static struct option longopts[] = {
 	{"enablepw",  1, NULL, 'e'},
 	{"path",      1, NULL, 'd'},
 	{"quiet",     0, NULL, 'q'},
+	{"ipv4",      0, NULL, '4'},
 	{}
 };
 
@@ -364,24 +366,66 @@ static void set_nonblock(int fd)
 	}
 }
 
+static int getSocket4(struct sockaddr_in &addr, socklen_t &slen)
+{
+	int sockfd = socket(AF_INET, SOCK_STREAM, 0);
+	if (sockfd < 0) {
+		xperror("IPv4 socket()");
+		return -1;
+	}
+	addr = {};
+	addr.sin_family = AF_INET;
+	addr.sin_addr.s_addr = INADDR_ANY;
+	addr.sin_port = htons(port);
+	slen = sizeof(addr);
+	return sockfd;
+}
+
 static int initSocket()
 {
 	int optval = 1;
 	int err;
 	int sockfd;
-	struct sockaddr_in6 address = {};
+	union {
+		struct sockaddr_in6 addr6;
+		struct sockaddr_in  addr4;
+	} addr;
+	socklen_t slen;
 
-	sockfd = socket(AF_INET6, SOCK_STREAM, 0);
-	if (sockfd < 0) {
-		xperror("socket()");
-		return -1;
+	if (isipv4) {
+		// Forced IPv4, set from command-line
+		if ((sockfd = getSocket4(addr.addr4, slen)) < 0)
+			return -1;
+	} else {
+		// Using AF_INET6 will also allow IPv4 to connect (v4-mapped-on-v6)
+		// We will try IPv6 first and hope it is available
+		sockfd = socket(AF_INET6, SOCK_STREAM, 0);
+		if (sockfd < 0) {
+			if (EAFNOSUPPORT == errno) {
+				// Someone disabled IPv6 on the machine. Try IPv4 instead.
+				isipv4 = true;
+				if ((sockfd = getSocket4(addr.addr4, slen)) < 0)
+					return -1;
+			} else {
+				xperror("IPv6 socket()");
+				return -1;
+			}
+		} else {
+			// Ensure that v4-mapped-on-v6 is enabled. In theory, it could fail if
+			// there is no IPv4 stack available, but we don't care in that case.
+			int zero = 0;
+			setsockopt(sockfd, IPPROTO_IPV6, IPV6_V6ONLY, &zero, sizeof(zero));
+
+			addr.addr6 = {};
+			addr.addr6.sin6_family = AF_INET6;
+			addr.addr6.sin6_addr = IN6ADDR_ANY_INIT;
+			addr.addr6.sin6_port = htons(port);
+			slen = sizeof(addr.addr6);
+		}
 	}
+
 	setsockopt(sockfd, SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval));
-	// Using AF_INET6 will also allow IPv4 to connect (v4-mapped-on-v6)
-	address.sin6_family = AF_INET6;
-	address.sin6_addr = IN6ADDR_ANY_INIT;
-	address.sin6_port = htons(port);
-	err = bind(sockfd, reinterpret_cast<struct sockaddr *>(&address), sizeof(address));
+	err = bind(sockfd, reinterpret_cast<struct sockaddr *>(&addr), slen);
 	if (err) {
 		close(sockfd);
 		xperror("bind()");
@@ -3424,6 +3468,13 @@ static void closeClient(connectionRecType &ctx)
 {
 	if (ctx.sock < 0)
 		return;
+	if(ctx.cmdtimeout > 0.0) {
+		// Happens when a client disconnects with an active SET command. This
+		// is bad because it desynchronizes the command channel from the error
+		// channel. We don't know what happened anymore.
+		info("Warning: Client '%s' disconnected while waiting for SET %s", ctx.hostname.c_str(), activeSetter.c_str());
+		setterClear(ctx);
+	}
 	info("Connection terminated '%s' (%d)", ctx.hostname.c_str(), ctx.sock);
 	close(ctx.sock);
 	ctx.sock = -1;
@@ -3493,8 +3544,11 @@ static int sockMain(int svrfd)
 		} else if (pfds[0].revents & POLLIN) {
 			// POLLIN on a listen socket means new connection available
 			int cfd;
-			struct sockaddr_in6 csa;
-			socklen_t csal = sizeof(csa);
+			union {
+				struct sockaddr_in6 csa6;
+				struct sockaddr_in  csa4;
+			} csa;
+			socklen_t csal = sizeof(csa);	// Will be the largest of the two
 			cfd = accept(svrfd, reinterpret_cast<struct sockaddr *>(&csa), &csal);
 			if (cfd < 0) {
 				switch (errno) {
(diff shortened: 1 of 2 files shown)


src/emc/usr_intf/emcrsh.cc
```diff
@@ -369,19 +370,41 @@ static int initSocket()
...
+			addr.addr4.sin_family = AF_INET;
+			addr.addr4.sin_addr.s_addr = INADDR_ANY;
+			addr.addr4.sin_port = htons(port);
+			slen = sizeof(addr.addr4);
+		} else {
+			xperror("IPv6 socket()");
+			return -1;
+		}
 	}
 	setsockopt(sockfd, SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval));
```
grandixximo, line 427: Dual-stack here relies on net.ipv6.bindv6only defaulting to 0. With that sysctl
set to 1, socket(AF_INET6, ...) still succeeds, so the fallback never fires and
IPv4 clients get ECONNREFUSED. Measured:

    v6only=0  IPv4 client: connected OK
    v6only=1  IPv4 client: connect FAILED (Connection refused)

Worth pinning the assumption rather than inheriting it, on the IPv6 branch only?

    int zero = 0;
    setsockopt(sockfd, IPPROTO_IPV6, IPV6_V6ONLY, &zero, sizeof(zero));

src/emc/usr_intf/emcrsh.cc
```diff
@@ -369,19 +370,41 @@ static int initSocket()
...
+
+	// Using AF_INET6 will also allow IPv4 to connect (v4-mapped-on-v6)
+	// We will try IPv6 first and hope it is available
+	addr.addr6 = {};
+	addr.addr6.sin6_family = AF_INET6;
+	addr.addr6.sin6_addr = IN6ADDR_ANY_INIT;
+	addr.addr6.sin6_port = htons(port);
+	socklen_t slen = sizeof(addr.addr6);
 
 	sockfd = socket(AF_INET6, SOCK_STREAM, 0);
```
grandixximo: Is EAFNOSUPPORT the only way the IPv6 listener can be unavailable in practice?
A restricted netns or an LSM/seccomp policy returns EACCES or EPERM here, and
there are setups where socket() succeeds but bind() to :: is what fails. Since
IPv4 is a strict fallback, any downside to taking it on any IPv6 failure?

src/emc/usr_intf/emcrsh.cc
```diff
@@ -3493,8 +3523,11 @@ static int sockMain(int svrfd)
 		} else if (pfds[0].revents & POLLIN) {
 			// POLLIN on a listen socket means new connection available
 			int cfd;
-			struct sockaddr_in6 csa;
-			socklen_t csal = sizeof(csa);
+			union {
+				struct sockaddr_in6 csa6;
+				struct sockaddr_in  csa4;
+			} csa;
+			socklen_t csal = isipv4 ? sizeof(csa.csa4) : sizeof(csa.csa6);
```
grandixximo: accept() writes the real length back, so does the family need to be known here?
sizeof(csa) is correct for both.

Broader: would sockaddr_storage plus getnameinfo(..., NI_NUMERICHOST |
NI_NUMERICSERV) let the union, isipv4, and the duplicated logging branch below
all disappear?

src/emc/usr_intf/emcrsh.cc
```diff
@@ -3424,6 +3447,13 @@ static void closeClient(connectionRecType &ctx)
 {
 	if (ctx.sock < 0)
 		return;
+	if(ctx.cmdtimeout > 0.0) {
```
grandixximo, line 3471: The latch clears, but the NML command is still in flight, so its completion or
error lands on whichever client polls next. Intended trade-off, or would
recording emcCommandSerialNumber at close time and discarding status up to it
be cheap enough to close the desync?

src/emc/usr_intf/emcrsh.cc
BsAtHome: Not really. The printing of the message just later on needs access to the `port` field. That is only accessible on the real `sockaddr_in{6}` structure. That makes the union a better match.

src/emc/usr_intf/emcrsh.cc
BsAtHome: It is an intended trade-off. The alternative, when not cleared, is a timeout. When that hits, you have exactly the same situation. The default timeout (if not set by the user) is about 10 years. My guess is you don't want to wait that long.
Other commands from other clients will have their own serial number recorded. If the command in flight blocks forever, then all others will block on the serial number too. That is a problem how the underlying system works and we cannot overcome that limitation.

src/emc/usr_intf/emcrsh.cc
BsAtHome: The setsockopt() was missing. Added.

Also added a forced IPv4 to the command-line (option -4) so you can force the system to use IPv4 from the start and not try IPv6 at all.

rene-dev:
Why on earth would you disable ipv6?

BsAtHome:
> Why on earth would you disable ipv6?

Normally, you wouldn't, at all.
But, there actually is a use-case in local networks that are very tightly controlled.

grandixximo:
> Why on earth would you disable ipv6?

Improve latency on your CNC which does not need it

PR #4404, opened 2026-08-18
SimConfigs: Add melfa-sim using 'three21' kinematics

Sigma1912:
- Create a new folder for serial manipulator simulations
- Move 'melfagui' vismach model which relys on the config stl files into the config folder
- Add a new sim that applies 'three21' kins
- Add and clarify README descriptions
- Fix startup warnings about nonexistent .tbl files


Sigma1912:
Is there concern about moving vismach models that rely on specific sim config stl files to those sim config folders?

IMO it doesn't make much sense in splitting vismach models from the stl model files they rely on.

Currently someone  wanting to reuse these vismach models already needs to copy the required stl files from the specific sim config. With this change the single vismach file would need to copied with it.

BsAtHome:
The question is whether the files/executables currently in the tree belong there if they are specialized for the sim(s) only. Files that are not generic may not belong in the normal tree, but need to be moved to where they belong.

You moved an executable `src/hal/user_comps/vismach/melfagui.py` into the config directory but the `debian/linuxcnc.install.in` still has it. Package build fails (not related to the debian sid failures).
Another question would be whether the man page still needs to be there if the program is part of the config.

Sigma1912:
There are some vismach models that create their own geometry and some that use specific stl files that live in the sim config folders (pumagui). 
A model that relies on specific stl data has little use without it.

IMO all models in `/src/hal/user_comps/vismach/` should be moved to their respective sim configs. Happy to follow up with a PR.

> You moved an executable src/hal/user_comps/vismach/melfagui.py into the config directory but the debian/linuxcnc.install.in still has it. Package build fails (not related to the debian sid failures).

Should be fixed now.


> Another question would be whether the man page still needs to be there if the program is part of the config.

I don't see a point in having these man pages. 


As someone who has created some of these sims I find it MUCH easier if all required files are in the sim config itself.

BsAtHome:
> IMO all models in `/src/hal/user_comps/vismach/` should be moved to their respective sim configs. Happy to follow up with a PR.

If this stuff is purely local to the sim(s), then it would be more logical to have it there.

> > Another question would be whether the man page still needs to be there if the program is part of the config.
> I don't see a point in having these man pages.

Yeah, but it is still referenced in the package build files. Every time you (re)move from previously installed files, then you may want to grep your way through the debian/ dir to see where else it is used.

> As someone who has created some of these sims I find it MUCH easier if all required files are in the sim config itself.

It is not really about how easy it is. The real test is where it belongs. If a file is (only) part of the sim, then it belongs in that sim directory. Files that are shared between sims should probably use a proper sim directory structure, unless its use is generic enough to become distributable.

The trouble we are in now is that these files were previously part of the binaries in the bin/ directory. That means that anybody could be using them for anything. Moving them would then break people's setup. That is something that needs to be checked against. If these executables make no sense outside the sim, then, by all means, we could move them. But if there is any chance that someone is (re)using them, then we must be very careful.

Sigma1912:
> The trouble we are in now is that these files were previously part of the binaries in the bin/ directory. That means that anybody could be using them for anything. Moving them would then break people's setup. That is something that needs to be checked against. If these executables make no sense outside the sim, then, by all means, we could move them. But if there is any chance that someone is (re)using them, then we must be very careful.


There is simply no way of knowing what people are doing out of tree. 
In the worst case they would have to copy the respective vismach model into their config folder which can also be done on a deb install. Anybody reusing these files would probably be capable of doing that. 

The only thing I can think of is putting these files in their respective sim configs which would then use the local copy.  Then we could add a line to these files in /src/hal/user_comps/vismach/ that prints a deprecation notice pointing to the respective sim config.
 Maybe even a tk popup for those who do not run from a terminal.
 Would that be acceptable?

BsAtHome:
Yes, copy the file to where it belongs and then add a deprecation warning (terminal/gui) to the old ones. Then after "some" time, we can get rid of them.

PR #4405, opened 2026-08-18
halrmt: Allow IPv4 as fallback when IPv6 is disabled.

BsAtHome:
The new halrmt server uses v4-mapped-on-v6 to have one socket accept both IPv6 and IPv4 connections. However, if IPv6 is disabled at the kernel command line (or not compiled in), then it would fail to create a listening socket. This PR adds an IPv4-only fallback in case IPv6 fails.

### DIFF
--- a/src/hal/utils/halrmt.cc
+++ b/src/hal/utils/halrmt.cc
@@ -68,6 +68,7 @@ using namespace linuxcnc;
 static std::string helloPwd   = "EMC";       // Connect password
 static std::string enablePwd  = "EMCTOO";    // Enable password
 static std::string serverName = "EMCNETSVR"; // Server name written in hello response
+static bool isipv4 = false;         // We try AF_INET6, but can fallback to AF_INET
 static int port = 5006;
 static std::string inifilename;
 static volatile int quitloop = 0;   //  Signal to main loop to exit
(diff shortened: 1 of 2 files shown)


src/hal/utils/halrmt.cc
```diff
@@ -2886,19 +2887,41 @@ static int initSocket()
...
+            addr.addr4.sin_family = AF_INET;
+            addr.addr4.sin_addr.s_addr = INADDR_ANY;
+            addr.addr4.sin_port = htons(port);
+            slen = sizeof(addr.addr4);
+        } else {
+            xperror("IPv6 socket()");
+            return -1;
+        }
     }
     setsockopt(sockfd, SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval));
```
grandixximo, line 2943: Same as #4403: this depends on net.ipv6.bindv6only being 0. With that sysctl at
1, socket(AF_INET6, ...) succeeds so the fallback never fires and IPv4 clients
get ECONNREFUSED. Confirmed:

    v6only=0  IPv4 client: connected OK
    v6only=1  IPv4 client: connect FAILED (Connection refused)

Explicit setsockopt(sockfd, IPPROTO_IPV6, IPV6_V6ONLY, &zero, ...) on the IPv6
branch?

src/hal/utils/halrmt.cc
```diff
@@ -2886,19 +2887,41 @@ static int initSocket()
...
+
+    // Using AF_INET6 will also allow IPv4 to connect (v4-mapped-on-v6)
+    // We will try IPv6 first and hope it is available
+    addr.addr6 = {};
+    addr.addr6.sin6_family = AF_INET6;
+    addr.addr6.sin6_addr = IN6ADDR_ANY_INIT;
+    addr.addr6.sin6_port = htons(port);
+    socklen_t slen = sizeof(addr.addr6);
 
     sockfd = socket(AF_INET6, SOCK_STREAM | SOCK_CLOEXEC | SOCK_NONBLOCK, 0);
```
grandixximo: Is EAFNOSUPPORT the only failure worth falling back on? EACCES/EPERM from a
restricted netns or LSM policy reach the same point, and bind() to :: can fail
independently of socket(). Reason not to fall back on any IPv6 failure?

src/hal/utils/halrmt.cc
```diff
@@ -2985,8 +3008,11 @@ static int sockMain(int svrfd)
         } else if (pfds[0].revents & POLLIN) {
             // POLLIN on a listen socket means new connection available
             int cfd;
-            struct sockaddr_in6 csa;
-            socklen_t csal = sizeof(csa);
+            union {
+                struct sockaddr_in6 csa6;
+                struct sockaddr_in  csa4;
+            } csa;
+            socklen_t csal = isipv4 ? sizeof(csa.csa4) : sizeof(csa.csa6);
```
grandixximo: accept4() writes the real length back, so sizeof(csa) works for both families.
Would sockaddr_storage plus getnameinfo(..., NI_NUMERICHOST | NI_NUMERICSERV)
remove the union, isipv4, and the duplicated logging branch below altogether?

BsAtHome: Indeed. Still need access to the specific sockaddr for the port.

src/hal/utils/halrmt.cc
BsAtHome: Added the setsockopt.

src/hal/utils/halrmt.cc
BsAtHome: Any other failure would be a bad sign. You can add `-4` now to force IPv4 only. If IPv6 fails due to other restrictions than not being available, then you can manually use IPv4 instead.
BTW, retrying on other errors may be futile because many will happen on IPv4 too if they hit IPv6.

src/hal/utils/halrmt.cc
```diff
@@ -2881,24 +2882,66 @@ static int parseCommand(connectionRecType &ctx, std::string &line)
 *                            MAIN PROGRAM                              *
 ************************************************************************/
 
+static int getSocket4(struct sockaddr_in &addr, socklen_t &slen)
+{
+    int sockfd = socket(AF_INET, SOCK_STREAM, 0);
```
grandixximo: The refactor dropped SOCK_CLOEXEC | SOCK_NONBLOCK from both socket() calls here
and at line 2918. Master has them on the listen socket:

    sockfd = socket(AF_INET6, SOCK_STREAM | SOCK_CLOEXEC | SOCK_NONBLOCK, 0);

and halrmt has no fcntl fallback anywhere, unlike emcrsh which calls
set_nonblock() after listen(). So the listener is now blocking and inheritable:

- accept4() after a spurious POLLIN blocks the poll loop indefinitely (accept(2)
  BUGS: the connection can be aborted between poll and accept, leaving nothing
  to take). That is the caveat from #4401.
- halrmt forks and execs at line 628, so every component it launches inherits
  the listening fd and keeps port 5006 bound past halrmt's exit.

The first push had the flags on the IPv6 socket. Restoring them on both calls
should be all that is needed.

BsAtHome: At least you are watching. I must have been asleep. May still require some head banging ;-)

Fixed.

grandixximo:
Flags restored, thanks.

One leftover: `usage()` (halrmt.cc:3218) does not list `-4`, so `halrmt --help` omits a flag the man page documents. `linuxcncrsh`'s `usage()` in #4403 does list it. One line:

```c
"  -4,--ipv4             Only use IPv4\n"
```

BsAtHome:
Must have hurt my head ;-)

Added the usage() line.

PR #4406, opened 2026-08-18
mtconnect: adding optional user pages / html files

multigcs:
so you can add your digital twin and other pages to your mtconnect agent.

### DIFF
--- a/docs/src/man/man1/mtconnect-agent.1.adoc
+++ b/docs/src/man/man1/mtconnect-agent.1.adoc
@@ -64,6 +64,7 @@ All configuration is read from the *[MTCONNECT]* section:
 *TRANSPORT*::    Comma-separated list of *http*, *mqtt*, *shdr* (default *http*).
 *HTTP_PORT*::    Embedded HTTP agent port (default 5000).
 *HTTP_BIND*::    Interface to bind (default 127.0.0.1; use 0.0.0.0 for the LAN).
+*USER_PAGES*::   Path to optional user pages / html files.
 *SHDR_PORT*::    Port for the SHDR adapter when *shdr* is in *TRANSPORT* (default
   7878).  SHDR feeds an external MTConnect agent (e.g. cppagent) `<ts>|id|value`
   lines; configure that agent with a Devices.xml from *--dump-probe*.
--- a/lib/python/mtc/agent.py
+++ b/lib/python/mtc/agent.py
@@ -68,6 +68,10 @@ class AgentState:
         self._last_in_spindle = None
         self._last_values = {}
         self._lock = threading.Lock()
+        self.user_pages = self.ini.find("MTCONNECT", "USER_PAGES", "")
+        if self.user_pages and self.user_pages[0] != "/":
+            # use relativ path
+            self.user_pages = os.path.join(os.path.dirname(ini_path), self.user_pages)
 
     # -- data collection -----------------------------------------------------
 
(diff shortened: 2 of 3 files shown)


lib/python/mtc/agent.py
```diff
@@ -112,6 +116,16 @@ def extension_schema(self):
         with open(path, "rb") as fh:
             return fh.read(), "application/xml"
 
+    def user_page(self, name):
+        """Return (bytes, content_type) for a served user file, or None."""
+        if not self.user_pages:
+            return None
+        path = os.path.realpath(os.path.join(self.user_pages, name))
+        if path.startswith(self.user_pages) and os.path.isfile(path):
```
grandixximo, line 124: What happens here when `USER_PAGES=/cfg/pages` and the request is `/html/../pages_evil/twin.html`? `realpath` gives `/cfg/pages_evil/twin.html`, and that string still starts with `/cfg/pages`. Also, `self.user_pages` keeps any `..`/symlink form from the ini-dir join while `path` is canonicalized, so the two sides are not in the same form. Suggest realpath-ing the base once in `__init__` and using `os.path.commonpath((base, path)) == base` (or `path.startswith(base + os.sep)`).

lib/python/mtc/agent.py
```diff
@@ -112,6 +116,16 @@ def extension_schema(self):
...
+            with open(path, "r") as fh:
```
grandixximo, line 125: Reading in text mode means any binary asset (png, ico, woff2, ...) raises `UnicodeDecodeError`, which the handler turns into a 500. `"r"` also uses the locale encoding rather than utf-8. Could this be `"rb"` like `extension_schema`/`model_file`?

lib/python/mtc/agent.py
```diff
@@ -112,6 +116,16 @@ def extension_schema(self):
...
+                return fh.read(), "text/html"
```
grandixximo, line 126: Every file goes out as `text/html`. Browsers refuse to execute `<script src>` served with that MIME type, and stylesheets too in standards mode, so a digital-twin page with its own .js/.css will not load. `mimetypes.guess_type(path)` with a fallback would cover this; `_send` already accepts bytes.

lib/python/mtc/agent.py
```diff
@@ -68,6 +68,10 @@ def __init__(self, ini_path, buffer_size=131072):
         self._last_in_spindle = None
         self._last_values = {}
         self._lock = threading.Lock()
+        self.user_pages = self.ini.find("MTCONNECT", "USER_PAGES", "")
+        if self.user_pages and self.user_pages[0] != "/":
+            # use relativ path
```
grandixximo, line 73: nit: typo ("relativ"), and the comment restates the code. Either drop it or say why (anchors the pages dir to the ini location).

lib/python/mtc/agent.py
```diff
@@ -112,6 +116,16 @@ def extension_schema(self):
         with open(path, "rb") as fh:
             return fh.read(), "application/xml"
 
+    def user_page(self, name):
```
grandixximo, line 119: Would you consider a small test next to the existing mtc tests: serve a page, 404 on a missing one, and 404 on a `../` escape attempt? That third case is what guards the containment check long-term.

lib/python/mtc/http_agent.py
```diff
@@ -96,6 +96,14 @@ def do_GET(self):
                     else:
                         data, content_type = result
                         self._send(data, content_type=content_type)
+                elif route.startswith(f"/html/"):
```
grandixximo, line 99: nit: f-string with no placeholders, drop the `f`.

docs/src/man/man1/mtconnect-agent.1.adoc
```diff
@@ -64,6 +64,7 @@ All configuration is read from the *[MTCONNECT]* section:
 *TRANSPORT*::    Comma-separated list of *http*, *mqtt*, *shdr* (default *http*).
 *HTTP_PORT*::    Embedded HTTP agent port (default 5000).
 *HTTP_BIND*::    Interface to bind (default 127.0.0.1; use 0.0.0.0 for the LAN).
+*USER_PAGES*::   Path to optional user pages / html files.
```
grandixximo, line 67: Where do these files appear on the wire? The route prefix `/html/<name>` only exists in the code right now; one sentence here would make it discoverable.

PR #4410, opened 2026-08-19
docs: fix the two remaining htmldocs CI warnings

grandixximo:
Clears the two remaining `htmldocs` CI warnings.

`python-hal-interface.adoc` used inter-document xref syntax, `xref:python-lcnc_realtime.adoc[]`. The docs build as one book, so it renders as `href="#python-lcnc_realtime.adoc"`, a fragment that does not exist; `htmlcheck.sh` flagged it in all eight languages. Now `<<cha:python-lcnc_realtime>>`, the anchor the target chapter declares.

The rc.local example in `hm2_rpspi.9.adoc` used a markdown ``` fence. po4a does not recognise fences and reflows the block onto one line, so every translated build hit `unterminated listing block` and lost the rest of the page (the German manpage drops NOTE, AUTHOR and LICENSE). A plain `----` listing block fixes it; po4a then keeps the line breaks.


grandixximo:
no warnings 🌞

BsAtHome:
One new bummer from me and one old one, it seems ;-)

PR #4413, opened 2026-08-19
Fix python site-packages dir

NTULINUX:
2.9 backport

### DIFF
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -1628,7 +1628,7 @@ fi
 AC_MSG_RESULT([$PYTHON_TK_VERSION])
 
 AC_MSG_CHECKING(for site-package location)
-SITEPY="$($PYTHON -c 'import sysconfig; s = sysconfig.get_scheme_names(); m=list(set(("deb_system", "rpm_prefix")) & set(s)); print(sysconfig.get_path("platlib", m.__getitem__(0))) if m else print("/usr/lib/python3/dist-packages");')"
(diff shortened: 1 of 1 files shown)


grandixximo:
CI failure on `package-arch (bullseye)` is real. Bullseye has python3.9: no `deb_system` (added in 3.10), no `rpm_prefix` (Fedora-only), and `sysconfig.get_default_scheme()` does not exist before 3.10, so the new line raises `AttributeError` and `SITEPY` comes out empty. Verified in a bullseye container. Empty `SITEPY` drops the python tree into `$(DESTDIR)` root, `usr/lib/python3` in `debian/linuxcnc.install.in` matches nothing, package build fails. Master never hit this; it no longer builds bullseye.

`get_default_scheme()` is the wrong fallback anyway: it yields `site-packages`, but Debian pythons without `deb_system` only import from `dist-packages`. Suggested fix, works on 3.9+:

```
SITEPY=`$PYTHON -c "import sysconfig; s=sysconfig.get_scheme_names(); print(next((sysconfig.get_path('platlib', m) for m in ('deb_system', 'rpm_prefix') if m in s), '/usr/lib/python3/dist-packages'))"`
```

NTULINUX:
Should I send a PR for master branch to match the new 2.9 python handling/SITEPY or leave it the way it is?

grandixximo:
We can merge 2.9 into master, there is another fix that can merge up that I'm waiting for coordination

PR #4414, opened 2026-08-19
G12.1, G13.1: select the kinematics type from G-code

grandixximo:
Switchable kinematics is selected today by writing `motion.switchkins-type` from HAL, in practice `M68 E- Q-` against an analog output netted in the INI. The interpreter never sees it, so a program is read, its limits checked and its path looked ahead in whatever kinematics the interpreter last knew about. This adds `G12.1 P-` and `G13.1` so the selection goes through the interpreter and motion together.

Part of #4374, which is where the rest of the multiaxis work is described. Nothing here depends on the other pieces.

## What it does

- `G12.1 P-` selects a kinematics type by number, `G13.1` cancels to identity kinematics. Both queue a synchronisation point, so no move is ever planned in one kinematics and executed in another.
- The module declares what each of its types IS: `switchkinsDeclare(ktype, flags)` with `KINSTYPE_IDENTITY` (no transform) and `KINSTYPE_PRIMARY` (the working transform) from kinematics.h. The flags travel through motion's status to the interpreter, so `G13.1` means identity whatever number the module lists it as. A module that declares nothing keeps working by number, but `G13.1` is an error with a pointer to the switchkins docs, since the number of the identity kinematics is then a guess. Two identity types, or a declaration naming a type the module does not provide, fail the module load.
- A `G12.1 P-` naming a type the module does not provide is a read-time error now; it was a mid-program motion abort.
- `#<_kins_type>` is read back from motion on every synch, so it stays right across an abort or a switch made behind the interpreter's back.
- New `motion.kins-type` status pin reports the type actually in force.
- Selecting from HAL is deprecated. `motion.switchkins-type` still works and warns once per session.
- Every switchkins sim config moves to G12.1, so none of them trips the deprecation warning. The nets from `motion.analog-out-03` go with them, and the trsrn vismach guis and the halshow watch lists follow `motion.kins-type`.

## Why these numbers

Out-of-tree patches have used G12.1 and G13.1 for kinematics switching for years, so configurations in the wild already speak them.

They also keep LinuxCNC parallel with other controllers, where G12.1 and G13.1 turn polar coordinate interpolation on and off. That is itself a change of kinematics, and if LinuxCNC ever gets it, switchkins is where it belongs rather than as a mechanism of its own. G12 and G13 are circular pocket milling on those controllers, so leaving the plain codes free keeps room to port that later.

## Renames

`adjustKinsVar0` was the kinematics type and `kinsType` was a char toggling between 'r' and 's'. Neither name described what it held. The traj-prefixed status fields moved into `EMC_TRAJ_STAT` where they belong.

No other pin, G-code or INI name changes.

## Testing

The kins-switch test drives a program across both directions of the switch on 5axiskins: the pose is preserved across it, a W move in identity touches no other joint, `#<_kins_type>` reports each selection, a negative or unprovided P word is refused at read time, and the abort routine's G13.1 runs to its end instead of dying on the queue wait.

Every switchkins sim config driven headless through its M-codes, correct type each time and no deprecation warning: millturn, scara, bridgemill, hexapod, melfa, puma, xyzac-trt, xyzbc-trt, xyzab-tdr, xyzacb-trsrn and xyzbca-trsrn, the last two including the TWP wrappers. The M428 remaps now spell `G12.1 P#<kinstype>` instead of assuming type 0, since identity is no longer necessarily the first type.

Also checked by hand: a switch made from HAL behind the interpreter's back, which the interpreter picks up on the next synch, with one warning per session; and abort of a program that had switched mid-run, after which motion and the interpreter still agree.


configs/sim/axis/vismach/millturn/remap_subs/428remap.ngc
```diff
@@ -1,29 +1,17 @@
...
-     M2
-o1 endif
-
-   M66  E0 L0                           ; force synch
-   M68  E#<SWITCHKINS_PIN> Q#<kinstype> ; set kinstype value
+   G12.1 P#<kinstype>                   ; select kinstype, syncs interp and motion
    M128                                 ; switch limits
    G10 L2 P7 X-290 Y0 Z-160 A0          ; reset home offset
    G59.1                                ; activate home offset
-   M66  E0 L0                           ; force synch
```
Sigma1912, line 18: This sync is needed for the limit values, that get changed in M128, to be read by the interpreter.

configs/sim/axis/vismach/millturn/remap_subs/429remap.ngc
```diff
@@ -1,29 +1,17 @@
...
-     M2
-o1 endif
-
-   M66  E0 L0                           ; force synch
-   M68  E#<SWITCHKINS_PIN> Q#<kinstype> ; set kinstype value
+   G12.1 P#<kinstype>                   ; select kinstype, syncs interp and motion
    M129                                 ; switch limits
    G10 L2 P8 X-160 Y0 Z-290 A0          ; reset home offset
    G59.2                                ; activate home offset
-   M66  E0 L0                           ; force synch
```
Sigma1912, line 18: This sync is needed for the limit values, that get changed in M129, to be read by the interpreter.

docs/src/man/man9/motion.9.adoc
```diff
@@ -253,9 +253,20 @@ Note: feed-inhibit applies to G-code commands -- not jogs.
...
+  select the machine kinematics functions. Extra G-code commands are
   required to synchronize task and motion before and after changes to
   the pin value.
+  *Deprecated*: the interpreter does not see this pin, so limits and
+  look ahead go on using the kinematics it last knew about. Use the
+  G-code words *G12.1 P-* and *G13.1*, which ask motion directly and
+  synchronize task and motion themselves. Motion reports the
+  deprecation once, the first time the pin is used to change the
+  kinematics. The pin is in a grace period: it keeps working for now,
+  and is meant to go.
```
Sigma1912: May I suggest:
'it keeps working for now,
  but is meant to be removed in the future.'

docs/src/motion/switchkins.adoc
```diff
@@ -135,20 +138,79 @@ A module providing more than three kinematics types has one
...
 
-=== G-/M-code commands
+[WARNING]
+Selecting the kinstype from HAL is deprecated and motion says so, once,
+the first time the pin is used to change it. The interpreter does not
+see the pin, so a program is read, its limits checked and its path
+looked ahead in whatever kinematics the interpreter last knew about,
+which is not necessarily the one that will run it. Use 'G12.1' and
+'G13.1'. The pin is in a grace period: it keeps working for now, and is
+meant to go.
```
Sigma1912: same suggested change as above

Sigma1912 (commented):
Is there a reason to keep the 'M428','M429' and 'M430' remaps in all those configs that do not have a `M128`/`M129` call to switch the limit values? 


The `M66` sync after `M128`, `M129` calls are required so the interpreter rereads the changed limit values. Once limit switching in this manner is no longer required both the `M128 / M129` and the following `M66` can go.

docs/src/man/man9/motion.9.adoc
grandixximo: reworded as suggested

docs/src/motion/switchkins.adoc
grandixximo: agreed

configs/sim/axis/vismach/millturn/remap_subs/428remap.ngc
grandixximo: my bad should not have removed without checking what they were for, restored with proper comment, what the sync is actually for.

configs/sim/axis/vismach/millturn/remap_subs/429remap.ngc
grandixximo: restored

grandixximo:
No reason beyond keeping this PR to one subject. Where there are no limits to switch the remap is now a `G12.1` and the guard that checks the switch took, so you are right that it is thin, but dropping it means moving that guard somewhere or losing it, and it is what the halui buttons and the docs point at.

I would rather do it once the limits move out of `M128`/`M129`, because that is the change that empties the millturn remaps too, and then all of them can be cleaned up the same way in one pass instead of two. It also keeps this diff to one thing: migrate the selection mechanism, change nothing else.

PR #4417, opened 2026-08-19
gmoccapy: Fix reading [DISPLAY]CYCLE_TIME when set as a real

### LINKED ISSUE #4415: Gmoccapy: Multiple sim configs fail to launch
Testing on current master:

### Example
None of the Gmoccapy sim configs in `configs/sim/gmoccapy/non_trivial_kinematics` launch.

### Possible problem
Comparing the terminal output to 2.9.10 when lauching `configs/sim/gmoccapy/non_trivial_kinematics/Gantry/gantry_mm.ini `   points to a possible clash between the gmoccapy-internal ini value check and the new linuxcnc ini parser

2.9.10 (config launches): 
` 
[Gmoccapy.GMOCCAPY.GETINIINFO][WARNING]  Wrong entry [DISPLAY] CYCLE_TIME in INI File! Will use gmoccapy default 150 (getiniinfo.py:56)
` 

current master (config fails)

`
/home/user/git/linuxcnc-master/configs/sim/gmoccapy/non_trivial_kinematics/Gantry/gantry_mm.ini:9: warning: Trailing character(s) in signed integer conversion ([DISPLAY]CYCLE_TIME='0.200')
` 

Further testing shows that a missing  [DISPLAY] CYCLE_TIME entry in `configs/sim/gmoccapy'/moccapy.ini`  is handled by gmoccapy and does not cause a launch failure.

` [Gmoccapy.GMOCCAPY.GETINIINFO][WARNING]  Wrong entry [DISPLAY] CYCLE_TIME in INI File! Will use gmoccapy default 150 (getiniinfo.py:56)`  

However in 2.9.10 the same missing ini value causes the launch to fail with this error:

``` 

  File "/home/user/git/linuxcnc-master/bin/gmoccapy", line 6575, in <module>
    app = gmoccapy(sys.argv)
          ^^^^^^^^^^^^^^^^^^
  File "/home/user/git/linuxcnc-master/bin/gmoccapy", line 514, in __init__
    GLib.timeout_add( cycle_time, self._periodic )  # time between calls to the function, in milliseconds
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/gi/overrides/GLib.py", line 612, in timeout_add
    return GLib.timeout_add(priority, interval, function, *user_data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Argument 1 does not allow None as a value
 (gmoccapy:75)
 ``` 
which indicates a problem with a value returned  from ini parsing/checking
https://github.com/LinuxCNC/linuxcnc/blob/0bb86dd6675ecdcd8646b14a4e5d5de2820bf386/src/emc/usr_intf/gmoccapy/gmoccapy.py#L513-L514

https://github.com/LinuxCNC/linuxcnc/blob/0bb86dd6675ecdcd8646b14a4e5d5de2820bf386/src/emc/usr_intf/gmoccapy/getiniinfo.py#L49-L57

### PR DESCRIPTION
**Pull request by BsAtHome**
Several sims use the wrong `[DISPLAY]CYCLE_TIME` being set in seconds, whereas the docs say it is in milliseconds. The old INI read would cause an exception, which no longer happens. The new INI-file reader warns of trailing context and return the value zero (0). It will now test for existence and whether it is larger than 0. Otherwise the default is used.

Also, the existence test "not val" should read "val is None", which has been fixed in two places where INI values are read and zero (0) is a valid value.

Fixes #4415

After discussion in #4415 gmoccapy now accepts both seconds and milliseconds bringing it in line with Axis.

### DIFF
--- a/share/gscreen/skins/silverdragon/silverdragon_handler.py
+++ b/share/gscreen/skins/silverdragon/silverdragon_handler.py
@@ -243,7 +243,7 @@ class HandlerClass:
         self.halcomp["search_vel"] = search_vel
         self.halcomp["sensor_height"] = sensor_height
         self.halcomp["maxprobe"] = maxprobe
-        if not xpos or not ypos or not zpos or not maxprobe:
+        if xpos is None or ypos is None or zpos is None or maxprobe is None:
             self.widgets.chk_use_auto_zref.set_active(False)
             self.widgets.chk_use_auto_zref.set_sensitive(False)
             self.widgets.btn_g30.set_sensitive(False)
@@ -271,7 +271,7 @@ class HandlerClass:
         # set up laser crosshair offsets
         xpos = self.gscreen.inifile.getreal("LASER", "X")
         ypos = self.gscreen.inifile.getreal("LASER", "Y")
-        if not xpos or not ypos:
+        if xpos is None or ypos is None:
             self.widgets.btn_laser_zero.set_sensitive(False)
             self.widgets.tbtn_laser.set_sensitive(False)
         else:
--- a/src/emc/usr_intf/gmoccapy/getiniinfo.py
+++ b/src/emc/usr_intf/gmoccapy/getiniinfo.py
@@ -47,14 +47,20 @@ class GetIniInfo:
             sys.exit()
 
     def get_cycle_time(self):
-        temp = self.inifile.getint("DISPLAY", "CYCLE_TIME")
-        try:
-            return temp
-        except:
-            message = ("Wrong entry [DISPLAY] CYCLE_TIME in INI File! ")
-            message += ("Will use gmoccapy default 150")
-            LOG.warning(message)
-            return 150
+        temp = self.inifile.getreal("DISPLAY", "CYCLE_TIME")
+        if temp is not None:
+            # Can be in seconds or milliseconds and, regardless,
+            # the result must be larger than zero to be valid.
+            # This brings gmoccapy in line with f.ex. AXIS.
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: share/gscreen/skins/silverdragon/silverdragon_handler.py [hunk 1]
```diff
@@ -243,7 +243,7 @@ def init_tool_measurement(self):
         self.halcomp["search_vel"] = search_vel
         self.halcomp["sensor_height"] = sensor_height
         self.halcomp["maxprobe"] = maxprobe
-        if not xpos or not ypos or not zpos or not maxprobe:
+        if xpos is None or ypos is None or zpos is None or maxprobe is None:
```
grandixximo, line 246: Same falsy-zero pattern just below this hunk: `silverdragon_handler.py:275` checks LASER offsets with `if not xpos or not ypos:`. `LASER X = 0` now returns `0.0` (falsy) and disables the laser buttons. Same `is None` fix applies.

Otherwise LGTM, verified locally: gantry_mm.ini launches, falls back to 150 ms, no more 99% CPU spin.

BsAtHome: Right, fixed.

There will probably be more of these lurking throughout the code. Must be future project ;-)

Issue #4408, opened 2026-08-18
Feature Request: add fields to tool table


alex-pres:
My proposal is to extend tool table with few fields that will allow to make LinuxCNC GUI more user friendly. 

For example it would be good to calculate and show feed per tooth for a milling tool, but we can't because we don't know tooth number. Also, for same goal it would be good to know tool material: HSS or Carbide. 

Of course every GUI may invent its own way to store additional values, but it is not reliable because tool.tbl may be externally replaced or changed and linked data will be corrupted.

There are two ways to implement it:

## 1. Define fixed fields and hope it will be enough

I'd add the following:
* Material: string  with recommended values like HSS, Carbide, Ceramic, CBN, PCD, Other
* Tooth Number: int [ useless on lathe but good to have same format anywhere ] 
* Tool Type: string with recommended values like : end mill, ball nose mill, bull nose mill, face mill, chamfer mill, slot mill, thread mill, twist drill, center drill, reamer, thread tap, other. 
* For lathe: insert type

However, this list may never be complete.

## 2. Define API for GUI to store values in tool table + document recommended field names and values, make it available from Python

I propose to add 1 new field named E to tool table
Format: FIELD NAME (up to 3 chars), then = then value (up to 12 chars). fields separated by ; 
like
`M=HSS;Z=4;T=EM;`
or for lathe
`M=CBN;IT=CCMT060204;`

If it is generally accepted as idea, I can implement it and submit as pull request. 

It has been said that it is CAM software job and it is true, but for simple tasks CAM is not always necessary. For example many lathe jobs are coded right in the shop, on LinuxCNC PC.

andypugh:
I believe that it is possible to use a database as a tool table. 
https://linuxcnc.org/docs/2.9/html/tooldatabase/tooldatabase.html

Part of the reason for this is to allow integrators to add any fields that they wish to the tool table.

alex-pres:
Looks good! It closes the problem.

PR #4422, opened 2026-08-19
tooldb: notify db_program of G10 tool offsets

alex-pres:
In db mode no tool table file is written, so offsets set with G10 L1/L10/L11 were lost at restart without the TOOL_OFFSET notification.

may be there was good reason to comment these lines, but now it seems to work ok. I added a test for it.

### DIFF
--- a/tests/tooldb/db-g10-offset/.gitignore
+++ b/tests/tooldb/db-g10-offset/.gitignore
@@ -1,0 +1,2 @@
+db_tools.txt
+db_cmds.log
--- a/tests/tooldb/db-g10-offset/README
+++ b/tests/tooldb/db-g10-offset/README
@@ -1,0 +1,10 @@
+Verify that tool offsets set with G10 L1/L10/L11 are persisted when an
+[EMCIO]DB_PROGRAM is configured.
+
+In db mode with a nonrandom toolchanger no tool table file is written, so the
+db program must be notified of the new offsets ('p' command).  Without that
+notification the offsets only live in memory and are lost on the next start.
+
+test.ini applies the offsets (one tool per G10 variant, and one of them while
+the tool is in the spindle), then verify.ini restarts LinuxCNC and checks that
+the tool data handed back by the db program still carries them.
--- a/tests/tooldb/db-g10-offset/base.inc
+++ b/tests/tooldb/db-g10-offset/base.inc
@@ -1,0 +1,98 @@
+[EMC]
+# The version string for this INI file.
+VERSION = 1.1
+DEBUG = 0x0
+
+[RS274NGC]
+PARAMETER_FILE = sim.var
+
+[EMCMOT]
+EMCMOT = motmod
+COMM_TIMEOUT = 4.0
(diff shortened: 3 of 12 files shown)


tests/tooldb/db-g10-offset/.gitignore
```diff
@@ -0,0 +1,3 @@
+sim.var*
```
BsAtHome: The `sim.var` and `sim.var.bak` are covered by `tests/.gitignore`. Is there a reason to have it here?

tests/tooldb/db-g10-offset/base.inc
```diff
@@ -0,0 +1,107 @@
...
+SERVO_PERIOD = 1000000
+
+[TASK]
+TASK = milltask
+CYCLE_TIME = 0.001
+
+[HAL]
+HALUI = halui
+HALFILE = core_sim.hal
+
```
BsAtHome: You should rename `core_sim.hal` to something that is not available from the default library. We've seen problems before. This is a contained test, so no reason not to use a local more relevant name.

If possible, to limit the copy-[pw]aste, you should use the library version as-is and add your own hal-file with local additions here.

alex-pres:
Done

tests/tooldb/db-g10-offset/README
```diff
@@ -0,0 +1,6 @@
+Verify that tool offsets set with G10 L1/L10/L11 are persisted when an
```
grandixximo, line 1: Nit (non-blocking): the README and PR body say L1/L10/L11, but the test only exercises L1. Same code path, but L10/L11 cases would be cheap to add.

grandixximo (commented):
Verified against the tree, this looks right to me.

Some history on the commented-out block: it was active code in the old ioControl.cc and was commented out (with the stale `io_db_mode` name) by 764655eb4d "moving all IO handling from iocontrol to task" during the taskio migration. So there was no deliberate reason to disable it, it was a casualty of that move. And the bug is real: in db mode with a nonrandom toolchanger `tooldata_save()` returns early without writing anything (tooldata_common.cc), so offsets set with G10 were lost at restart without this notification.

The pno logic matches the pre-move code, `db_mode` is in scope, and the 'p' notification is synchronous (send_and_verify), so the `wait_complete()` ordering in the test is sound. The test covers restart persistence, the loaded-tool case, and survival of untouched tools. Nice work.

One non-blocking nit inline.

grandixximo (approved):

alex-pres:
Done. Added tests for L10/L11 and squashed into one commit to keep history clean.

grandixximo:
@alex-pres Do you actually use this?

alex-pres:
> @alex-pres Do you actually use this?

Yes, I do .  And it what was suggested to me here https://github.com/LinuxCNC/linuxcnc/issues/4408

Although I still believe tool diameter and tooth count [ + may be tool grade ] must be system-global and standardised across LinuxCNC because every GUI will win from that.

PR #3781, opened 2026-02-02
Documentation - add diagram "Architecture - Context diagram"

zz912:
More information: https://github.com/LinuxCNC/linuxcnc/issues/3738

PR #4423, opened 2026-08-19
docs: Architecture C1 context diagram - halcmd and halshow attach to HAL

Zia-research:
@zz912 — the separate PR for the context diagram I promised in #3718, and only for it.
It carries the correction you accepted there.

## What changes

The Terminal box holds `halcmd`, `halshow` and `linuxcncrsh` and draws one link, to the
Core Runtime.

The box is split in two inside its own footprint: `linuxcncrsh` keeps the link to the
core, `halcmd / halshow` gain a dashed link to the HAL Interface Boundary. The
*Developers / Testers* link is drawn to both halves, so the split does not narrow it to
`linuxcncrsh` alone.

## Whose file this is

The published SVG is not an export of the `.drawio` beside it. It is the file
**@BsAtHome** rebuilt by hand in Inkscape and offered in #4057, which **@grandixximo**
merged there.

The second commit puts the same split in the `.drawio`, so the source does not contradict
the figure. The two have already drifted since #4057: the SVG reads
`(joints, spindles, IO, safety)`, `PNConf` and full stops where the source reads
`(axes, spindle, IO, safety)`, `PNCconf` and middle dots. None of the three is touched
here — but a re-export would revert all of them, so it is worth saying which file is
authoritative from now on.

### DIFF
--- a/docs/src/getting-started/images/LCNC_Architecture_C1.drawio
+++ b/docs/src/getting-started/images/LCNC_Architecture_C1.drawio
@@ -256,13 +256,28 @@
             <mxPoint x="1100" y="500" as="sourcePoint" />
           </mxGeometry>
         </mxCell>
-        <mxCell id="XWJ5TldJVoNbO8orDr2t-41" parent="1" style="rounded=1;whiteSpace=wrap;html=1;fillColor=#dae8fc;strokeColor=#6c8ebf;dashed=1;" value="Terminal&lt;br&gt;halcmd / halshow / linuxcncrsh&amp;nbsp;" vertex="1">
-          <mxGeometry height="80" width="200" x="1600" y="240" as="geometry" />
+        <mxCell id="XWJ5TldJVoNbO8orDr2t-41" parent="1" style="rounded=1;whiteSpace=wrap;html=1;fillColor=#dae8fc;strokeColor=#6c8ebf;dashed=1;" value="Terminal&lt;br&gt;linuxcncrsh&amp;nbsp;" vertex="1">
+          <mxGeometry height="35" width="200" x="1600" y="235" as="geometry" />
+        </mxCell>
+        <mxCell id="XWJ5TldJVoNbO8orDr2t-49" parent="1" style="rounded=1;whiteSpace=wrap;html=1;fillColor=#dae8fc;strokeColor=#6c8ebf;dashed=1;" value="Terminal&lt;br&gt;halcmd / halshow&amp;nbsp;" vertex="1">
+          <mxGeometry height="38" width="200" x="1600" y="282" as="geometry" />
+        </mxCell>
+        <mxCell id="XWJ5TldJVoNbO8orDr2t-50" edge="1" parent="1" source="XWJ5TldJVoNbO8orDr2t-49" style="edgeStyle=orthogonalEdgeStyle;rounded=0;orthogonalLoop=1;jettySize=auto;html=1;entryX=1;entryY=0.5;entryDx=0;entryDy=0;startArrow=classic;startFill=1;dashed=1;exitX=0.5;exitY=1;exitDx=0;exitDy=0;" target="XWJ5TldJVoNbO8orDr2t-4">
+          <mxGeometry relative="1" as="geometry" />
+        </mxCell>
+        <mxCell id="XWJ5TldJVoNbO8orDr2t-51" edge="1" parent="1" source="XWJ5TldJVoNbO8orDr2t-29" style="edgeStyle=orthogonalEdgeStyle;rounded=0;orthogonalLoop=1;jettySize=auto;html=1;dashed=1;startArrow=classic;startFill=1;exitX=0.65;exitY=1;exitDx=0;exitDy=0;exitPerimeter=0;entryX=1;entryY=0.5;entryDx=0;entryDy=0;" target="XWJ5TldJVoNbO8orDr2t-49">
+          <mxGeometry relative="1" as="geometry">
+            <Array as="points">
+              <mxPoint x="1450" y="160" />
+              <mxPoint x="1860" y="160" />
+              <mxPoint x="1860" y="301" />
+            </Array>
+          </mxGeometry>
         </mxCell>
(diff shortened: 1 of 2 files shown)


BsAtHome:
This change is trivial and unnecessary IMO. The change does not reflect the actual situation with respect to the rest of the graph.

To exemplify: the box "LinuxCNC Core Runtime" encapsulates a combination of non-realtime and realtime. For example, the motion controller is clearly in RT and is actually handled under a HAL thread/function. But still, it is not drawn in the HAL interface boundary because this image does not reflect the real ownership paths. It represents logical communication paths. In that context, the "LinuxCNC Core Runtime" may be seen as a much wider container than the technical reality.

This drawing is targeted as an introduction and not as a development aid. Therefore, adding more details is not necessary a good thing and may even be counter productive. I'd leave the drawing as-is.

And for the drawio/SVG drift. I think that the drawio version should be removed. It turns out that it is terrible at generating SVG. Much easier to use Inkscape and some hand-editing.

zz912:
I admit that the benefit of this PR is minimal. On the other hand, I would not spoil Zia-research's joy from the benefit to this community. His PR is correct.

linuxcncrsh and halcmd was added by smoe ask:
https://github.com/LinuxCNC/linuxcnc/pull/3781#issuecomment-3867216649

I have always struggled with Inkscape. If somebody wants to make changes to this image by only Inkscape, then drawio would be deleted.

grandixximo:
I land between the two positions. @BsAtHome is right that the diagram shows logical communication paths, not technical ownership. But by that same standard the current Terminal link is wrong: halcmd and halshow communicate with HAL shared memory only, they have no path to the core runtime, logically or technically. The motion-controller example is container placement; this is a nonexistent link. So I read this PR not as adding technical detail to a logical diagram, but as fixing a factual error at the diagram's existing abstraction level. The added visual load is one dashed line, which seems within budget for a C1.

That said, the benefit is small, and I would not object to either outcome. What I think does deserve attention regardless is the drift question this PR surfaced: the .drawio and the published SVG already disagree in three places, and any future re-export silently reverts the Inkscape fixes. I support removing the .drawio and keeping the hand-maintained SVG as the single source. Separately, the figure should read PNCconf everywhere, one of the two files has PNConf, which is simply a typo.

Zia-research:
Thanks @zz912, @BsAtHome, @grandixximo — appreciate the time each of you put into this.

Happy to make any change you'd like: say the word and I'll push it. That includes the two points raised here — dropping the `.drawio` and keeping the hand-maintained SVG as the single source, and `PNConf` → `PNCconf`. In this PR or a separate one, whichever you prefer.

And if the conclusion is that it isn't worth merging, that's genuinely fine by me — no ego attached. It's the maintainers' call.

grandixximo:
A separate PR to clean drawio, and fix typo will merge quick separately, this one we probably needs some more persuasion, if you fold in is fine, just will sit here for a bit longer.

One question that might unblock this....
@BsAtHome do you want to shape a mental model for new comers, where halcmd talks with interpreter?

BsAtHome:
But when we are taking about what is talking to what an subdivisions, then, f.ex., halshow is not a terminal program and misplaced there too. Halrmt is missing as well. Several GUIs are not named either.

But more damning, if we were to go into superficial detail, is the fact that linuxcncrsh/halcmd/halshow only connects to "Developers/Testers" and not to "Integrator" nor "Machine operator". Then, halcmd (and halrmt) is the interface to control HAL, which should connect to both "HAL Configuration" and "HAL Modules". After all, halcmd is designed to read HAL-files and "execute" the commands to build the RT content at runtime (not to be mistaken with the realtime script).
The motion controller is a RT component that is in the realm of HAL, yet it is now somehow part of the "LinuxCNC Core Runtime", but it is actually the interface layer between "HAL Modules" and the "LinuxCNC Core Runtime" (lets not start about the TP and kinematics).
The "INI Configuration" is led to believe to play a part in the "HAL Interface Boundary", which it does not. It is indirectly involved only because HAL-files can do INI-expansion via halcmd and halrmt. So that path is misrepresented too.

So, which way should we go?...

grandixximo:
Good catalog, @BsAtHome, and I think it points to an answer rather than away from one. The items are not all the same kind:

- Missing halrmt, unnamed GUIs: legitimate C1 elision. Box contents are examples, nobody reads them as exhaustive.
- Terminal linking only to Developers/Testers, halcmd not linking HAL Config/Modules: granularity choices, fine at intro level.
- halshow under "Terminal", PNConf: flat mislabels, one-word fixes.
- halcmd => Core Runtime, INI Configuration as a direct edge of the HAL boundary: false paths by the diagram's own semantics.

So "which way should we go" need **not** be coarse-everything or redraw-everything. State the rule once ("links are logical communication paths, box contents are representative examples"), then fix only the last two categories and leave elision and granularity alone. That keeps the C1 an introduction while making every drawn link true.

Concretely, I would expand this PR's scope rather than split it: the Terminal split, the halshow label, the INI edge, PNCconf, and the .drawio removal all touch the same figure, and once the .drawio is gone they are all edits to one hand-maintained SVG. One errata pass, one review, done. @Zia-research already offered to push whatever we agree on; @zz912, @BsAtHome, does that scope work?

BsAtHome:
This is my take on the architecture diagram:
LinuxCNC-architecture

zz912:
> This is my take on the architecture 

I like it.

grandixximo:
One addition worth the empty space under Terminal: the HAL inspection tools. halshow, halmeter and halscope attach directly to HAL in userspace, no NML, no core runtime path. They are basic enough that an integrator should see how they interact even at this level. Not under the Terminal label though, halshow is a Tk program, not a terminal one. A small "HAL Tools (halshow, halmeter, halscope)" box in that gap with a single edge down to the HAL Interface Boundary would cover it.

BsAtHome:
Agreed, update:

LinuxCNC-architecture

grandixximo:
Replaced by #4437 
@Zia-research thank you for spurring the motivation for the better graph, will be closing this, don't feel discouraged, you were great help actually :-)

Zia-research:
Thanks @grandixximo — no discouragement here at all, this is the right outcome.

The merged figure is better than what my PR proposed, and it closes the three things this thread had open

Happy to learn!!

PR #4411, opened 2026-08-19
posemath: split the header, move EmcPose, stop exporting gomath and sincos

grandixximo:
C, D and E of the plan agreed in #4387. Each commit stands alone.

**Split the header by language.** posemath.h declared the C++ classes and the C API in one file. The tree already has a shape for that, inifile.h/.hh and hal.h/.hh, so follow it: posemath.h keeps the C API and gains posemath_types.h for the types and constants, posemath.hh takes the classes, operators and copy templates and refuses to compile as C. posemath.h ends by including posemath.hh under __cplusplus, so C++ code including posemath.h still gets the classes and nothing out of tree changes. The gain is for C++ files wanting no more than PmCartesian, which can include posemath_types.h instead of four hundred lines of class declarations.

Declaration text moves verbatim. The one edit is PmAxis, an unnamed enum in C and a typedef of the C++ enum PM_AXIS in C++; the types come first now, so the enum takes its tag and one definition serves both.

**Move EmcPose.** emcpos.h held the type and emcpose.h the operations, and code wanting the type included the one that could not stand alone. Neither is an NML message. Both fold into src/libposemath/emcpose.h and emcpose.c moves with them. emcpos.h stays and includes the new header, so the 126 files that include it are untouched. The declarations gain an extern "C" block they never had.

**Stop exporting gomath.h, gotypes.h and sincos.h.** Five in-tree users between them, all kinematics, which include them by path instead. What remains installed is posemath.h plus its two headers, plus emcpose.h.

**Testing.** Build clean, no errors or warnings. runtests over blendmath, realtime-math, interp, ccomp and matrixkins: 91 run, 91 successful, 0 failed. All ten kinematics modules that link a posemath object load under halrun. Preprocessing posemath.h before and after, in C and in C++, gives the same declarations apart from that PmAxis line. libposemath.so.0 gains the fifteen emcPose entry points and liblinuxcnc.a loses them, and nothing in userspace calls them. Installed headers go from 38 to 37.

tpcomp.comp is filtered out of COMPS so nothing in tree compiles it, and it references emcpose.c by path; checked separately by setting TOPDIR and running halcompile on a copy.

PR #4425, opened 2026-08-20
untangle: break the src/ include cycle into three

grandixximo:
Following the dependency report in #4411. The tree had one include cycle covering 17 of its 32 directories. This takes it to three cycles over nine, and none of what is left is a stray include.

Four changes, each standing alone, each built clean on its own.

**`rtapi: drop the unused hal.h include`.** `uspace_rtapi_main.cc` includes `<hal.h>` and uses nothing declared in it: the one HAL entry point it reaches is resolved by `dlsym()` through a typedef written beside the call. That line was the whole of rtapi's dependency on hal, and removing it splits the 17 directories into 10 and 5.

**`hal: move setps_util beside the code that links it`.** `hal/Submakefile` already compiled `hal/utils/setps_util.c` into the python module, so `hal/` was reaching down into `hal/utils/` for a source file and a header. It moves up to `hal/`, its three users in `hal/utils/` name the new path, and the same object goes into the same four targets. Cycle down to 8.

**`emc: move the interpreter and motion interface types into nml_intf`.** `state_tag_t` is filled in by the interpreter and carried in motion messages and in `TC_STRUCT`; its own header says the info "isn't actually used by motion directly". `StateTag` is its C++ form, embedded in `canon.hh`, `emc.hh` and `emc_nml.hh`, and `emc/nml_intf/Submakefile` already compiled `modal_state.cc` from across the tree. Both move to `emc/nml_intf`.

`ACTIVE_G_CODES`, `ACTIVE_M_CODES` and `ACTIVE_SETTINGS` move with them into `emc/nml_intf/interp_codes.h`, because `emc_nml.hh` was including `rs274ngc.hh`, and so the whole interpreter class hierarchy, for three array sizes. Two includes go too: `modal_state.cc` included `interp_base.hh` and used nothing from it, and `emcmodule.cc` was reaching `debugflags.h` through the `rs274ngc.hh` chain and now names it.

**`emc: move emcmotcfg.h to the interface directory`.** No source changes at all: it is exported and every user takes it as `<emcmotcfg.h>` out of `include/`. It was the only reason `emc/kinematics` (9 includes), `emc/nml_intf` (3) and `emc/ini` (1) depended on `emc/motion`. `emc/motion/Submakefile` existed only to copy it into `include/`, so it goes and `emc/motion` comes off `SUBDIRS` with it.

The fifth commit adds `scripts/include-dep-report.py`, which is what the other four were measured with. It resolves each file under the rules of the compile it actually goes through, since userspace and realtime do not get the same `-I` and `.comp` files get a third set, and it reports the cycles, the cycle edges that are a single include, any include that would not resolve in realtime, the headers reached by both compiles, and the exported headers with no user outside their own directory.

**What is left.** Three cycles over nine directories: libnml's five among themselves, `emc/nml_intf` with `emc/tooldata`, and `emc/motion` with `emc/tp`. Each is a question about where an interface belongs rather than a stray include, so none of them is in here.

**Testing.** Configured `--with-realtime=uspace`. Every commit builds clean on its own, no errors and no warnings. Full `runtests` twice: 297 run, 293 successful, 0 shmem errors each time. `pyvcp` and `tooledit` fail in both, and fail the same way on a branch without any of this work. The other failures differ between the two runs, `hal-show` and `ui-smoke/gmoccapy` in the first and `raster` and `ui-smoke/qtdragon-quit` in the second, so they are flaky here rather than caused by this. `hal-show` in particular passes on its own and passed in the second full run; its one line of difference is a HAL component owner id, and nothing here touches HAL object creation.

Parallel to #4411 rather than stacked on it; the two merge clean.


src/emc/nml_intf/emcops.cc
```diff
@@ -15,6 +15,7 @@
 * Last change:
 ********************************************************************/
 
+#include <stdio.h>
 #include "emc.hh"
```
BsAtHome: I cannot see why stdio is needed in _this_ file. The header does not seen to include anything that comes from stdio, or am I missing something?

Or does stdio become necessary for one of the other headers? But then it would be obvious to include it there. This will influence the inclusion of stdio several places, probably.

grandixximo: Line 225, the only one in the file:

```c
fprintf(stderr,"UNEXPECTED idx %s %d\n",__FILE__,__LINE__);
```

You are asking the better question though, so I answered that instead. `emcops.cc` reports through `rcs_print_error()` now, the way `interpl.cc:37` already does in that directory, and its `<stdio.h>` is gone again. `emccanon.cc:3706` open coded the same line and follows it. Neither costs a dependency, still 136 edges and the same three cycles. And no header was feeding those includes: everything a grep for stdio finds in `emc.hh`, `rcs_print.hh`, `rtapi.h` and `hal.h` is `__attribute__((format(printf,1,2)))`.

Your header case is in as well, since `stdio.h` costs no edge: `rs274ngc_interp.hh:648`, `hal/classicladder/files.h:30` and `hal/utils/scope_usr.h:261` name `FILE *` and now say so. It does not make `files.h` standalone, its only guard is around the macro block and it uses `StrRung` without including `classicladder.h`, but that is classicladder's problem rather than an include path one.

The rest is a question for you. The same line is open coded at five more sites and `UNEXPECTED_MSG` is defined locally in six files, across `usr_intf`, `rs274ngc`, `sai` and `tooldata`. `emc/rs274ngc` and `emc/tooldata` include nothing from `libnml/rcs`, so `rcs_print_error()` would cost them a directory edge each. Both already reach rtapi, and `rtapi_print_msg()` is used from userspace in `halui.cc` and `taskclass.cc`, so that would cost nothing. One shared `rtapi_print_msg()` for the lot, or `rcs_print_error()` and accept the two edges?

BsAtHome: Right, now we are touching a new realm of problems, printing message.

I think we need to handle that discussion separate from this cleanup and revisit cleanup details when the message problem is "solved".

This PR will get us a significant improvement. It would be unwise to try fix it all in one go.

grandixximo:
Agreed, parking it. What is in here is two call sites rather than a policy, and if the messaging discussion lands elsewhere they are cheap to move. The other five open coded sites and the six local `UNEXPECTED_MSG` copies stay untouched.

On the count: eight is more than you usually want in one PR. I kept them apart because each builds on its own, but I am happy to fold. The smallest useful shape is six, the two script commits into one and the two moves into `nml_intf` into one. Say if you want that, or fewer, and I will force push it.

PR #4396, opened 2026-08-17
mesa_modbus: mark as superseded by hm2_modbus (2.9 backport)

grandixximo:
Mark mesa_modbus as superseded by hm2_modbus (2.9 backport)

Backport of #4386 with status adjusted for 2.9, per the review discussions: in 2.9 mesa_modbus is superseded, not deprecated. It remains available throughout the 2.9 series, while new configurations are encouraged to migrate to the hm2_modbus driver with mesambccc-compiled MBCCB files.

Notes added:

- modcompile.py: stderr note at build time
- README.adoc in the modbus source directory
- docs: NOTE in the mesa_modbus driver page, (SUPERSEDED) in the modcompile(1) man page, label in the docs index

No load-time warning on purpose: the driver is still fully supported in 2.9, so a build-time nudge suffices.

No functional changes; existing configurations keep working unchanged.

PR #4427, opened 2026-08-20
Merge branch '2.9' into master

grandixximo:
Merge 2.9 into master. Net new content:

- #4385: rs274 gets its own tool mmap instead of truncating `$HOME/.tool.mmap`
- #4413: improved SITEPY detection (works on python 3.9+, dist-packages fallback; master's `get_default_scheme()` fallback is wrong for Debian pythons without `deb_system`)

Everything else on 2.9 was already in master (NTULINUX TCLLIBPATH fix 78da77a9da, #4160 backport #4161, hm2_modbus doc fix #4380, mesa_modbus deprecation #4396), so conflicts there were resolved keeping master's side. `tooldata_mmap.cc` was resolved manually keeping master's NULL style.

Full RIP build and configure verified.


BsAtHome:
This should be squashed, I guess. Or at least squashed when merged.

grandixximo:
Squash works content-wise, but it drops the ancestry link. With a real merge, 2.9's tip becomes a common ancestor and the next 2.9=>master merge-up only brings new commits. With a squash, git never learns these 2.9 commits were integrated, so the same conflicts (mesa_modbus wording, the #4160 backport overlaps, the SITEPY line) resurface on every future merge-up and need manual resolution again.

No need to mandate one flow for everything: individual fixes can keep going master-first with backports as before. But when we do use a 2.9=>master merge-up like this one, it only works without squashing. The merge commit is what makes the flow repeatable.

BsAtHome:
Right, no more comments then.

grandixximo:
github shows a mess, but it lands clean, see here
image

PR #4431, opened 2026-08-21
5axiskins, maxkins: take the tool length through a pin like every other rotary module

grandixximo:
## Summary

Every trt module takes the tool length through a `.tool-offset` pin and turns it with the rotaries. `5axiskins` and `maxkins` have no such pin, so canon adds the offset along machine Z and there it stays: with the head tilted the tip misses the programmed point by `L sin(B)` in X and `L (1 - cos(B))` in Z, the full tool length at B90.

Both modules get the pin, folded into the pivot to tip distance with the same term taken back out of the joint Z, so what canon added along Z is moved onto the tool rather than counted twice.

## The two modules differ in what this fixes

**`maxkins` had no route at all** for a Z-column tool length. `max5kins.hal` sets `pivot-length` with a static `setp` and sends `motion.tooloffset.w` to the vismach display, which reads zero on a config with six joints and no W axis. That is why line 18 was a commented-out `setp max5gui.tool-length 35`: the model drew no tool. It now reads `motion.tooloffset.z`, the pin `max5triv.hal` already uses, and feeds the module as well as the display.

**`5axiskins` already had one**, spelled by hand. `5axisgui.hal` loads a `sum2` and adds the tool length into `pivot-length`, with a comment saying so. That arithmetic is identical to a real pin, so this is a cleanup: the sim keeps both quantities and stops routing one through a realtime component to reach the module that wanted it.

There is a third route on both, undocumented until now: W is already in the pivot to tip distance, so a length in the W column reaches the kinematics once a block commands W. All three add, so a length belongs in one of them. `kins.9.adoc` now says so, and documents `pivot-length` and `tool-offset` for both modules.

## Compatibility

`tool_offset` defaults to 0 and every term it enters is additive, so a config that does not connect the pin computes exactly what it computed before.

## Test

Live on both sims, headless.

`5axiskins`, bridge mill, `G43.1 Z5` then `G0 X10 Y20 Z-30 B0 C0` then `G0 B40 C25`. The pin reproduces the `sum2` config to the digit:

|  | joints |
|---|---|
| `sum2` into `pivot-length`, old module | `-138.553671 -49.271714 -84.658667` |
| `tool-offset` pin, new module | `-138.553671 -49.271714 -84.658667` |

`maxkins`, max5, `G43.1 Z5` then `G0 X10 Y20 Z-30 B0 C0` then `G0 B40`:

|  | joint X | joint Y | joint Z |
|---|---|---|---|
| before | 10.000000 | 20.000000 | -25.000000 |
| after | 6.786062 | 20.000000 | -26.169778 |
| hand calc | `10 - 5 sin40` | 20 | `-25 + 5 cos40 - 5` |

At B0, and with no tool loaded, both are unchanged.

### DIFF
--- a/configs/sim/axis/vismach/5axis/bridgemill/5axisgui.hal
+++ b/configs/sim/axis/vismach/5axis/bridgemill/5axisgui.hal
@@ -13,20 +13,12 @@ net :jy2 joint.6.pos-fb
 # Note: no joint or motor connections are used
 # for the w coordinate
 
-# compute pivot-length needed for 5axiskins
-# according to gui pivot_len and tool z offset
-loadrt sum2 names=pivotsum
-addf pivotsum servo-thread
-
 net :gui-pivot-len <= 5axisgui.pivot_len
-net :gui-pivot-len => pivotsum.in0
+net :gui-pivot-len => 5axiskins.pivot-length
 
 net :tool-len <= motion.tooloffset.z
-net :tool-len => pivotsum.in1
+net :tool-len => 5axiskins.tool-length
 net :tool-len => 5axisgui.tool_length
-
-net :pivot-len <= pivotsum.out
-net :pivot-len => 5axiskins.pivot-length
 
 net :tool-diam <= halui.tool.diameter
 net :tool-diam => 5axisgui.tool_diam
--- a/configs/sim/axis/vismach/5axis/max5/max5kins.hal
+++ b/configs/sim/axis/vismach/5axis/max5/max5kins.hal
@@ -15,9 +15,10 @@ loadrt not count=5
 loadusr -W max5gui
 
 # set a visible tool
-#setp max5gui.tool-length 35
 setp max5gui.tool-radius 3
-net tool-len motion.tooloffset.w max5gui.tool-length
+# the tool length is applied along the tool, not along Z, so the tip stays
+# on the programmed point as B tilts
+net tool-len motion.tooloffset.z max5gui.tool-length maxkins.tool-length
 
 # add motion controller functions to servo thread
 addf motion-command-handler servo-thread
--- a/docs/src/man/man9/kins.9.adoc
+++ b/docs/src/man/man9/kins.9.adoc
@@ -392,6 +392,15 @@ Note: These kinematics may be used with the vismach 5axisgui providing
 that the joint-letter assignments agree with the default ordering
 expected by it (XYZBCW `->` joints 0..5)
 
+*5axiskins.pivot-length*::
+  Distance from the B rotation point to the tool holder gauge line.
+
+*5axiskins.tool-length*::
+  Tool length, applied along the tool rather than along Z, so that the tip
+  stays on the programmed point as B and C move. Net it from
+  *motion.tooloffset.z*. Left unconnected, the tool length stays where canon
+  put it, along machine Z, which is only correct at B0.
+
 == SEE ALSO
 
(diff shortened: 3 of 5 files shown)


Sigma1912:
Isn't  `.tool-offset` a bit vague given that a tool can have 9 offset values (XYZABCUVW)?
Maybe we could make it `.tooloffset-z`  so it reflects what is used in  `motion.tooloffset.z

grandixximo:
Fair, and I went looking for a name rather than picking one.

The tree has no convention to appeal to. `trtfuncs.c`, `genhexkins` and `pentakins` call it `.tool-offset`; `xyzacb_trsrn`, `xyzbca_trsrn` and `xyzab_tdr_kins` call it `.tool-offset-z`. Three each, and the split does not even follow machine class, since trt and trsrn are both serial 5-axis mills. So whichever of the two I pick, these modules end up spelled differently from one of the two families already there.

What decided it for me is that both existing spellings are already contradicted by their own documentation. `genhexkins.c:43` describes the pin as "tool length offset", `pentakins.c:45` as "tool length from the origin along z axis". The configs agree: the vismach pins are `max5gui.tool-length` and `5axisgui.tool_length`, and the signal is `:tool-len`. Everything in tree calls this quantity a tool length except the pin names.

The controls do too. Fanuc's `G43.4 Hn` takes the corrector holding "the length of the tool... measured from spindle nose to tool end". Siemens TRAORI works from the tool length in the tool data. Heidenhain parameterises the head as Pivot Length and Pivot Offset, with the tool as length L, which is also where our own `pivot-length` comes from.

So I have gone with `.tool-length`, sitting beside the `pivot-length` that is already on both modules. It answers your objection better than `-z` does, I think: a length is not one of the nine offsets at all, so there is nothing to be vague about. And it does not claim a direction the value does not have, since the module applies it along the tool rather than along Z.

That does make a third spelling, which I am not pretending is an improvement on its own. What I would suggest is that it is the one to converge on, and that renaming the existing six is separate work with a release of both pins, not something to smuggle into this PR.

One small thing: you wrote `.tooloffset-z`, but your three modules use `.tool-offset-z`. If you would still rather have that spelling than mine, say so and I will switch, since you have more 5-axis modules in tree than anyone.

Sigma1912:
Thanks for the research, `tool-length` is fine for me.

Sigma1912:
> One small thing: you wrote .tooloffset-z, but your three modules use .tool-offset-z. If you would still rather have that spelling than mine, say so and I will switch, since you have more 5-axis modules in tree than anyone.

All the same to me, so you choose.

grandixximo:
Two things a reviewer would otherwise have to find on their own.

5axiskins has had a pin by nearly this name before. `6de7f0db21` (2013-04-09, Chris Radek) removed `5axiskins.tooloffset.z`: "The tool-length hal pin was vestigial and confusing." Reading the code as it stood, he was right. `tool_length` appeared exactly twice, at the struct member and at `hal_pin_float_new`, and neither `kinematicsForward` nor `kinematicsInverse` ever read it. So this is not a rejected idea returning; it is the working version of what was a stub.

I was also wrong above to say converging the existing six would want "a release of both pins". The project has no such mechanism, and the one deprecation of a shipped pin we do have argues against wanting it: plasmac keeps `requested_velocity` forever, marked "deprecated", and never reads it, so an old `net` loads and the feed rate silently stops arriving. A rename that makes the config refuse to start is the better failure. The pin is gone, the message says so, and the docs say what replaced it. Whenever the six converge, that plus fixing the in-tree configs is the shape I would argue for.

PR #4436, opened 2026-08-21
Gmoccapy Tool/Offset pages: Value verification and tool data checking

### LINKED ISSUE #4435: Gmoccapy: Tool number and pocket are not editable with calculator
When having the calculator enabled, the  entered value for tool number and for pocket isn't applied.

Further I noticed a different behavior for entering values in the offset page and the tool table.
For the tool table, you need to select the line first and then you can select a cell for editing.
For the offset page you can directly select a cell for editing without selecting the line first. It doesn't matter if the calculator is used or not.
@Sigma1912 Do you think we should unify that? If yes I tend to the version used in the offset page.

Sigma1912:
Discussion has shown that silent manipulation to make the user entered data fit the data type required by the cell is not wanted and that a rejected value should trigger a message to the user. 

Value entries by the user to the offset or tool tables are currently 'verified' and manipulated in several places:
- Calculator widget
- entry_dialog
- gmoccapy
- tooledit / offsetpage widgets

This PR moves value verification to the end recipient (ie the tooledit -  and offsetpage widgets). The value is verified in the same place regardless whether the user uses the keyboard directly,  the buttons in the calculator widget  or the keyboard in the calculator widget. The cell and it's data range is known and a clear user message can be generated.

Fixes https://github.com/LinuxCNC/linuxcnc/issues/4435

Also in this commit:
 - Tool table is checked for duplicate tool numbers on save.

The second commit:
 - The tool number for a new tool is preset to one above the highest number found in the table instead of '0'

The third commit:
- Lines that contain only white space are ignored instead of creating a new row with all '0'
A message is shown on startup if the tool file contains
- duplicated tool numbers 
- orientation values not in range 0...9
- malformed lines (missing descriptors or values)  these lines are copied to a file so they are not silently lost when the table is saved.

Example:
messages


hansu:
That was quick, thanks.

I haven't dug deep into this but I wonder if we should fix it better in tooledit (or additionally).
It fails in line `int(new_text) `with new text = "5.000" for example.

https://github.com/LinuxCNC/linuxcnc/blob/bae5676d037f5f68cb4ed7bb70a5cdd90e1a6f55/lib/python/gladevcp/tooledit_widget.py#L513-L523

This would also fix it in this case. But it might be bad if this function gets accidentally a float value which is cut then.

        if col in(1,2):
            try:
                import re
                self.model[path][col] = int(re.split(r'[.,]', new_text)[0])

Sigma1912:
I noticed that the pocket number column currently allows negative integer values, which I don't _think_ is correct?
So Tool and Pocket number columns will now truncate any float to the absolute integer value. I'm not sure we still need the dot,comma replacement but kept it just in case.

src/emc/usr_intf/gmoccapy/gmoccapy.py
```diff
@@ -2118,7 +2119,10 @@ def on_tool_col_edit_started(self, widget, filtered_path, new_text, col):
             # Clicking on a cell emits 'editing-started' which leads to the evaluation of the text in edit mode.
             # To use the return value of the calculator, it must be pretended that there is no editable (=no edit mode).
             self.widgets.tooledit1.editable = None
-            self.widgets.tooledit1.validate_input(row, f"{value:11.4f}", col)
+            if col in integer_cols:
+                self.widgets.tooledit1.validate_input(row, f"{value:11.0f}", col)
```
hansu: Why not as int? This makes it more clear:
```suggestion
                self.widgets.tooledit1.validate_input(row, f"{value:11d}", col)
```

hansu:
>I noticed that the pocket number column currently allows negative integer values, which I don't think is correct?

I don't know. Maybe one want to disable a tool by setting pocket to -1?
I think the bigger problem is that duplicated tool numbers and pocket numbers are allowed.

grandixximo:
I use duplicate pocket numbers, stacked tools

Sigma1912:
Just checked in Axis gui (if that is anything to go by):
- allows duplicate pocket numbers
- errors out on negative pocket numbers

And yes, there is definitely a use for duplicate pocket number.

hansu:
Okay, but duplicated tool numbers? You would never know which one will be used.

Sigma1912:
> Okay, but duplicated tool numbers?

Yes, those should indeed not be allowed.

grandixximo:
I used programs where the tool ID has to be unique, it is tricky to do it without getting in the way when the user is renumbering tools for whatever reason, some thought need be put into that. Although at the final state there should be no duplicates, when user is wanting to do a renumbering, not having any hard blocks (current behavior) allows to have a moment where there are two tools with the same number; the trick is finding a reasonable way to allow renumbering without getting in the way of the renumbering cleanup flow, some kind of message confirmation, same number already in the table, want to swap number? cancel?
Also when adding a tool, it would be nice to have the next available ID be ready for the user.
Also there could be tool tables with duplicates in the wild, a warning that those need be fixed should show up somewhere somehow without nagging the user too much...

lib/python/gladevcp/tooledit_widget.py
```diff
@@ -512,7 +512,7 @@ def validate_input(self, path, new_text, col):
 
         if col in(1,2):
             try:
-                self.model[path][col] = int(new_text)
+                self.model[path][col] = abs(int(float(new_text.replace(',', '.'))))
```
grandixximo: `abs()` silently flips the sign instead of rejecting negative input: typing `-3` is stored as `+3` with no feedback. Since the stated intent is to match Axis, note that Axis rejects negative pocket numbers with an error rather than flipping the sign. A `>= 0` check with reject would match that behavior. This widget is shared with axis, gscreen and qtdragon, so the sign-flip applies to direct cell edits there too. Also, hansu's point about pocket = -1 to disable a tool is still open in the discussion; `abs()` would make that impossible to express.

grandixximo (approved):

lib/python/gladevcp/tooledit_widget.py
grandixximo, line 583: This did not get the same float-tolerant treatment: it still uses plain int(new_text). Direct cell entry of 3.0 is now accepted for tool/pocket but still silently rejected for orientation. Same int(float(...)) pattern, keeping the existing range check, would unify it.

grandixximo (approved):

lib/python/gladevcp/tooledit_widget.py
hansu: >I can't say much about the use of pockets as I don't use a tool changer. But A valid point is to reject negative values rather than inverting them.

lib/python/gladevcp/tooledit_widget.py
hansu: What are the lines you are referring to? I think the above lines not.

hansu:
> I used programs where the tool ID has to be unique, it is tricky to do it without getting in the way when the user is renumbering tools for whatever reason, some thought need be put into that. Although at the final state there should be no duplicates, when user is wanting to do a renumbering, not having any hard blocks (current behavior) allows to have a moment where there are two tools with the same number; the trick is finding a reasonable way to allow renumbering without getting in the way of the renumbering cleanup flow, some kind of message confirmation, same number already in the table, want to swap number? cancel? Also when adding a tool, it would be nice to have the next available ID be ready for the user. Also there could be tool tables with duplicates in the wild, a warning that those need be fixed should show up somewhere somehow without nagging the user too much...

I could imagine to do the check on save. So it's easy to rename the IDs and have duplicate IDs in between.

lib/python/gladevcp/tooledit_widget.py
grandixximo: 525-531

grandixximo: GitHub displays it weird because is not in the diff

Sigma1912:
Generally I cannot really comment since I don't use the tool table much.  
Maybe we also shouldn't silently truncate a float entry to integer either.

Check on save seems the most straight forward to me but we would probably also have to check on load since a tool table can easily be edited outside the gui.


> This widget is shared with axis, gscreen and qtdragon

I thought this was only shared by gscreen.

lib/python/gladevcp/tooledit_widget.py
grandixximo: Correction on my earlier claim: the gladevcp ToolEdit widget is embedded only in gscreen and gmoccapy. Axis and qtvcp launch the standalone Tcl tooledit, which is a separate codebase, so this change does not affect them.

On rejecting vs truncating: refusing to accept wrongly formatted numbers seems fine to me, they might be genuine typos and a reject makes the user look twice. Silent truncation or sign-flip hides that.

zz912:
As a tester, I don't think silently truncating a decimal value is the right behavior here.

If the Tool Number or Pocket Number field expects an integer and a user accidentally enters, for example, `5.7`, I would expect the application to report a validation error rather than silently changing the value to `5`.

From a testing/UX perspective, silently modifying the user's input can hide an input error and potentially lead to a different tool or pocket being used than the operator intended.

In my opinion, it would be safer and clearer to reject the value and tell the user that an integer is required. The user can then correct the input explicitly.

Sigma1912:
I agree that silent changes to the values entered is not the right thing to do so I made some deeper changes.

1. The receiver of the data (eg the offsetpage_widget or the  tooledit_widget) is responsible for checking the entered values and will now inform the user if an entry is rejected.
2. The calculator no longer tries to shape the return values to what the receiver might want.
3. Offset editing using the calculator is now handled by the same code as when editing with the keyboard.

 Testers welcome.

lib/python/gladevcp/calculatorwidget.py
```diff
@@ -164,13 +164,15 @@ def set_value( self, value ):
...
-            if self.use_localization:
-                return locale.atof( value )
-            else:
-                return float(value)
+            value = float(self.entry.get_text())
         except:
             return None
+        if self.use_localization:
+            try:
+                value = locale.atof( value )
```
zz912: Just a small reminder regarding the use of `locale.atof()` — perhaps it simply got overlooked in the context of the other changes. 🙂

We have spent quite a bit of time dealing with `atof()` and decimal separator localization in LCNC, and we even have a recommendation about this in the documentation: https://linuxcnc.org/docs/devel/html/en/gui/gui-dev-reference.html#_localization_of_float_numbers_in_guis

`atof()` can of course work correctly in some situations, but in combination with locales I find it rather tricky and potentially a bit of a time bomb — especially because the problem may not show up under the usual EN locale.

If the intention here is to keep using `locale.atof()`, I would at least ask for it to be tested with the DE locale, where the decimal separator is a comma. That is where I would expect any potential problems to show up first.

Please don't take this as a blocker for the change, but rather as a reminder based on the experience we've already had with this in LCNC.

Sigma1912: That is current code that got moved around not something new. That is there for backwards compatibility aimed at Gscreen only not something that Gmoccapy uses.

zz912: Yes, I understand that this is existing code being moved around and that it is only there for Gscreen backwards compatibility.

The reason I mentioned it is precisely because this is how we ended up with problems in the past. `atof()` had been working without any apparent issues for a long time, and then some surrounding code changed and suddenly the locale-related problems started showing up.

So my concern is not that `atof()` is being introduced here as something new, but rather that moving it around can potentially change the context in which it runs. That's why I would still consider a quick test with the DE locale worthwhile, just to make sure the existing behaviour remains safe after the change.

If that has already been covered by the existing Gscreen compatibility testing, then of course that's fine.

Sigma1912:
> Also when adding a tool, it would be nice to have the next available ID be ready for the user.
Also there could be tool tables with duplicates in the wild, a warning that those need be fixed should show up somewhere somehow without nagging the user too much...

Both implemented.

gmoccapy:
Sorry for jumping in late.
I was surprised reading that it was mentioned duplicate tool and pocket numbers may be needed.

I can not imagine a situation where the same tool or pocket numbers may be used in one tool table. The G-code does name the tool number, the "change code" looks in the tool table for that tool and gets that way the corresponding pocket number. The mechanical tool changer doesn't care about the tool number, as it just gets the tool out of the pocket and puts it in the spindle.

If you use a sister tool (i.e. 2 drills 5 mm diameter) in one tool changer the tools should have different tool and pocket numbers. If drill 1 is worn out an dyou want to use drill 2, you will need to edit the tool table. Otherwise, how will you know witch tool will be used? 

So IMHO it should not be possible to have duplicate tool numbers and also not duplicate pocket numbers.


Norbert

hansu:
I could image a use case for duplicate pocket numbers: If you want to use the same tool for two different g-code files where they have different tool numbers, you could just give them the same pocket number.

grandixximo:
Multiple profiling wheels stacked on top of each other, on the same cone, all tools live in the same pocket, multiple tools with different height and diameters for different profiles, this is exactly how I use my edge profiling machines, when I call the tool change for a tool in the same pocket, the tool change routing is skipped, different tool same pocket is easy to use as conditional for skipping tool change, also simple to understand and manage, really don't see why it should be denied as a workflow, if you think is dangerous on some machines, might be worth an optional warning, I don't mind if you default it to warn about duplicate pockets, but removing the functionality altogether I don't think is wise.

lib/python/gladevcp/tooledit_widget.py
```diff
@@ -509,32 +564,40 @@ def validate_input(self, path, new_text, col):
...
-                self.model[path][col] = int(new_text)
+            try :
+                value = float(new_text)
+                if value != int(value) or value < 0:
+                    raise TypeError
+                else:
+                    self.model[path][col] = int(value)
             except:
-                pass
+                msg = (_(f"\nMust be a positive whole number"))
```
hansu, line 575:
```suggestion
                msg = (_(f"\nMust be a positive integer"))
```

hansu:
> > Also when adding a tool, it would be nice to have the next available ID be ready for the user.
> > Also there could be tool tables with duplicates in the wild, a warning that those need be fixed should show up somewhere somehow without nagging the user too much...
> 
> Both implemented.

Thanks.
1. It actually fills the gap of tool numbers. So if you have a tool no. 16 and 18, it creates a tool 17. Would be better if it creates a tool 19 in this case.
2. Can't we use the Gmoccapy dialog for the warnings?
    To be consistent we should also display a warning when entering e.g. 16.0 for the tool number (16,0 gives a warning).

Sigma1912:
> It actually fills the gap of tool numbers. So if you have a tool no. 16 and 18, it creates a tool 17. Would be better if it creates a tool 19 in this case.

Could you elaborate on the reasoning behind this?

> Can't we use the Gmoccapy dialog for the warnings?

Possibly, but why? The point is that the recipient widget verifies and displays the message as to why the entry was rejected.

lib/python/gladevcp/tooledit_widget.py
Sigma1912: Sure, note though that 'integer' is a term out of computer science, not every machine operator might be as familiar with it as we are.

hansu: As far as I know is Integer just the english word for "Ganzzahl". I think only german speaking people can imagine what you mean with "whole number" ;-)

hansu:
> > It actually fills the gap of tool numbers. So if you have a tool no. 16 and 18, it creates a tool 17. Would be better if it creates a tool 19 in this case.
> 
> Could you elaborate on the reasoning behind this?

First it messes up the order. Ok no big thing.
But I use to use not continuous tool numbers.
E.g. 30, 35, 40, 60 and so on. So adding a tool with number 31 would not be helpful. 
But the old version with new tool number = 0 has the advantage that it's easier to see that this is the newly added tool.


> 
> > Can't we use the Gmoccapy dialog for the warnings?
> 
> Possibly, but why? The point is that the recipient widget verifies and displays the message as to why the entry was rejected.

Because the Gmoccapy dialogs are touch optimized. And contribute to a consistent UI. For the exit of the tool table there is also the Gmoccapy dialog used.

lib/python/gladevcp/tooledit_widget.py
Sigma1912: https://www.oxfordlearnersdictionaries.com/definition/english/whole-number

Sigma1912:
> First it messes up the order. Ok no big thing.

Clicking on the header 'Tool#' will sort the table.

> But I use to use not continuous tool numbers.
E.g. 30, 35, 40, 60 and so on. So adding a tool with number 31 would not be helpful.

But in that case '61' would also not help, you would have to change the tool number either way.

> But the old version with new tool number = 0 has the advantage that it's easier to see that this is the newly added tool.

The problem is that there might already be a tool '0' in the table. Also tool number '0' has different usage depending on random/non-random tool changers.

Sigma1912:
> Can't we use the Gmoccapy dialog for the warnings?
To be consistent we should also display a warning when entering e.g. 16.0 for the tool number (16,0 gives a warning).

Ok, I'll have a look.

lib/python/gladevcp/tooledit_widget.py
grandixximo: The average machinist will understand "positive whole number" much faster and more naturally than "positive integer."

"Integer" is an academic term, I'm with Sigma on this one ;-)

Sigma1912:
Fixed the bug where an entry with a zero decimal (eg '16.0') in the calculator did not trigger an error. 

Modified as requested:

- Dialogs use the popups native to gmoccapy (the widgets can fall back to their own ).
- Tool numbers on new tools are preset to one above the highest number found in the table.

lib/python/gladevcp/tooledit_widget.py
hansu: ok

gmoccapy:
> Multiple profiling wheels stacked on top of each other, on the same cone, all tools live in the same pocket, multiple tools with different height and diameters for different profiles, this is exactly how I use my edge profiling machines, when I call the tool change for a tool in the same pocket, the tool change routing is skipped, different tool same pocket is easy to use as conditional for skipping tool change, also simple to understand and manage, really don't see why it should be denied as a workflow, if you think is dangerous on some machines, might be worth an optional warning, I don't mind if you default it to warn about duplicate pockets, but removing the functionality altogether I don't think is wise.

That is a valid usecase. I can imagine similar cases for some tools.

hansu:
> Fixed the bug where an entry with a zero decimal (eg '16.0') in the calculator did not trigger an error.

:+1: 



>* Tool numbers on new tools are preset to one above the highest number found in the table.

 :+1: 


> * Dialogs use the popups native to gmoccapy (the widgets can fall back to their own ).
> 

For this dialog it's not yet a gmoccapy dialog :
grafik

lib/python/gladevcp/calculatorwidget.py
```diff
@@ -164,20 +164,26 @@ def set_value( self, value ):
     def get_value( self ):
         self.compute()
         try:
-            value = self.entry.get_text()
+            value = float(self.entry.get_text())
+            if self.integer_only:
+                if value != int(value):
+                    raise ValueError
+                value = int(value)
             if self.use_localization:
```
hansu, line 172: ` locale.atof( value ) ` doesn't make sense here because localization is already handled in compute() and `value` is already a float at this point (for non integer mode).

lib/python/gladevcp/calculatorwidget.py
```diff
@@ -164,20 +164,26 @@ def set_value( self, value ):
...
+        return value
 
     def get_preset_value( self ):
         return self.preset_value
 
     def compute( self ):
         qualified = ''
         temp = self.eval_string.strip( " " ).replace("Pi", "math.pi")
+        if self.integer_only and ( '.' in temp or ',' in temp):
+            self.entry.set_text( "Error" )
```
hansu, line 185: What speaks against raising the `ValueError` right here instead of later in `get_value`()?

lib/python/gladevcp/tooledit_widget.py
```diff
@@ -502,39 +555,53 @@ def col_editted(self, widget, path, new_text, col, filter):
...
+                msg = (_(f"\nMust be a decimal number"))
         # validate input for orientation: check if int and valid range
         elif col == 15:
-            try:
-                value = int(new_text)
-                if value in range(10):
-                    self.model[path][col] = value
+            try :
+                value = float(new_text)
+                if value != int(value) or value not in range(10):
```
hansu, line 590: Why not  
```python
try :
                value = int(new_text)
```

like above?

If we want to accept inputs like this "1.000000" we could also use <float_type>.is_integer()

It is only a bit inaccurate:
```python
>>> float("1.000000000000001").is_integer()
False
>>> float("1.0000000000000001").is_integer()
True
```

Sigma1912:
> For this dialog it's not yet a gmoccapy dialog :

Should be fixed now. Thanks for testing.
duplicate tool nr

hansu (commented):
And please rebase to master to resolve the conflicts.

lib/python/gladevcp/calculatorwidget.py
hansu: Git messed up the line numbers again, it is about this  part

```python
            value = float(self.entry.get_text())
            if self.integer_only:
                if value != int(value):
                    raise ValueError
                value = int(value)
            if self.use_localization:
                value = locale.atof( value )
```

lib/python/gladevcp/calculatorwidget.py
hansu: Its about those lines:

```python
    def compute( self ):
        qualified = ''
        if self.integer_only and ( '.' in temp or ',' in temp):
            self.entry.set_text( "Error" )
            return
```

Sigma1912:
Alas, there have been other changes pushed directly that actually break this PR.  I don't have the time right now so I'm afraid this will have to wait.

hansu:
After the rebase I get this:

```
  File "/home/cnc/linuxcnc-master/bin/gmoccapy", line 2221, in on_tool_col_edit_started
    self.widgets.tooledit1.validate_input(row, cell_text, col, captations)
  File "/home/cnc/linuxcnc-master/lib/python/gladevcp/tooledit_widget.py", line 606, in validate_input
    self.ext_dialog(msg,header)
  File "/home/cnc/linuxcnc-master/bin/gmoccapy", line 2137, in tooltable_dialog
    self.dialogs.warning_dialog(self, header, message)
TypeError: Dialogs.warning_dialog() takes 1 positional argument but 4 were given
 (gmoccapy:86)

```
The two calls     `self.dialogs.warning_dialog(self, header, message)` :
```python

    def tooltable_dialog(self, message, header=None):
        if header is None:
            header = _("Tool page error:")
        self.dialogs.warning_dialog(self, header, message)
```
```python

    def offsetpage_dialog(self, message, header=None):
        if header is None:
            header = _("Offset page error:")
        self.dialogs.warning_dialog(self, header, message)
```

should probably be          `self.dialogs.show_warning_dialog(self, header, message)`

hansu:
> Alas, there have been other changes pushed directly that actually break this PR. I don't have the time right now so I'm afraid this will have to wait.

It seems Chris did some commits on master accidentally :/

lib/python/gladevcp/tooledit_widget.py
Sigma1912: I'll adopt the first suggestion to make it consistent with the code for columns 1 and 2. 
I think we agreed that we don't want to silently change floats to integers and I have already changed the calculator widget to not accept keyboard input containing decimals.

lib/python/gladevcp/calculatorwidget.py
Sigma1912: The check for decimal symbols in `compute()` is required to disallow keyboard entry of, say, '6.0' in the calculator, which we disallow for direct keyboard entries into the table. 

The check in `get_value()` is required to check the calculator result of 'b = str( eval( qualified ) )' because a user can enter an operation using integers into the calculator using the keyboard (eg '6/2' is allowed but '3/2' must raise an error).

lib/python/gladevcp/calculatorwidget.py
Sigma1912: Agreed.

Sigma1912:
I'm sorry, couldn't recover. See new https://github.com/LinuxCNC/linuxcnc/pull/4559

hansu:
> I'm sorry, couldn't recover. See new #4559

You could have force-pushed your new branch into this one, then this PR would have stayed alive.
But I don't care ;-)

PR #4438, opened 2026-08-21
tree: remove includes that nothing in the file uses

grandixximo:
104 `#include` lines across 82 files, of 30 distinct in-tree headers, that provide nothing the including file mentions. Removals only, no additions and no moves.

This is the sweep referred to in #4411, held back until #4425 landed.

**How they were found.** Take the names each in-tree header declares, macros, typedefs, tags, prototypes, extern variables and enum members, and check whether any of them appears in the file that includes it. That heuristic is going to be wrong somewhere, so nothing rests on it: 209 candidates came out, all 209 were removed at once, the tree was built, and whatever failed had its include put back. What is here is what survived a build with all of them gone together, `--with-realtime=uspace`, no errors and no warnings.

That build does not cover everything. Five of the files are outside what a uspace configuration compiles, so those were read by hand instead, and two keep their include: `rtai_ulapi.c` needs `rtapi_rtai_shm_wrap.h`, which declares nothing itself and only wraps `rtai_shm.h`, and `fifousr.c` takes `FIFO_KEY` and `FIFO_SIZE` from an anonymous enum in its own `common.h`.

**What it buys.** Against current master, `scripts/include-dep-report.py` goes from 137 directory edges and 1287 cross-directory include sites to 131 and 1213.

The cycles do not move: still 3, covering 9 directories and 18 edges. I think that is the honest result and worth stating plainly. Dead includes inflate the picture, they are not what holds the knots together. The three remaining cycles are interface questions, not stray includes.

The ones that repeat are `rtapi_math.h` 21 times, `rtapi_string.h` 15, `config.h` 10, `block.h` 7, `rtapi_slab.h` and `rtapi_mutex.h` 6 each.

### DIFF
--- a/src/emc/canterp/canterp.cc
+++ b/src/emc/canterp/canterp.cc
@@ -55,7 +55,6 @@
 #include <limits.h>
 #include <algorithm>
 #include <linuxcnc.h>
-#include "config.h"
 #include "nml_intf/interp_return.hh"
 #include "nml_intf/canon.hh"
 #include "rs274ngc/interp_base.hh"
--- a/src/emc/ini/inihal.cc
+++ b/src/emc/ini/inihal.cc
@@ -27,7 +27,6 @@ Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #include <rtapi_math.h>
 #include "inihal.hh"
 #include "iniaxis.hh"
-#include "inispindle.hh"
 
 static int debug=0;
 static int comp_id;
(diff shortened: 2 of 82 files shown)


BsAtHome:
But there seems to be a misalignment now for RTAI, that needed a removed include.

grandixximo:
`rip-rtai` caught two, and it caught them because my check could not: the compiler pass ran `--with-realtime=uspace`, so the five files that configuration does not build were never tested. Two of them needed their include, both for a reason a name scan cannot see, a header that declares nothing itself and an anonymous enum.

Restored, the other three read by hand. 104 removals across 82 files now, directory edges unchanged at 131.

PR #4440, opened 2026-08-21
Glnav: Add is_lathe method

### LINKED ISSUE #4426: Vismach: Model rotation is broken
Affects current master.

- Open a sim config with a vismach model. Example: `configs/sim/axis/vismach/3axis-tutorial`
- Try to rotate the 3D model in the vismach window  (middle mouse button + drag ) 

Issue:

The model rotation is broken (it's not possible to control the 3D rotation) and the following error is printed repeatedly:

```
AttributeError: 'O' object has no attribute 'is_lathe'
Exception in Tkinter callback
  File "/usr/lib/python3.12/tkinter/__init__.py", line 1967, in __call__
    return self.func(*args)
           ^^^^^^^^^^^^^^^^
  File "/home/user/git/alex_preview/lib/python/rs274/OpenGLTk.py", line 331, in tkRotateOrTranslate
    self.rotateOrTranslate(event.x, event.y)
  File "/home/user/git/alex_preview/lib/python/glnav.py", line 357, in rotateOrTranslate
    self.rotate(x, y)
  File "/home/user/git/alex_preview/lib/python/glnav.py", line 278, in rotate
    if self.is_lathe():
       ^^^^^^^^^^^^^

```

I have reproduced this issue in https://github.com/alex-pres/linuxcnc/tree/pr-gremlin-opengl-rewrite so it is unrelated to the recent work in 'vismach.py'
 
cc@alex-pres

Sigma1912:
Fixes https://github.com/LinuxCNC/linuxcnc/issues/4426

### DIFF
--- a/lib/python/glnav.py
+++ b/lib/python/glnav.py
@@ -173,6 +173,10 @@ class GlNavBase:
         self._totaly = 0.0
         self.modelview = identity_matrix()
 
+    def is_lathe(self):
+        # preview widgets that need it (gremlin, axis, qt5_graphics) override this
+        return False
+
     def basic_lighting(self):
         """\
         Reset the camera for the first expose.


lib/python/vismach.py
```diff
@@ -1332,9 +1332,8 @@ def basic_lighting(self):
         glEnable(GL_DEPTH_TEST)
         glMatrixMode(GL_MODELVIEW)
         glLoadIdentity()
-  
 
-                
+    def is_lathe(self): return False
```
grandixximo: `glnav.GlNavBase` calls `self.is_lathe()` from `rotate`, `set_view_y` and `set_view_p`, but never defines it. Would `def is_lathe(self): return False` on `GlNavBase` be the better home, with this line dropped? The preview widgets that care (gremlin, axis, qt5_graphics) already override it, and anything else that mixes in the navigation code gets a sane answer instead of an AttributeError. There is precedent in the file: 9c51028c75 ("put more useful defaults in glNavBase") made the same move, with the message "all subclasses need to set or define these, but most will be happy with the defaults." Happy to send that as a follow-up one-liner if you would rather keep this PR as it stands.

grandixximo: Worth one short comment on why this is unconditionally False? A line such as "vismach models never use the lathe preview rotation" saves the next reader a trip into glnav.

grandixximo: For the record, this single method covers all three `is_lathe()` call sites, so nothing is left half-done. `set_view_y` and `set_view_p` still would not work for vismach, but for an unrelated reason: they call `extents_info()`, which only `glcanon.GlCanonDraw` provides and which vismach has no bounding box to build from. Nothing in vismach binds those views, so it stays dormant and is not something this PR needs to touch.

Sigma1912:
Thanks for looking into this. 
I have now moved it into `GlNavBase`.

PR #4448, opened 2026-08-22
src: include exported headers with angle brackets where they are used

grandixximo:
@BsAtHome The include-type half of #4375, split out as you asked. The check stays there.

The build copies every `SRCHEADERS` entry into `include/`, so an exported header exists twice, the source under `src/` and the copy modules compile against. A quoted include searches the includer's own directory before the `-I` path and an angled include does not, so the two forms can reach different copies of the same header and both compile.

An exported header keeps the quoted form so it still finds its siblings once it lands in `include/`, and so does its implementation, which wants the source rather than a stale export. The seven files here are users, and build out of tree where only the exported copies exist.

### DIFF
--- a/src/emc/ini/iniaxis.cc
+++ b/src/emc/ini/iniaxis.cc
@@ -19,7 +19,7 @@
 #include "nml_intf/emcglb.h"
 #include "nml_intf/emccfg.h"
 #include "libnml/rcs/rcs_print.hh"
-#include "inifile.hh"
+#include <inifile.hh>
 
 #include "inihal.hh"
 #include "iniaxis.hh"
--- a/src/emc/ini/inijoint.cc
+++ b/src/emc/ini/inijoint.cc
@@ -17,7 +17,7 @@
 #include "libnml/rcs/rcs_print.hh"
 #include "nml_intf/emcglb.h"
 #include "nml_intf/emccfg.h"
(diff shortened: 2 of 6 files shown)


src/hal/hal.hh
```diff
@@ -4,7 +4,7 @@
 #include <map>
 #include <string>
 #include <variant>
-#include "hal.h"
+#include <hal.h>
 
 #warning "Do not use hal.hh. It will be removed (and, eventually, replaced)."
```
BsAtHome, line 10: This is redundant as it is about to be removed/changed anyway.
(see the #warning)

BsAtHome:
The inifile.hh fixes reminds me of the fact that liblinuxcncini.so should also be moved outside the emc tree. It is a separate library stuff links against. Yet another cleanup ;-)

src/hal/hal.hh
grandixximo: Reverted

BsAtHome: Check.

I now have a PR in the pipeline to remove the old hal.hh. Still working on removal of the pyhal.py module. Both are obsolete and use direct access.

grandixximo: Dropped, thanks. #4449 takes the header out entirely, so there is nothing left to flag.

PR #4455, opened 2026-08-23
kinematics: let a module report where the tool points

grandixximo:
A kinematics module reports the controlled point and nothing about which way the tool faces, so everything that needs the tool frame rebuilds the machine geometry for itself. For `xyzacb-trsrn` the same chain is written three times: closed-form in `xyzacb_trsrn.comp`, as homogeneous matrices in the config's `remap_funcs_twp.py`, and as signed `HalRotate` calls in the vismach model. The only thing keeping them in step is a comment in the Python saying its matrices "must be the same as the ones used to derive the kinematic model".

This adds an optional entry point a module can answer with instead, and a chapter saying what it answers in.

```c
int kinematicsToolFrame(const double *joint,
                        PmRotationMatrix *rot,
                        const KINEMATICS_FORWARD_FLAGS *fflags);
```

The columns are the tool frame axes in world coordinates. Its origin is the controlled point `kinematicsForward()` already gives, so the pair is the whole tool pose.

**Nothing is obliged to change.** `switchkinsSetup()` and `switchkinsRegister()` keep their signatures; a type opts in through a separate call. Nothing in motion calls the entry point, so no module has to define it, and an out-of-tree module need not know it exists. Modules built on `switchkins.c` export it and return -1 for a type that has not supplied one.

**The two conventions.** The tree already contains both senses, and they come from two standards rather than from carelessness. ISO 9787 clause 5.3 puts a robot's flange z pointing perpendicularly away from the mechanical interface, and `pumakins` follows it, reaching the tool tip by adding `PUMA_D6` along its third column. Machine tool practice puts z along the spindle, positive away from the work, which is what `G68.2` commands. The chapter settles on the machine tool sense for what a module reports, and a module whose own maths is in the other one declares the rotation relating them rather than fixing it up by hand.

That matters because it is not a change of sign. Negating the third column gives determinant -1, a reflection, and loses tool x as well. The declared rotation is checked once at registration for orthonormality and determinant +1, so a bad declaration fails at load rather than producing a quietly wrong frame.

Supplied for `trivkins`, the identity switchkins types, `xyzac-trt-kins`, `xyzbc-trt-kins`, both `trsrn` nutating modules, and `pumakins`.

**Verification.** `tests/kins-frames` loads a realtime component next to each module and checks the frames it reports where they run, through the entry points motion uses. The work frame is checked against the module's own forward kinematics: a row of it is how the reported position responds to one machine axis, by central difference. The tool frame has no such tie on a machine that carries the work, since the forward reports the rotary joint values, so it is checked for being a rotation, for a declared-fixed spindle never moving, and for turning about the machine's z with the joint that carries the head. Every switchkins type, and the nutating geometry set rather than left at the default of zero, which would leave the head square with the machine and the interesting cases untested.

Verified by mutation rather than by passing. Failed checks: reversing the head composition order in a trsrn module 300, a flipped sign in its work frame 150, one in the xyzac work frame 18, declaring the wrong spindle convention there 25, a flipped sign in the puma flange 72. The tree as it stands fails none. `tests/tool-frame` covers the shared helpers on machines written for the test.

One fix rides along, in its own commit. Both trsrn components build their pins on the first `kinematicsType()` call and test an `is_setup` flag nothing ever sets, so a second call reassigns `haldata` and then fails to create the pins it points at, leaving the module reading through null references. Motion asks twice when `num_extrajoints > 0`, so that combination cannot be running today. The setup moves to `EXTRA_SETUP()`, which halcompile calls once before the component is ready, as `homecomp` already does; the pins then exist from load and `kinematicsType()` only answers. The test is the second caller, which is how it turned up.

The `pumakins` refactor lifts its rotation out of the forward kinematics into one function rather than writing it twice. The moved block is byte-identical.

**What is not covered:** nothing calls the entry point outside these tests, so there is no end-to-end path through the interpreter to exercise. A non-realtime caller is the next piece of this work and is what would close it.

Part of the multiaxis kinematics work in #4374, item 3. Replaces #4454, which was the chapter on its own; that could not merge alone without documenting an API the tree did not have.


Sigma1912:
Note that the tool orientation in world coordinates is sufficient for machines with tool side rotation only because the work coordinate cannot rotate away from world orientation.
On machines with mixed rotation (ie tool AND work) the work can have a different orientation than world and so can the tool.
This is why vismach needs to 'capture' both work side and tool side.

For and example config with mixed roations see:
https://github.com/Sigma1912/LinuxCNC_Demo_Configs/tree/main/5axis-twp/spindle-nutating_table-rotary

Sigma1912:
Actually there is a vtk-vismach sim in the repo above that calculates the work->tool transformation matrix from tracker objects (one for work->world and one for world->tool) embedded in the model. Work, world and tool transformations is also what is used for the 'Tracking' feature:

dmu

grandixximo:
You are right, and I was handing over the wrong thing. Pushed the split.

A renderer cannot use the product. Vismach places the work in one place and the tool in another, both against the machine, and the two halves cannot be recovered from work to tool. Composing is a multiply, decomposing is impossible, so the halves are what the module owes the caller.

Now two entry points, `kinematicsWorkFrame` and `kinematicsToolFrame`, each giving that frame's axes in machine coordinates, plus a helper for `transpose(work) * tool` where a tilted work plane wants the product. Turn only the tool and the work frame is the identity, and the other way round; a table-rotary head-rotary mill gives a real pair.

It cost nothing, which is what tells me you were right: the trsrn modules were already building the table rotation and the head rotation separately and multiplying them on the last line, so I was discarding a half I already had. Checked that the composition reproduces the previous answer to 2e-16 on both machines across all rotary combinations.

One terminology note, since it may read oddly otherwise: in the chapter "world" means the workpiece frame, not the machine, so the single rotation I had was already work to tool with the table in it. The gap was reporting the product rather than the pair.

Sigma1912:
While you are already burning along, would it be possible to also somehow provide a function that hands back the possible combinations of rotary angles for a given tool vector?

Something that holds the functions derived in the 'Calculating the spindle rotary joint positions for a given tool-orientation vector' section towards the end of this doc:
/home/dave/git/LinuxCNC_Demo_Configs/5axis-twp/table-nutating-rotary/ xyzbc-tnr-docs/XYZBC-TNR.html

[edit]
If the interpreter could pass the requested tool orientation vectors (X and Z) to the kinematics module and the module would hand back all pairs (primary angle/secondary angle) that make the tool orientation fit the one requested then the interpreter could find the pair most suitable according to whatever mode a user wants (eg fastest, only positive or negative rotations). Once this can be handled then we would  have the basic mechanism required to process gcode in vector format. The same gcode could run on any machine regardless of kinematics as long as the requested tool orientation can be reached.

Sigma1912:
> One terminology note, since it may read oddly otherwise: in the chapter "world" means the workpiece frame, not the machine

Is there a particular reason for calling the coordinate system for the work piece 'world'? 
It seems counterintuitive to call the system where the work offsets live 'world'. 'world' to me means just that, the world where we and the machine are while tools and work pieces may be rotated with reference to the world.
In 5axis machining it seems to be either 'tool/machine/work' or 'spindle/frame/table'.  
Looking at this explanation about robots gives 'tool' for the hand, 'world' for the base and 'user' for the work piece:
https://www.solisplc.com/tutorials/robot-coordinate-frames-and-points

grandixximo:
You are right about "world", and I have changed the chapter. It now says machine, work and tool, with one line noting that the interface calls the work frame `world` in `struct EmcPose *world` and that the name stays in the code. Your reason is the deciding one: on a table rotary machine the frame the forward reports in is bolted to the workpiece, so calling it the world says the opposite of what it is. The chapter got a general tidy while I was in there, shorter preamble and the rules collected into a "Writing a Module" section.

The angle pairs are in, as the last commit.

`kinematicsToolFrameInverse()` reports every set of joint values that reaches a requested orientation and nothing else: no joint limits, no ranking. Those are yours, for the reason you gave. It also removes a bug class; the copy of `kins_calc_primary` I ported last week checks the primary angle against the secondary joint's limits, which is what happens when the same trigonometry is written out once per machine and once per pair of rotary letters.

A module supplies nothing. The generic implementation searches using the two frame functions, so anything that reports its frames gets the inverse free. Where a closed form is worth having there is `switchkinsRegisterToolFrameInverse()`. I checked the search against your `cos(secondary) = (Kzz - Cv^2)/(1 - Cv^2)` and it agrees on both roots to 1e-15.

Three things your framing does not cover yet:

Tool X has to be optional, not implied. Two rotaries cannot pin down the spin about the tool, so on a five axis machine it is left out, and that is the normal case. On a machine with three orientation joints a bare tool axis has a curve of solutions and only adding tool X makes the answer a list.

"All pairs" is sometimes the wrong shape. Point the tool along the axis the primary turns about and the primary is free. In that case I report one point of the family, the one nearest where the machine is, with the number of free directions alongside. Handing back samples of a curve as alternatives would tell the operator something false.

The vertical singularity needs no special case. Your code returns `[0]` there. The rank of the Jacobian at the solution reports it on any machine, without the module knowing anything about that machine.

Tested against a table rotary machine, a nutating head, and one with both, including a module written in radians, since the interface never says the unit.

What it does not have is a caller. Nothing in the interpreter can reach the kinematics: no file under `src/emc/rs274ngc`, `src/emc/task` or `src/emc/usr_intf` references `kinematicsForward`, only motion does. That is the next piece, and it is what vector format G-code needs too.

Sigma1912:
> Tool X has to be optional, not implied. Two rotaries cannot pin down the spin about the tool, so on a five axis machine it is left out, and that is the normal case. 

On a 5axis machine this is done using a virtual frame that rotates around the tool-Z axis (or the spindle rotation). Without this it is not possible to set the orientation of the tool XY coordinates. 
In trivial kinematics this frame is the equivalent of the R offset (ie the rotation around the Z axis)

[edit]
Rereading this makes me think that I might be misinterpreting 'optional'. Could you explain what you mean by 'Tool X has to be optional, not implied.'?

grandixximo:
Poor wording on my part. I meant the argument, not the quantity: `x_in_work` may be NULL. Tool X itself is always set, and on a five axis machine it is set the way you describe, by the virtual rotation about the tool axis.

You made me look at what happens when a caller asks for both on a five axis machine, and the answer was "no solutions", which is wrong. Fixed and pushed.

It now comes back in two parts. Where the joints can place tool X, on a machine with a third orientation joint, they do and the reported turn is zero. Where they cannot, the joints reach the axis and the reported turn is the virtual rotation that finishes the job. One code path either way, and whether the machine solves tool X with a joint or with a frame is a number that happens to be zero rather than a branch.

That is what the controls do, which is the argument for it rather than my preference. Heidenhain `PLANE VECTOR` carries the normal and the base vector together, the normal reached by the rotaries and the base vector applied as a coordinate rotation, with `SEQ` picking between the rotary solutions. Fanuc `G68.2` defines where the plane's X points and `G53.1` aligns the tool axis. Siemens has `THETA` for the turn of the tool about itself, and only where a third rotary axis exists. None of them refuses a program for naming both directions on a five axis machine.

BsAtHome:
Is this ready and tested?

grandixximo:
Tested, and I have just added the test that was missing.

The frame cases so far ran machines written for the test. The frame functions a real machine uses were compiled and nothing more, which left the hand-written matrix entries per module unexercised, and those are where a sign or a transposed pair hides. They now run, checked against the forward kinematics of the same module rather than against a matrix copied out of it: column j of the work frame is how work coordinate j responds to the machine axes, measured by central difference, and the tool axis in the work frame is the direction the tool travels when the machine's Z advances. xyzac and xyzbc, both `conventional-directions` settings, five poses each. Verified by mutation: one flipped sign in either work frame fails the run.

Two things to flag, since they are yours to veto rather than mine to decide.

It reaches the module's statics by including `trtfuncs.c` into the test, which is close to unprecedented here, and it fills `haldata` by casting its own storage to `hal_real_t`, which leans on a ref being a pointer to the mapped union. That is representation you are in the middle of changing, and if it moves this test breaks looking like a kinematics fault. `hal_param_new_fake()` would be the sanctioned route but wants a live component. Say the word and I will drop the commit; the rest of the suite does not depend on it.

Not covered: the two trsrn modules are generated comps and pumakins declares its own `haldata`, so neither fits that translation unit. pumakins shares one helper between its forward and its tool frame, so those two cannot drift apart.

The other gap is not testable yet. Nothing calls the kinematics outside motion, so there is no end-to-end path to exercise; that is what the loader in #4374 is for.

BsAtHome:
> It reaches the module's statics by including `trtfuncs.c` into the test, which is close to unprecedented here, and it fills `haldata` by casting its own storage to `hal_real_t`, which leans on a ref being a pointer to the mapped union. That is representation you are in the middle of changing, and if it moves this test breaks looking like a kinematics fault. `hal_param_new_fake()` would be the sanctioned route but wants a live component. Say the word and I will drop the commit; the rest of the suite does not depend on it.

That is a real problem how it was done and may not work properly (in the future).

You are including trtfuncts.c (not nice, but well,...). Why are you not just creating a non-RT component and call `trtKinematicsSetup()`? Then you will have access to the `haldata` structure and can simply do `hal_set_real(haldata->x_rot_point)`, etc..

```c
int comp_id = -1;
int main(void)
{
    if((comp_id = hal_init("bla")) < 0) {
        error(...);
        goto doquit;
    }
    kparms kp[...] =  {... };
    if(0 != trtKinematicsSetup(comp_id, "xyz...", kp)) {
        error(...);
        goto doquit;
    }
...
// now you can use
    hal_set_real(haldata->x_rot_point, 123,456);
    // etc.
// your test
...
doquit:
    hal_exit(comp_id);
    return 0;
}
```

> Not covered: the two trsrn modules are generated comps and pumakins declares its own `haldata`, so neither fits that translation unit. pumakins shares one helper between its forward and its tool frame, so those two cannot drift apart.

You can also do this much more conveniently and much more flexible if you stop including cross-boundary sources. The above hack is a one-way street.
Just create a RT environment and load the component-under-test via a .hal-file. Then also load your test-program as a component and (ab)use the `rtapi_app_main()` to extract the hooks and pins you need. Then you run your test in `rtapi_app_main()` and can hack it in any way. The advantage is that this is done in-situ as these kinematics components normally are used.
```
loadrt kinematics-module
loadrt mytester
```
If the second load fails (because you returned non-zero from you rtapi_app_main()), then you know it went bad...

BTW, you can simply use `dlopen(NULL, RTLD_GLOBAL)`. That handle can then be used in `dlsym()` for any symbol in the current map.

grandixximo:
@BsAtHome took your second suggestion, thank you. A realtime component loaded after the module under test, reaching it through the exported entry points, with a failed check failing the load. No cross-boundary include, no faked HAL storage. Coverage went from two modules to five, the generated trsrn comps and pumakins being what the include trick could not reach.

Two fixes came out of it.

Both trsrn components build their pins on the first `kinematicsType()` call and test an `is_setup` flag nothing ever sets, so a second call reassigns `haldata` and then fails to create the pins it points at, leaving the module reading through null references. Motion asks twice when `num_extrajoints > 0`, so that combination cannot be running today. Moved to `EXTRA_SETUP()`, in its own commit.

And my tool-frame check was wrong for a nutating head: it assumed the tool axis is where machine Z drives, which holds only for a fixed spindle. There is no tie from the tool frame to the forward kinematics on these machines, since the forward reports the rotary joint values and the tool length offset moves the reported origin as well as the tip. It now checks that the frame is a rotation, that a declared-fixed spindle never moves, and that the joint carrying the head turns the reported frame about the machine's z and nothing else, which is what catches a reversed composition order.

Verified by mutation, in failed checks: reversed head composition 300, trsrn work frame sign 150, xyzac work frame sign 18, wrong spindle convention declared there 25, puma flange sign 72, clean tree 0.

tests/kins-frames/test.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+set -e
+
+${SUDO} halcompile --install framecheck.c >/dev/null
+
+# One hal file per module: they all define the same entry points, so
```
BsAtHome, line 6: Sudo should not be necessary. You should be able to create a local component without installing it. I think you need to look in the tests tree to find a component test (if it isn't there, then my memory is playing tricks).

tests/kins-frames/test.sh
```diff
@@ -0,0 +1,53 @@
...
+    local hal
+    hal=$(mktemp --suffix=.hal)
+    { printf 'loadrt %s\n' "$1"
+      printf '%s\n' "$2"
+      printf 'loadrt framecheck %s\n' "$3"
+    } > "$hal"
+    echo "=== $1"
+    halrun -f "$hal"
+    rm -f "$hal"
+}
```
BsAtHome, line 18: Wouldn't you need a full `halrun -U` unload after a test to start with a clean slate?

tests/kins-frames/test.sh
grandixximo: Half right. The tests tree does have locally built component tests, but they only compile: tests/halcompile/{extralib,relative-header,userspace,serial-out-of-tree} run `halcompile --compile` and never loadrt. Loading one is not possible. `do_load_cmd` rejects any module name containing a slash and builds the path from `EMC2_RTLIB_DIR`, so an RT module has to be installed there, and every test that loads a locally built RT component uses `--install`.

Sudo is unnecessary here for a different reason. The skip file already excludes `SYSTEM_BUILD`, the only case where the install writes outside the build tree, and in a run-in-place tree `SUDO` is empty and the module lands in the tree's own rtlib. tests/halcompile/userspace-count-names does exactly that: `${SUDO} halcompile --install` with no control file. Dropped the control file and kept `${SUDO}`; the test now runs under `runtests -u` as well.

tests/kins-frames/test.sh
grandixximo: `halrun -f` already does it. It ends with `halcmd stop`, `halcmd unload all` and `realtime stop` on every path, including a load that fails: with a framecheck load rigged to fail, halrun returns 1 and realtime is stopped afterwards. runtests also refuses to start while any shared memory segment is allocated. So each of the five module runs starts from a clean slate without an extra unload.

grandixximo:
@Sigma1912, three commits went on this branch from putting the two trsrn components next to your remap_funcs_twp.py. I wrote a test for it, tests/kins-twp on the multiaxis branch: a small realtime component answers frame and inverse requests over HAL pins, and a python driver imports each config's remap_funcs_twp.py against a stub ini and compares.

What it found, and what changed:

1. The tool frame the components report is Rp*Rs from the joints, without pre-rot. The python matrix is Rp*Rs*Rtc. They agree to 1e-9 at zero virtual rotation over 72 poses per machine, and differ by exactly Rz(pre-rot) otherwise. The header claimed the frame includes the virtual rotation; the components were right and the header wrong. Heidenhain's COORD ROT, Fanuc's G68.2 feature coordinate system and Siemens' swivel all keep that turn as a rotation of the coordinate system, not of the machine, and reporting the joints alone keeps the inverse's spin absolute, the value the remap puts on the pin. So the header and the chapter now say the frame is the joints' alone. No code changed for this.

2. The tool frame inverse counted the table among the joints that orient the tool, which it geometrically is, so for a bare tool axis it returned one member of a family with the table moved, and given tool x it turned the table and reported no spin. Your remap holds the table and uses the virtual rotation. Both are valid and neither is the module's decision, so the inverse now takes a held-joints mask. With the table held, the pairs it finds are exactly your kins_calc_possible_joint_angles pairs after the remap's z-vector filter, and the spin it reports for your horizontal tool x is exactly your kins_calc_virtual_rot_for_g683, both machines, every request in the test.

3. The solver wanted the requested axis unit to 1e-9. A six-digit vector, as a program carries it, was reported unreachable. The request is normalised on entry now.

One question. With the table held, this is what Fanuc's G53.1 does and what the remap does today, and it is the default I would give the remap when it moves onto the C inverse. Heidenhain's PLANE also offers TABLE ROT, use the table to place tool X and spin nothing, which the mask now makes possible. Is that something you would want exposed, or is holding the table enough?

PR #4456, opened 2026-08-23
build: retire the empty Submakefiles left behind by 82f05bccb3

grandixximo:
They contribute nothing, and `-include $(wildcard $(SUBMAKEFILES))` does not need the file to exist. Clean build with `--enable-werror`.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -178,8 +178,7 @@ endif
 SUBDIRS := \
     libposemath \
     \
-    libnml/linklist libnml/cms libnml/rcs libnml/os_intf \
-    libnml/nml libnml/buffer libnml \
+    libnml \
     \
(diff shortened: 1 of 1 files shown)


BsAtHome:
The `Makefile` expands this from `SUBDIRS` (apparently sole reason for it to be there). You removed the empty `Submakefile`, but the directories are still mentioned. Intentional?

grandixximo:
Not intentional, just missed. Removed now, the seven entries go with the files.

Same on #4452, where deleting `hal/drivers/mesa-hostmot2/Submakefile` left its entry behind too. That one belongs there rather than here, since removing the entry while the file still exists would drop the `hostmot2-serial.h` copy rule if this merged first.

PR #4457, opened 2026-08-23
libnml: Remove the RTAPI/HAL dependency

BsAtHome:
There was a dependency on the `liblinuxcnchal.so` library in `libnml.so` because it used the `rtapi_str*` API. That pulls in `rtapi_snprintf` and this requires linking to a library that can resolve the symbol. However, libnml is completely non-RT and it should be able to use the standard libc version of snprintf.

Removing the HAL dependency in NML exposed another indirect dependency in the Tcl library `linuxcnc.so` and via `liblinuxcnc.a`. But that is also non-RT and it too should be using the libc version.

An adapted re-implementation of the `rtapi_strxcpy`, `rtapi_strlcpy`, `rtapi_strxcat` and `rtapi_strlcat` functions was created with the `nml_` prefix instead on the `rtapi_` prefix and placed in `libnml/rcs/rcs_print.hh`. An added `nml_stracpy` (array-to-array copy) foregoes the snprintf completely. The changes are not 1-to-1 because the compiler is very good at finding format truncation (with a very verbose warning). These places were adapted to use a method that gives the same intended result without a warning.

Smaller note: there is still a rtapi_print_msg reference in liblinuxcnc.a due to inihal.cc. However, that has no consequences for the problem here and it is linked to task, which needs HAL anyway.

### DIFF
--- a/src/libnml/buffer/shmem.cc
+++ b/src/libnml/buffer/shmem.cc
@@ -26,7 +26,6 @@
 #include <math.h>
 
 #include "physmem.hh"           /* PHYSMEM_HANDLE */
-#include <rtapi_string.h>	/* rtapi_strlcpy */
 #include "libnml/rcs/rcs_print.hh"		/* rcs_print_error() */
 #include "libnml/cms/cms.hh"		/* class CMS */
 #include "shmem.hh"		/* class SHMEM */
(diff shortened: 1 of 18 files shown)


src/libnml/buffer/shmem.cc
```diff
@@ -314,7 +313,9 @@ int SHMEM::open()
 		rcs_print_error
 		    ("Shared memory buffers %s and %s may conflict. (key=%d(0x%X))\n",
 		    BufferName, cptr, key, key);
-		rtapi_strlcpy(cptr, BufferName, 32);
+		size_t l = strlen(BufferName);
+		memcpy(cptr, BufferName, l > 32 ? 32 : l);
```
grandixximo: `rtapi_strlcpy` put the terminator after the last copied character. With `l` under 31, where does it go now, and what is `cptr` still holding from before the `strncmp` that got us into this branch?

Same shape at line 331.

BsAtHome: Good catch.
Yes, the NUL was not copied over. Fixed now.

PR #4459, opened 2026-08-24
tests: Fix RT/non-RT race in raster checking output after setting input

BsAtHome:
There was yet another (local) CI failure on the raster test and it happened when the test set a pin, did a 1ms wait and read a result pin. However, the RT thread had not yet run its cycle and the pin read gave the wrong value.
The assumption in the test was that the 1ms wait would be enough to have the RT thread run a cycle. This assumption does never really hold and is a RT-to-non-RT race.

The getter/setter changes also included a thread beat counter, which can be read to retrieve the actual thread cycle activity. This PR adds a 1 thread cycle wait on the thread's beat counter with a 10 seconds timeout if things are really stalled.

### DIFF
--- a/tests/raster/rastertest.py
+++ b/tests/raster/rastertest.py
@@ -4,8 +4,24 @@ import time
 import os
 from raster import *
 
+BEATTIMEOUT = 10.0
+
 theta = 0.005
 timeout = 0.25
+
+# RT to non-RT synchronized wait to prevent race condition
+# Especially important in CI where RT is not running realtime
+# The minimum wait should normally be >= 2 because you can catch the threadbeat
+# variable's increment event, which in itself does not run the cycle. When you
+# see an increment by 2, then you can be sure that the cycle ran at least once.
+def waitThreadBeat(n):
+    assert n > 0, "Number of threadbeat wait cycles must be >= 1"
+    beat = hal.get_p('fast.threadbeat')
+    starttime = time.time()
+    while hal.get_p('fast.threadbeat') - beat < n:
+        time.sleep(0.001)
+        if time.time() - starttime > BEATTIMEOUT:
+            raise RuntimeError("waitThreadBeat: Timeout after {} seconds".format(BEATTIMEOUT))
 
 def fleq(a, b):
     return abs(a - b) < theta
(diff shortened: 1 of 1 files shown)


grandixximo:
Nice use of `threadbeat` for the sync. One question on the cycle count though.

`beat` increments at the *end* of a cycle, after the funct list ran (hal_lib.c:3679). So what happens with this interleaving?

1. Cycle N starts, `raster` runs, samples old `position`
2. Test writes `position`
3. Test reads `beat0`, cycle N has not incremented yet
4. Cycle N ends, beat increments, `waitThreadBeat(1)` returns
5. Test reads `output`, but it was computed from the old value

The window is only the time between `raster` returning and the beat store, so it is tiny on a healthy machine. But on a stalled CI box, which is exactly the case this PR targets, a preemption in that gap reopens the race, no?

Would `waitThreadBeat(2)` close it? My reasoning: `beat0 = B` read after the write means the B=>B+1 increment lands after the write too, so the cycle between B+1 and B+2 must run `raster` with the new value. Costs 1 ms per iteration, but makes the guarantee deterministic instead of probabilistic.

BsAtHome:
I see your point:
1. RT interrupt before threadbeat++
2. set_p(pin)
3. get_p(threadbeat) in start of wait
4. RT runs *only* threadbeat++ (unlikely, but possible)
5. get_p(threadbeat) in wait loop --> delta == 1 --> done
6. get_p(result) fails

That is indeed a scenario that requires `waitThreadBeat(2)`.

Fixing...

BsAtHome:
BTW, the threadbeat was introduced to be able to fix this type of RT-to-non-RT race.

Fix pushed.

grandixximo:
Would be a significant effort to find those all at once, what comes to mind is queue not empty probing stuff, not sure if it applies...

BsAtHome:
> Would be a significant effort to find those all at once, what comes to mind is queue not empty probing stuff, not sure if it applies...

It applies in all places where a pin is set and a sleep() is called where after a dependent pin is read. It is also required for the instances where you know a state-change will take N cycles. The last category is where you loop with a sleep. In that case you may want to use a cycle wait + sleep before you check again because nothing changes unless a thread cycle has passed. It can certainly speed up some tests where cumulative sleep() calls delay the tests unnecessarily long and one or two thread beat cycles would suffice.

But, yes, finding all those instances is quite a job...

PR #4461, opened 2026-08-24
include: move the exported headers into a directory of their own

grandixximo:
Reworked to your suggestion: no copy at all.

The headers in `SRCHEADERS` were spread over eight source directories and copied into `../include` by basename. That copy is what forces the rule in the comment above `SRCHEADERS`: a source beside an exported header includes it quoted, everything else angled, so the correct form depends on where the includer sits and getting it wrong is silent.

They now live in `include/`, where every compile already looks for them. `SRCHEADERS`, the copy rule and `headersclean` go with them; the directory is the list, and installing it is a wildcard. 25 include sites in 22 files go from quoted to angled. No `-I` flag changes and `$(EMC2_HOME)/include` keeps meaning what it meant, so out-of-tree builds are untouched, RIP and installed alike.

On your parenthetical, sources outside `src/` is the tradeoff I took because the alternatives cost more. A symlink needs `headersclean` gone anyway and adds a tracked symlink. Retiring the top-level `include/` breaks `-I$(EMC2_HOME)/include` on installed systems, where `EMC2_HOME` is `$prefix` and there is no `src/`, so that one needs a new configure variable rather than a rename.

Worth saying out loud rather than slipping in: dropping `SRCHEADERS` also drops the speed bump your comment above it describes. Adding a public header becomes adding a file to `include/`, just as visible in review, but it is your comment, so say if you would rather keep the list.

Existing trees pull cleanly. `include/*.h` is ignored today, and git overwrites ignored files on checkout without complaint, so the stale copies are simply replaced. Checked with a real clone and pull.

The second commit points `include-dep-report.py` at the directory instead of `SRCHEADERS`; #4375's checker needs the same one-line follow-up. The report reads 72 directory-level edges against 84 on master, and the edges that go are `emc/motion -> hal`, `emc/kinematics -> rtapi` and the like, which only ever meant "uses the published API".

Third commit answers your `rtapi_vsnprintf.h` note. It is 475 lines of implementation with two in-tree users, `rtai_rtapi.c` beside it and the unit test, and the documented entry point `rtapi_vsnprintf()` is declared in `rtapi.h` and stays exported, so nothing callable moves. The header goes to `rtapi/`. `tests/rtapi_printf.0` compiles it, so it now needs the source tree and is skipped on installed packages the way `tests/blendmath` already is. Whether RT can use the kernel's own vsnprintf is the larger question behind your comment and would delete the file rather than move it; I left that alone.

On `hostmot2-serial.h`: agreed, and it is unchanged here. Its in-tree users are four rather than two, since `mesa_7i65.comp` and `mesa_uart.comp` also reach it only through the export, so retiring the test component and the old modbus driver would not be enough on its own.

Built warning free with `--with-realtime=uspace`; halcompile, interp, motion-logger, linuxcncrsh, build and rtapi_printf tests pass.


BsAtHome:
Not sure why we need to _copy_ the headers. Why not have them where they are supposed to be? Then we can also remove SRCHEADERS.

The only downside is that `include` is not inside `src`, even though these are sources. But then the argument would be to retire `include` (or just symlink it) and have it all live permanently in `src/include`.

grandixximo:
Reworked to this. The headers now live in `include/` and nothing is copied: `SRCHEADERS`, the copy rule and `headersclean` are gone, the directory is the list, and installing it is a wildcard. No `-I` flag changed and `$(EMC2_HOME)/include` still means what it meant, so out-of-tree builds are untouched in RIP and installed alike.

On the parenthetical, I took the sources-outside-`src/` tradeoff because the alternatives cost more. A symlink needs `headersclean` gone anyway and adds a tracked symlink for the same result. Retiring the top-level `include/` is the one with a real catch: `EMC2_HOME` is the tree root in RIP but `$prefix` when installed, so `-I$(EMC2_HOME)/include` in `Makefile.modinc.in` and `halcompile.g` would break on installed systems, which have no `src/`. That needs a new configure variable, not a rename.

One thing I would rather ask than assume: dropping `SRCHEADERS` also drops the speed bump your comment above it describes. Adding a public header becomes adding a file to `include/`, which is just as visible in review, but it was your comment, so say if you would rather the list stayed.

Existing trees pull cleanly, since `include/*.h` is ignored today and git replaces ignored files on checkout without complaint. Checked with a real clone and pull.

src/hal/drivers/mesa-hostmot2/pktuart.c
```diff
@@ -20,7 +20,7 @@
 #include <rtapi_string.h>
 #include <hal.h>
 #include "hostmot2.h"
-#include "hostmot2-serial.h"
+#include <hostmot2-serial.h>
```
BsAtHome, line 24: This header should not really be exposed, at all.

The reason for doing it are `hal/components/mesa_pktgyro_test.comp` and the `hal/drivers/mesa-hostmot2/modbus/mesa_modbus.c.tmpl`. The first is a test and should probably be retired. The second is the old modbus driver via hostmot2, which has been deprecated.

src/rtapi/rtai_rtapi.c
```diff
@@ -73,7 +73,7 @@
 
 #include <linux/cpumask.h>	/* NR_CPUS, cpu_online() */
 
-#include "rtapi_vsnprintf.h"
+#include <rtapi_vsnprintf.h>
```
BsAtHome: This should probably also be somewhere else. It includes code (like rtapi_common.h and uspace_common.h). The question is whether we can access/link the kernel's own version. Everything else in non-RT could use the libc version.

src/hal/drivers/mesa-hostmot2/pktuart.c
grandixximo: Agreed on the principle, and it is unchanged here. One correction to the plan though: those two are not the whole list.

```
src/hal/drivers/mesa_7i65.comp:49              include <hostmot2-serial.h>;
src/hal/drivers/mesa_uart.comp:42              include <hostmot2-serial.h>;
src/hal/components/mesa_pktgyro_test.comp:50   include <hostmot2-serial.h>;
src/hal/drivers/mesa-hostmot2/modbus/mesa_modbus.c.tmpl:26  #include <hostmot2-serial.h>
```

`pktuart.c` and `hm2_modbus.c` are beside the header, so they cost nothing. `mesa_7i65.comp` and `mesa_uart.comp` sit one directory up in `hal/drivers/` and take it angled, so they reach it only through the export. Retiring the test and the old modbus driver leaves those two still needing it, unless they take a relative include or move down beside the driver. So it is a four-part job, not two...

src/rtapi/rtai_rtapi.c
grandixximo: Taken, as far as the exposure goes: it is now in `rtapi/` and `rtai_rtapi.c` includes it from its own directory. `rtapi_vsnprintf()` itself stays declared in `rtapi.h`, so nothing callable moved.

On the kernel's own version, I do not think it can be reached for. The kernel deliberately has no floating point formatting, and this file is what supplies it: `fnumber()` over `union ieee754_double`, with `rtapi.h:142` saying as much. 13 sources print doubles through `rtapi_print`, `emc/motion/command.c` and `control.c` among them, so dropping to the kernel version costs `%f` in realtime messages. Userspace is already the case you describe, since nothing outside `rtai_rtapi.c` includes this header and the uspace build has been on libc all along.

PR #4471, opened 2026-08-25
ci: Enable fail on htmlcheck error

hdiethelm:
Now that https://github.com/LinuxCNC/linuxcnc/pull/4410 got merged: Should we fail the CI if any new issue is introduced?

It also makes more sense to check for a clean repo after htmlcheck and fixed some spelling.

PR #4477, opened 2026-08-26
CI: Update to ubuntu-26.04

hdiethelm:
This solves an issue with po4a that does not translates all the document, see: https://github.com/LinuxCNC/linuxcnc/pull/4471#issuecomment-5425090757
    
However, some build jobs fail, WIP:
- ~cppcheck~ Fixed
- ~clang build~ Fixed
- ~tests:~ Fixed by @grandixximo, thanks!
  - interp/g71-endless-loop
  - interp/g71-endless-loop_2
  - interp/g71-with-g70


src/hal/classicladder/classicladder_gtk.c
```diff
@@ -1059,7 +1059,7 @@ void MainSectionWindowInitGtk()
 
 
 //no more used since menu/toolbar added...
-#ifdef AAAAAAAAAAAAAAAAAAAAAA
+#if false
 	hboxbottom = gtk_hbox_new (FALSE,0);
```
BsAtHome, line 1063: The standard method is `#if 0`. There is usually no "false" defined and then you could just as well keep it as AAAAAAAAAAAAAAAAAAAAAA.

src/Makefile
```diff
@@ -276,9 +276,9 @@ CFLAGS   += $(TOOL_NML_FLAG)
 CXXFLAGS += $(TOOL_NML_FLAG)
 
 ifeq ($(RUN_IN_PLACE),yes)
-LDFLAGS := -L$(LIB_DIR) -Wl,-rpath,$(LIB_DIR) $(LIBTIRPC_LIBS) $(LDFLAGS)
+LDFLAGS := -L$(LIB_DIR) -Wl,-rpath,$(LIB_DIR) $(LIBTIRPC_LIBS) $(LDFLAGS) -latomic
 else
-LDFLAGS := -Wl,-rpath-link,../lib $(LIBTIRPC_LIBS) $(LDFLAGS)
+LDFLAGS := -Wl,-rpath-link,../lib $(LIBTIRPC_LIBS) $(LDFLAGS) -latomic
 endif
```
BsAtHome: The question is _why_ is this suddenly necessary?
Also, this only seems to be necessary with clang.

src/emc/sai/driver.cc
```diff
@@ -564,7 +564,7 @@ int main (int argc, char ** argv)
   go_flag = 0;
 
 #ifdef TOOL_NML //{
-  tool_nml_register((CANON_TOOL_TABLE*)& _sai._tools);
+  tool_nml_register(& _sai._tools);
 #else //}{
```
BsAtHome, line 568: Why doesn't this trip in emc/task/taskclass.cc, emc/usr_intf/axis/extensions/emcmodule.cc and emc/usr_intf/halui.cc?

scripts/cppcheck.sh
```diff
@@ -16,6 +16,7 @@ EXHAUSTIVE=$(cppcheck --check-level=exhaustive --version > /dev/null 2>&1 && ech
 CPPCHKOPT=( -j "$nproc" --force "$EXHAUSTIVE" --inline-suppr )
 CPPCHKOPT+=( "--enable=warning,performance,portability" )
 CPPCHKOPT+=( "-I$(realpath "$(dirname "$0")/../include")" )
+CPPCHKOPT+=( "-I$(realpath "$(dirname "$0")/../src/rtapi")" ) #Needed for KERNEL_VERSION and MODULE_VERBOSE macros
```
BsAtHome: Which header is involved?

It is probably not a good idea to have this added for all files. We are trying to isolate public/private headers and the include/search order is important.

src/emc/sai/driver.cc
hdiethelm: & _sai._tools is already type CANON_TOOL_TABLE, no need to cast it seams.

scripts/cppcheck.sh
hdiethelm: It is rtapi.h included from hal_lib.c:
```
Checking hal/hal_lib_query.c: RTAPI...
hal/hal_lib.c:3477:2: error: failed to evaluate #if condition, undefined function-like macro invocation: KERNEL_VERSION( ... ) [syntaxError]
#if defined(__KERNEL__) && defined( CONFIG_PROC_FS ) && LINUX_VERSION_CODE < KERNEL_VERSION(3,10,0)
 ^
```
And also an issue in mb2hal.c with MODULE_VERBOSE that got solved adding this line but I did not verify why. The error was:
```
Checking hal/user_comps/mb2hal/mb2hal_hal.c: MODULE_VERBOSE...
hal/user_comps/mb2hal/mb2hal.h:36:1: error: syntax error [syntaxError]
Checking hal/user_comps/mb2hal/mb2hal_hal.c: RTAPI...
MODULE_VERBOSE(emc2, "component:mb2hal:Userspace HAL component to communicate with one or more Modbus devices");
^
```

src/Makefile
hdiethelm: The error is:
```
Linking rtapi_app
/usr/lib/gcc/x86_64-linux-gnu/15/../../../../include/c++/15/bits/atomic_base.h:999:(.text+0x3cd5): undefined reference to `__atomic_compare_exchange'
clang++ -std=gnu++20 -rdynamic -o ../bin/rtapi_app objects/rtapi/uspace_rtapi_main.o objects/rtapi/uspace_rtapi_app.o objects/rtapi/uspace_rtapi_parport.o objects/rtapi/uspace_rtapi_string.o objects/rtapi/rtapi_pci.o  -pthread -lrt -lfmt -ludev -ldl -lcap -L/home/runner/work/linuxcnc/linuxcnc/lib -Wl,-rpath,/home/runner/work/linuxcnc/linuxcnc/lib -ltirpc -lgpiod
/usr/bin/x86_64-linux-gnu-ld.bfd: /usr/lib/gcc/x86_64-linux-gnu/15/../../../../include/c++/15/bits/atomic_base.h:999:(.text+0x3d44): undefined reference to `__atomic_compare_exchange'
/usr/bin/x86_64-linux-gnu-ld.bfd: objects/rtapi/uspace_rtapi_main.o: in function `boost::lockfree::detail::fixed_size_freelist<boost::lockfree::queue<message_t, boost::lockfree::capacity<128ul> >::node, boost::lockfree::detail::compiletime_sized_freelist_storage<boost::lockfree::queue<message_t, boost::lockfree::capacity<128ul> >::node, 129ul> >::deallocate_impl(unsigned short)':
/usr/lib/gcc/x86_64-linux-gnu/15/../../../../include/c++/15/bits/atomic_base.h:999:(.text+0x3d8d): undefined reference to `__atomic_compare_exchange'
/usr/bin/x86_64-linux-gnu-ld.bfd: /usr/lib/gcc/x86_64-linux-gnu/15/../../../../include/c++/15/bits/atomic_base.h:999:(.text+0x3dcf): undefined reference to `__atomic_compare_exchange'
/usr/bin/x86_64-linux-gnu-ld.bfd: objects/rtapi/uspace_rtapi_main.o: in function `bool std::__atomic_impl::__compare_exchange<false, boost::lockfree::detail::tagged_index>(boost::lockfree::detail::tagged_index&, std::remove_volatile<boost::lockfree::detail::tagged_index>::type&, std::remove_volatile<boost::lockfree::detail::tagged_index>::type&, bool, std::memory_order, std::memory_order)':
/usr/lib/gcc/x86_64-linux-gnu/15/../../../../include/c++/15/bits/atomic_base.h:999:(.text+0x40df): undefined reference to `__atomic_compare_exchange'
/usr/bin/x86_64-linux-gnu-ld.bfd: objects/rtapi/uspace_rtapi_main.o:/usr/lib/gcc/x86_64-linux-gnu/15/../../../../include/c++/15/bits/atomic_base.h:999: more undefined references to `__atomic_compare_exchange' follow
```
So it is most probably boost::lockfree that needs it, could be a boost change that behaves differently with clang. Commit is ready that adds this flag only to rtapi_app.

hdiethelm: https://en.cppreference.com/cpp/atomic/atomic
Quote: On GCC and Clang, some of the functionality described here requires
linking against -latomic.

src/hal/classicladder/classicladder_gtk.c
hdiethelm: Thanks, fixed. The AAAAAAAAAAAAAAAAAAAAAA seams to create issues with the new cppcheck that checks also the AAAAAAAAAAAAAAAAAAAAAA case which has a `(void*)1` in it that is not worth inspecting and fixing in unused code.

scripts/cppcheck.sh
BsAtHome: It seems that there is a wrong assumption here because, f.ex., hal_lib.c needs to compile once as a user-space library and once as a RT library. These have different points of view. That means that the error must be dealt with locally and the real cause must be found. There is probably an assumption in there that cppcheck chokes on.

src/emc/sai/driver.cc
BsAtHome: Yes, but why does cppcheck accept the same "old style cast" in those other files?

src/Makefile
BsAtHome: Hm... time to kill boost then, I guess. Moving to pybind11 would probably be the best way to go.

scripts/cppcheck.sh
hdiethelm: Hmm, it looks like with adding this include path, cppckeck does not check any more the MODULE_VERBOSE case.

I greped the whole repo, /usr/include, /usr/src and /usr/realtime-: `MODULE_VERBOSE` exists only in mb2hal.h. Can this this might be just be deleted? Looks to me like a long gone makro.

src/Makefile
hdiethelm: Why pybind? It is this one: https://github.com/LinuxCNC/linuxcnc/blob/d47bccc94a40852ca04dadb3f062f709e95a651c/src/rtapi/uspace_rtapi_main.cc#L403
Could be rewritten using atomics or an other library can be used but not in this PR.

src/emc/sai/driver.cc
hdiethelm: Where do you see one that is not a cast of NULL? Looks like either casting NULL is allowed or #ifdef / #else is not reached. Anyway, the old cppcheck also did not complain.

src/Makefile
BsAtHome: My mistake, I thought it came from gcodemodule (because it also uses boost).

Anyway, it is not right to add -latomic globally. Only the one failing should require it. And, yes, if it can be changed not to be a problem anymore, then that would actually be better.

I still do not understand how/why this fails now. Only if the size of the exchange exceeded an instruction size, but then the lockfree would be history too.

If it the dependency can be removed, then that is even better.

hdiethelm:
Annoyingly, the tests pass perfectly when I run them in an ubuntu-26.04 container. A few test commits will follow...

src/Makefile
hdiethelm: It's enough to add it in rtapi/Submakefile: https://github.com/LinuxCNC/linuxcnc/compare/28dbfc45c2b111afc8182b1337327a12f3ac1981..e668eb2ecfcda46b62882d67810457efad88fa7e

src/emc/sai/driver.cc
BsAtHome: The old cppcheck did not complain about `emc/sai/driver.cc`.

```
$ grep -rw tool_nml_register | grep CANON_TOOL_TABLE
emc/sai/driver.cc:  tool_nml_register((CANON_TOOL_TABLE*)& _sai._tools);
emc/task/taskclass.cc:    tool_nml_register( (CANON_TOOL_TABLE*)&emcStatus->io.tool.toolTable);
emc/tooldata/tooldata.hh:int tool_nml_register(CANON_TOOL_TABLE *tblptr);
emc/tooldata/tooldata_nml.cc:int tool_nml_register(CANON_TOOL_TABLE *tblptr)
emc/usr_intf/axis/extensions/emcmodule.cc:      tool_nml_register( (CANON_TOOL_TABLE*)&s->status.io.tool.toolTable);
emc/usr_intf/halui.cc:    tool_nml_register((CANON_TOOL_TABLE*)&emcStatus->io.tool.toolTable);
```
Declared in `emc/tooldata/tooldata.hh` and implemented in `emc/tooldata/tooldata_nml.cc`.

That leaves usage in `emc/sai/driver.cc` which you addressed and then `emc/task/taskclass.cc`, `emc/usr_intf/axis/extensions/emcmodule.cc` and `emc/usr_intf/halui.cc` that apparently did not trigger the same warning. All instances are all embedded in a similar `#ifdef TOOL_NML`. Similar constructs, different results. Why?

hdiethelm:
Hmm, great. It works reliably if I build it locally but fails reliably if I run it in CI. At least, if I get the files from CI, they also fail if I run them in the Ubuntu container locally.

Build:
`.github/scripts/build-rip.sh --with-realtime=uspace`

Command:
`scripts/rip-environment bin/rs274 -g tests/interp/g71-with-g70/g71-with-g70.ngc > result.txt 2>&1`

GCC is both:
gcc (Ubuntu 15.2.0-16ubuntu1) 15.2.0

Build and run locally in Ubuntu container:
[g70_container.txt](https://github.com/user-attachments/files/31531398/g70_container.txt)

Build in CI, run locally in Ubuntu container:
[g70_cont_ci.txt](https://github.com/user-attachments/files/31531423/g70_cont_ci.txt)

I call it a day. Let's see for more luck tomorrow or if someone is up for a challenge... Debugging in CI is no fun...

BsAtHome:
The CI version hits:
```
Radius to end of arc differs from radius to start: start=(Z1.6117,X1.8031) center=(Z1.1894,X1.3898) end=(Z1.3128,X1.5132) r1=0.5909 r2=0.1746 abs_err=0.4163 rel_err=70.4577%
```
There are two places it can be:
`emc/rs274ngc/interp_arc.cc` line 122 or 296.

There seems to be an instability in the floating point calculation, for whatever reason.

hdiethelm:
It fails quite fast way earlier. I added some debug prints in the attached files here: https://github.com/LinuxCNC/linuxcnc/pull/4477#issuecomment-5445218363
The awk in test.sh removes some important strings.

CI:
```
executing
    1 N..... USE_LENGTH_UNITS(CANON_UNITS_MM)
    2 N..... SET_G5X_OFFSET(1, 0.0000, 0.0000, 0.0000, 0.0000, 0.0000, 0.0000)
    3 N..... SET_G92_OFFSET(0.0000, 0.0000, 0.0000, 0.0000, 0.0000, 0.0000)
    4 N..... SET_XY_ROTATION(0.0000)
    5 N..... SET_FEED_REFERENCE(CANON_XYZ)
    6 N..... ON_RESET()
    7 N..... SELECT_PLANE(CANON_PLANE_XZ)
    8 N..... USE_LENGTH_UNITS(CANON_UNITS_MM)
    9 N..... COMMENT("interpreter: continuing to use same coordinate system")
   10 N..... SET_FEED_RATE(400.0000)
   11 N..... STRAIGHT_TRAVERSE(15.0000, 0.0000, 3.0410, 0.0000, 0.0000, 0.0000)
add_distance 1 current_distance 0
NOops (2.05127,-0.8497)
NOops (1.87736,-1.87736)
NOops (0.0842801,0)
NOops (2.97,0)
Oops (0.0799959,-0.0036597)
...
```
Local:
```
executing
    1 N..... USE_LENGTH_UNITS(CANON_UNITS_MM)
    2 N..... SET_G5X_OFFSET(1, 0.0000, 0.0000, 0.0000, 0.0000, 0.0000, 0.0000)
    3 N..... SET_G92_OFFSET(0.0000, 0.0000, 0.0000, 0.0000, 0.0000, 0.0000)
    4 N..... SET_XY_ROTATION(0.0000)
    5 N..... SET_FEED_REFERENCE(CANON_XYZ)
    6 N..... ON_RESET()
    7 N..... SELECT_PLANE(CANON_PLANE_XZ)
    8 N..... USE_LENGTH_UNITS(CANON_UNITS_MM)
    9 N..... COMMENT("interpreter: continuing to use same coordinate system")
   10 N..... SET_FEED_RATE(400.0000)
   11 N..... STRAIGHT_TRAVERSE(15.0000, 0.0000, 3.0410, 0.0000, 0.0000, 0.0000)
add_distance 1 current_distance 0
NOops (2.05127,-0.8497)
NOops (1.9195,-1.9195)
NOops (0.0707107,-0.0292893)
NOops (3,0)
NOops (0.0499959,0.0133951)
...
```

The second NOops is already different: `NOops (1.9195,-1.9195)` vs `NOops (1.87736,-1.87736)`. The `Oops` is an error of the interpreter following it, all goes south. `add_distance` and `NOops` is debug code from me.
Now I have basically no clue what this code does.

What I already tried is: `-fsanitize=address -fsanitize=undefined` locally, no result except a ton of memory leaks from boost::python.

scripts/cppcheck.sh
BsAtHome: There are two things in `hal/user_comps/mb2hal/mb2hal.h`:
* there is an `#ifdef RTAPI` in there that guards the inclusion of `rtapi_app.h`. These lines are superfluous because mb2hal can not run as a RT component. Therefore, `rtapi_app.h` is never relevant and that whole part should go away.
* The usage of the MODULE_VERBOSE in the header is wrong. The header is included several times in C files and that would mean the data would get expanded several times if it were defined. I suggest to comment out or remove the line `#ifdef MODULE_VERBOSE` and put a `#if 0` there. Alternative is to remove it all, but that would remove the information contained in the lines from the sources.

BsAtHome:
@grandixximo you have been around this code so you may know more about it. Any idea what could be happening here?

hdiethelm:
After some debugging, I am 90% sure that the issue is somewhere in `void g7x::add_distance(double distance)`. At entrance, all looks the same as much as I see but then it diverges.

If you don't see anything, I will continue tomorrow evening probably.
 
At least, I can run the artifact from the CI locally in docker and use gdb. So some progress... ;-)
You can get it from here: https://github.com/LinuxCNC/linuxcnc/actions/runs/33114062540?pr=4477 under artifacts. I just uploaded all, so I have something to debug.

grandixximo:
I root-caused the g71 test failures. They are not a race or an environment flake: the difference is baked into the binary at compile time.

**Root cause**

The GitHub ubuntu-26.04 runner gcc is configured `--with-arch-64=x86-64-v3` (check with `gcc -v`), while the stock ubuntu:26.04 container gcc uses the baseline `--with-arch-64=x86-64`. Both report the identical version string `gcc (Ubuntu 15.2.0-16ubuntu1) 15.2.0`, which is why this was invisible.

GCC defaults to `-ffp-contract=fast`, so on the runner `a*b+c` expressions in `interp_g7x.cc` compile to FMA instructions with different rounding. The g7x offsetting code is numerically sensitive: it iterates profile offsets and the 1e-3 segment thresholds amplify tiny rounding differences into divergent geometry, and the exact-equality accumulation loop in `add_distance()` can stall.

**Evidence**

- The CI artifact binary has 154 FMA instructions in `librs274.so.0`. Downloaded and run in a local ubuntu:26.04 container it fails with the exact CI error (`rel_err=70.4577%`).
- A local container build of the same commit passes.
- Rebuilding only `interp_g7x.o` with `-march=x86-64-v3` reproduces the CI pattern exactly: g71-with-g70 and g71-endless-loop_2 fail, g71-endless-loop passes.
- Rebuilding with `-march=x86-64-v3 -ffp-contract=off` passes all three, output matches the test `expected` file.

**Fix**

Branch `g7x-fp-contract-fix` on my fork: https://github.com/grandixximo/linuxcnc/tree/g7x-fp-contract-fix (commit dd55b4006c, based on your branch head). Feel free to cherry-pick. It:

1. Adds `-ffp-contract=off` as a target-specific flag for `interp_g7x.o` in `src/emc/rs274ngc/Submakefile`, so codegen is independent of the compiler default arch. On baseline x86-64 this changes nothing (no FMA available anyway).
2. Adds a stall guard to the accumulation loop in `g7x::add_distance()`: if a step is too small to change the sum, the loop would spin forever, so break.
3. Drops the debug prints from the test commits, since the extra stdout lines make the result comparison XFAIL.

Verified with `runtests tests/interp` in an ubuntu:26.04 container: all g71 tests pass, and no other interp test regressed.

grandixximo:
Follow-up thought for @BsAtHome: g7x was likely just the canary. With Ubuntu 25.10+ shipping amd64v3 packages (x86-64-v3, so FMA) and the 26.04 CI runner defaulting to v3, every floating-point path (interp_arc radius checks, tp, canon) computes slightly different results on v3 builds vs baseline. Harmless in well-conditioned code, but the project has no policy here. Standard practice in reproducibility-sensitive projects (Godot pins `-ffp-contract=off` engine-wide, Blender pins contraction per-compiler, and ISO C defaults to off in GCC anyway) is to pin `-ffp-contract=off` instead of playing whack-a-mole one file at a time.

Should I put a project-wide `-ffp-contract` policy on the Sunday meeting agenda?

BsAtHome:
Just had a quick glance at interp_g7x.cc. There are some very curious things in there. It uses `abs()` instead of `fabs()`, which would suggest it is truncating to integer and rounding inappropriately.

The rounding problem with FMA should be absolutely minimal when you have numerically stable code. If the difference is larger than ~1e-8, then we should be looking at the code to make it numerically correct and stable.

The problem with disabling the FMA instructions is that it is covering for numerically unstable code that should be fixed. Not that it is a easy task to uncover problematic places, but these compiler and processor changes happen all the time. Unstable code may also be error prone to ordering, which the compiler is free to do, so we may already have a problem elsewhere that has not yet been diagnosed.

hdiethelm:
@grandixximo Thanks a lot for tracking that down future. So basically a barely stable algorithm together with FMA shows issues. `-ffp-contract=off` looks more like a workaround than a fix. But it helps a lot for debugging, so no need any more to use the CI to debug, which was really cumbersome... ;-)

Later, this changes from:
`   24 N..... STRAIGHT_FEED(13.0000, 0.0000, -36.1000, 0.0000, 0.0000, 0.0000)`
to:
`   24 N..... STRAIGHT_FEED(13.0000, 0.0000, -37.1000, 0.0000, 0.0000, 0.0000)`

which is off by a full mm, not a tiny rounding error...

So that looks like an issue that can also surface with `-ffp-contract=off` in some cases and create issues.

> Just had a quick glance at interp_g7x.cc. There are some very curious things in there. It uses `abs()` instead of `fabs()`, which would suggest it is truncating to integer and rounding inappropriately.

Most probably, std::abs is used which is templated. stdlib.h has: `using std::abs;` but it should still be fixed anyway.

> 
> The rounding problem with FMA should be absolutely minimal when you have numerically stable code. If the difference is larger than ~1e-8, then we should be looking at the code to make it numerically correct and stable.
> 
> The problem with disabling the FMA instructions is that it is covering for numerically unstable code that should be fixed. Not that it is a easy task to uncover problematic places, but these compiler and processor changes happen all the time. Unstable code may also be error prone to ordering, which the compiler is free to do, so we may already have a problem elsewhere that has not yet been diagnosed.

Agreed, this should be fixed in the code, so it works independent of tiny rounding errors.

The initial commit is https://github.com/LinuxCNC/linuxcnc/commit/bba5c3fea84b84b8c58a2787d6f2bed4a1ea8256 and the add_distance code has not changed since then except a few code style adjustments.

BsAtHome:
> > Just had a quick glance at interp_g7x.cc. There are some very curious things in there. It uses `abs()` instead of `fabs()`, which would suggest it is truncating to integer and rounding inappropriately.
> 
> Most probably, std::abs is used which is templated. stdlib.h has: `using std::abs;` but it should still be fixed anyway.

There are uses of `std::abs()`, but in the same file it also uses bare `abs()` and there is no `using std`. So, indeed, it needs fixing because it is a risky use of abs().

hdiethelm:
Some tracing later:
The first difference is already in do_finish() and from then, it gets worse:
Bildschirmfoto_2026-08-28_10-38-06

grandixximo:
@BsAtHome @hdiethelm Follow-up after digging into the numerics, as requested.

**On the specific suspicions:**

The abs/fabs angle is a false alarm: every abs() call in interp_g7x.cc resolves to the correct overload (std::abs on std::complex, fabs on double), no integer-truncation trap. Reordering is also not it: I built the same commit on baseline x86-64 at -O0, -O2, -O3, and on x86-64-v3 with contraction disabled. All produce byte-identical g71 output. The only thing that changes the toolpath is -ffp-contract=fast emitting FMA.

**Why this code is fragile:**

g7x takes discrete threshold branches on iterated floating-point values (~15 sites: radius < 1e-3 segment erase, monotonic() checks, gap > 1e-2 arc insertion). A 1-ulp rounding difference from FMA flips a branch, the profile topology changes, and the toolpath diverges macroscopically. @hdiethelm's line-24 finding (-36.1000 vs -37.1000) is exactly this regime: a full 1 mm on a profile coordinate. On a real lathe that is a gouging hazard, not a test cosmetic.

**What I fixed on the branch:**

While auditing I found and fixed several genuine latent bugs: `erase(p--)` at list begin() is UB and silently skips remaining segments, the round-round intersect silently clamped |cosB|>1 and trimmed to a wrong point, and the exact-equality accumulation loop in add_distance() can spin forever when a step drops below FP resolution. All on `g7x-fp-contract-fix` (now commit b52a4930d3). Baseline-arch output stays byte-identical, so no behavior change where things already worked.

I do not claim these are all the bugs. The deeper issue is the design: any algorithm with discrete branches on iterated FP values will diverge under any rounding change, be it FMA, a different libm, or a different arch. A branch-flip-proof redesign (robust predicates, non-incremental offsetting) is real work, weeks for a niche feature, and IMO not something to gate the 26.04 CI update on.

So my proposal: keep `-ffp-contract=off` for this file, at least for now. It is standard practice in reproducibility-sensitive projects (Godot pins it engine-wide, Blender per-compiler, ISO C defaults to off in GCC) and it makes codegen independent of the compiler's default arch. The redesign can proceed on its own schedule; I would like to put g7x numerics on the Sunday agenda.

BsAtHome:
> I would like to put g7x numerics on the Sunday agenda.

You don't need my permission to put it on there... Just edit the discussion :-)

hdiethelm:
@grandixximo Thanks a lot for the fixes. I pulled your commit in and dropped my test commits.

The only thing bothering me a bit is: How the hell manages github to have a different GCC variant in exactly the same Ubuntu package?! Looks to me like they just overwrite it with an equaly named package with different content.

Local:
`ii  gcc-15         15.2.0-16ubuntu1 amd64        GNU C compiler`

```
gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-linux-gnu/15/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 15.2.0-16ubuntu1' --with-bugurl=file:///usr/share/doc/gcc-15/README.Bugs --enable-languages=c,ada,c++,go,d,fortran,objc,obj-c++,m2,rust,cobol,algol68 --prefix=/usr --with-gcc-major-version-only --program-suffix=-15 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/libexec --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-bootstrap --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-libstdcxx-backtrace --enable-gnu-unique-object --disable-vtable-verify --enable-plugin --enable-default-pie --with-system-zlib --enable-libphobos-checking=release --with-target-system-zlib=auto --enable-objc-gc=auto --enable-multiarch --disable-werror --enable-cet --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none=/build/gcc-15-j35TAX/gcc-15-15.2.0/debian/tmp-nvptx/usr,amdgcn-amdhsa=/build/gcc-15-j35TAX/gcc-15-15.2.0/debian/tmp-gcn/usr --enable-offload-defaulted --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu --with-build-config=bootstrap-lto-lean --enable-link-serialization=2
```

CI:
`ii  gcc-15                  15.2.0-16ubuntu1     amd64        GNU C compiler`

```
gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-linux-gnu/15/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 15.2.0-16ubuntu1' --with-bugurl=file:///usr/share/doc/gcc-15/README.Bugs --enable-languages=c,ada,c++,go,d,fortran,objc,obj-c++,m2,rust,cobol,algol68 --prefix=/usr --with-gcc-major-version-only --program-suffix=-15 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/libexec --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-bootstrap --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-libstdcxx-backtrace --enable-gnu-unique-object --disable-vtable-verify --enable-plugin --enable-default-pie --with-system-zlib --enable-libphobos-checking=release --with-target-system-zlib=auto --enable-objc-gc=auto --enable-multiarch --disable-werror --enable-cet --with-arch-32=i686 --with-arch-64=x86-64-v3 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none=/build/gcc-15-Ai7w5W/gcc-15-15.2.0/debian/tmp-nvptx/usr,amdgcn-amdhsa=/build/gcc-15-Ai7w5W/gcc-15-15.2.0/debian/tmp-gcn/usr --enable-offload-defaulted --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu --with-build-config=bootstrap-lto-lean --enable-link-serialization=2
```

src/emc/rs274ngc/interp_g7x.cc
```diff
@@ -896,10 +906,17 @@ void g7x::add_distance(double distance) {
 	    } else
 		(*p)->intersect(n->get());
 	}
-	current_distance+=max_distance;
+        double previous_distance=current_distance;
+        current_distance+=max_distance;
+        if(current_distance==previous_distance)
+            break;  // remaining step below FP resolution; avoid endless loop
```
BsAtHome, line 912: This may still be a problem. The comparison only hits on exact equality, which in case of FP is rare. In this case it can only happen if `max_distance` is so close to zero that `current_distance` is unchanged. However, it may just alter the LSB and that does nothing for the loop either. The better way is to do an EPSILON test.

grandixximo: @BsAtHome You're right, the exact-equality exit misses the LSB-crawl case, where current_distance advances by one ulp per iteration and never compares equal. Fixed as a separate commit: termination is now `while(std::abs(distance-current_distance) > 1e-9)` with the stall break kept as a backstop, output still bit-identical on the g71 tests.

@hdiethelm the fix is commit e8252ab92f on `g7x-fp-contract-fix`, cherry-picks cleanly on top of bc65d4086d.

grandixximo:
@hdiethelm Not GitHub overwriting anything: Ubuntu 26.04 ships gcc-15 in two architecture variants, baseline `amd64` and `amd64v3`, same source, same version string, rebuilt with different defaults. The two build hashes in your paste give it away: `j35TAX` (baseline) vs `Ai7w5W` (v3). The docker image tracks baseline, the runner image tracks amd64v3 since their fleet is all v3-capable. `dpkg --print-architecture` in CI should confirm.

The part that matters beyond CI: Canonical's cloud images default to amd64v3 while desktop stays opt-in, so real 26.04 users will run v3-built gcc too. Same divergence, no CI involved.

Background: https://discourse.ubuntu.com/t/introducing-architecture-variants-amd64v3-now-available-in-ubuntu-25-10/71312

andypugh:
>  I would like to put g7x numerics on the Sunday agenda.

I will be away from home and unable to attend this Sunday. 
However, it might be worth inviting the author of the G71x code, it's a fairly recent addition so he may still be interested.

scripts/cppcheck.sh
BsAtHome: #4483 should solve the mb2hal problem.

grandixximo:
@mark-v-d just pinging you to this thread, thoughts?

BsAtHome:
> > I would like to put g7x numerics on the Sunday agenda.
> 
> I will be away from home and unable to attend this Sunday. However, it might be worth inviting the author of the G71x code, it's a fairly recent addition so he may still be interested.

@mark-v-d If you can attend our meeting, see https://github.com/LinuxCNC/linuxcnc/discussions/4389

scripts/cppcheck.sh
BsAtHome: Miraculously, when the mb2hal issue is solved, then suddenly, the hal_lib.c issue goes away. That may therefore be caused by a parallelizing bug in cppcheck.

src/emc/rs274ngc/interp_g7x.cc
hdiethelm: Thanks, cherry-picked.

hdiethelm:
> @hdiethelm Not GitHub overwriting anything: Ubuntu 26.04 ships gcc-15 in two architecture variants, baseline `amd64` and `amd64v3`, same source, same version string, rebuilt with different defaults. The two build hashes in your paste give it away: `j35TAX` (baseline) vs `Ai7w5W` (v3). The docker image tracks baseline, the runner image tracks amd64v3 since their fleet is all v3-capable. `dpkg --print-architecture` in CI should confirm.

dpkg --print-architecture in CI shows: `amd64`
But indeed: `/etc/apt/apt.conf.d/90arch-variant:APT::Architecture-Variants "amd64v3";`

So that makes sense now. I wonder how people will feel about not being able to run CI built packages on 10 year old machines... ;-)

For CI checks it is good anyway to iron out bugs like the one with g7x we have right now.

scripts/cppcheck.sh
hdiethelm: See https://github.com/LinuxCNC/linuxcnc/pull/4483 for the underlying reason.

hdiethelm: I figured out the issue for the needed -I:
In the cppckeck stage, we do not build before cppckeck. So `include/rtapi.h` does not exist. After build on master, mb2hal.c / hal_lib.c check passes and only emc/sai/driver.cc / emc/rs274ngc/gcodemodule.cc fails. This is why `CPPCHKOPT+=( "-I$(realpath "$(dirname "$0")/../src/rtapi")" )` or a build is needed. As an alternative, we could move the cppcheck to the rip-and-test job but I think missing includes will be found by the compiler, so the -I does not hurt.

src/emc/sai/driver.cc
hdiethelm: Basically I have no clue.
```
Checking src/emc/sai/driver.cc ...
Checking src/emc/sai/driver.cc: TOOL_NML...
src/emc/sai/driver.cc:567:22: warning: Potentially invalid type conversion in old-style C cast, clarify/fix with C++ cast [dangerousTypeCast]
  tool_nml_register( (CANON_TOOL_TABLE*)&_sai._tools);
```
vs
```
Checking src/emc/task/taskclass.cc ...
Checking src/emc/task/taskclass.cc: TOOL_NML..
```
Looks like a bug in cppcheck.

scripts/cppcheck.sh
BsAtHome: Well, with the mb2hal fix you don't need to add the include anymore. It will pass without any problem. Tested in a Ubuntu 26.04 VM.

src/emc/sai/driver.cc
hdiethelm: If I change:
```
-CPPCHKOPT+=( "--enable=warning,performance,portability" )
+CPPCHKOPT+=( "--enable=warning,performance,portability,style" )
```
it is reported. But I am not sure if you are ready to fix the few 1000's of errors this generates... ;-)

Chatbot says cppcheck only reports these type of casts if they are dangerous. But I did not find any reference for this or any information on what is dangerous.

```
Checking src/emc/task/taskclass.cc: TOOL_NML...
src/emc/task/taskclass.cc:189:24: style: C-style pointer casting [cstyleCast]
    tool_nml_register( (CANON_TOOL_TABLE*)&emcStatus->io.tool.toolTable);
```
vs
```
Checking src/emc/sai/driver.cc: TOOL_NML...
src/emc/sai/driver.cc:567:22: warning: Potentially invalid type conversion in old-style C cast, clarify/fix with C++ cast [dangerousTypeCast]
  tool_nml_register( (CANON_TOOL_TABLE*)&_sai._tools);
```

scripts/cppcheck.sh
hdiethelm: Did you test before or after build? A build creates include/rtapi.h after which the -I is not needed any more.

BsAtHome: Ha, after the build. That narrows it down to the actual error!

You cannot run cppcheck on a clean clone. You must have run `autogen.sh`, `configure` and `make headers`.

The include directory _must_ be populated before any checks are done. This is important because the include order is _paramount_. We have been working on fixing `""` and `<>` includes for some time.

#4461 actually changes the placement of all exportable headers to move them permanently to the include directory. But that is not yet done and may require some more tweaking. Until the header copy rule is redundant (`make headers`), we must configure and copy them before cppcheck is run. That also means that all the necessary package dependencies need to be installed to make configure happy.

The only alternative to running a install deps/configure/make headers cycle is to extract the SRCHEADERS variable from the `Makefile` and copy them to include manually (until #4461 lands). That may be a lot faster than installing dependencies.

src/emc/sai/driver.cc
BsAtHome: It probably can't infer the type in taskclass.cc (because of the `->`) and can in driver.cc using local type tracking.

We'll leave the "style" issue for now... If you don't mind ;-)

scripts/cppcheck.sh
hdiethelm: Yes, it caught me also, took me some time to figure out that a clean repo vs. a repo after build makes a difference... ;-)

I would go an other route: Add all include paths that are used during build. cppcheck does not really care where the headers are but if it finds them, it extracts information to check the file. Right now, many headers are not found that are in the include paths during build. This can be verified by adding `--enable=missingInclude`. But then there are also complains about missing system headers. If you add them, cppcheck complains about issues in system headers, also not good.

During build, the includes are anyway checked by the compiler, so this should be no issue if the include path is to long. Except might be if there are equally named headers that are normally hidden by the include paths.

An alternative is, to use `cppcheck --project=compile_commands.json ...`. By using `bear -- make -j$(nproc)`, this can be created.

Otherwise, just switch to clang-tidy that does basically the same but is better documented. However, it's will take some time to set that up.

For now, to not make the CI to slow, I would just add this single include path and call it a day. It should not hurt. The rtapi.h is only copied to include/ right? So no change at all if I do this.

src/emc/sai/driver.cc
hdiethelm: Or there is some crude logic in cppcheck that thinks one of them is good. Fine for me.

scripts/cppcheck.sh
hdiethelm: Hmm, thinking about it, make headers should be done. There are headers also from other places, so only adding src/rtapi is not enough. I will do this.

BsAtHome: You can do in the .github workfile:
```bash
cd src  # Need to be in src directory
for i in $(get_srcheaders.awk < Makefile); do
    cp $i ../include/$(basename $i)
done
```

The `get_srcheaders.awk` script (needs chmod +x):
```awk
#!/usr/bin/awk -f
BEGIN { collect = 0; }

{
    if(collect == 1) {
        line = line $0;
        if(!match($0,/\\\s*$/)) {
            collect = 0;
        }
    }
}

/^\s*SRCHEADERS\s*[:]?=/ {
    collect = 1;
    gsub(/^\s*SRCHEADERS\s*[:]?=/, "");
    line = $0;
}

END { gsub(/\\/, "", line); print line; }
```

hdiethelm: So:
https://github.com/LinuxCNC/linuxcnc/compare/efcadca4fff32eb332b19b8b86c11e65be20ed61..a016563de59c3adeae2069510aad3b375a377eaf

And a CI mistake:
https://github.com/LinuxCNC/linuxcnc/compare/a016563de59c3adeae2069510aad3b375a377eaf..efab928fb7f8395ffec6558dbb8f8dd54850521f

Now the headers are created. After https://github.com/LinuxCNC/linuxcnc/pull/4461 goes in, this can be simplified again. But it should also not hurt, except the datarate on the ubuntu mirrors.

Your awk script is also nice but it will probably break with https://github.com/LinuxCNC/linuxcnc/pull/4461.

scripts/cppcheck.sh
```diff
@@ -15,7 +15,14 @@ EXHAUSTIVE=$(cppcheck --check-level=exhaustive --version > /dev/null 2>&1 && ech
...
 CPPCHKOPT+=( "--enable=warning,performance,portability" )
-CPPCHKOPT+=( "-I$(realpath "$(dirname "$0")/../include")" )
+
+INCLUDE_DIR=$(realpath "$(dirname "$0")/../include")
+if [ ! -f "$INCLUDE_DIR/linuxcnc.h" ]; then
+  echo "$INCLUDE_DIR/linuxcnc.h does not exist. Perform at least 'make headers' before calling cppcheck.sh"
+  exit 1
+fi
+CPPCHKOPT+=( "-I$INCLUDE_DIR" )
+
```
BsAtHome: Nothing needs to be changed here. This will fix itself when #4461 lands and then we need to undo all of this again. The include stuff in cppcheck.sh should stay the same.

scripts/cppcheck.sh
BsAtHome: Fine, but we don't need any changes in cppcheck.sh then.

scripts/cppcheck.sh
hdiethelm: Ok, so I take your AWK way that is easier to revert and will fail.

BsAtHome: No, your take is better, to install deps and run `make headers`. That will become a no-op later on, so that will not be a problem. Cleanup may be done later or together with #4461.

BsAtHome: However, my comment here stands. The cppcheck.sh script works without the changes now that you do a make headers. Why should we test then? For a lone user? Yeah, Right...

hdiethelm: Ok, done.

hdiethelm:
So, all checks passed and it looks good from my side.

scripts/cppcheck.sh
hdiethelm: BTW: The idea of the checks was for dev's cloning the repo and doing cppckeck without build and then wondering why they fail and after build they pass (like me and you). However, assuming https://github.com/LinuxCNC/linuxcnc/pull/4461 comes in soon, this is not needed.

grandixximo:
> So that makes sense now. I wonder how people will feel about not being able to run CI built packages on 10 year old machines... ;-)

I don't like the sound of this, I think we should go for a fat binary, switching at run time 

Edit: 
actually released Debs are still V1 are they not? This is only a rip test build in CI, V3 are discarded binaries for rip builds tests in CI, or am I missing something?

BsAtHome:
> > So that makes sense now. I wonder how people will feel about not being able to run CI built packages on 10 year old machines... ;-)
> 
> I don't like the sound of this, I think we should go for a fat binary, switching at run time
> 
> Edit: actually released Debs are still V1 are they not? This is only a rip test build in CI, V3 are discarded binaries for rip builds tests in CI, or am I missing something?

Please wake me when we are distributing each program in bin as as WASM compiled docker image of a snap package in flatpack format to be run in a browser. Then we can be certain that processors are too fast and RAM prices too low.

hdiethelm:
Yes, exactly. All uploaded deb's are built in debian dockers which are most probably v1, otherwhise the FMA issue would have surfaced already. I was more thinking about other people building binarys directly in the ubuntu runner.

Never the less, this could become a topic for sid one day.

grandixximo:
We can always go back to punch cards ;-) At least the FMA rounding would be deterministic.

Seriously though: the v3 binaries never leave CI, the debs build in Debian containers at baseline v1. The day sid goes amd64v3 we can revisit; until then a fat binary is a solution shopping for a problem.

grandixximo:
@hdiethelm could you rebase this, I think I still have 7 missing links, but not broken so you should be 0 warnings after rebase, still we might probably want to merge #4487 first...

hdiethelm:
> @hdiethelm could you rebase this, I think I still have 7 missing links, but not broken so you should be 0 warnings after rebase, still we might probably want to merge #4487 first...

Done. Yes, #4487 should be merged first.

grandixximo:
Zero html checks warnings, that is good 👍

grandixximo:
@hdiethelm #4487 landed, you can rebase we decided to fix the code to support amd-x64-v3 builds

hdiethelm:
> @hdiethelm #4487 landed, you can rebase we decided to fix the code to support amd-x64-v3 builds

Rebased but the tests fail again. Or should I have left something in from your commits?

There was a new issue with cppcheck but I fixed it: https://github.com/LinuxCNC/linuxcnc/compare/e4d38633bfb9d0e16a12ba4e23a5ccdb74220d5e..bd118f511bd5c0ab8e2b5c81d888dffc2c571fa1

grandixximo:
The g71 failures are not in LinuxCNC code. Root cause is a GCC 15.2 wrong-code bug: [GCC PR124802](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=124802) ("[15 Regression] Wrong code at -O2 with AVX2 complex double fmaddsub pattern"). The SLP vectorizer emits `vfmaddsub132pd` with a wrong multiplier for complex-double expressions when FMA is available, which is the case on the ubuntu-26.04 runners: they install from the amd64v3 archive and gcc defaults to `-march=x86-64-v3`. The bug report has a reproducer; I reproduced the exact CI failure locally in an `ubuntu:26.04` container and verified the g71 tests pass at `-O2 -march=x86-64-v3` with default FP contraction once the SLP vectorizer is off, so the `interp_g7x` numerics are fine and need no flag.

Fixed upstream in GCC 15.3 (released 2026-06-12) and 16.1. Ubuntu resolute still ships `15.2.0-16ubuntu1`, and the gcc-16 snapshot in the archive (`r16-8246`) predates the fix; I verified it miscompiles the same way.

I am holding this PR until Ubuntu publishes 15.3 to resolute-updates. The jobs install build deps from the live archive on every run, so a re-run should go green the day it lands. I have a daily reminder watching the archive, so I will re-trigger CI when it happens.

hdiethelm:
How long did you dig to figure this out? ;-)

Ok, so we wait until GCC has been fixed.

BsAtHome:
Did you rebase this to current master? At least one of the tests should no longer fail with current master.

hdiethelm:
> Did you rebase this to current master? At least one of the tests should no longer fail with current master.

Rebased, got a new cppckeck errors, fixed 3 of them and missed 1. However, one xfail test error is gone. 

The last cppcheck error has to wait for tomorrow:
```
emc/rs274ngc/gcode_renderer.cc:91:10: warning: Struct 'PreviewData' does not have a copy constructor which is recommended since it has dynamic memory/resource management. [noCopyConstructor]
1/31 files checked 0% done
    free(attrs);
         ^
emc/rs274ngc/gcode_renderer.cc:91:10: warning: Struct 'PreviewData' does not have a operator= which is recommended since it has dynamic memory/resource management. [noOperatorEq]
Checking emc/rs274ngc/pyarrays.cc ...
    free(attrs);
         ^
```

src/emc/rs274ngc/gcode_renderer.cc
```diff
@@ -86,6 +86,10 @@ double GCodeRenderer::external_angle_units() {
 // PreviewData: storage, growth, and the Python object it is handed over in
 // ---------------------------------------------------------------------------
 
+PreviewData::PreviewData() {
+
+}
+
```
BsAtHome: I'd suggest to do this inline.

src/emc/rs274ngc/gcode_renderer.hh
```diff
@@ -134,7 +134,18 @@ struct ToolOffsetRecord {
 };
 
 struct PreviewData {
+    PreviewData();
     ~PreviewData();
```
BsAtHome: Make it `PreviewData() {};` No need not to inline.

src/emc/rs274ngc/gcode_renderer.hh
```diff
@@ -134,7 +134,18 @@ struct ToolOffsetRecord {
...
+
+    //Fixes cppcheck error:
+    // warning: Struct 'PreviewData' does not have a copy constructor which is recommended
+    //  it has dynamic memory/resource management. [noCopyConstructor]
+    // Struct 'PreviewData' does not have a operator= which is recommended since
+    //  it has dynamic memory/resource management. [noOperatorEq]
+    //Implement when needed
+    PreviewData(PreviewData& other) = delete;
+    PreviewData& operator=(PreviewData b) = delete;
```
BsAtHome: May I suggest:
```c++
struct PreviewData {
    PreviewData() {};
    // Structure cannot be copied by assignment or copy-constructed because it uses dynamic memory.
    PreviewData(const PreviewData &) = delete;
    PreviewData &operator=(const PreviewData &) = delete;
...
```

No need to add verbose comments. This is a known and common construct.

BsAtHome: And, BTWm you need the `const`

hdiethelm: Thanks, that makes it simpler.

PR #4481, opened 2026-08-27
optimization: Build the G-code preview in C++

alex-pres:
Follows PR##4293 

Adds C++ GCodeRenderer, which builds the whole preview - transform, arcs, taps, suppression, vertices, extents, lengths and event records - during gcode.parse and hands it over once, replacing the Python rendering engine. C++ handles Vec[9] -> X,Y,Z transform completely, following all interpretator details, and Python renders it into OpenGL using shaders.

Parsing is ~2.9x faster on move-shaped programs and peaks at less memory; and specially for arcs, the last per-move Python call kept in code, go from 1.58 s to 0.075 s at 1.28M segments.

Even without speed benefit it is anyway better because resulting code is much more clean and easy to follow. Previous code was hard to read, understand and test. Right now we can test GCodeRenderer engine directly, it is independent of presentation. 

## Benchmark results, compare to current master

Headless, `fractal-1M.ngc` (1,000,149 moves):

| | master | this branch |
|---|---|---|
| parse+build | 2.705 s | **0.732 s** (3.7x) |
| total load | 2.712 s | **0.745 s** |
| peak RSS | 166.8 MB | 163.5 MB |
| VBO bytes | 20,003,000 | 20,003,000 |
| CPU geometry | 28,004,200 | 28,004,200 |
| extents | identical | identical |

Byte-identical geometry out, ~3.7x less time to produce it.

Results could be much better if ngc file use arcs (current master calculates arcs in Python which was bad idea, after moving to C++ it is 100x faster).


BsAtHome:
What are all those json files doing in there? It seems you want to do some "bake" test, but I did not see a `test{.hal,.sh,}` in that directory. Did I miss it?
You also need to disable any test(s) that cannot run in the headless CI run. All GUI/Graphics related dependencies are not installed in the headless CI run.

Another point, you are "redecorating" the gcodemodule. Would it be an idea if we move this to pybind11 at the same time? Doing this work twice may be more work and may also be error prone.

alex-pres:
> What are all those json files doing in there? It seems you want to do some "bake" test, but I did not see a `test{.hal,.sh,}` in that directory. Did I miss it? You also need to disable any test(s) that cannot run in the headless CI run. All GUI/Graphics related dependencies are not installed in the headless CI run.

My fault, forgot to add test.sh - done now. JSON files are reference results for related .ngc files from another folder. 

No problem with headless - these tests are for C++ routines that do math only - no GUI, no graphics. 

> Another point, you are "redecorating" the gcodemodule. Would it be an idea if we move this to pybind11 at the same time? Doing this work twice may be more work and may also be error prone.

Any WIP on this or samples? 

Yes, I can work on pybind11 conversion, but any existing refrerences to such change in LinuxCNC  will be helpful. If any

alex-pres:
I agree test files do not look good. I will rework it to easily see what each case is testing. 

In time being, I'd appreciate feedback on the rest of PR.

alex-pres:
> Another point, you are "redecorating" the gcodemodule. Would it be an idea if we move this to pybind11 at the same time? > Doing this work twice may be more work and may also be error prone.

OK, I have found halquery is using pybind now. I will try to rework gcodemodule to use it too in this branch.

src/emc/rs274ngc/gcodemodule.cc
```diff
@@ -40,16 +40,20 @@
   */
 
 
-#include <sys/time.h>
-
 #include <Python.h>
-#include <structmember.h>
+#include <pybind11/pybind11.h>
+
```
BsAtHome, line 48: You don't need to include `Python.h` when using pybind11 (also in the header).

Also, you include it in the header(s), if you need them there and then you do not need them here.

alex-pres:
Reworked with pybind11 , it is a pleasure to work with. Resulting code is much easier to read and follow! Added some modern C++  which gcodemodule.cc deserve )  

Tests reworked too.

src/emc/rs274ngc/Submakefile
```diff
@@ -47,7 +47,7 @@ TARGETS += ../lib/librs274.so ../lib/librs274.so.0
 	@rm -f $@
 	$(CXX) -g $(LDFLAGS) $(PYTHON_EXTRA_LDFLAGS) -Wl,-soname,$(notdir $@) -shared -o $@ $^ -lstdc++ $(BOOST_PYTHON_LIB) $(PYTHON_LIBS) $(PYTHON_EXTRA_LIBS)
 
-GCODEMODULESRCS := emc/rs274ngc/gcodemodule.cc
+GCODEMODULESRCS := emc/rs274ngc/gcodemodule.cc emc/rs274ngc/gcode_renderer.cc
 PYSRCS += $(GCODEMODULESRCS)
```
BsAtHome, line 51: Does adding gcode_renderer.cc to gcodemodule add a graphics dependency to the gcodemodule? If so, then it must be split or the headless case will fail to work.

alex-pres: There is no actual OpenGL "rendering" in gcode_renderer.{cc,hh} 

It contains only math transform inside so it will run fine on headless.

BsAtHome: Check.

src/emc/rs274ngc/gcodemodule.cc
alex-pres: Done

src/emc/rs274ngc/Submakefile
alex-pres: > Check.

Yes, of course, my mistake!

Fixed with last commit.

alex-pres:
Last commit is breaking compatibility - returns tuples instead of arrays is not good idea, breaks code on Python side. I will revert this at Monday.

alex-pres:
> Last commit is breaking compatibility - returns tuples instead of arrays is not good idea, breaks code on Python side. I will revert this at Monday.

It was false alarm, problem not caused by this code. Everything seems to be OK and ready for review and testing. 

Current live testing results: loading of 418Mb .ngc file (real work file sent by customer, with G2 arcs): 25 seconds to first render.

grandixximo:
> 418Mb .ngc file (real work file sent by customer, with G2 arcs): 25 seconds

:drooling_face:

alex-pres:
> > 418Mb .ngc file (real work file sent by customer, with G2 arcs): 25 seconds
> 
> 🤤

Together with PR #4491 : 12 seconds! 


P.S. As said before AXIS [specially AXIS] is using not optimised file editor code that now eats more memory and CPU than graphical preview. So for benchmarking I cut it off. Worth to be optimised too, may be later.

alex-pres:
And with both #4491 and #4492 - _5.7 seconds_ to first frame on same file. 
Peak RSS after file load:  +1.04GB.

grandixximo:
Could you please fetch master, then merge or rebase and push, just to let the CI run again, and we shall get the shots of all the UIs to make sure nothing regressed.

alex-pres:
> Could you please fetch master, then merge or rebase and push, just to let the CI run again, and we shall get the shots of all the UIs to make sure nothing regressed.

Done. And squashed to one commit.

BsAtHome:
And the screenshots seem to be similar to what we see elsewhere.

src/rtapi/rtapi_string.h
```diff
@@ -48,16 +48,24 @@ RTAPI_END_DECLS
 #endif
 RTAPI_BEGIN_DECLS
 static inline size_t rtapi_strlcpy(char *dst, const char *src, size_t size) {
-    return rtapi_snprintf(dst, size, "%s", src);
+    size_t len = strlen(src);
```
grandixximo: Drop this? #4493 landed and this branch is rebased past it, so nothing here calls these any more.

lib/python/rs274/glcanon.py
```diff
@@ -67,9 +58,64 @@ def getter(self):
...
+        to be raised from Python; ``notify``; and the foam Z levels, which
+        feed the draw side. The ``hide``/``show`` depth is not among them:
+        the renderer counts it off the same text after this returns;
+      * ``change_tool`` - not for the record, which C++ writes, but because
+        the interpreter reads the canon's tool table for a G43 after it, and a
+        GUI's override is what moves the simulated spindle slot;
+      * the dwell colours, attached in :meth:`adopt_geometry` from the table
+        below, which is the one thing C does not carry.
+
+    The offsets, the rotation, the plane and the feed rate are still
```
grandixximo: This contradicts the renderer, which forwards none of them. Which one is the contract we are keeping?

lib/python/rs274/glcanon.py
```diff
@@ -67,9 +58,64 @@ def getter(self):
...
+    not one of them any more.
+    """
+
     lineno = -1
 
+    #: What the C side reads to choose the protocol, once, at parse start.
+    #: Must be the bool; the C side ignores any other value, and a canon that
+    #: sets it without a callable ``adopt_geometry`` is a TypeError rather
+    #: than a silent fall back to per-move callbacks.
+    use_gcode_renderer = True
```
grandixximo: Opting out gives `RuntimeError: parse_file interp_error`, because `Translated.straight_feed` calls a `straight_feed_translated` that no longer exists. Worth failing in `make()` with the flag named instead?

Also: an out-of-tree screen subclassing `GLCanon` to hook `dwell` or `straight_feed` now renders fine and is never called. Does that want a `BREAKING:` note in `updating-linuxcnc.adoc`, like the GL rewrite got?

lib/python/rs274/glcanon.py
```diff
@@ -387,125 +373,26 @@ def run_time(self, max_feed_rate):
...
-        self.xo = xo
-        self.yo = yo
-        self.zo = zo
-        self.ao = ao
-        self.bo = bo
-        self.co = co
-        self.uo = uo
-        self.vo = vo
-        self.wo = wo
+    # -- the callbacks still forwarded during a rendered parse --------------
```
grandixximo: Neither of these fires on a rendered parse, so `self.plane` and `self.feedrate` keep their class defaults. Dead now, or is something still meant to call them? Note `gcode.arc_to_segments()` reads both off the canon.

src/emc/rs274ngc/gcode_renderer.cc
```diff
@@ -0,0 +1,859 @@
...
+    py::handle canon(canon_ptr);
+    py::object flag;
+    try {
+        flag = canon.attr("use_gcode_renderer");
+    } catch(py::error_already_set &e) {
+        if(!e.matches(PyExc_AttributeError)) { e.restore(); return nullptr; }
+        return nullptr;                 // no attribute: the callback protocol
+    }
+    // Anything that is not a bool is not an opt-in - see the note on
+    // catch-all `__getattr__` in the header. Callback protocol, no complaint:
```
grandixximo: The header note this points at is a one-line fragment; the rationale went out with `MoveBatch`. Restore it?

src/emc/rs274ngc/gcode_renderer.cc
```diff
@@ -0,0 +1,859 @@
...
+        }
+    for(int j = 0; j < P3_COUNT; j++) {
+        r->data_->drawn[BOX_MIN][j] = 9e99;
+        r->data_->drawn[BOX_MAX][j] = -9e99;
+    }
+    r->data_->tool_numbers.push_back(0);         // ordinal 0 is None
+    if(!r->read_planes()) return nullptr;
+    // getattr-with-default swallows whatever the read raised, which is what a
+    // canon that simply has no progress hook needs.
+    r->progress_ = py::getattr(canon, "renderer_progress", py::none());
```
grandixximo, line 403: This keeps the canon and its whole program alive until the next parse. Clear it in `hand_over()`?

src/emc/rs274ngc/gcode_renderer.cc
```diff
@@ -0,0 +1,859 @@
...
+    for(int i = 0; i < P9_COUNT; i++) {         // xo, yo, zo, ao .. wo
+        snprintf(name, sizeof name, "%co", AXES[i]);
+        try {
+            r->tool_[i] = canon.attr(name).cast<double>();
+        } catch(py::error_already_set &e) {
+            e.restore();
+            break;
+        }
+    }
+    if(PyErr_Occurred()) return nullptr;
```
grandixximo: A canon missing any of `xo`..`wo` (the tool length offsets) now fails the parse. Should those be in the opt-in contract stated in the header?

src/emc/rs274ngc/gcode_renderer.cc
```diff
@@ -0,0 +1,859 @@
...
+        if(steps == 1 && !jump) {
+            pt = p2;
+        } else {
+            double t = (double)sub / (double)steps;
+            for(int k = 0; k < P9_COUNT; k++)
+                pt[k] = t * p2[k] + (1.0 - t) * p1[k];
+        }
+        write_vertex(pt, line_number, sub == 0 ? KIND_NOOP : cat, nullptr);
+    }
+    data_->cur9 = p2;
```
grandixximo, line 580: After a failed `reserve()` this still claims the move completed, so the next one records no jump. Return after the first failed `write_vertex` instead?

src/emc/rs274ngc/gcodemodule.cc
```diff
@@ -64,423 +69,257 @@ struct _inittab builtin_modules[] = {
...
+static py::tuple int_array(const int *arr, int sz) {
+    py::tuple res(sz);
+    for(int i = 0; i < sz; i++) res[i] = arr[i];
+    return res;
 }
-static PyObject *LineCode_mcodes(LineCode *l, void *) {
-    return int_array(l->mcodes, ACTIVE_M_CODES);
+
+static void linecode_register(py::module_ &m) {
+    py::class_<LineCode> c(m, "linecode");
```
grandixximo, line 112: `gcode.linecode()` was constructible before and is not now. Deliberate?

src/rtapi/rtapi_string.h
alex-pres: Sorry for this, kept merge from local benchmark testing. Reverted.

grandixximo: No apology needed ;-)

lib/python/rs274/glcanon.py
alex-pres: Fixed comments with commit 712df5719c22eda06b2173672ce6d36d844ab788

lib/python/rs274/glcanon.py
alex-pres: It was bad idea to opt-in opt-out with object property. 

Reworked now to use class hierarchy which cannot be opt-out . Commit a0235200479264db141af8ce8c05ddeb21d0bedf

To use per-move classic protocol, just do not subclass from GlCanon. PrintCanon is good example that it still works.

alex-pres: Also removed subclassing from Translated because there is no use for it now. Classic canon may still subclass from Translated to use it. 

Commit d635bb9c6c11bc2d2bdb5309c4b3b1f9271f4984

lib/python/rs274/glcanon.py
alex-pres: Removed now useless attrs 5b90d0784ae83cf01ebeccfc12242b338f44cffa

src/emc/rs274ngc/gcode_renderer.cc
alex-pres: removed any gettattrs here and comments, it now checks for parent class instead.

src/emc/rs274ngc/gcode_renderer.cc
alex-pres: fixed  e5468a8aa6ddaaa3b37736cd81f07b5ac448143e

src/emc/rs274ngc/gcode_renderer.cc
alex-pres: Removed all state and exchange from Python object, now it is minimal. 
Added example of minimal useful canon to use C++ renderer

As additional bonus GCodeRenderer can be used directly from C++ code (cosmetic changes will be necessary but generally nothing stops it). 

296fa8390531f8394ca10feafaa6fbb86a680ccb

src/emc/rs274ngc/gcodemodule.cc
alex-pres: yes, it was broken. restored and added a test 
09730f19a245fe57285ef82766c6e686d24eff43

src/emc/rs274ngc/gcode_renderer.cc
alex-pres: You're right that it sets `cur9` either way. It's dead state though — `write_vertex` bumps `interp_error`, and `append()` (gcode_renderer.hh:394) and `arc_feed` both bail on that, so no further move reaches `fill()`. Nothing reads it again.

Does it worth adding bool return v and check for each vertex if it is handled anyway? If yes, I will do that.

alex-pres:
Finished with changes, hope all mentioned is fixed now

grandixximo:
You might want to look at CI the clang is reporting failures, we have it fail on warnings, gotta clean those up.

alex-pres:
> You might want to look at CI the clang is reporting failures, we have it fail on warnings, gotta clean those up.

fixed with last commit.

lib/python/rs274/glcanon.py
grandixximo: Opt-out half is gone, thanks. The `updating-linuxcnc.adoc` note is still open, and the rework widened what it would cover: `lo`, `first_move`, `xo`..`wo`, `plane`, `feedrate`, `g5x_offset_*`, `g92_offset_*` and `g5x_index` are off `GLCanon` now too. Nothing in tree reads them, so only out-of-tree screens would notice.

src/emc/rs274ngc/gcode_renderer.cc
```diff
@@ -0,0 +1,846 @@
...
+            "to send as initcode, a `G53 G0` per axis that the leading-traverse\n"
+            "drop repositions on rather than draws. Everything else such a\n"
+            "canon needs is the ordinary canon protocol, which is not this\n"
+            "protocol's business: next_line, comment, message, change_tool,\n"
+            "check_abort, the get_* queries and parameter_file.";
+    ns["arcdivision"] = py::int_(64);
+    py::object cls = py::reinterpret_borrow<py::object>((PyObject *)&PyType_Type)(
+            "RendererCanon", py::tuple(), ns);
+    m.attr("RendererCanon") = cls;
+    renderer_canon_type = (PyTypeObject *)cls.ptr();     // the module owns it
```
grandixximo: Only the module attribute holds this alive, so `del gcode.RendererCanon` leaves the pointer dangling into `PyObject_TypeCheck`. `inc_ref()` here?

src/emc/rs274ngc/gcode_renderer.cc
grandixximo: No, not worth threading a bool through. One correction though: the bail happens on the next canon call, so the tail of the current move or arc still runs, `render_arc` loops its segments and `fill()` finishes its sub-steps without rechecking. Bounded and OOM-only either way, so leaving it is fine.

src/emc/rs274ngc/gcode_renderer.cc
alex-pres: It's hard to imagine why one would want to run that. `del gcode.RendererCanon` , fixed anyway

NTULINUX (approved):
Very awesome, thank you!

alex-pres:
[BsAtHome](https://github.com/BsAtHome) sorry, I clicked "request review" by mistake.

alex-pres:
Pushed documentation update, must be good to go now.

alex-pres:
Last commit adds ability for Python code to request actual machine axis positions array along with usually returned visual XYZ points. It is optional and disabled by default. It may be useful for advanced rendering or program analysis.

alex-pres:
Tool offsets is useful to build correct re-presentation based on AXIS coordinates [ like vismach ] . So it is returned too as sorted tuples, records mark line range affected.  ProgramGeometry in python knowns how to parse it.

alex-pres:
> Tool offsets is useful to build correct re-presentation based on AXIS coordinates [ like vismach ] . So it is returned too as sorted tuples, records mark line range affected. ProgramGeometry in python knowns how to parse it.

Just pushed and understood that implementation is invalid - it must index not line# but idx# in move array. It'd be incorrect if tool offsets change mid line. Not sure if that is possible but anyway better to fix. I will submit better implementation next week.

src/emc/rs274ngc/gcode_renderer.cc
```diff
@@ -0,0 +1,986 @@
...
+
+#include <algorithm>
+#include <math.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "nml_intf/canon.hh"
+#include "posemath.h"
+
```
BsAtHome: This should read `#include <posemath.hh>` or `#include <posemath.h>`.
The include is an exported include that lives in the include directory.

src/emc/rs274ngc/gcode_renderer.cc
```diff
@@ -0,0 +1,986 @@
...
+    rotation_xy_ = degrees;
+    // `M_PI / 180.0` folded first, which is what `math.radians` multiplies by:
+    // the fill this replaced took its sin and cos from the canon, so keeping
+    // the argument bit-identical keeps the baked expectations so too.
+    double rad = degrees * (M_PI / 180.0);
+    rotation_cos_ = cos(rad);
+    rotation_sin_ = sin(rad);
+    // The angle back, as it has always been computed here.
+    double back = -degrees * M_PI / 180.0;
+    unrot_cos_ = cos(back);
```
BsAtHome: First you use parentheses `(M_PI / 180.0)` and not in the second. But why do you need the second calculation using `-degrees`? You could simply write `unrot_cos(-rad)` and `unrot_sin(-rad)`?
And, since `sin(-a) == -sin(a)` (odd function) and `cos(-a) == cos(a)` (even function) you can reduce the complexity even more.

src/emc/rs274ngc/gcode_renderer.cc
```diff
@@ -0,0 +1,986 @@
...
+
+bool PreviewData::reserve(size_t extra) {
+    size_t need = n + extra;
+    if(need <= cap) return true;
+    size_t want = cap * 2;
+    if(want < 1024) want = 1024;
+    while(want < need) want *= 2;
+    float *grown[MAX_PLANES] = {};
+    for(int i = 0; i < nplanes; i++) {
+        grown[i] = (float*)realloc(pos[i], want * P3_COUNT * sizeof(float));
```
BsAtHome: You should probably not use `sizeof(float)`, but `sizeof(*grown[0])`, which makes the size determination dependent of the actual type of the variable.
(several other places too)

You also lose the access to the previous pointer when realloc() returns NULL and will leak.

This is also C++ code. So why is this not using `std::vector<float>`?

Also, why is this float? Is that because of GL?

src/emc/rs274ngc/gcode_renderer.cc
```diff
@@ -0,0 +1,986 @@
...
+    std::unique_ptr<GCodeRenderer> r(new GCodeRenderer(canon));
+    r->data_ = new PreviewData();
+    for(int i = 0; i < EXTENT_KINDS; i++)
+        for(int j = 0; j < P3_COUNT; j++) {
+            r->data_->extents[i][BOX_MIN][j] = 9e99;
+            r->data_->extents[i][BOX_MAX][j] = -9e99;
+        }
+    for(int j = 0; j < P3_COUNT; j++) {
+        r->data_->drawn[BOX_MIN][j] = 9e99;
+        r->data_->drawn[BOX_MAX][j] = -9e99;
```
BsAtHome: You should use `+DBL_MAX` and `-DBL_MAX` for double or `+FLT_MAX` and `-FLT_MAX` for float.

tests/gcode-renderer/test_bake.py
```diff
@@ -0,0 +1,438 @@
+"""The parts built from a finished program record: markers, solids, palettes.
+
```
BsAtHome, line 3: This (and the other python programs) should have a shebang:
`#!/usr/bin/env python3`

It should also be `chmod +x`

tests/gcode-renderer/test.sh
```diff
@@ -0,0 +1,42 @@
...
+if python3 -c "import rs274.glcanon" 2>/dev/null; then
+    # The renderer: its transform against that reference, then its behaviour
+    # and the shape of the record it hands over.
+    python3 test_transform.py >&2
+    python3 test_renderer.py >&2
+    python3 test_record.py >&2
+
+    # What readers downstream ask a finished program.
+    python3 test_queries.py >&2
+else
```
BsAtHome: The test programs should be invokable using `./test_transfom.py` syntax.
The test starts with the test's directory as CWD.

BsAtHome:
BTW, you need to rebase because there is a conflict now.

src/emc/rs274ngc/gcode_renderer.cc
alex-pres: I'll fix possible leak here. 

I'd prefer vector here but this code is specially optimised for big arrays. I just ran some benchmarks to check: 

I replicated the exact growth pattern — four arrays doubling in lockstep, then the `shrink()` — in the trixie container (glibc, `g++ -O2`):

| moves | realloc peak RSS | vector peak RSS | realloc grow time | vector grow time |
|---|---|---|---|---|
| 1M | 48.8 MB | 63.7 MB | 6.2 ms | 9.1 ms |
| 8M | 369 MB | 492 MB | 30 ms | 80 ms |

Final arrays are 366 MB at 8M moves, so **realloc peaks at essentially the final size while vector peaks 33% above it**. That is the whole story: a `vector` reallocation must hold the old and new blocks at once, so the last doubling costs final + half-final, and `shrink_to_fit` pays it a second time. `realloc` on a block that big is served by `mremap`, which re-points page tables and copies no bytes, so the peak never bulges. For an 8M-move program that is 122 MB you don't have to have free — and given the rewrite's selling point was 2.35× less memory than master, that is not a margin to hand back.

Unfortunately C++ will not allow to implement it C++ way : _One thing we cannot do is get vector to match: `std::allocator` has no reallocate hook, so a custom allocator won't buy grow-in-place — we will have to rewrite the container. The only portable way to close the gap would be reserving the final size upfront, which we can't know before the parse runs._

alex-pres: > Also, why is this float? Is that because of GL?

Float is because it is 2x time less memory to consume in GPU and also we don't need to be exact here - these numbers are only for representation.

BsAtHome: > Float is because it is 2x time less memory to consume in GPU and also we don't need to be exact here - these numbers are only for representation.

Understood.

BsAtHome: Well, I didn't know that vector operation was that much slower. Again shows the strength of pure C ;-)

> Final arrays are 366 MB at 8M moves, so **realloc peaks at essentially the final size while vector peaks 33% above it**.

But for allocation at that size, the operation is by mmap and the pages are only put into the RSS when touched. Does vector actually touch these pages? That's be bad, I guess. I wonder how large the difference between DRS and RSS is with these allocations.

But I'll consider that purely academic because speed matters here. And when malloc/realloc does a better job, then that be it. Rewriting the vector allocator, even with its explicit possibility, is a pain we can avoid.

src/emc/rs274ngc/gcode_renderer.cc
alex-pres: I'll do this but it may break existing third-party code that rely on this. These constants were used in python code . Though I did not found any use of this in current code.

alex-pres: Also , the legacy "calc_extents" in gcodemodule.cc also used 9e99 for years. may be just keep it as is ? comparing to DBL_MAX in python code is not one word.

src/emc/rs274ngc/gcode_renderer.cc
alex-pres: > You also lose the access to the previous pointer when realloc() returns NULL and will leak.

Checked for leaks. There is a grown[..] array specially to prevent this. if `realloc` failed, it will only use previous "grown" values  and it will exit realloc and parse will stop once limit of points is reached. already allocated memory will be freed as usual.

src/emc/rs274ngc/gcode_renderer.cc
alex-pres: fixed with 15a6a0ccc5765c48fa1d36b008e1eed0641de3b1

src/emc/rs274ngc/gcode_renderer.cc
alex-pres: it is rarely called [ coordinate system rotation gcode only ] . fixed anyway here 15a6a0ccc5765c48fa1d36b008e1eed0641de3b1

src/emc/rs274ngc/gcode_renderer.cc
alex-pres: fixed sizeof(..) shape 15a6a0ccc5765c48fa1d36b008e1eed0641de3b1

src/emc/rs274ngc/gcode_renderer.cc
BsAtHome: Comparing equality with these large values is always wrong. You can only use them for min/max operation. They are really bad as a signalling value. Its value is inexact (~8.9999999999999999489e+99).

The gcodemodule will use min/max on them. The question is whether any of the values can remain at the init value. If not, then they also should be fixed. If they can, the we need to examine whether the values are used anywhere.

tests/gcode-renderer/test_bake.py
alex-pres: fixed 58965d95a2e8d69171984c7e50c6882a750cf9de

tests/gcode-renderer/test.sh
alex-pres: fixed 58965d95a2e8d69171984c7e50c6882a750cf9de

alex-pres:
> BTW, you need to rebase because there is a conflict now.

done

src/emc/rs274ngc/gcode_renderer.cc
alex-pres: done 9dd8985a19fe415e85763508221826b06da65012

alex-pres:
Probably need one more step here: I almost implemented TimeEstimateCanon in C++ which does close-to-real time estimate during parse and return it to GUI. 
Compared in SIM and got ~98% time match which I believe is quite enough.  

It would deserve separate PR but because it changes gcodemodule.cc internals it may be here too. Let me know if separate PR is better choice.

BsAtHome:
> Probably need one more step here: I almost implemented TimeEstimateCanon in C++ which does close-to-real time estimate during parse and return it to GUI. Compared in SIM and got ~98% time match which I believe is quite enough.

A 98% match is better than most people can guess, I guess.
Unless we're talking about a long tail that is "off by progress dialogs from a know Redmond software firm pretending it could calculate time". If that is not the case, then we'd be stupid to complain at this stage.

> It would deserve separate PR but because it changes gcodemodule.cc internals it may be here too. Let me know if separate PR is better choice.

I think this PR has been boiling quite enough for now. The speed improvement is a good thing. Improving on it can be the next update.

alex-pres:
I completely agree: this PR is ready to be served.

I'll add one small refactoring : split GcodeRenderer to 2 classes, one that is base and second is preview-targeted. And it will be enough for this branch and ready to merge. 

Next PR for time estimate will use the base class to do its own job, and it will be add-on only not touching the preview itself, I think it will be best that way. 

And  time estimate definitely will work much better than Win 3.11 install progress )

BsAtHome:
And please squash the commits, if you can.

alex-pres:
After trying I decided to abandon idea of the class split - code becomes too noisy and hard to follow. So I squashed all commits  already reviewed. 

Last problem I found accidentally - LineCode pybind11 rewrite slow down it. It does not affect new rendering code, but those who will use old python canon protocol (where all moves are reported to python methods) will be affected. And it is not cosmetic (numbers below). So I rewrote it back to C++ struct shared to python without pybind11.  I made this a separate commit. After review, it is ready to merge [ I can do squash again ] 

## Benchmark

8M-move pure-G1 file, dev container, both `gcode.so` builds kept side by side and **alternated rep by rep** so drift lands on both. Minimum of 5 reps.

| canon's `next_line` | before | after | delta |
|---|---|---|---|
| drops the object | 5.04 s | 4.24 s | **−0.80 s** (−16 %) |
| reads `sequence_number` | 6.06 s | 4.40 s | **−1.67 s** (−27 %) |
| reads four fields | 8.60 s | 4.71 s | **−3.88 s** (−45 %) |

The first row is exactly the 0.80 s the plan attributed to the object. The other two are the part the plan listed as unmeasured, and it was the larger half: a field read went from **~125 ns** (pybind11 property lambda) to **~19 ns** (descriptor over a struct offset).
.

Issue #4480, opened 2026-08-27
Kinematics: Switchable kins should default to identity (joint) mode


Sigma1912:
Switchable kinematics  eg 'genserkins', 'pumakins' currently default to 'world'  (or 'genserkins') mode. 
If the kinematic model fails (eg in a singularity) the controller switches off and cannot be recovered because there is no way to switch to joint mode and move away from the problematic pose.

PR #4482, opened 2026-08-28
switchkins: allow starting in identity kinematics

grandixximo:
`sparm=identityfirst` makes switchkins-type 0 identity kinematics and type 1 the module's own. It exists in `xyzac-trt-kins`, `xyzbc-trt-kins` and `5axiskins`. `genserkins`, `pumakins`, `scarakins`, `three21kins` and `genhexkins` hardcode type 0 to their own kinematics, so those machines can only start in the kinematics that can fail. This adds the option to all five.

Refs #4480.

## Why type 0 is the one that matters

`motion.switchkins-type` is an input pin that rests at 0, and the shipped configs drive it from `motion.analog-out-03`, which is also 0 until an `M68` runs. Whatever is registered as type 0 is therefore both the startup type and the type a machine falls back to, so it is the one that has to solve at any pose.

Other controls treat the kinematic transformation as a modal state that is off unless the program turns it on: `TRAORI`/`TRAFOOF` on Sinumerik, `FUNCTION TCPM`/`FUNCTION RESET TCPM` on Heidenhain, `#TRAFO ON`/`#TRAFO OFF` on TwinCAT CNC, where `auto_enable_kin_trafo` decides what a program start selects. Robot controls take the other route and keep joint jog available at all times, which is the documented way out of a singularity on a FANUC. LinuxCNC has both routes; this closes the gap in the second one.

## Iterative forward kinematics

A type with iterative forward kinematics restarts from the pose it last produced, and that pose only accumulates while the type is running. A type that has never run therefore had nothing to start from. With identity kinematics selected first, genhexkins reported the hexapod platform at the origin instead of the place it is at, and switching to it adopted that position, so the next `G0` dropped the platform 20 mm.

switchkins now tracks whether it holds a pose for a type at all. Without one it starts from the estimate the caller passes, which motion already seeds from `[TRAJ]HOME`. That covers any iterative type, not just genhexkins.

A gui type that cannot solve also no longer fails the running type. Those pins drive a vismach display, and a stale reading there was being reported to motion as a forward-kinematics failure of the type actually running.

## Position after a failed switch

`handle_kinematicsSwitch()` ran the new type's forward kinematics straight into `carte_pos_cmd` and ignored the result, so a type that cannot solve at the current pose left whatever the failed solver had written there and `tpSetPos()` adopted it. The next move then ran from a position the machine is not at. It now solves into a local pose and commits only on success.

## Verification

Each sim config run twice, stock and with `sparm=identityfirst`, comparing the world position after switching to the module kinematics:

| config | module | stock type0 pose | identityfirst, after switch to type1 |
|---|---|---|---|
| puma560 | genserkins | 17.0038 5.5000 7.2000 180 -0.1 0 | identical |
| puma | pumakins | 450 100 -495 180 0 0 | identical |
| scara | scarakins | 640 0 465 0 0 0 | identical |
| melfa_321 | three21kins | 509.9893 0.0041 595.0000 -179.999 0.001 -0.001 | identical |
| hexapod | genhexkins | 0.0000 0.0003 20.0000 -0.0006 -0.0000 0.0000 | identical |

For the hexapod the `skgui.*` pins also read the true platform pose while identity kinematics are running, and switching back to identity returns the same joint values as the stock ordering.

`runtests`: 296 run, 296 successful, 0 failed, 3 skipped.

## Open question

The setting exchanges type 0 and type 1, so G-code or HAL that selects a type by number has to match. #4480 asks for identity to be the default rather than an option, which would mean that renumbering on every existing genserkins and pumakins config. Sinumerik makes the equivalent choice a machine-data bit, so a configurable default is not unusual, and I have kept this opt-in. Say the word if the default should flip instead.

@Sigma1912 one question on the report itself. In AXIS `$` toggles joint jog mode, and free mode never calls the inverse kinematics, so after the controller drops out you can re-enable and jog the joints clear without changing kinematics type. Which GUI were you on, and what stopped the recovery there? Starting in identity is worth having either way, I would just like to know whether there is a second problem underneath.


Sigma1912:
> Which GUI were you on, and what stopped the recovery there? 

You are of course correct about the `$` in AXIS, however I usually disable keyboard shortcuts on all my GUIs. 


Thanks for the usual prompt and thorough fix.

Sigma1912:
> Say the word if the default should flip instead.

No, I think a configurable default is the right choice. Startup behavior remains unchanged and the user can choose which should be the default.

PR #4483, opened 2026-08-28
mb2hal: Fix module info placement and cppcheck problem

BsAtHome:
The new cppcheck in Ubuntu 26.04 (for future CI, see #4477) would trip on a define test of `MODULE_VERBOSE`. The placement was wrong in the header and the usage should be replaced with `MODULE_INFO` as used everywhere else.

This PR addresses the issue. It also removes the `RTAPI` dependent inclusion check of `rtapi_app.h` (probably copy-pasted from a compiled component). This is a user-space program that does not need `rtapi_app_main()` and uses plain `main()`.

### DIFF
--- a/src/hal/user_comps/mb2hal/mb2hal.h
+++ b/src/hal/user_comps/mb2hal/mb2hal.h
@@ -7,9 +7,6 @@
 #include <pthread.h>
 
 #include <rtapi.h>
-#ifdef RTAPI
-#include <rtapi_app.h>
-#endif
 #include <rtapi_string.h>
 #include <rtapi_errno.h>
 #include <hal.h>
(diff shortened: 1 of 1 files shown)


grandixximo:
Good catch...

BsAtHome:
Well, there is still a problem. Just installed a Ubuntu 26.04 in a VM and it still complains. Need to look deeper what is happening.

BsAtHome:
The real issue here is that the `MODULE_INFO` stuff is only available when RTAPI is defined. However, that is never the case with user-space components and therefore, the whole `MODULE_xxx` exercise is a hopeless endeavor. It simply needs to be removed.

hdiethelm:
The issue looks to be how cppcheck handles `#ifdef`. cppcheck has to figure out which cases it should check.

You see that here, it sees `#ifdef MODULE_VERBOSE` and does a check with MODULE_VERBOSE=1 due to --force, which of course fails:
```
Checking src/hal/user_comps/mb2hal/mb2hal.c ...
Checking src/hal/user_comps/mb2hal/mb2hal.c: MODULE_VERBOSE...
src/hal/user_comps/mb2hal/mb2hal.h:36:1: error: syntax error [syntaxError]
MODULE_VERBOSE(emc2, "component:mb2hal:Userspace HAL component to communicate with one or more Modbus devices");
^
Checking src/hal/user_comps/mb2hal/mb2hal.c: RTAPI..
```

If there are missing includes, this generates more errors, due to cppckeck does not find KERNEL_VERSION and just assumes KERNEL_VERSION=1 for example which of course also fails. As soon as rtapi.h is found, problem solved.

See:
https://cppcheck.sourceforge.io/manual.html `Automatic configuration of preprocessor defines`

PR #4487, opened 2026-08-29
interp_g7x: make the G7x roughing cycle independent of rounding

grandixximo:
Follow-up to #4477, which hit G71 test failures on the ubuntu-26.04 runners because their gcc defaults to x86-64-v3 and contracts `a*b+c` into an FMA. That PR disables contraction for this file, which unblocks CI but leaves the fragility in place, as @BsAtHome pointed out.

## Root cause

The offsetting code decided discrete cases by comparing floats that are meant to be equal. Lathe profiles produce those ties by construction: an arc starting at its own apex, an exactly flat facing cut, a scan line landing on top of an arc. FMA is not the problem, it is just a disturbance small enough to expose one.

Measured by perturbing every coordinate in `tests/interp/g71-with-g70` by a relative 1e-12, six orders of magnitude below the 1e-6 tolerance the cycle works to:

| | diverging | hanging |
|---|---|---|
| master | 16/20 | 4/20 |
| this branch | 0/30 | 0 |

The hangs are the failure mode the `g71-endless-loop` tests exist for: with the apex tie broken the wrong way, `pocket()` re-enters the same pocket forever.

## Part size

The thresholds in the file are plain numbers, so they only fit a part of a few tens of millimetres. The tangential intersections the offsetting relies on carry an error of about `r*sqrt(2*eps)`, which passes the tolerance on a larger part. Same perturbation, profile scaled up:

| part size | master | this branch |
|---|---|---|
| 38 mm | 16/20 + 4 hangs | 0/20 |
| 190 mm | 10/20 | 0/20 |
| 305 mm | 20/20 | 0/20 |
| 3.8 m | 16/20 | 0/20 |
| 381 m | 19/20 | 0/20 |

The profile is normalised into a canonical range on entry and the tool path scaled back on output. The factor is a power of two, so it adds no rounding of its own, and it is 1 at the size the thresholds were chosen for. Working range goes from 20 to 114 mm, to roughly 0.4 mm to 381 m. Below that the interpreter rejects the arcs itself as zero radius, which is an error rather than a wrong path. Inch mode is covered by the same mechanism.

## Verification

- 86/86 in `tests/interp` and `tests/m70-m73`
- `interp_g7x.o` gives identical output built with `-march=x86-64-v3`, `-march=native`, `-ffp-contract=off`, `-mfpmath=387`, `-O0` and `-O3`
- both `nc_files/lathe_g7x_*` demos byte-identical to master
- timing unchanged, 1486 against 1488 ms per 1000 cycle sets

## Notes for review

- The first commit is the one already in #4477. It rebases away once that merges.
- Two lines are added to the `g71-with-g70` expected output, from making `straight_segment::climb_only()` use the same flatness test as `climb()`. That contour segment is now climbed like the rest; both lines are zero length repeats of moves already present, so the path is unchanged.
- The test dropping an arc that goes nowhere moves out of `motion_machine`, where it ran on scaled output against the interpreter's fixed `TOLERANCE_EQUAL`, into `scaled_motion`, where it sees the normalised profile. At unit scale it is the same decision on the same numbers.
- With this in, `-ffp-contract=off` is redundant. I left it in place because only CI can confirm that on gcc 15 with a v3 default. Happy to drop it in a separate commit if you would rather have CI prove the point.


BsAtHome:
> Working range goes from 20 to 114 mm, to roughly 0.4 mm to 381 m.

That would probably cover most use cases. Only machines for Dyson sphere construction and macro-molecular-manipulators would no longer fit without scaling on the hardware side. Can't have it all, I guess.

> * Two lines are added to the `g71-with-g70` expected output, from making `straight_segment::climb_only()` use the same flatness test as `climb()`. That contour segment is now climbed like the rest; both lines are zero length repeats of moves already present, so the path is unchanged.

Are these final position moves to end at the correct place and not a tolerance influenced end-point?


> * With this in, `-ffp-contract=off` is redundant. I left it in place because only CI can confirm that on gcc 15 with a v3 default. Happy to drop it in a separate commit if you would rather have CI prove the point.

I really don't think we should make exceptions for any file. This patch fixes the problem and that is the correct way to handle it. Any other places in the code that will be exposed as containing instabilities must get the same treatment. No problems should be hidden. And this patch fixes the problem, so why cover up for potential future problems?

grandixximo:
On the two added lines: they are not new positions. The path visits the same points in the same order; only the record count changes.

```
motion records:                   239 master, 241 branch
distinct successive positions:    222 master, 222 branch
position sequence identical:      yes
branch records with an (x,z) not already in master: none
```

Both are exact repeats of a position the path already occupies. The endpoints are the offset contour itself: X11.0000 is the programmed X10 plus the D1 allowance, Z-25.1780 and Z-27.1780 come straight from the profile. No tolerance enters the endpoint. The tolerance only decides whether `climb_only()` traverses that segment at all, and it now agrees with `climb()` that a segment flat to within the tolerance is climbing.

On the contraction flag: agreed, dropped in the commit on top, Submakefile is back to identical with master. Kept separate from the earlier commit because that one is also in #4477 and has to stay byte identical there to rebase away cleanly.

Note that this PR builds on ubuntu-24.04, so its own CI never emits an FMA and cannot exercise the case. To prove the maths instead of waiting on CI I wrote a small harness that injects perturbations far larger than contraction does: 100 lines, Python and shell, one script rewriting every coordinate in a .ngc by a given relative amount and one comparing the resulting tool paths numerically. At a relative 1e-12, four orders of magnitude above what contraction changes, master diverges in 16 of 20 runs and hangs in 4; the branch is unchanged in 30 of 30. A second pair of scripts, 51 lines, sweeps the same profile across part sizes. Happy to put either in tree if you want it, though it needs a tolerance-comparing `checkresult` rather than `cmp`, since the 4-decimal output has rounding ties that move under any perturbation. Once #4477 lands, 26.04 CI becomes the real guard.

PR #4488, opened 2026-08-29
fix safety and correctness bugs in hostmot2 drivers

sleepybishop:
Hello! This PR fixes some subtle safety and correctness bugs in the hostmot2 HAL drivers.

A few things this cleans up:
- Hardened socket send/receive checks in `hm2_eth` to ensure packet sizes match exactly, and added some bounds checking to prevent overflows.
- Plugged a firmware memory leak on failure in `setsserial.c` and made `krealloc` pointer handling safer across the board.
- Fixed some integer overflows when bit-shifting in `pktuart.c`.
- Fixed a copy-paste bug in the Three-Phase PWM config (`tp_pwmgen.c`) and an off-by-one bug in `hm2_sserial_cleanup`.

Please let me know if you have any feedback or see any issues with these fixes!

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/abs_encoder.c
+++ b/src/hal/drivers/mesa-hostmot2/abs_encoder.c
@@ -262,13 +262,20 @@ int hm2_absenc_parse_format(hm2_sserial_remote_t *chan,  hm2_absenc_format_t *de
                         "not get the pins you expected\n");
             }
             else if (strchr("bBuUsSeEfFpPgGhHlLmM", *format)){
+                hm2_sserial_data_t *confs;
                 hm2_sserial_data_t *conf;
-                chan->num_confs++;
-                chan->confs = (hm2_sserial_data_t *)rtapi_krealloc(chan->confs,
-                        chan->num_confs * sizeof(hm2_sserial_data_t),
-                        RTAPI_GFP_KERNEL);
 
-                conf = &(chan->confs[chan->num_confs - 1]);
+                if (q > RTAPI_UINT8_MAX) {
+                    HM2_ERR_NO_LL("Field length %d exceeds the supported maximum\n", q);
+                    return -EINVAL;
+                }
+                confs = rtapi_krealloc(chan->confs,
+                        (chan->num_confs + 1) * sizeof(*confs),
+                        RTAPI_GFP_KERNEL);
+                if (confs == NULL) return -ENOMEM;
+                chan->confs = confs;
+
+                conf = &(chan->confs[chan->num_confs++]);
                 conf->DataDir = LBP_IN;
                 conf->DataLength = q;
                 rtapi_strxcpy(conf->NameString, name);
@@ -357,6 +364,10 @@ int hm2_absenc_parse_format(hm2_sserial_remote_t *chan,  hm2_absenc_format_t *de
         else
         {
             // Not a % format, append name
+            if (nameptr == &name[sizeof(name) - 1]) {
+                HM2_ERR_NO_LL("Absolute encoder field name is too long\n");
+                return -EINVAL;
+            }
             *nameptr++ = *format++;
             *nameptr = 0;
         }
@@ -400,12 +411,6 @@ int hm2_absenc_parse_md(hostmot2_t *hm2, int md_index) {
 
     if (hm2->absenc.num_chans == 0) { // first time though
         hm2->absenc.clock_frequency = md->clock_freq;
-        hm2->absenc.ssi_busy_flags = rtapi_kmalloc(sizeof(rtapi_u32), RTAPI_GFP_KERNEL);
-        *hm2->absenc.ssi_busy_flags = 0;
-        hm2->absenc.biss_busy_flags = rtapi_kmalloc(sizeof(rtapi_u32), RTAPI_GFP_KERNEL);
-        *hm2->absenc.biss_busy_flags = 0;
-        hm2->absenc.fabs_busy_flags = rtapi_kmalloc(sizeof(rtapi_u32), RTAPI_GFP_KERNEL);
-        *hm2->absenc.fabs_busy_flags = 0;
     }
     
     for (index = 0 ; index < md->instances ; index ++){
@@ -422,12 +427,18 @@ int hm2_absenc_parse_md(hostmot2_t *hm2, int md_index) {
                 goto fail1;
             }
             if (index == def->index && md->gtag == def->gtag){
+                hm2_sserial_remote_t *chans;
+
                 has_format = true;
-                hm2->absenc.num_chans += 1;
-                hm2->absenc.chans = rtapi_krealloc(hm2->absenc.chans,
-                        hm2->absenc.num_chans * sizeof(hm2_sserial_remote_t),
+                chans = rtapi_krealloc(hm2->absenc.chans,
+                        (hm2->absenc.num_chans + 1) * sizeof(*chans),
                         RTAPI_GFP_KERNEL);
-                chan = &hm2->absenc.chans[hm2->absenc.num_chans - 1];
+                if (chans == NULL) {
+                    HM2_ERR("out of memory allocating absolute encoder channel\n");
+                    goto fail1;
+                }
+                hm2->absenc.chans = chans;
+                chan = &hm2->absenc.chans[hm2->absenc.num_chans++];
                 memset(chan, 0, sizeof(hm2_sserial_remote_t));
                 chan->index = index;
                 chan->myinst = md->gtag;
@@ -517,6 +528,8 @@ void hm2_absenc_process_tram_read(hostmot2_t *hm2, long period) {
                 err_tag[i] = 1;
             }    
         } 
+        // These pointers are supplied by hm2_absenc_register_tram() whenever
+        // the corresponding global-start address enables this dereference.
         if (((chan->myinst == HM2_GTAG_SSI)
                 && (*hm2->absenc.ssi_busy_flags & (1 << chan->index)))
                 || ((chan->myinst == HM2_GTAG_BISS)
--- a/src/hal/drivers/mesa-hostmot2/bspi.c
+++ b/src/hal/drivers/mesa-hostmot2/bspi.c
@@ -121,6 +121,10 @@ int hm2_tram_add_bspi_frame(char *name, int chan, rtapi_u32 **wbuff, rtapi_u32 *
         HM2_ERR_NO_LL("Can not find BSPI instance %s.\n", name);
         return -1;
     }
+    if (chan < 0 || chan > 15) {
+        HM2_ERR("BSPI %s: channel number %i is out of range\n", name, chan);
+        return -EINVAL;
+    }
     if (hm2->bspi.instance[i].conf_flag[chan] != true){
         HM2_ERR("The selected write channel (%i) on bspi instance %s.\n" 
                 "Has not been configured.\n", chan, name);
@@ -205,6 +209,10 @@ int hm2_bspi_write_chan(char* name, int chan, rtapi_u32 val)
         HM2_ERR_NO_LL("Can not find BSPI instance %s.\n", name);
         return -1;
     }
+    if (chan < 0 || chan > 15) {
+        HM2_ERR("BSPI %s: channel number %i is out of range\n", name, chan);
+        return -EINVAL;
+    }
     if (hm2->bspi.instance[i].conf_flag[chan] != true){
         HM2_ERR("The selected write channel (%i) on bspi instance %s.\n" 
                 "Has not been configured.\n", chan, name);
@@ -246,6 +254,10 @@ int hm2_bspi_setup_chan(char *name, int chan, int cs, int bits, double mhz,
         HM2_ERR("BSPI %s: Number of bits for chan %i (%i) is out of range, "
                 "BSPI only supports 1-64 bits\n", name, chan, bits);
         return -1;
+    }
+    if (mhz <= 0) {
+        HM2_ERR("BSPI %s: clock frequency must be greater than zero\n", name);
+        return -EINVAL;
     }
     if (delay < 0 || delay > 1e6){
         HM2_ERR("The requested frame delay on channel %i of %inS seems "
@@ -388,6 +400,3 @@ void hm2_bspi_write(hostmot2_t *hm2)
 {
     (void)hm2;
 }
-
-
-
--- a/src/hal/drivers/mesa-hostmot2/bspi.c
+++ b/src/hal/drivers/mesa-hostmot2/bspi.c
@@ -123,7 +123,7 @@ int hm2_tram_add_bspi_frame(char *name, int chan, rtapi_u32 **wbuff, rtapi_u32 *
     }
     if (chan < 0 || chan > 15) {
         HM2_ERR("BSPI %s: channel number %i is out of range\n", name, chan);
-        return -EINVAL;
+        return -1;
     }
     if (hm2->bspi.instance[i].conf_flag[chan] != true){
         HM2_ERR("The selected write channel (%i) on bspi instance %s.\n" 
@@ -211,7 +211,7 @@ int hm2_bspi_write_chan(char* name, int chan, rtapi_u32 val)
     }
     if (chan < 0 || chan > 15) {
         HM2_ERR("BSPI %s: channel number %i is out of range\n", name, chan);
-        return -EINVAL;
+        return -1;
     }
     if (hm2->bspi.instance[i].conf_flag[chan] != true){
         HM2_ERR("The selected write channel (%i) on bspi instance %s.\n" 
@@ -257,7 +257,7 @@ int hm2_bspi_setup_chan(char *name, int chan, int cs, int bits, double mhz,
     }
     if (mhz <= 0) {
         HM2_ERR("BSPI %s: clock frequency must be greater than zero\n", name);
-        return -EINVAL;
+        return -1;
     }
     if (delay < 0 || delay > 1e6){
         HM2_ERR("The requested frame delay on channel %i of %inS seems "
--- a/src/hal/drivers/mesa-hostmot2/hm2_eth.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_eth.c
@@ -1418,23 +1418,8 @@ static int hm2_eth_probe(hm2_eth_t *board) {
         // DB25, 17 pins used, IO 34 to IO 50
         board->llio.ioport_connector_name[0] = "P1";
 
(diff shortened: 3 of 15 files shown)


grandixximo:
Hardware tested?

sleepybishop:
> Hardware tested?

Yes, but my own not mesa hardware. Most of my robots use a mix of artix and ecp5's. I do have an old cnc router with a 6i25 but that machine is currently buried under a pile of boxes.

src/hal/drivers/mesa-hostmot2/pktuart.c
```diff
@@ -300,24 +301,24 @@ static int config_tx(const char *name, const hostmot2_t* hm2, hm2_pktuart_instan
 	}
 
 	if((cfg->flags & HM2_PKTUART_CONFIG_FORCECONFIG) || bitrate != inst->tx_bitrate) {
-			inst->tx_bitrate = bitrate;
-		if((r = writefn(hm2->llio, inst->tx_bitrate_addr, &bitrate, sizeof(bitrate))) < 0) {
+		if(!(r = writefn(hm2->llio, inst->tx_bitrate_addr, &bitrate, sizeof(bitrate)))) {
```
grandixximo: This conversion is repeated about forty times across `pktuart.c`, `uart.c` and `bspi.c`, and it assumes every llio `read`/`write` returns 1 or 0. `hm2_spi.c` does not:

```c
static int do_read(hm2_lowlevel_io_t *llio, rtapi_u32 addr, void *buffer, int size) {
    hm2_spi_t *this = (hm2_spi_t*) llio;
    int r = queue_read(llio, addr, buffer, size);
    if(r < 0) return r;
    return do_pending(this);
}
```

`queue_read`/`queue_write` return `-EINVAL` on a misaligned size, and `do_pending()` returns `-errno` when the `SPI_IOC_MESSAGE` ioctl fails. A negative value is truthy, so `!r` reads a genuine SPI failure as success; the `r < 0` test it replaces caught exactly that case. Net effect: error detection for pktuart, uart and bspi disappears on Raspberry Pi `hm2_spi` boards.

`r <= 0` covers both conventions. Worth auditing the other llio backends before settling on one.

src/hal/drivers/mesa-hostmot2/bspi.c
```diff
@@ -187,11 +191,12 @@ int hm2_bspi_clear_fifo(char * name)
     }
     rtapi_u32 zero = 0;
     r = hm2->llio->write(hm2->llio, hm2->bspi.instance[i].count_addr, &zero, sizeof(rtapi_u32));
-    if (r < 0) {
+    if (!r) {
```
grandixximo: Beyond the `!r` question above, this also flips the success return from 1 to 0, and both `hm2_bspi_clear_fifo` and `hm2_bspi_write_chan` are `EXPORT_SYMBOL_GPL`. In tree only `mesa_7i65` calls them and it tests `< 0`, so nothing here breaks. Is the change intentional for out-of-tree callers, and if so should `hostmot2-serial.h` or the man page state the new contract?

src/hal/drivers/mesa-hostmot2/pktuart.c
```diff
@@ -867,11 +882,12 @@ int hm2_pktuart_queue_get_frame_sizes(const char *name, rtapi_u32 fsizes[])
...
 		r = hm2->llio->queue_read(hm2->llio, hm2->pktuart.instance[inst].rx_fifo_count_addr,
 				&fsizes[j], sizeof(rtapi_u32));
-		if(r < 0) {
+		if(!r) {
 			HM2_ERR("Unable to queue Rx FIFO count read %d of %d (error %d))\n", j, nfs, r);
+			return -EIO;
 		}
 	}
-	return j - 1;
+	return j;
```
grandixximo, line 885: `j - 1` => `j` (and `i - 1` => `i` in `queue_read_data`) is another exported-symbol semantic change. `hm2_modbus` only tests `< 0`, and `hm2_pktuart.3.adoc` documents no return value at all, so nothing in tree notices. The old `j - 1` does look like a bug rather than a contract, but since this silently changes an API other comps may rely on, could it get its own commit plus a line in the man page?

src/hal/drivers/mesa-hostmot2/sserial.c
```diff
@@ -1439,17 +1434,21 @@ int hm2_sserial_register_tram(hostmot2_t *hm2, hm2_sserial_remote_t *chan){
                  case 1:
                     ret = hm2_sserial_wait(hm2, inst, period);
                     if (ret > 0) break;
-                    // FIXME: Assigns 100 to state3 and then unconditionally assigns 2
-                    if (ret < 0) hal_set_ui32(inst->state3, 100); // quit and tidy up
+                    if (ret < 0) {
+                        hal_set_ui32(inst->state3, 100); // quit and tidy up
```
grandixximo: This is the #4144 fix; I have raised it in the review body so @BsAtHome and @andypugh can weigh in on direction.

The concrete gap: `case 100` at line 1589 does not set `written = param`. With the `break` in place, a parameter whose write keeps failing stays dirty and is retried on every pass through the loop, printing `Problem found writing sserial parameter %s` and issuing a stop-all each time. Today it is written once, silently, and never repeated. Which failure mode would a user with an intermittent sserial cable rather have?

src/hal/drivers/mesa-hostmot2/sserial.c
```diff
@@ -672,7 +674,7 @@ int hm2_sserial_parse_md(hostmot2_t *hm2, int md_index){
     }
 
     // Now iterate through the sserial instances, seeing what is on the enabled pins.
-    for (i = 0 ; i < md->instances ; i++) {
+    for (i = 0 ; i < hm2->sserial.num_instances ; i++) {
```
grandixximo: Agreed there is a real bug here: with `num_sserials` set below `md->instances`, `inst = &hm2->sserial.instance[count]` can walk past the end of an array `hal_malloc`'d at `num_instances` entries.

Does capping the loop change which ports get probed, though? The old sweep visited every hardware instance and packed the populated ones into slots 0..count-1, and the comment further down still says the slot is reused "to keep active ports contiguous in the array". With `num_sserials=1` on a board where port 0 is empty and port 1 is populated, you now find nothing. Would `if (count >= hm2->sserial.num_instances) break;` inside the loop keep the bounds fix and the original packing?

src/hal/drivers/mesa-hostmot2/sserial.c
```diff
@@ -2145,13 +2155,17 @@ void hm2_sserial_force_write(hostmot2_t *hm2){
 void hm2_sserial_cleanup(hostmot2_t *hm2){
     int i,r;
     rtapi_u32 buff;
-    for (i = 1 ; i < hm2->sserial.num_instances; i++){
+    if (hm2->sserial.instance == NULL) return;
+
+    for (i = 0 ; i < hm2->sserial.num_instances; i++){
```
grandixximo, line 2155: Starting at 0 is right, and the `memset` plus the `command_reg_addr != 0` guard make the untouched slots safe.

One thing this now exposes more widely: `hm2_sserial_cleanup` runs from the parse_md failure path at line 846 and again from `hm2_cleanup`, and none of these pointers are NULLed after `rtapi_kfree`. Since you are already in this function, would you mind zeroing `confs`, `modes`, `globals`, `remotes` and their counts after freeing?

src/hal/drivers/mesa-hostmot2/sserial.c
```diff
@@ -1408,26 +1407,22 @@ int hm2_sserial_register_tram(hostmot2_t *hm2, hm2_sserial_remote_t *chan){
...
-                                    shift = 0;
-                                }
-                            // FIXME: This overlayed s64 read is very wrong!
-                            if (xlabs((hal_get_sint(p->param.s) - p->s64_written) >> shift) > 2) break;
-                            return hal_set_ui32(inst->state2, 2); // increment indices
+                                    if (hal_get_real(p->param.r)
+                                            != p->float_written) break;
+                                    return hal_set_ui32(inst->state2, 2);
+                                default:
+                                    HM2_ERR("Non IEEE float type parameter of length %i\n",
```
grandixximo, line 1427: `DataLength` 8 and 16 had defined shifts before and now land in the `default` arm, which is an `HM2_ERR` inside the RT parameter loop. Since the state machine only increments indices, a remote reporting a 16 bit float parameter would print on every pass through the list.

Do you know of a remote that reports one, or is it safe to say only 32 and 64 exist in practice? If the latter, worth saying so in the comment so the next reader does not have to rediscover it.

src/hal/drivers/mesa-hostmot2/abs_encoder.c
```diff
@@ -400,12 +411,6 @@ int hm2_absenc_parse_md(hostmot2_t *hm2, int md_index) {
 
     if (hm2->absenc.num_chans == 0) { // first time though
```
grandixximo, line 412: Correct that these were leaked: `hm2_register_tram_read_region()` stores `&hm2->absenc.ssi_busy_flags` and the TRAM allocator overwrites the pointer, so the `kmalloc`ed word never survived.

Removing them also removes a NULL guard for the dereference further down in `hm2_absenc_process_tram_read()`. That holds today only because `hm2_absenc_setup_ssi()` and its siblings always set the matching `*_global_start_addr` non-zero, and that is what gates the registration. Worth a comment at the dereference noting the pointer comes from TRAM allocation, so nobody reintroduces a path where a channel exists without its global register.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
```diff
@@ -1369,23 +1418,8 @@ static int hm2_eth_probe(hm2_eth_t *board) {
...
-        // terminal block, 10 pins used, Step & Dir 4,5, serial Rx/Tx/Txen 0,1
-        board->llio.ioport_connector_name[2] = "TB4";
-
-        // terminal block, 8 inputs, 6 SSR outputs
-        board->llio.ioport_connector_name[3] = "TB5";
-
-        // terminal block, 16 inputs
-        board->llio.ioport_connector_name[2] = "TB6";
+        // terminal block, TB1 to TB6
+        board->llio.ioport_connector_name[1] = "TB1-TB6";
```
grandixximo: The original was clearly wrong: `[2]` and `[3]` overwritten three times each, TB4 and TB6 never landing anywhere.

Both these boards set `num_ioport_connectors = 2` and supply `io_connector_pin_names`, though, so `ioport_connector_name` is only read by `hm2_print_pin_usage` on the branch these boards never take. Given that, is collapsing to `"TB1-TB6"` better than deleting the dead assignments and leaving `[0]` and `[1]` as they were? The new string names a connector that does not exist.

grandixximo (commented):
Two things I would want resolved before this goes in.

**The `if (r < 0)` => `if (!r)` conversion is not safe for every llio.** Details inline on `pktuart.c`; short version is that `hm2_spi.c` returns negative errnos, and a negative value is truthy.

**The sserial state machine hunks revisit #4144.** Those six `if (ret < 0) { ...; break; }` additions are the fix proposed there. #4144 was closed without the behavior change, on the concern that smart-serial might start quitting with spurious errors.

The half that made that fix safe is also missing here: `case 100` does not set `written = param`, so with the `break` in place a parameter whose write keeps failing stays dirty and is retried on every pass through the list, with an `HM2_ERR` each time. Today the code marches on after a failed wait, and the write is silently lost but never repeated.

@BsAtHome @andypugh you both have the context on #4144. Should these hunks land here with the `case 100` half added, or go back to that issue?

src/hal/drivers/mesa-hostmot2/pktuart.c
BsAtHome: According to the llio interface, the read/write functions return 1 on success and 0 on failure. The hm2_spi driver is wrong. The hm2_rpspi driver is wrong too (only partially). The hm2_spix (and subordinate) driver does it correctly.

BsAtHome:
> @BsAtHome @andypugh you both have the context on #4144. Should these hunks land here with the `case 100` half added, or go back to that issue?

I think it is necessary to split this PR and address the issue classes separately.

The fixes on the tests of llio write/read failure are important. But it must be consistent. The llio drivers must do the right thing too.
And then, a huge swat of the read/write calls never get checked at all. Implementing guards everywhere would not necessarily be a good thing because it is sometimes unclear what you can do to resolve the issue and most would bug the user at thread-cycle rates. So changing it all is not really an option.

I suggest to take out all the llio drivers' write/read issues (where the return value is checked, like pktuart) and fix the llio read/write in the hm2_spi/hm2_rpspi drivers as well. However, it must be said that both are superseded by the hm2_spix driver. There may be other llio drivers that need to be checked for the correct return values.

src/hal/drivers/mesa-hostmot2/spix_rpi3.c
```diff
@@ -850,8 +850,8 @@ static const spix_port_t *rpi3_open(int port, const spix_args_t *args)
 	if(!rpp->spiport) {
 		rpp->clkdivw = spi0_clkdiv_calc(spiclk_base, args->clkw);
 		rpp->clkdivr = spi0_clkdiv_calc(spiclk_base, args->clkr);
-		ccw = spiclk_base / rpp->clkdivw;
-		ccr = spiclk_base / rpp->clkdivr;
+		ccw = spiclk_base / (rpp->clkdivw ? rpp->clkdivw : 65536);
+		ccr = spiclk_base / (rpp->clkdivr ? rpp->clkdivr : 65536);
```
BsAtHome, line 854: Correct. Would be a division by zero in showing the speed. It is is the slowest possible speed. Not that anyone would ever want to select that...

BsAtHome:
Fixed the llio SPI drivers in #4489.

src/hal/drivers/mesa-hostmot2/pktuart.c
sleepybishop: Thank you, I moved them to their own commit and updated the man pages.

src/hal/drivers/mesa-hostmot2/sserial.c
sleepybishop: Agree, I removed this change.

src/hal/drivers/mesa-hostmot2/sserial.c
sleepybishop: Updated, it still probes every physical port and stops only after the configured number of populated ports has been packed.

src/hal/drivers/mesa-hostmot2/sserial.c
sleepybishop: Agree, fixed.

src/hal/drivers/mesa-hostmot2/hm2_eth.c
sleepybishop: Agree, removed the dead assignments and retained the two logical connector entries, P1 and TB1

src/hal/drivers/mesa-hostmot2/abs_encoder.c
sleepybishop: I added a comment explaining that these pointers are supplied by TRAM registration

src/hal/drivers/mesa-hostmot2/sserial.c
sleepybishop: AFAICT, the current read/write paths support only 32 and 64-bit floats. I documented that and made unsupported widths warn once only.

sleepybishop:
Thanks. I removed the LLIO convention changes from this PR and defer to #4489 to handle the legacy SPI backends. I tried to reorganize the remaining fixes into focused commits, but agree splitting is also fair. Let me know what would make the most sense.

BsAtHome:
Will need to do a deeper dive to see how a split would be best.

I'm especially wary about changes in sserial and the encoder stuff. That has also been a pain in porting to the getter/setter infrastructure. Any changes in there _must_ be checked against running hardware. So these qualify for a very limited separate PR.

grandixximo:
Before you dive into the split, is it worth getting PCW one pass over the branch first? Would one branch be less work to test than two or three separate PRs?

Results recorded against commit hashes would survive whatever split you land on, as long as the branch is not rebased before then.

src/hal/drivers/mesa-hostmot2/bspi.c
```diff
@@ -121,6 +121,10 @@ int hm2_tram_add_bspi_frame(char *name, int chan, rtapi_u32 **wbuff, rtapi_u32 *
         HM2_ERR_NO_LL("Can not find BSPI instance %s.\n", name);
         return -1;
     }
+    if (chan < 0 || chan > 15) {
+        HM2_ERR("BSPI %s: channel number %i is out of range\n", name, chan);
+        return -EINVAL;
+    }
```
BsAtHome, line 127: All other error exits return `-1`.

Same problem for changes in `hm2_bspi_write_chan()` and `hm2_bspi_setup_chan()`

BsAtHome:
For example, the pktuart stuff I can test locally (hm2_modbus uses it). That also includes the hm2_eth (for at least 7i96s). But the eth changes should be manageable. Adding in the uart fixes should be alright in this batch.

That what really worries me is sserial.c and abs_encoder.c. I will need to do a deep dive what is being done. It has surprised me before how the old hacks work. Changing things can have consequences.

The parts inm, inmux, pins, resolver and bspi can be joined in a PR for separate review.

Both spix_rpi3 and tp_pwmgen can be applied without further notice. These are clear and simple bugfixes.

KentChenY:
test

src/hal/drivers/mesa-hostmot2/setsserial.c
```diff
@@ -176,7 +176,7 @@ int sslbp_read_cookie(void){
     return res;
 }
 
-rtapi_u8 sslbp_read_byte(rtapi_u32 addr){
```
andypugh, line 179: This is called read_byte and the return value is an unsigned byte. What is the reasoning behind changing to signed int?

sleepybishop: The intention was to preserve the pre-existing `return -1`'s error path. With an unsigned return type, -1 is converted to 255. Returning int preserves successful byte values from 0–255 while allowing the caller to detect an error before using the value as a shift exponent.

src/hal/drivers/mesa-hostmot2/bspi.c
sleepybishop: Fair, I've updated them match existing convention.

sleepybishop:
I'll leave this alone for the time being, please let me know how to proceed wrt to the split.

PR #4489, opened 2026-08-30
hostmot2: Fix the (superseded) SPI driver's low-level IO return values

BsAtHome:
The old and superseded SPI drivers for `hm2_rpspi` and `hm2_spi` did not return the correct low-level IO read/write return value. Even for these superseded drivers, as long as they are there, they should do the correct thing.

Note: for anyone using these old superseded drivers: you should be using `hm2_spix` instead.

### DIFF
--- a/src/hal/drivers/mesa-hostmot2/hm2_spi.c
+++ b/src/hal/drivers/mesa-hostmot2/hm2_spi.c
@@ -144,12 +144,12 @@ static int send_queued_writes(hm2_lowlevel_io_t *llio) {
 static int queue_write(hm2_lowlevel_io_t *llio, rtapi_u32 addr, const void *buffer, int size) {
     hm2_spi_t *this = (hm2_spi_t*) llio;
     if(size == 0) return 0;
-    if(size % 4) return -EINVAL;
+    if(size % 4) return 0; // -EINVAL
 
     int wsize = size / 4;
     if(wsize + this->nbuf + 1 > MAX_TRX) {
         int r = do_pending(this);
-        if(r < 0) return r;
+        if(r < 0) return 0;
     }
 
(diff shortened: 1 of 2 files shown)


src/hal/drivers/mesa-hostmot2/hm2_spi.c
```diff
@@ -190,15 +190,15 @@ static int queue_read(hm2_lowlevel_io_t *llio, rtapi_u32 addr, void *buffer, int
 static int do_write(hm2_lowlevel_io_t *llio, rtapi_u32 addr, const void *buffer, int size) {
     hm2_spi_t *this = (hm2_spi_t*) llio;
     int r = queue_write(llio, addr, buffer, size);
-    if(r < 0) return r;
+    if(!r) return 0;
```
grandixximo, line 193: `size == 0` returns 0 here, where `hm2_eth` and the three spix backends return 1. For `hm2_spi` that is also a flip: `do_write(size=0)` used to reach `do_pending() >= 0` and come back 1.

Unreachable today, every call site passes a nonzero `sizeof`. But if this commit is where the convention gets pinned down, should zero-size be success or failure, and should `hostmot2-lowlevel.h:70` say which?

BsAtHome: That is indeed a good question. I'm not sure we should bother. Both the hm2_spi and hm2_rpspi are obsolete drivers. Both drivers have survived with the current "wrong" assumptions.

For the hm2_spi, both `queue_write()` and `queue_read()` always returned zero on `size==0`. They returned `-EINVAL` on `size%4 != 0`, which now returns zero. The `do_write()` is the llio interface, where zero is error, so that should work out properly.

PR #4490, opened 2026-08-30
tests: Update all components in tests to getter/setter

BsAtHome:
There are components used in the tests. These also need to be converted to getter/setter or the tests will (soon) fail.

This PR addresses that issue. Additionally, several components generated compiler warnings. All of the unused arg/variable/function warnings. These have been quelled by mostly obvious means.

### DIFF
--- a/tests/halcompile/names/names_match.comp
+++ b/tests/halcompile/names/names_match.comp
@@ -1,6 +1,7 @@
 component names_match;
 license "GPL";
-pin out bit out;
+pin out bool out;
+option period no;
 function _;
 ;;
-FUNCTION(_) {}
+FUNCTION(_) { out_set(0); }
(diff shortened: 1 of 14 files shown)


tests/halcompile/names/names_match.comp
```diff
@@ -1,6 +1,6 @@
 component names_match;
 license "GPL";
-pin out bit out;
+pin out bool out;
```
grandixximo, line 3: This one still compiles with `-Wunused-parameter` on `period` (and `__comp_inst`). The other converted comps got `option period no`; left out here on purpose? names_dont_match is a negative test so the noise matters less, but here the build output stays noisy.

BsAtHome: In this case you don't see the noise, normally. But, for consistency, fixed.

PR #4492, opened 2026-09-01
optimisation: rtapi: copy strings without going through vsnprintf

alex-pres:
_Another bottleneck found, and probably it is time to stop here with optimisation._ 

`rtapi_strlcpy` ran a full `vsnprintf` per call, and `rtapi_strxcpy` is on the interpreter hot path (`write_state_tag`, once per motion block): 

Benchmarked this fix: _parse-only over a 13.7M-line file drops 11.19s to 9.21s (1.21x)._  

`rtapi_strlcat` is not on hotpath but also fixed to avoid `vnsprintf`. Behaviour changed to return BSD return value, and avoid overrun when size-l underflows. No callers in linuxcnc code used return value of `rtapi_strlcat` so the fix is safe.

### DIFF
--- a/src/rtapi/rtapi_string.h
+++ b/src/rtapi/rtapi_string.h
@@ -48,7 +48,13 @@ RTAPI_END_DECLS
 #endif
 RTAPI_BEGIN_DECLS
 static inline size_t rtapi_strlcpy(char *dst, const char *src, size_t size) {
-    return rtapi_snprintf(dst, size, "%s", src);
+    size_t len = strlen(src);
+    if (size) {
+        size_t n = (len < size - 1) ? len : size - 1;
+        memcpy(dst, src, n);
+        dst[n] = '\0';
+    }
+    return len;
 }
 #define rtapi_strxcpy(dst, src) ({ \
     rtapi_static_assert(rtapi_is_array(dst), "dst must be non-const array"); \
@@ -56,8 +62,10 @@ static inline size_t rtapi_strlcpy(char *dst, const char *src, size_t size) {
 })
 
 static inline size_t rtapi_strlcat(char *dst, const char *src, size_t size) {
-    size_t l = strlen(dst);
-    return rtapi_snprintf(dst+l, size-l, "%s", src);
+    size_t l = strnlen(dst, size);
+    if (l == size)
+        return l + strlen(src);
+    return l + rtapi_strlcpy(dst+l, src, size-l);
 }
 
 #define rtapi_strxcat(dst, src) ({ \


BsAtHome:
I'm not sure this fix belongs in RTAPI. The interpreter is completely non-RT, user-space only and should not be using RTAPI calls for string manipulation, at all.

alex-pres:
```
#0  ___vsnprintf_chk (... format=0xffff90501f20 "%s" ...)
#2  rtapi_snprintf (...) at ./rtapi/uspace_common.h:272
#3  rtapi_strlcpy (dst=..., src=..., size=256) at ../include/rtapi_string.h:51
#4  Interp::write_state_tag (...) at emc/rs274ngc/interp_write.cc:229
#5  Interp::write_canon_state_tag (...) at emc/rs274ngc/interp_write.cc:364
#6  Interp::convert_straight (...) at emc/rs274ngc/interp_convert.cc:5484
```

the same Interp code working for preview parsing and for realtime which I believe is correct. 

I believe string copy must never been implemented like that, and specially in realtime.

BsAtHome:
Well, in realtime it is actually never in the RT path. It is only invoked at rtapi_app_main() initialization or when the shit hits the fan and an error pops up (which kills realtime anyway). The reason for its existence is (old) kernel compatibility that has been missing some string functionality. Apparently no real problem in newer kernels, except for the kprintf lacking floating point, which the RTAPI clone does implement.

And, it confirms what I said... It should not be using RTAPI, but it does. So, the fix is to stop the interpreter from using it? FWIW, when you look deeper into the interpreter code, then you also discover that it uses quite a few static buffers and a lot of C code with a small coating of C++.

Cleaning up RTAPI is a different project. Necessary, probably, but change in there has larger ramifications and potential for trouble (you cannot use any libc functionality or many of the builtins).

alex-pres:
> Cleaning up RTAPI is a different project. Necessary, probably, but change in there has larger ramifications and potential for trouble (you cannot use any libc functionality or many of the builtins).

It'd be really challenging to touch it. 

This fix however must be safe enough if not 100.0% safe.

alex-pres:
> And, it confirms what I said... It should not be using RTAPI, but it does. So, the fix is to stop the interpreter from using it? FWIW, when you look deeper into the interpreter code, then you also discover that it uses quite a few static buffers and a lot of C code with a small coating of C++.

> Can you please test with this branch: https://github.com/BsAtHome/linuxcnc/tree/refs/heads/fix_rs274ngc-no-rtapi
> 
> It removes RTAPI completely from the interpreter and uses plain strncpy() via similar define/inline function. It does not use any printf.

I checked and yes, it gives the same result on benchmarks and that is definitely right solution for performance problem. 


However, keep snprintf("%s", ...) in realtime lib for string copy, is it good? I believe it is better to get fixed too.

BsAtHome:
> I checked and yes, it gives the same result on benchmarks and that is definitely right solution for performance problem.

Good, that is in #4493 now. This also removes the RTAPI dependency in the interpreter's librs274.so, if I'm not mistaking. And that is a good thing on the path of untangling the code.

> However, keep snprintf("%s", ...) in realtime lib for string copy, is it good? I believe it is better to get fixed too.

I agree that the printf in that path should be removed too. However, it should be part of a larger RTAPI cleanup, I think.

BsAtHome:
Superseded by #4493.

Issue #3878, opened 2026-03-25
Discussion: Real Time Cores (Preempt-RT, RTAI, Xenomai and others)


hdiethelm:
After the long discussions in https://github.com/LinuxCNC/linuxcnc/issues/3873 and https://github.com/LinuxCNC/linuxcnc/issues/3875, I investigated the real time options a bit more in depth and created a summary.

LinuxCNC has quite a number of available real time cores. Some of them are broken, others work fine. What I found so far:

All userspace variants are configured by `--with-realtime=uspace` and built if the needed library is installed. During runtime, a variant is detected using the kernel name, version and other hints and selected automatically or also not if your kernel name doesn't match expectations.

Format: {Text shown at start of LinuxCNC}:{My short description}

- User Space: `--with-realtime=uspace`
  - POSIX non-realtime: Any old kernel or you forgot `make setuid`
    - Works, but big latency
    - Basically what starts when you don't have any other realtime or it is not properly detected
  - POSIX realtime:  PREEMPT_RT
    - Works
    - Probably the most used option
    - Debian stock RT kernel works fine
  - XENOMAI (posix-skin) realtime: Xenomai2: Might be / Xenomai3: Works / Xenomai4: Not compatible
    - Works: I managed to get it running with Xenomai3
    - Probably seldom used due to no up to date compiled kernel package is available
  - LXRT realtime: RTAI Userspace
    - Seams to be broken after 2.8.4. I had no success, 2.8.4 is buster where I don't have a VM. In Bookwork/Trixie I was not able to compile 2.8.4. 2.9.* and master builds but has an error starting.
- Kernel Space: `--with-realtime=/usr/realtime-$VERSION`
  - Nothing: RTAI Kernelspace
    - Works in 2.9
    - Broken in Master: https://github.com/LinuxCNC/linuxcnc/issues/3875
    - @NTULINUX maintains a kernel which is available in the LinuxCNC Debian repo: https://github.com/NTULINUX/RTAI

RTAI is not maintained any more, last commit ~4 years ago.

To help others wanting to try Xenomai3, I created a repo with all you need to get it running:
https://github.com/hdiethelm/xenomai3-linuxcnc

Xenomai4 has a new API: EVL. I did not found a posix skin. I managed to get the kernel and tools built for anybody interested in creating a Xenomai4 module for LinuxCNC:
https://github.com/hdiethelm/xenomai4-linuxcnc

[...]

PR #4498, opened 2026-09-02
hm2_eth: Create support library's

hdiethelm:
This PR has tree parts. If desired, I can split them also up in to separate PR's.

For now it is more about the concepts. No need to review the code from a style standpoint.

**Part1: Create support library's**
The issue is, that as soon as hm2_eth is linked to libevl, this module can not be used any more without libevl installed.
By creating liblinuxcnc-hm2_eth_net_evl.so and liblinuxcnc-hm2_eth_net_xenomai.so and loading them only when needed, the hard dependency is broken and the library's can be deployed in a linuxcnc-uspace-xenomai / linuxcnc-uspace-evl Debian package.

The Debian packages are not yet created but I am planning to do so if this approach is approved.

The firewall functions are moved out of the Ethernet support modules, so not so many exports are needed for the library's. The function order changes slightly but this should have no side effects.

Alternatives:
- hm2_eth / hm2_eth_xenomai / hm2_eth_evl
  - Three hm2_eth components, main core but different Ethernet code linked. 
  - Disadvantages
    - You can not run one Ethernet card with EVL and the other with posix. Niche application if you have a low priority thread communicating with a second mesa card.
    - Three components that could have been a config flag.
- hm2_eth_xenomai / hm2_eth_evl linuxcnc components
  - Disadvantages: 
    - More components
    - Empty rtapi_app_main 
    - Needs entry in hal file
    - Makefile makes it hard to define custom C / LD flags for specific modules. I did not figure out how to do this.
- Just leave it as it is and depend on libevl
   - Disadvantage: Users would have to compile it from source or three different full linuxcnc-uspace packages are needed.

What do you think about the support library approach?

**Part 2: Xenomai3 support**
~It was easy to do. Note that sendto(), recvfrom() and other syscalls are overwritten by the Xenomai C / LD flags, so even thoutght it looks like posix, behind are xenomai syscalls. This is just how xenomai3 posix skin works.~

~TBD: if anyone needs this. I just wanted to try it to see any differences to xenomai4. It needs more testing but on my hardware, performance is similar. However, setup is a bit more cumbersome. The two scripts rtnet_up.sh / rtnet_down.sh are just temporary and should go to the man page with some description how to set it up when done.~
This will go to a separate PR. Code is here: https://github.com/hdiethelm/linuxcnc-fork/tree/hm2_eth_lib_xenomai3

**Part 3: Xenomai4 OOB improvements**
Moved to https://github.com/LinuxCNC/linuxcnc/pull/4503 which should be merged first.


BsAtHome:
To me it certainly feels like a bad choice doing dlopen in a realtime component.

Is the target for compilation ever both uspace/posix, Xenomai3 and EVL _at the same time_? If the answer is no, then there should be only one `hm2_eth` component/driver, which adapts properly at compile time to the necessary layout(s) and format(s). That can still be supported by libraries, but whether they should be .so is doubtful. The compilation generates a (mostly) static driver. Only upwards are symbols matched (to the hostmot2 component). Requiring this also downwards is unnecessary when you know at build-time what your target system is.

So, what is the answer to the question, whether the build can only target one system at a time or multiple simultaneously?

hdiethelm:
> To me it certainly feels like a bad choice doing dlopen in a realtime component.
> 

Why? It only changes from 2x to 3x dlopen in the rtapi_app main thread. But of course there is no dlopen in the realtime tasks which would be bad.
First two:

Load RT extension:
https://github.com/LinuxCNC/linuxcnc/blob/fa910f12394ba1b76edc4c5b3ac67955d9d79ace/src/rtapi/uspace_rtapi_main.cc#L1533

Load Component:
https://github.com/LinuxCNC/linuxcnc/blob/fa910f12394ba1b76edc4c5b3ac67955d9d79ace/src/rtapi/uspace_rtapi_main.cc#L576

With this PR, there is a third dlopen in the same thread in the rtapi_app_main path called just after load:
https://github.com/LinuxCNC/linuxcnc/blob/fa910f12394ba1b76edc4c5b3ac67955d9d79ace/src/rtapi/uspace_rtapi_main.cc#L583

> Is the target for compilation ever both uspace/posix, Xenomai3 and EVL _at the same time_? If the answer is no, then there should be only one `hm2_eth` component/driver, which adapts properly at compile time to the necessary layout(s) and format(s). That can still be supported by libraries, but whether they should be .so is doubtful. The compilation generates a (mostly) static driver. Only upwards are symbols matched (to the hostmot2 component). Requiring this also downwards is unnecessary when you know at build-time what your target system is.
> 

At build time, if you have libxenomai / libevl installed, all three (+posix) variants are compiled in. But that also means that if you create a debian package, it will automatically depend on these libs.

I use this all day for testing, just boot a different kernel and it works out of the box with the correct back-end, no rebuild needed.

Looks like the reason liblinuxcnc-uspace-xenomai.so / liblinuxcnc-uspace-xenomai-evl.so for rtapi_app also exist to break this dependency, and there is even a control file to pack these in a different debian package: https://github.com/LinuxCNC/linuxcnc/blob/master/debian/control.uspace-xenomai.in But it looks like the rest was either removed or never finalized, so these packages are not built. If this is finalized, you would have:

linuxcnc-uspace: Posix, not depending on libxenomai / libevl
linuxcnc-uspace-xenomai: Xenomai support, depending on libxenomai
linuxcnc-uspace-evl: Xenomai EVL support, depending on libevl

> So, what is the answer to the question, whether the build can only target one system at a time or multiple simultaneously?

It can target all of the existing uspace realtime systems at the same time but right now, linuxcnc-uspace will then depend on libxenomai / libevl which is unnecessary but only an issue if these libs are not available in the linuxcnc package repo.

grandixximo:
This is best argument I heard so far, for splitting the deb...

hdiethelm:
> This is best argument I heard so far, for splitting the deb...

This is also the main intent besides being able to use linuxcnc built with libevl on a system withouth. 

However, with a single deb, it will still depend on libevl.

Might be it can be changed to recommended, i could look into how debian packages are configured and if shlibdeps can be modifyed. But I think separate packages are nicer anyway.

zz912:
Hello,

I have two questions from a user’s perspective:

1. What do you see as the main user-facing reason for supporting Xenomai 3 in `hm2_eth`? Is there a specific use case for it in LinuxCNC, or is it more about keeping the door open for future use?

2. If Xenomai 4/EVL is the direction LinuxCNC’s realtime support is heading, wouldn’t it be more beneficial for users to focus the effort on a ready-to-use Xenomai 4 + LinuxCNC image/Live CD rather than supporting multiple realtime backends?

hdiethelm:
> Hello,
> 
> I have two questions from a user’s perspective:
> 
>   1. What do you see as the main user-facing reason for supporting Xenomai 3 in `hm2_eth`? Is there a specific use case for it in LinuxCNC, or is it more about keeping the door open for future use?

Basically, Xenomai3 support is already in since some time (2016). Now that the framework is in to have different backends for hm2_eth, it took only 140 lines of code for Xenomai3 support.

It might be of use for some people, so why not add it, especially due to it took only a few hours to create it.

All I have read so far about realtime performance, there is no way to predict which system works best on your PC. The only real way is to test it. And you can only do that if the support is there. It might well be that for some people, Xenomai3 works better. This was also the main reason for me to implement this: Being able to test how well it works.

> 
>   2. If Xenomai 4/EVL is the direction LinuxCNC’s realtime support is heading, wouldn’t it be more beneficial for users to focus the effort on a ready-to-use Xenomai 4 + LinuxCNC image/Live CD rather than supporting multiple realtime backends?

This is a question for the maintainers creating live images. Basically, the actual framework would also support a live image with tree different kernels and depending on which you start, the matching RT framework is used.

However, to have generic Debian packages which can be installed with or without libevl / libxenomai, part 1 of this PR (or one of the alternatives) is needed.

zz912:
Thanks for the explanation. I now understand why you want to keep the individual variants separate – mainly because of the dependencies and because, as a developer/tester, you can have a single build and test it against different kernels.

I’m looking at it a little differently, though, from the perspective of a beginner who is using LinuxCNC for the first time.

Someone coming from Windows or Mac is often also working with Debian for the first time. Such a user really doesn’t know what POSIX, Xenomai 3 or EVL mean. I don’t think it’s ideal to expect them to make this decision before they have even run LinuxCNC for the first time.

A Debian beginner also tends to have exactly the opposite approach from an experienced developer: they would rather have “everything” installed so that it works. For a developer this may be a nightmare, but for a beginner it is completely normal. :-)

So I could imagine something like a simple meta-package called `linuxcnc` (just a suggestion for the name), which would install all the required variants:

```
linuxcnc
   ├── linuxcnc-uspace
   ├── linuxcnc-uspace-xenomai3
   └── linuxcnc-uspace-xenomai4
```

An experienced user could of course still install only `linuxcnc-uspace` or only `linuxcnc-uspace-xenomai4` if they know exactly what they need. But for a beginner there would be a simple path: install `linuxcnc` and they don’t need to understand the differences between the various realtime environments beforehand.

This is just a suggestion from a UX perspective. I’m not saying that it is technically a better solution than your package separation.

One more small point about the naming. Personally, I would prefer:

```
linuxcnc-uspace
linuxcnc-uspace-xenomai3
linuxcnc-uspace-xenomai4
```

instead of `linuxcnc-uspace-xenomai` and `linuxcnc-uspace-evl`.

Again, the main reason is the beginner. “EVL” doesn’t mean anything to them, whereas `xenomai3` and `xenomai4` immediately make it clear that these are two generations of Xenomai. I know that Xenomai 4 uses EVL and that technically it is a bit more complicated, but from a user’s perspective `xenomai4` seems like a much more understandable name to me.

So perhaps it makes sense to keep the packages separate for clean dependencies, while also providing one simple “default” path for people who are just getting started with LinuxCNC and Debian.

hdiethelm:
Basically, not keeping the add-ons separate would mean either:
- linuxcnc-uspace depends on libevl / libxenomai which are not available for install (for now) except from my github
- Build from source or deploy separately as it is the case right now

If you don't know what this means, either google it or read the doc. I still have to update the main section about real time tough. For beginners, it's probably the best to stay with PREEMPT_RT, at least until there is some experience from more advanced users.

linuxcnc -> this is already taken by the RTAI variant

Suggestion which (mostly) matches the naming already used in the source:
linuxcnc-uspace suggests: linuxcnc-uspace-xenomai, linuxcnc-uspace-xenomai-evl in the case these packages are built
linuxcnc-uspace-xenomai
linuxcnc-uspace-xenomai-evl

hdiethelm:
@BsAtHome Would you prefer if I separate the Xenomai4 OOB improvements to a separate PR?

The library part seams to need some more discussion. But it is also needed for Xenomai3 due to Xenomai3 uses a many C and LD flags to replace posix functionality which would probably be a bad idea to apply to all components.

hdiethelm:
To simplify review and testing, I decided to create a PR only concerning the Xenomai4 improvements: https://github.com/LinuxCNC/linuxcnc/pull/4503 It should be merged first due to there will be some conflicts but they are easy to resolve.

BsAtHome:
Separating the PR's library question from other things would be a good thing because it separates features. I do have some worries about the proper separation of the CFLAGS/LDFLAGS necessities.

Having this in-tree means it needs to be maintained. And, you already noticed, some parts have been bitrotting. Fixing that is a good thing. Adding to it makes things more complex. That should be no problem if there is a way to maintain it properly.

Question 1: Can the combined packages for uspace/xenomai/evl be built in CI? You don't need a running kernel for the build process afaik.
Question 2: How is this going to be tested in real life? Is there a live/install image coming with three kernels so? And, how is that being maintained?

hdiethelm:
> Separating the PR's library question from other things would be a good thing because it separates features. I do have some worries about the proper separation of the CFLAGS/LDFLAGS necessities.

Done, I removed all code not needed for the library creation. However, the commit moving the firewall functionality to hm2_eth (also included in https://github.com/LinuxCNC/linuxcnc/pull/4503) is needed. The library change is: https://github.com/LinuxCNC/linuxcnc/pull/4498/commits/9978feee354123124b0716ef4fd6e8949f86a50f. The original branch is at https://github.com/hdiethelm/linuxcnc-fork/tree/hm2_eth_lib_xenomai3 until this has settled.

Yes, I also don't like the way it is now with CFLAGS/LDFLAGS for components. There is no way to change them for a single component as much as it looks. All files are added to a huge list and then compiled with the same settings, except the path based filter for cruckig. So it's either all, nothing or a complex filter.

If there is a way to build certain components with different flags, an other viable option would be to build hm2_eth + back-end in 3 variants: hm2_eth / hm2_eth_evl / hm2_eth_xenomai and then instead of using a config flag, select the back-end by loading the component. This was indeed my first approach and it would not need `dlopen()`. I discarded it because it is less flexible, you can not mix posix and evl cards.

For evl, CFLAGS/LDFLAGS is not a big issue, the only flag is -levl but the linker removes it if there is nothing to link. For Xenomai3, there are many flags which I really don't want to apply to all components.

> 
> Having this in-tree means it needs to be maintained. And, you already noticed, some parts have been bitrotting. Fixing that is a good thing. Adding to it makes things more complex. That should be no problem if there is a way to maintain it properly.
> 
> Question 1: Can the combined packages for uspace/xenomai/evl be built in CI? You don't need a running kernel for the build process afaik. 

Yes, I supply the packages needed here:
https://github.com/hdiethelm/xenomai3-linuxcnc
https://github.com/hdiethelm/xenomai4-linuxcnc

I can install them in CI similar than we do for the RTAI build. To build the linuxcnc:
Xenomai4 needs libevl / libevl-dev / linux-libc-evl-dev
Xenomai3 needs libxenomai / libxenomai-dev

However, with the package structure right now, linuxcnc-uspace will then depend on libevl + libxenomai due to shlibdeps detecting the dependency. This is somewhat hidden, as soon as the libraries are installed, configure detects them and thus builds the additional files which then add dependency's to the debian packages. Creating separate packages will fix this.

BTW: This is nothing introduced by my changes. This issue exists probably since the separate xenomai packages where removed: https://github.com/LinuxCNC/linuxcnc/commit/6f285604ac1a

> Question 2: How is this going to be tested in real life? Is there a live/install image coming with three kernels so? And, how is that being maintained?

Right now, I maintain the two repos above. A live image would surely be possible but I did not look yet into how this is done. Otherwise, you can install all which is needed from my repo or build it from source with the supplied scripts and configs.

For Xenomai4, I run my machine with this Kernel + hm2_eth, so this will stay up to date at least for the near future and is tested.

For Xenomai3, I just use this for testing. There it depends on the interest. If there is no feedback or tickets incoming, it might start to bitrot. The main intent there was to compare it to Xenomai4 but if there is interest, I will update it from time to time.

BsAtHome:
With respect to CI, I don't think we should build all the .deb variants in one runner. Most people won't see that case. I was more thinking about the way the RTAI runner works. That leaves the normal builds without added dependencies. A dedicated CI runner will build the xenomai/evl debs (and only targeted at latest Debian stable). We need to get that right so no one gets confused which deb is which. Importantly the linuxcnc-uspace deb from either runner targeting the same Debian release should be bit-equal (i.e. reproducible build).
That surely means, we need to do the package split right now so we never ever get into the problem in the first place. Confusion is the last thing we want to create.

It may be a necessity to support Xenomai3 as long as Xenomai4 lacks some features (or not?). That is, unless there is a reason only to support the head, which may be less problematic for us in the long run. But that only works if we really don't need Xenomai3 for supporting older/other installs. It also depends on the kernel support for each Xenomai version and the prospects for future support. Once we formalize a way, we have a hard time getting rid of it again, so we need to be very sure to go one or another path.

hdiethelm:
> With respect to CI, I don't think we should build all the .deb variants in one runner. Most people won't see that case. I was more thinking about the way the RTAI runner works. That leaves the normal builds without added dependencies. A dedicated CI runner will build the xenomai/evl debs (and only targeted at latest Debian stable). We need to get that right so no one gets confused which deb is which. Importantly the linuxcnc-uspace deb from either runner targeting the same Debian release should be bit-equal (i.e. reproducible build). That surely means, we need to do the package split right now so we never ever get into the problem in the first place. Confusion is the last thing we want to create.

So are you OK using `dlopen()` to load the different support library's? Or do you think it's better to have 3 components 
hm2_eth / hm2_eth_evl / hm2_eth_xenomai? Both variants will be viable except the Makefile issue with components I am having right now.

The builds with Xenomai will not be bit identical than the builds without due to there are many ifdefs in rtapi_app and also hm2_eth.

I added a rip-xenomai stage which builds Xenomai3/4 to check for compile issues. I don't run the tests due to they don't test anything that is different in the xenomai build and there is no xenomai kernel running. If desired, I can add them.

As soon as a proper Debian package structure is in place, where linuxcnc-uspace will not depend on the lib's when they are active, one or also both variants can be built in the packaging stage by default without hurting anyone.

> 
> It may be a necessity to support Xenomai3 as long as Xenomai4 lacks some features (or not?). That is, unless there is a reason only to support the head, which may be less problematic for us in the long run. But that only works if we really don't need Xenomai3 for supporting older/other installs. It also depends on the kernel support for each Xenomai version and the prospects for future support. Once we formalize a way, we have a hard time getting rid of it again, so we need to be very sure to go one or another path.

I am not aware if someone even uses Xenomai3. It could be useful together with ether-cat: https://github.com/LinuxCNC/linuxcnc/issues/3878#issuecomment-5430604872. Most non-posix uspace realtime implementations where basically bitrotten and broken before I fixed them, except Xenomai3 on master branch worked. But without Ethernet support and no working kernel packages, it was most probably to much effort for many people.

Do you know if someone actively uses Xenomai3? 

Xenomai3 is stable and maintained by Siemens, while Xenomai4 is mostly maintained by Philippe Gerum and still somewhat work in progress. There where a few incompatible API changes in the ~6 months I am working on it.

Due to this and due to the additional code for full Xenomai3 support is not that much, it is probably worth keeping both variants at least for some time and also add Xenomai3 hm2_eth support.

Meanwhile, LXRT (RTAI Userspace) support can probably be removed. It was fully broken until I fixed it and it needs a change in the support headers supplied by @NTULINUX to be able to build this variant. At least on my PC, it does not work well and results in full system freezes from time to time.

BsAtHome:
> So are you OK using `dlopen()` to load the different support library's? Or do you think it's better to have 3 components hm2_eth / hm2_eth_evl / hm2_eth_xenomai? Both variants will be viable except the Makefile issue with components I am having right now.

I still don't like it. What are the alternatives, besides having three differently named components/drivers?

> The builds with Xenomai will not be bit identical than the builds without due to there are many ifdefs in rtapi_app and also hm2_eth.

This is a problem. The packages have the same name and should therefore be identical when built. Having the same package name with different functionality is a disaster waiting to happen.

When there are differences, due to the conditional compilation, then the packages must carry different names and add a conflict so that you cannot install them together. The alternative is to split the package into more parts, but that may be even more problematic due to rtapi_app. Having yet another level of exec indirection is just making complex things even more complicated and I cannot recommend that (and may be prone to errors too).

> I added a rip-xenomai stage which builds Xenomai3/4 to check for compile issues. I don't run the tests due to they don't test anything that is different in the xenomai build and there is no xenomai kernel running. If desired, I can add them.

At least the build should be tested if we are to support these in the long run, which we seem to do.

> As soon as a proper Debian package structure is in place, where linuxcnc-uspace will not depend on the lib's when they are active, one or also both variants can be built in the packaging stage by default without hurting anyone.

That will need some good thinking.

> Do you know if someone actively uses Xenomai3?

No.

> Xenomai3 is stable and maintained by Siemens, while Xenomai4 is mostly maintained by Philippe Gerum and still somewhat work in progress. There where a few incompatible API changes in the ~6 months I am working on it.

Well, it seems there is some backing. Then it will not die just tomorrow ;-)

> Due to this and due to the additional code for full Xenomai3 support is not that much, it is probably worth keeping both variants at least for some time and also add Xenomai3 hm2_eth support.

Xenomai 3 would probably the go-to version as long as Xenomai 4 is in flux. At least, for those seeking (some kind of) stability.

> Meanwhile, LXRT (RTAI Userspace) support can probably be removed. It was fully broken until I fixed it and it needs a change in the support headers supplied by @NTULINUX to be able to build this variant. At least on my PC, it does not work well and results in full system freezes from time to time.

There is probably no one who uses it. I wouldn't mind it gone. Maybe others want to pitch in on removing LXRT support? Otherwise it is left to bitrot until RTAI gets the boot.

grandixximo:
> LXRT (RTAI Userspace) support can probably be removed

+1

BsAtHome:
How many executables/libraries are affected by Xenomai{3,4} that alter the link behaviour/dependencies?

hdiethelm:
> > So are you OK using `dlopen()` to load the different support library's? Or do you think it's better to have 3 components hm2_eth / hm2_eth_evl / hm2_eth_xenomai? Both variants will be viable except the Makefile issue with components I am having right now.
> 
> I still don't like it. What are the alternatives, besides having three differently named components/drivers?

Except making linuxcnc-uspace always depend on libevl / libxenomai none that I am aware of. But I need to read in a bit more into how the dynamic linker works.

Compared to all the other GB's of dependency's linuxcnc needs, these are small, using 200kB together, but they would have to be supplied by the linuxcnc deb server. Separate deb's are nicer in my opinion.

> 
> > The builds with Xenomai will not be bit identical than the builds without due to there are many ifdefs in rtapi_app and also hm2_eth.
> 
> This is a problem. The packages have the same name and should therefore be identical when built. Having the same package name with different functionality is a disaster waiting to happen.
> 
> When there are differences, due to the conditional compilation, then the packages must carry different names and add a conflict so that you cannot install them together. The alternative is to split the package into more parts, but that may be even more problematic due to rtapi_app. Having yet another level of exec indirection is just making complex things even more complicated and I cannot recommend that (and may be prone to errors too).
> 

This can be solved by just removing the ifdef's. In rtapi/uspace_rtapi_main.cc, all ifdefs are to enable/disable the checks for a specific realtime backend, no dependency needed there. When the checks are always active, the behavior will be a bit different when you have a Xenomai kernel running but support is deactivated / support package not installed:

Now: Will start posix

Removed ifdefs: Will crash with liblinuxcnc-uspace-xenomai.so.0 not found.

But this can be improved. To variants: 
- Crash with a message to start a standard kernel or install support package
  - Give an environment variable to force a specific backend, someone might want to run a xenomai kernel and run posix LinuxCNC for whatever reason.
  - My preferred variant, but it seams many people like autotdetect.
- Just start posix non-rt variant (like before but also check for the existence of the support library, not only the kernel type)

Then there are the ifdef added by me in hal/drivers/mesa-hostmot2/hm2_eth.c. They can just be removed. If you set board_rtnet=evl and don't have it installed, a crash is probably fine. But I have to improve error message for this case.

Should be easy to do. I will look into it. Which variant do you prefer? Fail and give hints to resolve or automatically do something most people don't want to do?

> > I added a rip-xenomai stage which builds Xenomai3/4 to check for compile issues. I don't run the tests due to they don't test anything that is different in the xenomai build and there is no xenomai kernel running. If desired, I can add them.
> 
> At least the build should be tested if we are to support these in the long run, which we seem to do.

Tests can be easily enabled.

> 
> > As soon as a proper Debian package structure is in place, where linuxcnc-uspace will not depend on the lib's when they are active, one or also both variants can be built in the packaging stage by default without hurting anyone.
> 
> That will need some good thinking.
> 

I pushed a first variant for debian packages. But configure.ac needs an added `--with-xenomai` / `--with-xenomai-evl` to force the package config to be consistent with the configure result. Otherwise, the packaging will fail if you do:

`debian/configure xenomai xenomai-evl` and the libs are not installed.

I will do that next.

> > Do you know if someone actively uses Xenomai3?
> 
> No.
> 
> > Xenomai3 is stable and maintained by Siemens, while Xenomai4 is mostly maintained by Philippe Gerum and still somewhat work in progress. There where a few incompatible API changes in the ~6 months I am working on it.
> 
> Well, it seems there is some backing. Then it will not die just tomorrow ;-)
> 

Yes, the Xenomai lives since 2001 and is up to date with a 7.2 kernel. The community is also nice. Now there are even a few patches in from my side and a few pending... ;-)

> > Due to this and due to the additional code for full Xenomai3 support is not that much, it is probably worth keeping both variants at least for some time and also add Xenomai3 hm2_eth support.
> 
> Xenomai 3 would probably the go-to version as long as Xenomai 4 is in flux. At least, for those seeking (some kind of) stability.
> 

It depends. Both variants are probably less well tested just due to they are less often used as PREEMT_RT. I would just go for Xenomai4 and update it from time to time. If it works and you don't need always the latest kernel release, it should be fine.

Xenomai3 is more cumbersome to setup network. You basically need a custom start/stop script. Unload some kernel modules, load some others and so on. I will add this to de doc when xenomai3 hm2_eth gets in.

> > Meanwhile, LXRT (RTAI Userspace) support can probably be removed. It was fully broken until I fixed it and it needs a change in the support headers supplied by @NTULINUX to be able to build this variant. At least on my PC, it does not work well and results in full system freezes from time to time.
> 
> There is probably no one who uses it. I wouldn't mind it gone. Maybe others want to pitch in on removing LXRT support? Otherwise it is left to bitrot until RTAI gets the boot.

Ok, i can prepare a PR removing LXRT. RTAI is still used I think and it's also nice having it still in: It opens the door creating an in-kernel Xenomai version. It should be possible. Might be if I find time to spare, I will try this, sounds like a challenge...

hdiethelm:
> How many executables/libraries are affected by Xenomai{3,4} that alter the link behaviour/dependencies?

Existing: rtapi_app loads one of:
- liblinuxcnc-uspace-posix.so.0 -> linuxcnc-uspace.deb
- liblinuxcnc-uspace-xenomai.so.0 -> linuxcnc-uspace-xenomai.deb
- liblinuxcnc-uspace-xenomai-evl.so.0 -> linuxcnc-uspace-xenomai-evl.deb

New: hm2_eth loads zero to both:
- liblinuxcnc-hm2_eth_net_xenomai.so.0 -> linuxcnc-uspace-xenomai.deb
- liblinuxcnc-hm2_eth_net_evl.so.0 -> linuxcnc-uspace-xenomai-evl.deb

BsAtHome:
There are a few things to consider:
1. The same name .deb(s) that you build with or without Xenomai need to be identical. This is the principle of reproducible builds.
2. Standard builds of distributable .deb(s) can not depend on third party downloads. This ensures that the build is contained to the what the Debian distribution provides. This is a requirement or Debian cannot rebuild their own repo(s) and our package would be removed.
3.  No .deb(s) can add dependencies that are not provided by the standard Debian repository. Adding remote dependencies is problematic (see 2). We can put /additions/ in the LinuxCNC repository, but they need to be self-contained or we would be divergent from the general distribution, which is bad.

It means that we can only enable Xenomai in the normally distributed build when it is not interfering with the normal build and operation. Building the standard .deps with Xenomai support means that Xenomai must be effectively invisible and transparent for all practical purposes and intends.

Building Xenomai support libraries/.deb(s) can only be distributed via the LinuxCNC repository because, afaik, no Xenomai support/package is present in standard Debian. Doing this kind of distribution means to have procedures in place to update the LinuxCNC repo. Note that the RTAI build is not distributed via Debian. It is only present in the LinuxCNC repo. If we cannot "just" add the Xenomai support libraries/.deb(s) in a separate repo, then we need to supply a parallel package just for Xenomai and then it would be better to have a different name (and a conflict with the standard package) to ensure proper versioned install.

If all the `#ifdef` can be removed without repercussion, then they were probably not necessary. But then someone needs to explain why they were put there in the first place? Or do we need some reorganisation first?

hdiethelm:
> There are a few things to consider:
> 
>1. The same name .deb(s) that you build with or without Xenomai need to be identical. This is the principle of reproducible builds.
> 
> 2. Standard builds of distributable .deb(s) can not depend on third party downloads. This ensures that the build is contained to the what the Debian distribution provides. This is a requirement or Debian cannot rebuild their own repo(s) and our package would be removed.
> 
> 3. No .deb(s) can add dependencies that are not provided by the standard Debian repository. Adding remote dependencies is problematic (see 2). We can put /additions/ in the LinuxCNC repository, but they need to be self-contained or we would be divergent from the general distribution, which is bad.
> 
> 
> It means that we can only enable Xenomai in the normally distributed build when it is not interfering with the normal build and operation. Building the standard .deps with Xenomai support means that Xenomai must be effectively invisible and transparent for all practical purposes and intends.
> 
> Building Xenomai support libraries/.deb(s) can only be distributed via the LinuxCNC repository because, afaik, no Xenomai support/package is present in standard Debian. Doing this kind of distribution means to have procedures in place to update the LinuxCNC repo. Note that the RTAI build is not distributed via Debian. It is only present in the LinuxCNC repo. If we cannot "just" add the Xenomai support libraries/.deb(s) in a separate repo, then we need to supply a parallel package just for Xenomai and then it would be better to have a different name (and a conflict with the standard package) to ensure proper versioned install.

Hmm, does that mean that even trying to load a non existing library will be an issue?
Basically two options:
1. If no: Go forward with support library packages, linuxcnc-uspace will not suggest linuxcnc-uspace-xenomai or linuxcnc-uspace-xenomai-evl. But they will be available in the linuxcnc package repo some day. With no ifdef's, I think I can get the linuxcnc-uspace packages identical. You will get a warning or an error if you run LinuxCNC on a Xenomai kernel suggesting to install the support library's. No difference on a RT or standard kernel.
2. If yes: Create more full-fledged packages that contain all files, most of them identical, similar to RTAI: linuxcnc-uspace-xenomai / linuxcnc-uspace-xenomai-evl
    - In this case, the ifdef's should stay in, the packages will have a different name and (slightly) different content
    - This will take more space and longer to build due to you build linuxcnc essentially 3 times, but a non-issue with modern hardware.
    - No way to install all at the same time (this is really nice to test and compare different variants)

Both options are not available now, so I think one of these have to be implemented anyway. What do you think?

> 
> If all the `#ifdef` can be removed without repercussion, then they were probably not necessary. But then someone needs to explain why they were put there in the first place? Or do we need some reorganisation first?

They where added here:
https://github.com/LinuxCNC/linuxcnc/commit/46b3593ed096fd1781cda3ab184412cb99f5397b
https://github.com/LinuxCNC/linuxcnc/commit/0c7934bfdba1acb86df69f014a9d7b0e9fd6b88e
I would say they where never really needed except might be due to the Debian packaging issues you mentioned. Possibly, the intent was also to not touch the existing code while adding an experimental feature, but there is no real explanation visible anywhere in these two commits.

From my side, I just followed the existing pattern.

I added two commits that show 1. but I can also switch to the 2. variant, no issue.
- Get rid of ifdef's
- Improve error messages and add LINUXCNC_FORCE_REALTIME_TYPE

If I run LinuxCNC on a Xenomai kernel without the libs:
```
../scripts/latency-test
dlopen: liblinuxcnc-uspace-xenomai-evl.so.0: cannot open shared object file: No such file or directory
    Please install the matching support package:
    linuxcnc-uspace-xenomai or linuxcnc-uspace-xenomai-evl
    To manually select the realtime type, use: FORCE_REALTIME_TYPE
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Could not load rtapi dll
rtapi_app: recv_result 1 failed: Connection reset by peer
rtapi_app: failed to read from master
```

To force PREEMT_RT and circumvent the autodetection:
```
LINUXCNC_FORCE_REALTIME_TYPE=3 ../scripts/latency-test
Note: Using POSIX realtime
```

An alternative in this case would be to test for the library's in `detect_xenomai()` and return false if a Xenomai kernel is running but no support libraries are available. With matching warnings, this is also a possible way to go.

BsAtHome:
I was wondering... Why exactly do you need an _additional_ set of .so libraries to load? You already have the distinction in rtapi_app, which dlopen's the posix/xenomai libraries.

You can use these existing libraries to export an API for hm2_eth, which will automatically link to the loaded .so. This is also how hal_lib.so and hostmot2.so provides all the infrastructure to all the drivers.

Thus, you only need to define an API that is independent of the RT system. Then you export that API from the rtapi_app's loaded RT support system .so libraries. You then need to use an abstracted API in hm2_eth that will satisfy all RT systems.

This will also make it much easier to package. The stock linuxcnc_uspace supplies the posix library and can be built without any other dependencies. Then, you can build the Xenomai3/4 libraries _separately_ and package them accordingly. They will not conflict and the proper one will be loaded at runtime by rtapi_app. If it isn't available, then a message may be issued (running Xenomai RT without the specific RT library available).

hdiethelm:
> I was wondering... Why exactly do you need an _additional_ set of .so libraries to load? You already have the distinction in rtapi_app, which dlopen's the posix/xenomai libraries.
> 
> You can use these existing libraries to export an API for hm2_eth, which will automatically link to the loaded .so. This is also how hal_lib.so and hostmot2.so provides all the infrastructure to all the drivers.
> 
> Thus, you only need to define an API that is independent of the RT system. Then you export that API from the rtapi_app's loaded RT support system .so libraries. You then need to use an abstracted API in hm2_eth that will satisfy all RT systems.
> 
> This will also make it much easier to package. The stock linuxcnc_uspace supplies the posix library and can be built without any other dependencies. Then, you can build the Xenomai3/4 libraries _separately_ and package them accordingly. They will not conflict and the proper one will be loaded at runtime by rtapi_app. If it isn't available, then a message may be issued (running Xenomai RT without the specific RT library available).

I don't "need" them, there are always different options. Figuring out the best one is more difficult... ;-) The concept in this PR already does proper packing and with the removed ifdef's, linuxcnc-uspace should be binary identical. But I agree the second dlopen and the two additional libs are not that nice.

Two alternatives:
1. hm2_eth_xenomai / hm2_eth_evl linuxcnc components. These need to be loaded before hm2_eth similar like hostmot2 needs to be loaded before hm2_eth.
    - Disadvantages: 
      - More components
      - Needs entry in hal file, fails with link error if you forget it. Not loading hostmot2 shows: `hm2_eth: dlopen: hm2_eth.so: undefined symbol: hm2_register`
      - Makefile makes it hard to define custom C / LD flags for specific modules. I did not yet figure out how to do this.
    - Advantages: 
      - Follows the components concept
2. Add UDP functionality as rtapi_* functions, available always.
    - Disadvantages: 
      - Globally available in all components
    - Advantages: 
      - This would also make it way easier to add new components which need realtime UDP.

UDP functionality as rtapi_* functions was an intuition I tried to implement until this started to look quite complex. However, with https://github.com/LinuxCNC/linuxcnc/pull/4503, it should be doable again.

Before, this API would have had to follow a specific initialization order, needed only for Xenomai EVL.

It would look roughly:
```
int rtapi_udp_init(rtapi_udp_t *udp, rtapi_net_t type, const char *ip, const char *ifname)
int rtapi_udp_send(rtapi_udp_t *udp, const void *buffer, int len)
int rtapi_udp_recv(rtapi_udp_t *udp, void *buffer, int len, int recv_timeout_ns)
int rtapi_udp_close(rtapi_udp_t *udp)
```

All variants would support type = RTAPI_NET_POSIX while only Xenomai and Xenomai EVL would support type = RTAPI_NET_EVL or type = RTAPI_NET_XENOMAI.

BsAtHome:
Making infrastructure available to all components is not a bad thing (that is what hal_lib does too). You could even hook your ICMP ping and ethercat into it when designed carefully. That is something to think about. I'd seriously consider that an improvement when an abstract API can help more than one client/component.

It certainly removes the incentives for duplication. And, as you said, a second dlopen (in hm2_eth) is just not nice, which was also my initial objection. This is cleanly solved by putting it in rtapi_app's decision. 

Your hm2_eth_XYZ component option is quite awkward because it requires user interaction when you load this via loadrt. It is not compatible with any existing config.

hdiethelm:
> Making infrastructure available to all components is not a bad thing (that is what hal_lib does too). You could even hook your ICMP ping and ethercat into it when designed carefully. That is something to think about. I'd seriously consider that an improvement when an abstract API can help more than one client/component.

Possibly, then it would be more like rtapi_net_init() with socket options. Ping needs a special socket `IPPROTO_ICMP` for POSIX but a raw socket for EVL due to `IPPROTO_ICMP` is not supported there. The main reason there is not yet a EVL ping component: It needs a whole lot of code to create the raw Ethernet packet bit by bit... ;-) I have a working variant tough...

> 
> It certainly removes the incentives for duplication. And, as you said, a second dlopen (in hm2_eth) is just not nice, which was also my initial objection. This is cleanly solved by putting it in rtapi_app's decision.
> 
> Your hm2_eth_XYZ component option is quite awkward because it requires user interaction when you load this via loadrt. It is not compatible with any existing config.

Agreed. I also did not like this option.

So I will investigate option 2. However, https://github.com/LinuxCNC/linuxcnc/pull/4503 should be merged first. Meanwhile, I can re base this branch temporarily.

PR #4499, opened 2026-09-04
hal: Add halfileupdate to convert HAL configuration files

grandixximo:
@BsAtHome This is the .hal side of what we discussed in #4256, as a separate binary: `halfileupdate`.

It should be safe because it does not match strings. It reads the `loadrt` lines, works out which instances exist, and converts only names that resolve to one of those instances. `count=`, `names=` (the instance name is the user's, only the `loadrt` line changes) and components whose HAL prefix is not the module name, `mux-gen` and friends, all fall out of that instead of being special cases. Anything that does not resolve is left alone and reported.

Given an INI file it walks the `[HAL]` files in the order LinuxCNC runs them and follows `source`, which is the part that matters in practice: in stock configs the `loadrt` and the pins that use it are regularly in different files, `sim_vmc.hal` loads `conv_float_s32 count=3` and `toolchange_gray.hal` uses `conv-float-s32.2.in`. It also picks up HAL files behind a GUI's own key such as `CUSTOM_HALFILE`, and rewrites `HALCMD` lines in the INI itself. On a single .hal file it converts what that file loads and says to run the INI for the rest.

Signals, aliases and comments are never rewritten, though a comment still naming something that changed is reported. Names built with a substitution or in a haltcl loop are reported rather than guessed at. Library files are skipped. Diff by default, `-i` keeps a .bak.

Per your last comment: no table in `hal_lib`, no hooks on the halcmd error paths, no CI mode. The renames are plain tables in the tool, taken from your `wip_hal-types-and-isolation` branch: the eleven `conv_*` converters, the `mux_generic` and `demux_generic` pins, `demux` `sel-u32` and the `reset` pins. The four converters whose two sides collapse to one type, `conv_s32_s64` and friends, are gone with no replacement there, so their uses are reported instead of rewritten. I left `abs_s32`, `scaled_s32_sums` and the tristate pair alone since you have not renamed them; if that changes the tables get regenerated.

I have not written a removal release into it, since 2.12 or 3.0 is not settled; it needs remembering at some point.

Testing: all 326 stock config INI files dry-run clean, 65 lines changed, no false positives. The test asserts that the old names are gone and that the replacements are consistent, not how they are spelled, so regenerating the tables will not invalidate it.

### DIFF
--- a/src/hal/utils/halfileupdate.py
+++ b/src/hal/utils/halfileupdate.py
@@ -1,0 +1,931 @@
+#!/usr/bin/env python3
+#
+#    halfileupdate - update HAL configuration files after component,
+#                    pin and parameter renames
+#    Copyright 2026 Luca Toniolo
+#
+#    This program is free software; you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation; either version 2 of the License, or
+#    (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program; if not, write to the Free Software
+#    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# Killing off the 32-bit HAL types renames components, pins and
+# parameters whose name spells a type that no longer exists
+# ('conv_s32_float' -> 'conv_sint_real', 'mux-gen.00.in-s32-00' ->
+# 'mux-gen.00.in-sint-00').  Configurations using those names stop
+# working.  This tool rewrites them.
+#
+# Nothing is renamed by pattern-matching a name.  A name is rewritten
+# only when it resolves to an instance of a component that a 'loadrt'
+# in the processed files created, so 'names=' and 'count=' instances,
+# and components whose HAL prefix differs from the module name, are
+# handled without guessing.  A name that cannot be resolved is left
+# alone and reported.
+#
+# By default a unified diff is printed and nothing is written; --in-place
+# rewrites, keeping a .bak.
+#
+# The tool is a migration aid.  It carries the rename table itself, no
+# part of HAL knows about the old names, and both are meant to be
+# dropped once the release that renamed things is a few releases old.
+
+import argparse
+import difflib
+import os
+import re
+import sys
+import tempfile
+
+# ---------------------------------------------------------------------------
+# The renames.  The 32-bit HAL types are gone, so the type words in a name
+# follow the types that are left: bit -> bool, float -> real, s32 -> sint,
+# u32 -> uint.  Regenerate these tables when the rename set changes;
+# nothing else in HAL knows the old names.
+# ---------------------------------------------------------------------------
+
+# Renamed components (the loadrt/loadusr module name).  None means the
+# component is gone with no replacement: two old converters map to one new
+# one (s32 and s64 both become sint) and a converter whose two sides
+# collapse to the same type is not a conversion any more.  Uses of such a
+# component are reported and left unchanged.
+MODULE_RENAMES = {
+    'conv_bit_float':   'conv_bool_real',
+    'conv_bit_s32':     'conv_bool_sint',
+    'conv_bit_s64':     'conv_bool_sint',
+    'conv_bit_u32':     'conv_bool_uint',
+    'conv_bit_u64':     'conv_bool_uint',
+    'conv_float_s32':   'conv_real_sint',
+    'conv_float_s64':   'conv_real_sint',
+    'conv_float_u32':   'conv_real_uint',
+    'conv_float_u64':   'conv_real_uint',
+    'conv_s32_bit':     'conv_sint_bool',
+    'conv_s32_float':   'conv_sint_real',
+    'conv_s32_u32':     'conv_sint_uint',
+    'conv_s32_u64':     'conv_sint_uint',
+    'conv_s32_s64':     None,
+    'conv_s64_bit':     'conv_sint_bool',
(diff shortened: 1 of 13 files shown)


BsAtHome:
> The four converters whose two sides collapse to one type, conv_s32_s64 and friends, are gone with no replacement there, so their uses are reported instead of rewritten.

That is good. Both conv_s32_s64, conv_s64_s32, conv_u32_u64 and conv_u64_u32 are superfluous now.

> I left abs_s32, scaled_s32_sums and the tristate pair alone since you have not renamed them; if that changes the tables get regenerated.

I'm still working on them. The abs_s32 and abs_s64 will collapse into abs_sint. Anything else I find will also need to be added.

Also working on changing the components to replace si32/ui32 with sint/uint where ever possible.

src/hal/utils/halfileupdate.py
```diff
@@ -0,0 +1,913 @@
...
+FUNC_RENAMES = {
+}
+
+# Commands that carry HAL object names, and which of their arguments do.
+# Argument 0 is the command itself.  Signal names are never rewritten,
+# they are the user's own.
+PIN_ARGS = {
+    'linkps': (1,),         # linkps pin [signal]
+    'linksp': (2,),         # linksp signal pin
+    'linkpp': (1, 2),       # linkpp pin pin
```
BsAtHome: `linkpp` has been removed from halcmd. If you see this, then you should produce an error.
Both `linkps` and `linksp` are superseded and replaced by `net`. Probably need to warn the user.

src/hal/utils/halfileupdate.py
```diff
@@ -0,0 +1,913 @@
...
+                self.errors += 1
+                print("halfileupdate: %s" % e, file=sys.stderr)
+            return
+        tcl = path.endswith('.tcl')
+        if tcl and re.search(r'(?m)^\s*(proc|for|foreach|while|if)\b', text):
+            self.rep.at(path)
+            self.rep.note("haltcl file that builds HAL names in code; the "
+                          "plain commands are converted, a name a proc or a "
+                          "loop builds is not, and has to be converted by "
+                          "hand")
```
BsAtHome, line 766: This makes me wonder... Why should we continue to support Tcl based hal files? Is there a (very) pressing case for having them?

grandixximo: TCL other than innumerable parentheses, allows you to have multi model machines swappable configs with few INI changes per model, meaning a manufacturer could manage a single TCL, and have model only INI files, I am actually work in progress on moving my configs to TCL, but I'm too comfortable in Hal, and I find TCL less readable, Hal is easier to debug and hack new features on, so I'm conflicted about it, unfortunately I don't see other scripting languages coming to the rescue here...
Why wouldn't you want a way to script your Hal? I think is a valuable construct, just a painful language.

The biggest gripe is not being able to search a specific pin directly in the TCL file because it has been created from a variable, which is not listed anywhere, is that reason enough to drop the "scripted  Hal" construct all together?

BsAtHome: Sure, I get it why it is there ;-)
That said, not being able to grep your way to a pin is (very) annoying.

Anyway, there has been a continual push to move from Tcl to Python. That is my point... The Tcl support libraries are withering and Python would be the better choice in that light. Also seen in the light of our own updating of the python modules.

FWIW, I see a version 3 where TP is non-RT and all Tcl has been retired (rtai killed too; python modules in a namespace,...).

BsAtHome:
Do you also update the `newsig` lines?
`newsig sig.name bit|float|s32|u32|s64|u64` --> `newsig sig.name bool|real|sint|uint`

src/hal/utils/halfileupdate.py
BsAtHome: On second thought... warning on `linkps` and `linksp` will also affect many sim configs. Maybe we better should let that warning be.

andypugh:
Why is this needed in addition to the existing update_ini script?

src/hal/utils/halfileupdate.py
andypugh: I think that some of the shared files in the sample configs are Tcl HAL. 
To be honest I didn't support that change, I think that all the sample configs should be flat files, to be easy to modify and for new users to comprehend.  Yes that leads to a lot of redundancy, but so what?

BsAtHome:
> Why is this needed in addition to the existing update_ini script?

Factoring this out is a good thing because you may not hit all .hal files in the ini upgrade. There can be files located many places and a user may need to do some things more manually.

There is no problem of calling this from the ini update script to automate the 1.1->1.2 version upgrade. But when that fails, for whatever reason, you still want to offer some kind of automation for the hal files.

src/hal/utils/halfileupdate.py
BsAtHome: There already is a huge copy/paste redundancy in the sample configs. It is no problem as long as they are maintained. But that is the hard problem.

grandixximo:
Both done.

`newsig` lines now convert the type word, `newsig sig s32` to `newsig sig sint`, case-insensitive like halcmd, with the same map as the rest. It turns out to be the most common thing in a config by a distance: over the stock configs it takes the run from 65 changed lines to 201, and 148 of those are `newsig`.

`linkpp` now gets a warning that says the line will fail whatever its names are and has to become a `net`. The names on it are still converted so a hand rewrite lands on the right ones. I left `linkps` and `linksp` silent, per your second thought.

Tables get regenerated when `abs_sint` and the rest of your renames land; say the word when the set is settled and I will redo them in one go.

src/hal/utils/halfileupdate.py
grandixximo: Having default TCL configs in tree is IMO quite user unfriendly, TCL configs IMO should be something a more advanced user reaches for, rather than being forced in the introductory configs, although that is how I learned they exist in the first place, but then again if we are going all Python... This feels like a larger topic for the agenda...

src/hal/utils/halfileupdate.py
grandixximo: Done

grandixximo:
Draft till the renaming is settled...

PR #4513, opened 2026-09-10
Remove the RTAI-kernel-only hardware drivers

### LINKED ISSUE #4505: pluto_servo / pluto_step / pcl720: not built on uspace since 2014, docs links silently missing
While investigating missing links in the CI docs build (see #4471, `components_gen.adoc` lacks `pcl720`, `pluto_servo`, `pluto_step`), I found the root cause: these three driver comps have not been built on uspace since 2b0c9aa839 (June 2014), when the `BUILD_SYS=uspace` whitelist in `src/hal/components/Submakefile` was introduced with only `serport` (later also `mesa_7i65`, `mesa_uart`).

Since every modern install (preempt-rt) is a uspace build, these drivers are effectively unavailable to all current users. Only an RTAI/kbuild build produces them. Their man pages are also absent from the published docs (linuxcnc.org returns 404 for `pluto_servo.9.html` and `pcl720.9.html`), so the entries in `components.adoc` point to nothing.

The situation differs per driver:

- **pluto_servo / pluto_step**: no technical blocker. `pluto_common.h` has had a userspace I/O path (`sys/io.h`) since 2007, the comp compiles cleanly with uspace flags, and its dependency `hal_parport` builds on uspace. It was simply never whitelisted and never hardware-tested on uspace.
- **pcl720**: genuinely kernel-only. It includes `<asm/io.h>` unconditionally and uses kernel types, so a uspace build needs an actual port.

Questions for direction:

1. pluto_servo/pluto_step: whitelist for uspace (compile-tested, needs someone with Pluto-P hardware to verify), or leave as-is?
2. pcl720: worth a port, or formally deprecate?
3. Independently: should the docs note that these drivers are only available on kbuild/RTAI builds? Right now the docs present them as normal components.

Side note: `make clean-manpages` only removes `$(GENERATED_MANPAGES)`, which on uspace excludes these drivers' man pages, so stale `pluto_servo.9` etc. survive forever in old build trees and confuse the generated component list (that is why a local Trixie build showed links the CI build did not).

grandixximo:
Reshaped after the weekly-meeting decision: remove the drivers outright instead of deprecating. Zero users found in the poll, no testers, Vital Systems has no Motenc cards left.

Two commits:

1. **build: wipe generated manpage tree on clean.** `clean-manpages` removed manpages by the BUILD_SYS-dependent `$(GENERATED_MANPAGES)` list, so kbuild-only driver manpages survived forever on uspace trees and confused gen_complist.py (the local-vs-CI divergence hdiethelm hit in #4471). Everything under docs/build/man is generated, so wipe the tree.
2. **Remove the RTAI-kernel-only hardware drivers.** hal_stg, hal_motenc, hal_vti, hal_evoreg, hal_tiro, hal_ax5214h, opto_ac5, pci_8255, pcl720, pluto_servo, pluto_step, all excluded from uspace builds since 2014 and serving early-2000s ISA/PCI/parport cards. Also removes hal_skeleton (kbuild-era code template), rbf2h.py (only used by the pluto firmware rules), the pluto firmware trees (unbuildable without an ancient Quartus), the driver doc pages and man pages, the po4a master entries, and the components.adoc rows. Sample configs for these cards are removed too (by_interface/pluto, servotogo, vigilant, vitalsystems, by_machine/boss, attic/demo_mazak).

Verified: full build plus htmldocs pass, components_gen no longer mentions any of the removed drivers, repo-wide grep finds only historical comments.

Closes #4505.


src/hal/drivers/pcl720.comp
```diff
@@ -1,5 +1,8 @@
 component pcl720 """Driver for the Advantech PCL 720 card.""";
-description """This driver supports the Advantech PCL720 ISA card. It might
+description """
+.SH DEPRECATED
+This driver is deprecated and will be removed in a future release. It is only built for RTAI kernels; it is not available in the standard preempt-rt (uspace) builds.
```
BsAtHome: Doc is in asciidoc format... Not troff.

(I know, some files never got converted properly, but no need to add troff to asciidoc)

src/hal/drivers/pluto_servo.comp
```diff
@@ -17,6 +17,8 @@
 component pluto_servo """Hardware driver and firmware for the Pluto-P parallel-port FPGA, for use with servo machines.""";
 
 description """
+.SH DEPRECATED
+This driver is deprecated and will be removed in a future release. It is only built for RTAI kernels; it is not available in the standard preempt-rt (uspace) builds.
```
BsAtHome: Same troff/asciidoc problem

src/hal/drivers/pluto_step.comp
```diff
@@ -62,6 +62,8 @@ may give improved step regularity at low step speeds.
 .RE""";
 
 description """
+.SH DEPRECATED
+This driver is deprecated and will be removed in a future release. It is only built for RTAI kernels; it is not available in the standard preempt-rt (uspace) builds.
```
BsAtHome: And more troff/asciidoc

docs/src/drivers/ax5214h.adoc
```diff
@@ -4,6 +4,12 @@
 [[cha:ax5214-driver]]
 = AX5214H Driver
 
+WARNING: This driver is deprecated and will be removed in a future
+release. It is only built for RTAI kernels; it is not available in the
+standard preempt-rt (uspace) builds. If you still use this hardware,
+let us know on the LinuxCNC forum or the emc-developers mailing list,
+see https://github.com/LinuxCNC/linuxcnc/issues/4505.
```
BsAtHome: Shouldn't these warnings be added as admonitions?
```asciidoc
[WARNING]
.Deprecation warning
====
This driver is deprecated and will be...
====
```

also other places.

docs/src/hal/components.adoc
```diff
@@ -116,6 +116,13 @@ To search in the man pages, use the UNIX tool `apropos`.
 | hal_ppmc | Pico Systems <<cha:pico-drivers,driver>> for analog servo, PWM and Stepper controller ||
 | link:../man/man9/hal_bb_gpio.9.html[hal_bb_gpio] |Driver for BeagleBone GPIO pins ||
 | link:../man/man9/hal_parport.9.html[hal_parport] |Realtime HAL component to communicate with one or more PC parallel ports ||
+| hal_ax5214h |Axiom AX5214H <<cha:ax5214-driver,driver>> for the 48 channel ISA digital I/O board (deprecated, RTAI kernel builds only) ||
+| hal_evoreg |Driver for the Siemens EVOREG motion control board (deprecated, RTAI kernel builds only) ||
+| hal_motenc |Vital Systems <<cha:motenc,driver>> for the Motenc-100 and Motenc-LITE servo boards (deprecated, RTAI kernel builds only) ||
+| hal_stg |Servo-To-Go <<cha:servo-to-go,driver>> for the ISA motion control card (deprecated, RTAI kernel builds only) ||
+| hal_vti |Driver for the Vigilant VTI card (deprecated, RTAI kernel builds only) ||
+| pci_8255 |Driver for PCI cards based on the 8255 digital I/O chip (deprecated, RTAI kernel builds only) ||
+| pcl720 |Driver for the Advantech PCL720 ISA digital I/O card (deprecated, RTAI kernel builds only) ||
```
BsAtHome: Why add them now? They haven't been here since...forever. Do these require an extra mention?

Also more below.

debian/changelog
```diff
@@ -6,6 +6,12 @@ linuxcnc (1:2.10.0~pre1) UNRELEASED; urgency=medium
   [ andypugh ]
   * Add a finite-jerk trajectory planner
 
+  [ Luca Toniolo ]
+  * Deprecate the RTAI-kernel-only hardware drivers (hal_stg, hal_motenc,
+    hal_vti, hal_evoreg, hal_ax5214h, opto_ac5, pci_8255, pcl720,
+    pluto_servo, pluto_step). They are not built for uspace and will be
+    removed in a future release.
+
```
BsAtHome, line 14: This is quite a lone addition. There is a list ...very... long that should have been  added here. What reasoning is behind this that it must now be added to the Debian changelog?

src/hal/drivers/pcl720.comp
grandixximo: Fixed in c060966aaf, plain text note now, no troff.

src/hal/drivers/pluto_servo.comp
grandixximo: Fixed in c060966aaf.

src/hal/drivers/pluto_step.comp
grandixximo: Fixed in c060966aaf.

docs/src/drivers/ax5214h.adoc
grandixximo: Admonition blocks everywhere in c060966aaf, thanks for the example.

docs/src/hal/components.adoc
grandixximo: Fair point, dropped in c060966aaf. They were never listed and adding rows for drivers that cannot even be linked just advertises dead hardware. The three existing rows (opto_ac5, pluto_servo, pluto_step) keep the deprecation marker.

debian/changelog
grandixximo: Kept, because this is exactly the audience that matters for a deprecation. Users do not read driver doc pages until something breaks, but people scanning the release notes do read the changelog. A deprecation that only lives in the docs of the deprecated thing itself is invisible until it is too late. If the changelog is the wrong channel, where should release-level deprecations go?

BsAtHome: > but people scanning the release notes do read the changelog

Optimist?

grandixximo: Just trying to be as loud as I can about the deprecation, I think is better to have it than not?

BsAtHome:
Remove the drivers completely.

src/hal/drivers/Submakefile
```diff
@@ -1,41 +0,0 @@
-hal/drivers/pluto_servo_rbf.h: hal/drivers/pluto_servo_firmware/pluto_servo.rbf hal/drivers/rbf2h.py
-	$(PYTHON) hal/drivers/rbf2h.py -g \
-    -c "This is a component of pluto_servo, a PWM servo driver and" \
-    -c "quadrature counter for emc2" \
-    -c "Copyright (C) 2006, 2007 Jeff Epler" \
-	    $< > $@
-
```
BsAtHome: This Submakefile is now empty. Why not remove it?

BsAtHome:
Why move the sample configs? They are no longer necessary and not used in any meaningful way. The "Attic" is an old CVS construct. Why do we need it, at all?

grandixximo:
Wrong call, fixed in dd394d90d0: the configs are deleted outright, demo_mazak included.

